PR #172, opened 2015-01-15
Introduce new flag: KEY_FLAGS

manuelm:
Bitwise disjunction of flags, which don't require one or more values.
Recommended way to set multiple flags. Overrides previously defined flags.

### DIFF
--- a/examples/keyNew.c
+++ b/examples/keyNew.c
@@ -86,6 +86,18 @@ keyDel(k);
 
 }{
 
+//! [With Flags]
+Key *k=keyNew("user/tmp/ex3",
+	KEY_FLAGS, KEY_BINARY | KEY_CASCADING_NAME, // flags
+	KEY_SIZE, 7,			// assume binary length 7
+	KEY_VALUE, "some data",		// value that will be truncated in 7 bytes
+	KEY_END);			// end of args
+//! [With Flags]
+printf ("%.7s\n", (char*)keyValue(k));
+keyDel(k);
+
+}{
+
 //! [With Everything]
 Key *k=keyNew("user/tmp/ex4",
 	KEY_BINARY,			// key type
--- a/src/libelektra/keyhelpers.c
+++ b/src/libelektra/keyhelpers.c
@@ -345,108 +345,111 @@ int keyInit(Key *key)
  * initialized, but really empty, object
  * @param va the variadic argument list
  */
-void keyVInit (Key *key, const char *name, va_list va)
+void keyVInit(Key *key, const char *name, va_list va)
 {
 	keyswitch_t action = 0;
+	size_t value_size = 0;
 	void *value = 0;
-	int valueSizeChanged = 0;
-	size_t valueSize = 0;
+	void (*func) (void) = 0;
+	uint64_t flags = 0;
 	char *owner = 0;
-	option_t nameOptions = 0;
-	void (*p) (void) = 0;
 
 	if (!key) return;
 
 	if (name) {
-		action=va_arg(va, keyswitch_t);
-		while (action) {
+		while ((action = va_arg(va, keyswitch_t))) {
 			switch (action) {
+				/* flags with an argument */
 				case KEY_SIZE:
-					valueSize=va_arg(va, size_t);
-					valueSizeChanged = 1;
-					break;
-				case KEY_BINARY:
-					keySetMeta (key, "binary", "");
+					value_size = va_arg(va, size_t);
 					break;
 				case KEY_VALUE:
 					value = va_arg(va, void *);
-					if (valueSizeChanged && keyIsBinary(key))
-					{
-						keySetBinary(key,value, valueSize);
-					} else if (keyIsBinary(key)) {
-						valueSize = elektraStrLen (value);
-						keySetBinary(key,value, valueSize);
-					} else {
-						keySetString(key,value);
-					}
+					if (value_size && keyIsBinary(key))
+						keySetBinary(key, value, value_size);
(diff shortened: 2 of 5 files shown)


manuelm:
```
The following tests FAILED:
	  4 - testmod_ccode (Failed)
	  6 - testmod_hexcode (Failed)
	 73 - test_key (Failed)

/home/jenkins/workspace/workspace/elektra-mergerequests/src/plugins/ccode/testmod_ccode.c:73: error in test_decode: string not correctly encoded

/home/jenkins/workspace/workspace/elektra-mergerequests/src/plugins/hexcode/testmod_hexcode.c:62: error in test_decode: string not correctly encoded

test_keyFixedNew: key value user/hello is not equal user/hello
/home/jenkins/workspace/workspace/elektra-mergerequests/tests/ctest/test_key.c:1143: error in test_keyFixedNew: Compare key "mmk1" with "mmk2" failed, too many meta keys found (k1 < k2)
```

what's wrong with these tests?

examples/keyNew.c
```diff
@@ -86,6 +86,18 @@ keyDel(k);
 
 }{
 
+//! [With Flags]
+Key *k=keyNew("user/tmp/ex3",
+	KEY_FLAGS, KEY_BINARY & KEY_CASCADING_NAME, // flags
```
markus2330: &?

src/libelektra/keyhelpers.c
```diff
@@ -345,108 +345,113 @@ int keyInit(Key *key)
...
 					break;
-				case KEY_MODE:
-					/* Theoretically this should be mode_t, but prefer using
-					   int to avoid troubles when sizeof(mode_t)!=sizeof(int)
-					 */
-					keySetMode(key,va_arg(va, int));
+
+				/* flags without an argument */
+				case KEY_FLAGS:
+					flags = va_arg(va, uint64_t);
```
markus2330: what about flags |=?

src/libelektra/keyhelpers.c
```diff
@@ -345,108 +345,113 @@ int keyInit(Key *key)
...
+						flags |= action;
+					if (test_bit(flags, KEY_BINARY))
+						keySetMeta(key, "binary", "");
+					if (test_bit(flags, KEY_LOCK_NAME))
+						keyLock(key, KEY_LOCK_NAME);
+					if (test_bit(flags, KEY_LOCK_VALUE))
+						keyLock(key, KEY_LOCK_VALUE);
+					if (test_bit(flags, KEY_LOCK_META))
+						keyLock(key, KEY_LOCK_META);
 					break;
```
markus2330: shouldn't this block be at the very end similar to keySetName?

markus2330:
tests: a new test case should be written to exactly reproduce the change of behaviour, but it seems to be related to KEY_SIZE or KEY_VALUE.

additionally, some testcases should be added to test KEY_FLAGS

examples/keyNew.c
manuelm: nice catch

src/libelektra/keyhelpers.c
manuelm: Questionable, this would prevent overriding. Since this currently works: keyNew(..., KEY_NAME, "user(foo", KEY_NAME, "user/bar"), I decided to make KEY_FLAGS behave the same.

src/libelektra/keyhelpers.c
manuelm: This is related to above, with the exception it's behaving the other way round ;) - So you cannot override/unset them anymore.

All in all I don't care much as I think adding the same key flag twice is undefined behavior anyway.

manuelm:
Tests are fixed. However I think the current behavior of KEY_VALUE with a KEY_SIZE >0 and key data being string is wrong or at least not intuitive.

Take https://github.com/ElektraInitiative/libelektra/blob/master/tests/ctest/test_key.c#L1142: You can change the value of KEY_SIZE to any value you like. Because the key data is a string the KEY_SIZE isn't used at all. Only elektraStrLen(key_value) is what's used.

markus2330:
Hello,

you wrote:
> Tests are fixed.

Thank you.

> However I think the current behavior of KEY_VALUE with a
> KEY_SIZE >0 and key data being string is wrong or at least not intuitive.

Is there any reason to use KEY_SIZE for non-binary?

The more general problem with ksNew is that there is no error reporting even 
when internal calls fail or arguments are inconsistent. Another pitfall are 
invalid key names.

Please write down such findings in doc/todo/FUTURE. Hopefully symbol versioning 
is there soon, then we can start fixing these things ;)

> Take
> https://github.com/ElektraInitiative/libelektra/blob/master/tests/ctest/te
> st_key.c#L1142: You can change the value of KEY_SIZE to any value you like.
> Because the key data is a string the KEY_SIZE isn't used at all. Only
> elektraStrLen(key_value) is what's used.

Yeah, its weird. I do not think that was the indent of the test.

best regards

src/libelektra/keyhelpers.c
```diff
@@ -345,108 +345,113 @@ int keyInit(Key *key)
...
-	size_t valueSize = 0;
+	void (*func) (void) = 0;
+	uint64_t flags = 0;
 	char *owner = 0;
-	option_t nameOptions = 0;
-	void (*p) (void) = 0;
 
-	if (!key) return;
+	if (!key)
+		return;
```
markus2330: use one-liner

manuelm:
everything fixed + rebased. ready to merge

Issue #138, opened 2014-11-06
ini plugin makes hierarchies flat


markus2330:
E.g. I have:

kdb ls system/passwd

    system/passwd/yacy/uid
    system/passwd/zeroinst
    system/passwd/zeroinst/gid

Actual outcome (excerpt):
kdb export system/passwd ini

    ...
    shell = /bin/false
    uid = 114
    yacy = 
    gid = 265
    home = /var/lib/yacy
    ...

Makes the hierachy flat.

Expected outcome:

    [yacy]
    uid=..
    [zeroinst]
    gid=..

fberlakovich:
What is the planned release date for 0.8.11?

markus2330:
Hello,

You wrote:
> What is the planned release date for 0.8.11?

Asap, hopefully before the 5th.

fberlakovich:
First of all: sorry for the late response. I wrote an answer 2 days ago, but obviously something went wrong.

> If not its straight forward: keyDirectBelow parentKey are sections, others are values.

This is also problematic because as far as I know INI may contain key value pairs without a section. E.g.

```
nosectionvalue = value
[section1]
sectionvalue = value1
```
The value could even be empty:
``` 
nosectionvalue = 
[section1]
sectionvalue = value1
``` 
So even using the value as a differentiator is not possible. We could simply reject parsing such INI files, but then we would have yet another incomplete INI parser implementation :(. Unfortunately INI is hardly (if at all) standardized. Another possibility would be to make the behaviour configurable by allowing the user to specify section strategies (e.g. all keys below the parent key are treated as sections). However, this increases the configuration burden and would also be a fair amount of work. Which option do you like best?

markus2330:
Hello,

you wrote:
> This is also problematic because as far as I know INI may contain key value
> pairs without a section.

The plugin "ni" solves this problem that keys without value are sections, with 
value are key=value, e.g.:

```
abc = value
abc/x = value1
```

would be:


```
abc = value
[abc]
x = value1
```

w/o `abc = value` would be without first line.

> The value could even be empty:
>
> So even using the value as a differentiator is not possible. We could
> simply reject parsing such INI files, but then we would have yet another
> incomplete INI parser implementation :(. Unfortunately INI is hardly (if
> at all) standardized. Another possibility would be to make the behaviour
> configurable by allowing the user to specify section strategies (e.g. all
> keys below the parent key are treated as sections). However, this
> increases the configuration burden and would also be a fair amount of
[...]

PR #178, opened 2015-01-22
Ini hierarchies

fberlakovich:
A first proposal of how the improved ini section handling discussed in #138 could look like. The plugin behaves as follows:
* keys with a NULL value (i.e. binary keys with value 0) are treated as sections
* all other keys are treated as plain ini keys
* if the autosection option exists, the plugin will automatically create a section entry on write for keys with a level greater than 1 relative to the parent key (see the autosection testcase for a illustrative example)

If you like it this way, I will update the README accordingly. Additionally an option would be thinkable that prevents uses `keyChangeName` to directly use the names of ini keys / sections (see #138) for an example. If you have any further ideas on how to make the the use of the plugin more intuitive, please don't hesitate to ask.

### DIFF
--- a/src/plugins/ini/ini.c
+++ b/src/plugins/ini/ini.c
@@ -41,6 +41,10 @@ static void flushCollectedComment (CallbackHandle *handle, Key *key)
 	}
 }
 
+// TODO defined privately in internal.c, API break possible.
+// Might consider moving this to the public API as it might be used by more plugins
+size_t elektraUnescapeKeyName(const char *source, char *dest);
+
 // TODO: this is very similar to elektraKeyAppendMetaLine in keytometa
 static int elektraKeyAppendLine (Key *target, const char *line)
 {
@@ -279,7 +283,7 @@ static short isSectionKey(Key *key)
 	return keyIsBinary(key) && !keyValue(key);
 }
 
-static const char *getIniName(KeySet *keys, Key *parent, Key *key)
+static char *getIniName(KeySet *keys, Key *parent, Key *key)
 {
 	cursor_t currentCursor = ksGetCursor(keys);
 
@@ -292,7 +296,7 @@ static const char *getIniName(KeySet *keys, Key *parent, Key *key)
 
 		if (!keyCmp(temp, parent))
 		{
-			// we reached the parent key, there won't be any more section keys
+			/* we reached the parent key, there won't be any more section keys */
 			section = parent;
 			break;
 		}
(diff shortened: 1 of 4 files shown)


src/plugins/ini/ini.c
```diff
@@ -16,6 +16,7 @@
 #include <stdlib.h>
 #include <kdberrors.h>
 #include <kdbextension.h>
+#include <kdbproposal.h>
```
markus2330: Please add reason why it is included, so that it can be removed later

src/plugins/ini/ini.c
```diff
@@ -273,6 +287,56 @@ void writeMultilineKey(Key *key, FILE *fh)
...
+			break;
+		}
+
+		section = ksLookup(keys, temp, KDB_O_NONE);
+	} while(!isSectionKey(section));
+
+	ksSetCursor(keys, currentCursor);
+	keyDel(temp);
+
+	char *buffer = elektraMalloc(keyGetNameSize(key));
```
markus2330: please comment that new allocated memory is returned

src/plugins/ini/testmod_ini.c
```diff
@@ -70,14 +70,18 @@ static void test_plainIniWrite(char *fileName)
 			keyNew ("user/tests/ini-write/nosectionkey",
 					KEY_VALUE, "nosectionvalue",
 					KEY_END),
-			keyNew ("user/tests/ini-write/section1", KEY_DIR, KEY_END),
+			keyNew ("user/tests/ini-write/section1",
+					KEY_BINARY,
```
markus2330: Please write a test case for this KEY_BINARY feature + add to docu

markus2330:
Looks good, thank you!

markus2330:
Please also update README.md

src/plugins/ini/testmod_ini.c
fberlakovich: As far as I can tell the KEY_BINARY flag is already part of the keyNew documentation and the feature is also tested in testabi_key.c

markus2330: Felix Berlakovich wrote:
> As far as I can tell the KEY_BINARY flag is already part of the keyNew
> documentation and the feature is also tested in testabi_key.c

Fixed in 79868f66088a136340e469520a4554313e1ef4dc

markus2330:
Please rebase, a large change was merged.

Issue #177, opened 2015-01-22
Header files of libtool installed incorrectly


markus2330:
The header files related to merging are located in a subfolder "merging". However, they are not installed in such a subdirectory. So trying to include them fails when Elektra is installed.

I would propose to not add them in the public API at all, but instead only provide the simple new high-level API we discussed.

Additionally, files should ideally be prefixed with kdb, because elektra enforces the user to put /usr/include/elektra into the include path (otherwise recursive inclusion fails). So I would suggest kdbmerge.h as new name for the high level merge API.

fberlakovich:
Should the private headers (e.g. automergestrategy.hpp) also be renamed (e.g. kdbautomergestrategy.hpp)? Am I right that making those headers private simply means not installing them via CMake?

In addition I would be glad to make them public again somewhere in the future when the merge API is more mature. The whole framewwork was designed from ground up to be highly flexible and modular and removing this flexibility by hiding the building blocks hurts. However, I totally understand that the framework is currently hard to use and the headers won't be of any use at the moment.

markus2330:
Hello,

you wrote:
> Should the private headers (e.g. automergestrategy.hpp) also be renamed
> (e.g. kdbautomergestrategy.hpp)? Am I right that making those headers
> private simply means not installing them via CMake?

Not installing it is the most private way. kdbprivate.h is only installed, 
because the data structures are needed by debuggers.

> In addition I would be glad to make them public again somewhere in the
> future when the merge API is more mature. The whole framewwork was
> designed from ground up to be highly flexible and modular and removing
> this flexibility by hiding the building blocks hurts. However, I totally
> understand that the framework is currently hard to use and the headers
> won't be of any use at the moment.

I would be glad to have that API with more flexibility once the issues are 
resolved. Having two different APIs for the same thing is also not ideal, thus 
hiding the other is the best thing at the moment (also considering that hiding 
is the least effort). Do you agree?

best regards, Markus

fberlakovich:
> Having two different APIs for the same thing is also not ideal, thus hiding the other is the best thing at the moment (also considering that hiding is the least effort). Do you agree?

Yes, absolutely.

markus2330:
I installed all files because kdb.h is the only file that guarantees compatibility anyway.

PR #180, opened 2015-01-25
Merging simplification

fberlakovich:
This changeset introduces merge configurations. They are simply combinations of strategies and configure a three way merger instance. As can be seen in the example this drastically eases the use of the merge framework. If you like it this way, I will complement the documentation.

Could you please give me some hints on how to seperate the headers into private and public headers? What would go into kdbmerge.h? The idea would be to make only the configurations public, while the more fine grained strategies stay private.


markus2330:
start build

fberlakovich:
This pull request will also deal with #177 once finished

src/tools/kdb/cmdline.cpp
```diff
@@ -148,21 +148,14 @@ Cmdline::Cmdline (int argc,
...
-			"                         theirs        .. use always their version in case of conflict\n"
-			"                         base          .. use always the base version in case of conflict\n"
-			"                         newkey        .. merge just those keys added by one side only\n"
-			"                         ourvalue      .. use our value if theirs is unmodified\n"
-			"                         theirvalue    .. use their value if ours is unmodified\n"
+			"                         preserve      .. automerge only those keys where just one side deviates from base (default)\n"
+			"                         ours          .. like preserve, but in case of conflict use our version\n"
+			"                         theirs        .. like preserve, but in case of conflict use their version\n"
+			"                         cut           .. primarily used for import. removes existing keys below the import point and always takes the imported version\n"
+			"                         import        .. primarily used for import. preserves existing keys if they do not existing in the imported keyset. in all other\n"
```
markus2330: line too long, spelling: "they do not exist"

src/tools/kdb/import.cpp
```diff
@@ -63,16 +63,9 @@ int ImportCommand::execute(Cmdline const& cl)
 
 	KeySet base;
 
-	// TODO: this should not be neccessary, but
-	// applying threeway strategies to a twoway merge is hard
-	base = KeySet();
+	base = existingKeys;
```
markus2330: useless copy: directly use existingKeys (you may rename it) or use KeySet const& base

src/tools/kdb/import.cpp
```diff
@@ -63,16 +63,9 @@ int ImportCommand::execute(Cmdline const& cl)
...
 
-	// TODO: for now we have to position this strategy manually
-	// to avoid meta information loss
-	MetaMergeStrategy metaStrategy(merger);
-	merger.addConflictStrategy(&metaStrategy);
-
-	helper.parseStrategies (cl, merger);
+	helper.configureMerger (cl, merger);
```
markus2330: Is this really the same as for merge.cpp?

markus2330:
Please also write docu and release notes for that change.

src/tools/kdb/import.cpp
fberlakovich: To be honest I didn't read into all the import stuff again when patching this part. However, the tests still pass, so I guess it should be ok. It was the same as merge.cpp even before the merge configuration patch, wasn't it?

markus2330: Felix Berlakovich wrote:
> To be honest I didn't read into all the import stuff again when patching
> this part. However, the tests still pass, so I guess it should be ok. It

I am not aware of any tests that check if "kdb import" merges correctly.

> was the same as merge.cpp even before the merge configuration patch,
> wasn't it?

You told me about the differences of two-way merging which made it hard to 
implement "kdb import" ;)

fberlakovich: > I am not aware of any tests that check if "kdb import" merges correctly.

check_import.sh should do this

> You told me about the differences of two-way merging which made it hard to implement "kdb import" ;)

Yes, but the differences should lie in the used strategies. For the two way stuff I had to implement a few strange strategies rarely used in any other scenario. But these strange strategies should be bundled in the import specific configurations (i.e. import and cut)

markus2330:
Please rebase, a large change was merged.

PR #181, opened 2015-01-27
fix ambiguity of KeySet(alloc, ...) constructor

### LINKED ISSUE #42: Ambiguity in the C++ KeySet varargs constructors
Currently the C++ bindings of KeySet provides, among the others, two contructors with varargs:
```c++
	inline explicit KeySet(size_t alloc, va_list ap);
	inline explicit KeySet(size_t alloc, ...);
```
the problem is that on some architectures `va_list` is represented as `void*` as public type, which makes the two constructor ambiguous when called with `alloc` greater than 0 and no additional arguments. In such case, the `va_list` version gets used, causing `ksVNew` to try to read arguments from a null (or garbage) `ap`.

The following patch for `src/bindings/cpp/tests/testcpp_ks.cpp`
```diff
@@ -821,6 +821,18 @@ void test_duplicate()
        // ks4.toStream(stdout, 0);
 }

+void test_alloc_no_keys()
+{
+       cout << "testing test_alloc_no_keys" << endl;
+
+       KeySet ks (5,
+               KS_END);
+       succeed_if (ks.size() == 5, "size not correct");
+
+       KeySet ks2 (ks);
+       succeed_if (ks2.size() == 5, "size not correct");
+}
+

 int main()
 {
@@ -844,6 +856,7 @@ int main()
        test_ksrelease();
        test_lookuppop();
        test_duplicate();
+       test_alloc_no_keys();

        cout << endl;
        cout << "test_key RESULTS: " << nbTest << " test(s) done. " << nbError << " error(s)." << endl;
```
is enough to trigger the issue on e.g. mips(el) or hppa. Currently this is exposed by the lua and python bindings, since all do:
```c++
  KeySet(size_t alloc) {
    return new kdb::KeySet(alloc, KS_END);
  }
```
In those it can be easily worked around by adding the `KS_END` marker a second time (making them really as vararg), but of course it wouldn't fix the issue with the C++ contructors of KeySet.

manuelm:
Make the KeySet(alloc, va_list) constructor available as KeySet::fromVA instead.
Usage:
  KeySet ks;
  ks.fromVA(...);

Additional mark the constructor as deprecated and make it available on x86
and x64 platforms only. On this platforms the ambiguity is no problem.

Fixes #42

### DIFF
--- a/src/bindings/cpp/include/keyset.hpp
+++ b/src/bindings/cpp/include/keyset.hpp
@@ -31,7 +31,9 @@ public:
 	inline KeySet(ckdb::KeySet *k);
 	inline KeySet(const KeySet &other);
 
-	inline explicit KeySet(size_t alloc, va_list ap);
+#if defined(__amd64__) || defined(__x86_64__) || defined(__i386__)
+	inline explicit KeySet(size_t alloc, va_list va)  __attribute__((deprecated("use 'KeySet::fromVA' instead!")));
+#endif
 	inline explicit KeySet(size_t alloc, ...);
 
 	inline ~KeySet ();
@@ -40,6 +42,7 @@ public:
 
 	ckdb::KeySet * getKeySet() const;
 	void setKeySet(ckdb::KeySet * k);
+	void fromVA(size_t alloc, va_list va);
 
 	KeySet& operator=(KeySet const& other);
 
@@ -387,14 +390,17 @@ inline KeySet::KeySet (const KeySet &other)
  * @brief Create a new keyset.
  *
  * @param alloc minimum number of keys to allocate
- * @param ap variable arguments list
+ * @param va variable arguments list
  *
  * @copydoc ksVNew
+ * @deprecated use fromVA instead!
  */
-inline KeySet::KeySet (size_t alloc, va_list av)
+#if defined(__amd64__) || defined(__x86_64__) || defined(__i386__)
+inline KeySet::KeySet (size_t alloc, va_list va)
 {
-	ks = ckdb::ksVNew (alloc, av);
+	ks = ckdb::ksVNew (alloc, va);
 }
+#endif
 
 /**
  * @brief Create a new keyset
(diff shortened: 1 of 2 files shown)


src/bindings/cpp/include/keyset.hpp
```diff
@@ -31,7 +31,9 @@ class KeySet
 	inline KeySet(ckdb::KeySet *k);
 	inline KeySet(const KeySet &other);
 
-	inline explicit KeySet(size_t alloc, va_list ap);
+#if defined(__amd64__) || defined(__x86_64__) || defined(__i386__)
```
markus2330: obviously thats not exactly the list that do not have va_list defined in a way that its not a pointer

actually, maybe its better to only provide fromVA then

manuelm: I have no problem with removing the va constructor. Shall I remove it right away and rebase?

markus2330: Hello,

manuel wrote:
> I have no problem with removing the va constructor. Shall I remove it right
> away and rebase?

Actually, I still prefer to distinguish by useless argument. C++11 also uses 
this technique, see std::piecewise_construct_t. The advantage is that it does 
not impose any significant runtime overhead, opposed to creating and 
destroying a KeySet that would happen with fromVA. When a static object is 
provided, its also easy to use.

But if you think that from the bindings perspective fromVA is the better 
solution you can rebase right away ;)

best regards,

manuelm: So e.g. `inline explicit KeySet(int unused, size_t alloc, va_list va)` is what you prefer?

I think the va_list-constructor won't be used much. Bindings are unable to use it. They are even unable to use the variadic arguments constructor. So right now we don't use it all.

markus2330: See 55ae0b7ce62019048c323a06ad3bfc2809f80871

I also expect that it will be used rarely, but it should work properly when needed.

PR #191, opened 2015-03-02
qt-gui 0.0.5

0003088:
New features:

* what's this mode
* allow  to create keys below other keys in the keyareaview
* escaped slashes will be part of the keyname and not create a new child
* if nodes that have no key are edited, a new key will be created, so they will contain a key afterwards
* added indicator in treeview if node has children
* allow user to select colors, gui is elektrified

Known issues:

* synchronize only works below user 

Other:
* many bugfixes


src/tools/qt-gui/qml/AboutWindow.qml
```diff
@@ -62,21 +62,500 @@ BasicWindow {
...
 						onLinkActivated: {
 							Qt.openUrlExternally(link)
 						}
 					}
 				}
+				Tab {
+					id: licenseTab
+
+					title: qsTr("License")
+					TextArea {
```
markus2330: I dont think we need creative commons full text here (especially not this long variant). Just state that the icons are used and put a link to the licence.

src/tools/qt-gui/src/guibackend.cpp
```diff
@@ -13,18 +13,13 @@ using namespace kdb;
...
 
-GUIBackend::GUIBackend(const GUIBackend &other)
-	: QObject()
-{
-	Q_UNUSED(other);
-}
-
 void GUIBackend::createBackend(const QString &mountpoint)
 {
 	m_backend = new Backend();
```
markus2330: prefer to use smart ptr

src/tools/qt-gui/src/guisettings.cpp
```diff
@@ -0,0 +1,162 @@
+#include "guisettings.hpp"
+#include <QDebug>
+#include <QPalette>
+
+using namespace kdb;
+
+namespace kdb
+{
+template <>
```
markus2330: add some comments

src/tools/qt-gui/src/guisettings.cpp
```diff
@@ -0,0 +1,162 @@
...
+	else
+		m_nodeWithKeyColor = lookup("nodeWithKeyColor");
+
+	if(!lookup("nodeWithoutKeyColor").isValid())
+		append("nodeWithoutKeyColor", m_nodeWithoutKeyColor);
+	else
+		m_nodeWithoutKeyColor = lookup("nodeWithoutKeyColor");
+
+	//make sure there is a configuration, even on the very first start of the gui
+	setKDB();
```
markus2330: You can only write into user and there only user-made changes should be reflected. So its not useful to write here already.

src/tools/qt-gui/src/undomanager.cpp
```diff
@@ -47,10 +47,16 @@ void UndoManager::createEditKeyCommand(TreeViewModel* model, int idx, QVariantLi
...
+
+		foreach (ConfigNodePtr node, tmpModel->model())
+		{
+			oldMDMap.insert(node->getName(), node->getValue());
+		}
 	}
+	else
+		tmpModel = new TreeViewModel;
 
 	data.insert(2, oldMDMap);
```
markus2330: why 2 (see also data.takeAt(2))? Please avoid magic constants and use const variables with name instead.

src/tools/qt-gui/src/treeviewmodel.hpp
```diff
@@ -240,6 +216,7 @@ class TreeViewModel : public QAbstractListModel
 signals:
 	void						showMessage(QString title, QString text, QString detailedText) const;
 	void						expandNode(bool);
+	void						updateIndicator() const;
```
markus2330: Please add some comments, its unclear what indicator is meant.

src/tools/qt-gui/src/treeviewmodel.cpp
```diff
@@ -562,7 +515,7 @@ void TreeViewModel::synchronize()
 
 	try
 	{
-		kdb.set(keySet, "/");
+		kdb.set(keySet, "/user");
```
markus2330: That should be "user". But I have to admit a good tutorial for namespaces is missing!
Please also mark that as TODO because obviously we want go back to "/" later.

src/tools/qt-gui/src/main.cpp
```diff
@@ -12,15 +12,19 @@
 #include "confignode.hpp"
 #include "undomanager.hpp"
 #include "guibackend.hpp"
+#include "guisettings.hpp"
 
 int main(int argc, char* argv[])
 {
 	QApplication app(argc, argv);
+	//keynames are only split at unescaped slashes
+	app.setProperty("KEY_DELIMITER", QRegularExpression("(?<!\\\\)/"));
```
markus2330: Thats only a heuristic which should be avoided. Please use keyGetBaseName() instead. Or do you need an iterator over key names?

src/tools/qt-gui/src/guisettings.cpp
```diff
@@ -0,0 +1,162 @@
...
+void GUISettings::setKDB()
+{
+	//won't set config without user prefix
+	try
+	{
+		m_kdb.set(m_config, "user" + m_base);
+	}
+	catch(const KDBException &ex)
+	{
+		qDebug() << ex.what();
```
markus2330: actually error+merging code should be reused here.

src/tools/qt-gui/src/guisettings.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+	emit nodeWithoutKeyColorChanged();
+}
+
+void GUISettings::append(const QString &keyName, const QColor &color)
+{
+	m_config.append(Key("user" + m_base + keyName.toStdString(), KEY_VALUE, color.name().toStdString().c_str(), KEY_END));
+}
+
+QColor GUISettings::lookup(const QString &keyName) const
```
markus2330: The method name is too generic (respective to implementation), either use templates or rename to lookupColor.

src/tools/qt-gui/src/treeviewmodel.cpp
markus2330: (There is however doc/NAMESPACES.md)

src/tools/qt-gui/src/guisettings.cpp
```diff
@@ -0,0 +1,162 @@
...
+	catch(const KDBException &ex)
+	{
+		qDebug() << ex.what();
+	}
+
+	//check if stored colors exist, if so, load them,else create them
+	if(!lookup("highlightColor").isValid())
+		append("highlightColor", m_highlightColor);
+	else
+		m_highlightColor = lookup("highlightColor");
```
markus2330: avoid that these strings are present multiple times. Also make them lowercase with underscores.

src/tools/qt-gui/src/guisettings.cpp
```diff
@@ -0,0 +1,162 @@
...
+	//set base path for all colors
+	m_base = "/sw/libelektra.org/qt-gui/#0/";
+
+	try
+	{
+		m_kdb.get(m_config, m_base);
+	}
+	catch(const KDBException &ex)
+	{
+		qDebug() << ex.what();
```
markus2330: give any context about this error (e.g. could not retrieve GUISettings).

You should think about a unified strategy (propagate to user vs. log)

src/tools/qt-gui/src/guibackend.cpp
```diff
@@ -108,11 +107,11 @@ void GUIBackend::serialise(TreeViewModel *model)
...
 		if (root == "system")
 		{
-			model->sink(model->model().at(0), keys, "system", k);
+			model->sink(model->model().at(0), keys, k);
 		}
 		else if (root == "user")
 		{
-			model->sink(model->model().at(1), keys, "user", k);
+			model->sink(model->model().at(1), keys, k);
 		}
```
markus2330: there are some more namespaces now (see doc/NAMESPACES.md). You can use

     elektraNamespace keyGetNamespace(Key const* key)

do make sure you won't miss one (compiler warns you if you haven't handeled all cases in switch)

src/tools/qt-gui/src/findvisitor.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef FINDVISITOR_HPP
+#define FINDVISITOR_HPP
+
+#include "visitor.hpp"
+#include "confignode.hpp"
+
+class FindVisitor : public Visitor
```
markus2330: Please add some documentation. What is this FindVisitor for?

src/tools/qt-gui/src/findvisitor.hpp
```diff
@@ -0,0 +1,19 @@
...
+{
+public:
+	explicit FindVisitor(TreeViewModel* searchResults, const QString& term);
```
markus2330: Is the TreeViewModel modified? (add const?)

markus2330:
Again, impressive work!

src/tools/qt-gui/src/undomanager.cpp
0003088: You complained that the signature of the command was too long, so I put the parameters into a map. Of course the index needs stay the same if a parameter is edited. I will add a comment that maps the indices to the items (or return to the long signature?).

src/tools/qt-gui/src/main.cpp
0003088: This regex is needed to create a correct hierarchy in the model and has nothing to to with the name of the Key object. It is used to split the QString that contains the keyname to create the new nodes.

src/tools/qt-gui/src/findvisitor.hpp
0003088: It implements the search function.

src/tools/qt-gui/src/findvisitor.hpp
0003088: It sure is, the nodes that contain the search term are added to this model.

src/tools/qt-gui/src/undomanager.cpp
markus2330: 0003088 wrote:
> You complained that the signature of the command was too long, so I put the
> parameters into a map. Of course the index needs stay the same if a
> parameter is edited. I will add a comment that maps the indices to the
> items (or return to the long signature?).

I am not sure if map is a good idea. A class/struct is more appropriate to 
avoid too many parameters.

The avoidance of magic values is, however, independent of this issue.

src/tools/qt-gui/src/main.cpp
markus2330: 0003088 wrote:
> This regex is needed to create a correct hierarchy in the model and has
> nothing to to with the name of the Key object. It is used to split the
> QString that contains the keyname to create the new nodes.

I am confused. Why does splitting the name has nothing to do with the name?

src/tools/qt-gui/src/undomanager.cpp
0003088: I chose a map because this function is called im QML and a map is natively supported in javascript.

src/tools/qt-gui/src/findvisitor.hpp
markus2330: 0003088 wrote:
> It sure is, the nodes that contain the search term are added to this model.

Please answer these questions in the comments.

src/tools/qt-gui/src/main.cpp
0003088: Let's say there is a key "user/sw/app/foo/baloo". To create a the correct hierarchy in the model I need to know where to split the string into parts to create the according nodes (earlier the delimiter was "/" which will work fine in this case). But what if the keyname is "user/sw/app/foo\/baloo"?  We decided that escaped slashed should be part of the keyname, so the "/" delimiter will split the string at the wrong character. To avoid this, the regex is used to split the string at "/" but not at "\/". I do not see why keyGetBaseName() would help me to decide where to split the string?

markus2330: 0003088 wrote:
> Let's say there is a key "user/sw/app/foo/baloo". To create a the correct
> hierarchy in the model I need to know where to split the string into parts
> to create the according nodes (earlier the delimiter was "/" which will
> work fine in this case). But what if the keyname is
> "user/sw/app/foo\/baloo"?  We decided that escaped slashed should be part
> of the keyname, so the "/" delimiter will split the string at the wrong
> character. To avoid this, the regex is used to split the string at "/" but
> not at "\/". I do not see why keyGetBaseName() would help me to decide
> where to split the string?

The problem is that user/sw/app/foo\\/baloo *should* be splitted, because the 
otherwise escaping \ is escaped. keyGetBaseName() correctly returns baloo and 
thus might help. If you want to iterate over all parts (user, sw, app, foo\ 
and baloo) you would need the key name iterator.

0003088: Then the key name iterator is the way to go, since I need all parts of the name.

0003088: Could you give me a hint where to find the iterator? I was not able to locate it.

markus2330: Hi,

0003088 wrote:
> Could you give me a hint where to find the iterator? I was not able to
> locate it.

It was not there, thats why I asked if you need it.
I now added a bidirectional (reverse) iterator in key.hpp
d3ab632e209e7df2f3dba1f1a39a2ed234c24c8e

Tests are here
src/bindings/cpp/tests/testcpp11_iter_name.cpp

You find examples how to use it in
src/bindings/cpp/examples/cpp_example_iter_name.cpp

best regards

0003088: Thank you, I thought it already existed.

markus2330:
For your information: The merging simplification is also merged.

Please ping me here if its ready to merge.

0003088:
Addressed the issues above. I implemented synchronize for "user" only, because kdb.set() for keys other than user is broken at the moment anyway. Should merging with cascading keys work in the future, its only necessary to change the root key from "user" to "/"

markus2330:
Perfect, thank you! I work on that problem and will change "user" to "/" when 
its fixed.

Issue #186, opened 2015-02-22
OS X build fixes


beku:
The osx-01 branch contains some fixes. The state is not final. Build stops at around 60%. The commit format is not always best.

Here the current issue:
Linking CXX executable ../../../bin/kdb-static
cd /Users/kuwe/programme/oyranos/libelektra/build3/src/tools/kdb &&
/opt/local/bin/cmake -E cmake_link_script
CMakeFiles/kdb-static.dir/link.txt --verbose=1
/usr/bin/c++    -std=c++98  -Wno-deprecated-declarations
-Wstrict-null-sentinel -Wno-missing-field-initializers -Wold-style-cast
-Woverloaded-virtual  -Wsign-promo  -pedantic -Wno-variadic-macros -Wall
-Wextra -Wno-overlength-strings -Wsign-compare -Wfloat-equal
-Wformat-security -Wshadow -Wcomments -Wtrigraphs -Wundef
-Wuninitialized -O2 -g -DNDEBUG -Wl,-search_paths_first
-Wl,-headerpad_max_install_names   CMakeFiles/kdb-static.dir/check.cpp.o
CMakeFiles/kdb-static.dir/cmdline.cpp.o
CMakeFiles/kdb-static.dir/command.cpp.o
CMakeFiles/kdb-static.dir/convert.cpp.o
CMakeFiles/kdb-static.dir/cp.cpp.o
CMakeFiles/kdb-static.dir/export.cpp.o
CMakeFiles/kdb-static.dir/external.cpp.o
CMakeFiles/kdb-static.dir/file.cpp.o
CMakeFiles/kdb-static.dir/fstab.cpp.o
CMakeFiles/kdb-static.dir/get.cpp.o
CMakeFiles/kdb-static.dir/import.cpp.o
CMakeFiles/kdb-static.dir/info.cpp.o
CMakeFiles/kdb-static.dir/list.cpp.o CMakeFiles/kdb-static.dir/ls.cpp.o
CMakeFiles/kdb-static.dir/main.cpp.o
CMakeFiles/kdb-static.dir/merge.cpp.o
CMakeFiles/kdb-static.dir/mergehelper.cpp.o
CMakeFiles/kdb-static.dir/metaget.cpp.o
CMakeFiles/kdb-static.dir/metals.cpp.o
CMakeFiles/kdb-static.dir/metaset.cpp.o
CMakeFiles/kdb-static.dir/mount.cpp.o
[...]

PR #192, opened 2015-03-03
OS X 01

beku:
Here comes a bunch of fixes for OS X. They are tested on  SL, which is slightly aged but should be otherwise fine with current environments. Elektra builds with these change set fine, except some undefined inbuild functions from missed string.h and stdio.h inclusions, which I did not hunt for.

See as well issue #186 .

The src/bindings/cpp/include/keyset.hpp could be reverted. I guess the clang fix is better.

The changes in src/plugins/resolver/resolver.c provide a finer grained feature detection than the previously assumed Linux style API set. The last compile error was fixed by providing a getter function instead of accessing a global variable.


cmake/ElektraCompiling.cmake
```diff
@@ -84,9 +84,9 @@ set (COMMON_FLAGS "${COMMON_FLAGS} -Wsign-compare -Wfloat-equal")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wformat-security")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wshadow")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wcomments -Wtrigraphs -Wundef")
-set (COMMON_FLAGS "${COMMON_FLAGS} -Wuninitialized -Winit-self -Wmaybe-uninitialized")
```
markus2330: Is Wmaybe-uninitialized unknown to the apple compiler? Ideally we should have a ```if (apple)``` and move non-common flags to the specific places.

src/include/kdbprivate.h
```diff
@@ -518,10 +518,13 @@ int elektraValidateKeyName(const char *name, size_t size);
  */
 #define ELEKTRA_MAX_ARRAY_SIZE (21)
 
+#if !defined(__cplusplus)
```
markus2330: was this necessary?

src/include/kdbtypes.h
```diff
@@ -85,7 +89,7 @@ typedef unsigned char             kdb_octet_t;
 typedef signed short              kdb_short_t;
 typedef unsigned short            kdb_unsigned_short_t;
 
-#if SIZEOF_LONG == 4
+#if (LONG_MAX == +2147483647)
```
markus2330: What is the value of SIZEOF_LONG?

src/libelektra/keymoremeta.c
```diff
@@ -20,6 +20,8 @@
 
 #ifdef HAVE_ERRNO_H
 #include <errno.h>
+#elif defined(__APPLE__)
```
markus2330: I think we should fix HAVE_ERRNO_H for APPLE

src/plugins/resolver/mutex.c
```diff
@@ -3,5 +3,8 @@
 #include <pthread.h>
 
 // every resolver should use the same mutex
-pthread_mutex_t elektra_resolver_mutex =
-	PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
```
markus2330: Actually I think this won't work. The mutex won't be recursive, that means multiple nested KDB instances won't work anymore (but will deadlock).

If there is no equivalent for PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP on apple, we need an ifdef so that the problem only exists on apple.

src/plugins/resolver/resolver.c
```diff
@@ -68,6 +80,16 @@ static void resolverInit (resolverHandle *p, const char *path)
...
 	p->path = path;
+
+	{
+		pthread_mutexattr_t mutex_attr;
+
+		int result = pthread_mutexattr_init(&mutex_attr);
+		if(result == 0)
+			result = pthread_mutexattr_settype(&mutex_attr, PTHREAD_MUTEX_RECURSIVE);
+		if(result == 0)
+			result = pthread_mutex_init(getResolverMutex(), &mutex_attr);
```
markus2330: This seems not very robust. Failures of pthread_mutexattr_init are simply ignorend?

Additionally it seems like the mutex is reinitialized for every active resolver or did I miss something?

src/plugins/resolver/resolver.c
```diff
@@ -51,8 +52,19 @@
...
+#if defined(__APPLE__)
+#define statSeconds(status) status.st_mtime
+#define statNanoSeconds(status) status.st_mtimespec.tv_nsec
+#elif defined(WIN32)
+#define statSeconds(status) status.st_mtime
+#define statNanoSeconds(status) 0
+# else
+#define statSeconds(status) status.st_mtim.tv_sec
+#define statNanoSeconds(status) status.st_mtim.tv_nsec
+#endif
```
markus2330: nice solution, thanks

markus2330: did it actually also compile with WIN32?

cmake/ElektraCompiling.cmake
petermax2: Clang only understands `-Wuninitialized` on OS X. Will use the `APPLE` variable to check for OS  X.

src/include/kdbprivate.h
petermax2: yes, otherwise the compiler confuses the types on OS X

src/include/kdbtypes.h
petermax2: SIZEOF_LONG: 8
LONG_MAX: 9223372036854775807
on a 64 bit Mac

petermax2: what was the problem with SIZEOF_LONG anyway?

src/include/kdbtypes.h
```diff
@@ -101,14 +105,14 @@ typedef unsigned int              kdb_unsigned_long_t;
...
 
-#if SIZEOF_LONG == 8
+#if (LONG_MAX == +9223372036854775807)
 #define ELEKTRA_LONG_LONG_F "%ld"
 #define ELEKTRA_LONG_LONG_S strtol
 typedef long                      kdb_long_long_t;
 #define ELEKTRA_UNSIGNED_LONG_LONG_F "%lu"
 #define ELEKTRA_UNSIGNED_LONG_LONG_S strtoul
 typedef unsigned long             kdb_unsigned_long_long_t;
-#elif defined(HAVE_SIZEOF_LONG_LONG) && (SIZEOF_LONG_LONG == 8)
```
petermax2: both `HAVE_SIZEOF_LONG_LONG` and `SIZEOF_LONG_LONG` are correctly set. I do not like the notion of having magic numbers (comparing LONG_MAX). what is the reason for this change?

src/bindings/cpp/include/keyset.hpp
```diff
@@ -21,10 +21,10 @@ class KeySetReverseIterator;
  */
 struct Va
 {
-	Va(){}
```
petermax2: this change will be undone as requested by @markus2330

src/libelektra/keymoremeta.c
petermax2: `HAVE_ERRNO_H` is defined on my setup. Is there a certain configuration where `HAVE_ERRNO_H` fails / is not being defined?

markus2330:
@petermax2: if you move around the compiler flags, please also fix mingw: ```cc1plus: error: unrecognized command line option "-Wmaybe-uninitialized"```

src/include/kdbprivate.h
markus2330: I did not talk about to int64_t -> kdb_long_long_t but about the ```#if !defined(__cplusplus)```. Why should the compiler confuse the types in C++?

src/include/kdbtypes.h
markus2330: If it runs into the ```# error``` we should add an additional apple branch and not messing up with existing typedefs.

src/include/kdbtypes.h
markus2330: Yes, I agree. Same thing here, if needed we can add another branch.

src/include/kdbtypes.h
beku: @petermax2 SIZEOF_LONG is not defined
LONG_MAX is defined in ANSI C
http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
So every compiler should support those numbers. lcms2.h and other projects have those values and they are used on so many platforms including OS X, Windows, Embedded ...

src/plugins/resolver/mutex.c
beku: the recursive aware initialisation is handled by PTHREAD_MUTEX_RECURSIVE

src/plugins/resolver/resolver.c
beku: Oh yes. A 
   static int initialised = 0;
might do the trick. The calls are tacken from the POSIX documentation. So my guess is that the dynamic allocation should not fail except for the stack being limited in size for the pthread_mutex_t.

src/include/kdbtypes.h
beku: SIZEOF_LONG_LONG is not defined on my system (OS X SL).
//Result of TRY_COMPILE
HAVE_SIZEOF_LONG_LONG:INTERNAL=FALSE

src/plugins/resolver/resolver.c
beku: I want to check that.

src/include/kdbtypes.h
markus2330: SIZEOF_LONG is set by cmake and should be fixed that it works on apple, too

    src/include/CMakeLists.txt:check_type_size(long            SIZEOF_LONG)

src/include/kdbtypes.h
markus2330: Same thing here:

    src/include/CMakeLists.txt:check_type_size("long long"     SIZEOF_LONG_LONG)

Can you show the log of TRY_COMPILE? Is long long available at all?

src/plugins/resolver/mutex.c
markus2330: So we simply need an ifdef to either use PTHREAD_MUTEX_RECURSIVE or PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP

src/plugins/resolver/resolver.c
markus2330: Its not that simple, int is not thread-safe. So we would need a mutex to protect the recursive mutex initialization. If PTHREAD_MUTEX_RECURSIVE works, its the ideal solution. Otherwise (if no such thing for Apple exists or works) I would like to have the double-mutex thing *only* for Apple or other OS not supporting proper recursive mutex initialization.

markus2330:
gcc4.4 does not support the flag, too: ```cc1plus: error: unrecognized command line option "-Wmaybe-uninitialized"``` See http://community.markus-raab.org:8080/

src/include/kdbtypes.h
petermax2: as mentioned before, it works on my OS X setup:

	//Result of TRY_COMPILE
	HAVE_SIZEOF_LONG:INTERNAL=TRUE
	//Result of TRY_COMPILE
	HAVE_SIZEOF_LONG_DOUBLE:INTERNAL=TRUE
	//Result of TRY_COMPILE
	HAVE_SIZEOF_LONG_LONG:INTERNAL=TRUE

the values are set as expected on a 64 bit machine

	//CHECK_TYPE_SIZE: sizeof(long)
	SIZEOF_LONG:INTERNAL=8
	//CHECK_TYPE_SIZE: sizeof(long long)
	SIZEOF_LONG_LONG:INTERNAL=8

src/include/kdbprivate.h
petermax2: Removing the `#if !defined(__cplusplu)` line (and the corresponding `#endif`) results in 

	Linking CXX shared module ../../../lib/libelektra-dump.so
	[ 40%] Built target elektra-dump
	In file included from /Users/pnirschl/Documents/libelektra/src/libtools/src/plugin.cpp:16:
	/Users/pnirschl/Documents/libelektra/src/include/kdbprivate.h:525:50: error: unknown type name 'kdb_long_long_t'; did you mean 'kdb::long_long_t'?
	int elektraReadArrayNumber(const char *baseName, kdb_long_long_t *oldIndex);
	                                                 ^~~~~~~~~~~~~~~
	                                                 kdb::long_long_t
	/Users/pnirschl/Documents/libelektra/src/include/kdbtypes.h:47:18: note: 'kdb::long_long_t' declared here
	        typedef int64_t long_long_t; // default: 0
	                        ^
	In file included from /Users/pnirschl/Documents/libelektra/src/libtools/src/plugin.cpp:16:
	/Users/pnirschl/Documents/libelektra/src/include/kdbprivate.h:526:44: error: unknown type name 'kdb_long_long_t'; did you mean 'kdb::long_long_t'?
	int elektraWriteArrayNumber(char *newName, kdb_long_long_t newIndex);
	                                           ^~~~~~~~~~~~~~~
	                                           kdb::long_long_t
	/Users/pnirschl/Documents/libelektra/src/include/kdbtypes.h:47:18: note: 'kdb::long_long_t' declared here
	        typedef int64_t long_long_t; // default: 0
	                        ^
	2 errors generated.
	make[2]: *** [src/libtools/src/CMakeFiles/elektratools.dir/plugin.cpp.o] Error 1
	make[1]: *** [src/libtools/src/CMakeFiles/elektratools.dir/all] Error 2
	make: *** [all] Error 2

petermax2: are these functions only exposed in a `C` environment?

	int elektraArrayValidateName(const Key *key);
	int elektraReadArrayNumber(const char *baseName, kdb_long_long_t *oldIndex);
	int elektraWriteArrayNumber(char *newName, kdb_long_long_t newIndex);

petermax2:
CMake supports checkinf or compiler flags using the `check_c_compiler_flag` function. See [Stackoverflow](http://stackoverflow.com/questions/3134660/how-to-apply-different-compiler-options-for-different-compilers-in-cmake)

src/include/kdbtypes.h
beku: It's possible that I take the route to fix the cmake issue.

IMHO a standards based approach like ANSI C's limits.h content appears much more straight forward to me. That is even independent of build systems.

src/include/kdbtypes.h
beku: Appears to work now, without having a glue why. Perhaps wrong CFLAGS.

beku:
Appears to work now, without having a glue why. Perhaps wrong CFLAGS. errno-h and the sizeof macros are all set fine now.

petermax2:
@markus2330 do we use the standard `LONG_MAX` variant or do we keep the CMake `SIZEOF_LONG` macros?

src/plugins/resolver/mutex.c
petermax2: is it possible, that `PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP` is something Windows-specific? I did a little Google research and found that `PTHREAD_MUTEX_RECURSIVE` is the POSIX way for recursive mutex creation, that should be supported on all platforms.

markus2330:
Hello,

Peter Nirschl wrote:
> @markus2330 do we use the standard `LONG_MAX` variant or do we keep the
> CMake `SIZEOF_LONG` macros? 

SIZEOF_LONG is fine, but we should find out why try_compile fails in older mac 
versions.

best regards
Markus

src/plugins/resolver/mutex.c
markus2330: Hello,

Peter Nirschl wrote:
> is it possible, that `PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP` is something
> Windows-specific? I did a little Google research and found that
> `PTHREAD_MUTEX_RECURSIVE` is the POSIX way for recursive mutex creation,
> that should be supported on all platforms.

Its rather GNU libc specific. PTHREAD_MUTEX_RECURSIVE seems to be only the 
type, but not an initialization. [The manpage](http://manpages.debian.org/cgi-
bin/man.cgi?query=pthread_mutex_lock) does not mention any portable way to do 
it.

Best regards,
Markus

petermax2:
I added a proposal to fix the recursive mutex initialization quirks under OS X. Please have a look at [5b217bc78a54e6eeafbc097872988c6c994ff24b](https://github.com/petermax2/libelektra/commit/5b217bc78a54e6eeafbc097872988c6c994ff24b) and tell me if this makes sense or not.

markus2330:
Hello,

Peter Nirschl wrote:
> I added a proposal to fix the recursive mutex initialization quirks under
> OS X. Please have a look at
> [5b217bc78a54e6eeafbc097872988c6c994ff24b](https://github.com/petermax2/li
> belektra/commit/5b217bc78a54e6eeafbc097872988c6c994ff24b) and tell me if
> this makes sense or not.

Nicely done! If the Linux part is unchanged and the apple part works it surely 
makes sense ;)

For error handling: simply fail (ELEKTRA_SET_ERROR and return -1 respective 
cleanup). It should only fail in ulimit scenarios anyway. Same for 
pthread_mutexattr_init.

Btw. resolverInit is the completely wrong place to do the initialization stuff 
because its called n times (per namespace).

So please move everything directly to
int ELEKTRA_PLUGIN_FUNCTION(resolver, open)
(then return -1 works, too)

Also please make all your methods static (if you need them at all, they are 
quite trivial and apple only.  Then you do not need a mutex.h, too), otherwise 
some compilation variants will fail because of duplicate symbols.

Additionally, make sure that all test cases run under Mac OS X, I cannot test 
them.

These things considered we finally can merge the mac os x fixes!

Thank you!

best regards
Markus

petermax2:
the test-cases succeed on my OS X setup as well as on my ArchLinux setup.

markus2330:
Hello,

Peter Nirschl wrote:
> the test-cases succeed on my OS X setup as well as on my ArchLinux setup.

Perfect. Thank you, nice job!

Can you prepare the merge request then?

best regards

cmake/ElektraCompiling.cmake
```diff
@@ -92,7 +94,13 @@ set (COMMON_FLAGS "${COMMON_FLAGS} -Wshadow")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wcomments -Wtrigraphs -Wundef")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wuninitialized -Winit-self")
 
-set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--unresolved-symbols=ignore-in-shared-libs")
+# Not every compiler understands -Wmaybe-uninitialized
+check_c_compiler_flag(-Wmaybe-uninitialized HAS_CFLAG_MAYBE_UNINITIALIZED)
+if (HAS_CFLAG_MAYBE_UNINITIALIZED)
+	set (COMMON_FLAGS "${COMMON_FLAGS} -Wmaybe-uninitialized")
+endif (HAS_CFLAG_MAYBE_UNINITIALIZED)
```
markus2330: Nice solution! Is better than the gcc4.4 check in master. Thank you!

petermax2:
I think we are good to go.

src/include/kdbprivate.h
markus2330: This was the case some time ago (because mingw was not able to build them because of int64_t). But now they should work everywhere.

In kdbprivate.h there was still a int64_t, I fixed this in 9734b95067cfb3b588cc7352b464e730b891e5cc

tests/cframework/tests.c
```diff
@@ -66,6 +67,15 @@ int init (int argc, char**argv)
 	{
 		tmpvar = "/tmp";
 	}
+	// check tempvar for trailing slash / 
+	if(strlen(tmpvar) > 2)
+	{
+		if(tmpvar[strlen(tmpvar) - 1] == '/')
```
markus2330: If trailing slashes are a problem then this should be a loop.

PR #207, opened 2015-04-01
Fixed an issue causing the rename plugin to handle the parentKey incorrectly

### LINKED ISSUE #206: rename: changing parent key does not work
Suppose we mount openicc the usual way:
```
# kdb mount --resolver=resolver_fm_xhp_x color/settings/openicc-devices.json /org/freedesktop/openicc yajl rename cut=org/freedesktop/openicc
```
And then set something:
```
$ kdb set user/org/freedesktop/openicc INIT
```
Then the content of the file is:
```
{
    "org": {
        "freedesktop": {
            "openicc": {
                "user": "INIT"
            }
        }
    }
}
```

But expected would be:
```
{
    "org": {
        "freedesktop": {
            "openicc":"INIT"
        }
    }
}
```

It seems like rename somehow wrongly puts the namespace for the top-level parent key.

The problem is only present for the parent key.

Btw. even when we execute a commando again with different values:

```
$ kdb set user/org/freedesktop/openicc BUG
```

The file is unchanged.

fberlakovich:
This pull request should fix #206. A test that failed before this commit is appended

### DIFF
--- a/src/plugins/rename/rename.c
+++ b/src/plugins/rename/rename.c
@@ -93,8 +93,13 @@ static Key *restoreKeyName(Key *key, const Key *parentKey, const Key *configKey)
 			Key *result = keyDup(key);
 			keySetName(result, keyName(parentKey));
 			keyAddUnescapedBasePath(result, keyString(configKey));
-			const char *relativePath = keyName(key) + keyGetNameSize(parentKey);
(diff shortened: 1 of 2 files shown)


markus2330:
Thank you for fast response! There was a memory issue in the test case. I fixed it and explained the issue and how to avoid it in the ksAppendKey() docu.

PR #211, opened 2015-04-02
Add PLUGIN_DEPENDENCIES.md 

iandonnelly:
Added document that list packages needed to build plugins with dependencies.


markus2330:
Can one of the admins verify if this patch should be build?

markus2330:
Can one of the admins verify if this patch should be build?

iandonnelly:
I split the dependencies into each plugin's readme as Markus and I agreed it would be better this way.

markus2330:
thank you!

PR #213, opened 2015-04-04
kdbget example improved

KurtMi:
done some TODOs in the kdbget exmaple, to show proper error handling.


markus2330:
Can one of the admins verify if this patch should be build?

markus2330:
Can one of the admins verify if this patch should be build?

markus2330:
add to whitelist

examples/kdbget.c
```diff
@@ -8,12 +9,52 @@ int main()
 	KDB *handle = kdbOpen(key);
 
 	kdbGet(handle, myConfig, key);
-	// TODO: check for errors and warnings in key
 
-	keyDel(key);
+	//check for error
+	if(keyGetMeta (key,"error"))
```
markus2330: actually it would be better to check the return value of kdbGet() and do error handling based on that

markus2330:
Thank you! Always nice to see the examples improved. Is it possible to make two examples out of it: 1.) minimal kdbget to see the basic idea (could be more commented though and you can refer to the other example)
2.) kdbget with error handling as done in your example

Maybe its also a good idea to have an introductory README.md to guide the user through the examples.

examples/kdbget.c
```diff
@@ -8,12 +9,52 @@ int main()
...
+	{
+		const char * warning_count = keyString (keyGetMeta (key,"warnings"));
+		char buffer[] = "warnings/#00/description";
+
+		do {
+			const Key *warnkey = keyGetMeta (key,buffer);
+			printf ("Warning occurred: %s\n",keyString (warnkey));
+			//fields for more information are listed in the value from
+			//the Key returned by keyGetMeta(key,"warnings/#00")
+			buffer[11]++;
```
markus2330: avoid to hardcode 11, prefer to use a variable, e.g. indexOfUnitPosition indexOfTensPosition

examples/kdbget.c
```diff
@@ -8,12 +9,52 @@ int main()
...
+			{
+				buffer[11] = '0';
+				buffer[10]++;
+			}
+		} while (strncmp (&buffer[10],warning_count,2)<=0);
+	}
+
+	keyDel (key);
+
+	//insert a key so lookup has something to lookup
```
markus2330: rephrase e.g.: "lookup always will find something, even if kdbGet() did not return the value"

the appending can also be done before the kdbGet(), then a user supplied configuration would win and make the program execution of the example more interesting

markus2330: (using ```kdb set user/sw/MyApp/Tests/TestKey1 value``` could be mentioned then)

examples/kdbget.c
```diff
@@ -8,12 +9,52 @@ int main()
...
+					KEY_VALUE, "foo",
+					KEY_META, "comment", "bar",
+					KEY_END));
+
+	//lookup
+	Key * result = ksLookupByName (myConfig,"/sw/MyApp/Tests/TestKey1", 0);
+	if(result)
+	{
+		//do something with the key
+		keySetString(result,"foobar");
```
markus2330: keySetString without later kdbSet() has no effect. Because it is the kdbGet() example, printing the value out would be expected

markus2330:
further improvements please as discussed

markus2330:
jenkins build please

examples/kdbget_error.c
```diff
@@ -0,0 +1,90 @@
...
+	keyDel (key);
+
+	//lookup
+	Key * result = ksLookupByName (myConfig,"/sw/MyApp/Tests/TestKey1", 0);
+	if(result)
+	{
+		//do something with the key
+	} else printf("Key not found in KeySet\n");
+
+	const char * key_name = keyName(result);
```
markus2330: ```do something with the key``` should be the place where the next lines are placed.

With the given code the program might print
```
Key not found in KeySet
key: (null) value: (null) comment: (null)
```

PR #222, opened 2015-04-19
Fixed an issue in rename that caused the parentKey value to be discarded

### LINKED ISSUE #206: rename: changing parent key does not work

fberlakovich:
Should fix #206

### DIFF
--- a/src/plugins/rename/README.md
+++ b/src/plugins/rename/README.md
@@ -18,17 +18,13 @@ in the required format. If keys are renamed, their original name is stored in th
 ### OPERATION ###
 
 The cut operation can be used to strip parts of a keys name. The cut operation is able to cut anything starting after the path
-of the parent key. However, the resulting name must not be equal to the parent key. For example consider a KeySet with the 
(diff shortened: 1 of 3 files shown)


markus2330:
Thank you!

PR #228, opened 2015-05-24
fix compilation of race tool under oS-11.4

beku:
add missed POSIX include


markus2330:
Thank you for the nice and easy fix!

beku:
glad it is useful

Issue #174, opened 2015-01-16
libtools: cascading merging


markus2330:
Have high level strategies for importing and kdb conflicts

markus2330:
Thank you for resolving this issue!

As already mentioned, cascading merging is still missing. I will rename the issue.

markus2330:
Any progress on this issue?

If following patch is applied:
```
diff --git a/src/tools/qt-gui/src/treeviewmodel.cpp b/src/tools/qt-gui/src/treeviewmodel.cpp
index e3b151a..3628559 100644
--- a/src/tools/qt-gui/src/treeviewmodel.cpp
+++ b/src/tools/qt-gui/src/treeviewmodel.cpp
@@ -474,7 +474,7 @@ void TreeViewModel::synchronize()
        KeySet base = GUIBasicKeySet::basic();
        KeySet resultKeys;
 
-       Key root("user", KEY_END);
+       Key root("/", KEY_END);
 
        KDB kdb;
```

synchronization leads to:
```
terminate called after throwing an instance of 'kdb::tools::helper::InvalidRebaseException'
  what():  the supplied key is not below the old parent
```

because cascading is not supported in merging.

So I will release the qt-gui with only "user" writeable.

markus2330:
#231 did not fully adress this issue, see comments within pull request

PR #231, opened 2015-07-10
Fixed merge problems with cascading keys

fberlakovich:
This patch fixes the problem discussed in #174 as well as another issue that caused the synchronization in the qt-gui to fail. The problem discussed in #174 is not really about merging cascading keys, but about handling cascading parent keys (cascading parent keys caused the InvalidRebaseException).

More special treatment for cascading keys in the ThreeWayMerger is thinkable. For example what should happen if one of the keys in the merged keysets is a cascading key? This issue is currently not addressed, but for now I concentrated on fixing the issues affecting the qt-gui. I would suggest renaming #174 and creating another issue that is really about the treatment of cascading keys during a merge.

### DIFF
--- a/src/libtools/src/helper/comparison.cpp
+++ b/src/libtools/src/helper/comparison.cpp
@@ -23,10 +23,18 @@ namespace helper
 bool keyDataEqual(const Key& k1, const Key& k2)
 {
 	if (!k1 || !k2) return false;
-	if (k1.getString () != k2.getString ())
+
+	if (k1.isBinary() != k2.isBinary()) return false;
+
+	if (k1.isBinary() && k2.isBinary())
 	{
-		return false;
+		return k1.getBinary() == k2.getBinary();
 	}
+	else
+	{
+		return k1.getString() == k2.getString();
+	}
+
 	return true;
 }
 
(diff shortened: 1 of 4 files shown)


markus2330:
Thank you for your quick fix, that's great!

markus2330:
Btw. you understood #174 correctly, it is about cascading parent keys. Cascading keys within the keyset are possible, but merging them is most likely not needed because cascading keys are never persistent. So the cascading keys come from only one source and hence no merge is needed.

markus2330:
There is, however, an issue with this fix. The rebasePath with parents that are '/' creates cascading keys. (e.g. in keyhelper.cpp:50 result.setName (newPath); might make "result" a cascading key). Then the internal assumption that no cascading keys exist is broken and e.g. "source" in 
automergestrategy.cpp:40 is a null-key, making source.getString() throw an exception.

Some higher level test cases would be needed, too.

To reproduce these issues you only have to change 

    -       Key root("user", KEY_END);
    +       Key root("/", KEY_END);

in src/tools/qt-gui/src/treeviewmodel.cpp:~500

PR #232, opened 2015-07-12
Another cascading merge fix

fberlakovich:
Another try to fix the issue discussed in #174. When I try the qt-gui now with "/" instead of "user" the only problem left is an error to set information below the version mountpoint. However, this seems unrelated to the merging framework. When stepping through the code in the debugger, added keys are merged fine.


markus2330:
Thank you, great work! The version problem is fixed in 09129b72ab0ce0674abd35e110cb977e7638ca3e

PR #236, opened 2015-07-19
tests: fix build with external gtest

pinotree:
Move the own gtest-elektra.h out of the embedded copy of gtest, to a specific directory used whichever version (external or embedded) of gtest is used.

P.S.: this was one of the reasons why keeping an embedded copy of something else produces bad effects...


markus2330:
Thank you for the fix!

I am also not happy with the internal copy, but hardly any system has the source installed (nor has it available in its distribution).

PR #241, opened 2015-07-23
tests: fix a couple of bashisms

pinotree:


markus2330:
Thank you!
(Actually I use zsh, so I changed the git comment when merging)

PR #242, opened 2015-08-02
cmake: switch from FindElektra.cmake to ElektraConfig.cmake

pinotree:
Instead of providing a `FindElektra.cmake` module to find an installed elektra, provide a `ElektraConfig.cmake` module with the configuration of the installed elektra. This has few advantages:
- `ElektraConfig.cmake` imports the targets and sets variables for paths, since they are known when building/installing elektra
- the configuration files are not installed among the private CMake module path

The additional generated `ElektraConfigVersion.cmake` file takes care of the version handling, so specifying a version in
```cmake
find_package(Elektra <version>)
```
now will do the proper thing.

Following CMake' style, now exported variables are `Elektra_*`, with `ELEKTRA_*` provided for compatibility.


markus2330:
Thank you, but I'll have to update examples/external/cmake/ before I can merge this pull request.

pinotree:
> I'll have to update examples/external/cmake/ before I can merge this pull request.

How so? The new module maintains "compatibility" in its name and how variables are exported; at least for me, examples/external/cmake/ worked fine by building it manually after installing the patched elektra. If that isn't the case, then please do share the results, and I will fix accordingly.

markus2330:
Ok, I just thought that examples should use the latest technique, but if the compatibility is here to stay, then its ok.

PR #243, opened 2015-08-02
Copy COPYING-CMAKE-SCRIPTS and Copyright.txt from CMake

pinotree:
Some of the CMake modules shipped refer to these two files for their license, so copy them to have the proper license pointers.


markus2330:
Thank you for fixing this!

PR #244, opened 2015-08-02
swig-python: add support for copy module

manuelm:
subjects says it all

### DIFF
--- a/src/bindings/swig/python3/examples/example_keyset.py
+++ b/src/bindings/swig/python3/examples/example_keyset.py
@@ -39,3 +39,18 @@ print("")
 print("You asked for slices? You get slices:")
 print("  KeySet1[1:3]={0}".format([ str(k) for k in ks1[1:3] ]))
 print("")
+
+print("We can create shallow copies and remove keys without affecting other keysets:")
+ks2 = kdb.KeySet(ks1.dup())
+# or ks2 = copy.copy(ks1)
+ks2.pop()
+print("  KeySet2 now has {0} keys while KeySet1 still has {1} keys".format(len(ks2), len(ks1)))
+
+print("In Python we can even create deep copies and modify the keys inside:")
+ks1[0].setMeta("foo", "bar")
+import copy
+ks2 = copy.deepcopy(ks1)
+ks2[0].setMeta("foo", "changed")
+print("  KeySet1[0].getMeta('foo').value={0}".format(ks1[0].getMeta("foo").value))
+print("  KeySet2[0].getMeta('foo').value={0}".format(ks2[0].getMeta("foo").value))
+print("")
--- a/src/bindings/swig/python3/kdb.i
+++ b/src/bindings/swig/python3/kdb.i
@@ -124,6 +124,10 @@
     return ckdb::keyCmp($self->getKey(), o->getKey());
   }
 
+  kdb::Key *__copy__() {
+    return new kdb::Key($self->dup());
+  }
+
   %pythoncode %{
     def get(self):
       """returns the keys value"""
@@ -209,6 +213,19 @@
 %extend kdb::KeySet {
   KeySet(size_t alloc) {
    return new kdb::KeySet(alloc, KS_END);
+  }
+
+  kdb::KeySet *__copy__() {
+    return new kdb::KeySet(*$self);
+  }
+
+  kdb::KeySet *__deepcopy__(PyObject *memo) {
+    (void) PyDict_Check(memo);
+    ssize_t size = $self->size();
+    kdb::KeySet *ks = new kdb::KeySet(size, KS_END);
+    for(cursor_t cursor = 0; cursor < size; ++cursor)
+      ks->append($self->at(cursor)->dup());
+    return ks;
   }
 
   %pythoncode %{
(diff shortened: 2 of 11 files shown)


pinotree:
Sounds good at a first glance.
Wouldn't be better to also hide `dup()` in the python bindings, so it is not called accidentally anymore?

manuelm:
I see dup() as a companion to getKey(), which might be useful for passing the underlying c structure to 3rd party functions. But one might argue I could easily copy dup() by using the constructor + getKey(). So I'll leave this for someone else to decide :-)

markus2330:
1.) Could you please document the various types of duplications (incRef, dup, deep copy), ideally with examples how to use each of them.
2.) Are the copy sematics consistent to C++, Java and Lua semantics?

Regarding to if dup() returning a raw pointer is needed: Could you please benchmark creation of a key object compared to the swig proxy object? If there is a benefit in providing it, maybe renaming it to getDupKey(), cdup() or rawDup() would be less confusing?

src/bindings/swig/python3/kdb.i
```diff
@@ -124,6 +124,10 @@
     return ckdb::keyCmp($self->getKey(), o->getKey());
   }
 
+  kdb::Key *__copy__() {
+    return new kdb::Key(*$self);
```
markus2330: Isn't this just taking a reference of the key?

src/bindings/swig/python3/kdb.i
```diff
@@ -211,6 +215,19 @@
...
+  kdb::KeySet *__copy__() {
+    return new kdb::KeySet(*$self);
+  }
+
+  kdb::KeySet *__deepcopy__(PyObject *memo) {
+    (void) PyDict_Check(memo);
+    ssize_t size = $self->size();
+    kdb::KeySet *ks = new kdb::KeySet(size, KS_END);
+    for(cursor_t cursor = 0; cursor < size; ++cursor)
+      ks->append($self->at(cursor)->dup());
```
markus2330: why not using ksDup? ksDup will have optimizations in future, e.g. use the same hashmap without rebuilding it

manuelm: swig is c++. and dup is just calling ksDup anyway

src/bindings/swig/python3/kdb.i
manuelm: yep. my bad.

manuelm:
re 1:
* help(kdb.Key) documents __copy__. No need for any more documentation. It's the standardized way for (deep) copying.
* _incRef is hidden (prefixed by _). I see no reason why anyone should use it.
* dup shouldn't be used too and I'll hide it as well so no one else will use it unless there's some nasty hacks needed in which case reading code is probably the least sort of problem.

In general I see no reason for documentation of bindings if the underlying API call with the exact same name has already been sufficiently documented.

re 2: please elaborate

markus2330:
Thank you for your answers. The question 2 was: The copy constructor in C++ creates a new reference of a Key, but a duplicate in the case of the KeySet (creating new references of Keys for it). Does Python and Lua have the same behaviour?

Btw. The deep duplication code would be useful in C++, too. Don't you think?

manuelm:
Yes, both bindings are duplicating the C++ semantics . The key constructor creates a new object holding the same underlying key, the ks constructor creates a copy of the set with the same keys.

manuelm:
If we add some sort of deepcopy to C++ KeySet all swig bindings will benefit from this. Right now this is Python only.

I'll add tests/example after #259 has been merged.

manuelm:
jenkins build bindings please

manuelm:
tests + example added

manuelm:
jenkins build bindings please

manuelm:
No idea why github says the build failed. both builds are fine. @markus2330 any idea?

markus2330:
jenkins build bindings please

markus2330:
for whatever strange reason the ssh plugin suddenly needed write access to the parent folder to which it should copy files to (obvious given the "Auth fail" message). It should work now.

But the build job for the bindings should be not effected by this problem.

manuelm:
Thanks for looking into this.

The bindings build was fine. The auto triggered merge build took longer than the bindings build.

src/bindings/swig/python3/examples/example_keyset.py
```diff
@@ -39,3 +39,18 @@
 print("You asked for slices? You get slices:")
 print("  KeySet1[1:3]={0}".format([ str(k) for k in ks1[1:3] ]))
 print("")
+
+print("We can create shallow copies and remove keys without affecting other keysets:")
+ks2 = kdb.KeySet(ks1.dup())
+# or ks2 = copy.copy(key1)
```
markus2330: should this be ks2 = copy.copy(ks1)?

src/bindings/swig/python3/kdb.i
markus2330: Will you move this code to C++-binding in this pull request?

markus2330:
Thanks for the examples!

markus2330:
Was there a specific problem with the triggering of the buildings?

manuelm:
```
Setting commit status on GitHub for https://github.com/ElektraInitiative/libelektra/commit/5f5ed6b277a1b3b9127a9e1d88bea15f8c4f0fe1
ERROR: Publisher 'Set build status on GitHub commit' aborted due to exception: 
java.io.IOException: {"message":"No commit found for SHA: 5f5ed6b277a1b3b9127a9e1d88bea15f8c4f0fe1","documentation_url":"https://developer.github.com/v3/repos/statuses/"}
```

??

PR #245, opened 2015-08-08
doc: use @retval more

pinotree:
Switch from `@return` to `@retval` in more places, so return values are properly marked.


markus2330:
Thank you! Its always good to see documentation improved!

PR #246, opened 2015-08-08
Switch from __FUNCTION__ to __func__

pinotree:
`__FUNCTION__` is a non-standard extension, and GCC 5 starts warning about it in pedantic mode; instead, use `__func__` which is standard since C99.


markus2330:
Thanks! I did not try to compile it with GCC 5 up to now.

Issue #233, opened 2015-07-13
qt-gui: issues with searching


markus2330:
There are various issues with the searching feature.

1.) user/guitest/a/b is not found when searched for guitest

2.) searching for guitest/a/b yields nothing (is / allowed?)

3.) too many hits causes problems with the scroll bar (e.g. when searching for "b"). When scrolling around the hits suddenly disappear.

Some errors that occur during searching (maybe unrelated):
```
qrc:/qml/KeyWindow.qml:21: TypeError: Cannot read property 'path' of null
qrc:/qml/main.qml:812: TypeError: Cannot call method 'get' of undefined
```

Feel free to split the issues, I am not sure if and how they relate to each other.

0003088:
1) That's strange, I have to investigate.
2) No, slashes are not supported atm, but could be implemented of course. Maybe if the user enters a key with slashes the result contains only the expected key if present and no further keys that include parts of the name? Like an exact search? Or should the string be split and the result contain all keys that have one of the parts in their name?
3) The problem seems to be that the model is still populated with keys while it is already shown to the user, looks like this confuses qt. The only quick solution I can think of atm is presenting the user some kind of progress bar while the search is performed and to add the model to the view only after it is fully populated?

0003088:
1) and 2) should be fixed, the reason I suggested as explanation for 3) is not correct, the issue seems to be caused by some caching problem in the list view or something. Still need to find some solution.

markus2330:
Great! Thanks for your efforts!

Hope to see the other fix, too.

markus2330:
Any progress on the third issue? Should we open a new bug for it?

0003088:
> Any progress on the third issue? Should we open a new bug for it?

I think a new bug might be a good idea. I'm pretty sure the issue is a problem in Qt, I will file a bug report with them.

PR #266, opened 2015-08-31
qt-gui 0.0.8

### LINKED ISSUE #235: qt-gui: rewrites all configuration files
Due to internal handling in qt-gui all keys are marked as not synchronized. As such all configuration files are rewritten.

This is a major problem for many usecases, e.g. when the user of the gui does not have access permissions to all files.

Currently, workarounds exist for version, uname and constants plugins.

The first step would be to find all places where keys are changed unnecessarily. (e.g. string and meta data is reapplied even though it has not changed and places where keys are duplicated)

The solution does not necessarily only effect qt-gui, maybe we can make keys more intelligent so that they only marked as synchronized when it is really necessary. E.g. setting the same string again could be detected.

0003088:
* made sure keys can only be renamed if the new name/value/metadata is different from the existing ones

* fixed 1) and 2) of #233 

* fixed #235, is the current situation with duplication ok (in Confignode::setName() etc) or should this be changed?

* fixed qml warning when deleting key


markus2330:
Very nice, looks like a breakthrough regarding #235.
In some tests it only wrote modified configuration!

It seems, however, to not correctly detect any conflicts anymore, because an unconditional kdb.get() follows an unconditional kdb.set(), and this will always succeed. Is this on purpose until the merging code is fixed? But you do not have to worry about that, its trivial to fix.

Additionally, outside changes are not pulled in anymore. (I will have a look at that, too)

The initial startup window size seems to be quite large for me. Or is this some local configuration?

And I saw a qml TypeError:

   qrc:/qml/main.qml:688: TypeError: Property 'get' of object  is not a function

Is this a problem?

But anyway, thanks for the update!

src/tools/qt-gui/qml/KeyWindow.qml
```diff
@@ -116,13 +116,13 @@ BasicWindow {
 					metaValueField.placeholderText: valuePlaceHolder
 
 					metaNameField.onTextChanged:  {
-						if(metaName !== metaNameField.text){
+                        if(metaName.trim() !== metaNameField.text.trim()){
```
markus2330: Is this trimming useful? Doesn't that mean metaNames cannot start and end with spaces?

src/tools/qt-gui/src/confignode.cpp
```diff
@@ -104,15 +104,17 @@ void ConfigNode::setName(const QString& name)
...
 	if(!m_key)
 		m_key = Key(m_path.toStdString(), KEY_END);
-
+	else
+		m_key = m_key.dup();
 
 	try{
-		m_key.setBaseName(name.toStdString());
+        if(m_key.getBaseName().compare(name.toStdString()) != 0)
+            m_key.setBaseName(name.toStdString());
```
markus2330: Space/Tabs seem to be inconsistent with code above.

src/tools/qt-gui/src/confignode.cpp
```diff
@@ -126,9 +128,13 @@ void ConfigNode::setValue(const QVariant& value)
 {
 	if(!m_key)
 		m_key = Key(m_path.toStdString(), KEY_END);
+	else
+		m_key = m_key.dup();
 
-	m_key.setString(value.toString().toStdString());
-	m_value = value;
+    if(m_key.getString().compare(value.toString().toStdString()) != 0){
```
markus2330: please use enter before {

src/tools/qt-gui/src/treeviewmodel.cpp
```diff
@@ -540,73 +544,79 @@ void TreeViewModel::synchronize()
...
-
-	MergeResult result = merger.mergeKeySet(MergeTask(BaseMergeKeys(base, root),
-													  OurMergeKeys(ours, root),
-													  TheirMergeKeys (theirs, root),
-													  root));
-	if (!result.hasConflicts ())
+	try
 	{
-		resultKeys.append(result.getMergedKeys());
+		kdb.set(ours, root);
```
markus2330: set ours before merging is done? theirs is ignored?

src/tools/qt-gui/src/treeviewmodel.cpp
```diff
@@ -540,73 +544,79 @@ void TreeViewModel::synchronize()
...
 	}
-	else
+	catch (KDBException const& e)
 	{
```
markus2330: unreachable code

markus2330:
Actually, the necessary changes were quite minimal. The design with ConfigNode seems to be very good.

Btw. would be interesting if the memory consumption now decreased noticeable again.

src/tools/qt-gui/src/confignode.cpp
0003088: Sure.

src/tools/qt-gui/src/treeviewmodel.cpp
0003088: Wasn't that the point of the whole thing? Set ours and in case of conflict merge the sets? Shouldn't that be the same as merging without conflicts?

src/tools/qt-gui/src/treeviewmodel.cpp
0003088: Will change accordingly

src/tools/qt-gui/src/treeviewmodel.cpp
markus2330: The most important thing was to get rid of wrong duplications which is done by this merge request.

We should do the same for the merging and not build a workaround.

src/tools/qt-gui/qml/KeyWindow.qml
0003088: I wanted to avoid that a white space causes an edit action, but was not aware that keynames with whitespaces are desired.

src/tools/qt-gui/src/confignode.cpp
0003088: Reinstalled QtCreator and missed to change the default settings

src/tools/qt-gui/qml/KeyWindow.qml
markus2330: You are right that most users spaces enter spaces unintentionally. (Especially the ones which will be removed with trim)

But they are valid parts of key names and as such there should be *some way* to enter them. Silently trimming them seems a bit dangerous. Maybe a popup? Or highlight them to reduce the chance that they are unintentional?

markus2330:
Thank you, good job!

PR #275, opened 2015-09-09
WIP: Draft for crypto plugin using libgcrypt

petermax2:
I implemented an interface to the gcrypt library enabling the Elektra user to encrypt and decrypt binary values of arbitrary length. The plugin itself does not do anything yet. There are just the test cases and the code for encryption and decryption.

There is no support for string values yet, as they need another setter function. I will fix this in the future by adding meta-data to the key as required.


manuelm:
instead of adding FindLibGcrypt cmake module, wouldn't FindPkgConfig/pkg_check_modules be sufficient?

petermax2:
@manuelm it seems like pkg-config does not find libgcrypt:

```pkg-config --libs lgcrypt```
```Package lgcrypt was not found in the pkg-config search path.```
```Perhaps you should add the directory containing `lgcrypt.pc'```
```to the PKG_CONFIG_PATH environment variable```
```No package 'lgcrypt' found```

I tried it with ```libgcrypt```, ```gcrypt``` and ```lgcrypt```.  Also I have no "pc" file for libgcrypt under ```/usr/lib64/pkgconfig```.

**Comment by manuelm** [comment]
Yeah, looks like upstream doesn't want to move to pkgconfig:
`Upstream-Status: Rejected [upstream have indicated they don't want a pkg-config dependency]`

File: src/liberror/specification [hunk 1]
```diff
@@ -648,3 +648,27 @@ severity:normal
 ingroup:plugin
 module:storage
 see:9 75 109
+
+number:111
+description:Libgcrypt failed to initialize due to a version mismatch.
```
markus2330: also here you can easily abstract from libgcrypt. libgcrypt can be mentioned in the reason (e.g. "libgrypt 11 != libgrypt 20")

src/liberror/specification
```diff
@@ -648,3 +648,27 @@ severity:normal
...
+severity:normal
+ingroup:plugin
+module:crypto
+
+number:112
+description:Failed to set up the crypto handle correctly.
```
markus2330: that a handle is used is interna.

What about the other initialization errors?

src/liberror/specification
```diff
@@ -648,3 +648,27 @@ severity:normal
...
+severity:normal
+ingroup:plugin
+module:crypto
+
+number:113
+description:Encryption failed due to an error reported by the backend.
```
markus2330: the word "backend" is quite overloaded. Do you mean crypto-backend?

src/liberror/specification
```diff
@@ -648,3 +648,27 @@ severity:normal
...
+severity:normal
+ingroup:plugin
+module:crypto
+
+number:114
+description:Decryption failed due to an error reported by the backend.
```
markus2330: Is there some way to distinguish between password wrong and any other decryption error? Is it written in the reason?

src/plugins/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+find_package (LibGcrypt)
+
+if (LIBGCRYPT_FOUND)
+	include(LibAddPlugin)
```
markus2330: please indent the if and else body by one space

src/plugins/crypto/README.md
```diff
@@ -0,0 +1,58 @@
...
+The encryption and decryption of values is a straight forward process, once the key and IV are supplied.
+The key-derivation process in a library is a bit tricky.
+
+The crypto plugin itself can hold configuration data about the order of the applied key derivation functions.
+For example, if a password is provided by a meta-key, it will be used, otherwise we look for a key file.
+If no key file has been configured, we try to trigger the PGP agent, etc.
+
+The following example configuration illustrates this concept:
+
+	system/elektra/crypto/config/key-derivation/#0#meta
```
markus2330: the #NR#REF was needed to refer back to other array entries. I do not think it is needed here.
So better:
```
system/elektra/crypto/config/key-derivation/#0 = meta
```
(the value of the key is meta, #0 would be highest priority)

src/plugins/crypto/README.md
```diff
@@ -0,0 +1,58 @@
...
+	system/elektra/crypto/config/key-derivation/#1#file
+	system/elektra/crypto/config/key-derivation/#2#agent
+	system/elektra/crypto/config/key-file/#0#path=~/.elektra/id_aes
```
markus2330: similar here:
```
system/elektra/crypto/config/key-file/path/#0
```
so `system/elektra/crypto/config/key-file/ path` is an array with many different path names

src/plugins/crypto/README.md
```diff
@@ -0,0 +1,58 @@
...
+	system/elektra/crypto/config/key-file/#1#path=/etc/elektra/id_aes
+
+Since it does not make sense to encrypt every key in a KeySet, only Keys marked with a certain
```
markus2330: I don't know if it does not make sense, but it certainly makes sense to have keys that contain public information not encrypted.

src/plugins/crypto/README.md
```diff
@@ -0,0 +1,58 @@
...
+meta-key will be considered for encryption/decryption.
+
+## Examples ##
+
+TBD
```
markus2330: Examples would be important to get an idea how it should work.

src/liberror/specification
```diff
@@ -648,3 +648,27 @@ severity:normal
...
+severity:normal
```
markus2330: I am unsure if failed decryption always should be an error. What if someone only wants access to a not encrypted key next to an encrypted one?

Maybe passing out the encrypted string makes sense then (because the application is not interested anyway).

Most likely both warning and error makes sense for *de*cryption.

src/plugins/crypto/crypto.c
```diff
@@ -0,0 +1,129 @@
...
+			keyNew ("system/elektra/modules/crypto/exports/get", KEY_FUNC, elektraCryptoGet, KEY_END),
+			keyNew ("system/elektra/modules/crypto/exports/set", KEY_FUNC, elektraCryptoSet, KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos/version", KEY_VALUE, PLUGINVERSION, KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos", KEY_VALUE, "Information about crypto plugin is paced in the keys below", KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos/author", KEY_VALUE, "Peter Nirschl <peter.nirschl@gmail.com>", KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos/licence", KEY_VALUE, "BSD", KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos/needs", KEY_VALUE, "", KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos/provides", KEY_VALUE, "", KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos/placements", KEY_VALUE, "postgetstorage presetstorage", KEY_END),
+			keyNew ("system/elektra/modules/crypto/infos/description", KEY_VALUE, "filter plugin for cryptographic operations", KEY_END),
```
markus2330: readme_crypto not embedded here (generated from REAME.md)

src/plugins/crypto/gcrypt_operations.c
```diff
@@ -0,0 +1,243 @@
...
+}
+
+int elektraCryptoGcryInit()
+{
+	if (!gcry_check_version(GCRYPT_VERSION))
+	{
+		ELEKTRA_SET_ERRORF(111, NULL, "Libgcrypt version check failed, looking for version: %s", GCRYPT_VERSION);
+		return (-1);
+	}
+	gcry_control(GCRYCTL_DISABLE_SECMEM, 0);
```
markus2330: does this have global effects?

src/plugins/crypto/gcrypt_operations.c
```diff
@@ -0,0 +1,243 @@
...
+		gcry_cipher_close(*handle);
+		free(handle);
+	}
+}
+
+int elektraCryptoGcryInit()
+{
+	if (!gcry_check_version(GCRYPT_VERSION))
+	{
+		ELEKTRA_SET_ERRORF(111, NULL, "Libgcrypt version check failed, looking for version: %s", GCRYPT_VERSION);
```
markus2330: With NULL as argument nothing will happen. You need to supply a key where the warning/error is attached.

src/plugins/crypto/crypto.c
```diff
@@ -0,0 +1,129 @@
...
+{
+	return 1;
+}
+
+/**
+ * @brief initialize the crypto backend
+ * @retval 1 on success
+ * @retval -1 on failure
+ */
+int elektraCryptoInit()
```
markus2330: This should also happen in elektraCryptoHandleCreate.
Otherwise API is much nicer.

src/plugins/crypto/crypto.c
```diff
@@ -0,0 +1,129 @@
...
+{
+	return elektraCryptoGcryInit();
+}
+
+/**
+ * @brief clean up the crypto backend
+ *
+ * Some libraries may need extra code for cleaning up the environment.
+ */
+void elektraCryptoTeardown()
```
markus2330: The same plugin can be instanciated several types. Global teardown is not possible, you do not know if you are last.

src/plugins/crypto/crypto.h
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+#ifndef ELEKTRA_PLUGIN_CRYPTO_H
+#define ELEKTRA_PLUGIN_CRYPTO_H
+
+#include <kdbplugin.h>
+#include <stdio.h>
+
+// We may support other libraries in the future so we make the crypto-handle exchangeable
+#include <gcrypt.h>
```
markus2330: Its easy to hide gcrypt.h from here. See kdb.h.in how to do it.

src/plugins/crypto/gcrypt_operations.c
```diff
@@ -0,0 +1,243 @@
...
+	if(valueLen % ELEKTRA_CRYPTO_GCRY_BLOCKSIZE == 0)
+	{
+		outputLen = valueLen / ELEKTRA_CRYPTO_GCRY_BLOCKSIZE;
+	}
+	else
+	{
+		outputLen = (valueLen / ELEKTRA_CRYPTO_GCRY_BLOCKSIZE) + 1;
+	}
+	outputLen *= ELEKTRA_CRYPTO_GCRY_BLOCKSIZE;
+	output = (unsigned char*)malloc(outputLen);
```
markus2330: use elektraMalloc instead of malloc. Is cast necessary?

src/plugins/crypto/gcrypt_operations.c
```diff
@@ -0,0 +1,243 @@
...
+	if(output == NULL)
+	{
+		ELEKTRA_SET_ERROR(87, NULL, "Memory allocation failed");
```
markus2330: no key provided -> no error attached

src/plugins/crypto/gcrypt_operations.h
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+#ifndef ELEKTRA_PLUGIN_GCRYPT_OPERATIONS_H
+#define ELEKTRA_PLUGIN_GCRYPT_OPERATIONS_H
+
+#include <kdb.h>
+
+#define ELEKTRA_CRYPTO_GCRY_KEYSIZE (32)
+#define ELEKTRA_CRYPTO_GCRY_BLOCKSIZE (16)
+
```
markus2330: gcrypt.h + internal definition of handle could be here

src/plugins/crypto/testmod_crypto.c
```diff
@@ -0,0 +1,224 @@
...
+	}
+	return 0;
+}
+
+void test_init()
+{
+	succeed_if( elektraCryptoInit() == 1, "crypto initialization failed" );
+}
+
+void test_handle_init()
```
markus2330: all tests should be static

src/plugins/crypto/testmod_crypto.c
```diff
@@ -0,0 +1,224 @@
...
+		{
+			return 1;
+		}
+	}
+	return 0;
+}
+
+void test_init()
+{
+	succeed_if( elektraCryptoInit() == 1, "crypto initialization failed" );
```
markus2330: are other tests self contained?

markus2330:
What about values encrypted with different keys?

markus2330:
add to whitelist

markus2330:
jenkins start build please

markus2330:
The build fails because of a memleak within libgcrypt.so. I might be a problem within libgcrypt.so (then just add something to tests/valgrind.suppression) or it might be a problem in how libgcrypt.so is used.

petermax2:
@markus2330  I considered your feddback. Thank you! However I somehow messed up my git history, so I will reopen this merge request.

PR #282, opened 2015-09-14
Fix GI binding for Python3

davispuh:
Currently on Arch Linux it fails with
```
CMake Error at src/bindings/gi/python3/cmake_install.cmake:44 (file):
  file INSTALL cannot find
  "/mnt/GitHub/ElektraInitiative/libelektra/src/bindings/gi/python3/python/gi".
  src/bindings/gi/cmake_install.cmake:61 (include)
  src/bindings/cmake_install.cmake:41 (include)
  src/cmake_install.cmake:42 (include)
  cmake_install.cmake:40 (include)
```

this PR fixes that.


manuelm:
thanks

PR #286, opened 2015-09-15
Python plugin

manuelm:
This is an initial PR mainly intended for review work. Please do not pull this in yet.

open todos:
- [x] python2 build environment support


manuelm:
@markus2330 I tried to add support for python2 (the plugin supports python2) but failed. I tried to add a 2nd add_plugin call and make use of ELEKTRA_VARIANT, but this way I'm missing the generated entry in exported_symbols.

Adding a separate mainly empty python2 directory and `#include`-ing the cpp file (which is the better idea anyway) however failed because I was unable to generate the expected symbols using c macros.

markus2330:
What do you mean with not pulling in? Don't you want it to be part of this release? If not, I will review it immediately after the release is done.

For variants you need `ELEKTRA_PLUGIN_FUNCTION(resolver, open)` also for exporting, see src/plugins/resolver.
The problem you had with `#include`-ing the cpp file is exactly the same. All exported names of every plugin must be different, it does not matter if you use the variants or not. Variants make it compact. What is easier, is a point of view.

markus2330:
If the variants are too complex for you, you can also make it like it is done in yajl: it compiles with different major versions, but it does not allow both versions at once. I can change the code to variants, then. (Its nearly the same anyway)

manuelm:
No, this is for the next release and your university project.

I'm actually using `ELEKTRA_PLUGIN_FUNCTION` but as explained this won't generate the required reference in exported_symbols.{h,c}. As a consequence it's not possible to load the mangled plugin.

markus2330:
Ok, I will review it then. ELEKTRA_PLUGIN_FUNCTION obviously only works with C-functions. Use `extern "C"`

manuelm:
I'm using `extern "C"` :smile:

src/plugins/python/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
...
+# - it's impossible to compile this plugin twice
+# - however it's also impossible to create a python2 plugin and include the
+#   cpp files of this plugin because we also need to mangle the exported symbols
+# - although elektra has support for symbol mangling (ELEKTRA_VARIANT) we can't
+#   make us of this because the exported symbols will also be generated by the
+#   build environment for the elektra-full build (magic is happening in exported_symbols)
+# - and it's impossible to remove the plugin from elektra-full
+
+find_package(PythonLibs 3)
+find_package(SWIG 2)
```
markus2330: Is this searching needed again? Depending on CMake order, SWIG_FOUND and others might already be present.

If this does not work, find_PACKAGE(SWIG) should be consistent with src/bindings/CMakeLists.txt, i.e. done at one place.

src/plugins/python/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
+# NOTE: this plugins requires python 3 although it's compatible with python2
+# because of the following build systems issues:
+# - it's impossible to compile this plugin twice
+# - however it's also impossible to create a python2 plugin and include the
+#   cpp files of this plugin because we also need to mangle the exported symbols
+# - although elektra has support for symbol mangling (ELEKTRA_VARIANT) we can't
+#   make us of this because the exported symbols will also be generated by the
+#   build environment for the elektra-full build (magic is happening in exported_symbols)
+# - and it's impossible to remove the plugin from elektra-full
```
markus2330: why is this impossible? Maybe its even useful to exclude them, e.g. kdb-full starting the java interpreter causes a lot of trouble.

src/plugins/python/CMakeLists.txt
manuelm: Afaik cmake only export it's variables to subfolders. Furthermore it's possible to not compile bindings but still compile the plugin (although it's currently disabled because tests won't work)

src/plugins/python/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
+# NOTE: this plugins requires python 3 although it's compatible with python2
+# because of the following build systems issues:
+# - it's impossible to compile this plugin twice
+# - however it's also impossible to create a python2 plugin and include the
+#   cpp files of this plugin because we also need to mangle the exported symbols
+# - although elektra has support for symbol mangling (ELEKTRA_VARIANT) we can't
+#   make us of this because the exported symbols will also be generated by the
```
markus2330: Ok, I understand. So you get link errors because of duplicated symbols that are generated by SWIG. Why not putting the SWIG stuff in an anonymous namespace?

src/plugins/python/python.cpp
```diff
@@ -0,0 +1,401 @@
...
+	PyList_Append(sysPath, pyPath);
+	Py_DECREF(pyPath);
+	PyGILState_Release(gstate);
+	return 1;
+}
+
+static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+static unsigned open_cnt = 0;
+
+int ELEKTRA_PLUGIN_FUNCTION(Python, Open)(ckdb::Plugin *handle, ckdb::Key *errorKey)
```
markus2330: should be exern "C"?

src/plugins/python/CMakeLists.txt
manuelm: No, that's wrong. SWIG is no problem. libelektra/exported_symbols.c is

src/plugins/python/python.cpp
manuelm: It is. The whole plugin is.

src/plugins/python/python.cpp
```diff
@@ -0,0 +1,401 @@
...
+	PyObject *pModule = Python_ImportModule(bname);
+	if (pModule == NULL)
+	{
+		std::cerr << "ERROR: Failed to import python script "
+			<< keyString(script) << std::endl;
+		PyErr_Print();
+		free(tmpScript);
+		return -1;
+	}
+	free(tmpScript);
```
markus2330: should use elektraFree

src/plugins/python/python.cpp
```diff
@@ -0,0 +1,401 @@
...
+	if (kdbModule == NULL)
+	{
+		std::cerr << "ERROR: Failed to import kdb module" << std::endl;
+		PyErr_Print();
+		return -1;
+	}
+	Py_XDECREF(kdbModule);
+
+	/* extend sys path */
+	char *tmpScript = strdup(keyString(script));
```
markus2330: should use elektraStrDup

basically all memory allocations should use the elektra wrappers.

Wrappers are useful for profiling and memory testing (overwrite elektraMalloc to return 0)

src/plugins/python/CMakeLists.txt
manuelm: I already tried to explain this in a comment in this PR. The referenced symbol in kdb_exported_syms from libelektra/exported_symbols.c is cause the most headache.

I also tried to find a good way to exclude the plugin from -full/-static without any luck. Tried that at least for 4 hours.

src/plugins/python/python.cpp
```diff
@@ -0,0 +1,401 @@
...
+	{
+		Py_Initialize();
+		PyEval_InitThreads();
+		if (!Py_IsInitialized())
+		{
+			pthread_mutex_unlock(&mutex);
+			return -1;
+		}
+	}
+	open_cnt++;
```
markus2330: Isn't such a counter integrated for Py_IsInitialized() anyway?

manuelm: No. Counter is need for destruction. Py_Finalize will immediately shutdown python.

src/plugins/python/CMakeLists.txt
markus2330: Comments are good, thanks. But I still wonder what the difference from this plugin to others is. Seems like I have to find out myself.

src/plugins/python/testmod_python.c
```diff
@@ -0,0 +1,145 @@
...
+	init(argc, argv);
+
+	test_variable_passing();
+	test_two_scripts();
+
+	printf("NOTE: The following errors are intended. We're testing error conditions!\n");
+	test_fail();
+	test_wrong();
+
+	printf("\ntest_hosts RESULTS: %d test(s) done. %d error(s).\n", nbTest, nbError);
```
markus2330: should be test_python

src/plugins/python/testmod_python.c
```diff
@@ -0,0 +1,145 @@
...
+	Plugin *plugin = elektraPluginOpen(ELEKTRA_PLUGIN_NAME, modules, conf, 0);
+	succeed_if (plugin == NULL, "python plugin shouldn't be loadable"); \
+
+	elektraModulesClose(modules, 0);
+	ksDel(modules);
+}
+
+int main(int argc, char** argv)
+{
+	printf("MOUNT       TESTS\n");
```
markus2330: should be PYTHON TESTS

src/plugins/python/testmod_python.c
```diff
@@ -0,0 +1,145 @@
...
+
+	KeySet *modules = ksNew(0, KS_END);
+	elektraModulesInit(modules, 0);
+
+	KeySet *conf = ksNew(1,
+		keyNew("user/script", KEY_VALUE, srcdir_file("python/python_plugin_wrong.py"), KEY_END),
+		KS_END);
+
+	Plugin *plugin = elektraPluginOpen(ELEKTRA_PLUGIN_NAME, modules, conf, 0);
+	succeed_if (plugin == NULL, "python plugin shouldn't be loadable"); \
```
markus2330: maybe check for error in errorKey (unused argument of elektraPluginOpen)

src/plugins/python/testmod_python.c
```diff
@@ -0,0 +1,145 @@
...
+	PyGILState_STATE gstate = PyGILState_Ensure();
+	PyObject *sysPath = PySys_GetObject((char *)"path");
+	PyObject *pyPath  = PyUnicode_FromString(path);
+	PyList_Append(sysPath, pyPath);
+	Py_DECREF(pyPath);
+	PyGILState_Release(gstate);
+	return 1;
+}
+
+void init_python()
```
markus2330: Why not static? Why having this function at all? Does the initialization in open() work?

manuelm: static is missing.
The function is part of the workaround explained in CMakeLists.

src/plugins/python/testmod_python.c
```diff
@@ -0,0 +1,145 @@
...
+{
+	setenv("PYTHONDONTWRITEBYTECODE", "1", 1);
+	if (!Py_IsInitialized())
+		Py_Initialize();
+	Python_AppendToSysPath(".");
+}
+
+// test simple variable passing
```
markus2330: You can print this text when test starts.

src/plugins/python/python.cpp
```diff
@@ -0,0 +1,401 @@
...
+}
+
+int ELEKTRA_PLUGIN_FUNCTION(Python, Error)(ckdb::Plugin *handle, ckdb::KeySet *returned,
+	ckdb::Key *parentKey)
+{
+	int ret = 0;
+	moduleData *data = static_cast<moduleData *>(elektraPluginGetData(handle));
+	if (data != NULL)
+	{
+		/* call python function */
```
markus2330: calling python functions looks identical, why not wrap this in a function for Get, Set, Error?

src/plugins/python/python.cpp
```diff
@@ -0,0 +1,401 @@
...
+static unsigned open_cnt = 0;
+
+int ELEKTRA_PLUGIN_FUNCTION(Python, Open)(ckdb::Plugin *handle, ckdb::Key *errorKey)
+{
+	KeySet *config = elektraPluginGetConfig(handle);
+	Key *script = ksLookupByName(config, "/script", 0);
+	/* success if no script to execute */
+
+	if (script == NULL || keyString(script) == NULL)
+		return 0;
```
markus2330: It does not distinguish between: user messed up and did not provide script and (the necessary) bootstrap phase where no script is provided.

    if (ksLookupByName(config, "/module", 0))  return 0;
    if (script == NULL || keyString(script) == NULL) //set error+ return -1

src/plugins/python/python.cpp
markus2330: Nice, should have looked before :-)

src/plugins/python/CMakeLists.txt
markus2330: Yes, if you do `SET()` it only goes to subfolders. But the find_package use cache variables, visible outside, too.

The main critique is that the searching is different, one might find swig 2, the other swig 3.

src/plugins/python/testmod_python.c
manuelm: Please elaborate. The python c-plugin itself doesn't set the errorkey parameter. And no test does. What's the purpose of checking something which won't change?

src/plugins/python/CMakeLists.txt
manuelm: swig 2 is the minimum version. swig 3 is fine too (I'm only using swig 3) :smile:

src/plugins/python/python.cpp
markus2330: Yes, we discussed this a lot. A flag for not doing Py_Finalize would be nice.

markus2330:
Looks very nice! Thanks for putting so much effort in python2 and improving cmake macros, e.g. add_plugin.

Basically only minor stuff+build server should run.

src/plugins/python/testmod_python.c
markus2330: You should set an error, if an error happens. There should be no cerr, but only ELEKTRA_SET_ERROR and ELEKTRA_ADD_WARNING..

markus2330: For PyErr_print: if there is no function to print to a string, simply configure using "/print" if the error should be printed or not. (See jni binding). Per default a library should never print something.

manuelm:
review notes done. python2 still missing. build error war unrelated to this PR

markus2330:
Looks, much better, thanks!

I am missing a src/plugins/python/README.md?
Where does readme_python.c comes from?

manuelm:
It's in my local directory because it contains top secret information :smile_cat:

manuelm:
erm. python 2 support is missing and this PR was specifically not intended for this release.

markus2330:
> No, this is for the next release

I think this statment is crystal clear? Why change of mind? Did you forget a *not*?

If I do not mention it in the release notes, it is, however, equal to not being published. Is this ok for you then?

manuelm:
Uff, I really forgot the *not*... Sorry for confusion.

Can we remove the python plugins from the `PLUGINS=ALL`-target instead? So the python plugins don't get pulled in unless selected by hand? I've at least one more commit, but it's a rather small one.

markus2330:
> Can we remove the python plugins from the `PLUGINS=ALL`-target instead?

I do not really like the idea, but its your plugin ;)

9583f5ced4c2f3c56f649c0efb91d4d144a151a0

PR #294, opened 2015-09-17
Python fixes

manuelm:
The first two commits are optional for this release. I don't care much about them.

The last however IS mandatory.


manuelm:
jenkins build bindings please

manuelm:
And I really don't like these kind of modifications without doing a PR

manuelm:
latest commit removes duplicated test data for python2.

If you want to pull 7cfead in without the other commits I can create a single PR for you

manuelm:
And the last one, which removes the duplicated README.md

Apart from CMakeLists.txt python2 plugin is only a small stub now.

markus2330:
Mmh, I am afraid the python plugin story does not terminate.

Its strange to need to set the key exactly to `0` without any docu saying so even though not doing so most likely leading to crashes (if library is loaded multiple times, having libgetenv, using python somewhere else and so on,...).

The debian package already uses "python2" as path for testdata and is finally working.

I dont think that the assumption that python2 programs are binary compatible to python3 programs is a valid one. (Even though it can be done) In the proposed situation it is useless to have a separate folder for the python2 plugin at all -- variants should be used instead. But then the release will be delayed even longer.

Can you compile a minimal merge request with only providing important bug fixes?

manuelm:
> Its strange to need to set the key exactly to 0 without any docu saying so even though not doing so most likely leading to crashes (if library is loaded multiple times, having libgetenv, using python somewhere else and so on,...).

The shutdown flag can be set to:
0 ... no engine shutdown / default
1 (or any other value) ... engine shutdown

This is exactly the reason why a *value* makes sense here. To force the behavior without worrying about default values. Changing the default value to *no shutdown* is no problem for me. It's changing the flags behavior which I don't like.

> The debian package already uses "python2" as path for testdata and is finally working.

I have no idea what you're talking about. The testdata either get's rewritten if it's executed from the build directory or not touched at all because the install phase is copying the test data from python to python2

> I dont think that the assumption that python2 programs are binary compatible to python3 programs is a valid one. (Even though it can be done) In the proposed situation it is useless to have a separate folder for the python2 plugin at all -- variants should be used instead. 

First of all python progams can never be binary compatible. There's no binary data, it's a scripting language. Second, python2 bindings are a subset of python3. So they *will* be compatible. Third, we can always expand the rewrite magic to first look in the origin directory.

And I already explained why variants are not working. In short: It's broken.

markus2330:
One more thing about 7cfeada 

I am sorry to have changed it without your consent, that its clearly wrong and I would not have done it if we are not weeks overdue the release and it crashed in every situation I tried to use it with shutdown (except for the unit tests where a workaround is present). Additionally, its not documented, so the described behaviour actually was not changed.

markus2330:
Shutdown flag: are you sure that your described behaviour as implemented with the `!!` expression?

In a small test the shutdown was true if the key was absent, or set to "". It seems, however, we want to achieve the same. Please make the implementation less obfuscated. Its funny its a one-line, but thats not really something we want to achieve here. And please describe it in the README.

Ok  if the python2 directory still exists, I did not look into it.

Point 3: we are discussing irrelevant terminology here. Just because ABI == API does not mean ABI does not exist.

About variants: Lets discuss after the release how to repair it.

manuelm:
> Shutdown flag: are you sure that your described behaviour as implemented with the !! expression?

Yes, I'm sure. And for me it's easy to read. The `!!` is just an expression to force a boolean in c. Since strcmp returns `!=0` if the string doesn't match, the expression is the boolean of `shutdown = (x != '0')`.

Shall I change the default value of shutdown from `No` to `Yes`, or not?

> Point 3: we are discussing irrelevant terminology here. Just because ABI == API does not mean ABI does not exist.

I still don't get it. The python scripts/plugins will be portable, at least if you don't add python3 specifics to it on purpose. All c files of python2 are just a `#include` statement. So they are already the same.
As long as these assumptions hold true, my commit fully makes sense.

markus2330:
Important is that the default, i.e. no key, is *no* shutdown.
Currently: ksLookupByName returns 0, keyString is "(null)", strcmp() will be some negative number, !! will make it true and this will lead to a shutdown? Or is the variable name misleading?

Thats not what you said in the description above:
> The shutdown flag can be set to:
> 0 ... no engine shutdown / default
> 1 (or any other value) ... engine shutdown

Please document and fix it. I am ok with value-based decision. I would prefer to have "off", "no", "0" also as turned off, but for this  release (as technical preview) it does not really matter.

Also python -> python3 would be better (and consistent to swig names). But also not now please.

manuelm:
> Important is that the default, i.e. no key, is no shutdown.
Currently: ksLookupByName returns 0, keyString is "(null)", strcmp() will be some negative number, !! will make it true and this will lead to a shutdown? Or is the variable name misleading?

No the variable is not misleading. I just mixed the default value. So, once again: The current default value is `shutdown the engine`. Chosen because I prefer a clean environment.

So I'll change the default value.

> Also python -> python3 would be better (and consistent to swig names). But also not now please.

Actually I will rather re-rename the swig bindings. `re-rename` because you've renamed them a few hundred commits ago while adding the python2 stuff. Will be one of my next PR after this release.

markus2330:
Ok, so I get the default value, you get the name "python" ;)

markus2330:
Btw. I prefer a clean environment, too. But not in the case of potential crashes.

manuelm:
I've removed the commits to reduce the python2 directory. All other commits are safe to merge.

Please ping me as soon as you've tagged the release

manuelm:
jenkins build bindings please

markus2330:
Thanks for all the effort!

I will write the release notes to our internal mailinglist, too. (Consider this as ping)

markus2330:
Btw. python2 directory name, I was talking about: http://build.libelektra.org:8080/job/elektra-git-buildpackage-wheezy/297/console

manuelm:
you fixed that by adding the `install`-call to CMakeLists.txt, right? I didn't change that (except it's source)

markus2330:
Yes, I initially fixed it and broke it again by removing the python plugins from `ALL` for the release, because the debian package was built with `-DPLUGINS=ALL` but expected the python plugin to be present. Now its fine again.

PR #303, opened 2015-09-21
Draft for crypto plugin using libgcrypt

petermax2:
I implemented an interface to the gcrypt library enabling the Elektra user to encrypt and decrypt binary values of arbitrary length. The plugin itself does not do anything yet. There are just the test cases and the code for encryption and decryption.

There is no support for string values yet, as they need another setter function. I will fix this in the future by adding meta-data to the key as required.

- [x] consider the feedback from @markus2330 from #275.
- [x] move crypto header to internal header
- [x] remove duplicate entries from valgrind suppressions file
- [x] enhance metakey documentation (document how encryption changes the type to binary and how it is being restored)
- [x] move gcrypt initialization to ``elektraCryptoHandleCreate()``


src/plugins/crypto/README.md
```diff
@@ -0,0 +1,57 @@
+- infos = Information about crypto plugin is in keys below
+- infos/author = Peter Nirschl <peter.nirschl@gmail.com>
+- infos/licence = BSD
+- infos/needs = 
+- infos/provides = storage
```
markus2330: This information is wrong. (The testcases also fail because of that)

There is no real category for such a plugin up to now. Maybe filefilter?

markus2330:
The feature for encrypting strings and iterating over all keys that have some metadata is quite essential and should be trivial, don't you think?

Btw. I think its ok that the result is binary, even when the input was a string. You just have to remember somehow what it has been initially.

Or do you only want feedback?

src/plugins/crypto/README.md
petermax2: Well it's not exactly bound to any files, as you can pass configuration and content payload in-memory. Are you ok with "filter"?

petermax2:
It is simple indeed. I will implement it this evening.

The result MUST be binary, even if the value was of type String. I will add a flag in the Key's metadata.

src/plugins/crypto/README.md
markus2330: Sorry, I was thinking about a plugin that encrypts a whole file. In this case filter is good ;)

markus2330:
MUST is a strong word, you could encode it base64 or similar. But this should be done by another plugin, and not this one.

I dont think you should leak the information if it was binary or not. Instead  you should encode the bit in the encrypted value. E.g. /env/option/debug is binary in the case when the logging is to stderr, when its a string it will write to a file given by that string. You could construct other examples where this information is really confidential. Null pointer are binary and represent missing information, this should be somehow encoded in the encrypted value, too.

Please document all metadata in doc/METADATA.ini.

petermax2:
Encoding this information is a great idea. I guess I was too crazy about the metakeys :) Thanks!

petermax2:
Any ideas about the valgrind issues yet? With version ``valgrind-3.10.1`` on Fedora no errors occur. On Debian with version ``valgrind-3.10.0`` 38 problems are being reported (only jumps on uninitialized values):

	==12244== Conditional jump or move depends on uninitialised value(s)
	==12244==    at 0x4027E4: cmp_buffers (testmod_crypto.c:59)
	==12244==    by 0x4027E4: test_enc_and_dec_with_binary (testmod_crypto.c:224)
	==12244==    by 0x4027E4: main (testmod_crypto.c:274)

However the C code is:

	static int cmp_buffers(const unsigned char *b1, size_t len1, const unsigned char* b2, size_t len2)
	{
		if (len1 != len2)
		{
			return 1;
		}

		for(unsigned int i = 0; i < len1 && i < len2; i++)
		{
			// THIS IS LINE 59 -> i has been initialized in the loop head
			if (b1[i] != b2[i])
			{
				return 1;
			}
		}
		return 0;
	}

What am I missing?

markus2330:
From the code I cannot say that, it could go outside of array boundaries, but when the version later does not report an error anymore, its most likely wrongly reported. Then you can add an suppression to `tests/valgrind.suppression`. This can be done using `--gen-suppressions=all`. But why not use `memcmp` instead of the loop?

petermax2:
> But why not use memcmp instead of the loop?

At first I had other checks in mind, but now ``memcmp`` is enough. Btw. ``memcmp`` also gets valgrind to freak out about "conditional jumps on uninitialized values". After checking and re-checking the code I finally added suppressions.

@markus2330  How do I get the source of the potential memory leak? After running

	sudo ctest -T memcheck --output-on-failure -LE memleak

all test cases pass but at the end there is the output, that memory leaks have been found:

	Total Test time (real) =  55.09 sec
	-- Processing memory checking output: #
	Memory checking results:
	Potential Memory Leak - 10

Valgrind only complains about the unconditional jumps mentioned above, but no potential memory leaks in ``bin/testmod_crypto``.

markus2330:
Source can be found in DynamicAnalysis.xml in a folder below build directory/Testing.

petermax2:
Let's blame it on Libgcrypt's initialization function:

	==7686==    at 0x4C28C20: malloc (vg_replace_malloc.c:296)
	==7686==    by 0x54C7ED0: ??? (in /lib/x86_64-linux-gnu/libgcrypt.so.20.0.3)
	==7686==    by 0x54C922B: ??? (in /lib/x86_64-linux-gnu/libgcrypt.so.20.0.3)
	==7686==    by 0x5562DF1: ??? (in /lib/x86_64-linux-gnu/libgcrypt.so.20.0.3)
	==7686==    by 0x5562E6E: ??? (in /lib/x86_64-linux-gnu/libgcrypt.so.20.0.3)
	==7686==    by 0x54C7E2F: ??? (in /lib/x86_64-linux-gnu/libgcrypt.so.20.0.3)
	==7686==    by 0x54C8044: ??? (in /lib/x86_64-linux-gnu/libgcrypt.so.20.0.3)
	==7686==    by 0x4E7E1CF: elektraCryptoGcryInit (gcrypt_operations.c:28)
	==7686==    by 0x401E05: test_init (testmod_crypto.c:100)
	==7686==    by 0x401E05: main (testmod_crypto.c:251)

markus2330:
Yes, that is a clear situation, just add a suppression.

memcmp should not cause any valgrind problem, it is used often within Elektra.

petermax2:
Again these problems do not occur on Fedora 22. I guess all these issues have already been fixed but did not make it into the Debian repository yet. I will add suppressions and commit again.

petermax2:
Could you send me the suppressions for Jenkins, please?

markus2330:
Please resolve merge conflict because of error numbers. Sorry for that, its not good have PR which are pending too long. If you add the basic functionality in get/set, I can merge this PR. You can modify everything later as desired. Also note that #305 adds even more numbers and will be merged soon (so take higher numbers).

I cannot reproduce the error locally (which is strange, because I also run jessie) and do not have access to the -mm agents.

It should be easy, however, to write the suppression by hand, that might work:
```
{
   <insert_a_suppression_name_here>
   Memcheck:*
   ...
   fun:elektraCryptoGcryInit
   fun:test_init
   fun:main
}

```

petermax2:
Is it a problem if error numbers in ``specification`` are not sequential? I started by ``150`` to leave room for #305.

petermax2:
Apparently it is a problem for ``liberrer``. #305 will need its own error numbers, I'm afraid. :(

petermax2:
Sorry, the suppression did not work. I wonder what Jenkins is complaining about. Do you know which Debian version is running the build server? My Jessie VM does not report any errors any more and Fedora 22 has been fine, anyway.

markus2330:
Peter Nirschl wrote:
> Sorry, the suppression did not work. I wonder what Jenkins is complaining
> about. Do you know which Debian version is running the build server? My
> Jessie VM does not report any errors any more and Fedora 22 has been fine,
> anyway.

I also did not have errors. It should be jessie.
@manuelm Can you check valgrinds version on stable-mm?

petermax2:
I managed to find another leak (this time in ``libstdc++``). I guess it can be suppressed? 

	==6932== 72,704 bytes in 1 blocks are still reachable in loss record 1 of 1
	==6932==    at 0x4C28C50: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
	==6932==    by 0x5C391EF: ??? (in /usr/lib64/libstdc++.so.6.0.21)
	==6932==    by 0x400F639: call_init.part.0 (in /usr/lib64/ld-2.21.so)
	==6932==    by 0x400F74A: _dl_init (in /usr/lib64/ld-2.21.so)
	==6932==    by 0x4000CB9: ??? (in /usr/lib64/ld-2.21.so)
	==6932==    by 0x1: ???
	==6932==    by 0xFFF0005F2: ???
	==6932==    by 0xFFF00062B: ???
	==6932==

petermax2:
Jenkins build please

markus2330:
Seems good now regarding the build server. (only a network connection problem)

Do you have libelektragetenv active? I think this memleak results from it.

petermax2:
Yes, ``getenv``was active in my CMakeCache

markus2330:
jenkins build please

markus2330:
@petermax2 I think the problem was that "Jenkins" does not listen do you, only "jenkins" does.

It is documented in doc/GIT.md

petermax2:
@markus2330 ready to merge

markus2330:
Thanks, as discussed, there might be one more review.

doc/METADATA.ini
```diff
@@ -501,3 +501,16 @@ status= implemented
...
+status= implemented
+usedby/plugins= crypto
+type= string
+description= used to mark a key as encrypted. If value is not empty, the value has been encrypted.
+
+[crypto/encrypt]
+status= implemented
+usedby/plugins= crypto
+type= string
+description= used to mark a key for encryption. The crypto plugin is supposed to encrypt the key.
```
markus2330: can you document that it changes the semantics of "binary" (because it actually does not tell anymore if the key is actually binary)

src/plugins/crypto/crypto.c
```diff
@@ -0,0 +1,220 @@
...
+{
+	return elektraCryptoInit(errorKey);
+}
+
+/**
+ * @brief close the crypto plugin
+ * @retval 1 on success
+ * @retval -1 on failure
+ */
+int elektraCryptoClose(Plugin *handle ELEKTRA_UNUSED, Key *errorKey ELEKTRA_UNUSED)
```
markus2330: Function could be avoided.

src/plugins/crypto/crypto.c
```diff
@@ -0,0 +1,220 @@
...
+{
+	return 1;
+}
+
+/**
+ * @brief initialize the crypto backend
+ * @retval 1 on success
+ * @retval -1 on failure
+ */
+int elektraCryptoInit(Key *errorKey)
```
markus2330: for the API to be universal, it should somehow accept elektraCryptoHandle as parameter
(not important)

src/plugins/crypto/crypto.h
```diff
@@ -0,0 +1,53 @@
...
+{
+	unsigned short flags;
+	unsigned long contentLen;
+};
+
+#define ELEKTRA_CRYPTO_META_ENCRYPT ("crypto/encrypt")
+#define ELEKTRA_CRYPTO_META_ENCRYPTED ("crypto/encrypted")
+
+// We may support other libraries in the future so we make the crypto-handle exchangeable
+#include <gcrypt.h>
```
markus2330: split to other (internal) header file

tests/valgrind.suppression
```diff
@@ -62,3 +62,397 @@
...
+   match-leak-kinds: reachable
+   fun:malloc
+   obj:*libgcrypt.so.20*
+   obj:*libgcrypt.so.20*
+   obj:*libgcrypt.so.20*
+   obj:*libgcrypt.so.20*
+   fun:elektraCryptoGcryInit
+   fun:test_init
+   fun:main
+}
```
markus2330: isn't that a duplicate?

src/plugins/crypto/crypto.c
petermax2: can you give me an example, please? is there already a plugin that uses these functions with internal parameters?

src/plugins/crypto/crypto.c
petermax2: how? where would you put the initialization of gcrypt?

markus2330: I am talking about elektraCryptoClose, which is doing nothing.

src/plugins/crypto/crypto.c
markus2330: E.g. kdbOpen() returns the handle
or elektraModulesInit() takes the handle as first parameter

But I see the problem, you are already doing that in elektraCryptoHandleCreate()

Why not doing init there?

petermax2:
@markus2330 How do I provide the internal header file to the ``testmod_crypto`` build target? Having a CMake config like this:

	configure_file (
		"${CMAKE_CURRENT_SOURCE_DIR}/crypto_internal.h.in"
		"${CMAKE_CURRENT_BINARY_DIR}/crypto_internal.h"
	)

	install (FILES
		"${CMAKE_CURRENT_BINARY_DIR}/crypto_internal.h"
		DESTINATION include/${TARGET_INCLUDE_FOLDER}
	)

results in a compiling ``elektra-crypto-objects`` and a **failig** ``testmod_crypto`` build target.

markus2330:
include_directories("${CMAKE_CURRENT_BINARY_DIR}") should do the trick.

petermax2:
@markus2330  thank you for the feedback (again)! Please have a closer look at ``crypto.h`` and ``crypto_internal.h.in``. I hope the API is better now.

PR #305, opened 2015-09-21
Enum-, Conditionals- and Calculate-Plugin

tom-wa:


src/liberror/specification
```diff
@@ -660,3 +660,93 @@ description:python warning
...
+module:conditionals
+
+number:126
+description:Syntax error
+severity:error
+ingroup:plugin
+module:conditionals
+
+number:127
+description:Validation failed
```
markus2330: It would be good to have a possibility to check if any type of validation failed. We could use some inheritance schema for that.

See #307

src/plugins/calculate/README.md
```diff
@@ -0,0 +1,18 @@
...
+- infos/needs =
+- infos/provides = check
+- infos/placements = presetstorage
+- infos/description =
+
+## Introduction ##
+
+Compares a key value to a mathematical expression using polish prefix notation defined in the `check/calc` metakey. 
+Operations are `+ - / *` . Operants are keys with names relative to the parent key.
+How the values are compared is specified at the begining of the Metakey using the conditions `<, <=, ==, !=, =>, >`
```
markus2330: Why compare?

markus2330:
Can you update this PR, its difficult to read because already accepted merges are displayed, too.

src/plugins/calculate/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Examples ##
+
+`check/calc = "== + testval1 + testval2 testval3"` compares the keyvalue to the sum of testval1-3 and yields an error if the values are not equal.
```
markus2330: Inconsistent/confusing: should be `calculate` and not `calc`

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+#include "calculate.h"
+
+typedef enum{ERROR, ADD, SUB, MUL, DIV, NOT, EQU, LT, GT, LE, GE, RES, VAL, END, EMPTY}Operation;
+typedef struct{
+	double value;
+	Operation op;
+}PNElem;
+
+
+int elektraCalculateGet(Plugin *handle ELEKTRA_UNUSED, KeySet *returned ELEKTRA_UNUSED, Key *parentKey ELEKTRA_UNUSED)
```
markus2330: marked ELEKTRA_UNUSED, but is used.

src/plugins/calculate/README.md
markus2330: Additionally, how to call the plugin that will do string operations? Should it be added here?

Is `+` overloaded then, or would it have other operators?

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+#include ELEKTRA_README(calculate)
+				keyNew ("system/elektra/modules/calculate/infos/version",
+					KEY_VALUE, PLUGINVERSION, KEY_END),
+				KS_END);
+		ksAppend (returned, contract);
+		ksDel (contract);
+
+		return 1; /* success */
+	}
+	/* get all keys */
```
markus2330: wrong comment

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+	if(stackPtr < stack)
+		result.op = ERROR;
+	while(stackPtr >= stack)
+	{	
+		if(stackPtr->op == VAL && stackPtr != stack)
+		{
+			--stackPtr;
+			continue;
+		}
+		else if(stackPtr->op == VAL && stackPtr == stack)
```
markus2330: Please use `{}` here.

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+	if(stackPtr->op != VAL)
+		result.op = ERROR;
+	else
+		result.op = RES;
+	result.value = stackPtr->value;
+	return result;
+}
+static PNElem parsePrefixString(const char *prefixString, KeySet *ks, Key *parentKey)
+{
+	char *regexString = "((([[:alnum:]]*/)+[[:alnum:]]+))|([-+/<>=!{*])";
```
markus2330: Should be `const char`. Isn't the compiler saying something here?

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+	char *ptr = (char *)prefixString;
+	regex_t regex;
+	Key *key;
+	//PNElem stack[100];
+	
+	PNElem *stack = malloc(3*sizeof(PNElem));
```
markus2330: Use of magic number. Why 3?

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+		nomatch = regexec(&regex, ptr, 1, &match, 0);
+		if(nomatch)
+		{
+			break;
+		}
+		len = match.rm_eo - match.rm_so;
+		start = match.rm_so + (ptr - prefixString);
+		if(len == 1)
+		{
+			switch(prefixString[start])
```
markus2330: This statemachine could be in a separate function?

src/plugins/calculate/calculate.c
tom-wa: Going to add a constant.
A valid PPN has at least 3 elements

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+		{
+			ksRewind(ks);
+			searchKey = realloc(searchKey, len+2+strlen(keyName(parentKey)));
+			strcpy(searchKey, keyName(parentKey));
+			strcat(searchKey, "/");
+			strncat(searchKey, ptr+1, len);
+			key = ksLookupByName(ks, searchKey, 0);
+			if(!key)
+			{
+				ELEKTRA_SET_ERRORF(124, parentKey, "Couldn't find Key %s", searchKey);
```
markus2330: Give more context to the user: Why was the key searched (which meta key, which operation,...)

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+	result = doPrefixCalculation(stack, stackPtr);	
+	if(result.op != ERROR)
+		result.op = resultOp;
+	free(stack);
+	return result;
+}
+
+int elektraCalculateSet(Plugin *handle ELEKTRA_UNUSED, KeySet *returned ELEKTRA_UNUSED, Key *parentKey ELEKTRA_UNUSED)
+{
+	/* set all keys */
```
markus2330: wrong comment + add more real comments

src/plugins/calculate/calculate.c
```diff
@@ -0,0 +1,331 @@
...
+		if(!meta)
+			continue;
+		result = parsePrefixString(keyString(meta), ksDup(returned), parentKey);
+		if(result.op == ERROR)
+		{
+			return 1;
+		}
+		else if(result.op == EQU)
+		{
+			if(fabs(atof(keyString(cur)) - result.value) > 0.00001)
```
markus2330: Use of constant + epsilon should be configureable?

markus2330: Btw. from README.md it was not clear to me that it will be interpreted as floating point. That should be documented.

src/plugins/calculate/testmod_calculate.c
```diff
@@ -0,0 +1,90 @@
...
+	keyNew("user/tests/calculate/bla/val2", KEY_VALUE, "50", KEY_END),
+	keyNew("user/tests/calculate/bla/val3", KEY_VALUE, "3", KEY_END),
+		KS_END);
+	KeySet *conf = ksNew(0, KS_END);
+	PLUGIN_OPEN("calculate");
+	ksRewind(ks);
+	succeed_if(plugin->kdbSet(plugin, ks, parentKey) == (1), "error");
+	ksDel(ks);
+	keyDel(parentKey);
+	PLUGIN_CLOSE();
```
markus2330: again: you should find a more abstract and easier way to specify test cases: Only keyset+expected results should suffice.

src/plugins/conditionals/README.md
```diff
@@ -0,0 +1,28 @@
...
+Condition:  `Key` *Operation* `('String' | '1234.56' | Key | '')`
+
+Operations: `!=, ==, <, <=, =>, >`
+
+
+## Example ##
+
+`(if/this/key  != 'value') ? (then/key == some/other/key) : (or/key <= '125')` 
+
+Meaning: IF `if/this/key` NOT EQUAL TO `'value'` THEN `then/key` MUST EQUAL `some/other/key` ELSE `or/key` MUST BE LESS THAN `125`
```
markus2330: `if/` should be removed in explanation

src/plugins/conditionals/conditionals.c
```diff
@@ -0,0 +1,413 @@
...
+#include <stdlib.h>
+#include <kdberrors.h>
+#include <regex.h>
+#include <ctype.h>
+#include <math.h>
+#include "conditionals.h"
+
+typedef enum{EQU, NOT, LT, LE, GT, GE}Comparator;
+
+static int isNumber(const char *s)
```
markus2330: document what this function should do.

src/plugins/conditionals/conditionals.c
```diff
@@ -0,0 +1,413 @@
...
+{
+	char *ptr = (char *)s;
+	int ret = 1;
+	while(*ptr != '\0')
+	{
+		if(*ptr == '.' && ret == 1)
```
markus2330: What about multiple `.`? Why not use `strtol` or similar to find out if its a number?

atof is used short afterwards anyway.

src/plugins/conditionals/conditionals.c
```diff
@@ -0,0 +1,413 @@
...
+	else if(s2 == NULL)
+	{
+		retval = 1;
+	}
+	else if((ret = isNumber(s1)) && (ret2 = isNumber(s2)))
+	{
+		if(ret == 2 || ret2 == 2)
+		{
+			result = (fabs(atof(s1)-atof(s2)));
+			if(result < 0.00001)
```
markus2330: again epsilon

src/plugins/conditionals/conditionals.c
```diff
@@ -0,0 +1,413 @@
...
+		memset(compareTo, 0, endPos-rightSide);
+		strncat(compareTo, rightSide+1, endPos-rightSide-1);
+	}
+	else if(rightSide && elektraStrLen(rightSide) > 1)
+	{
+		len = keyGetNameSize(parentKey)+elektraStrLen(rightSide);
+		if(elektraRealloc((void **)&lookupName, len) < 0)
+		{
+			ELEKTRA_SET_ERROR(87, parentKey, "Out of memory");
+			result = -1;
```
markus2330: why not goto Cleanup here?

src/plugins/conditionals/conditionals.c
```diff
@@ -0,0 +1,413 @@
...
+Cleanup:
+	if(lookupName)
+		elektraFree(lookupName);
+	if(compareTo)
+		elektraFree(compareTo);
+	return result;
+}
+
+
+static int checkCondition(const char *condition, KeySet *ks, Key *parentKey)
```
markus2330: generic name, no comment

src/plugins/calculate/README.md
tom-wa: True, but actually i've already added support for setting the key value. +=, -= etc will follow soon
Validation was just my first priority. 
i think string operations should be handled by a different module?

src/plugins/calculate/calculate.c
tom-wa: The epsilon is only there to prevent gcc from complaining about unsafe floating point comparison.
Gonna use a constant, but i don't think it should be configureable.

markus2330: There is a very good reason why floating point comparision should be done that way.
Yes, constant should be fine! You can export the epsilon using the contract below "constants" so (for documentation purposes).
Do not forget to document that floating point is used.

tom-wa:
@markus2330 done, i guess.

markus2330:
Can you increase the error numbers (please also consider other merge requests)

tom-wa:
@markus2330 done

PR #313, opened 2015-10-01
Python: subinterpreter support + some fixes

manuelm:
after a lot of crashes I finally got subinterpreter support working. In general one shouldn't mix pythons thread state APIs with the GIL API. Thankfully a lot of them is already deprecated (haven't used them).


manuelm:
jenkins build bindings please

src/plugins/python/python.cpp
```diff
@@ -47,45 +47,52 @@ static PyObject *Python_fromSWIG(ckdb::KeySet *keyset)
 	return SWIG_NewPointerObj(new kdb::KeySet(keyset), ti, 0);
 }
 
-extern "C"
+class Python_LockSwap
```
markus2330: nice solution

PR #315, opened 2015-10-02
bindings: in tests don't check KDB_VERSION if CHECK_VERSION is set

### LINKED ISSUE #272: Disable Version Test for python+lua tests
While generally it makes sense to check if system/elektra/version/constants/KDB_VERSION matches the compiled in version, there should also a way to ignore these tests, e.g. this is needed for API compatibility tests (running old tests against new library).

E.g. one occurrence is here: src/bindings/swig/python2/tests/testpy2_kdb.py

In the shell tests, such tests are disabled after setting the environment variable CHECK_VERSION.
Could you please implement the same for python and lua tests?

manuelm:
fixes #272

### DIFF
--- a/src/bindings/gi/lua/testgi_kdb.lua
+++ b/src/bindings/gi/lua/testgi_kdb.lua
@@ -28,8 +28,10 @@ do
 	local ks = kdb.KeySet(100)
 	db:get(ks, "system/elektra")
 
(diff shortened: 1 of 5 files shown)


manuelm:
rebased and ready

markus2330:
Thanks!

Issue #318, opened 2015-10-03
Segmentation Fault on `kdb lsmeta` when no argument provide


iandonnelly:

iandonnelly:
I think the issue was just not catching when no argument was provided. I updated this in 3fc4c930c0b8d87620c713c432bec072d3274d7a and updated the error handling. Look good to you @markus2330 ?

PR #319, opened 2015-10-03
Fix `kdb lsmeta` Segfault.

iandonnelly:
This commit should fix the error I reported in Issue #318

### DIFF
--- a/src/tools/kdb/metals.cpp
+++ b/src/tools/kdb/metals.cpp
@@ -13,11 +13,15 @@ MetaLsCommand::MetaLsCommand()
 
 int MetaLsCommand::execute (Cmdline const& cl)
 {
+
+	if(cl.arguments.size() != 1){
+		throw invalid_argument("1 argument required");
+	}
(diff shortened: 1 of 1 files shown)


src/tools/kdb/metals.cpp
```diff
@@ -13,11 +13,15 @@ MetaLsCommand::MetaLsCommand()
...
+
+	if(cl.arguments.size() != 1){
+		throw invalid_argument("1 argument required");
+	}
+
 	Key root (cl.arguments[0], KEY_END);
 	if (!root.isValid())
 	{
-		cerr << "Not a valid name supplied" << endl;
-		return 1;
```
iandonnelly: The documentation doesn't currently mention the return value of 1. Is this used anywhere? If not this new solution is better, right?

markus2330: You are right!

markus2330:
Thank you, very good fix!

PR #321, opened 2015-10-03
getenv example: do not link to elektra/elektratools

pinotree:
This small C application just dumps the content of the environment, printing also the environment variables requested as arguments.
Since this is usually run as tool in a `elektrify-getenv` context, it will get the effects of elektragetenv, and thus linking elektra and elektratools is not needed.


markus2330:
You are completely right, thank you!

PR #322, opened 2015-10-03
Stop installing scripts not useful outside of sources

pinotree:
Do not install anymore few scripts which are basically useless when installed, and indeed they are just useful when working on the elektra sources:
- configure-*: used to run cmake for building elektra
- copy-template: copy the "template" plugin to create a new plugin
- generate-news-entry: small helper to print few details for NEWS


markus2330:
They were installed because I thought that in the tools-folder they would not disturb. But that assumption seems to be wrong. Thanks for merge request.

pinotree:
> They were installed because I thought that in the tools-folder they would not disturb.

They don't "disturb" :) It's just that they are useful only among the sources, not really when shipped to the installation (none of the affected scripts have been installed in Debian packages, for example).

markus2330:
In the unofficial package they were ;)

You are right that they only operate within source, but nevertheless they save you from typing `../../../scripts/XXX` but instead you could write `kdb XXX`.

If no one misses them, I will fix the unofficial packages.

PR #326, opened 2015-10-05
bindings: rename python3 to python

manuelm:
finally rename back the python3 bindings


manuelm:
@markus2330 If the renaming is ok for you, please pull this in *before* #315. I'll rebase #315 afterwards

markus2330:
It might create problems in the future (think of potential incompatible python4) but, as said, if you would like to have it, I will accept it.

manuelm:
I did some investigation regarding upcoming major python versions and possible backwards incompatibilities. Didn't find much but everyone agrees there won't be another big break and the current one was kind of a mistake in the backsight. ref: http://www.curiousefficiency.org/posts/2014/08/python-4000.html

Additional apart from binary keys we don't use much (read: none) python3 features anyway.

Issue #327, opened 2015-10-07
mathcheck segfault on i386


markus2330:
@tom-wa 
There is a  segfault on i386, see:
http://community.markus-raab.org:8080/job/elektra-gcc-i386/lastFailedBuild/console

In http://community.markus-raab.org:8080/job/elektra-gcc-i386/ws/ you might find more information (e.g. coredump)

```
58/92 Test #58: testscr_check_kdb_internal_check ...***Failed    0.46 sec

ELEKTRA KDB INTERNAL TEST SUITE

Check if script tests the correct version
error: check of plugin mathcheck failed
error: check of plugin mathcheck produced: "Segmentation fault (core dumped)"
testscr_check_kdb_internal_check.sh RESULTS: 76 test(s) done 2 error(s).
```

manuelm:
```
$ gdb --args ./build/bin/kdb check mathcheck
[...]
(gdb) bt f
#0  0xf7fa9e5f in keySetString (key=0x812e2b8, newStringValue=0x88e368f1 <error: Cannot access memory at address 0x88e368f1>) at /home/jenkins/workspace/workspace/elektra-gcc-i386/src/libelektra/keyvalue.c:376
        ret = 0
#1  0xf7fb18f4 in keyVInit (key=0x812e2b8, name=0xf7f17b04 "system/elektra/modules/mathcheck/export/constants/EPSILON", va=0xffffd3cc "\265\370\344>") at /home/jenkins/workspace/workspace/elektra-gcc-i386/src/libelektra/keyhelpers.c:344
        name_options = <optimized out>
        action = <optimized out>
        value_size = 0
        value = 0x88e368f1
        func = 0x0
        flags = 0
        owner = 0x0
#2  0xf7fb1a26 in keyVNew (name=0xf7f17b04 "system/elektra/modules/mathcheck/export/constants/EPSILON", va=0xffffd3c4 "\002") at /home/jenkins/workspace/workspace/elektra-gcc-i386/src/libelektra/key.c:224
        key = 0x812e2b8
#3  0xf7fb1a73 in keyNew (name=0xf7f17b04 "system/elektra/modules/mathcheck/export/constants/EPSILON") at /home/jenkins/workspace/workspace/elektra-gcc-i386/src/libelektra/key.c:206
        k = <optimized out>
        va = 0xffffd3c4 "\002"
[...]

PR #330, opened 2015-10-08
fixed mathcheck export value

tom-wa:


src/plugins/mathcheck/mathcheck.c
```diff
@@ -21,8 +21,8 @@
 #include "mathcheck.h"
 
 #define MIN_VALID_STACK 3
-#define EPSILON 0.00001
-
+#define EPSILON 0.00001f
```
markus2330: Why not simply use `"0.00001"` (or put quotes using two macros)?

markus2330:
Thanks for quick response!

- Please also add a testcase for a wrong and correct calculation.
- Please also add fix #327 in the comment of the commit.

src/plugins/mathcheck/mathcheck.c
tom-wa: had a lot of trouble decidin

markus2330: https://gcc.gnu.org/onlinedocs/cpp/Stringification.html

markus2330:
I meant the git commit of git commit, but never mind ;)

Thanks for the fix!

PR #332, opened 2015-10-17
set fix number of column count for Csvstorage

tom-wa:
added an option to set a fixed number of columns, everything that has more or less columns gets rejected.


markus2330:
Thanks!

PR #336, opened 2015-10-26
Ini bugfix

tom-wa:


markus2330:
Please write (at least short) description. Make sure the text `fix #nr` is in the respective commits.

@fberlakovich: Can you also take a look at this.

src/plugins/ini/ini.c
```diff
@@ -226,15 +228,15 @@ int elektraIniGet(Plugin *handle, KeySet *returned, Key *parentKey)
 	{
 		switch (ret)
 		{
-		case -1:
```
markus2330: please do not indent.

src/plugins/ini/ini.c
```diff
@@ -84,6 +84,8 @@ static int iniKeyToElektraKey (void *vhandle, const char *section, const char *n
 
 	keyAddBaseName (appendKey, name);
 
+	if(value == NULL)
+		keySetMeta(appendKey, "ini/empty", "");
```
markus2330: Can a null key be used unambiguously? If yes, it should be preferred. There are already 3 states of emptyness (key not present, value not present, string emtpy), we should not over-complicate everything.

src/plugins/ini/inih-r29/inih.c
```diff
@@ -147,8 +147,11 @@ int ini_parse_file(FILE* file,const struct IniConfig* config, void* user)
                     error = lineno;
             }
             else if (!error) {
-                /* No '=' or ':' found on name[=:]value line */
-                error = lineno;
+        	end = find_char_or_comment(start, '\0');
```
markus2330: has spaces

src/plugins/ini/testmod_ini.c
```diff
@@ -103,7 +103,49 @@ static void test_plainIniWrite(char *fileName)
...
+			keyNew ("user/tests/ini-write/section2",
+					KEY_BINARY,
+					KEY_END),
+			keyNew ("user/tests/ini-write/section2/key3",
+					KEY_VALUE, "value3",
+					KEY_END),
+			keyNew ("user/tests/ini-write/section2/emptykey", KEY_META, "ini/empty", "", KEY_END),
+			KS_END);
+
+	succeed_if(plugin->kdbSet (plugin, ks, parentKey) >= 1,
```
markus2330: could you test full round-trip? also get it again and also check if keyset is equal.

fberlakovich:
Looks nice. I am very happy that this was finally fixed and even with a much simpler solution than I could think of.

src/plugins/ini/ini.c
tom-wa: i couldn't really differentiate between 
`key = ` and `key` without the metakey when writing back to the config file

markus2330: tom-wa wrote:
> i couldn't really differentiate between 
> `key = ` and `key` without the metakey when writing back to the config file

Ok, I see. So is the following correct?

- null keys are section (status quo)
- `key` is an empty string with ini/empty set
- `key = ` is an empty string

Please correct my statements and add them to README.md.
You also need to rebase, I modified the README.md.

tom-wa: Yes, you are correct. 
Will rebase and update.

PR #339, opened 2015-11-01
qt-gui: fix search results bug

### LINKED ISSUE #328: qt-gui: large list of search results becomes empty when scrolling
When a search term has many hits, the listview will become empty when the user scrolls fast through the search results. This does not happen when the user scrolls slowly. Probably a bug in Qt. Will try to find a solution.

0003088:
Fix #328


markus2330:
Thank you!

PR #340, opened 2015-11-02
Markdown link converter for Doxygen

KurtMi:
Markdown link converter for doxygen and some changes in documentation.
- [x] done


doc/CMakeLists.txt
```diff
@@ -57,6 +57,8 @@ if (DOXYGEN_FOUND)
 
 	find_package(Perl)
 
+	set (MARKDOWN_LINK_CONVERTER ${CMAKE_BINARY_DIR}/bin/markdownlinkconverter)
```
markus2330: See cmake/Modules/LibAddMacros.cmake (e.g. exporterrors), you should use find_util and DEPENDS. This also works in the CMAKE_CROSSCOMPILING case.

doc/CODING.md
```diff
@@ -96,7 +96,12 @@ with `ks`, `key` or `kdb`.
 
 ### Doxygen Guidelines ###
 
-Use `doxygen` to document APIs, if available.
+`doxygen` is used to document the API, if available. We support also the import
```
markus2330: "if available" is not from you but its wrong. Either fix to "is used to build" or remove "is available"

witch -> which

avoid camelcase for link (MarkdownLinkConverter, folder is correctly named markdownlinkconverter)

doc/CODING.md
```diff
@@ -104,6 +109,7 @@ Use `\copydoc`, `\copybrief` and `\copydetails` intensively.
 
 Files should start with:
 
+\verbatim
```
markus2330: This is not understood by github.

Why is code not verbatim by default?

doc/Doxyfile
```diff
@@ -768,7 +768,10 @@ INPUT                  = @PROJECT_SOURCE_DIR@/src/libelektra \
                          @PROJECT_SOURCE_DIR@/src/libtools \
                          @PROJECT_SOURCE_DIR@/src/bindings/cpp/include \
                          @PROJECT_SOURCE_DIR@/src/plugins/doc/doc.h \
-                         @PROJECT_SOURCE_DIR@/src/libloader/doc.c
+                         @PROJECT_SOURCE_DIR@/src/libloader/doc.c \
+                         @PROJECT_SOURCE_DIR@/doc \
+                         @PROJECT_SOURCE_DIR@/src/plugins \
```
markus2330: won't this pull in a lot of documentation from internal plugin functions?

doc/markdownlinkconverter/README.md
```diff
@@ -0,0 +1,27 @@
...
+  to be compatible with github, where you can show the content of a folder in
+  combination with the README.md of the containing folder. Links ending with
+  `/` will be changed to `/README.md`.
+* Anchors wont work in imported markdown pages.
+
+## About
+
+The Markdown Link Converter (MLC) is used as input filter in doxygen, if a markdown
+file is desired to be in the API documentation it only must have the extension
+`md` and be in the `INPUT` path.
```
markus2330: `.md`

doc/markdownlinkconverter/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+The MLC gives each markdown file a header `{ #header }` witch is attached to a title
```
markus2330: s/witch/which/g

doc/markdownlinkconverter/README.md
```diff
@@ -0,0 +1,27 @@
...
+and converts the links of other markdown files to refer to this header. This conversions
```
markus2330: This conversion

doc/markdownlinkconverter/README.md
```diff
@@ -0,0 +1,27 @@
...
+happens in 2 passes, witch is needed because there can be files with no title.
```
markus2330: sed -i s/witch/which/g */*

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+#define FOLDER_DELIMITER '\\'
+#else
+#define FOLDER_DELIMITER '/'
+#endif
+
+/* Slashes or backslashes to the elektra root.
+ * Depends on the place of the executable, example:
+ * /foo/bar/elektra/build/bin/executable
+ * E:\\foo\bar\elektra\build\bin\executable
+ * In both cases the there are 3 slashes or backslashes
```
markus2330: I thinks its better to pass the source directory as arguments and not expect the build directory to be a subdirectory of the source.

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+ */
+#define FOLDER_DELIMITER_TO_ELEKTRA_ROOT 3
+#define TEMP_FILENAME "temp"
+
+// Link Blacklist: do not convert links with the following starting and ending
+char * ignoreTargetEnd [] = { ".h", ".c" , ".cpp" , ".hpp" , "\0"};
+// ignore http, @ref and #anchor
+char * ignoreTargetStart [] = { "#", "@" , "http" , "\0"};
+// both need to be null terminated
```
markus2330: Do you mean null termination in addition to the element `"\0"` that contains two null bytes?

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+
+// helpers
+void printTarget(FILE * output, char * target, char * filenameInElektra);
+void printConvertedPath (FILE * output ,char * path);
+char * getPathInElektraRoot (char * executablename, char * filename);
+void exitError (FILE * f1, FILE * f2, char * mes);
```
markus2330: Every argument can be modified? (Please use const)

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+ */
+
+#define LINK_START 0
+#define LINK_POSSIBLE(old, new) ((old == 0 || old == 5) && new == 2)
+#define LINK_FOUND(old, new) (state == 4 && newstate == 5)
+#define LINK_NAME(old, new) (((old == 2 || old == 3) && (new == 2  || new == 3)) \
+								|| (old == 3 && new == 4))
+#define LINK_TARGET(old, new) (state == 4 && newstate == 4)
+#define LINK_TRAP(old, new) (new == 6)
+#define LINK_NOLINK(old, new) (newstate == 0 || newstate == 1)
```
markus2330: I think you can should use inline functions here.
These macros have a number of problems. Think of `LINK_POSSIBLE(a++, b)` or even worse: `LINK_POSSIBLE(a&b, c)`.

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+/* The next transition array matches the first title,
+ * with the following regex:
+ *
+ * ^' '*#+.*$
+ *
+ * TITLE_START defines the start state and TITLE_ISGOAL ()
+ * checks whether a title has matched.
+ */
+
+#define TITLE_START 0
```
markus2330: again: should be const int, no macro

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+		{
+			// first [, possible link
+			// position is saved for setting back
+			if (fgetpos (input, &pos))
+				exitError (input, NULL, "fgetpos");
+
+			index = 0;
+			len = 0;
+		}
+		if (LINK_FOUND (state, newstate))
```
markus2330: why check LINK_FOUND if a link is not possible?

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+			char target [len * sizeof (char) + 1];
+			if (fread (&target[0], sizeof (char), len, input) != len)
+				exitError (input, NULL, "fread");
+
+			target[len] = '\0';
+
+			//check target
+			bool targetOK = true;
+			//start
+			for (int i = 0;strcmp (ignoreTargetStart[i], "\0") != 0;++i)
```
markus2330: Isn't this the same: `strcmp (ignoreTargetStart[i], "")`? (strcmp should stop at first null-character)

The function gets really long.

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+			}
+			//end
+			for (int i = 0;strcmp (ignoreTargetEnd[i], "\0") != 0;++i)
+			{
+				if(len < strlen (ignoreTargetEnd[i]))
+					continue;
+
+				int j = len - strlen (ignoreTargetEnd[i]);
+				if (strncmp (ignoreTargetEnd[i], &target[j],
+					strlen (ignoreTargetEnd[i])) == 0)
```
markus2330: Please use variables for multiple usage of this strlen, strcmp expressions.

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+		}
+		state = newstate;
+	}
+}
+
+int main (int argc, char *argv[])
+{
+	if (argc < 2)
+	{
+		fprintf (stderr, "args error\n");
```
markus2330: Could be a bit more verbose ;)

doc/CODING.md
KurtMi: doxygen parses markdown files first with the doxygen parser and afterwards with the markdown parser, so it recognizes the @commands as doxygen commands. This behaviour produces an error and therefore is no output between /** and */ in the doxygen html. I did not find a nice solution for this problem :/

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+	strcpy (executablename, argv[0]);
+	strcpy (filename, argv[1]);
+
+	char * filenameInElektra = getPathInElektraRoot (executablename, filename);
+
+	//1st pass
+	FILE * input = fopen (filename, "r");
+	if (!input)
+	{
+		fprintf (stderr ,"fopen Error\n");
```
markus2330: at least tell the user which file it tried to open.

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+			exitError (input, NULL, "parsed file path invalid");
+
+		// print Title + Header
+		fprintf (output, "# %s # {#", &title[1]);
+		printConvertedPath (output, filenameInElektra);
+		fprintf (output, "}\n");
+		//2nd pass
+		convertLinks (input, output, filenameInElektra);
+		fclose (input);
+	} else
```
markus2330: please `} else {`.

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+	{
+		fclose (input);
+		//reset temp file
+		if (fsetpos (output, &startTempFile))
+			exitError (output, NULL, "fsetpos");
+
+		//2nd pass
```
markus2330: why a second pass?

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+	{
+		++target; // remove starting /
+	}
+	printConvertedPath (output, target);
+}
+
+/* This helper converts and prints a given path, all slashes or backslashes
+ * and points will be replaced by underscore.
+ */
+void printConvertedPath (FILE * output, char * path)
```
markus2330: method could be static, or?

doc/markdownlinkconverter/markdownlinkconverter.c
```diff
@@ -0,0 +1,452 @@
...
+ * the the parsed file path in the Elektra root folder.
+ * Example:
+ * /foo/bar/test/elektra_src/build/bin/exe
+ * /foo/bar/test/elektra_src/doc/help.md
+ * with the FOLDER_DELIMITER_TO_ELEKTRA_ROOT set to 3 will return:
+ * doc/help.md
+ */
+char * getPathInElektraRoot (char * executablename, char * filename)
+{
+	int countDeliToFsRoot = -(FOLDER_DELIMITER_TO_ELEKTRA_ROOT-1);
```
markus2330: Deli -> Delim

doc/CODING.md
markus2330: Lets give the example to a separate file not parsed or lets use a newline between code block and doxygencommands so that github can at least make a code block.

doc/markdownlinkconverter/markdownlinkconverter.c
KurtMi: There can be markdown files with no headings and only at the end of a file you can be sure that there is no heading at all in a file. Adding the convention that every .md file must have at least one heading seemed harsh.

markus2330: The questions should also be answered in the code. The idea is that you add a code-comment to contain the information you just stated in git hub.

KurtMi: This details are described in the README.md of the markdown link converter.

markus2330: So `//2nd pass (see README.md)` would suffice

doc/markdownlinkconverter/markdownlinkconverter.c
KurtMi: I think there is no possibility (without a hack) to pass an extra argument to the input filter, it will be called by popen with <filter> <input-file> (https://github.com/doxygen/doxygen/blob/389dc4c06c31a5da26a55bf2f6a8ba8316dbc1ff/src/definition.cpp#L747). But there is the possibility to write the project directory path in a file with cmake, read it with the input filter and delete it after the doxygen call.

markus2330: Or Elektra_SOURCE_DIR in CMakeCache.txt already contains the value?

KurtMi: nice!

markus2330:
I changed merge request build that it also builds docu:
http://doc.libelektra.org/api/mergerequest/

markus2330:
jenkins build please

markus2330:
jenkins build please

markus2330:
Build fails when building latex: `Package inputenc Error: Unicode char \u8:ā not set up for use with LaTeX.`

Can you fix that wrong umlaut?

KurtMi:
I will have a look at it.

KurtMi:
jenkins build please

KurtMi:
I think on the  debian-stretch-mr slave is latex not installed, because:
Latex not found, PDF Manual can't be created.

doc/markdownlinkconverter/markdownlinkconverter.c
KurtMi: yes that is right, even more functions can be defined as static, but I think there will be no files which use the markdownlinkconverter

doc/CMakeLists.txt
```diff
@@ -57,6 +57,9 @@ if (DOXYGEN_FOUND)
 
 	find_package(Perl)
 
+	add_executable (markdownlinkconverter markdownlinkconverter/markdownlinkconverter.c)
+	find_util (markdownlinkconverter MARKDOWN_LINK_CONVERTER "")
+
```
KurtMi: I am not sure if the find_util marco is used right here

doc/markdownlinkconverter/markdownlinkconverter.c
markus2330: > even more functions can be defined as static
The idea is to apply it to all functions

> will be no files which use the markdownlinkconverter
`static` also has the advantage that the compiler can warn about not-used functions. And there is no disadvantage in doing so.

For tests we also mark everything as static to see to be sure all tests are actually executed.

markus2330:
I installed latex-recommended and latex-extra on testing agent now.

markus2330:
jenkins build please

doc/CMakeLists.txt
markus2330: There are two output parameters, you should request both and concatinate them to MARKDOWN_LINK_CONVERTER.

I added some docu for find_util: 9ee141a06df3ab3e8e0933f22c72c498f5483557

Can you give arguments to the doxygen filter? If not, it won't work with mingw.

markus2330:
The PDF has 512 pages. Many are, however, completely empty chapters only titled "README". Isn't markdown to latex implemented in doxygen?

Btw. I pinned down the build job that all merge requests will be built on debian-stable-mm. (even though testing has latex now, too)

doc/tutorials/plugins.md
```diff
@@ -11,45 +11,45 @@ and what various methods exists within one.
 
 ### The Interface ###
 
-All plug-ins use the same basic interface. This interface consists of five basic functions,
```
markus2330: The fixes are ok, but its not good to have them all in one merge request.

doc/NEWS.md
```diff
@@ -247,7 +247,7 @@ Thanks to Manuel Mausz for to testing and improving portability!
 - make dbus example work again
 - fix compiler warnings for gcc and clang
 - fix valgrind suppressions
-- Installation of GI binding is fixed, thanks to Dāvis
+- Installation of GI binding is fixed, thanks to Davis
```
markus2330: Is it somehow possible to allow UTF8 by default?
Basically only a latex preamble would be needed, woudn't it?

src/plugins/path/README.md
```diff
@@ -13,18 +13,18 @@ This plugin checks whether the value of a key is a valid filesystem path.
...
 The motivation to write this plugin is given by the two paths that exist
-in /etc/fstab: the device ﬁle and the mountpoint. A missing ﬁle is
-not necessarily an error, because the device ﬁle may appear later when
+in /etc/fstab: the device file and the mountpoint. A missing file is
+not necessarily an error, because the device file may appear later when
 a device is plugged in and the mountpoint may be there when another
 subsequent mount was executed. So only warnings are yielded in that
 case. One situation, however, presents an error: Only an absolute path
 is allowed to occur for both device and mount point. When checking for
-relative ﬁles, it is not enough to look at the ﬁrst character if it is
```
markus2330: urgh, thanks for fixing all this. Seems like someone did a copy and paste from a pdf...

doc/CMakeLists.txt
KurtMi: > Can you give arguments to the doxygen filter? If not, it won't work with mingw.
No, there is no possibility to pass arguments to a input filter in doxygen, it will be executed by doxygen with popen. I think there is no need to take care of the arguments because doxygen does it.

> I added some docu for find_util: 9ee141a
"EXE_SYM_LOC ... Note: you must pass this value to COMMAND of add_custom_command"
Why should cmake execute the markdownlinkconverter?

doc/tutorials/plugins.md
KurtMi: is there a possibility to remove a commit form the pull request? So I can make an extra pull request with this fixes?

markus2330: I think you can do `rebase -i` and just drop the commit. (Warning: its a potentially dangerous operation: you might lose data)

If you cannot do it: I can still accept the PR, just split the PR (in such situations) in the future.

doc/CMakeLists.txt
markus2330: > No, there is no possibility to pass arguments to a input filter in doxygen,

Ok, then it won't work with mingw. So there is no real reason to use `find_util`. Sorry for the unnecessary extra work.

> Why should cmake execute the markdownlinkconverter?

Its the description of the typical usage of `find_util` and does not apply to this situation.

KurtMi: > Ok, then it won't work with mingw

So if I would use the mingw gcc to generate the docu it will fail?

So I can leave the CmakeList.txt with the find _util or should I change it back?

KurtMi:
> The PDF has 512 pages. Many are, however, completely empty chapters only titled "README". Isn't markdown to latex implemented in doxygen?

512 is not so optimal but I planed to optimze the pdf output after this pull request. I made a note for this task in the README.md of the markdownlinkconverter.

doc/CMakeLists.txt
markus2330: > So if I would use the mingw gcc to generate the docu it will fail?

Yes, it would call `wine` without your program as argument.

It is possible to workaround this argument limitation. We could generate a shell script which calls `wine markdownlinkconverter`. So better leave the usage of `find_util`: then we have a reminder that this place needs to be modified when we adapt `find_util` to work without arguments.

markus2330:
Yes, this should be done in another PR. Ping me when you are finished with this one.

KurtMi:
all done, build will fail, because utf8 stuff

markus2330:
What do you mean by "build will fail"? Even the pdf was built without errors.

markus2330:
Thanks for all the effort!

PR #343, opened 2015-11-06
fixed invalid free in iconv/validateFile

tom-wa:
made a stupid mistake in my code and forgot that kdbbUTF8Engine frees the buffer.

### DIFF
--- a/src/plugins/iconv/iconv.c
+++ b/src/plugins/iconv/iconv.c
@@ -103,7 +103,7 @@ setlocale (LC_ALL, "");
  *
  * Binary values are not effected.
  *
- * If iconv() or nl_langinfo() is not available on your system, or if iconv()
+ * If iconv() or [M`?[M`?nl_langinfo() is not available on your system, or if iconv()
  * this plugin can't be used.
  *
  * @param direction must be @c UTF8_TO (convert from current non-UTF-8 to
@@ -114,7 +114,7 @@ setlocale (LC_ALL, "");
  * 	leading NULL; after the call: the size of the converted string including
  * 	leading NULL
  * @retval 0 on success
- * @retval byte position of the wrong encoded character on failure
+ * @retval -1 on failure 
  * @ingroup backendhelper
  *
  */
@@ -137,19 +137,8 @@ int kdbbUTF8Engine(Plugin *handle, int direction, char **string, size_t *inputOu
 
 	if (converter == (iconv_t)(-1)) return -1;
 
-	int Multiplier;
-
-	if(kdbbNeedsUTF8Conversion(handle))
-	{
-		Multiplier = 4;
-	}
-	else
-	{
-		Multiplier = 1;
-	}
-
 	/* work with worst case, when all chars are wide */
-	bufferSize=*inputOutputByteSize * Multiplier;
+	bufferSize=*inputOutputByteSize * 4;
 	converted=malloc(bufferSize);
 	if (!converted) return -1;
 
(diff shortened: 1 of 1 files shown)


src/plugins/iconv/iconv.c
```diff
@@ -186,11 +186,13 @@ static int validateFile(Plugin *handle, Key *parentKey)
...
 	const char *fileName = keyString(parentKey);
 	FILE *fp = fopen(fileName, "rb");
-	char inputBuffer[4096];
+	char *inputBuffer = NULL;
+	const int inputBufferSize = 4096;
 	unsigned long counter = 0;
 	while(!feof(fp))
 	{
-		size_t bytesRead = fread(inputBuffer, CHAR_SIZE_MAX, (sizeof(inputBuffer)/CHAR_SIZE_MAX), fp);
+		inputBuffer = realloc(inputBuffer, inputBufferSize);
```
markus2330: can you use elektraRealloc here? Why is it reallocated every time? inputBufferSize is not changed?

markus2330:
Mmmh, you do not need to free arrays on the stack. Arrays on the stack should be preferred whenever possible.

Why do you think you need realloc and free?

Can you describe the problem you tried to fix here?

tom-wa:
> Mmmh, you do not need to free arrays on the stack. Arrays on the stack
> should be preferred whenever possible.
>
> Why do you think you need realloc and free?
>
> Can you describe the problem you tried to fix here?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ElektraInitiative/libelektra/pull/343#issuecomment-154460864>.
>
I've been using an array on the stack in my initial version.
We later decided to use kdbbUTF8Engine for the file validation instead
of iconv directly but i overlooked the part where the buffer gets free'd
at the end of kdbbUTF8Engine.
A lot of overhead is caused that way though.

markus2330:
Yes, please rethink for a better way. Ideally on the stack, and if this is not possible with a single elektraMalloc.

tom-wa:
> Yes, please rethink for a better way. Ideally on the stack, and if
> this is not possible with a single elektraMalloc.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ElektraInitiative/libelektra/pull/343#issuecomment-154466225>.
>
currently get and set are allocating and freeing memory in a loop too.
actually for every kdbbutf8engine call(so every key) there are 3 frees,
3 mallocs, 1 iconv_open() and 1 iconv_close().
i'd have to rewrite a lot of the code, but i could reduce the overhead
to a fraction.

markus2330:
Please refrain from rewriting everything, but only find a good solution for this particular issue.

src/plugins/iconv/iconv.c
```diff
@@ -103,7 +103,7 @@ setlocale (LC_ALL, "");
  *
  * Binary values are not effected.
  *
- * If iconv() or nl_langinfo() is not available on your system, or if iconv()
+ * If iconv() or [M`?[M`?nl_langinfo() is not available on your system, or if iconv()
```
markus2330: ?

src/plugins/iconv/iconv.c
```diff
@@ -137,19 +137,8 @@ int kdbbUTF8Engine(Plugin *handle, int direction, char **string, size_t *inputOu
...
-		Multiplier = 4;
-	}
-	else
-	{
-		Multiplier = 1;
-	}
-
 	/* work with worst case, when all chars are wide */
-	bufferSize=*inputOutputByteSize * Multiplier;
+	bufferSize=*inputOutputByteSize * 4;
```
markus2330: avoid magic constant

PR #344, opened 2015-11-06
Fix annoying error messages when deleting key (Qt5.5)

0003088:
I do not know why the error appeared only in Qt5.5 and not in Qt5.4, the index after deleting a key was computed wrong, now it should never go below 0.


markus2330:
Great you found the problem! Can you please reformat the spaces to tabs?

markus2330:
Thanks for the fix! Btw. Your editor is still configured to use spaces. I fixed it for this PR.

Issue #295, opened 2015-09-17
3-way merge: marks every key as sync


markus2330:
just a reminder, because TODO in src/tools/qt-gui/src/treeviewmodel.cpp is not very visible

markus2330:
solution might not actually effect 3-way mering (reconstruct minimal changed keyset as additional API)

fberlakovich:
Could you please give an example of what keys are marked as sync? The TODO in treeviewmodel does not seem to be present anymore.

markus2330:
I think at least all of "our". Its best if you create a testcase with keys not marked sync. After merging only merged keys should be marked sync.

I think the problem is in rebaseKey, which always creates a new key. And it seems to be called for every key in "our"? (see ThreeWayMerge::detectConflicts)

markus2330:
(or simply unsync and check for sync in the already existing test cases)

fberlakovich:
Which API should I use to check whether a key needs sync (especially in C++)?

markus2330:
I added Key::needSync. For changing the bit you need an internal API (kdbprivate.h), `clear_bit(k->flags, KEY_FLAG_SYNC);` should do for test cases.

fberlakovich:
Am I right that this is mainly a problem when either ours or theirs overlaps with the merge target? Otherwise the sync flag is justified because the merged keys have a new parentKey and strictly spoken each merged key is a new one.

markus2330:
I am not sure, its the best if you debug it with the GUI.
Even if it can be argued its a new key -- when theirs and ours is identical they should not have the sync flag set.

Rationale: sync flag is only an optimization to detect if a keyset needs to be rewritten. If we know it needs not, we should not set the flag.

The current status, that everything is rewritten in the merge case even though neither theirs nor ours changed, is unsatisfactory.

markus2330:
Thanks for fixing the issue for this release

PR #346, opened 2015-11-07
merger sync key problem

fberlakovich:
The commit should fix the issue discussed in #295. However, I could not test the fix with the gui. The gui eats up all my system ressources until the system grinds to a halt. When it finally starts it is so slow that I can hardly click anything. Please let me know if the issue is resolved.

### DIFF
--- a/src/libtools/include/merging/mergeresult.hpp
+++ b/src/libtools/include/merging/mergeresult.hpp
@@ -48,7 +48,10 @@ public:
 
 	void addMergeKey(const Key& key)
 	{
-		mergedKeys.append (key);
+		if (!mergedKeys.lookup(key))
+		{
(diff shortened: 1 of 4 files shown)


markus2330:
Is the problem with the GUI also without any mountpoints?

markus2330:
thanks for the fix! I opened #347 which might be caused due to merging.

fberlakovich:
I will try the GUI without any mountpoints. I only had two very small files mounted at the time and a `kdb ls` was very fast. I already had it running successfully previously, however, it never was very responsive on my development machine.

markus2330:
Problem is discussed in #348, fix will be done soon

markus2330:
Is fixed. It shouldn't matter if you have mountpoints.

PR #353, opened 2015-11-10
Improved the augeas plugin error messages

fberlakovich:
Partly fixes the issue discussed in #206. As you have guessed, the error messages stem from Augeas. However, the augeas plugin now includes all the information available from Augeas. The message reported in #206 now looks like this:
```
The command ls failed while accessing the key database with the info:
Error (#85) occurred!
Description: an Augeas error occurred
Ingroup: plugin
Module: storage
At: /path/to/augeas.c:425
Reason: parse_failed
	position: 58:0
	message: Iterated lens matched less than it should
	lens: /raw/tree
Mountpoint: system/limits
Configfile: /etc/security/limits.conf

### DIFF
--- a/src/plugins/augeas/augeas.c
+++ b/src/plugins/augeas/augeas.c
@@ -66,22 +66,44 @@ static const char *getLensPath(Plugin *handle)
 
 static const char *getAugeasError(augeas* augeasHandle)
 {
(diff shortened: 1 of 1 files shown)


markus2330:
Thank you, nicely done!

Issue #365, opened 2015-11-12
mathcheck: failure in unit tests


markus2330:
@tom-wa The double comparision seems to be broken, did you always use the epsilon?
```
src/plugins/mathcheck/testmod_mathcheck.c:60: error in main: error
src/plugins/mathcheck/testmod_mathcheck.c:63: error in main: error
```

tom-wa:
Sorry, that's a locale problem ( , vs . ) in testmod_mathcheck

PR #367, opened 2015-11-13
fix #363 #365

### LINKED ISSUE #363: mathcheck: avoid unnecessary decimal points
mathcheck seems to print `.000000` even for "integers".

	kdb setmeta user/example/mathcheck/k check/math ":= + a '5'"
	kdb set user/example/mathcheck/a 8
	kdb get user/example/mathcheck/k

With the result that 5+8 is `13.000000` and not 13 as one could expect (not knowing of internals)

In the long run I think we should make compilation variants out of mathcheck, so that the user can also choose to use int and maybe even fixed point decimals instead of double.

But for now, maybe we can suppress `0` at the end?

tom-wa:


src/plugins/mathcheck/testmod_mathcheck.c
```diff
@@ -42,7 +42,11 @@ int main(int argc, char** argv)
 {
 	printf ("MATHCHECK     TESTS\n");
 	printf ("==================\n\n");
-
+	char *oldlocale;
+	char *savedlocale;
+	oldlocale = setlocale(LC_NUMERIC, NULL);
+	savedlocale = strdup(oldlocale);
+	setlocale(LC_NUMERIC, "de_DE");
```
markus2330: thats not a good solution, should be locale-independent.

src/plugins/mathcheck/testmod_mathcheck.c
```diff
@@ -59,13 +63,13 @@ int main(int argc, char** argv)
 	ks = create_ks("1", "== / bla/val1 bla/val3");
 	test(ks, (-1));
 	ksDel(ks);
-	ks = create_ks("3", "== + '1.5' '1.5'");
```
markus2330: I think `.` would be preferred in general.

src/plugins/mathcheck/testmod_mathcheck.c
```diff
@@ -68,12 +69,19 @@ int main(int argc, char** argv)
 	ks = create_ks("1", "== + '1.5' '1.5'");
 	test(ks, (-1));
 	ksDel(ks);
-	ks = create_ks("10", "== + bla/val3 '7'");
+	
+    ks = create_ks("10", "== + bla/val3 '7'");
```
markus2330: spaces

src/plugins/mathcheck/testmod_mathcheck.c
```diff
@@ -68,12 +69,19 @@ int main(int argc, char** argv)
...
 	test(ks, 1);
 	ksDel(ks);
 	printf ("\ntestmod_mathcheck RESULTS: %d test(s) done. %d error(s).\n",
 			nbTest, nbError);
 
+	char buffer[24];
+	elektraFtoA(buffer, sizeof(buffer), (1.5));
+	succeed_if(!(strcmp(buffer, "1.5")), "elektraFtoA failed");
+    fprintf(stderr, "elektraFtoA: val: %g, ret: %s\n", (1.5), buffer);
```
markus2330: spaces again

PR #379, opened 2015-11-19
INI draft: relative ordered ini with metadata based tree-style structure

tom-wa:
needs "some" refactoring...


markus2330:
I will look at it after the release! But its clear that its very ambitious!

Issue #391, opened 2015-11-21
crypto: compiler warning


markus2330:
```
/home/markus/Projekte/Elektra/current/src/plugins/crypto/testmod_crypto.c:99:25: warning: 
      expression which evaluates to zero treated as a null pointer constant of type 'const char *'
      [-Wnon-literal-null-conversion]
        Key *errorKey = keyNew(KEY_END);
                               ^~~~~~~
/home/markus/Projekte/Elektra/current/src/plugins/crypto/testmod_crypto.c:108:25: warning: 
      expression which evaluates to zero treated as a null pointer constant of type 'const char *'
      [-Wnon-literal-null-conversion]
        Key *errorKey = keyNew(KEY_END);
                               ^~~~~~~
/home/markus/Projekte/Elektra/current/src/plugins/crypto/testmod_crypto.c:141:25: warning: 
      expression which evaluates to zero treated as a null pointer constant of type 'const char *'
      [-Wnon-literal-null-conversion]
        Key *errorKey = keyNew(KEY_END);
                               ^~~~~~~
/home/markus/Projekte/Elektra/current/src/plugins/crypto/testmod_crypto.c:176:25: warning: 
      expression which evaluates to zero treated as a null pointer constant of type 'const char *'
      [-Wnon-literal-null-conversion]
        Key *errorKey = keyNew(KEY_END);
                               ^~~~~~~
/home/markus/Projekte/Elektra/current/src/plugins/crypto/testmod_crypto.c:216:25: warning: 
      expression which evaluates to zero treated as a null pointer constant of type 'const char *'
      [-Wnon-literal-null-conversion]
        Key *errorKey = keyNew(KEY_END);
                               ^~~~~~~
6 warnings generated.
```

markus2330:
Should be `keyNew (0, KEY_END)`

petermax2:
interesting, I was not able to reproduce these warnings in my build containers. but I see the problem. PR is on its way.

markus2330:
Should have mentioned: only clang-modernize gave this warning.

petermax2:
Thx for the clarification!

PR #393, opened 2015-11-21
crypto-plugin: fix compiler warning (issue #391)

petermax2:
fixed the warnings reported in #391

### DIFF
--- a/src/plugins/crypto/testmod_crypto.c
+++ b/src/plugins/crypto/testmod_crypto.c
@@ -96,7 +96,7 @@ static void getIncompleteConfiguration(KeySet **ks)
 
 static void test_init()
 {
-	Key *errorKey = keyNew(KEY_END);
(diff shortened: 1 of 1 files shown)


src/plugins/crypto/testmod_crypto.c
```diff
@@ -213,7 +213,7 @@ static void test_enc_and_dec_with_null()
 {
 	elektraCryptoHandle *handle;
 	KeySet *config;
-	Key *errorKey = keyNew(KEY_END);
+	Key *errorKey = keyNew ("", KEY_END);
```
petermax2: speaking of code formatting. do we have a decision for blanks and brackets, etc.?

markus2330: Yes, the space is ok! Actually, there also should be a space around the `*`. Kurt is writting a style checker to formalize it. But it won't be too strict.

Issue #394, opened 2015-11-22
changes in bindings for library split


markus2330:
I split the libraries as explained in src/libs/README.md. 

Can you please review the merge 2803083b15c39b9ec47ec46a7099b14335119e21, especially src/libs/elektra/CMakeLists.txt, if the library (symbol) exports are still correct? And can you check if the bindings are linked against the correct libelektra-*?

Then I added kdbGetNamespace to the API. Maybe its useful to add to the bindings, too.

manuelm:
Restricting exported symbols is broken. e.g. ini plugin is exporting `writeComments` which shouldn't be allowed.

bindings link against `-plugin` for no reason. Also `-ease`, `-meta` and `-proposal` can be removed.

markus2330:
Thank you for looking into it!

@tom-wa can you make writeComments static? Please also fix the compilation warning in the test cases.

@manuelm Can you fix the linkage of the bindings? How will you provide the elektraPluginGetConfig for the lua/python plugins?

I think there is a lot more optimization potential. Especially there is still junk in the core. Everyone is welcome to report how the libraries can be improved.

manuelm:
> @manuelm Can you fix the linkage of the bindings?
I can take a look but I wish you'd use merge requests instead of just pushing like the rest of us.

> How will you provide the elektraPluginGetConfig for the lua/python plugins?
?! I'm talking about the bindings. Not plugins. Imho they're linking against all libs though only core is required.

markus2330:
Didn't think anyone was interesting in reviewing the pull request. I thought its no issue, because we have a full release time for all the details of the library split. Nothing is in stone yet, we can even rename the libs and move functions around.

> Not plugins.

To clarify: When the plugin interface does not provide the plugin configuration you would need in the binding a way to issue `elektraPluginGetConfig` and thus also the bindings would need to be linked against libplugin.

manuelm:
> To clarify: When the plugin interface does not provide the plugin configuration you would need in the binding a way to issue elektraPluginGetConfig and thus also the bindings would need to be linked against libplugin.

"No" is the current answer. Just take a look at the `open`-signature of the python or lua plugin and you'll see there's no handle (which wouldn't work anyway) or config passed. But to fix this we only need to change the signature and pass the config keyset. No need for changing anything else or linking against `-plugin`.

Btw, can you please tell someone (not me) to write a elektra plugin using the python plugin? Such bugs won't be found unless someone starts using them.

[...]

PR #398, opened 2015-11-23
lua+python: add config keyset to open-call

manuelm:
As discussed in #394 this adds the config keyset (from elektraPluginGetConfig) to the open-call in both scripting plugins


markus2330:
Thank you! Looks very nice!

PR #399, opened 2015-11-23
bindings: only link against lib-core and lib-kdb

manuelm:
Last one for today. Please merge *after* libproposal issue has been solved. Otherwise gi-bindings won't work anymore.

Edit: Or reject in case you don't think it's worth it :-)

### DIFF
--- a/src/bindings/glib/CMakeLists.txt
+++ b/src/bindings/glib/CMakeLists.txt
@@ -22,7 +22,7 @@ else()
 
 	add_library (${GELEKTRA_LIBRARY} SHARED ${SOURCES})
 
-	tool_link_elektra (${GELEKTRA_LIBRARY})
+	target_link_libraries (${GELEKTRA_LIBRARY} elektra-core elektra-kdb)
 	target_link_libraries (${GELEKTRA_LIBRARY} ${GOBJECT_LIBRARIES})
 
 	install (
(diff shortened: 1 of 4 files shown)


markus2330:
libproposal issue should be resolved now. Can you check if it works now?

Is it necessary to specifiy elektra-core if you already link against elektra-kdb?

manuelm:
jenkins build bindings please

manuelm:
jenkins will find out for us.

Ofc the resolver will pull in elektra-core. But I prefer to explicitly link against all required libraries.

markus2330:
It compiles and ldd shows the correct information. The question remains if this is a cmake-feature (then it would be supported on every cmake-platform) or if its a gnu-ld feature.

manuelm:
If what is a gnu-ld feature? Linking against multiple libraries?

markus2330:
I am talking about that maybe [GNU ld from binutils](https://sourceware.org/binutils/docs/ld/) has a feature which magically links all depend libraries, too.

manuelm:
I'm sorry, I don't get it. Bindings depend on elektra-core + elektra-kdb + language interpreter. I'm linking against all of them + std c++ lib. Which other lib are you seeing or referring to?

markus2330:
Ok, we are talking past each other. I was talking if it is *necessary* to specify all libs. I will merge this here, because obviously it won't hurt to specify it.

manuelm:
In fact explicitly specifying is also some kind of documentation :-)

markus2330:
If we all use it the same way, then yes. Would you specify libcore if you only use `kdb*` functions? (indirect usage)

manuelm:
No, because I didn't use any of its symbols.

Issue #403, opened 2015-11-23
With preserveorder, sections are lost on `kdb set`


omnidan:
I have an ini file mounted on `system/test` with `preserveorder=1`:

```
[test]
hello = world
```

Running `kdb set system/test/test/hello world2` results in:

```
test/hello = world2
```

Without `preserveorder=1` it works fine:

```
[test]
hello = world2
```

markus2330:
This should be resolved

Issue #355, opened 2015-11-10
ini: allow repeated entries


markus2330:
automatically build up arrays on repeated entries:
```
[sec]
a=1
a=2
```
(when parameter array is given), see also #182 

So above would result in the following keyset:
```
sec/a   with metadata array="#1"
sec/a/#0=1
sec/a/#1=2
```

The keyset without the metadata `array`, the ini file would be written to:
```
[sec]
a=
a/#0=1
a/#1=2
```

@strahlex any opinions on that?

strahlex:
Looks fine for me. Context is representing a array in an INI file e.g.:
````ini
[DISPLAY]
program = .ngc GCode file
program = .py Python file
````
resulting in list like
````python
program = ['.ngc GCode file', '.py Python file']
````

**Comment by markus2330** [comment]
Yes, exactly, an array is conceived to be transformed to container datastructures, such as a list in python.

**Comment by markus2330** [comment]
Should work now (needs mount option "array"), e.g.:
```
$ kdb mount example.ini /example ini array=
$ cat `kdb file user/example`
key = value1
key = value2
key = value3
$ kdb export user/example ini
key = 
key/#0 = value1
key/#1 = value2
key/#2 = value3
```

**Comment by markus2330** [comment]
Seems to work now with option "array"
```
[sec]
a=1
a=2
```
in Elektra with:
```
$ kdb mount  $(realpath test.ini) system/ini/array ini array=
$ kdb ls system/ini/array
system/ini/array
system/ini/array/sec
system/ini/array/sec/a
system/ini/array/sec/a/#0
system/ini/array/sec/a/#1
$ kdb set system/ini/array/sec/a/#1 x
```

Questionable, however is:
```
kdb set system/ini/array/sec/a/#2 new 
```
leads to the file:
```
[sec]
a = 1
a = x
a/#2 = new
```

strahlex:
It would be preferable if writing works too. However, read-only should be sufficient right now.

markus2330:
Of course, writing configuration is an equally important goal for Elektra. I add the missing features in the issue  #388.

markus2330:
Seems to work now in both directions!

Issue #354, opened 2015-11-10
ini: removes parts of names


markus2330:
there is a critical bug that destroys names in the ini files.

```
> kdb rm -r user/inibug
> kdb set user/inibug/MyApp/mykey new_value
> kdb set user/inibug/binarytest
> kdb set user/inibug/debienna/test value
> kdb export user/inibug ini
MyApp/mykey = new_value
[binarytest]
st = value
```
@tom-wa I am not sure when it was introduced, but I think its better if you debug/fix it because you worked with ini recently.

fberlakovich:
Some info I could gather: the bug is rooted in `getIniName`. The bug happens when the previously written key is a section (in this case `user/inibug/binarytest`), but the next key is not below this section (in this case `user/inibug/debienna/test`). `getIniName` then incorrectly treats the key as below the section key and the expression `(char *)keyName(key)+strlen(keyName(section))+1` starts at a wrong position.

markus2330:
@tom-wa please also provide a testcase for regressions.

markus2330:
Build server still has not built it (see #160)

@manuelm if you find time, would be good if you can install ruby-ronn

omnidan:
This still isn't fixed, running those commands with the latest elektra build results in:

```
MyApp/mykey = new_value
[binarytest]
debienna/test = value
```

`debienna/test` shouldn't be under the `[binarytest]` section.

tom-wa:
this particular case should work using the autosections flag.
the problem can't really be fixed in the current version because the plugin has ne concept of ordering, but i've already found a solution for it in the newer version that should be ready in about a week.

fberlakovich:
@markus2330: since this bug is rooted in the new implementation of ini, could you please assign @tom-wa. For some reason I cannot assign him.

markus2330:
@fberlakovich thanks for the hint. Could you please also comment on #377?

markus2330:
Seems to work nicely now, new output:
```
[MyApp]
mykey = new_value
[binarytest]
[debienna]
test = value
```

omnidan:
Yep, works for me too.

PR #405, opened 2015-11-27
Enhanced ini 2

tom-wa:
TODO: 
write a README

Fixed:
#403 
#355 
#354
#424


markus2330:
@fberlakovich can you also take a look over it?

src/plugins/ini/README.md
```diff
@@ -1,156 +1,42 @@
 - infos = Information about ini plugin is in keys below
-- infos/author = Felix Berlakovich <elektra@berlakovich.net>
+- infos/author = Name <name@libelektra.org>
```
markus2330: There should be a name here

src/plugins/ini/ini.c
```diff
@@ -23,17 +26,20 @@ int elektraIniClose(Plugin *handle, Key *parentKey);
 
 #include "contract.h"
 
+#define INTERNAL_ROOT_SECTION "GLOBALROOT"
```
markus2330: What about nameconflicts (a key could be named GLOBALROOT)?

src/plugins/ini/ini.c
```diff
@@ -86,58 +92,192 @@ static Key *createUnescapedKey(Key *key, const char *name)
...
+}
+
+static void insertNewKeyIntoExistendOrder(Key *key, KeySet *ks)
+{
+	if(keyGetMeta(ksLookup(ks, key, KDB_O_NONE), "order"))
+		return;
+	ksRewind(ks);	
+	Key *curKey;
+	Key *prevKey = NULL;
+	while((curKey = ksNext(ks)) != NULL)
```
markus2330: use spaces between key words and `(`

src/plugins/ini/ini.c
```diff
@@ -198,13 +357,11 @@ int elektraIniOpen(Plugin *handle, Key *parentKey ELEKTRA_UNUSED)
 	KeySet *config = elektraPluginGetConfig (handle);
 	IniPluginConfig *pluginConfig = (IniPluginConfig *)elektraMalloc (sizeof (IniPluginConfig));
 	Key *multilineKey = ksLookupByName (config, "/multiline", KDB_O_NONE);
-	Key *autoSectionKey = ksLookupByName(config, "/autosections", KDB_O_NONE);
-	Key *toMetaKey = ksLookupByName(config, "/meta", KDB_O_NONE);
-	Key *preserveKey = ksLookupByName(config, "/preserveorder", KDB_O_NONE);
-	pluginConfig->preserverOrder = preserveKey != 0;	
+	Key *simpleKey = ksLookupByName(config, "/simple", KDB_O_NONE);
```
markus2330: What is a simple key?

src/plugins/ini/ini.c
tom-wa: Any suggestions for a better name ? 
The name should be something easy to remember, because it can be used to create global ini keys with kdb set having the same name as a section key. e.g. having `[section1]` and `section1 = i_am_a_global_key`

src/plugins/ini/ini.c
tom-wa: proably not the best name for the variable, but the idea was to use the `simple` as a configuration parameter to output the keyset the same way simpleini does (no "real" sections, just section/key = value pairs)

src/plugins/ini/ini.c
markus2330: Why would you need the GLOBALROOT then? I would expect that
`kdb set user/inimountpoint/section1 i_am_a_global_key`
already writes into the global section?

src/plugins/ini/ini.c
markus2330: Maybe `nosection`? Or a enum `/section` which is `none`, `null`, or `always`. (`null` means that for null-keys you get a section, which is the current behaviour of ini-plugin)

src/plugins/ini/ini.c
tom-wa: only if no section key `user/inimountpoint/section1`  - ( `[section1]` ) exists, otherwise kdb set would just overwrite the key, and i'm not sure if it would be possible to restore it.

src/plugins/ini/ini.c
```diff
@@ -86,58 +92,210 @@ static Key *createUnescapedKey(Key *key, const char *name)
...
+		return;
+	ksRewind(ks);	
+	Key *curKey;
+	Key *prevKey = NULL;
+	while((curKey = ksNext(ks)) != NULL)
+	{
+		if(!strcmp(keyName(curKey), keyName(key)))
 		{
-			if (!(keyGetMeta(sectionKey, "order")))
+			const char *oldOrder = "000000001";
```
markus2330: use array syntax

src/plugins/ini/ini.c
```diff
@@ -86,58 +92,210 @@ static Key *createUnescapedKey(Key *key, const char *name)
...
+			if(keyGetMeta(prevKey, "order"))
+				oldOrder = keyString(keyGetMeta(prevKey, "order"));
```
markus2330: use {} here

src/plugins/ini/ini.c
```diff
@@ -86,58 +92,210 @@ static Key *createUnescapedKey(Key *key, const char *name)
...
-	snprintf(buf, sizeof(buf), "%ld", lastIndex);
-	++lastIndex;
-	keySetMeta(appendKey, "order", buf);
-	if (value == NULL)
+	existingKey = ksLookup(handle->result, appendKey, KDB_O_NONE);
+	if(existingKey)
+	{
+		if(strcmp(keyString(appendKey), "") || keyGetMeta(existingKey, "ini/array"))
+		{
+			if(handle->array)
```
markus2330: function for array handling? please add some comments

src/plugins/ini/ini.c
```diff
@@ -86,58 +92,210 @@ static Key *createUnescapedKey(Key *key, const char *name)
...
+			{
+
+				keySetMeta(appendKey, "ini/section", 0);
+				if(keyGetMeta(existingKey, "ini/array"))
```
markus2330: Give some overview what the meta data is for.

src/plugins/ini/ini.c
```diff
@@ -435,94 +926,38 @@ int elektraIniSet(Plugin *handle, KeySet *returned, Key *parentKey)
...
-
-		if (!strcmp (keyName(current), keyName(parentKey))) continue;
-		if (isSectionKey(current))
+		if(keyGetMeta(cur, "order"))
 		{
-			keyDel(sectionKey);
-			sectionKey = keyDup(current);
-			keySetBinary(sectionKey, 0, 0);
-			current = sectionKey;
+			if(atoi(keyString(keyGetMeta(parentKey, "order"))) < atoi(keyString(keyGetMeta(cur, "order"))))
```
markus2330: What if a / is in order?

src/plugins/ini/ini.c
```diff
@@ -15,6 +15,11 @@
 #include <stdlib.h>
 #include <string.h>
 #include <kdberrors.h>
+#include <kdbproposal.h>
+#include <kdbprivate.h>
```
markus2330: Is this really necessary? Can you write as comment for which functions?

tom-wa:
@markus2330 done

PR #406, opened 2015-11-29
qt-gui: 0.0.9

0003088:
I completely forgot to create this pull request, it was supposed to be in the last release.


markus2330:
Thank you for PR!

Ohh, ok, some fixes and the update to "0.0.9" was already merged. I think it should be called 0.0.10 or 0.1.0 then?

markus2330:
I now only get:
```
QQmlApplicationEngine failed to load component
qrc:/qml/main.qml:231 Type GUIActions unavailable
qrc:/qml/GUIActions.qml:1 module "QtQuick" version 2.4 is not installed
```
(no main window shown)

Is it possible to also support an older version of QtQuick? (At least the one in Debian Jessie, I think its 2.2)

markus2330:
Seems like changing the header is enough:
```
diff --git a/src/tools/qt-gui/qml/GUIActions.qml b/src/tools/qt-gui/qml/GUIActions.qml
index bad41ed..c3708be 100644
--- a/src/tools/qt-gui/qml/GUIActions.qml
+++ b/src/tools/qt-gui/qml/GUIActions.qml
@@ -1,5 +1,5 @@
-import QtQuick 2.4
-import QtQuick.Controls 1.3
+import QtQuick 2.2
+import QtQuick.Controls 1.2
 import "MainFunctions.js" as MFunctions
 
 Item {
diff --git a/src/tools/qt-gui/qml/MainMenuBar.qml b/src/tools/qt-gui/qml/MainMenuBar.qml
index 74b853f..e653970 100644
--- a/src/tools/qt-gui/qml/MainMenuBar.qml
+++ b/src/tools/qt-gui/qml/MainMenuBar.qml
@@ -1,4 +1,4 @@
-import QtQuick 2.3
+import QtQuick 2.2
 import QtQuick.Controls 1.2
 import QtQuick.Controls.Styles 1.2
```

0003088:
I'll just increase the version number then (on a sidenote, it seems that I do not get any notifications anymore. My tu email address is probably expired, will have to change it)

markus2330:
Thank you!

PR #408, opened 2015-12-05
getenv: fix wrapping on powerpc

pinotree:
On powerpc architectures `__libc_start_main` has different arguments, so fix the signature for the wrapper `__libc_start_main` and the arguments passed to the real one.


markus2330:
Thank you, great fix!

PR #409, opened 2015-12-05
markdownlinkconverter: fix char/int mismatch

pinotree:
The return value of `fgetc` and parameter for `isblank` is `int`, not `char`.
This fixes markdownlinkconverter on architectures where `char` is unsigned (e.g. powerpc, s390, arm).


markus2330:
Thanks for the fix!

@KurtMi here a real-life example for portability

PR #410, opened 2015-12-05
wresolver: use KDB_MAX_PATH_LENGTH instead of PATH_MAX

pinotree:
We have the own define to use for a maximum path length, so use it to build also on platforms without `PATH_MAX`.


markus2330:
Thank you, nice catch!

PR #413, opened 2015-12-06
lua: fix Key:tostring()

pinotree:
`self->getName()` returns a new temporary `std::string`, and passing its internal buffer as return value of `__tostring` means using free'd memory.

Copy a bit of the `kdb::Key::getName()` implementation, returning the raw const data from the underlying `Key` object.

### DIFF
--- a/src/bindings/swig/lua/kdb.i
+++ b/src/bindings/swig/lua/kdb.i
@@ -229,7 +229,9 @@
   }
 
   const char *__tostring() {
-    return self->getName().c_str();
+    ckdb::Key *rawKey = self->getKey();
+    if (!rawKey) throw kdb::KeyException();
+    return ckdb::keyName(rawKey);
   }
 
   myswig::LuaSTLIterator_T<kdb::Key::iterator> *name_iterator() {


markus2330:
thank you!

@manuelm any objections?

manuelm:
Thank you. But the better fix is to just return std::string and let swig do the rest.

pinotree:
> But the better fix is to just return std::string and let swig do the rest.

Sorry, but the included patch is still wrong. In the generated `kdbLUA_wrap.cxx` I see the snippet for the lua `__tostring` function:
```c++
static int _wrap_Key___tostring(lua_State* L) {
  int SWIG_arg = 0;
  kdb::Key *arg1 = (kdb::Key *) 0 ;
  char *result = 0 ;
  
  SWIG_check_num_args("kdb::Key::__tostring",1,1)
  if(!SWIG_isptrtype(L,1)) SWIG_fail_arg("kdb::Key::__tostring",1,"kdb::Key *");
  
  if (!SWIG_IsOK(SWIG_ConvertPtr(L,1,(void**)&arg1,SWIGTYPE_p_kdb__Key,0))){
    SWIG_fail_ptr("Key___tostring",1,SWIGTYPE_p_kdb__Key);
  } 
  
  {
    try {
      result = (char *)kdb_Key___tostring(arg1);
    }    KEY_EXCEPTIONS    catch (const std::exception & e) {
      SWIG_exception(SWIG_RuntimeError, e.what());     
    }    catch (...) {
      SWIG_exception(SWIG_UnknownError, "unknown error in kdb::Key::__tostring()");
    } 
  } 
  lua_pushstring(L,(const char *)result); SWIG_arg++;
  return SWIG_arg;
  
  if(0) SWIG_fail;
  
fail:
  lua_error(L);
  return SWIG_arg;
} 
```
So, please note:
```c++
      result = (char *)kdb_Key___tostring(arg1);
```
which is basically doing the same, i.e. talking the pointer to the internal buffer of the temporary (!) `std::string`. Please reopen this issue, and consider my original patch for this, thanks.

manuelm:
Make sure to regenerate the generated code. You're probably looking at the old one.
SWIG is using `lua_pushlstring` for `std::string`: https://github.com/swig/swig/blob/master/Lib/lua/std_string.i#L54

pinotree:
Oh, I see. Thanks for the notice, and the fix.

PR #414, opened 2015-12-06
python: fix flags in Key()

pinotree:
One of the constructors of `Key` takes `uint64_t`, coming from what swig maps for an `int` value. OTOH, passing a `uint64_t` as vararg value for `KEY_FLAGS` will lead to mismatches later, as `KEY_FLAGS` is read as `int`.

Hence, check that the value passed in the `uint64_t` can fit into an `int`, and cast it as such when passing it to `KEY_FLAGS`.

### DIFF
--- a/src/bindings/swig/common.i
+++ b/src/bindings/swig/common.i
@@ -23,6 +23,8 @@
   #include "keyset.hpp"
   #include "kdb.hpp"
   using namespace kdb;
+
+  #include <limits>
 %}
 
 %apply long { ssize_t }
--- a/src/bindings/swig/python/kdb.i
+++ b/src/bindings/swig/python/kdb.i
@@ -120,11 +120,17 @@
   $result = PyBytes_FromStringAndSize((const char*)$1, (size > 0) ? size : 0);
 }
 
+%typemap(check) (char const *name, uint64_t flags) {
+  if ($2 > std::numeric_limits<int>::max()) {
+    SWIG_exception(SWIG_ValueError, "Too big value for flags in Key()");
+  }
+}
+
 // add some other useful methods
 %extend kdb::Key {
   Key(const char *name, uint64_t flags = 0) {
     return new kdb::Key(name,
-      KEY_FLAGS, flags,
+      KEY_FLAGS, static_cast<int>(flags),
       KEY_END);
   }
 


markus2330:
thank you!

@manuelm any objections?

pinotree:
Ah, the lua swig binding has a basically the same issue. I can send a new PR for that as well, but I'd like to get this merged first (so it has been reviewed and eventually changed).

manuelm:
I'm not fine with that.

`KEY_FLAGS` always has been `uint64_t` (see c25bf8eb57f3966163d2fec982cad770ca4171d2 + 9401282fa45a10388c2061e5cf7cbafa7bbc5130). aca7c1e86285319bb41dde8eb3e5feb71a902931 changed it to an `int` without checking further implications.

Changing everything (SWIG+gobject+GI) to `int` is wrong as we'll have the same bug report again as soon as the available flags exceed the 32bit range.

So the correct fix is to change `KEY_FLAGS` back to `uint64_t` and make sure the compilers will always pass the right type, even if the uint32_t range hasn't been reached.

Does anyone know of some compiler magic or is the only fix to either always cast or introduce a noop-enum which will set the MSB?

markus2330:
enum beyond 32bit is compiler-specific, so I am afraid hacks to disable or enable the extension will be specific, too.

it would not be wise to support flags that need more than 31bits and I do not intend to do so.

Which compiler makes a type larger than 4 bytes if enums are or-ed?

pinotree:
> KEY_FLAGS always has been uint64_t (see c25bf8e + 9401282). aca7c1e changed it to an int without checking further implications.

I'm sorry that aca7c1e apparently brought this issue, but the code was already broken. In particular, we have three cases here:

1. ```new Key(name, KEY_FLAGS, int_variable, ...)```
2. ```new Key(name, KEY_FLAGS, KEY_FOO | KEY_BAR, ...)```
3. ```new Key(name, KEY_FLAGS, uint64_t_variable, ...)```

(1) and (2) are what is usually done in the code, either in the own elektra one (C, C++), or by 3rd party code. (3) is what current lua and python3 swig-based do.

If you pass a `int` variable, or that is implicitly represented as such, it will be read as `int` (and thus aca7c1e). If you pass a `uint64_t` code and the vararg read as `int`... code breaks, and this is what this patch wants to fix.

> So the correct fix is to change `KEY_FLAGS` back to `uint64_t` and make sure the compilers will always pass the right type, even if the `uint32_t` range hasn't been exceeded.

You cannot do this. Even if you revert aca7c1e, you still need to manually cast values passed to `KEY_FLAGS` **everywhere** (missing one would imply a bug), but happily breaking 3rd party code.

You cannot really pick yourself the size of an enum in C (only C++ >= 11 allows that IIRC), so the only reliable way is to use `int` for integer-like values passed to varargs.
Regarding more than 31 flags, in this case the only reliable solution is to introduce a new set of flags, e.g. `KEY_FLAGS_2`, and have users pass them like that. Sure, it is ugly as hell, but that's the only portable way to do this.

Btw:
* http://stackoverflow.com/questions/366017/what-is-the-size-of-an-enum-in-c
* http://stackoverflow.com/questions/9972345/sizeof-enum-in-c-language
* http://stackoverflow.com/questions/1113855/is-the-sizeofenum-sizeofint-always

manuelm:
Fixed. I really hate the va_args usage in elektra. It's way to easy to get into undefined behavior.

markus2330:
thank you both for detailed analysis and fixing it!

PR #417, opened 2015-12-09
	Cleaning up #ifdefs that break statements

RomeroMalaquias:


src/plugins/yajl/yajl_parse.c
```diff
@@ -395,12 +396,11 @@ int elektraYajlGet(Plugin *handle ELEKTRA_UNUSED, KeySet *returned,
 		{
 			stat = yajl_parse(hand, fileData, rd);
 		}
-
-		if (stat != yajl_status_ok
+		test_status = (stat != yajl_status_ok);
```
markus2330: variable could also be declared here

markus2330:
Thanks for the contribution! Did you test if yajl1 still works?

markus2330:
jenkins build please

RomeroMalaquias:
Hi @markus2330, i didn't tested, sorry. 
If you approve that change and want the declaration of variable in another place, i can replace it.

markus2330:
I agree that the change is an improvement (with variable declarations in either way, even though I prefer the declaration as late as possible). Testing it would be good though, I do not have yajl1 easily available anymore. Is there any specific reason you need the change, or is it just to check if your PRs are accepted for further contributions?

RomeroMalaquias:
it's just to check if developers care about the discipline of conditional annotations

markus2330:
Interesting study. Was this checked and/or repaired by a tool? Seems to be correct. Would be nice if you can move the variable within the block.

RomeroMalaquias:
We have an tool to point possible candidates to refactoring, but still don't have one to test.
Because of this, i look for refactoring that aren't much complex.

markus2330:
Ok, thanks for the fix. Further improvements are welcome.

Issue #418, opened 2015-12-10
Provide global initialization and destruction


markus2330:
As initially proposed by @manuelm and now by @petermax2  in #276 we could add global initialization and destruction.

Use case: Some plugins need libraries (or interpreters) which need to be initialized exactly once. (currently: crypto, syslog, python)

Ways to resolve it:
1. (Current way) when plugin is opened (+ will only be done when plugin is used, + users do not have to care about it, - need reference counting library facilities to work correctly with other libraries doing the same thing, as workaround a `noinit` plugin configuration can be provided)
2. provide some form of `kdbInit()` and `kdbFini()` that callback what plugins want them to do (- API for user gets more difficult and gets more state, - setup might happen for non-used plugins, - Cannot be used in within-plugin context)
3. global __constructors__ (- also needs ref counting for libraries, - not portable)
4. @manuelm mentioned something about apache?

If we agree to (1) we should also agree to a convention how the `noinit` plugin configuration should be named.

petermax2:
As for the crypto plugin number 1 would be great. I just want to know if I'm the first plugin of my kind. Some global destruction mechanism could be useful in the future (like cleaning up secure memory ares and stuff like that)

markus2330:
Do you really mean 1.? (thats status quo)

Knowing that you are the first plugin still gives no information if the application or other libraries have already initialized libssl before.

I am afraid that there is no solution for the general problem: You cannot fix every broken API design by an additional layer. But I am looking forward that someone can convince me otherwise and we find a nice solution for python+libssl+syslog init and especially cleanup!

I think the best solution is to avoid global state all together (current approach in Elektra) and the second best solution is a reference-counted atomic "give me the handle if it already exists or create one otherwise" similar to the JNI API. Its only the second best because 1. forces some multi-thread decisions to the application and/or 2. fail if you need (different) configuration for the initialization.

Nevertheless, I think the discussion is necessary because it is a recurring problem with many APIs.

manuelm:
What about moving the decision about initializing and final destruction to the end-user/application developer (API consumer)? As said many times we - as a library - are not able to fix this, just so some workarounds.

petermax2:
@markus2330 do we have some reference counting facility in Elektra right now? Because knowing at `OPEN` that this is instance 1 and knowing at `CLOSE` that 0 instances are remaining after this plugin, would be sufficient for my needs. I don't know about python tough.

manuelm:
[...]

PR #419, opened 2015-12-10
Crypto-Plugin: ref counting for library initialization (WIP)

petermax2:
I prepared the library initialization using a simple ref counter (protected by a mutex) as preparation for the planned OpenSSL support.

- [x] investigate crypto API based ref counting
- [ ] check if gcrypt has been initialized (possibly by an application)

### DIFF
--- a/doc/todo/FUTURE
+++ b/doc/todo/FUTURE
@@ -114,6 +114,9 @@ maybe kdbConstruct and kdbDestruct needed too?
 
 allow non-file/resolver based backends
 
+read function symbols from contract (and not duplicated)
+	have elektraPluginContract separated from elektraPluginGet?
+
 
 
 == Error handling ==
--- a/src/bindings/glib/gelektra-key.c
+++ b/src/bindings/glib/gelektra-key.c
@@ -192,7 +192,7 @@ static void gelektra_key_gi_init_va(GElektraKey *key, const gchar *name, ...)
  * \note This is for GObject Introspection.
  * \note Do NOT use! Use gelektra_key_new instead
  */
-void gelektra_key_gi_init(GElektraKey *key, const gchar *name, guint64 flags,
+void gelektra_key_gi_init(GElektraKey *key, const gchar *name, int flags,
 	const gchar *value, const void *data, gsize data_size)
 {
 	gelektra_key_gi_init_va(key, name,
--- a/src/bindings/glib/gelektra-key.h
+++ b/src/bindings/glib/gelektra-key.h
@@ -71,7 +71,7 @@ GElektraKey *gelektra_key_make(Key *key);
 GElektraKey *gelektra_key_gi_make(GElektraKey *key);
 
 /* initialization */
-void gelektra_key_gi_init(GElektraKey *key, const gchar *name, guint64 flags,
+void gelektra_key_gi_init(GElektraKey *key, const gchar *name, int flags,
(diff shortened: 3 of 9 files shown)


markus2330:
Thank you!

I think the abstraction can be improved: the ref counting schema is needed for openssl, but for others there might be a better solution (atomic refcounting already as part of a crypto-API).

So why not put the refcounting to the gcrypt and openssl provider of `elektraCryptoInit`? (No library has such ref counting is a valid answer: then the choice of abstraction is correct.)

markus2330:
Btw. a way to avoid init/cleanup might be needed to. (e.g. If application already configured openssl)

petermax2:
gcrypt does not seem to have any ref counting built into its public API. They might be using it internally  (I read some old discussions about it in developer forums). Libssl uses ref counting internally but they don't seem to make it public either. Libcrypto does not use any ref counting at all but provides a lock mechanism through its API. [1, 2] However this is the part of the API that we try to initialize here. ;)

[[1] The Libgcrypt Reference Manual](https://gnupg.org/documentation/manuals/gcrypt/index.html#Top)
[2] Viega, Messier and Chandra. Network Security With OpenSSL. O'Reilly. 2002.

markus2330:
Ok, then thank you for the PR! Some plugin options and maybe also as metadata in kdbOpen()/kdbClose() to enable/disable tearup/down should be added later, too. (and the check if gcrypt has been initialized)

I merged it, because it was a nicely coupled unit, please create a new PR then.

petermax2:
Thank you! 

OK, I noted the metadata check. Checking if gcrypt has already been set up is not supported by the API. In general the manual suggests to leave all of the initialization stuff to the application developer. So OpenSSL will make some things easier for us.

markus2330:
An additional remark to the internal ref counting in gcrypt/libssl: If they do it correctly, it should not matter if you reopen/reclose it as long as the number of open and close matches?

What happens if the crypto libs have multiple init/teardown? Did you test it?
The default behaviour has to be chosen carefully based on that.

For plugin configuration we should,however, agree to a naming convention in #418 first.

petermax2:
For now it does not matter if the setup routines have been called multiple. I tested it for both Gcrypt and libcrypto. That may change in the future. The gcrypt folks are planning on dropping the initialization alltogether in the future but they didn't make it clear when that will be.

markus2330:
That sounds good! Did you also test the case with complete shutdown and restart? That caused problems in python.

So the ref counting introduced in this PR is actually not needed?

petermax2:
gcrypt does not have to shutdown, I'm not sure about OpenSSL yet. The one thing I did not test yet is how these libraries behave when multiple threads are running and the library gets re-initialized in between operations. OpenSSL shouldn't have any problems as they internally just update a function pointer.

The thing with ref counting is: I want to comply to the guidelines and manuals of the crypto APIs as much as possible. And they all suggest to initializing the libraries as early as possible and call these routines only once. So maybe in a future update something breaks if we call the setup routines multiple times. Also if libgcrypt starts in FIPS-enabled mode the library does additional setup, that is supposed to run only once in the lifetime of the application. The same is true for libcrypto. I think the self tests in FIPS-mode may throw errors if executed multiple times. It all depends on versions and compile/startup options and we can not be careful enough.

markus2330:
Ok, thank you! So better we have the ref-counting in place!

PR #430, opened 2015-12-27
shell recorder

tom-wa:


tests/shell/shell_recorder/script.dat
```diff
@@ -0,0 +1,13 @@
+Storage: dump
+Mountpoint: system/testmount/test
+File: /tmp/test.dump
+MountArgs:
```
markus2330: If I understand it correctly "Storage:" is the specification in which format the recorder does import/export. What do we need MountArgs for?

tests/shell/shell_recorder/script.dat
```diff
@@ -0,0 +1,13 @@
...
+DiffType: File
+< mount $File $Mountpoint $Storage
```
markus2330: Is there also an interactive way for using the recorder?
Why is the `<` needed here? Why not syntactically distuingish the header and commands (e.g. via newline)?

tests/shell/shell_recorder/shell_recorder.sh
```diff
@@ -0,0 +1,120 @@
...
+    echo "RET: $RETVAL" >> ./protocol
+    echo "ERRORS: $ERRORS" >> ./protocol
+    echo "WARNINGS: $WARNINGS" >> ./protocol
+    echo -e "STDOUT: $STDOUT\0" >> ./protocol
+    echo -e "STDERR: $STDERR\0" >> ./protocol
+    echo -e "DIFF: $DIFF\0" >> ./protocol
+    echo >> ./protocol 
+}
+
+rm ./protocol 2>/dev/null
```
markus2330: What is the `cwd` here?

markus2330:
Thank you, very nicely done, certainly goes in a good direction!

tests/shell/shell_recorder/script.dat
tom-wa: Storage is the storage plugin, i was planing on renaming it later . 
MountArgs is just a String that gets attached after the storage plugin name, like additional plugins, plugin configs, etc
All those variables are optional.

tests/shell/shell_recorder/script.dat
tom-wa: yeah, i was already thinking about splitting the header and the commands, i'm not completely set on the syntax yet.

tests/shell/shell_recorder/script.dat
markus2330: So they are only variables? Why not allow any variable `*: *`?

tom-wa: i was actually planing on doing this, but i couldn't find a way without using eval(1) yet.

tests/shell/shell_recorder/globaltest.dat
```diff
@@ -0,0 +1,34 @@
...
+< set system/elektra/globalplugins/precommit list
+< set system/elektra/globalplugins/pregetstorage list
+< set system/elektra/globalplugins/postgetstorage list 
+< set system/elektra/globalplugins/presetstorage list
+< set system/elektra/globalplugins/prerollback list
+
+STDOUT: (tracer.*){3}
+STDERR: (set.*di.*pos.*){3}
+RET: 0
+< set system/globaltestkey
```
markus2330: To which command does the STDOUT/STDERR refers to? Is the order important (command first, or STDOUT/STDERR first?)? How to specify the effects in KDB?

markus2330: Can you give a highlevel overview which commands do I need to enter to record, which output files do I get then; and how do I replay?

tom-wa: STDOUT/STDERR/RET/WARNINGS/ERRORS/DIFF always refer to the next command (< ...)
Their syntax is posix extended regex and they are matched against the output/return values of the next command.

The only thing that needs to be specified is "Mountpoint" because it's used to lookup the configuration file, for exports/comparisons, and for backups/restoring of the original state.

i've added some more examples:

the shell recorder creates a copy of the configuration/key file before running a command and then diffs them to keep track of all changes. 
db_changes.dat is an example of the "DIFF"-Operation which can be used to check if the changes are what we expected.


replay_test_fail.dat and replay_test_succeeds.dat are both examples of the replay function.
if the script contains the string "REPLAY!" the shell recorder runs the script twice, keeps a copy of the protocol and the configuration file from the first iteration, and compares them against the results from the second one.

tests/shell/shell_recorder/script.dat
markus2330: Why do you think untrusted code could be a problem? Isn't everything within the repo?

tests/shell/shell_recorder/globaltest.dat
markus2330: Thank you! Can you add a README.md with that content?

markus2330:
Is this finished? Can you please create a small tutorial how to use it?

markus2330:
@tom-wa Is this finished? Can you please create a small tutorial how to use it?

tests/shell/shell_recorder/README.md
```diff
@@ -0,0 +1,331 @@
+# Shell Recorder #
+
+## Configuration ##
+
+### Mountpoint: ###
+
+This is the only configuration Variable that has to be set. It is used to determine where the shell_recorder should look for changes.
```
markus2330: Please add some introduction what a "configuration Variable" is in the context of a shell_recorder.

tests/shell/shell_recorder/README.md
```diff
@@ -0,0 +1,331 @@
...
+e.g. `Mountoint: user/test` tells the shell_recorder that you will be working under `user/test`.
+
+### REPLAY! ###
+
+When `REPLAY!` is set the shell_recorder runs the input script twice and compars the results.
```
markus2330: I think this should work differently: The idea is that you record once the output of an input script. Then we add this recorded file to libelektra git repo. From then on, the build server replays the same input script and sees if the output still matches.

tests/shell/shell_recorder/README.md
tom-wa: I'm still searching for a more appropriate term

tests/shell/shell_recorder/README.md
```diff
@@ -0,0 +1,285 @@
...
+* STDOUT:
+* STDERR:
+* RET:
+* WARNINGS:
+* ERRORS:
+* DIFF:
+
+## Commands ##
+
+A command starts with `<` followed by the usual kdb commands.
```
KurtMi: You should write also that kdb is omitted

KurtMi:
In my point of view the README is ok, but could have less examples and more explanations.

markus2330:
@KurtMi Thank you for reviewing!

markus2330:
Not all review comments are worked in.

./protocol modifies itself, that should be fixed (either not hardcode the path, using a tempfile or at least not adding the file with this name)

tests/shell/shell_recorder/README.md
```diff
@@ -0,0 +1,285 @@
...
+DiffType: Dump
+Mountpoint: user
+
+DIFF: (> keyNew.*user/test/key)
+< set /test/key val
+
+DIFF: (> keyNew.*user/test/key3)
+< set /test/key2 val2
+
+ % ./shell_recorder.sh db_changes.dat
```
markus2330: Some explanation would be useful would you wanted to achieve here. Its unclear that a file "protocol" is recorded here.

markus2330:
How can I output the mounted file? (I did a set and a get and I did not get what I wanted, how to debug it?)

markus2330:
And how to cleanup temporary files? How to execute non-kdb shell commandos?

markus2330:
Looks really nice!

markus2330:
Any progress here?

tom-wa:
working on it right now.

markus2330:
Great, thank you!

markus2330:
Please ping if ready

markus2330:
jenkins build please

markus2330:
@tom-wa can you start working with the shell recorder and give us tomorrow a tutorial about how it works? I think it will be really useful for many things (especially real-world and global plugin tests)

markus2330:
No change here? Would be nice to integrate it before the meeting.

markus2330:
Thank you, for new improvements, please open a new PR

Issue #439, opened 2016-01-09
PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP is not portable


haraldg:
While musl libc supports recursive mutexes, the initaliziation macro appears to be specific to glibc, causing a build error:
```
/home/lambda/olinuxino/openwrt/build_dir/target-arm_arm926ej-s_musl-1.1.11_eabi/elektra-0.8.15dev/src/plugins/resolver/resolver.c:52:49: error: 'PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP' undeclared here (not in a function)
```
See this link for the same issue on an other project: https://github.com/Itseez/opencv/issues/5476

markus2330:
Thank you for reporting!

@petermax2 already wrote a solution for that problem (that exists in OS X <= 10.6, too). Later we reverted it, even though we knew it is actually not portable, because we thought it would not effect us anymore.

@petermax2 Can you add your same code again with a `#ifndef PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP`?

petermax2:
I'll check the history and I'll try to resolve this issue.

markus2330:
Thank you! It would be ideal if you try to compile Elektra with `CC=musl-gcc  ./scripts/configure-debian ~e` to see if the problem is gone for musl.

@haraldg You can easily avoid the issue by removing all resolver variants that contain `m` (for mutex).

haraldg:
markus2330 writes:
> @haraldg You can easily avoid the issue by removing all resolver variants that contain `m` (for mutex).

I tried it with:
        -DKDB_DEFAULT_RESOLVER=resolver_l_h_b \
        -DPLUGINS="DEFAULT;-resolver;resolver_l_h_b"

but got an error:
CMake Error at cmake/ElektraCache.cmake:338 (message):
  selected default resolver (resolver_l_h_b) is not selected in PLUGINS,
  please change KDB_DEFAULT_RESOLVER or PLUGINS

markus2330:
@haraldg: Sorry, the workaround does not work as I thought it would. First I will tell you a workaround that works, then I will go into the details why the other one did not work.

A solution that compiles  is to use `wresolver` instead of `resolver`:

`cmake ~e -DKDB_DEFAULT_RESOLVER=wresolver -DPLUGINS="dump;error;wresolver"`
or longer as I tried it:
[...]

PR #443, opened 2016-01-10
resolver: restore recursive mutex init (see #439)

petermax2:
as discussed in #439 I restored the manual initialization code for the recursive mutex in the resolver. The resolver compiles now under `musl-gcc` with the `ELEKTRA_LOCK_MUTEX` compile definition set.

### DIFF
--- a/src/plugins/resolver/CMakeLists.txt
+++ b/src/plugins/resolver/CMakeLists.txt
@@ -18,7 +18,7 @@ add_plugin(resolver
 		ELEKTRA_PLUGIN_NAME=\"resolver\"
 		ELEKTRA_LOCK_FILE
 		ELEKTRA_LOCK_MUTEX
-)
+	)
 
 add_plugintest(resolver)
 
--- a/src/plugins/resolver/CMakeLists.txt
+++ b/src/plugins/resolver/CMakeLists.txt
@@ -8,12 +8,17 @@ find_package(Threads)
 add_plugin(resolver
 	SOURCES
 		${SOURCES}
+	LINK_LIBRARIES
+		${CMAKE_THREAD_LIBS_INIT}
+		${CMAKE_REALTIME_LIBS_INIT}
 	COMPILE_DEFINITIONS
 		ELEKTRA_VARIANT_BASE=\"\"
 		ELEKTRA_VARIANT_USER=\"hpu\"
 		ELEKTRA_VARIANT_SYSTEM=\"b\"
 		ELEKTRA_PLUGIN_NAME=\"resolver\"
-	)
+		ELEKTRA_LOCK_FILE
+		ELEKTRA_LOCK_MUTEX
+)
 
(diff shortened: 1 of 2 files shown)


src/plugins/resolver/CMakeLists.txt
```diff
@@ -8,12 +8,17 @@ find_package(Threads)
 add_plugin(resolver
 	SOURCES
 		${SOURCES}
+	LINK_LIBRARIES
+		${CMAKE_THREAD_LIBS_INIT}
+		${CMAKE_REALTIME_LIBS_INIT}
```
markus2330: Would be better if this is in another commit (reverting my change). Not very important though.

src/plugins/resolver/resolver.c
```diff
@@ -47,10 +47,15 @@
 #endif
 
 #ifdef ELEKTRA_LOCK_MUTEX
-#ifndef PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP
+#if defined(PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP)
+static pthread_mutex_t elektra_resolver_mutex = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
```
markus2330: It should be camel case (elektraResolverMutex). Rename preferable also in a separate commit.

src/plugins/resolver/resolver.c
```diff
@@ -370,6 +375,37 @@ int ELEKTRA_PLUGIN_FUNCTION(resolver, open)
 	resolverInit (&p->user, path);
 	resolverInit (&p->system, path);
 
+#if defined(ELEKTRA_LOCK_MUTEX) && defined(ELEKTRA_RESOLVER_RECURSIVE_MUTEX_INITIALIZATION)
```
markus2330: `#ifdef ELEKTRA_RESOLVER_RECURSIVE_MUTEX_INITIALIZATION` should be enough here.

src/plugins/resolver/resolver.c
```diff
@@ -370,6 +375,37 @@ int ELEKTRA_PLUGIN_FUNCTION(resolver, open)
...
+	// so we use another mutex for the initialization of the recursive mutex,
+	// since this section must be thread safe.
+	pthread_mutex_lock(&elektra_resolver_init_mutex);
+	if (!elektra_resolver_mutex_initialized)
+	{
+		pthread_mutexattr_t mutex_attr;
+
+		if (pthread_mutexattr_init (&mutex_attr))
+		{
+			ELEKTRA_SET_ERROR(35, errorKey, "Could not initialize recursive mutex");
```
markus2330: the error message should contain where the error happened (attr_init, settype, init), and ideally also the error code.

markus2330:
Thank you, very nicely done!

markus2330:
Thanks!

PR #445, opened 2016-01-11
posix shell plugin

tom-wa:


src/plugins/shell/shell.c
```diff
@@ -0,0 +1,138 @@
...
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/types.h>
+#include <errno.h>
+#include <kdberrors.h>
+#include <kdbhelper.h>
+
+
+int elektraShellOpen (Plugin * handle ELEKTRA_UNUSED, Key * errorKey ELEKTRA_UNUSED)
```
markus2330: You can remove these!

src/plugins/shell/shell.c
```diff
@@ -0,0 +1,138 @@
...
+{
+	char *cmd[4];
+	cmd[0] = "/bin/sh";
+	cmd[1] = "-c";
+	cmd[2] = cmdline;
+	cmd[3] = NULL;
+	pid_t pid = fork();
+	if(pid == 0)
+	{
+		return execv("/bin/sh", cmd);
```
markus2330: is there any specific reason not to use `system(3)`?
Or is this work-in-progress and you want to get stdin/stdout?

tom-wa: I just pushed it because i had some problems building it and tried to figure out what was wrong. 
Actually, i was planning on redirecting stdout/stderr, but i can't think of any reason to use that now. 
Do you think parsing the output using regex or loading parts of it into elektra keys would be useful in any way ?

markus2330: It would be good to have a way to suppress output and, on errors, putting all text into the error reason would be nice.

Not really must-have features though.

Another usecase is to send/receive a keyset on stdin/stdout and use a shell script as filter. But I think this would be a different story.

tom-wa:
@markus2330 done

markus2330:
Thank you!

Issue #438, opened 2016-01-08
Doxygen creates directories ( ) in source


markus2330:
@sanssecours wrote: One thing I noticed while I tested the changes was that `cmake` also seems to produce two (useless?) directories with the names `(` and `)` inside `libelektra/doc`. Is this intentional?

markus2330:
No, the directories are not created intentionally, should be fixed.

sanssecours:
I looked a little bit into the issue. It seems that the macro `do_doc` creates the folders `(` and `)`, in the directory `doc` inside the root of the repo. Lets start with steps to reproduce the issue:

---

1. Go to the root folder of the cloned repository
2. Create a folder called `build`

  ```bash
  mkdir -p build
  ```
3. Change the directory to the folder you just created
  
  ```bash
  cd build
  ```
4. Run `cmake`
  
  ```bash
  cmake ..
  ```

After the last step `/doc` –  here `/` specifies the root of the repository – contains the directories `(` and `)`. 

---

For the following explanation, please keep in mind that – although I read both the [Wikipedia article](https://en.wikipedia.org/wiki/CMake) about CMake and the [tutorial](https://cmake.org/cmake-tutorial) on the official website – I do not know a lot about this build system 🙃.

As already mentioned above, the cause of the issue seems to be `do_doc`, more specifically [line 14](https://github.com/ElektraInitiative/libelektra/blob/master/doc/CMakeLists.txt#L14) of '/doc/CMakeLists.txt'. The command `file(MAKE_DIRECTORY(${folder}))` creates a directory even if `folder` is empty i.e. `folder` contains the empty list `()`. As to why the variable `folder` is not set, I do not know. 

One (super ugly option) to fix the bug would be to just ask if `folder` is set to something before we create it: 

```CMake
if (${folder})
  file(MAKE_DIRECTORY(${folder}))
endif ()
```

Another option would be to just delete line 14. That of course might bring back the issue that line 13 mentions:

> sometimes doxygen is too slow and fails with "Could not create output directory .../doc/html"

The two time I tried `make html` I encountered not problems. That might be a different story if we run the whole build process multiple times though.

markus2330:
You are on the right track! It won't, however, be fixed by adding an if (folder is always set). When you read about the syntax and how the cmake command `file` works, it will immediately obvious what the problem is. I will leave it to you, I think fixing this bug will be a good way to learn CMake ;)

PR #446, opened 2016-01-11
CMake: Do not create directories `(` and `)` inside `doc`

sanssecours:
From issue #438:

> You are on the right track! It won't, however, be fixed by adding an if (folder is always set). When you read about the syntax and how the `cmake` command file works, it will immediately obvious what the problem is. I will leave it to you to find out, I think fixing this bug will be a good way to learn CMake ;)

Thank you for the useful hints. Hope the description in the commit message is correct. If not, please just comment below. This pull request (hopefully) fixes issue #438.


markus2330:
Thanks, well done! The commit description is also very good.

PR #451, opened 2016-01-12
require Lua 5.2 or above

### LINKED ISSUE #450: lua plugins depends on lua 5.2 but doesn't detect so
lua.cpp uses LUA_OK, which was only defined in 5.2. however cmake will try to build the plugin if an older version of lua is installed.
(No idea why OpenWRT is still stuck with lua 5.1.5 ...)

I'm manually disabling the lua plugin to avoid triggering a build error should lua happen do be selected, so this is no big deal.

manuelm:
fixes #450


manuelm:
jenkins build bindings please

markus2330:
Thank you! What was the initial reason for providing the FindLua52.cmake? Or since when does cmake provide FindLua in the core distribution?

manuelm:
Afair I copied the find lua cmake stuff from another project. Didn't bother looking for better alternatives.

FindLua replaces FindLua52: https://github.com/Kitware/CMake/commit/89cfa2005207ae9423d7a200cd22ed62ff733161
Since 3.0

markus2330:
Thanks for looking it up. Actually, we should not take existence of FindLua for granted then, because we only require cmake 2.8.8

manuelm:
please vote:
- [ ] require cmake 3.0
- [ ] include a copy of FindLua.cmake

markus2330:
Are external cmake files preferred? If yes, lets ship it.
Debian Wheezy (where we said we would support for some time) only has cmake (2.8.9-1)...

manuelm:
Wheezy is oldstable and according to their rules the security team should drop it on April 26th, 2016. However than the LTS team will take over until May 2018. :dancers: 

Anyway, I will add a copy as an extra commit.

manuelm:
jenkins build bindings please

manuelm:
jenkins build bindings please

markus2330:
Thank you!

PR #455, opened 2016-01-18
Conditionals: added new syntax for value assignments

tom-wa:


markus2330:
Thank you! Looks really nice. I was thinking, however, that not the metadata `check/condition` (its name would be contradictory) should be used for this feature, but instead the new metadata `assign`.

And it would be good to have examples for the feature, too. (Or even better: shell test cases embedded in the README.md.)

markus2330:
jenkins build please

tom-wa:
@markus2330 would you prefer "assign/condition" or "condition/assign" as metakey ?

markus2330:
Why not simply use `assign`? Do you plan to implement `assign` for math-plugin, too (somewhere in future)? Then use `assign/condition` and `assign/math`.

Btw. in the contract you now have to state all metadata a plugin can take care of.

Thank you!

tom-wa:
yeah, that's the plan (+ probably some other plugins in future), so i'll go with assing/condition

markus2330:
Thank you! Please ping me if you are ready!

tom-wa:
@markus2330 done

markus2330:
Thank you! The docu sometimes talks about `assign`, but I will fix it.

tom-wa:
sorry, i initially used `assign` but thought it would be ambiguous if other plugins would provide an assign operation too. Forgot to change it in the README.

PR #466, opened 2016-02-02
shorten config path in crypto plugin

petermax2:
as requested per mail today, I shortened the config paths for the crypto plugin


markus2330:
Thank you, nicely done!

Issue #459, opened 2016-01-22
Use Clang-Format to prettify code


sanssecours:
Basically finished, small issues:

- [ ] after `typedef enum` and `extern "C"` there is no newline. But from the docu `BreakBeforeBraces: Allman` already should be the right thing?

# Proposal

[Clang-Format][] is a tool that formats code according to a given set of rules. In my opinion it would make sense to require everyone to use this tool to format their code, before they add something to the repository. This way most of the source files follow a strictly defined style and no one has to care about how many spaces or tabs they need to add to certain parts of their code. For this to work we need to define a code style beforehand. According to [CODING.md](https://github.com/ElektraInitiative/libelektra/blob/master/doc/CODING.md) `.clang-format` should at least contain the following attributes:

```yaml
BreakBeforeBraces: Allman
IndentWidth: 8
ColumnLimit: 72
UseTab: Always
```

I am pretty sure that we need to specify some additional options, and maybe also change the value of `IndentWidth`, for a style that everyone agrees on. The documentation about the various options of `clang-format` is available [here](http://clang.llvm.org/docs/ClangFormatStyleOptions.html).

[Clang-Format]: http://clang.llvm.org/docs/ClangFormat.html

markus2330:
I actually tried to use clang-format, but the predefined styles are quite different to the current code base and creating a new style is a lot of effort.

petermax2:
I think a new style may be worth the effort, if the code formatting happens in an automated way. I still have to read my Elektra code twice before pushing it to find all spaces between function names and parentheses.

fberlakovich:
I agree. A tool that automatically formats code in a final way (e.g. after each commit) would be desirable. I use Eclipse which provides automatic code formatting. However, the configuration options are limited and some formatting still has to be done by hand.

markus2330:
Step 1: @KurtMi please write propose guideline
Step 2: Implement in clang-format

sanssecours:
I recently checked out the various options `clang-format` provides. Here is the current configuration:

```yaml
AccessModifierOffset: -8
AlignAfterOpenBracket: true
AlignConsecutiveAssignments: false
AlignEscapedNewlinesLeft: false
AlignOperands: false
AlignTrailingComments: true
AllowAllParametersOfDeclarationOnNextLine: false
AllowShortBlocksOnASingleLine: false
AllowShortCaseLabelsOnASingleLine: true
AllowShortFunctionsOnASingleLine: Empty
[...]

PR #467, opened 2016-02-03
New Coding Style

KurtMi:


doc/CODING.md
```diff
@@ -51,33 +51,56 @@ intent to add a new rule here.
...
-### C Guidelines ###
+- Limits
 
  * Functions should not exceed 100 lines.
  * Files should not exceed 1000 lines.
- * A line should not be longer than 72 characters.
- * Split up when those limits are reached.
- * Rationale: Readability with split windows.
- * The compiler shall not emit any warning (or error).
+ * A line should not be longer than 80 characters.
```
KurtMi: We should vote @markus2330 proposed 150 and @sanssecours 80

markus2330:
Thank you! I am ok with the change. Should we wait for release, next meeting or the vote; or should I merge it?

doc/CODING.md
markus2330: What about 120? Do you really have a workflow/screen/editor where 120 characters are problematic?

sanssecours: > What about 120? Do you really have a workflow/screen/editor where 120 characters are problematic?

I do not know if this questions is directed at me (too), but will answer it anyway. I usually use two windows side by side. With a text width of 80 characters, the editor window takes up a little less than half of my screen. I use the other half for stuff related to the coding process (documentation browser, Markdown preview, web browser, compiler output, second code window,…). A maximum of 120 character breaks this workflow, since I only have a small medium resolution screen (15 inch, 2880x1800).

Here are other reasons why a maximum of 80 characters per line might make sense:
- The default window width of all terminal emulators I know is 80 characters
- All but one (WebKit) of the default coding styles of Clang-Format use this value: LLVM, Google, Chromium, Mozilla
- [IBM punchcards](http://stackoverflow.com/questions/110928/is-there-a-valid-reason-for-enforcing-a-maximum-width-of-80-characters-in-a-code) use a default column width of 80 characters. Why would we use a different value if IBM considered this one ideal in 1928!

sanssecours:
> Should we wait for release, next meeting or the vote; or should I merge it?

The changes look fine to me. I think it would be okay to merge them right now. I do not think that it is that urgent though. Most of the people involved in the project know about the coding standard anyway.

By the way: Should I open a pull request for the `.clang_format` file posted [here](https://github.com/ElektraInitiative/libelektra/issues/459#issuecomment-174335686)?

doc/CODING.md
markus2330: Yes, the question was directed to everyone. So ok, if you cannot live with 120, we won't use it. I can live with 80, and agree that too long lines are hard to read.

Nevertheless, one last try: What about something around 100? The rationale behind 100 it is: If we agree that 80 characters in a line are ideal to read, but we often have 2 tabs (takes 16 spaces, 2 chars), something around 100 would be ideal.

Btw. Many setups, including e-mail editors are often configured with 72, so linewraps can also occur with 80 chars.

markus2330:
Yes, please open a PR, but not exactly as posted, but as discussed and as described in this PR. (e.g. PointerAlignment to Middle)

doc/CODING.md
sanssecours: > So ok, if you cannot live with 120, we won't use it.

Nah, I do not care that much 😀. I can totally live with 120 characters too. I still prefer 80 character (or less) though. Since you are the top contributor anyway, I think it would make sense that you have the last word on the decision. An other option would be – as KurtMi already suggested – to just let everyone vote on their favorite value.

markus2330: @KurtMi Yes, lets vote about it.

KurtMi:
I have added the poll result, can be merged.

PR #468, opened 2016-02-04
Add style file for `clang-format`

sanssecours:
Hi,

for now, the style file still uses a maximum column width of 80 characters. If there are any (other) problems with one of the attribute values, then please just comment below.

Thanks,
  René


markus2330:
Hi, thanks for calculating the vote result in #469 (actually I wanted to ask you first, but then I thought you are distracted from your main topic enough already).

Can you update the style file?

sanssecours:
> Can you update the style file?

I already changed the value of `ColumnLimit`, after I calculated the result of the poll.

Issue #464, opened 2016-02-01
crash with non-available plugin in global mountpoint with list plugin


markus2330:
Hi @tom-wa 

When I use
```
kdb set system/elektra/globalplugins
kdb set system/elektra/globalplugins/postcommit list
kdb set system/elektra/globalplugins/postcommit/user
kdb set system/elektra/globalplugins/postcommit/user/placements
kdb set system/elektra/globalplugins/postcommit/user/placements/set "presetstorage precommit postcommit"
kdb set system/elektra/globalplugins/postcommit/user/placements/get "pregetstorage postgetstorage"
kdb set system/elektra/globalplugins/postcommit/user/placements/error "prerollback postrollback"
kdb set system/elektra/globalplugins/postcommit/user/plugins
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0 counter
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/placements
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/placements/set "presetstorage"
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/placements/get "postgetstorage"
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/conflict/get "INFO"
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/conflict/get/range "INFO"
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/conflict/set "WARNING"
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/conflict/set/range "WARNING"
kdb set system/elektra/globalplugins/postcommit/user/plugins/#0/conflict/set/conflict "IGNORE"
kdb set system/elektra/globalplugins/postrollback list
kdb set system/elektra/globalplugins/precommit list
kdb set system/elektra/globalplugins/pregetstorage list
kdb set system/elektra/globalplugins/postgetstorage list
kdb set system/elektra/globalplugins/presetstorage list
kdb set system/elektra/globalplugins/prerollback list
```
but the plugin `counter` is not installed, Elektra crashes in the `list` plugin.

Backtrace:
```
#0  0x00007f69ef658147 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f69ef659528 in __GI_abort () at abort.c:89
[...]

PR #472, opened 2016-02-05
Bugfix 464

tom-wa:


markus2330:
Thank you! I think you have to write fix #464 (with `#`) so that github will close the bug automatically? (Only for the next time, I will close it manually this time if it won't work.)

PR #474, opened 2016-02-07
Keytometa append to samelevel fix

fberlakovich:
This PR fixes an issue with the samelevel setting of keytometa. In addition it adds the samelevel setting to the keytometa configuration shipped with the augeas plugin. The included testcase proves that the samelevel setting was not working as intended before the patch.


markus2330:
Thank you, looks fine!

Issue #470, opened 2016-02-04
Building of (LaTeX) Reference Manual Fails


sanssecours:
It looks like somewhere in the build process we add non UTF-8 characters to the LaTeX files produced by Doxygen. According to the output of `find . -name '*.tex' -exec grep -naxv '.*' '{}' /dev/null \;`:

```
./build/doc/latex/doc_GOALS_md.tex:139:{\itshape Trie} or prefix tree is an ordered tree data structure. In Elektra, it provides the information to decide in which backend a key resides. The algorithm, presented in \mbox{[}�����������
./build/doc/latex/md_doc_help_elektra-glossary.tex:32:Already available is the so-\/called X\+DG Base Directory Specification. Perhaps it will define file formats in the future, but currently only the paths are specified. To do so, it introduces some environment variables that help to find the actual configuration. Only if they are unset or empty, a built-\/in fallback should be used. If desired, elektrified applications behave in a X\+DG conforming way. (See the configuration x for \mbox{[}��������������
./build/doc/latex/md_doc_help_elektra-mounting.tex:9:\item everything else \mbox{[}���������
./build/doc/latex/md_doc_help_elektra-plugins-ordering.tex:33:\item \mbox{[}���������������������
./build/doc/latex/md_doc_help_kdb-ls.tex:25:If the user would also like to see the values of the keys below {\ttfamily path} then you should consider the \mbox{[}���������������
./build/doc/latex/md_doc_tutorials_compilation-variants.tex:34:\item \mbox{[}���������������������
./build/doc/latex/md_src_plugins_error_README.tex:107:Also explained \mbox{[}������
./build/doc/latex/md_src_plugins_README.tex:49:\item \mbox{[}�����������
``` 

`markdownlinkconverter` might produce these invalid characters.

Here are steps to reproduce this error:

1. `cmake .. -DBUILD_PDF=ON`
2. `make -j 8`
3. Enter `EOF` (<kbd>^</kbd> + <kbd>D</kbd>) after the CMake output shows a progress value of `[100%]`
4. Output:
```
[100%] Built target testtool_specreader
^D
! Emergency stop.
 ...

l.24 \item \mbox{[}�
                    ���������������������
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on refman.log.
make[2]: *** [doc/latex/refman.pdf] Error 1
make[1]: *** [doc/CMakeFiles/pdf.dir/all] Error 2
make: *** [all] Error 2
```

[...]

PR #476, opened 2016-02-08
add dependency for pdf target

KurtMi:
fix for #470?


markus2330:
Why do you think this does not resolve the bug?

KurtMi:
Because it looks like make executes the `${PDFLATEX_COMPILER}` command before all tex files are ready, so adding a dependency to latex which depends on `${OUTPUT_DIRECTORY}/latex/refman.tex` will not help.

sanssecours:
I Just applied the changes to my local copy of the repo. KurtMi is right: Even with the new dependency, the bug reported in issue #470 does still exist.

PR #478, opened 2016-02-09
fix for #470

KurtMi:


markus2330:
Thank you!

Btw. The CLI-Commandos github shows do not work with zsh when there is `#` the branch name.

KurtMi:
> Btw. The CLI-Commandos github shows do not work with zsh when there is # in the branch name.
Good to know!

PR #481, opened 2016-02-09
doku: change the welcome message in the top of README

tryge:
I found that the first thing I read about Elektra wasn't helping me much to understand what it can do. The current message uses a lot of buzzwords, that can mean about anything.

I tried to concretely describe a few things Elektra can help with, use repetition so someone might actually remember something after having read this, and use a language that invites the reader to continue reading (similar to advertising). (I don't claim any skills that I can actually do all that, but I tried...)

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,30 @@ parameters in a global, hierarchical key database._
 
 <img src="https://cdn.rawgit.com/ElektraInitiative/libelektra/master/doc/images/circle.svg" alt="Elektra" width="50" />
 
-The core is a small library implemented in C. The plugin-based framework fulfills many
-configuration-related tasks to avoid any unnecessary code duplication
-across applications while it still allows the core to stay without any
-external dependency. Elektra abstracts from cross-platform-related issues
-with a consistent API, and allows applications to be aware of other
-applications' configurations, leveraging easy application integration.
+Elektra provides a mature and easily comprehensible API for applications to access 
+and update configuration data with a focus on modularity and application integration.
+**Modularity** effectively eliminates code duplication across applications regarding 
+configuration tasks while easy **application integration** is obtained by allowing 
+applications to be aware of each others configuration.
+
+To highlight a few concrete things about Elektra, configuration data can come from any
+data source, but usually comes from configuration files that are [_mounted_](doc/help/elektra-mounting.md) into Elektra 
+similar to mounting a file system. As Elektra is a plugin based framework, there are a 
+lot of _storage plugins_ that support various configuration formats like ini, json, xml, 
+etc. However, there's a lot more to discover like executing scripts (`python`, `lua` or 
+`shell`) when a configuration value changes, or, enhanced validation plugins that won't 
+allow corrupted configuration to reach your application.
+
+As an application developer you get instant access to various configuration formats and the ability
+to fallback to a default configuration without having to deal with this on your own. As an administrator
+you can choose your favourite configuration format and _mount_ this configuration for the application.
+This features easy application integration as any application using Elektra can access any _mounted_ 
+configuration. You can even _mount_ `/etc` files such as `hosts` or `fstab` if needed, no need to 
+configure the same data twice in different files.
+
+In case you're worried about linking to such a powerful library. The core is a small library 
+implemented in C, works cross-platform, and does not need any external dependencies. There are 
+bindings for other languages in case C is too low-level for you.
 
 [Why should I use Elektra?](#goals)
 


markus2330:
Yes, I agree that a longer abstract is needed to get more than a very vague thought what it could be. But I am not sure that its a good idea to start introducing new terminology in the first paragraph.

Maybe there is a way to keep the first paragraph much more as it currently is (high-level and abstract) and explain in subsequent paragraphs more concrete with examples?

tryge:
The new terminology is by design, they are anchors that keep readers on the site and also give them concrete keywords to look for. They raise a question and usually readers seek to answer the questions (though they won't spend a lot of time on it).

It always depends on the message you want to send. I want to attract people and keep them reading.

What do you want to get across with the abstract?

tryge:
Btw. I'm not clinging to my changes, so we can just close the PR if you prefer the existing abstract.

markus2330:
The main purpose of the abstract is to be copied and pasted to places where a short description of Elektra is required. What should come across is that Elektra aims at improving modularity (reduces code duplication across applications) and integration (usage of configuration between applications). The other sentences are to attract specific important readers.

And I see your point that this aim might be contradicting to what we want in the beginning of the README.md, i.e. the main entry point of the "homepage".  I fully agree that it should change. Maybe two different paragraphs are needed to fulfil both goals?

README.md
```diff
@@ -5,12 +5,24 @@ parameters in a global, hierarchical key database._
...
 <img src="https://cdn.rawgit.com/ElektraInitiative/libelektra/master/doc/images/circle.svg" alt="Elektra" width="50" />
 
-The core is a small library implemented in C. The plugin-based framework fulfills many
-configuration-related tasks to avoid any unnecessary code duplication
-across applications while it still allows the core to stay without any
-external dependency. Elektra abstracts from cross-platform-related issues
-with a consistent API, and allows applications to be aware of other
-applications' configurations, leveraging easy application integration.
+Elektra provides a simple API for applications to access and update configuration data. The
+configuration data comes from configuration files that are _mounted_ into Elektra similar to
```
markus2330: This is a good example and important for getting an idea but not necessarily true. While Elektra favours the usage of configuration files (because of its resolvers and its current set of storage plugins), it actually also supports to integrate any other data sources as well.

README.md
```diff
@@ -5,12 +5,24 @@ parameters in a global, hierarchical key database._
...
-external dependency. Elektra abstracts from cross-platform-related issues
-with a consistent API, and allows applications to be aware of other
-applications' configurations, leveraging easy application integration.
+Elektra provides a simple API for applications to access and update configuration data. The
+configuration data comes from configuration files that are _mounted_ into Elektra similar to 
+mounting a file system. As Elektra is a plugin based framework, there are a lot of _storage 
+plugins_ that support various configuration formats like ini, json, xml, etc. However, 
+there's a lot more to discover like executing python or lua scripts when a configuration
+value changes or enhanced validation plugins that won't allow corrupted configuration to 
+reach your application.
```
markus2330: I like this part.

README.md
```diff
@@ -5,12 +5,24 @@ parameters in a global, hierarchical key database._
...
-configuration-related tasks to avoid any unnecessary code duplication
-across applications while it still allows the core to stay without any
-external dependency. Elektra abstracts from cross-platform-related issues
-with a consistent API, and allows applications to be aware of other
-applications' configurations, leveraging easy application integration.
+Elektra provides a simple API for applications to access and update configuration data. The
+configuration data comes from configuration files that are _mounted_ into Elektra similar to 
+mounting a file system. As Elektra is a plugin based framework, there are a lot of _storage 
+plugins_ that support various configuration formats like ini, json, xml, etc. However, 
+there's a lot more to discover like executing python or lua scripts when a configuration
```
markus2330: What about: "like executing scripts, e.g. written in python, lua or shell,"

README.md
```diff
@@ -5,12 +5,24 @@ parameters in a global, hierarchical key database._
...
+value changes or enhanced validation plugins that won't allow corrupted configuration to 
+reach your application.
+
+As an application developer you get instant access to various configuration formats and the ability
+to fallback to a default configuration without having to deal with this on your own. As an administrator
+you can choose your favourite configuration format and _mount_ this configuration for the application.
+This features easy application integration as any application using Elektra can access any _mounted_ 
+configuration. You can even _mount_ your operating systems `hosts` or `fstab` configuration if needed,
+no need to configure the same data twice in different files.
```
markus2330: I also like this part, for `hosts` and `fstab` same concerns as above. Maybe: "`/etc` files such as `hosts` and `fstab`"?

README.md
```diff
@@ -5,12 +5,24 @@ parameters in a global, hierarchical key database._
...
+
+In case you're worried about linking to such a powerful library. The core is a small library 
+implemented in C, works cross-platform, and does not need any external dependencies. There are 
+bindings for other languages in case C is too low-level for you.
```
markus2330: Also a good sentence.

tryge:
ok, I'll target both goals.

markus2330:
Btw. I think it will be very helpful for the reader that you give a hint when you start with the concrete part, like you did in the message of the pull request.

In particular I refer to "concretely describe a few things Elektra can help with". This would explain a lot and would solve the issue with the use of "configuration files" (Its a thing Elektra can help with, among others).

README.md
tryge: How about: "like executing scripts (python, lua or shell)"

README.md
tryge: yes.

README.md
tryge: I'll highlight that this is only one option.

markus2330: Its not the only option, but a focus of Elektra.

README.md
markus2330: yes, its shorter! (some people do not read text in parenthesis, though)

tryge:
New version.

README.md
```diff
@@ -5,12 +5,30 @@ parameters in a global, hierarchical key database._
...
-configuration-related tasks to avoid any unnecessary code duplication
-across applications while it still allows the core to stay without any
-external dependency. Elektra abstracts from cross-platform-related issues
-with a consistent API, and allows applications to be aware of other
-applications' configurations, leveraging easy application integration.
+Elektra provides a simple API for applications to access and update configuration 
+data with a focus on modularity and application integration. Modularity, in order 
+to effectively eliminate code duplication throughout applications regarding 
+configuration, and application integration, so applications can leverage the 
+configuration of other applications leading to well integrated systems.
```
markus2330: The last sentence is too long and is trying to say too much.

README.md
```diff
@@ -5,12 +5,30 @@ parameters in a global, hierarchical key database._
 
 <img src="https://cdn.rawgit.com/ElektraInitiative/libelektra/master/doc/images/circle.svg" alt="Elektra" width="50" />
 
-The core is a small library implemented in C. The plugin-based framework fulfills many
-configuration-related tasks to avoid any unnecessary code duplication
-across applications while it still allows the core to stay without any
-external dependency. Elektra abstracts from cross-platform-related issues
-with a consistent API, and allows applications to be aware of other
-applications' configurations, leveraging easy application integration.
+Elektra provides a simple API for applications to access and update configuration
```
markus2330: "simple" is a bit vague and also not really true (think about `getenv` API), maybe "mature and generic"? The API is more a building block (or data structure) for other high-level APIs. That the access is *global* is an important property not mentioned.

markus2330: Ok, global is in the very first sentence. (Above the img)

markus2330:
Thank you, I think the main part is now very nice, only the first paragraph needs some tuning.

README.md
tryge: I guess I'm a bit biased against the word generic. How about "mature and easily comprehensible", but if you still prefer generic, I'll adopt it.

tryge:
I adapted the first paragraph. While the last sentence is not much shorter, I think it's easier to read than before so the length might not be a problem anymore - it's also closer to the original text. For now I used "easily comprehensible", but I can still change that.

markus2330:
Thank you! I will do a bit more fine-tuning. Further PRs for the main README.md are highly appreciated..

PR #486, opened 2016-02-10
doku: add ronn as dependency for building the documentation

tryge:
Added a link to ronn's install instructions so it can be easily installed.


markus2330:
Thank you!

PR #490, opened 2016-02-10
fix broken links

omnidan:
* doc/help/elektra-introduction.md:66 - not sure if `backends` link means `elektra-backends.md`
 * doc/tutorials/compilation-variants.md:5 - not sure what `contract.md` should be

All other broken links should be fixed now.

BTW: There seems to be an off-by-one error (line number is always +1 of what the error message shows)


KurtMi:
> BTW: There seems to be an off-by-one error (line number is always +1 of what the error message shows)

I had cases where it was also -1, so I was not sure if it is always +1.

markus2330:
Thank you! Do you have a schedule for the http links?

Issue #484, opened 2016-02-10
merging: "binary" metakey is removed


markus2330:
The 3-way merge seems to use `setString` internally which removes the `binary` metakey and also values after null bytes.

Especially, the `binary` metakey is important for INI (they represent sections, ask @tom-wa for details).

Can you change the `setString` to `if (k.isString()) k.setString(..) else k.setBinary()`?

markus2330:
Thank you, should be fixed now.

PR #493, opened 2016-02-10
merge strategies now respect binary data #484

fberlakovich:
This PR should fix the problem discussed in #484 along with unit tests that verify the changed behaviour in the strategies and an extended shell merge test that verifies the behaviour on the commandline. However, there is no specific INI test for now. This is because 1. I have no example of a failed merge 2. the problem was not really INI specifc and easy to understand.


markus2330:
Thank you! Are you sure that the nullptr case is considered? (keyValue() might return null) Which one is the testcase for it?

markus2330:
@tom-wa The problem is now fixed, sections in INI files are preserved correctly (in kdb editor)

@fberlakovich thanks for the fast fix, also the testcases were useful to copy from

PR #496, opened 2016-02-11
fix linking for OS X #495

### LINKED ISSUE #495: Building on OS X Fails
Since commit cc2b57bc building on OS X fails. Here is the error output:

```
[ 24%] Built target elektra-ease
[ 24%] Linking CXX shared library ../../../../lib/libelektratools.dylib
  "_elektraPluginClose", referenced from:
      kdb::tools::Plugin::uninit() in plugin.cpp.o
  "_elektraPluginOpen", referenced from:
      kdb::tools::Plugin::Plugin(kdb::tools::PluginSpec const&, kdb::KeySet&) in plugin.cpp.o
make[2]: *** [lib/libelektratools.dylib] Error 1
make[1]: *** [src/libs/tools/src/CMakeFiles/elektratools.dir/all] Error 2
make: *** [all] Error 2
```

If you need any additional information, then please just comment below.

mpranj:
this fixes #495 for me
@markus2330 is this an acceptable fix? I'm not sure if you intended not to link against elektra-kdb

### DIFF
--- a/src/libs/tools/src/CMakeLists.txt
+++ b/src/libs/tools/src/CMakeLists.txt
@@ -18,7 +18,7 @@ if (BUILD_SHARED)
 		"ELEKTRA_SHARED"
 		)
 
-	target_link_libraries (elektratools elektra-core elektra-plugin elektra-ease)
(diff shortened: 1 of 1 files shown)


markus2330:
Yes, thank you for fixing it!

PR #499, opened 2016-02-12
Fix Minor Mistakes in Documentation

sanssecours:
Hi,

the following commits contain minor updates to the documentation. If there are any questions about the changes, please just comment below. Thanks.

### DIFF
--- a/doc/help/elektra-algorithm.md
+++ b/doc/help/elektra-algorithm.md
@@ -54,12 +54,12 @@ the backend enclosing the plugin is not mounted at all.
 
 ### Removing Keys
 
-In Elektra version $0.6$, removing keys was an explicit request.  Only a
+In Elektra version 0.6, removing keys was an explicit request.  Only a
 single `Key` object could be removed from the database.  For configuration
 files this method is inapplicable.  For `filesys`, however, it was easy
 to implement.
 
-In Elektra version $0.7$, the behaviour changed.  Removing keys was
+In Elektra version 0.7, the behaviour changed.  Removing keys was
 integrated into `kdbSet()`.  The user tagged keys that should be removed.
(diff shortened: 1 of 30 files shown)


doc/help/elektra-algorithm.md
```diff
@@ -138,7 +138,7 @@ to determine if they are responsible for a key or not.	Consequently, it
 can happen that more than one backend delivers a key with the same name.
 
 `kdbGet()` ensures that a key is uniquely identified by its name.
-Elektra's core will pop keys that are outside of the backend's
+Elektra's core will drop keys that are outside of the backend's
```
markus2330: Its better to keep `pop` and put a link to the glossary, because `pop`  is also the term used in the API.

doc/help/elektra-algorithm.md
```diff
@@ -222,8 +222,8 @@ needs to store the number of received keys of each backend.
 - Additionally, for every key it is checked if it belongs to this
 backend.  This makes sure that every key comes from a single source
 only as designated by the `Trie`.  In this process, all duplicated and
-overlapping keys will be popped in favour of the responsible backend as
-described below in responsibility.
+overlapping keys will be dropped in favour of the responsible backend as
```
markus2330: see above; maybe the sentence is better if put in active

markus2330:
Thank you, these are really great fixes!

doc/help/elektra-algorithm.md
sanssecours: Should be fixed now.

BTW: This is the first time I heard that there is a glossary. It might be a good idea to put the glossary somewhere after the table of contents. As it stands today it is at page 273 of the reference manual. That is not exactly a place where I would search for a glossary.

markus2330:
Thanks, the documentation keeps getting better!

PR #504, opened 2016-02-14
Fix building of certain plugins on OS X

sanssecours:
There might be another (and better) option to fix this problem. However, this one at least works on my machine. For reference, the command

```fish
cmake ..  -DPLUGINS='dump;resolver;sync;error;spec'; and make
```

prints the following error message before this commit on OS X:

```
  "_elektraPluginExport", referenced from:
      _elektraPluginSymbol in spec.c.o
  "_elektraPluginGetConfig", referenced from:
      _elektraSpecGet in spec.c.o
      _elektraSpecSet in spec.c.o
make[2]: *** [lib/libelektra-spec.so] Error 1
make[1]: *** [src/plugins/spec/CMakeFiles/elektra-spec.dir/all] Error 2
make: *** [all] Error 2
```


markus2330:
Thank you, this is an important fix! (Also for non-OS X)

PR #511, opened 2016-02-14
Fix #509

### LINKED ISSUE #509: building plugin doc fails on Mac OS X
When running the command `cmake .. -DPLUGINS='dump;resolver;sync;error;doc'`, `make` fails with the following message:

```
[ 19%] Linking C shared module ../../../lib/libelektra-doc.so
  "_kdbSet", referenced from:
      _elektraUsercodeUselessSymbol in doc.c.o
make[2]: *** [lib/libelektra-doc.so] Error 1
make[1]: *** [src/plugins/doc/CMakeFiles/elektra-doc.dir/all] Error 2
make: *** [all] Error 2
```

tryge:
As @markus2330 suggested, his fix worked (and I like things that actually compile and link :-) ).

### DIFF
--- a/src/plugins/doc/CMakeLists.txt
+++ b/src/plugins/doc/CMakeLists.txt
@@ -3,4 +3,7 @@ include(LibAddMacros)
 add_plugin(doc
 	SOURCES
 		doc.h doc.c
-	)
+	LINK_ELEKTRA
+		elektra-kdb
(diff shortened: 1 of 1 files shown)


markus2330:
Thanks for checking it! I prefer when everything compiles+links, too.

PR #512, opened 2016-02-14
add library split benchmark details + script

mpranj:
@markus2330 adding a few details, ping me if you need more


markus2330:
Thank you, its much better now!

PR #513, opened 2016-02-14
fixes #510

### LINKED ISSUE #510: building plugin list fails on Mac OS X
When running the command `cmake .. -DPLUGINS='dump;resolver;sync;error;list'`, `make` fails with the following message:

```
[ 18%] Linking C shared module ../../../lib/libelektra-list.so
  "_elektraPluginClose", referenced from:
      _elektraListClose in list.c.o
      _runPlugins in list.c.o
  "_elektraPluginOpen", referenced from:
      _runPlugins in list.c.o
make[2]: *** [lib/libelektra-list.so] Error 1
make[1]: *** [src/plugins/list/CMakeFiles/elektra-list.dir/all] Error 2
make: *** [all] Error 2
```

mpranj:
fixes #510, however unit tests for module `list` fail on my machine (OS X):

```
src/plugins/list/testmod_list.c:43: error in doTest: kdbget failed
src/plugins/list/testmod_list.c:45: error in doTest: key not found
src/plugins/list/testmod_list.c:47: error in doTest: meta key not found
src/plugins/list/testmod_list.c:48: error in doTest: kdbset failed

testmod_list RESULTS: 12 test(s) done. 4 error(s).
```

### DIFF
--- a/src/plugins/list/CMakeLists.txt
+++ b/src/plugins/list/CMakeLists.txt
@@ -4,6 +4,8 @@ add_plugin(list
 	SOURCES
 		list.h
 		list.c
-	)
+	LINK_ELEKTRA
+		elektra-kdb
+)
 
 add_plugintest(list)


markus2330:
Thank you for fixing the compilation. Can you create a new issue with these errors? Can you check with the debugger what goes wrong? (Maybe the rename or keytometa plugin is missing? They are used in the testcase)
Does global mounting works? (it uses the list plugin)

mpranj:
I will open a new issue. It seems that even more plugins fail. Some are just OS X specifics I guess, like testscr_check_real_world:
```
For dir w /a/b we got /private/tmp//a/b
error: resolving of dir/test/script did not yield /tmp//a/b but /private/tmp//a/b
```
which is actually fine for OS X.

markus2330:
Yes, for `/private/tmp` you could add an `if` in the testcases in order to expect the results correct for Mac OS X. But for this release lets concentrate on whats broken.

PR #514, opened 2016-02-15
Allow range without "-"

KurtMi:
```
kdb set spec/test/#
kdb setmeta spec/test/# array 1
kdb set /test ""
kdb set /test/#0 ""
kdb set /test/#1 ""
```
This setup caused a segfault, by fixing I added the possibility to set the range without "-".
Removed also some ELEKTRA_UNUSED


markus2330:
Thank you!

PR #518, opened 2016-02-15
fixes #516

### LINKED ISSUE #516: list plugin (test) fails on OS X
steps to reproduce:
```
cd build && cmake -DPLUGINS='dump;resolver;sync;error;list' .. && make
sudo make run all
```

Then `testmod_list` fails with:
```
src/plugins/list/testmod_list.c:43: error in doTest: kdbget failed
src/plugins/list/testmod_list.c:45: error in doTest: key not found
src/plugins/list/testmod_list.c:47: error in doTest: meta key not found
src/plugins/list/testmod_list.c:48: error in doTest: kdbset failed

testmod_list RESULTS: 12 test(s) done. 4 error(s).
```

All other tests are passing.

mpranj:
fixes #516 but doesn't look extremely nice

I'm thinking maybe we should make a macro for describing the dependencies of a plugin. I haven't seen dependency checking in any of the other plugins. Any thoughts?

### DIFF
--- a/src/plugins/list/CMakeLists.txt
+++ b/src/plugins/list/CMakeLists.txt
@@ -11,7 +11,7 @@ add_plugin(list
 list(FIND PLUGINS "rename" RENAME_INDEX)
 list(FIND PLUGINS "keytometa" KEYTOMETA_INDEX)
 
-if(BUILD_TESTING AND RENAME_INDEX AND KEYTOMETA_INDEX GREATER -1)
+if(BUILD_TESTING AND RENAME_INDEX GREATER -1 AND KEYTOMETA_INDEX GREATER -1)
 	add_plugintest(list)
 else()
 	message (WARNING "rename and keytometa plugins are required for testing, test deactivated")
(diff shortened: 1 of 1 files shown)


markus2330:
Hello, you can do this only for the test, but not for the plugin itself. The plugin works perfectly without keytometa and rename.

Manuel made similar things in lua/python plugins, where the tests depent on something, but not the plugin themselves.

mpranj:
Will fix.

markus2330:
Thanks, a warning if the test is deactivated would be nice. As said, keep it as close as possible to the existing two places.

mpranj:
Sorry, I don't know what you meant by

> keep it as close as possible to the existing two places.

markus2330:
I was talking about the lua/python plugins.
They have:
```
if (BUILD_TESTING AND FINDEX GREATER -1)
add_plugintest (...
else ()
message (WARNING "... are required for testing, test deactivated")
endif ()
```

mpranj:
@markus2330 this is finished if you have no further objections

markus2330:
thank you!

PR #532, opened 2016-02-18
update status of crypto-plugin

petermax2:
I added the "discouraged" flag to the status as for now "crypto" might suggest a false sense of security. As long as there is no key derivation interface there is no way of keeping the secret stuff (cryptographic keys) secret.


markus2330:
I think you worry too much, with `experimental unfinished memleak` nobody will think it is secure ;)

PR #537, opened 2016-02-20
Benchmark script and performance improvement for augeas plugin

fberlakovich:
The PR contains a small script (should be sh compatible :)) for benchmarking the augeas plugin performance with a provided hosts file. While analysing the plugin with callgrind I realised that an easy performance improvement can be made which is also included.

### DIFF
--- a/scripts/benchmark_augeas_hosts.sh
+++ b/scripts/benchmark_augeas_hosts.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+if [ -z "$KDB" ]; then
+    KDB=kdb
+fi
+
+if [ -z "$1" ]; then
+    echo "Usage: $0 <hosts file name>"
+    exit 1
+fi
+
+
+measure_time () {
+    { time -f "%e" $1 > /dev/null; } 2>&1
+}
+
+HOSTSFILE=$(echo "$(cd "$(dirname "$1")" && pwd)/$(basename "$1")")
+AUGEASPATH="system/benchmarks/$$/augeasplugin-hostsfile"
+KEYTOMETAPATH="system/benchmarks/$$/augeaskeytometa-hostsfile"
+HOSTSPATH="system/benchmarks/$$/hostsplugin-hostsfile"
+
+$KDB mount $HOSTSFILE $AUGEASPATH augeas lens=Hosts.lns
+$KDB mount $HOSTSFILE $HOSTSPATH hosts
+$KDB mount $HOSTSFILE $KEYTOMETAPATH augeas lens=Hosts.lns glob keytometa
+
+for run in $(seq 0 11); do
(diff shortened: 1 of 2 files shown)


scripts/benchmark_augeas_hosts.sh
```diff
@@ -0,0 +1,34 @@
...
+
+for run in $(seq 0 11); do
+	echo "RUN: #$run"
+	augeaswalltime=$(measure_time "$KDB ls $AUGEASPATH")
+	hostswalltime=$(measure_time "$KDB ls $HOSTSPATH")
+	keytometawalltime=$(measure_time "$KDB ls $KEYTOMETAPATH")
+	echo "augeas: $augeaswalltime; hosts: $hostswalltime; augeas with keytometa: $keytometawalltime"
+done
+
+for x in ` kdb mount | grep system/benchmarks/$$ | awk '{print $3}'`; do sudo kdb umount $x; done
```
markus2330: Its strange that mounting is done without sudo, but umount with. Also $KDB is not used and the whitespaces could be improved.

src/plugins/augeas/augeas.c
```diff
@@ -198,7 +198,7 @@ static int removeOrphan (augeas * handle, const char * treePath, void * data)
 static int foreachAugeasNode (augeas * handle, const char * treePath, ForeachAugNodeClb callback, void * callbackData)
 {
 	char * matchPath;
-	asprintf (&matchPath, "%s/*", treePath);
+	asprintf (&matchPath, "%s//*", treePath);
```
markus2330: Thanks, but this should be a different commit.

src/plugins/augeas/augeas.c
```diff
@@ -216,9 +216,6 @@ static int foreachAugeasNode (augeas * handle, const char * treePath, ForeachAug
 
 		result = (*callback) (handle, curPath, callbackData);
 
-		/* handle the subtree */
-		result = foreachAugeasNode (handle, curPath, callback, callbackData);
```
markus2330: Are you sure that this is right?

markus2330:
Thank you, I added some comments!

src/plugins/augeas/augeas.c
fberlakovich: At least the tests say so :). This can be done because ```path//*``` is used instead of ```path/*````. While ```path/*``` returns only the nodes below path, ```path//*``` returns all subtrees below path. See https://github.com/hercules-team/augeas/wiki/Path-expressions for more details.

markus2330: Ok, nice, thank you. Is this also faster or otherwise unrelated to the benchmark?

fberlakovich: It is about 20% faster than the old solution

PR #538, opened 2016-02-21
fix typo in type/README.md

petermax2:


markus2330:
Thank you, very good catch! I thought Ian copied this 1:1, but it seems like this sentence is original ;)

PR #542, opened 2016-02-25
dpkg plugin

tom-wa:
mounts /var/lib/dpkg/{available,status} (read only)

### DIFF
--- a/cmake/ElektraCache.cmake
+++ b/cmake/ElektraCache.cmake
@@ -63,6 +63,7 @@ set (PLUGINS_LIST_NODEP
 	wresolver
 	ini
 	list
+	dpkg
 	logchange
 	iterate
 	spec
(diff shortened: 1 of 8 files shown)


markus2330:
Thank you! Looks very nice and useful!

See buildserver:
There is a small syntactic problem in `cmake/ElektraCache.cmake` (should be tab). Can you reformat the whole source (clang-format)? And we should add a status `unfinished` and do not test plugins (with the tests that now fail for `dpkg`) that have this status.

Can you do this little fixes?

tom-wa:
forgot to turn off the storage tests for the plugin. 
i think we should split infos/provides = storage into storage and limited storage or something for plugins that aren't intended to be used as universal storage plugins ? 
there are tons of different config file formats, and we probably need nearly as much plugins for reading(and maybe writing to) them, but there'd be no need for them to work as universal storage plugins and the not_rw_storage list in include_common.sh.in would become quite long.

markus2330:
Yes, the idea about "storage/limited" is good. It can be used in cases where the problem is not only temporarily, but it is simply not possible to fully roundtrip any keyset. When the implementation is unfinished, however, other flags in `status` should be used. In this case we seem to have both: unfinished (write path) and "storage/limited"

Thank you! Hope to see more soon!

Issue #452, opened 2016-01-13
Issues compiling elektra on OpenBSD 5.8


tom-wa:
* OpenBSD doesn't use/know EBADMSG
```
/home/thomas/libelektra/src/plugins/resolver/resolver.c: In function 'elektraAddErrnoText':
/home/thomas/libelektra/src/plugins/resolver/resolver.c:278:20: error: 'EBADMSG' undeclared (first use in this function)
  else if (errno == EBADMSG)
                    ^
/home/thomas/libelektra/src/plugins/resolver/resolver.c:577:12: error: 'EBADMSG' undeclared (first use in this function)
    errno = EBADMSG;
            ^
```

* OpenBSD doesn't have librt
should be 
```
if (UNIX AND NOT APPLE AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD")
	set(CMAKE_REALTIME_LIBS_INIT rt)
endif()
```
in cmake/ElektraCompiling.cmake

markus2330:
Thanks for reporting! These are really easy fixes then! (EBADMSG was chosen arbitrary, we can take any other value not issued by the syscalls in questions, too)

markus2330:
@petermax2 Do you have OpenBSD installed?

petermax2:
@markus2330 I have an OpenBSD VM somewhere

petermax2:
Found another issue on OpenBSD (tried to compile a67c05505552c57861125c3ed4bf26cda7553fdc )

	[  1%] Building CXX object src/error/CMakeFiles/elektra-export-errors.dir/exporterrors.cpp.o
	cc1plus: warning: -Wuninitialized is not supported without -O
	In file included from /home/pnirschl/elektra/src/error/exporterrors.cpp:9:
	/home/pnirschl/elektra/src/error/parser.hpp:29: error: expected ';' before 'override'
	/home/pnirschl/elektra/src/error/parser.hpp:33: error: expected `;' before '}' token
	/home/pnirschl/elektra/src/error/parser.hpp: In constructor 'parse_error::parse_error(std::string, int)':
	/home/pnirschl/elektra/src/error/parser.hpp:22: error: 'move' is not a member of 'std'
	/home/pnirschl/elektra/src/error/exporterrors.cpp: In function 'int main(int, char**)':
	/home/pnirschl/elektra/src/error/exporterrors.cpp:281: error: 'to_string' was not declared in this scope
[...]

PR #543, opened 2016-02-25
Introduce FindLibRt module

petermax2:
See #452 for full discussion.

Verified build and unit tests for fb64dc44a7a47def349968bcdce4cf684b83e5b1 on:

- OpenBSD 5.8 with eg++ and egcc 4.3.9
- Apple Mac OS X 10.11.3 (15D21) with Apple LLVM version 7.0.2 (clang-700.1.81)


src/tools/race/CMakeLists.txt
```diff
@@ -1,14 +1,20 @@
 find_package(Threads)
+find_package(LibRt)
 
 if (APPLE)
 	remove_tool(race "race tool does not compile under MAC OS X")
 else()
+	if (NOT LIBRT_FOUND)
+		remove_tool(race "race tool requires librt support")
```
markus2330: Is it possible that librt is not found but nevertheless available? Why doing something different than for the resolver here?

markus2330:
Thank you! Only one little question.

src/tools/race/CMakeLists.txt
petermax2: > Is it possible that librt is not found but nevertheless available?

I hope not.

>  Why doing something different than for the resolver here?

My assumption was that the resolver compiles without librt. Is that correct or do we remove the resolver if no librt is found?

The race tool seems to require librt, so removing it in case of a missing librt was the only way.

markus2330: Some variants of resolver might work without it, but there would be no use to link against it, if it was not needed anyway. What was the behaviour on OpenBSD?

petermax2: > I am not so sure: librt functionality might be present in some libc. Maybe thats even the case in OpenBSD.

How do we find out if that is the case? Do we use some `try_compile` magic again?

> No need to have assumptions about if something links, thats really easy to check.

The assumption was not about whether it links or not but rather if it works without librt.

petermax2: > What was the behaviour on OpenBSD? 

The unit tests were successful on OpenBSD 5.8 without linking the resolver against librt.

markus2330: And they failed for the `race` tool?

src/plugins/resolver/CMakeLists.txt
```diff
@@ -3,22 +3,41 @@ include(LibAddPlugin)
...
-		ELEKTRA_PLUGIN_NAME=\"resolver\"
-		ELEKTRA_LOCK_FILE
-		ELEKTRA_LOCK_MUTEX
+if (LIBRT_FOUND)
+	add_plugin(resolver
+		SOURCES
+			${SOURCES}
+		LINK_LIBRARIES
+			${CMAKE_THREAD_LIBS_INIT}
+			${LIBRT_LIBRARIES}
```
markus2330: This could be written which much less duplication by the using an `if` outside the `add_plugin`.

src/tools/race/CMakeLists.txt
petermax2: Are there unit tests for the race tool? There is only `race.c` and `CMakeLists.txt` in `src/tools/race/`.

src/plugins/resolver/CMakeLists.txt
petermax2: That was my original idea. I will change it back.

src/tools/race/CMakeLists.txt
petermax2: You are right about OpenBSD providing some of librt features. Compiling and running `race` in OpenBSD without librt does not cause problems.

On Linux these symbols are provided by librt:

    /elektra/src/tools/race/race.c:102: undefined reference to `shm_unlink'
    /elektra/src/tools/race/race.c:104: undefined reference to `shm_open'
    /elektra/src/tools/race/race.c:120: undefined reference to `shm_unlink'
    /elektra/src/tools/race/race.c:178: undefined reference to `shm_unlink'

petermax2: It is still not compiling under Mac OS X. `pthread_barrier_t` and all related functions to it are not to be found.

markus2330: Thank you for checking! Can you update the PR? Please also document why `race` is excluded for APPLE.

Tests that use the race tool are in `tests/shell/check_race.sh` (currently disabled by default). Its easier to run the tool manually though.

petermax2:
@markus2330 did I break something? The link to "Details" refers to [http://build.libelektra.org:8080/job/elektra-ini-mergerequests/45/](http://build.libelektra.org:8080/job/elektra-ini-mergerequests/45/) and seems to be broken. But the [merge request build](http://build.libelektra.org:8080/job/elektra-mergerequests/858/) was successful.

petermax2:
jenkins build please

petermax2:
`elektra-ini-mergerequests` never succeeded. it wasn't me :blush:

markus2330:
Thank you for reporting! I now changed

jenkins build please

to not build the ini job.
@tom-wa Instead we now have `jenkins build ini please` if you want to test ini as default.

cmake/Modules/FindLibRt.cmake
```diff
@@ -0,0 +1,31 @@
...
+if (NOT LIBRT_FOUND)
+	try_compile (HAS_LIBRT_4SURE
+		"${CMAKE_BINARY_DIR}"
+		"${PROJECT_SOURCE_DIR}/tests/librt_test.c"
+	)
+
+	if (HAS_LIBRT_4SURE)
+		set (LIBRT_FOUND on)
+		set (LIBRT_LIBRARIES "")
+		message (STATUS "using system's built-in librt functions")
```
markus2330: That is a nice solution, thank you!

petermax2:
I am ready to merge if there is nothing left to improve.

markus2330:
Thank you! Btw. there is always something to improve, but we should prefer to spent time on more important issues.

PR #547, opened 2016-02-29
Documentation Updates

sanssecours:
This pull request contains minor fixes for the documentation. If there are problems with any of the commits, then please let me know.


doc/decisions/relative.md
```diff
@@ -4,8 +4,8 @@
 
 There is a different behaviour of various plugins whether their name is
 absolute or relative, including:
-1.) mounting the same file somewhere else does not work
-2.) importing somewhere else (other than from where it was exported) does not work
+1. mounting the same file somewhere else does not work
```
markus2330: Its not a new issue you introduce, but now that I see it: I think according to markdown there must be a newline before enumerations.

doc/help/elektra-algorithm.md
```diff
@@ -304,7 +304,7 @@ The user passes the configuration using the `KeySet` `returned`.  The key
 set will not be changed by `kdbSet()`.	The `parentKey` provides a way
 to limit which part of the configuration is written out.  For example,
 the `parentKey` `user/sw/org/app/#0/current` will induce `kdbSet()` to
-only modify the key databases below `user/sw/apps/myapp` even
+only modify the key databases below `user/sw/apps/app` even
```
markus2330: thanks, nice catch! Maybe "org" instead of "apps" would be better? (again the problem already existed earlier)

markus2330:
You are a *really* careful reader! I love your fixes! I am looking forward reading original text from you!

markus2330:
Btw. please do not create branches in the main repo, it easily leads to orphans. Instead prefer to use your own repo, there you can create as many branches as you would like to have.

doc/decisions/relative.md
sanssecours: > Its not a new issue you introduce, but now that I see it: I think according to markdown there must be a newline before enumerations.

You are right. Should be fixed now.

doc/help/elektra-algorithm.md
sanssecours: > Maybe "org" instead of "apps" would be better? (again the problem already existed earlier)

I totally overlooked that mistake 😳. The new version of the pull request contains the correct key.

sanssecours:
> Btw. please do not create branches in the main repo, it easily leads to orphans. Instead prefer to use your own repo, there you can create as many branches as you would like to have.

Okay, I will do that for the next pull request.

PR #549, opened 2016-03-03
Fix #544

### LINKED ISSUE #544: crypto tests
The crypto tests now expect both variants to be present. It would be good to have a clever method so that the test cases run only against the variant actually included.

I also changed the add_plugin heavily, I hope I did not introduce a problem in the crypto plugin, please report if the behaviour changed. See the commit 3121a8ad1fb05eceeed313cbae511069eeb2b0e0 and especially doc/decisions/cmake_plugins.md.

petermax2:
Minor improvements on crypto plugin's CMakeLists.txt and test cases.

See #544 to see the full discussion.


markus2330:
Thank you!

PR #555, opened 2016-03-07
WIP: spec fix

### LINKED ISSUE #554: conditionals: assignment does not work on new keys
```
kdb set user/example/highlevel/battery/level low
kdb set user/example/highlevel/device smartphone 
#kdb set user/example/highlevel/powersaving/gps
kdb get /example/highlevel/powersaving/gps
BLUBBERBLASEN!
```

Expected: low

tom-wa:
crude hacked together code, just a possible approach


tom-wa:
@markus2330 is this kind of approach to fix the problem with getting non-existing keys ok with you ?

markus2330:
It seems like you try to solve an ordering issue in an unappropriate way.
Certainly not every plugin is fine with getting executed twice.
Why not add an additional hook as discussed?

tom-wa:
couldn't fine any other way to do it.
the keyset needed for global plugins (spec) only exists after elektraGetDoUpdate, but the backends plugins are all executed within elektraGetDotUpdate.
placement 6-9 need to be executed after the hook.
an other way to do it would be only running 1-5 within elektraGetDoUpdate, then executing the hook, and then running 6-9, but there is the possibility that some of those needs to be executed before the hook or independent from the hook.
this was the only approach i could find without doing any bigger changes to kdbGet

src/libs/elektra/kdb.c
```diff
@@ -693,6 +726,15 @@ int kdbGet (KDB * handle, KeySet * ks, Key * parentKey)
 
 	ksRewind (ks);
 
+	if (elektraGetDoUpdate2 (split, ks, parentKey) == -1)
```
markus2330: I think the problem is: this is called too late, you cannot call doUpdate after postprocessing of the keysets.

why not:
```
elektraGetDoUpdate
hook MIDDLEGETSTORAGE
elektraGetDoUpdate2
elektraSplitGet // postprocessing
...
hook POSTGETSTORAGE
```

and spec will be called in MIDDLEGETSTORAGE

src/libs/elektra/kdb.c
```diff
@@ -508,13 +508,14 @@ static int elektraGetDoUpdate (Split * split, Key * parentKey)
 		keySetName (parentKey, keyName (split->parents[i]));
 		keySetString (parentKey, keyString (split->parents[i]));
 
-		for (size_t p = 1; p < NR_OF_PLUGINS; ++p)
+		for (size_t p = 1; p < 6; ++p)
```
markus2330: please avoid magic number 6, use a macro instead.

src/libs/elektra/kdb.c
tom-wa: elektraSplitMerge needs to be called before spec, and elektraSplitGet before elektraSplitMerge. So i guess i'd have to adjust elektraSplitGet but i'm not sure how to do the part with the syncbits

tom-wa:
@markus2330 can you take a look at the currrent approach ? 
it seems to work fine, but elektraSplitMerge is called twice. not sure how much bad the performance impact really is.

src/libs/elektra/kdb.c
```diff
@@ -526,7 +527,39 @@ static int elektraGetDoUpdate (Split * split, Key * parentKey)
 	return 0;
 }
 
+static int elektraGetDoUpdate2 (Split * split, KeySet * ks, Key * parentKey)
```
markus2330: I do not like the duplication of the code, can you add the NR_OF_PLUGINS as parameter?

markus2330:
elektraSplitMerge should be quite cheap, but better to benchmark it. Actually, having a more efficient ksAppend is in TODO. (You can take advantage of that keysets are sorted when giving them together.) If it is expensive we could avoid it to be done twice if no such hook exists?

Doesn't it create a problem if a plugin removed some keys after position 6? (Because they would stay in the merged key set then).Don't you need to work on a duplicated keyset temporarily for the first elektraSplitMerge?

markus2330:
But looks better now and less invasive, but there are still code duplication and formatting issues. Also the PR is too big, there are many unrelated parts here.

tom-wa:
@markus2330 done.

markus2330:
Thank you!

PR #557, opened 2016-03-08
JavaDoc for JNA binding + jUnit tests for Key and KeySet

Namoshek:
Added some JavaDoc for most code, but still some return values are missing (couldn't find and evaluate them). Unit tests cover most important parts of Key and KeySet class.


markus2330:
Thank you, very nice! Is there some automated way to run the unit tests?

src/bindings/jna/elektra/KDB.java
```diff
@@ -14,26 +17,50 @@
...
 
-	// wrapped methods
+
+	/*
+	 * Wrapped methods
+	 */
+
+	/**
+	 * 
+	 * @param ks
```
markus2330: Its not very useful to have an empty generated docu. What about `@copydoc`?

PR #560, opened 2016-03-12
Add redrawn version of logo

sanssecours:
I hope I correctly replaced each version of the old logo. Markus also asked about a way to convert the `.svg` file to other formats. I tried to convert them via [ImageMagick][] and [Inkscape][]. ImageMagick was not able to properly convert the logo with shadows (`circle_with_shadows.svg`). The command

```sh
convert circle_with_shadows.svg  -resize 100x100 circle.png
```
produced the following white square:

circle

Inkscape failed on the same SVG. The command 

```sh
inkscape -z -e circle.png -w 100 -h 100 circle_with_shadows.svg
```

produced the following transparent square:

circle

After these test I finally tried [svgexport][], which actually worked \o/. For example, the command

```sh
svgexport circle_with_shadows_and_background.svg circle.png 400:400
```

produces the following result:

circle

[ImageMagick]: http://www.imagemagick.org 
[Inkscape]: https://inkscape.org
[svgexport]: https://github.com/shakiba/svgexport


markus2330:
Thank you! Please prefer to put such explaining text into README.md files and not into PR descriptions where it is basically lost (or at least not so found easily) after some time.

PR #562, opened 2016-03-14
Ini fixes

tom-wa:


markus2330:
Thank you! Please always run clang-format for every PR, otherwise we will format the same things back and forth.

PR #563, opened 2016-03-18
curlget plugin

tom-wa:


markus2330:
Greatly done, thank you! Nice that you also added an example!

PR #564, opened 2016-03-24
Draft: VALIDATION: linevalidation, wordvalidation, ignorecase

tom-wa:


src/plugins/validation/validation.c
```diff
@@ -34,13 +38,34 @@ int elektraValidationSet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned, Key
...
 	{
-		const Key * meta = keyGetMeta (cur, "check/validation");
+		const Key * regexMeta = keyGetMeta (cur, "check/validation");
+
+		if (!regexMeta) continue;
 
-		if (!meta) continue;
+		const Key * icaseValidation = keyGetMeta (cur, "check/validation/ignorecase");
+		const Key * lineValidation = keyGetMeta (cur, "check/validation/line");
+		const Key * wordValidation = keyGetMeta (cur, "check/validation/word");
```
markus2330: What happens if both line and word are set?
What about having "check/validation/match" or "check/validation/boundary" which contains line, word,.. (default or none).
This would also better fit to new features like "check/validation/type" that can contain basic, extended, perl, fixed

tom-wa: word > line. line splits on newlines, word on all whitespaces.
grouping line, word and anywhere makes sense, but icase, invert have to be extra keys because any combination of the matching type, ignore case and invert is possible.

pcre works a little bit differently though, so i'm not sure if this plugin should handle pcre too or if it would be better to write a dedicated pcre plugin

markus2330:
Thank you!

Issue #565, opened 2016-03-24
keyIs(Direct)Below doesn't work with cascading keys


tom-wa:
keyIs(Direct)Below ( /test , user/test/key1 ) returns 0, expected 1

markus2330:
Thank you for reporting!

PR #566, opened 2016-03-24
INI: section fix

tom-wa:


src/plugins/ini/ini.c
```diff
@@ -268,7 +268,7 @@ static void insertKeyIntoKeySet (Key * parentKey, Key * key, KeySet * ks)
 	cursor_t savedCursor = ksGetCursor (ks);
 
 	// trying to find the keys section or sections parent section
-	const char * parent = findParent (parentKey, key, ksDup (ks));
+	char * parent = (char *)findParent (parentKey, key, ksDup (ks));
```
markus2330: Could you explain this change please?

markus2330:
Please give a short description which problem is fixed here.

src/plugins/ini/ini.c
tom-wa: findParent was previously returning the keyName of a deleted key, which, although it worked was bad code. but i should probably change the declaration of findParent instead of casting it to avoid a warning.

tom-wa:
it's mostly a workaround related to #565, exporting from dump to ini caused problems for some keys resulting in global keys getting sorted into the wrong positions because their relation to the rootkey isn't detected correctly

markus2330:
Its better to fix #565 then!

tom-wa:
@markus2330 removed #565 part, there's a fix for the findParent part and a segfault handling the "/" in this PR

markus2330:
Thank you!

PR #570, opened 2016-03-25
RESOLVER: fix #567

### LINKED ISSUE #567: Retain user/group
Resolver should retain user/group when writing to config files as root.

tom-wa:


src/plugins/resolver/resolver.c
```diff
@@ -73,6 +73,9 @@ static void resolverInit (resolverHandle * p, const char * path)
 	p->tempfile = 0;
 
 	p->path = path;
+
+	p->uid = getuid ();
+	p->gid = getgid ();
```
markus2330: I think thats wrong, we are not interested in chown to our own user (thats the case anyway), but to do a chown in the case of root to the uid/gid that was present in the file before.

src/plugins/resolver/filename.c
```diff
@@ -590,7 +650,7 @@ static int elektraResolveMapperSystem (resolverHandle * p, Key * warningsKey)
  *         be found
  * warnings will be reported to warningsKey
  */
-int ELEKTRA_PLUGIN_FUNCTION (resolver, filename) (Key * forKey, resolverHandle * p, Key * warningsKey)
+int ELEKTRA_PLUGIN_FUNCTION (resolver, filename)(Key * forKey, resolverHandle * p, Key * warningsKey)
```
markus2330: there seem to be plenty of whitespace changes, please rerun scripts/reformat-source

src/plugins/resolver/filename.c
```diff
@@ -193,7 +193,37 @@ static int elektraResolveSystem (char variant, resolverHandle * p, Key * warning
...
+		ssize_t bufSize = sysconf (_SC_GETPW_R_SIZE_MAX);
+		if (bufSize == -1) bufSize = 16384; // man 3 getpwuid
+
+		char * buf = elektraMalloc (bufSize);
+		if (!buf) return -1;
+		struct passwd pwd;
+		struct passwd * result;
+		int s;
+
+		s = getpwuid_r (getuid (), &pwd, buf, bufSize, &result);
```
markus2330: I am not sure if this is useful because it only checks for the current user, not for every possible user. You should at least avoid this code duplication.

src/plugins/resolver/resolver.c
markus2330: See comment  #567

tom-wa: Those 2 lines are just to initialize uid and gid, they get set the the actuall files owner in after the stat call in resolver, get

src/plugins/resolver/resolver.c
```diff
@@ -470,6 +473,8 @@ int ELEKTRA_PLUGIN_FUNCTION (resolver, get) (Plugin * handle, KeySet * returned,
 	{
 		// successful, remember mode
 		pk->filemode = buf.st_mode;
+		pk->gid = buf.st_gid;
```
markus2330: also update comment here

src/plugins/resolver/resolver.c
markus2330: Yes, seems to be that those two syscalls are done useless in this initialization.

You should remember them somewhere else, so that the chown call is only issued when we are root (and chown will not fail anyway). Maybe there is even a better way to check if we are able to issue chown (root comparision seems to be inaccurate, actually we want to check the capability SYS_CHOWN, but this seems to be very unportable,..).

src/plugins/resolver/filename.c
tom-wa: What if multiple users have a file with the same name and relative path in their home directories ? 
maybe we should also save the id of the user who initially mounted to file ?

markus2330: Actually, mounting should abort if the same file is mounted again. But that would be a different issue.

src/plugins/resolver/resolver.c
tom-wa: does it really matter if chown fails ? if it succeeds, the permissions are restored, if not, we most likely don't have the rights to change/restore them.

Issue #264, opened 2015-08-26
link warning in libelektra-journald.so


sl1pkn07:
when build the project:

    [ 54%] Linking C shared module ../../../lib/libelektra-journald.so
    CMakeFiles/elektra-journald.dir/journald.c.o: In the function `elektraJournaldSet':
    journald.c:(.text+0x2e1): warning: sd_journal_send_with_location was moved to libsystemd. Do not use libsystemd-journal.
    [ 54%] Built target elektra-journald

systemd 224

markus2330:
Thanks for reporting!

fberlakovich:
I am afraid you have to either reassign this issue to someone with a newer development environment (mine is Debian 7.8) or give me access to a suitable test machine. This warning does not occurr on my machine.

sl1pkn07:
same with systemd 227 (Archlinux)

markus2330:
@sl1pkn07 Which distribution? Is the issue still present with more recent versions?

@fberlakovich Where did you get FindSystemdJournal.cmake from? Maybe there is a more recent version available?

sl1pkn07:
Archlinux. And still happen with systemd 227

markus2330:
@sl1pkn07 Can you provide an Archlinux login for @fberlakovich somewhere?

To understand the bug we need much more information: What is in CMakeCache.txt, what is the gcc command directly after `Linking C shared module` (VERBOSE=1),... To fix it, we most likely need an Archlinux installation anyway.

sl1pkn07:
[ 62%] Linking C shared module ../../../lib/libelektra-journald.so
    cd /tmp/makepkg/elektra-git/src/build/src/plugins/journald && /usr/bin/cmake -E cmake_link_script CMakeFiles/elektra-journald.dir/link.txt --verbose=1
[...]

PR #572, opened 2016-04-04
Updated FindSystemdJournal CMakeModule

fberlakovich:
This could fix the warning described in #264. However, as the warning did not occur on my development machine I am not able to test it. Note also that this change might break the build for older Debian systems (which I am also not able to test)

### DIFF
--- a/cmake/Modules/FindSystemdJournal.cmake
+++ b/cmake/Modules/FindSystemdJournal.cmake
@@ -35,17 +35,18 @@ else (LIBSYSTEMD_JOURNAL_INCLUDE_DIR AND LIBSYSTEMD_ID128_INCLUDE_DIR)
     /usr/local/include
   )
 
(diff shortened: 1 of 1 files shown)


markus2330:
Why not using the search for the old one as fallback? Can you update the PR?

@sl1pkn07 can you verify if the warning is gone?

fberlakovich:
> Why not using the search for the old one as fallback? Can you update the PR?

Forget what I have done before. A more detailed look revealed that the cmake module already searches for both variantes. However, up to now the old variant was preferred. The only explanation for the warning in #264 I can find is a leftover libsystemd-journal.so.

markus2330:
So maybe we should switch the preference?

fberlakovich:
> up to now the old variant was preferred

It is switched in the PR ;).

markus2330:
Thank you!

sl1pkn07:
@markus2330 yes, is gone

greetings

PR #576, opened 2016-04-06
Ni Plugin: Fix elektraKeyGetRelativeName

sanssecours:
I hope this fix has no unintended consequences. What is the reason behind the decision to cut off the first  two characters if the plugin is mounted at `/`?


markus2330:
Thank you, thats really non-sense.

ini should have similar code, maybe you can move this function to libease and let both ini and ni use it. It would be nice to have tests for it, too.

PR #578, opened 2016-04-08
Lock plugin

KurtMi:


src/plugins/lock/README.md
```diff
@@ -0,0 +1,15 @@
+- infos = Information about the lock plugin is in keys below
+- infos/author = Kurt Micheli <e1026558@student.tuwien.ac.at>
```
markus2330: Are you sure you want to use that e-mail address? I think they shut it down when you leave TU.

src/plugins/lock/README.md
```diff
@@ -0,0 +1,15 @@
...
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/placements = pregetstorage postcommit
+- infos/status = maintained unittest tested nodep libc 0
+- infos/description =
+
+## LOCK ##
+
+This small plugin lock for each file. The granularity is huge and the logic very simple.
```
markus2330: What means "for each file"? It should be useable as global plugin, too.

src/plugins/lock/README.md
KurtMi: I am not sure but I think it will also work as global plugin.

src/plugins/lock/README.md
KurtMi: I am not sure but I think it will also work as global plugin

src/plugins/lock/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+It locks at the `GET` the accessed file with a `$FILENAME$.lock` file.
+Other `GET` request end up busy waiting until a `SET` removes the lock file.
```
markus2330: Which usecase did you have in mind to lock from get to set?

It should be enough to read-lock from PREGETSTORAGE to POSTGETSTORAGE (in get path) and a write-lock between PRESETSTORAGE (we should introduce a lock even earlier) and POSTCOMMIT/POSTROLLBACK.

src/plugins/lock/lock.c
```diff
@@ -0,0 +1,116 @@
...
+
+	char * lockFilePath = createLockFilePath (keyString (parentKey));
+	if (!lockFilePath)
+	{
+		ELEKTRA_SET_ERRORF (145, parentKey, "malloc fail in %s\n", "get");
+		return -1;
+	}
+	FILE * lockFile;
+
+	while (1)
```
markus2330: busyloop is not so ideal.

src/plugins/lock/lock.c
```diff
@@ -0,0 +1,116 @@
...
+		return -1;
+	}
+	fclose (lockFile);
+	free (lockFilePath);
+	return 1; // success
+}
+
+int elektraLockSet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_UNUSED, Key * parentKey)
+{
+	char * lockFilePath = createLockFilePath (keyString (parentKey));
```
markus2330: I think `keyString (parentKey)` is empty (or whatever was there before) for global plugins.

src/plugins/lock/testmod_lock.c
```diff
@@ -0,0 +1,148 @@
...
+	{
+		fprintf (stderr, "ERROR: can not wirte file %s\n", TESTFILELOCK);
+		goto endset;
+	}
+	fclose (file);
+
+	plugin->kdbSet (plugin, ks, parentKey);
+
+	file = fopen (TESTFILELOCK, "r");
+	succeed_if (!file, "kdbGet did not removed the lock file");
```
markus2330: nice test, but I am afraid it won't work when it releases already during kdbSet.

src/plugins/lock/lock.c
```diff
@@ -0,0 +1,116 @@
...
+	if (!lockFilePath)
+	{
+		ELEKTRA_SET_ERRORF (145, parentKey, "malloc fail in %s\n", "set");
+		return -1;
+	}
+	FILE * lockFile = fopen (lockFilePath, "r");
```
markus2330: there should a pid in the file.

markus2330:
Thank you!

src/plugins/lock/README.md
markus2330: Yes, looks like. But it was intended to be *only* a global plugin.

The idea is to completely lock down any access do the database if a single `kdbSet` is active. Use some global file, e.g. /run/elektra/kdb.pid

markus2330:
Btw. nice plugin, is this really your first one? Where there any difficulties?

markus2330:
Btw. I think we found a use case for this kind of plugin: When we do not have or should not (NFS) rely on timestamps this could be useful.

Your strategy would already work if only the `kdb` tool is used, but it would not handle long-living processes that do multiple kdbSet?

So following tasks would be required to make a plugin for NFS-conflict detections out of it:
- [ ] rename it to something less generic, unless you plan to have multiple locking features combined in this plugin
- [ ] hold lock from first kdbGet() until kdbClose() (for multiple kdbSet())
- [ ] release lock when pid behind the lock is already gone
- [ ] avoid the busy loop

@tom-wa do you have some further suggestions regarding NFS?

src/plugins/lock/README.md
KurtMi: I don not have a proper email, can I have a `@libelektra.org` mail?

KurtMi:
I update the plugin in this PR to the ordered global plugin, the `simple-lock` plugin as you proposed will follow soon.

KurtMi:
Btw. yes the first plugin, got the knowledge from reviewing and the copy-template works fine.

src/plugins/lock/README.md
markus2330: I have sent you an e-mail to your new address!

src/plugins/lock/README.md
```diff
@@ -0,0 +1,19 @@
...
+- infos/description =
+
+## LOCK ##
+
+This global lock plugin introduces a read lock while `GET` and a read/write lock
+while `SET`.
+
+A semaphore is used for the synchronisation and the implemented algorithm favors the writer,
+because updates should be propagated soon as possible and a `kdbSet` is more expansive than
+a `kdbGet`. This expansiveness comes from merging and conflict solving.
```
markus2330: It makes sense to favor the writers (because readers should get up-to-date information), but I cannot follow your explanation. Do you mean expensive instead of expansive? What has the run-time cost to do with preference?

merging is not a part of `kdbSet`, but validation is.

src/plugins/lock/lock.c
```diff
@@ -0,0 +1,209 @@
...
+ * @brief Source for lock plugin
+ *
+ * @copyright BSD License (see doc/COPYING or http://www.libelektra.org)
+ *
+ */
+
+#include "lock.h"
+
+#include <kdbhelper.h>
+#include <kdbprivate.h>
```
markus2330: is this really necessary? which functions are used from there?

markus2330:
Looks really nice! Can you rename it to semlock? Btw. the  simple-lock: plugin names should be a-z0-9 only, so maybe longlock?

Did you try to run it with the race test?

src/plugins/lock/README.md
KurtMi: Thank you

KurtMi:
> Did you try to run it with the race test?

 I tried but failed, it seems that the global hooks are being called a little weird.

I inserted printfs in the kdb.c, when the hooks get called and configured the global plugins that semlock is the only plugin and I get this output:

```
> kdb set /test/asd "asd"
OPEN
OPEN
OPEN
OPEN
PREGETSTORAGE
24966 PRE GET START
24966 PRE GET END
POSTGETSTORAGE
24966 POST GET START
24966 POST GET END
PREGETSTORAGE
24966 PRE GET START
24966 PRE GET END
PREGETSTORAGE         <---why is here a PREGETSTORAGE? 
24966 POST GET START
24966 POST GET END
CLOSE
CLOSE
PREGETSTORAGE
24966 PRE GET START
24966 PRE GET END
POSTGETSTORAGE
24966 POST GET START
24966 POST GET END
Using name system/test/asd
Set string to asd
24966 PRE SET START
24966 PRE SET END
POSTCOMMIT
24966 POST SET START
24966 POST SET END
CLOSE
CLOSE
```
Note that the `PRESETSTORAGE` is not printed.

Also I am worried about:
```
// resolver says that sync is
// not needed, so we
// skip other pre-commit
// plugins
```
in kdb.c:770. Without releasing set call no other process can make a get.

markus2330:
The placement of hooks is an ongoing discussion, see #555. I even thought about giving special placement especially for lock plugins (lockgetstorage, ..., unlocksetstorage). This would ensure that locks are always executed before and after any other plugins (without needing to resort of some weird ordering constructs).

Thus these global locking plugins are not really useful as "normal" plugins, we would not really lose something that way. And this way we could place the global plugins specifically in mind that it will *always* unlock if it was locked before. (and avoid such problems as in kdb.c:770)

@KurtMi  @tom-wa what do you think?

KurtMi:
We should discuss that topic after the meeting tomorrow. I will create an issue for the longlock, as reminder for me.

KurtMi:
I adjusted the tags and rating. I think it can be merged.

markus2330:
Thank you! Let us discuss it tomorrow.

PR #579, opened 2016-04-09
topological sorting

tom-wa:


markus2330:
Thank you, looks great!

Nice idea to use "order" when there are no topological constraints.

I have merged now and might open issues if there are some improvements needed.

PR #582, opened 2016-04-10
Libmeta: Fix linking process

sanssecours:
Please keep in mind that I do not know much about CMake. If there is a better way to fix the error:
```
  "_elektraArrayIncName", referenced from:
      _elektraMetaArrayAdd in meta.c.o
      _elektraMetaArrayToKS in meta.c.o
      _elektraSortTopology in meta.c.o
      _elektraMetaArrayToString in meta.c.o
  "_elektraKsToMemArray", referenced from:
      _elektraSortTopology in meta.c.o
make[2]: *** [lib/libelektra-meta.0.8.15.dylib] Error 1
make[1]: *** [src/libs/meta/CMakeFiles/elektra-meta.dir/all] Error 2
make: *** [all] Error 2
```
, then please tell me. Does the above error message only show up on my machine, or has anyone else seen something similar?


markus2330:
Thank you for fixing it, gcc only had linking problems when using elektraKsToMemArray dynamically (in new code not yet pushed).

Mac OS X seems to be much more strict about undefined symbols.
There are some options for ld, e.g. `--unresolved-symbols=report-all`
but I still always had problems that some symbols problematic for Mac OS X are not reported.

@tom-wa can you avoid cross-dependencies between libease and libmeta? I.e. now that libmeta links against libease it is not allowed that libease uses something from libmeta.

tom-wa:
@markus2330 I'll try to find a way to do it without elektraKsToMemArray

markus2330:
@tom-wa Its not so bad as long as the dependency is not circular. So you can use elektraKsToMemArray for now.

PR #586, opened 2016-04-11
Remove Useless CMake Linking Code

sanssecours:
Hope this change does not break anything.

### DIFF
--- a/tests/ctest/CMakeLists.txt
+++ b/tests/ctest/CMakeLists.txt
@@ -4,23 +4,3 @@ foreach (file ${TESTS})
 	do_test (${name})
 	target_link_elektra(${name} elektra-kdb)
 endforeach (file ${TESTS})
-
-target_link_elektra(test_array elektra-ease)
-target_link_elektra(test_backend elektra-plugin)
-
-target_link_elektra(test_mount elektra-plugin)
-target_link_elektra(test_plugin elektra-plugin)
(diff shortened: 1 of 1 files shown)


markus2330:
This change breaks setups where both BUILD_FULL and BUILD_STATIC are disabled. It is a goal that all tests also run with BUILD_SHARED, but this is not yet the case. The non-working ones should be disabled, though. So it should run with a reduced set but without errors. Please report issues if this is not the case (e.g. for Mac OS X).

sanssecours:
> This change breaks setups where both BUILD_FULL and BUILD_STATIC are disabled.

The following build commands seem to work fine on my computer (using GCC as compiler) after I remove the linking code:
```
cmake BUILD_FULL=OFF -DBUILD_STATIC=OFF ..
make -j 9
```
The test seem to work too. The output of `find -E 'bin' -regex 'bin/test_.[^./]+$' -exec './{}' \;` shows no errors. Do I need to do something different to reproduce the problems you mentioned? Sorry, if this a stupid question 😅…

markus2330:
Yes, you need a `-D` in front of BUILD_FULL.

Don't worry, there are no stupid questions. Its much easier to spot such `-D` errors together than finding them on your own.

markus2330:
P.s. I am keen on docu/comments that explain these coherences in the build system. You are not the first to tap into such issues, e.g. #113

PR #587, opened 2016-04-11
Ignore `Makefile` in Project Root

sanssecours:
This is a rather specific change. If you do not like it, then please just close this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ build*
 
 #ignore emacs backup files
 \#*#
+
+#ignore Makefile in project root
+/Makefile


markus2330:
While in-source builds are supported, I would not recommended them.

sanssecours:
Thanks for merging!

> While in-source builds are supported, I would not recommended them.

Actually I never tried in-source builds. I just use the Makefile to call various build commands via <kbd>⌘</kbd> + <kbd>B</kbd> inside my editor. Currently my temporary `Makefile` looks like this:

```make
run:
	cd build && make -j 9 test_keyname > /dev/null && bin/test_keyname
```

markus2330:
Ok, this makes sense. Thanks for sharing. I use
`set makeprg=make\ -j\ 5\ --no-print-directory\ -C\ /path/to/Elektra/build`.

PR #588, opened 2016-04-11
Move elektraKeyGetRelativeName to Libease

sanssecours:
(This time) I added some tests for `elektraKeyGetRelativeName`. If I should change anything, then please just comment below.

### DIFF
--- a/tests/ctest/test_keyname.c
+++ b/tests/ctest/test_keyname.c
@@ -53,6 +53,9 @@ static void test_relative_cascading ()
 	parent = keyNew ("/cascading\\/mountpoint/");
 	child = keyNew ("user/cascading\\/mountpoint/\\/dot");
 	test_relative ("\\/dot", child, parent);
+
+	child = keyNew ("user/second_level/cascading\\/mountpoint/\\/dot");
+	test_relative ("\\/dot", child, parent);
 }
 
 static void test_relative_generic ()
--- a/tests/ctest/test_keyname.c
+++ b/tests/ctest/test_keyname.c
@@ -1,0 +1,96 @@
+/**
+ * @file
+ *
+ * @brief Test suite for Libease functions accessing key name data.
+ *
+ * @copyright BSD License (see doc/COPYING or http://www.libelektra.org)
+ */
+
+#include <kdbease.h>
+
+#include "tests.h"
+
+#define ERROR_STRING_SIZE 1024
+
+#define test_relative(expected, child, parent)                                                                                             \
+	{                                                                                                                                  \
+		char error[ERROR_STRING_SIZE];                                                                                             \
+		const char * result = elektraKeyGetRelativeName (child, parent);                                                           \
+		snprintf (error, sizeof (error),                                                                                           \
+			  "Relative name of key was wrong.\nParent:   %s\nChild:    %s\nExpected: %s\nResult:   %s\n", keyName (parent),   \
+			  keyName (child), expected, result);                                                                              \
+		succeed_if (strcmp (result, expected) == 0, error);                                                                        \
+	}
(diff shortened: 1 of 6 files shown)


tests/ctest/test_keyname.c
```diff
@@ -0,0 +1,87 @@
...
+ * @copyright BSD License (see doc/COPYING or http://www.libelektra.org)
+ */
+
+#include <kdbease.h>
+
+#include "tests.h"
+
+#define ERROR_STRING_SIZE 1024
+
+static void test_relative (const char * expected, const char * result)
```
markus2330: It might be useful to let test_relative be a macro: then the line information would be more useful.

markus2330:
Thank you, looks great! I am only missing test cases with escaped slashes. It looks like that they are not even handled correctly.

If you want I can accept this PR and improvements  regarding escaped slashes can follow at a later PR?

tests/ctest/test_keyname.c
sanssecours: > It might be useful to let test_relative be a macro: then the line information would be more useful.

Thanks for the advice. I just changed `test_relative` accordingly.

sanssecours:
> Thank you, looks great! I am only missing test cases with escaped slashes. It looks like that they are not even handled correctly.

I added some test cases for keys containing escaped slashes. It looks like the code handles escaped slashes correctly, as long as the following two preconditions hold:

1. A cascading key maps to a single level namespace. So `/cascading/key` can be `system/cascading/key` or `user/cascading/key`, but not `user/secondlevel/cascading/key`.
2. The first level that a cascading key maps to – `user`, `system` in the examples above –, contains no escaped slashes. This seems to be the case, since it is not possible to create a (non cascading) key containing an escaped slash in the first level, via `keyNew` or `keySetName()`.

markus2330:
Looks good, thank you.

What happens if `/cascading/key` `user/secondlevel/cascading/key` are used?

The second precondition is irrelevant, as you noted the API takes care of it.

A small note: The label "in progress" is still there.

sanssecours:
> What happens if `/cascading/key` `user/secondlevel/cascading/key` are used?

Then the function returns `scading/key`. This happens cause the function “removes” the first level of the key resulting in `secondlevel/cascading/key`. After that it “removes” 14 (length of `/cascading/key`) characters.

The absolute name of a child always contains more levels, than its parent. So

Parent: `user/parent`
Child: `user/parent/key`

is fine, while 

Parent: `user/parent`
Child: `user`

should never happen.

So let us take a look at an example, where this property holds.

Parent Key: `/cascading/cas/` (Absolute name: `user/cascading/cas/`)
Child Key 1: `user/cascading/cas/key1`
Child Key 2: `user/second_level/cascading/cas/key1`

The function works correctly for “Child Key 1”:

1. Initial name: `user/cascading/cas/key1`
2. Add offset for first level (`user` — 4 chars): `cascading/cas/key1`
3. Add offset for parent (`/cascading/cas/` — 15 chars):  `key1`

It fails for  “Child Key 2”:

1. Initial name: `user/second_level/cascading/cas/key2`
2. Add offset for first level (`user` — 4 chars): `/second_level/cascading/cas/key2`
3. Add offset for parent (`/cascading/cas/` — 15 chars):  `ascading/cas/key2`

markus2330:
Ok, can you create an issue with this detailed analysis? When I merge this PR we might forget about it.

sanssecours:
> Ok, can you create an issue with this detailed analysis?

Since I did not want to create another issue, I just fixed the problem mentioned above (Commit 4d64e48).

markus2330:
Thanks, thats even better!

PR #590, opened 2016-04-13
fixed singned/unsigned comparison

tom-wa:


src/libs/meta/meta.c
```diff
@@ -1211,7 +1211,7 @@ int elektraSortTopology (KeySet * ks, Key ** array)
 					retVal = -1;
 					break;
 				}
-				else if (i == j)
+				else if ((long)i == (long)j)
 				{
```
markus2330: Is there no better solution than a cast?

Please also document that this branch is empty and purpose and that the case statement above has a fallthrough.

markus2330: Btw. is it difficult to ignore missing deps?

I am thinking about an optimization where I could add new keys on demand, thus deps that did not exist before might exist for the next sort.

From usability point you could check if deps are correct keynames (try if a keyName() has the same content as what you gave with keySetName()) and return -1 on the cases where there is such a problem.

markus2330: (Btw. if it is expensive for topology sort to have missing deps then it would of course better to remember missing deps and add them when needed)

tom-wa: i'll try it.

tom-wa: so, if a dependency is a valid keyName, but doesn't exist it should be ignored ? 
that'd be no problem.

markus2330:
Yes, exactly: "dependency is a valid keyName, but doesn't exist it should be ignored "

Please ping if ready.

tom-wa:
@markus2330 done, i'll add documentation later, i'm currently working on rename

markus2330:
There is now a memory leak, please fix it. Also please add docu now, then its done ;)

tom-wa:
the memleak is in key_testname.c
```
==30426== 110 (64 direct, 46 indirect) bytes in 1 blocks are definitely lost in loss record 50 of 59
==30426==    at 0x4C29C0F: malloc (vg_replace_malloc.c:299)
==30426==    by 0x4E61C5A: elektraKeyMalloc (key.c:85)
==30426==    by 0x4E61C80: keyVNew (key.c:215)
==30426==    by 0x4E61D30: keyNew (key.c:199)
==30426==    by 0x401C04: test_relative_root (test_keyname.c:34)
==30426==    by 0x401A12: main (test_keyname.c:92)
```

i'll write a fix for it after i'm done with rename

markus2330:
It would be great when you can reorder the tasks, I would like to do some benchmarks they would be invalid if something relevant changes afterwards.

markus2330:
(bzw. if you want you can keep the benchmarks and oldTopologicalSort) in a new file in the benchmarks folder)

markus2330:
Please ping if ready.

tom-wa:
unless someone is going to implement the keysets as hashtables i don't think that the old version would be of any use because ~90% of the lookups will fail.

tom-wa:
@markus2330 done

markus2330:
yes, Kurt implements a hash set for keysets. But it still might not matter much: It looks like you also modify the key sets heavily, thus the hash set would not be rebuild every time.

Yes, so better to discard it, the new one seems to be better anyway.

markus2330:
Thank you, thats great! But I think it needs a rebase because I just merged the memleak fix.

tom-wa:
@markus2330 does it still need the rebase ?

markus2330:
No, the page only needed a refresh, then github figured it out.

markus2330:
All tests working again, well done!

PR #592, opened 2016-04-13
keyIsBelow fix + tests

tom-wa:
rewrote keyIsBelow because there were still some cascading edge cases missing + tests

### DIFF
--- a/src/libs/elektra/keytest.c
+++ b/src/libs/elektra/keytest.c
@@ -31,7 +31,6 @@
 #include "kdbinternal.h"
 #include "kdbprivate.h"
 
-#include <stdio.h>
 
 /**
  * @defgroup keytest Methods for Making Tests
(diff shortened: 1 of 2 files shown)


markus2330:
Thank you! Can you add more test cases with escaped slashes? In the implementation please use the next key level iteration, otherwise it certainly will get inconsistent over time. (Do not try to detect yourself if a `/` is escaped)

tom-wa:
@markus2330 done

src/libs/elektra/keytest.c
```diff
@@ -419,68 +449,68 @@ succeed_if (keyRel (key, check) >= 0, "should be the same or below");
...
+ user/key/myself
+ user/key/sibling
+ @endverbatim
  *
  * @code
-keySetName (key, "user/key/folder");
+ keySetName (key, "user/key/folder");
 keySetName (check, "user/notsame/folder");
 succeed_if (keyRel (key, check) < -2, "key is not below, but same namespace");
- * @endcode
```
markus2330: there are a lot of formatting issues here! Did you use clang-format 3.8?

src/libs/elektra/keytest.c
```diff
@@ -31,6 +31,8 @@
 #include "kdbinternal.h"
 #include "kdbprivate.h"
 
+#include <stdio.h>
```
markus2330: This shouldn't be included here.

markus2330:
Thank you, looks much nicer now. But there are bunch of weird formatting issues and you included stdio.h?

src/libs/elektra/keytest.c
tom-wa: i used scripts/reformate-source which seems to call clang-format-3.8 + i have clang-format-3.8 installed.
just ran the script again, same issue

Issue #404, opened 2015-11-25
dbus and rename plugin do not work together


machinekoder:
There is a problem when the dbus and rename plugins are used at the same time. The dbus plugin seems not be aware of the renamed keys and notifies everyone with the raw key name. Furthermore it looks like the change detection does not work either.

``` bash
$ kdb mount `pwd`/mkwrapper.ini system/machinekit/cnc ini type dbus rename tolower=0
$ kdb set system/machinekit/cnc/display/display test2
```

``` bash
$ dbus-monitor --system type='signal',interface='org.libelektra',path='/org/libelektra/configuration'
ignal sender=org.freedesktop.DBus -> dest=:1.544 serial=2 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameAcquired
   string ":1.544"
signal sender=:1.545 -> dest=(null destination) serial=2 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
   string "system/machinekit/cnc/AXIS_0"
signal sender=:1.545 -> dest=(null destination) serial=3 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
   string "system/machinekit/cnc/AXIS_0/BACKLASH"
signal sender=:1.545 -> dest=(null destination) serial=4 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
   string "system/machinekit/cnc/AXIS_0/FERROR"
signal sender=:1.545 -> dest=(null destination) serial=5 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
   string "system/machinekit/cnc/AXIS_0/HOME"
signal sender=:1.545 -> dest=(null destination) serial=6 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
   string "system/machinekit/cnc/AXIS_0/HOME_FINAL_VEL"
signal sender=:1.545 -> dest=(null destination) serial=7 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
   string "system/machinekit/cnc/AXIS_0/HOME_IGNORE_LIMITS"
signal sender=:1.545 -> dest=(null destination) serial=8 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
   string "system/machinekit/cnc/AXIS_0/HOME_IS_SHARED"
[...]

PR #593, opened 2016-04-14
rename updates

tom-wa:
like this ?
get/case will be applied initially to the keyset, then the normal transformations are done.
on set the keyname will be either:
 restored and converted to lower or uppercase depending on set/case
of
if set/case == keyname: not restored and the transformed name kept.

### DIFF
--- a/src/plugins/list/testmod_list.c
+++ b/src/plugins/list/testmod_list.c
@@ -31,6 +31,7 @@ static void doTest ()
 			       keyNew ("user/plugins/#1/placements/set", KEY_VALUE, "presetstorage", KEY_END),
 			       keyNew ("user/plugins/#1/placements/get", KEY_VALUE, "pregetstorage", KEY_END), KS_END);
 	Key * parentKey = keyNew ("user/tests/list", KEY_END);
+	ksAppendKey (ks, parentKey);
 	PLUGIN_OPEN ("list");
 
 	Plugin * check = elektraPluginOpen ("keytometa", modules, ksNew (5, KS_END), errorKey);
--- a/src/plugins/list/testmod_list.c
+++ b/src/plugins/list/testmod_list.c
@@ -31,7 +31,6 @@ static void doTest ()
 			       keyNew ("user/plugins/#1/placements/set", KEY_VALUE, "presetstorage", KEY_END),
 			       keyNew ("user/plugins/#1/placements/get", KEY_VALUE, "pregetstorage", KEY_END), KS_END);
 	Key * parentKey = keyNew ("user/tests/list", KEY_END);
-	ksAppendKey (ks, parentKey);
 	PLUGIN_OPEN ("list");
 
 	Plugin * check = elektraPluginOpen ("keytometa", modules, ksNew (5, KS_END), errorKey);
--- a/src/plugins/rename/README.md
+++ b/src/plugins/rename/README.md
@@ -16,6 +16,34 @@ with the required names or case.
 
 If keys are renamed, their original name is stored in the `origname` MetaKey.
 
+There are 2 types of transformations: 
+* basic
+* advanced
+
+
+## BASIC TRANSFORMATIONS ##
+
+are applied before and after the advanced transformations.
+
+### GET ###
+`get/case`
+* toupper
+* tolower
+* unchanged    // this is the default value if no configuration is given
+converts the whole keyname below the parentKey to upper- or lowercase. if no configuration or `unchanged` is used no transformation is done here.
+
+### SET ###
+`set/case`
+* toupper
+* tolower
+* keyname
+* unchanged   // this is the default value if no configuration is given
+
+`toupper` or `tolower` tells the rename plugin to convert the whole keyname below below the parentKey to lower or uppercase.
+`unchanged` returnes the key to it's original name
+`keyname` tells the plugin to keep the name of the advanced transformation
+
+## ADVANCED TRANSFORMATIONS ##
 
 ## CUT ##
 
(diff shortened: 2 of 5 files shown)


src/plugins/rename/rename.c
```diff
@@ -315,38 +333,84 @@ int elektraRenameSet (Plugin * handle, KeySet * returned, Key * parentKey)
...
 	ksDel (iterateKs);
 
+	ksRewind (returned);
+	elektraFree (parentKeyName);
 	return 1; /* success */
 }
 
 Plugin * ELEKTRA_PLUGIN_EXPORT (rename)
 {
-	// clang-format off
```
markus2330: I think its nicer when you leave this customized formatted, don't you agree?

markus2330:
Thank you! Now only the profile plugin and we should be done with new features for this release!

Can you add test cases for the new features?

There are memory problems reported by build server.

I do not have time for a closer look today, sorry!

src/plugins/rename/rename.c
tom-wa: i honestly have no idea why that line always disappears...

tom-wa:
@markus2330 done

src/plugins/rename/rename.c
```diff
@@ -1,9 +1,9 @@
 /**
- * @file
+ * \file
```
markus2330: @ is the correct one, see doc/CODING.md

markus2330:
What does the `keyname` do? Please add info in the README.

markus2330:
Any progress here? The issues should only be minor?

tom-wa:
@markus2330 done

src/plugins/rename/rename.c
```diff
@@ -203,23 +228,16 @@ static Key * restoreKeyName (Key * key, const Key * parentKey, const Key * confi
 	{
 		if (strcmp (keyString (origNameKey), keyName (key)))
 		{
-			int hasSync = keyNeedSync (key); // test_bit(key->flags, KEY_FLAG_SYNC);
```
markus2330: Why did you remove these lines which restore the sync bit? They are needed for #404. We should always put comments above lines that fix a bug so that everyone know what they are for.

src/plugins/rename/rename.c
```diff
@@ -1,9 +1,9 @@
 /**
  * @file
  *
- * @brief A plugin that converts keys to metakeys and vice versa
+ * \brief A plugin that converts keys to metakeys and vice versa
```
markus2330: please change, it should be `@`

src/plugins/rename/README.md
```diff
@@ -16,6 +16,34 @@ with the required names or case.
...
+* basic
+* advanced
+
+
+## BASIC TRANSFORMATIONS ##
+
+are applied before and after the advanced transformations.
+
+### GET ###
+`get/case`
```
markus2330: one sentence here would be good.

src/plugins/rename/README.md
```diff
@@ -16,6 +16,34 @@ with the required names or case.
...
+* toupper
+* tolower
+* keyname
+* unchanged   // this is the default value if no configuration is given
+
+`toupper` or `tolower` tells the rename plugin to convert the whole keyname below below the parentKey to lower or uppercase.
+`unchanged` returnes the key to it's original name
+`keyname` tells the plugin to keep the name of the advanced transformation
+
+## ADVANCED TRANSFORMATIONS ##
```
markus2330: Correct level of title? (`#` instead of `##`?)

src/plugins/list/testmod_list.c
```diff
@@ -31,7 +31,6 @@ static void doTest ()
 			       keyNew ("user/plugins/#1/placements/set", KEY_VALUE, "presetstorage", KEY_END),
 			       keyNew ("user/plugins/#1/placements/get", KEY_VALUE, "pregetstorage", KEY_END), KS_END);
 	Key * parentKey = keyNew ("user/tests/list", KEY_END);
-	ksAppendKey (ks, parentKey);
```
markus2330: should not be here?

src/plugins/rename/rename.c
```diff
@@ -207,11 +234,6 @@ static Key * restoreKeyName (Key * key, const Key * parentKey, const Key * confi
 			Key * result = keyDup (key);
 			keySetName (result, keyString (origNameKey));
 			keySetMeta (result, ELEKTRA_ORIGINAL_NAME_META, 0);
-
-			if (!hasSync)
```
markus2330: Still removed. Could you also please add a comment that it is here because of #404. (At all places)

src/plugins/rename/README.md
```diff
@@ -8,14 +8,51 @@
...
+* toupper
+* tolower
+* keyname
+* unchanged   // this is the default value if no configuration is given
+
+`toupper` or `tolower` tells the rename plugin to convert the whole keyname below below the parentKey to lower or uppercase.
+
+`unchanged` returnes the key to it's original name
+
+`keyname` tells the plugin to keep the name of the advanced transformation
```
markus2330: one example would be great!

src/plugins/rename/testmod_rename.c
```diff
@@ -395,6 +395,26 @@ static void test_mixCase ()
 	PLUGIN_CLOSE ();
 }
 
+static void test_write ()
```
markus2330: can you also test the other direction (upper/lower)

tom-wa:
@markus2330 done

markus2330:
Thank you!

PR #594, opened 2016-04-14
Libease: Minor Changes for `test_keyname`

sanssecours:
Hope this does not break anything :).


markus2330:
It only breaks my editor ;)

Thanks for readding, such codes are an interesting test.

PR #598, opened 2016-04-17
elektraMetaArrayToKS fix array parent

tom-wa:

Issue #51, opened 2014-08-12
use relative keynames in storage and export files


iandonnelly:
- [X] simpleini
- [X] ni
- [x] tcl
- [ ] dump
- [ ] yajl with cascading keys
- [ ] xmltool (not really relevant, is superseded by xerces plugin)

I am currently having a problem with kdb import. When I try to import a file:
    cat backup.ecf | sudo kdb import system/restore
None of the keys are added from backup.ecf (which is in dump format).

If I have a file such as backup.txt in line format,
    cat backup.txt | sudo kdb import system/restore line
Works as expected.

Can anybody confirm?

markus2330:
ni is fixed in b91ba11a9f9841db082653ed66c50e274979c088

so we still have (at least): tcl, xmltool and dump

markus2330:
Some notes: `doc/decisions/relative.md` describes a bit more the background.

And as scattert in some todos (e.g. TOOLS) the idea is that `kdb export` does an export for each namespace, making development of plugins easier.

E.g. when you do `kdb export /hello ni` you would get following output:
```
kdb import user/hello ni << UNIQUE_EOF
relative/path/below/user/hello
...
UNIQUE_EOF
kdb import system/hello ni << UNIQUE_EOF
relative/path/below/system/hello
...
UNIQUE_EOF
```

So one can simply execute the exported file and gets everything reimported without having to care about which parts are exported with which format.
(we

As said for dump we need some compatibility so that old dump files (with absolute pathes) can still be read and even be written if not configured to use the new-style.

markus2330:
@sanssecours do you have something ready for this PR? E.g. the fix for cascading export/import?

sanssecours:
> @sanssecours do you have something ready for this PR? E.g. the fix for cascading export/import?

No. Otherwise I would have opened a pull request.

markus2330:
Ok, I actually meant "which could be ready for this release" (which is unlikely to happen today or tomorrow, too many issues are still open). Maybe we should at least output a warning that cascading import/export currently is broken for non-relative plugins? Or even tag plugins if they are relative or absolute and have different import/export code. (I would prefer to avoid it, but it seems unlikely that all plugins will be fixed anytime soon, or will they?)

sanssecours:
> Ok, I actually meant "which could be ready for this release" (which is unlikely to happen today or tomorrow, too many issues are still open).

Sorry, I do not think so. Unless the release is in a few weeks/months.

[...]

PR #599, opened 2016-04-17
Ini fix key is below

tom-wa:


markus2330:
Test case fails:
```
Compare <
>, with <globalkey1 = globalval1
>
in file /home/jenkins/workspace/workspace/elektra-mergerequests/src/plugins/ini/ini/emptyLines, line 1.
/home/jenkins/workspace/workspace/elektra-mergerequests/tests/cframework/tests.c:174: error in compare_line_files: comparing lines failed
/home/jenkins/workspace/workspace/elektra-mergerequests/src/plugins/ini/testmod_ini.c:426: error in test_preserveEmptyLines: files do not match as expected
```

tom-wa:
this problem is because it depends on #598 
anyway, please don't merge now, i found a problem which makes preserving order while a keyset is appended impossible, currently trying to find a best afford solution.

```
system/section   binary  order #0
system/section/zkey1 "1" order #1
system/section/zkey2 "2" order #2
system/section2/key "bla" order #3
```
- append keyset
```
system/section/subsection
system/section/subsection/key
system/section/zkey3
```
it's impossible to get the right order number for `zkey3` from the keyset.

markus2330:
Why? There might be more than one "right" number: either you give it `#2#0`, i.e. always appending keys to their respective section (if available) OR give it `#4` and reopen `[section]` for the last key. The first solution might be better because it is "only" an ordering problem and not also a problem of how to restore multiple sections correctly.

So first you have:
```
[section]
zkey=1
zkey2=2
[section2]
key=blah
```

and then

```
[section]
zkey=1
zkey2=2
zkey3 =3
[section/subsection]
key=whatever
[section2]
key=blah
```

But its not really an issue how to do it at long it properly roundtrips and the users ordering will be honored. If no order was given, you can order them as you think it is at least surprising.

tom-wa:
i want to prevent reopening the sections whenever it's possible because it can become quite a mess.

the issue is that `section/subsection`'s order number should be #2/#1 (inserted directly below zkey2 and above `section2`) and `section/subsection/key`'s #2/#2, but when trying to insert `zkey3` it's position within the keyset would also be directly below `zkey2` and it's order number #2/#1

i've already solved that issue by using "order" just for sections, and having an additional metakey for the key order relative to their sections. i'm currently just trying to figure out if it's possible to get rid of the additional metakeys at the end of kdbGet

markus2330:
Ok, thank you!

tom-wa:
```
 % kdb export /examples
[highlevel/gps]
status =
[highlevel/powersaving]
gps =
[]
highlevel =
highlevel/device =
highlevel/device = vehicle
```

the first `highlevel/device` key is the `spec` key, the second one the `user`
how should this case be treated ?

tom-wa:
to clarify: `/examples` is the sectionKey here, `[]` kinda "nullifies" the previous section (all keys above don't have a namespace) and both `spec/examples/highlevel/device` and `user/examples/highlevel/device` are below `/examples`

markus2330:
This should be fixed within `kdb export`: every namespace should be exported separately. There is nothing useful you can do within the storage plugin. See doc/decisions/relative.md and #51

Issue #603, opened 2016-04-18
qt-gui fails to synchronize because of readonly plugins


fberlakovich:
When trying to synchrinize in the qt-gui (even without changes) it displays the following error message

`Error (#84) occurred!
Description: Read only plugin, kdbSet not supported
Ingroup: kdb
Module: 
At: /home/felixl/development/elektra/libelektra/src/libs/elektra/plugin.c:390
Reason: the key system/elektra/version/constants/KDB_VERSION (expected system/elektra/version) was modified to 0.8.15 (expected Below are version information of the Elektra Library you are currently using)
Mountpoint: system/elektra/version
Configfile: 
`

This may be a regression of #208, but it seems that this time no saving at all is possible.

markus2330:
I can confirm the issue:
```
1 Warning was issued:
 Warning number: 36
	Description: could not unlink file
	Ingroup: plugin
	Module: resolver
	At: /home/markus/Projekte/Elektra/current/src/plugins/resolver/resolver.c:982
	Reason: the file "/etc/hosts.8005:1460995548.766905.tmp" because of "Datei oder Verzeichnis nicht gefunden"
	Mountpoint: system/hosts
	Configfile: 
Error (#84) occurred!
Description: Read only plugin, kdbSet not supported
Ingroup: kdb
Module: 
At: /home/markus/Projekte/Elektra/current/src/libs/elektra/plugin.c:390
Reason: the key system/elektra/version/constants/KDB_VERSION (expected system/elektra/version) was modified to 0.8.15 (expected Below are version information of the Elektra Library you are currently using)
Mountpoint: system/elektra/version
Configfile: 
```

I cannot reproduce it using the kdb tool, setting the same text again works without error:
```
kdb set system/elektra/version/constants/KDB_VERSION 0.8.15
```

@0003088 do you have any idea? Does it also occur on your installation?

0003088:
I can reproduce the bug with qt5.6. I will try to figure out what happens. Any clue as to the time the bug appeared for the first time?

markus2330:
It slipped in during 0.8.14..0.8.15. So 0.8.15 was unfortunately released with this bug. Do you have some idea which commit it could be or do we need git-bisect?

0003088:
No idea, really. It's a strange error, could it have something to do with encoding or something? The string from the gui is initially a  QString and the error occurs in a c file, maybe something went wrong on the way.

fberlakovich:
[...]

PR #604, opened 2016-04-18
Kdb sync in libtools

fberlakovich:
Because of #603 I could not fully test the qt-gui update, however the wrapper should provide a good starting ground to easily test KDB synchronisations.

### DIFF
--- a/src/libs/tools/include/merging/mergingkdb.hpp
+++ b/src/libs/tools/include/merging/mergingkdb.hpp
@@ -25,11 +25,18 @@ namespace merging
  * Provides a merging wrapper around a KDB instance. The wrapper allows to pass
  * a three way merger instance that is used to resolve conflicts during KDB set.
  */
-class MergingKDB : KDB{
+class MergingKDB : KDB
+{
 public:
-	MergingKDB() { }
-	MergingKDB(KDB & kdb) : KDB(kdb) { }
-	~MergingKDB() { }
+	MergingKDB ()
+	{
+	}
+	MergingKDB (KDB & kdb) : KDB (kdb)
+	{
+	}
+	~MergingKDB ()
+	{
+	}
 
 	/**
 	 * Behaves like the KDB get function
@@ -75,7 +82,7 @@ private:
 class MergingKDBException : public KDBException
 {
 public:
-	MergingKDBException (Key key, KeySet conflicts) : KDBException(key), m_conflicts(conflicts)
+	MergingKDBException (Key key, KeySet conflicts) : KDBException (key), m_conflicts (conflicts)
 	{
 	}
 
@@ -88,20 +95,17 @@ public:
 		return "Exception while merging conflicting KeySets";
 	}
 
-	KeySet getConflicts() const {
+	KeySet getConflicts () const
+	{
 		return m_conflicts;
 	}
 
 private:
 	KeySet m_conflicts;
 };
-
-
 }
 }
 }
 
 
 #endif /* KDBMERGE_HPP_ */
-
-
--- a/src/libs/tools/include/merging/mergingkdb.hpp
+++ b/src/libs/tools/include/merging/mergingkdb.hpp
@@ -25,7 +25,7 @@ namespace merging
  * Provides a merging wrapper around a KDB instance. The wrapper allows to pass
  * a three way merger instance that is used to resolve conflicts during KDB set.
  */
-class MergingKDB : KDB
+class MergingKDB : public KDB
 {
 public:
 	MergingKDB ()
@@ -34,7 +34,7 @@ public:
 	MergingKDB (KDB & kdb) : KDB (kdb)
 	{
 	}
-	~MergingKDB ()
+	virtual ~MergingKDB ()
 	{
 	}
 
--- a/src/libs/tools/include/merging/mergingkdb.hpp
+++ b/src/libs/tools/include/merging/mergingkdb.hpp
@@ -25,9 +25,10 @@ namespace merging
  * Provides a merging wrapper around a KDB instance. The wrapper allows to pass
  * a three way merger instance that is used to resolve conflicts during KDB set.
  */
-class MergingKDB {
+class MergingKDB : KDB{
 public:
-	MergingKDB(KDB * kdb) : innerKdb(kdb) { }
+	MergingKDB() { }
+	MergingKDB(KDB & kdb) : KDB(kdb) { }
 	~MergingKDB() { }
 
 	/**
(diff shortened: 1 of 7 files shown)


markus2330:
Thank you! Very nice solution! Actually, I also face #603 and thought it could be related to merging? So even with this PR #603 persists?

src/libs/tools/include/merging/mergingkdb.hpp
```diff
@@ -0,0 +1,105 @@
...
+{
+
+namespace merging
+{
+
+/**
+ * Provides a merging wrapper around a KDB instance. The wrapper allows to pass
+ * a three way merger instance that is used to resolve conflicts during KDB set.
+ */
+class MergingKDB {
```
markus2330: why not derive from KDB?

src/libs/tools/include/merging/mergingkdb.hpp
```diff
@@ -0,0 +1,105 @@
...
+public:
+	MergingKDB(KDB * kdb) : innerKdb(kdb) { }
+	~MergingKDB() { }
+
+	/**
+	 * Behaves like the KDB get function
+	 *
+	 * @see KDB
+	 */
+	virtual int get (KeySet & returned, std::string const & keyname);
```
markus2330: then you can use "override" here.

src/libs/tools/include/merging/mergingkdb.hpp
```diff
@@ -0,0 +1,105 @@
...
+	 * @throws MergingKDBException
+	 */
+	virtual int set (KeySet & returned, Key & parentKey, ThreeWayMerge & merger);
+
+private:
+	KDB * innerKdb;
+	KeySet base;
+};
+
+class MergingKDBException : public Exception
```
markus2330: Also, should be derived from KDBException (or even use KDBException)?
Should contain key, because maybe other warnings are present, too.

fberlakovich:
> Thank you! Very nice solution! Actually, I also face #603 and thought it could be related to merging? So even with this PR #603 persists?

Yes, the issue exists with and without this PR.

src/libs/tools/include/merging/mergingkdb.hpp
markus2330: Btw. did you use clang-format?

src/tools/qt-gui/src/treeviewmodel.hpp
```diff
@@ -290,8 +292,8 @@ class TreeViewModel : public QAbstractListModel
 	QList<ConfigNodePtr> m_model;
 	kdb::Key m_root;
 	kdb::KDB * m_kdb;   // only held by root node of TreeViewModel
-	kdb::KeySet m_base; // only held by root node of TreeViewModel
 	kdb::Key m_metaModelParent;
+	MergingKDB mergingKdb;
```
markus2330: why not using MergingKDB for `m_kdb`?

src/tools/qt-gui/src/treeviewmodel.hpp
```diff
@@ -27,6 +27,8 @@
 
 class Visitor;
 
+using namespace kdb::tools::merging;
```
markus2330: using in `hpp` is a no-go, please remove it.

src/tools/qt-gui/src/treeviewmodel.cpp
```diff
@@ -671,48 +631,35 @@ void TreeViewModel::synchronize ()
...
-		{
-			KeySet theirs = ours.dup ();
-			m_kdb->get (theirs, m_root);
-
-			KeySet result = handleConflict (theirs, m_base, ours);
-
-			m_kdb->set (result, m_root);
 
 #if DEBUG && VERBOSE
 			std::cout << "guitest: exception: now after set" << std::endl;
```
markus2330: indentation

src/libs/tools/include/merging/mergingkdb.hpp
fberlakovich: > why not derive from KDB?

I was uncertain, but maybe inheritance is a cleaner solution in this case. However, the `set` interface should stay different from the base variant because it has a slightly different behaviour (due to the get the passed KeySet is updated). Do you agree?

src/libs/tools/src/merging/mergingkdb.cpp
```diff
@@ -0,0 +1,86 @@
...
+	return ret;
+}
+
+int MergingKDB::set (KeySet & returned, std::string const & keyname, ThreeWayMerge & merger)
+{
+	Key parentKey (keyname.c_str (), KEY_CASCADING_NAME, KEY_END);
+	return set(returned, parentKey, merger);
+}
+
+int MergingKDB::set (KeySet & returned, Key & parentKey, ThreeWayMerge & merger)
```
markus2330: rename to sync?

src/libs/tools/include/merging/mergingkdb.hpp
fberlakovich: > Also, should be derived from KDBException (or even use KDBException)?
> Should contain key, because maybe other warnings are present, too.

Agree.

src/tools/qt-gui/src/treeviewmodel.hpp
markus2330: still not fixed, any specific reason why not using the name `m_kdb`? `mergingKdb` does not comply with naming convention.

markus2330:
There was no ping, so I did not see the new changes. Rename to `m_kdb` seems to be the only leftover. Please ping if finished.

src/libs/tools/include/merging/mergingkdb.hpp
```diff
@@ -0,0 +1,111 @@
...
+{
+
+namespace merging
+{
+
+/**
+ * Provides a merging wrapper around a KDB instance. The wrapper allows to pass
+ * a three way merger instance that is used to resolve conflicts during KDB set.
+ */
+class MergingKDB : KDB
```
markus2330: Is it a private inheritance on purpose? Isn't it a subtype? (otherwise one would write `public` there)

src/tools/qt-gui/src/treeviewmodel.hpp
fberlakovich: Sry, I misunderstood the comment.

src/libs/tools/include/merging/mergingkdb.hpp
fberlakovich: > Is it a private inheritance on purpose?

No, I am just a C++ noob :D

fberlakovich:
Fixes are pushed. I also made the destructor virtual as this was an Eclipse recommendation due to virtual methods in the class.

markus2330:
Thank you, well done! (Btw. virtual destructor in KDB is still missing)

Issue #628, opened 2016-04-26
rename: testcase fails at wheezy


markus2330:
http://community.markus-raab.org:8080/job/elektra-gcc-configure-debian-wheezy/lastFailedBuild/console
```
src/plugins/rename/testmod_rename.c:412: error in test_write: key1s name was not correctly saved
```

Issue #613, opened 2016-04-23
KDB Conflict Tests Fail


sanssecours:
Below is the current output of `testkdb_conflict` on my machine. If you need additional details, please just comment below.

```
Running main() from gtest_main.cc
[==========] Running 10 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 10 tests from Conflict/Conflict
[ RUN      ] Conflict/Conflict.ConflictWithFile/0
/tests/kdb/testkdb_conflict.cpp:75: Failure
Expected: first.set (firstReturned, parent) throws an exception of type KDBException.
  Actual: it throws nothing.
[  FAILED  ] Conflict/Conflict.ConflictWithFile/0, where GetParam() = true (18 ms)
[ RUN      ] Conflict/Conflict.ConflictWithFile/1
[       OK ] Conflict/Conflict.ConflictWithFile/1 (20 ms)
[ RUN      ] Conflict/Conflict.ConflictWithFileLoop/0
/tests/kdb/testkdb_conflict.cpp:103: Failure
Expected: first.set (firstReturned, parent) throws an exception of type KDBException.
  Actual: it throws nothing.
/tests/kdb/testkdb_conflict.cpp:103: Failure
Expected: first.set (firstReturned, parent) throws an exception of type KDBException.
  Actual: it throws nothing.
/tests/kdb/testkdb_conflict.cpp:103: Failure
Expected: first.set (firstReturned, parent) throws an exception of type KDBException.
  Actual: it throws nothing.
/tests/kdb/testkdb_conflict.cpp:103: Failure
Expected: first.set (firstReturned, parent) throws an exception of type KDBException.
  Actual: it throws nothing.
/tests/kdb/testkdb_conflict.cpp:103: Failure
Expected: first.set (firstReturned, parent) throws an exception of type KDBException.
  Actual: it throws nothing.
[  FAILED  ] Conflict/Conflict.ConflictWithFileLoop/0, where GetParam() = true (27 ms)
[ RUN      ] Conflict/Conflict.ConflictWithFileLoop/1
[       OK ] Conflict/Conflict.ConflictWithFileLoop/1 (23 ms)
[ RUN      ] Conflict/Conflict.ConflictWithFileSameKey/0
/tests/kdb/testkdb_conflict.cpp:126: Failure
Expected: first.set (firstReturned, parent) throws an exception of type KDBException.
  Actual: it throws nothing.
[...]

PR #629, opened 2016-04-26
Profile Plugin: Add dependency Libease

sanssecours:
Before this change the build system was not able to translate Elektra on OS X.

### DIFF
--- a/src/plugins/profile/CMakeLists.txt
+++ b/src/plugins/profile/CMakeLists.txt
@@ -4,6 +4,8 @@ add_plugin (profile
 	SOURCES
 		profile.h
 		profile.c
+	LINK_ELEKTRA
+		elektra-ease
 	)
 
 add_plugintest (profile)


markus2330:
Thank you!

@tom-wa seems like your linker also does not tell you if dependencies are missing.

markus2330:
@sanssecours can you take a look (check with Mac OS X) at ad537b36cb68722bb0f6a179fa726c02ed2be256? It should be the last breaking change of the release -- other changes should be fixes in the more narrow sense.

sanssecours:
> @sanssecours can you take a look (check with Mac OS X) at ad537b3?

Building – with `cmake .. -DENABLE_TESTING=ON -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5 -DTOOLS='qt-gui;kdb' -DBUILD_PDF=ON`  – works fine. The command `make test` currently reports two errors (see issues #613 and #628).

markus2330:
Thank you for checking again! Those two issues are not specific to Mac OS X and should be solved for the release.

PR #630, opened 2016-04-26
Fix profile.d script sourcing on some distros

sirblackheart:
Currently the elektraenv.sh was added to the /etc/profile.d/ folder
as kdb without .sh extension. Distributions like Arch Linux use *.sh
globing to source all files in the /etc/profile.d/ folder and
therefore do not read the kdb file without extension.
Adding sh when renaming the file on cmake install should fix this.


markus2330:
Thank you, sounds useful!

PR #632, opened 2016-04-26
Plugin Tests: Fix Silent Failure

sanssecours:
Before this change `ctest -R testscr_check_plugins` would fail silently
if the path of the build directory contained spaces.


markus2330:
Thank you, good catch!

PR #642, opened 2016-04-27
INI Plugin: Add Dependency Libmeta

sanssecours:
Before this change building of the plugin would fail on OS X.


markus2330:
Thank you!

PR #660, opened 2016-04-27
fix #603

### LINKED ISSUE #603: qt-gui fails to synchronize because of readonly plugins

0003088:
Sorry for the mess, I had to revert the button changes which ended in a little chaotic situation. Should be fine now though.

### DIFF
--- a/doc/NEWS.md
+++ b/doc/NEWS.md
@@ -42,7 +42,7 @@ See [http://libelektra.org](http://libelektra.org)
 - Resolver can also handle conflicts that happen
   within a single time tick (resolution of your clock)
   and also better handles NFS and older file systems
-- Overall many important fixes and improvements
+- Default storage and resolver can be changed by symlink.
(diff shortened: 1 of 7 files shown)


markus2330:
So it is a single line change after all? In github "Files changed" you can check if the overall outcome is as you expected.

markus2330:
Thank you!

PR #661, opened 2016-04-27
fix #523 again

### LINKED ISSUE #523: spec/conditionals: error message wrong
```
$ kdb rm -r user/example/highlevel/
$ kdb set /example/highlevel/device smartphone
$ kdb get /example/highlevel/powersaving/gps
Did not find key
$ kdb set /example/highlevel/powersaving/gps
Using name user/example/highlevel/powersaving/gps
Create a new key user/example/highlevel/powersaving/gps with null value
$ kdb get /example/highlevel/powersaving/gps
```

Only when I assign a non-null key, it will work:

```
$ kdb set /example/highlevel/powersaving/gps s
Using name user/example/highlevel/powersaving/gps
Set string to s
$ kdb get /example/highlevel/powersaving/gps
empty
```

tom-wa:
ok, that was a really stupid mistake...

### DIFF
--- a/src/libs/elektra/mount.c
+++ b/src/libs/elektra/mount.c
@@ -252,6 +252,8 @@ int elektraMountGlobals (KDB * kdb, KeySet * keys, KeySet * modules, Key * error
 		ksDel (keys);
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


markus2330:
We really need functionality checks for global plugins. I hope you are convinced now, too.

PR #662, opened 2016-04-27
Tests: Simplify Code in Resolver Check

sanssecours:
I did **not test** this change with `dash`! However the simplified code works fine in `sh` on OS X:

```sh
> TMPPATH=`cd /tmp; pwd -P`
> echo $TMPPATH
/private/tmp
```

### DIFF
--- a/tests/shell/check_resolver.sh
+++ b/tests/shell/check_resolver.sh
@@ -5,11 +5,7 @@ echo ELEKTRA CHECK RESOLVER
 echo
 
 #set tmp path (mainly for OS X compatibility)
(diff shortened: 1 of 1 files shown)


markus2330:
Looks nice, I am not sure how portable `-P` is, though.

markus2330:
Thank you!

sanssecours:
> Looks nice, I am not sure how portable `-P` is, though.

The switch is defined in [POSIX 1003.1](http://www.unix.com/man-page/posix/1posix/pwd/). So it should work in every POSIX compatible shell.

PR #669, opened 2016-04-28
Add install target for *.desktop file for qt-gui

sirblackheart:
This commit adds a desktop file for the elektra qt gui, install target
for it in cmake as well as install target for the icon. The desktop file
still misses i18n.


markus2330:
Thank you, great that you could do it for this release.

Issue #671, opened 2016-04-28
vfat support


markus2330:
Currently the temporary files created by the resolver are too long for vfat.

We could check the _PC_NAME_MAX/_PC_PATH_MAX dynamically using fpathconf, pathcon in the filename resolution of the resolver

markus2330:
Btw. as not-so-bad workaround you can use wresolver which works nicely. Only the known limitations apply (does not create a directory and does not remove config file on empty config).

markus2330:
Implementation Hint: you have to make sure that correct bits are removed when file name is make shorter.

markus2330:
Closed as irrelevant (use wresolver if you want vfat)

markus2330:
moved forgotten issue to #2531

PR #685, opened 2016-05-01
Resolver Plugin: Fix Compiler Warning on OS X

sanssecours:
If you know a better solution to fix this problem, then please let me know. Here is the warning message:

```
[126/925] Building C object src/plugins/resolver/CMakeFiles/elektra-resolver_fm_ub_x-objects.dir/filename.c.o
../src/plugins/resolver/filename.c:78:84: warning: format specifies type 'long' but the argument has type '__darwin_suseconds_t' (aka 'int') [-Wformat]
        snprintf (where + len, POSTFIX_SIZE - 1, ".%d:%ld.%ld.tmp", getpid (), tv.tv_sec, tv.tv_usec);
                                                          ~~~                             ^~~~~~~~~~
                                                          %d
/usr/include/secure/_stdio.h:57:62: note: expanded from macro 'snprintf'
  __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
                                                             ^~~~~~~~~~~
1 warning generated.
```

### DIFF
--- a/src/include/CMakeLists.txt
+++ b/src/include/CMakeLists.txt
@@ -24,6 +24,10 @@ check_type_size("long long"     SIZEOF_LONG_LONG)
 check_type_size("long double"   SIZEOF_LONG_DOUBLE)
 check_type_size(mode_t          SIZEOF_MODE_T)
 
+set(CMAKE_EXTRA_INCLUDE_FILES "sys/time.h")
+	check_type_size("((struct timeval*)0)->tv_usec" SIZEOF_TV_USEC)
+set(CMAKE_EXTRA_INCLUDE_FILES)
+
 set (BUILTIN_EXEC_FOLDER
 	"${CMAKE_INSTALL_PREFIX}/${TARGET_TOOL_EXEC_FOLDER}")
 set (BUILTIN_DATA_FOLDER
--- a/src/include/kdbconfig.h.in
+++ b/src/include/kdbconfig.h.in
@@ -82,6 +82,7 @@
 #define SIZEOF_INT               @SIZEOF_INT@
 #define SIZEOF_LONG              @SIZEOF_LONG@
 #define SIZEOF_MODE_T            @SIZEOF_MODE_T@
+#define SIZEOF_TV_USEC           @SIZEOF_TV_USEC@
 
 #cmakedefine HAVE_SIZEOF_LONG_LONG
 #ifdef HAVE_SIZEOF_LONG_LONG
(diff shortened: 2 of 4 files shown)


markus2330:
I also did not find a portable way, the cast seems to be necessary. Maybe someone else has an idea?

This code, however, is subject to be changed, see #671. (the `snprintf` generates too long file names to be written to vfat).

markus2330:
jenkins build please

markus2330:
jenkins build please

markus2330:
jenkins build please

sanssecours:
> The cast seems to be standard-way to do it. An alternative is to define a ELEKTRA_TIME_T which contains the necessary format specifier as string (see `src/include/kdbtypes.h` how its done for other types).

The code now uses `ELEKTRA_TIME_F` to store the format string. Is this what you wanted?

> Also note that `src/plugins/wresolver/wresolver.c:420` should be fixed in the same way, too. There I used the cast to long.

Should I just paste the preprocessor code from 4a96bcef there too? Or do you want me to put it into a common header file. If so, which header file would be appropriate?

markus2330:
Yes, please put it into kdbtypes.h and also use it for wresolver.

The name is also too generic, maybe change to `ELEKTRA_TIME_NSEC_F`?
And ideally you can fix it for `time_t`, it has the same problem on some platforms.

Btw. checking for size of the types would be better that checking for an OS.
systems that have `time_t` with 32bit signed now will change till year 2038.

Thanks!

sanssecours:
> Yes, please put it into kdbtypes.h and also use it for wresolver. The name is also too generic, maybe change to `ELEKTRA_TIME_NSEC_F`?
 
I changed the name for the format specifier of `tv_usec` to `ELEKTRA_TIME_USEC_F` and put it into `kdbtypes.h`. The code now checks the size of `((struct timeval*)0)->tv_usec` and not the type of the OS.

Since `wresolver` does not use the struct `timeval` but the struct `stat`, I did not change anything there. I do not think `stat.st_mtime`/`stat.st_mtimespec.tv_sec` (type `time_t` – 8 bit long on my system) and `timeval.tv_usec` (type `suseconds_t` – 4 bit on my system) are related. On my system I can remove the casts to `long` in line 420 of `wresolver.c` and the compiler does not complain about incorrect format specifiers.

markus2330:
Thank you! Removing the casts to long in wresolver would lead to warnings on Linux (time_t is 64bit there). So we would need the same technique implemented here to find the correct specifier.

PR #687, opened 2016-05-01
Build: Only Run Doxygen Once

sanssecours:
Before this change a parallel build started multiple instances of `doxygen` concurrently. Each instance of Doxygen invokes the Markdown Link Converter as filter program. Some of the Markdown Link Converter processes accessed and wrote the same files concurrently. This would cause some of them to fail, since each Markdown Link Converter assumes, that it has exclusive access to its input file.

I tested this both with `ninja` and `make -j 9`. Everything seems to work so far. Ninja builds work without the `add_dependencies` definitions added in this commit too. However, parallel invocations of `make` [do not](https://cmake.org/Bug/view.php?id=10082).

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -13,14 +13,6 @@ macro (do_doc target folder file install)
 	# sometimes doxygen is too slow and fails with "Could not create output directory .../doc/html"
 	file(MAKE_DIRECTORY ${folder})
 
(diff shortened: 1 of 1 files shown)


markus2330:
jenkins build please

markus2330:
add to whitelist

markus2330:
Thanks, nice that ninja works now!

PR #693, opened 2016-05-02
error to warning

tom-wa:


markus2330:
The `testscr_check_formatting` unittest finally works ;)

I think the resetting of an error between `kdbGet`/`kdbSet` calls is still missing. (And restoring of old errors if no new error happened).

tom-wa:
it's currently just turning every error into a warning after the first error is set. 
about the `kdbGet` / `kdbSet` error resetting part, i'm not really sure what you mean

markus2330:
Let us discuss this today. It should implement two rules:

1. *Within* an individual `kdbOpen`, `kdbGet` or `kdbSet`-calls only the first error will be used
2. *Between* individual `kdbOpen`, `kdbGet` or `kdbSet`-calls the error is only set iff the call is non-successful  (those returning -1)

Rule 2 means that that a new call with a new error overwrites the previous error *and also means* a new successful call keeps the previous error. (errno semantics)

This should also be documented in the API and kdb testcases should be added.

markus2330:
jenkins build please

markus2330:
any update here?

markus2330:
jenkins build please

markus2330:
jenkins build fast please

markus2330:
jenkins build fast please

markus2330:
jenkins build please

markus2330:
jenkins build stable please

markus2330:
jenkins build unstable please

markus2330:
any update here?

tom-wa:
I'm having a few issues getting the second part to work

markus2330:
Which are? You simply write guards at beginning/end of `kdb*`-functions and make sure that errors are only set according the API description and transform them to warnings otherwise.

tom-wa:
@markus2330 can you take a look at kdbSet here ? 
it basically just backs up the error, removes it if a new error could be set and -1 returned and restores it otherwise.

markus2330:
I did not look into the details, but yes it is something like this.

Can you add test cases for different errors scenarios in tests/kdb/testkdb_simple.cpp
(Btw. you can use the C API too, you only need `using namespace ckdb;`)
It needs to be in this file, because the text fixture Simple makes sure that a new mountpoint exists in every testcase for `testRoot`.

markus2330:
We need extensive tests here, let us do it for next release.

markus2330:
@mpranj Is it okay if we merge the PR for this release?

src/libs/elektra/kdb.c
```diff
@@ -633,6 +633,34 @@ static int elektraGetDoUpdateWithGlobalHooks (KDB * handle, Split * split, KeySe
 	}
 	return 0;
 }
+
+static int copyError (Key * dest, Key * src)
+{
+	keyRewindMeta (src);
+	Key * metaKey = (Key *)keyGetMeta (src, "error");
```
markus2330: avoid cast by using `const Key`

mpranj:
@markus2330 you mean this PR, or were you hinting at the hooks?
This one here can be merged!

markus2330:
Yes, I mean this PR.

@tom-wa Can you rebase, fix the small issue, and add some test cases?

tom-wa:
isn't that error actually expected on the second call to kdbSet ?
resolver (with elektraOpenFile) gets called before the error plugin and because kdbSet fails the new error should win

markus2330:
> isn't that error actually expected on the second call to kdbSet

No, from user perspective the last error should win (like working with errno). Only from plugin perspective the first error should win (and others should be attached as warning). So you need to clean the error in the beginning and restore the error on success (like errno is not touched on success).

Please also add some more test cases, it would be great to have this for 0.8.19.

markus2330:
Thanks, you are right about error 30, we have to investigate it. Seems we had a silent failure.

@tom-wa But there are still test cases missing. Could you please add them?

markus2330:
Thanks, your improved error handling actually led to a fix 4637e1f5fa9a410aeb65928ee5b74ff273020d94 of a quite serious issue (conflict of later kdbSet() against non-missing file if earlier kdbSet() had no file and is aborted).

Can you add some more test cases here? Maybe we find another issue ;)

markus2330:
jenkins build mergerequests please

markus2330:
jenkins build please

markus2330:
jenkins build debug please

markus2330:
Thank you!

PR #694, opened 2016-05-02
metaini to c array

tom-wa:
is there any use for that script ? 
it basically parses METADATA.ini und outputs c arrays. One is just the metakey names for storage plugins to check if the data should be written or not, one is metakey name + usedby/plugin, and one metakey + usedby/api


markus2330:
Thank you, yes, sounds useful.

Issue #656, opened 2016-04-27
OpenBSD: testscr_check_get_set fails


petermax2:
64/104 Test  #64: testscr_check_get_set ..............***Failed    5.50 sec

	ELEKTRA BASIC COMMAND SCRIPTS TESTS

	Check if script tests the correct version
	ccode not a read-write storage
	conditionals not a read-write storage
	constants not a read-write storage
	counter not a read-write storage
	csvstorage not a read-write storage
	curlget not a read-write storage
	dpkg not a read-write storage

	-- Testing dump
	find: -quit: unknown option
	find: -quit: unknown option
	testing ls command
	testing rm command
	Did not find keyDid not find keytesting array
	/root/Projekte/libelektra/build/tests/shell/testscr_check_get_set.sh: seq: not found
	testing ls command
	testing rm command
	Did not find keyDid not find keytesting array
	/root/Projekte/libelektra/build/tests/shell/testscr_check_get_set.sh: seq: not found
	enum not a read-write storage
	error not a read-write storage
	filecheck not a read-write storage
	glob not a read-write storage
	hexcode not a read-write storage
	hidden not a read-write storage
	hosts not a read-write storage
	iconv not a read-write storage

	-- Testing ini
	find: -quit: unknown option
	find: -quit: unknown option
	testing ls command
	testing rm command
	Did not find keyDid not find keytesting array
	/root/Projekte/libelektra/build/tests/shell/testscr_check_get_set.sh: seq: not found
	testing ls command
	testing rm command
	Did not find keyDid not find keytesting array
	/root/Projekte/libelektra/build/tests/shell/testscr_check_get_set.sh: seq: not found
	iterate not a read-write storage
	keytometa not a read-write storage
	line not a read-write storage
	lineendings not a read-write storage
	list not a read-write storage
	logchange not a read-write storage
	mathcheck not a read-write storage
	network not a read-write storage

	-- Testing ni
	find: -quit: unknown option
	find: -quit: unknown option
	testing ls command
	testing rm command
	Did not find keyDid not find keytesting array
	/root/Projekte/libelektra/build/tests/shell/testscr_check_get_set.sh: seq: not found
	testing ls command
	testing rm command
	Did not find keyDid not find keytesting array
	/root/Projekte/libelektra/build/tests/shell/testscr_check_get_set.sh: seq: not found
	noresolver not a read-write storage
	null not a read-write storage
	path not a read-write storage
	regexstore not a read-write storage
[...]

PR #696, opened 2016-05-05
fix libsplit benchmark: POSIX compatibility

### LINKED ISSUE #530: libsplit benchmark script is bash dependent
The benchmark script introduced in 15657e4d seems to be dependent on bash. When executed with the shebang provided it throws the following error on my development system (Debian 7.9):

```libelektra/scripts/benchmark_libsplit.sh: 18: libelektra/scripts/benchmark_libsplit.sh: Syntax error: Missing '))'```

The issue seems to be rooted in the line ```WALLTIME=$((time -p $BENCHMARK) 2>&1)```.
Can anyone confirm this issue?

### PR DESCRIPTION
**Pull request by mpranj**
Should fix #530 .

Tested on FreeBSD which is:
> close to the IEEE	Std 1003.1 (``POSIX.1'')

Prints just to stdout now, instead of creating logfiles which I have used to evaluate the results in R.

### DIFF
--- a/scripts/benchmark_libsplit.sh
+++ b/scripts/benchmark_libsplit.sh
@@ -6,13 +6,13 @@ fi
 
 BENCHMARK=./build/bin/large
 CURRTIME=`date +"%H-%M-%S_%m-%d-%Y"`
-LOGFILE="./doc/benchmark/logs/$(echo $KDB)_log_$CURRTIME.log"
+#LOGFILE="./doc/benchmark/logs/$(echo $KDB)_log_$CURRTIME.log"
+RUNS=21
 
-
-echo "Runtime as IEEE Std 1003.2-1992 (``POSIX.2'')" >> $LOGFILE
-for run in {0..20}; do
-	echo "RUN: #$run"
-	WALLTIME=$((time -p $BENCHMARK) 2>&1)
-	echo $WALLTIME | awk -F"real" '{print $1;}' # print programs stdout
-	echo $WALLTIME | awk -F"real" '{print $2;}' | awk '{print $1;}' >> $LOGFILE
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/benchmark_libsplit.sh [hunk 1]
```diff
@@ -6,13 +6,13 @@ fi
...
-
-echo "Runtime as IEEE Std 1003.2-1992 (``POSIX.2'')" >> $LOGFILE
-for run in {0..20}; do
-	echo "RUN: #$run"
-	WALLTIME=$((time -p $BENCHMARK) 2>&1)
-	echo $WALLTIME | awk -F"real" '{print $1;}' # print programs stdout
-	echo $WALLTIME | awk -F"real" '{print $2;}' | awk '{print $1;}' >> $LOGFILE
+echo "Runtime as IEEE Std 1003.2-1992 (``POSIX.2'')"
+echo "Doing $RUNS runs:"
+for run in $(seq 1 $RUNS); do
```
markus2330: seq seems to not exist on OpenBSD #656 

So POSIX would be:

1. count manually: `i=$((i+1));`
2. use awk to generate the seq `awk "BEGIN{for (i = 1; i <= $RUNS; i ++) print i}"`

Maybe you can fix #656, too?

scripts/benchmark_libsplit.sh
```diff
@@ -11,8 +11,10 @@ RUNS=21
 
 echo "Runtime as IEEE Std 1003.2-1992 (``POSIX.2'')"
 echo "Doing $RUNS runs:"
-for run in $(seq 1 $RUNS); do
+i=0
+while [ $i -lt $RUNS ] ; do
```
mpranj: I'm not sure how this behaves on OpenBSD. It works on FreeBSD but it appears from the man-pages that they have different `/bin/sh` implementations.

PR #697, opened 2016-05-05
remove seq, blind fix for #656

mpranj:
should fix at least the seq problem from #656


mpranj:
@nd2s tested it on OpenBSD, seems to work now

markus2330:
Thank you!

Issue #689, opened 2016-05-02
global caching plugin


markus2330:
As discussed in last meeting with @Namoshek it is surprising that `kdbGet` returns more than requested and it does not return anything when called again with an empty keyset.

An internal cache in a global plugin could fix this unexpected behaviour. It should cache the whole received configuration and do the proper `ksCut` so that the user exactly receives what was requested.

PR #706, opened 2016-05-08
Global cache filter plugin (issue #689)

Namoshek:
I started writing on a global cache plugin that filters unwanted keys and caches them until they are added to the set of keys to write-out during `kdbSet()`. Suggestions for further tests (and how to write them) are highly appreciated.


markus2330:
add to whitelist

markus2330:
jenkins build please

src/plugins/cachefilter/README.md
```diff
@@ -0,0 +1,16 @@
+- infos = Information about the cachefilter plugin is in keys below
+- infos/author = Marvin Mall <e1225943@student.tuwien.ac.at>
```
markus2330: Do you really want to use this e-mail address?

src/plugins/cachefilter/README.md
```diff
@@ -0,0 +1,16 @@
...
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
```
markus2330: should provide filter

src/plugins/cachefilter/README.md
```diff
@@ -0,0 +1,16 @@
...
+- infos/recommends =
+- infos/placements =
```
markus2330: Needs some placement in get/set direction

It should be a position late in kdbGet(), but early in kdbSet()

src/plugins/cachefilter/README.md
```diff
@@ -0,0 +1,16 @@
...
+- infos/status = experimental global
```
markus2330: Various status information missing, please look through doc/CONTRACT.ini and add everything that fits.

src/plugins/cachefilter/README.md
```diff
@@ -0,0 +1,16 @@
...
+- infos/metadata =
+- infos/description = A global plugin that steps in during kdbGet() process to filter the results in a way, so that no other keys than the requested one or descendants of it are returned. During kdbSet() the filtered keys are added back to the output, so that they don't get lost during the storage process. In other words, the plugin caches filtered keys so that they do not get lost by accident.
```
markus2330: Instead of "lost by accident", e.g. eases the usage of the API.

src/plugins/cachefilter/cachefilter.c
```diff
@@ -0,0 +1,73 @@
...
+			keyNew ("system/elektra/modules/cachefilter/infos/version", KEY_VALUE, PLUGINVERSION, KEY_END), KS_END);
+		ksAppend (returned, contract);
+		ksDel (contract);
+
+		return 1; // success
+	}
+
+	// filter keys that need to be cached
+	KeySet * temp = 0;
+	KeySet * toBeCached = ksNew (ksGetSize (returned), KS_END);
```
markus2330: What if a kdbGet() was executed before? toBeCached should be fetched from elektraPluginGetData.

src/plugins/cachefilter/README.md
Namoshek: I'll not use my personal email for this, that's for sure. My alias shall not be connected to my email in public.

src/plugins/cachefilter/cachefilter.c
```diff
@@ -0,0 +1,73 @@
...
+
+	// cache the filtered keys
+	elektraPluginSetData (handle, toBeCached);
+
+	return 1; // success
+}
+
+int elektraCachefilterSet (Plugin * handle, KeySet * returned, Key * parentKey ELEKTRA_UNUSED)
+{
+	// get cached keys (duplicate in case of kdbSet() error!)
```
markus2330: kdbSet() already does an duplication of returned, so you are only working on a copy.

src/plugins/cachefilter/cachefilter.c
```diff
@@ -0,0 +1,73 @@
...
+	KeySet * cachedKeys = ksDup ((KeySet *)elektraPluginGetData (handle));
```
markus2330: Do you really need to duplicate? ksAppend already takes care of incrementing the ref counter.

src/plugins/cachefilter/testmod_cachefilter.c
```diff
@@ -0,0 +1,94 @@
...
+}
+
+static KeySet * createTestKeysToNotCache ()
+{
+	return ksNew (3, keyNew ("user/tests/cachefilter/will/not/be/cached/key1", KEY_VALUE, "not-cached1", KEY_END),
+		      keyNew ("user/tests/cachefilter/will/not/be/cached/key2", KEY_VALUE, "not-cached2", KEY_END),
+		      keyNew ("user/tests/cachefilter/will/not/be/cached", KEY_VALUE, "not-cached", KEY_END), KS_END);
+}
+
+static void compareKeySets (KeySet * ks, KeySet * expected)
```
markus2330: compare_keyset already exists in tests.h

src/plugins/cachefilter/testmod_cachefilter.c
```diff
@@ -0,0 +1,94 @@
...
+
+
+int main (int argc, char ** argv)
+{
+	printf ("CACHEFILTER     TESTS\n");
+	printf ("==================\n\n");
+
+	init (argc, argv);
+
+	test_successfulCache ();
```
markus2330: testcases with longer kdbGet()/kdbSet() sequences are missing.

src/plugins/cachefilter/README.md
markus2330: Yes of course, but the student e-mail address will vanish after you finished. What about a `@libelektra.org` address where you can choose to which address the e-mails go to? Which address do you want?

markus2330:
Looks nice, thank you!

src/plugins/cachefilter/README.md
Namoshek: In this case I'll take `namoshek@libelektra.org`.

src/plugins/cachefilter/cachefilter.c
Namoshek: If I don't duplicate the `KeySet`, I would add more `Keys` to the cache with the next line `ksAppend (cachedKeys, returned);`, what is not what I want.

src/plugins/cachefilter/cachefilter.c
Namoshek: Is it guaranteed that `elektraPluginGetData(..)` returns NULL in case that no data has been stored before? From what I know, C does not guarantee that memory allocated dynamically for a struct will be zeroed during initialization. `elektraPluginExport` does this: `returned = elektraCalloc (sizeof (struct _Plugin));`, what looks to me like I cannot expect the plugin data to be NULL, but that also means I cannot cast it to a `KeySet` without running into problems...?

markus2330: Yes, calloc is used and it would not be a good idea to return uninialized memory, so its a guarantee. I updated the docu in 5b0c21f9369404ee5eedf67ddd719db276991759

src/plugins/cachefilter/cachefilter.c
KurtMi: Its a good practice to check the return value of `elektraPluginGetData`. Since only a correct working behaviour does not produce a segfault.

KurtMi: Duplicating is in my point of view also not needed. For ever application accessing the Key Database an instance of this plugin will be created and every instance has this own memory which you can access with `elektraPlugin{Get,Set}Data`. Therefore it should be enough to retrieve the data from the plugin and join it with the `returned` KeySet.

Namoshek: You are right, I was thinking too complex. As both, the cache and the returned KeySet, only store pointers, it shouldn't matter in which order I append them (returned to cachedKey or the other way round). So I'll just swap the logic and remove the `ksDup()`.

markus2330:
jenkins build please

markus2330:
jenkins build please

markus2330:
test from jenkins

markus2330:
test from jenkins

markus2330:
jenkins build unstable please

markus2330:
@KurtMi please review (after exam)

markus2330:
@mpranj Could you check if this plugin works with global plugins?

KurtMi:
I reviewed the code and logic, everything looks fine. Btw I will also try it as global plugin.

markus2330:
Thank you!

KurtMi:
I thought a little bit and I think it will not work in all cases, the Set should be called every time, which is not the case with the actual hooks.
While Testing I altered a example and made a `kdbGet` with no `kdbSet` afterwards, but the cutted key where there and `elektraCachefilterSet` was not called neither. Am I missing something?

Code:
```
KeySet * myConfig = ksNew (0, KS_END);
Key * key = keyNew ("system/test/asd", KEY_END);

KDB * handle = kdbOpen (key);

kdbGet (handle, myConfig, key);

Key * result = ksLookupByName (myConfig, "/test/asd/a", 0);
const char * key_name = keyName (result);
const char * key_value = keyString (result);
printf ("key: %s value: %s\n", key_name, key_value);

//~ kdbSet (handle, myConfig, key);
// maybe you want kdbSet() myConfig here

kdbClose (handle, 0); // no more affairs with the key database.
```
The Database:
```
>kdb export /test ini
[asb]
a = k
b = k
[asd]
a = s
b = s
```

Namoshek:
The plugin will cache everything from [consecutive] calls to kdbGet that is not part of the requested key tree. But if you do not call kdbSet, you do also not alter anything in the database at all, meaning that the plugin will not have to do its job, namely add cached keys to the keys to write to the database. This is no problem, but intended behavior. Think of it as a storage plugin that is only used for Get (in this case); there is also no problem when not calling Set afterwards, as long as you don't want to save (override, add or delete) anything.

KurtMi:
Ok then. My fault.

KurtMi:
I rethought my last thought and tested the plugin, it works as global plugin!

markus2330:
Thank you then both!

Issue #730, opened 2016-05-17
Speed up PR builds by only building/testing changed module


manuelm:
As explained here https://github.com/ElektraInitiative/libelektra/issues/160#issuecomment-219752143 we need a script which inspects the changed paths and dynamically builds the BINDINGS + PLUGINS variable.

To better integrate the script into the different build environments the script shouldn't start the configuration phase itself. Instead it should just output the configuration defines.

e.g. `scripts/magic.sh $TRAVIS_COMMIT_RANGE`
will output: `-DBINDINGS="..." -DPLUGINS="..."`

Regarding the bindings we can make things a little easier:
- touching anything in bindings/swig rebuilds swig_python+swig_python2+swig_lua
- touching anything in bindings/gi or bindings/glib rebuilds glib+gi_python+gi_lua

To make thinks more complicated again:
- touching src/plugins/python should also trigger src/plugins/python2

manuelm:
Scratch the independent script. Please make it a bash function. We can just source the file and call the function.

markus2330:
Sorry, the issue has to wait, others are more important.

markus2330:
A related problem is to know when then build directory needs to be discarded (needed for fast+incremental build jobs). It seems like adding errors/warnings is not safe, cmake does not detect it now, maybe because of #857. I would say, if a CMake file was changed, the script should purge the build directory.

(Btw. the problem is related because the script discussed here -- that only builds plugins as needed -- would be useful for fast+incremental, too)

markus2330:
I think the fast build job now gives timely feedback, and for the others it takes a while anyway. Please reopen if you think the issue is still relevant.

markus2330:
For the plugins and tools it would be quite straight-forward to realize: If src/plugins/<PLUGIN> or src/tools/<TOOL> has modifications, the shell function would set PLUGINS and TOOLS accordingly.

For bindings it seems to be a bit problematic, because there is no convention of how the folders to map to BINDINGS. What about restructuring the bindings to keep the function more simple?

> touching src/plugins/python should also trigger src/plugins/python2

Maybe the `-MM` option helps to find such deps:
```
> cd src/plugins/python2
> gcc -MM -DSWIG_TYPE_TABLE "-DSWIG_RUNTIME=<stdio.h>" "-DELEKTRA_README=<stdio.h>"  python.cpp
python.o: python.cpp ../python/python.cpp ../python/python.hpp
```

[...]

PR #742, opened 2016-05-22
WIP: Travis CI

mpranj:
Travis builds for OS X, thanks @manuelm a lot!
I opened the PR is just to confirm how many things I broke on linux builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,57 @@
+#
+# Turn off email notifications entirely
+#
+notifications:
+  email: false
+
+language: generic
+
+#
+# Define the build matrix
+#
+matrix:
+  include:
+#    - services: docker
+#      env: OS_RELEASE=debian_unstable
+#      sudo: required
+#      script:
+#        - docker-compose -f scripts/docker/docker-compose.yml build $OS_RELEASE
+#        - docker-compose -f scripts/docker/docker-compose.yml run $OS_RELEASE /bin/bash /workspace/code/scripts/travis.sh
+    - os: osx
+
+#
+# Some of the OS X images don't have cmake, contrary to what people
+# on the Internet say
+#
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      brew update &&
+      which cmake || brew install cmake &&
+      which swig || brew install swig &&
+      which python3 || brew install python3 &&
+      which lua || brew install lua &&
+      brew install gobject-introspection &&
+      brew install yajl &&
+      brew install augeas;
+    fi
+
+#
+# Source is checked out in $TRAVIS_BUILD_DIR
+# Build directory will be TRAVIS_BUILD_DIR/../build
+#
+before_script:
+  - cd $TRAVIS_BUILD_DIR/..
+  - mkdir build && cd build
+  - CMAKE_OPT=()
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      libffi_ver=$(brew ls --versions libffi | awk '{print $2}') &&
+      cp -nv /usr/local/Cellar/libffi/${libffi_ver}/lib/pkgconfig/libffi.pc /usr/local/lib/pkgconfig/ &&
+      python3_ver=$(python3 -c 'import sys; print(".".join(map(str, sys.version_info[:2])))') &&
+      CMAKE_OPT+=("-DPYTHON_INCLUDE_DIR:PATH=$(python3-config --prefix)/include/python${python3_ver}m") &&
+      CMAKE_OPT+=("-DPYTHON_LIBRARY:FILEPATH=$(python3-config --prefix)/lib/libpython${python3_ver}.dylib");
+    fi
+  - cmake -DBUILD_STATIC=OFF -DBINDINGS=ALL -DPLUGINS="ALL;-crypto_openssl" -DCMAKE_INSTALL_PREFIX=$TRAVIS_BUILD_DIR/../install -DKDB_DB_SYSTEM=$TRAVIS_BUILD_DIR/../kdbsystem ${CMAKE_OPT[@]} $TRAVIS_BUILD_DIR
+
+script:
+  - make -j3
+  - make -j3 test
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,23 +18,39 @@ matrix:
 #        - docker-compose -f scripts/docker/docker-compose.yml build $OS_RELEASE
 #        - docker-compose -f scripts/docker/docker-compose.yml run $OS_RELEASE /bin/bash /workspace/code/scripts/travis.sh
     - os: osx
+      osx_image: xcode7.3
+    - os: osx
+      osx_image: xcode7
+    - os: osx
+      osx_image: beta-xcode6.1
 
-#
-# Some of the OS X images don't have cmake, contrary to what people
-# on the Internet say
-#
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      brew update &&
-      which cmake || brew install cmake &&
-      which swig || brew install swig &&
-      which python3 || brew install python3 &&
-      which lua || brew install lua &&
-      brew install gobject-introspection &&
-      brew install yajl &&
+      export PATH=/usr/local/bin:/usr/local/sbin:$PATH;
+      brew update;
+      brew tap caskroom/cask;
+      brew install cmake;
+      brew install swig;
+      brew install gobject-introspection;
+      brew install python3;
+      brew install python;
+      brew install pygobject3 --with-python --with-python3;
+      brew install lua;
+      brew install yajl;
       brew install augeas;
+      brew install dbus;
+      brew install Caskroom/cask/java;
+      brew outdated libffi || brew install libffi;
+      libffi_ver=$(brew ls --versions libffi | awk '{print $2}') &&
+      gi_ver=$(brew ls --versions gobject-introspection | awk '{print $2}') &&
+      PKG_CONFIG_PATH=/usr/local/Cellar/libffi/${libffi_ver}/lib/pkgconfig/ &&
+      PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/local/Cellar/gobject-introspection/${gi_ver}/lib/pkgconfig &&
+      luarocks install lgi;
+      export JAVA_HOME=$(/usr/libexec/java_home -v 1.8);
+      java -version;
+      echo $JAVA_HOME;
+      which luarocks;
     fi
-
 #
 # Source is checked out in $TRAVIS_BUILD_DIR
 # Build directory will be TRAVIS_BUILD_DIR/../build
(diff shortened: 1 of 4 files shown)


mpranj:
jenkins build bindings please

mpranj:
it appears to me that 5e2adad6165e2b80216482b4786014b1d2b80b18 fixed java bindings tests on OS X ... 
Here's a log from travis:
```
        Start  63: testmod_jni
 63/128 Test  #63: testmod_jni ........................   Passed    0.66 sec
```

.travis.yml
```diff
@@ -0,0 +1,64 @@
+#
+# Turn off email notifications entirely
+#
+notifications:
+  email: false
```
markus2330: Why not sending notifications? Jenkins does this quite well (only mails when the build breaks)

If only one e-mail address is allowed we could use some forward e-mail address, e.g. travis@libelektra.org

.travis.yml
```diff
@@ -0,0 +1,64 @@
...
+  include:
+#    - services: docker
+#      env: OS_RELEASE=debian_unstable
+#      sudo: required
+#      script:
+#        - docker-compose -f scripts/docker/docker-compose.yml build $OS_RELEASE
+#        - docker-compose -f scripts/docker/docker-compose.yml run $OS_RELEASE /bin/bash /workspace/code/scripts/travis.sh
+    - os: osx
+
+osx_image: xcode7.3 # use latest OS X
```
markus2330: Does everyone has the latest OS X? Are we interested in building for older OS X, too?

.travis.yml
```diff
@@ -0,0 +1,64 @@
...
+      which python3 || brew install python3 &&
+      which lua || brew install lua &&
+      brew install gobject-introspection &&
+      brew install yajl &&
+      brew install augeas &&
+      brew cask install java;
+    fi
+#
+# Source is checked out in $TRAVIS_BUILD_DIR
+# Build directory will be TRAVIS_BUILD_DIR/../build
```
markus2330: I would prefer if you assign some variables for this, maybe some ROOT_DIR, BUILD_DIR, SOURCE_DIR and so on.

scripts/docker/Dockerfile.debian_unstable
```diff
@@ -0,0 +1,2 @@
+FROM manuelm/libelektra_build:debian_unstable
```
markus2330: Can we copy this to ElektraInitiative/libelektra_build:debian_unstable?

src/bindings/jna/CMakeLists.txt
```diff
@@ -27,6 +27,21 @@ if (Java_VERSION_MAJOR GREATER 0 AND Java_VERSION_MINOR GREATER 7)
 		NAMES jna
 		PATHS /usr/share/java
 		)
+	
+	if(NOT jna) # try fetching it, remember to update SHA1 if you update the jar
+		file(DOWNLOAD 
+			"https://maven.java.net/content/repositories/releases/net/java/dev/jna/jna/4.2.2/jna-4.2.2.jar"
```
markus2330: Is this jar then bundled with elektra.jar? Or how does the installed version work then?

src/bindings/jna/elektra/Elektra.java
```diff
@@ -11,7 +11,7 @@
 	 * Overall native binding to the elektra library
 	 */
 	Elektra INSTANCE = (Elektra)
-		Native.loadLibrary(("libelektra.so"), Elektra.class);
+		Native.loadLibrary(("elektra"), Elektra.class);
```
markus2330: Good fix, seems to work with Linux, too.

markus2330:
Nice work by both of you!

But the travis build itself fails ;( Should we exclude the py/lua bindings or can someone fix the LD_LIBRARY_PATH setting for these unit tests?

And it seems like no debian build happens?

manuelm:
@markus2330 As I said I'll take over the travis work as soon as the OS X build works. Don't - I repeat - don't merge any travis commits.

.travis.yml
manuelm: Because it's annoying while setting up travis..

.travis.yml
manuelm: No

scripts/docker/Dockerfile.debian_unstable
manuelm: This won't stay...

src/bindings/jna/CMakeLists.txt
manuelm: @mpranj Why aren't you fetching the jar in .travis.xml/before_install and just move it to /usr/share/java there?

manuelm:
Yeah, looks like bindings are still broken.
And crypto_openssl is still excluded? Any chance to get this working?

markus2330:
I think it is ok to exclude crypto_* for now: It has currently no functionality exposed to users and it is also tagged as experimental.

Btw. did you try to install elektra and exec `kdb run_all`? (Like in jenkins elektra-local-installation) Maybe we can this even do as root installed directly into system pathes as the container is recreated every time anyway? Then the result of the travis build would be a container where the latest Elektra is installed.

manuelm:
Travis is unable to find the openssl dev headers. That's why I've excluded the plugin. Most probably this is just a missing path someone with a real OS X machine can fix quickly.

@markus2330 Can we please only discuss OSX related things here. And stop thinking about new things after everything (incl #730) is set up. There are multiple reasons why this is a bad idea.

markus2330:
The installation and `kdb run` is related to our current problem (in Mac OS X): If we get the lua/python tests running more easily in the installed system, we could exclude them when running `make test` in the build directory.

On jenkins it was adding about two lines, `make install` and then `kdb run_all`. So it might be the easiest solution for this problem and additionally makes the tests more realistic. If it is not such a simple change, we should of course not add it in this PR.

markus2330:
Btw. build tests should be after the installed tests because some build tests require an installed Elektra (external+context tests).

.travis.yml
mpranj: Sure we can build on all available versions on travis once we get it to work at least on the current version.

src/bindings/jna/CMakeLists.txt
mpranj: Isn't the jar bundled with elektra later such that the original jna.jar can be thrown away?
Otherwise yes, I can move it somewhere. I was trying not to magically move files into system directories just for `make` to work.

markus2330: It is not bundled at the moment, but I think it makes sense because as far as I know the java world wants a jar to be self-contained?

mpranj: I have no idea how they do it. The "clean" way I know is using maven but that doesn't seem to be a (clean) option here.

markus2330: It is weird, https://cmake.org/cmake/help/v3.0/module/UseJava.html says that "For backwards compatibility, jar files listed as sources are ignored.", but it does not say anything about how to include them. INCLUDE_JARS only seems to set the include path, but not actually include them.

Actually we would only need to remove the line 451 of /usr/share/cmake-3.0/Modules/UseJava.cmake which does this strange ignoring of jar because of "backward compatibility".

src/plugins/crypto/CMakeLists.txt
```diff
@@ -5,83 +5,33 @@ include (LibAddPlugin)
...
+endif(APPLE)
 
-		try_compile (HAS_OPENSSL_4SURE
-			"${CMAKE_BINARY_DIR}"
-			"${PROJECT_SOURCE_DIR}/src/plugins/crypto/compile_openssl.c"
-			CMAKE_FLAGS
-				-DINCLUDE_DIRECTORIES:STRING=${OPENSSL_INCLUDE_DIR}
-				-DLINK_LIBRARIES:PATH=${OPENSSL_LIBRARIES}
-		)
+find_package(OpenSSL)
```
mpranj: I had to find this outside of `DEPENDENCY_PHASE` otherwise `OPENSSL_FOUND` etc. get reset.

markus2330: It does not matter if the options are reset, the add_plugin remembers the arguments it got.

find_package should only be used in the DEPENDENCY_PHASE.

I think the OPENSSL_MANUAL_OSX_HACK already does what you are trying to achieve.

And as said: the crypto plugin has no priority for this PR, its better if you revert these changes.

mpranj: Sure.

mpranj:
@markus2330 @manuelm just a quick update: `dlopen()` works now (FreeBSD/OS X). @markus2330 had almost done it in da243f9e25d8fa14f8286c48b4338a73c1e7242d but the path was wrong.

Also, I got it to build successfully etc. but it's not ready yet for @manuelm to take over, I will ping him later for that.

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+  - mkdir build && cd build
+  - CMAKE_OPT=()
+  - echo $PATH
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      python2_ver=$(python2 -c 'import sys; print(".".join(map(str, sys.version_info[:2])))') &&
+      CMAKE_OPT+=("-DPYTHON2_INCLUDE_DIR:PATH=$(python2-config --prefix)/include/python${python2_ver}") &&
+      CMAKE_OPT+=("-DPYTHON2_LIBRARY:FILEPATH=$(python2-config --prefix)/lib/libpython${python2_ver}.dylib") &&
+      python3_ver=$(python3 -c 'import sys; print(".".join(map(str, sys.version_info[:2])))') &&
+      CMAKE_OPT+=("-DPYTHON_INCLUDE_DIR:PATH=$(python3-config --prefix)/include/python${python3_ver}m") &&
+      CMAKE_OPT+=("-DPYTHON_LIBRARY:FILEPATH=$(python3-config --prefix)/lib/libpython${python3_ver}.dylib");
```
mpranj: While this does fix the builds on travis I would suggest/like to solve the same with cmake. It would make building on OS X a litte less painful.

The nice thing with travis is: we now have a raw description of the build dependencies which we could document in `doc/COMPILE.md` or somewhere.

manuelm: I really don't like putting platform specific paths into *our* build system. If cmake is unable to find it, try to fix this at cmakes side or leave it. Otherwise you'll have to constantly keep up with platform specifics of other people, which fills our build system with obscure paths nobody can verify or clean up. Heck, right now cmake doesn't even find python3 on debian jessie or wheezy. And nobody cares about that, unless you're building this on our own. And even if you build elektra on your own, you usually pull the dependencies and configure line from the package files of your distribution and be done.

Beside that nobody stops a user from downloading python and installing into different paths. It just happens that brew is the easiest way on OSX so most users are using this packager right now. Might be another packager and other paths a few years later.

So instead, I suggest we do the same thing like on Linux: You (or a OSX maintainer) just write a definition file for brew in which you can put the special brew specific paths.

I have the same opinion regarding the automatic jna jar download.

mpranj: I do not want to put OS X specific paths all over the build system.
I do however think that a python user will have python/python2/python3 somewhere in their PATH.

Using python-config works fine for me on FreeBSD and OS X:
```
mpranj@beastie ~ % python2-config --includes
-I/usr/local/include/python2.7 -I/usr/local/include/python2.7
mpranj@beastie ~ % python2-config --libs
-lpython2.7 -L/usr/local/lib -lintl -lutil -lm
mpranj@beastie ~ % python2-config --cflags
-I/usr/local/include/python2.7 -I/usr/local/include/python2.7 -fno-strict-aliasing -O2 -pipe -fstack-protector -fno-strict-aliasing -DNDEBUG
mpranj@beastie ~ % python2-config --ldflags
-L/usr/local/lib -lpython2.7 -L/usr/local/lib -lintl -lutil -lm -Wl,--export-dynamic
```

manuelm: Than please fix it on cmakes side: https://cmake.org/Bug/bug_relationship_graph.php?bug_id=9897
However I do not know why cmake doesn't call this in the first place. Might be related to cross-compiling.

mpranj:
I think this is failing only because of the deprecated github status updater, which I am currently removing from all jobs where it might be still in use.

This should be ready for @manuelm to implement the linux builds.

It solely concentrates on building for OS X on travis and introduces **no** fixes. For instance, I removed the automatic JNA download. It's a different matter. CMake doesn't even find JNA on FreeBSD where there is a nice port available. If you want us to take a look at this and make it work on more platforms, I will open a new issue.

markus2330:
Yes, elektra-mergerequests-unstable seems to fail only because of the Github Status setter.

So should I merge it? It looks great: it seems like everything finally works on Mac OS X. Or are there also leftovers for Mac OS X?

Please create an issue for every known problem, it does not mean you have to work for it. When someone needs JNA on FreeBSD we can fix it.

mpranj:
I removed the Github status setter. Let's see if it works well now.

No, @manuelm explicitly wanted us **not to merge**, but to take over the work and finish it. I made the PR just so it could be reviewed.

Other than that: yes I'd like to see this in master asap so we can detect regressions on OS X automatically.

manuelm:
> So should I merge it? 

Once again: no!

@mpranj 
Will take at look next (or in two weeks).

I see not much sense in testing multiple xcode variants. Apple supports the latest only. And apart from different clang versions we don't use any SDK shipped with xcode. All it take is more time to build.

mpranj:
@manuelm the different xcode variants are basically different VMs/different OS X versions. The three tested versions are still supported/get security updates and are in use. Not everyone has the latest OS X.

It makes sense because apple tends to change lots of things. One example: they introduced SIP https://en.wikipedia.org/wiki/System_Integrity_Protection which doesn't allow us to write to just any system directory. (also different library versions, OS X not using openssl any more, ..)
It might be interesting for the future if things get broken with OS X upgrades and it doesn't cost us anything right now. But you can remove it as you wish.

markus2330:
I am also strongly in favour of testing all variants, it often gives insightful clues in which situation something breaks. We can always decide to drop support for some old system if it is hard to fix something (and it is obviously the systems fault), but I would not exclude it only to reduce build time.

It would be very useful to have Mac OS X builds now, and not to wait 1-2 weeks for some other feature which is not urgent. It is no problem if you completely rewrite the travis file later (if needed).

manuelm:
Then for gods sake merge it. The .travis.xml only. Rest is not needed.

And yes, I'll definitely rewrite the whole damn thing.

markus2330:
It does not really make sense to say someone should not merge but its okay to cherry-pick exactly the same commits?

I only added the travis file, as asked for.

We should not invest too much time in build-stuff but rather invest time into the code, if a buildjob works it is good enough.

PR #744, opened 2016-05-22
Fix error on empty LUABIN_VERSION_STRING

sirblackheart:
Cmake has a configuration error on an empty LUABIN_VERSION_STRING in
both the SUBSTRING and COMPARE function. So lets asume we have a version
mismatch on an empty LUABIN_VERSION_STRING to fix this.

### DIFF
--- a/cmake/Modules/FindLua.cmake
+++ b/cmake/Modules/FindLua.cmake
@@ -114,9 +114,12 @@ function (verify_lua_executable_version)
 	)
 	
 	# Compare only MAJOR.MINOR
-	string (SUBSTRING ${LUABIN_VERSION_STRING} "4" "3" LUABIN_VERSION_STRING)
-	string (COMPARE EQUAL ${LUABIN_VERSION_STRING} "${_LUA_VERSION_MAJOR}.${_LUA_VERSION_MINOR}" VERSION_MATCHES)
-	
+	if (NOT LUABIN_VERSION_STRING STREQUAL "")
+		string (SUBSTRING ${LUABIN_VERSION_STRING} "4" "3" LUABIN_VERSION_STRING)
+		string (COMPARE EQUAL ${LUABIN_VERSION_STRING} "${_LUA_VERSION_MAJOR}.${_LUA_VERSION_MINOR}" VERSION_MATCHES)
+	else()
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindLua.cmake
```diff
@@ -114,9 +114,12 @@ function (verify_lua_executable_version)
 	)
 	
 	# Compare only MAJOR.MINOR
-	string (SUBSTRING ${LUABIN_VERSION_STRING} "4" "3" LUABIN_VERSION_STRING)
-	string (COMPARE EQUAL ${LUABIN_VERSION_STRING} "${_LUA_VERSION_MAJOR}.${_LUA_VERSION_MINOR}" VERSION_MATCHES)
-	
+	if (NOT LUABIN_VERSION_STRING STREQUAL "")
+		string (SUBSTRING ${LUABIN_VERSION_STRING} "4" "3" LUABIN_VERSION_STRING)
+		string (COMPARE EQUAL ${LUABIN_VERSION_STRING} "${_LUA_VERSION_MAJOR}.${_LUA_VERSION_MINOR}" VERSION_MATCHES)
+	else()
```
markus2330: The coding guidelines say that there should be a space before `(`.

markus2330:
Thank you! On which system do you get an empty string there?

@mpranj can you take a look?

sirblackheart:
On windows using the official cmake gui.

markus2330:
And is lua installed? Which version?

sirblackheart:
Lua is not installed so it is normal that the version string is empty, but cmake generates an error when it gets an empty string in the substring and compare equal functions.

markus2330:
Yes, maybe its better to fail even earlier and not pretend to have a version mismatch. Lets see what @mpranj says.

Nevertheless, there should not be a cmake configuration error, but only the bindings+plugins should be removed.

mpranj:
Yep that's my fault, but we should fail earlier as @markus2330 said.
Thanks for discovering this @sirblackheart!

sirblackheart:
Ok i let @mpranj handle this

markus2330:
I reopen it as issue #745. It breaks quite a view builds.

Its good to always have a issue, even if a PR is created which immediately fixes it. (for situations like this)

Issue #714, opened 2016-05-11
FreeBSD: testpy2_kdb.py fails


mpranj:
Seems very related to #631. `testpy2_kdb.py` is failing because `dlopen` failed.

```
 66/118 Test  #66: test_kdb.py ........................***Failed    0.08 sec
..EEEE
======================================================================
ERROR: test_ctor (__main__.KDB)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mpranj/workspace/libelektra/src/bindings/swig/python/tests/test_kdb.py", line 25, in test_ctor
    self.assertIsInstance(kdb.KDB(), kdb.KDB)
  File "/home/mpranj/workspace/libelektra/build/src/bindings/swig/python/kdb.py", line 1742, in __init__
    _kdb.KDB_swiginit(self, _kdb.new_KDB(*args))
kdb.KDBException: 1 Warning was issued:
 Warning number: 1
    Description: could not load module, dlopen failed
    Ingroup: modules
    Module: dl
    At: /home/mpranj/workspace/libelektra/src/libs/loader/dl.c:82
    Reason: of module: libelektra-resolver.so, because: Shared object "libelektra-resolver.so" not found, required by "python3"
    Mountpoint:
    Configfile:
Error (#40) occurred!
Description: Failed to open default backend (see warnings for more information)
Ingroup: kdb
Module:
At: /home/mpranj/workspace/libelektra/src/libs/elektra/kdb.c:282
Reason: could not open default backend
Mountpoint:
Configfile:
```

[testpy2_kdb.py.txt](https://github.com/ElektraInitiative/libelektra/files/260196/testpy2_kdb.py.txt)

markus2330:
Thanks, isn't it a complete duplicate?

Is it also broken when the installed Elektra was built with `TARGET_PLUGIN_FOLDER` as an empty string? And can you check the RPATH `readelf -a lib/libelektra-core.so.0.8.16 | grep RPATH` (see also #631)

mpranj:
on FreeBSD I get:
` 0x000000000000000f (RPATH)              Library rpath: [::::::::::::::::::::::]`

markus2330:
Maybe it is easier to investigate the issue on FreeBSD than on Mac OS X?

Issue #631, opened 2016-04-26
Some Tests Fail on OS X


sanssecours:
Looks like `test_kdb.lua` fails on OS X:

```
        Start  69: test_kdb.lua
 69/117 Test  #69: test_kdb.lua .......................***Failed    0.00 sec
        Start  70: test_key.lua
 70/117 Test  #70: test_key.lua .......................   Passed    0.00 sec
        Start  71: test_keyset.lua
 71/117 Test  #71: test_keyset.lua ....................   Passed    0.00 sec
```

If I run `lua ../src/bindings/swig/lua/tests/test_kdb.lua`, then the script just exits with return value `0`. If you need any additional information, then please just comment below.

manuelm:
I'm sorry but I don't have OS X. However ctest -V is your friend.

manuelm:
btw calling `lua $file` isn't the same as running the tests. You're using the system installed kdb library while the tests are obviously using the library located in the build directory. You should at least set `LUA_CPATH`. see https://github.com/ElektraInitiative/libelektra/blob/master/src/bindings/swig/lua/tests/CMakeLists.txt#L12

90% of swig related crashes were related to build environment faults. So first try to regenerate(!) + recompile the swig bindings.

markus2330:
This issue might be related to ad537b3 (at least on Linux kdb*lua|python were crashing, but fix for Linux is part of ad537b3). Without output of the test one cannot really tell.

You can use `make run_all` which suppresses the output of successful test cases but shows the output for the failed ones.

manuelm:
@sanssecours Any update on this?

sanssecours:
> This issue might be related to ad537b3 (at least on Linux kdb*lua|python were crashing, but fix for Linux is part of ad537b3). Without output of the test one cannot really tell.

> @sanssecours Any update on this?

Sorry for the late response. I thought I already wrote a comment containing extended information about this issue. Let me start by saying that `testpy2_kdb.py` also fails on my machine. So it might be a problem with my specific setup. I installed [SWIG](http://braumeister.org/formula/swig) via Homebrew and currently use the following `cmake` command to generate the build files for Elektra:

```sh
cmake .. -GNinja -DENABLE_TESTING=ON -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5 -DTOOLS='qt-gui;kdb' -DBUILD_PDF=ON -DBINDINGS=SWIG
```

[...]

PR #755, opened 2016-05-28
fix LD_LIBRARY_PATH for SWIG and GI bindings tests

mpranj:
Fixes issues #714 and #631.


markus2330:
I am puzzled: how can this make a difference?
The main CMakeLists.txt does:
```
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
```

I would prefer if we do not hardcode pathes at multiple places.

mpranj:
I just printed it. For me `CMAKE_LIBRARY_OUTPUT_DIRECTORY` was resolving to the current plugin/binding (build)directory instead of the `build_dir`/lib where the libraries are located.

markus2330:
Ok, cmake seems to reset it internally. Thanks for fixing it!

PR #756, opened 2016-05-28
Fix 754

### LINKED ISSUE #754: make install fails with CMAKE_INSTALL_PREFIX
Using `-DCMAKE_INSTALL_PREFIX` make install fails because it wants to write to system directories.

```
-- Installing: /etc/bash_completion.d/kdb
-- Installing: /etc/profile.d/kdb.sh
```

Is this intended? I'd like to use it without root permissions.

sirblackheart:
This has been confirmed to fix #754 
In addition it exposes both paths so be user set and an option so install the scripts in corresponding `XDG_CONFIG_HOME` folders.

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -20,23 +20,21 @@ install (PROGRAMS upgrade-bootstrap DESTINATION ${TARGET_TOOL_EXEC_FOLDER})
 
 install (PROGRAMS umount-all DESTINATION ${TARGET_TOOL_EXEC_FOLDER})
 
-set(USE_HOME_FOLDER_FOR_SCRIPTS OFF BOOL "if xdg home directorys should be used for script install")
-#There is no default profile.d in $HOME/.config, but using .config/profile seems ok to use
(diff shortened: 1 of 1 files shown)


markus2330:
I think this is a overly complex solution which will cause problems in maintainance. We already have a build job to test INSTALL_SYSTEM_FILES, with this PR we would need at least 2 more build jobs setting either HOME or XDG_CONFIG_HOME when USE_HOME_FOLDER_FOR_SCRIPTS is enabled.

Additionally, my bash docu does not mention $XDG_CONFIG_HOME/profile to be a place that is sourced. Is this a new bash feature?

See also #754

sirblackheart:
> I think this is a overly complex solution.
Yes that is trough.

> Additionally, my bash docu does not mention $XDG_CONFIG_HOME/profile to be a place that is sourced. Is this a new bash feature?

No, for this reason i added the warning to source the folder. But like you said it is a overly complex solution that grew trough trying to solve a problem that was already solved -> INSTALL_SYSTEM_FILES

> See also #754
Yes, lets stay with INSTALL_SYSTEM_FILES

PR #757, opened 2016-05-29
fix jna load library

mpranj:
Fix for OS X / BSD.


markus2330:
Wasn't this a part of #742 earlier? Nevertheless, thank you, its a good fix!

mpranj:
Indeed, but I'd like to keep #742 travis-only.

PR #753, opened 2016-05-28
Qt gui xdg theme support

sirblackheart:
This is a rework of #751 divided into several smaller commits as requested. I also tried to fix all the concerns brought up in the previous pull request. As this is a rather big pull request, I hope we get enough people to test this, so we can make sure no regressions are introduced trough it.

I tried to describe all changes and the reasoning behind them in the commit messages. If i missed any feel free to comment on it.

And as a reminder, this change introduces a new dependency for the qt-gui tool, the qt5 `svg` module.

close #720

PR #759, opened 2016-05-29
Use new MountBackendBuilder Interface when mounting backends

0003088:
* Add feature to remove already added plugins
* Add feature to auto-add recommended plugins
* Plugins can be removed from the list or from a button next to the plugin-dropdown.

How about the new icons? Should I change them myself? I suggest you merge #753 before this one?


0003088:
btw, do you have an astyle template or something for correct formatting?

mpranj:
Yes, clang-format. The configuration is in the project root. Watch out for qml I'm not sure how it's handled.

markus2330:
@sirblackheart Can you review and try it please?

sirblackheart:
This looks good to me and this definitely a useful addition to the backend mount dialog.

@0003088  A short tough regarding backend mounting in the gui. What do you think about to either hide it when the `Elektra Qt Editor` is started in user mode or request elevated rights for the mounting?

markus2330:
Thank you for testing it!

The disable-mounting-when-not-available discussion is exactly the same as the "isWritable" discussion, we should do it in a different issue.

Issue #731, opened 2016-05-18
crypto memleak


markus2330:
In 8df617494ed5cde4fc2a2c5f4492ad2f1fd33f73 I marked the crypto test as MEMLEAK because of:
```
==10002== 8 bytes in 1 blocks are indirectly lost in loss record 1 of 63
==10002==    at 0x4C28C20: malloc (vg_replace_malloc.c:296)
==10002==    by 0x6012ED0: ???
==10002==    by 0x601422B: ???
==10002==    by 0x60ADDF4: ???
==10002==    by 0x60ADE23: ???
==10002==    by 0x60ADF31: ???
==10002==    by 0x6012E2F: ???
==10002==    by 0x6013044: ???
==10002==    by 0x5E0330A: ???
==10002==    by 0x5E04261: ???
==10002==    by 0x4E41110: elektraPluginOpen (plugin.c:295)
==10002==    by 0x4020FE: test_init (testmod_crypto.c:127)
==10002==    by 0x4020FE: main (testmod_crypto.c:250)
```

Please remove the tag (revert the commit) and fix the memleak with the next PR.

markus2330:
(You do not (should not) look at it now, it is only a reminder for the next PR)

markus2330:
A *very* minor issue:

```
cmake -DPLUGINS="dump;resolver_fm_hpu_b;crypto_openssl" 
...
-- checking for one of the modules 'openssl'
-- Exclude Plugin crypto_openssl because OpenSSL development files not found
-- Found libgcrypt: -lgcrypt
```

Searching for gcrypt is not really needed, if not requested to add this plugin.

petermax2:
I had a quick look at the memleak. It is most probably caused by libgcrypt itself.

I agree with the CMake check. I will adapt the `CMakeList.txt`.

markus2330:
It does not need to be fixed for the next milestone. Yes, likely its only about valgrind suppressions.

petermax2:
Note (for myself): the memory of the configuration KeySets is lost at

    static void test_config_errors ()

by calling `newWorkingConfiguration()` without saving the pointer.

FIX by calling `ksDel()`...

petermax2:
Interesting fact: if you compile `dl.c` with the following change, no memleaks occur at all:

	diff --git a/src/libs/loader/dl.c b/src/libs/loader/dl.c
	index 2e3018a..03c7e33 100644
	--- a/src/libs/loader/dl.c
[...]

Issue #675, opened 2016-04-29
qt-gui segfaults when elektra is build with crypto plugins


sirblackheart:
Building elektra with
```
cmake -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DBUILD_SHARED=ON \
    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
    -DTOOLS=ALL \
    -DPLUGINS=ALL ..
```

leads to a segfault when starting the qt-gui.
Building with `-DPLUGINS="ALL;-crypto_gcrypt;-crypto_openssl"` solves this for now.

lldb gives me the following backtrace:
```
* thread #1: tid = 19176, 0x00007fffd12018d0, name = 'qt-gui', stop reason = signal SIGSEGV: address access protected (fault address: 0x7fffd12018d0)
  * frame #0: 0x00007fffd12018d0
    frame #1: 0x00007ffff1a93079 libcrypto.so.1.0.0`ERR_get_state + 57
    frame #2: 0x00007ffff1a94269 libcrypto.so.1.0.0`ERR_peek_error + 9
    frame #3: 0x00007fffd0dee891 libelektra-crypto_openssl.so`elektraCryptoOpenSSLInit(errorKey=0x0000000001823d80) + 177 at openssl_operations.c:93
    frame #4: 0x00007fffd0defb72 libelektra-crypto_openssl.so`libelektra_cryptoopenssl_openssl_LTX_elektraPluginopen [inlined] elektraCryptoInit(errorKey=0x0000000001823d80) + 34 at crypto.c:38
    frame #5: 0x00007fffd0defb6a libelektra-crypto_openssl.so`libelektra_cryptoopenssl_openssl_LTX_elektraPluginopen(handle=<unavailable>, errorKey=0x0000000001823d80) + 26 at crypto.c:191
    frame #6: 0x00007ffff5a9655b libelektra-kdb.so.4`elektraPluginOpen(name=0x00007fffffffb400, modules=<unavailable>, config=<unavailable>, errorKey=0x0000000001823d80) + 251 at plugin.c:295
    frame #7: 0x00007ffff6b27f9f libelektratools.so.0`kdb::tools::Plugin::Plugin(this=0x000000000180a590, spec_=<unavailable>, modules=<unavailable>) + 239 at plugin.cpp:39
[...]

PR #767, opened 2016-06-02
add shutdown option to crypto plugin config

petermax2:
In order to prevent the qt-gui from crashing (see #675 ) we disable the crypto library teardown in kdbClose() by default. In order to enable it, the key /crypto/shutdown must be set to true within the plugin configuration.

See #675 and #731 for full discussion.

@markus2330 wrote in #731 :
> Ok, then we should not cleanup by default. (Use a plugin config as done in python plugin). In the unit tests you should use cleanup to make valgrind happy.

What do you think about this solution? At least `qt-gui` should be running now, altough the memory leaks in `testmod_crypto` still occur.

### DIFF
--- a/src/plugins/crypto/README.md
+++ b/src/plugins/crypto/README.md
@@ -126,6 +126,16 @@ The following example configuration illustrates this concept:
 
 Only keys marked with a certain meta-key will be considered for encryption/decryption.
 
+## Configuration ##
+
+The following key must be set to `"true"` within the plugin configuration,
+if the plugin should shut down the crypto library:
+
+	/crypto/shutdown
+
+Per default shutdown is disabled to prevent applications like the qt-gui from crashing.
+Shutdown is enabled in the unit tests to prevent memory leaks.
+
 ## Examples ##
 
 ### Metadata based encyption ###
--- a/src/plugins/crypto/README.md
+++ b/src/plugins/crypto/README.md
@@ -131,7 +131,7 @@ Only keys marked with a certain meta-key will be considered for encryption/decry
 The following key must be set to `"1"` within the plugin configuration,
 if the plugin should shut down the crypto library:
 
-	/crypto/shutdown
+	/shutdown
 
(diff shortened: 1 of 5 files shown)


src/plugins/crypto/README.md
```diff
@@ -126,6 +126,16 @@ The following example configuration illustrates this concept:
 
 Only keys marked with a certain meta-key will be considered for encryption/decryption.
 
+## Configuration ##
+
+The following key must be set to `"true"` within the plugin configuration,
+if the plugin should shut down the crypto library:
+
+	/crypto/shutdown
```
markus2330: Why use prefix for configuration? If you want it plugin-specific, you can pass it only to plugin. And if you want it for the backend, you pass it as backend config.

src/plugins/crypto/crypto.c
```diff
@@ -205,15 +205,35 @@ int CRYPTO_PLUGIN_FUNCTION (open) (Plugin * handle ELEKTRA_UNUSED, Key * errorKe
...
+	}
+
+	Key * shutdown = ksLookupByName (pluginConfig, ELEKTRA_CRYPTO_PARAM_SHUTDOWN, 0);
+	if (!shutdown)
+	{
+		return 1; // applying default behaviour -> success
+	}
+	else
+	{
+		if (strcmp (keyString (shutdown), "true") != 0)
```
markus2330: See python implementation and decision, it should not behave differently here.

petermax2: The python plugin uses a ref counter (protected by a mutex). The crypto plugin does actually the same thing.

There is no mentioning of "python" in `doc/decisions`.

markus2330: I mean the decision doc/decisions/bool.md

markus2330:
Looks good now, please ping me if ready.

petermax2:
@markus2330 I like it for now. #675 is fixed with this PR. I will fix the remaining 18 memory leaks in a future PR.

Issue #762, opened 2016-05-31
GNOME Integration: ElektraSettings (GSettingsBackend)


raetiacorvus:
# ElektraSettings (GSettingsBackend)

I will use this ticket to track the state of the GSettingsBackend implementation:
## Interface Implemenation
- [x] read () method to read a key's value
- [x]  read_user_value () virtual method to read user's key value
- [x]  get_writable () method to get if a key is writable
- [x]  write () method to change key's value
- [x]  write_tree ()  method to change a tree of keys
- [x]  reset () virtual method to reset state -- _#761_ --
- [x]  subscribe () method to subscribe to key changes
- [x]  unsubscribe () method to unsubscribe to key changes
- [ ]  sync () virtual method to sync state -- _needs to handly conflicts and errors_ --
- [x]  get_permission ()  method to get permission of a key -- _not implemented_ --
## Code Quality
- [ ]  cleaner cmake code
- [ ]  comments -- _partially_ --
- [ ]  error handling
- [ ]  peer reviews
- [x] pass gsettings own unit tests
## Research
- [ ] which applications actually use g_settings_new_with_backend ()
## Note

Currently the backend is developed [outside](https://github.com/sirblackheart/elektrasettings) of the Elektra Source Tree to grant flexibility during early development cycles. It will be integrated once it reaches a mature and stable state.
## Discussion
- Specific behaviors relating getKDB, setKDB and append Key
  - those are only bugs in the gsettingsbackend, probably related to threading
- dbus as global plugin?
  - offenes issue #404 
- convenience plugin (bool 'true', 'false')

markus2330:
Nice overview, well done!

Did you already try to run the gsettings unit tests?

sirblackheart:
Just did and it looks rather good for the current state of the code.
The first two errors turn up in tests where ElektraSettings is still missing the implementation (`writeTree`). The others seem to come from calling the single test directly. But that is the only way to run all tests currently, as the glib tests segfaults on any error.

```
/gsettings/basic: OK
/gsettings/basic-types: OK
/gsettings/complex-types: OK
/gsettings/changes: OK
/gsettings/l10n: warning: translation is not working... skipping test. OK
/gsettings/l10n-context: warning: translation is not working... skipping test.  OK
/gsettings/delay-apply: **
GLib-GIO:ERROR:gsettings.c:492:test_delay_apply: assertion failed: (!changed_cb_called)
/gsettings/delay-revert: OK
[...]

PR #768, opened 2016-06-02
GNOME Integration: decision to GSettings values

raetiacorvus:
Decision on how to handle GSettings values when using Elektra as
backend.

### DIFF
--- a/doc/decisions/gsettings_values.md
+++ b/doc/decisions/gsettings_values.md
@@ -2,51 +2,69 @@
 
 ## Issue
 
-`GSettings` uses `Glibs` `GVariant` type for it values. The `GVariant` type can hold many
+`GSettings` uses `GLibs` `GVariant` type for it values. The `GVariant` type can hold many
 different variations of types but `Elektra` only distinguishes between strings
 and binary data.
 
 ## Constraints
 
  - saved values have to be readable by `GSettings` but also trough `Elektra` itself
- - keep `Elektra` as `Glib` independent as possible
+ - keep `Elektra` and `GLib` as independent as possible
  - values should be human readable
+ - values should be rather self explaining about their type
+ - saved values should be as portable as possible
 
 ## Assumptions
 
  - `Elektras` strength is its modularity to provide multiple backends that decide how to save values
- - `GVariants` can be printed as Strings and parsed back again if we know the type
- - `GVariants` can be also serialized and deserialized
- - representing `GVariants` values trough strings fulfills all three constraints
+ - applications using `GSettings` are well aware about the types they use
+ - `GVariant` types can be represented as strings
+ - `GVariant` types are rather self explaining about their type to some degree e.g. (255, 144) is a tuple of int
+ - `GVariant` string representation can be parsed back into a `GVariant` reliable if we know their type
+ - `GVariant` string representation can be parsed back into a `GVariant` unreliable on type guessing by parsing
+ - `GVariant` can be serialized and deserialized, meaning it is possible to save all types binary
+ - `Elektra` offers types trough templates in its c++ api
 
 ## Considered Alternatives
 
- - save serialized values as binarys,
+ - save serialized `GVariant` values as binarys
  - create a mapping between GVariants types and `spec` types and transform them when saving
+ - use metadata to always exactly know the `GVariant` type
 
 ## Decision
 
- - Use GVariants string representation
+ - use GVariants string representation
+ - do not use GVariants additional type information in the string
+ - do not use type mapping in the `GSettings` backend layer
+ - keep usage of metadata as minimal as possible in favor of portability
 
 ## Argument
 
-GVariants string representationat is the most useful for all parts involved. It is immediately human readable and usable in code trough string compares.
-If a party involved knows how GVariant represents a type it can use its own parser to recreate a data type from it.
-Serialized values would create a dependency on `GVariants` deserialization or would need a custom implementation, without they would not be human readable.
-Mapping too `GVariant` would need spec to always be up to date and changes in types could break the mapping.
-In addition it will be hard to match all `GVariant` types.
+GVariants string representation is the most useful for all parts involved. It is immediately human readable and usable in code trough string compares.
+If a party involved knows how GVariant represents a type and its format it can use its own parser to recreate the data.
+
+GVariant offers to print additional type info included to the string but this defeats the purpose of having a
+value that is easily human readable and the output seems to be only generated in special cases e.g. if an int is unsigned.
+
+Serialized values would either create a dependency on `GVariants` deserialization or need a custom implementation in Elektra.
+Without one of both they would not be human readable.
+
+Direct mapping to `GVariant` would need `spec` to always be up to date and changes in types could break the mapping.
+In addition it will be hard to match all possible `GVariant` types.
+
+The idea is also to keep the `GSettings` backend layer as thin as possible and handle further problems in `Elektra` itself.
+
+We could consider to save the GVariant type as metadata, but would prefer to keep metadata at a minimum.
 
 ## Implications
 
+An application has to know what type it wants to save and what type it expects.
 It only solves the problem how to handle direct writes and reads from GSettings.
 It does not solve the problem of defaults and constrains when an application accesses GSettings values in Elekra.
 
 ## Related decisions
 
 This influences the future follow up decision on how to handle defaults and constrains of GSettings in Elektra.
+Decisions on how to handle specific types in Elektra will have influence on this decission (e.g. handling of boolean values in string form)
 
 ## Notes
-
-We could consider to save the GVariant type as metadata, but would prefer to keep metadata at a minimum.
-GVariant offers to print additional type info included to the string but this defeats the purpose of having a
-value that is easily human readable and the output seems to be only generated in special cases e.g. if an int is unsigned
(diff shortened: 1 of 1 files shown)


doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,52 @@
...
+ - saved values have to be readable by `GSettings` but also trough `Elektra` itself
+ - keep `Elektra` as `Glib` independent as possible
+ - values should be human readable
+
+## Assumptions
+
+ - `Elektras` strength is its modularity to provide multiple backends that decide how to save values
+ - `GVariants` can be printed as Strings and parsed back again if we know the type
+ - `GVariants` can be also serialized and deserialized
+ - representing `GVariants` values trough strings fulfills all three constraints
```
markus2330: two line above it says that you need to know the type, so do you need meta-data, too?

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Considered Alternatives
+
+ - save serialized values as binarys,
+ - create a mapping between GVariants types and `spec` types and transform them when saving
+
+## Decision
+
+ - Use GVariants string representation
```
markus2330: Is the string representation of basic types (int, bool) compatible with Elektra's types?

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Argument
+
+GVariants string representationat is the most useful for all parts involved. It is immediately human readable and usable in code trough string compares.
```
markus2330: typo: representationat

doc/decisions/gsettings_values.md
sirblackheart: No GSettings tells me what type it expects

doc/decisions/gsettings_values.md
sirblackheart: boolean is repesented as 'true', 'false' that would need the convenience plugin globally to work with elektras c++ api.
Everything is a pure string even integers.

And i also do not want to do type conversion as it would be inconsistent with other GVariant types. like nested arrays of boolean for example.

If we would start doing that we would have to write our own GVariant converter, which is probably enough work for another thesis.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Related decisions
+
+This influences the future follow up decision on how to handle defaults and constrains of GSettings in Elektra.
+
+## Notes
+
+We could consider to save the GVariant type as metadata, but would prefer to keep metadata at a minimum.
+GVariant offers to print additional type info included to the string but this defeats the purpose of having a
+value that is easily human readable and the output seems to be only generated in special cases e.g. if an int is unsigned
```
markus2330: These are not notes, but important parts of the decision.

doc/decisions/gsettings_values.md
sirblackheart: The idea is also to keep this layer as thin as possible and add needed functionality as plugins to Elektra.

markus2330: It was more as a question, the implication should describe how compatible the gvariants types are to other types and which additional plugins would be needed.
(It does not mean you have to write every plugin related to types)

sirblackheart: Yes. You are right i probably should elaborate this problem better. I have to admit i totally forgot about Elektras c++ api and types while using the glib binding which are nearly identical as using the c api. Because of that I am probably also still lacking the proper knowledge about Elektras behavior with different types. This also brings up my question how the different plugin handle 'true' and 'false' strings. But I would rather like to discuss that at our next meeting so everyone benefits from it. I am already putting together a list inside #762 regarding other questions.

Again thank you for you thorough feedback, and sorry for my sometimes harsh answers.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+ - keep `Elektra` and `GLib` as independent as possible
+ - values should be human readable
+ - values should be rather self explaining about their type
+ - saved values should be as portable as possible
+
+## Assumptions
+
+ - `Elektras` strength is its modularity to provide multiple backends that decide how to save values
+ - applications using `GSettings` are well aware about the types they use
+ - `GVariant` types can be represented as strings
```
markus2330: As discussed: maybe link some material or give some more information about their representation.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+ - `GVariant` types are rather self explaining about their type to some degree e.g. (255, 144) is a tuple of int
```
markus2330: I thought there are some `'` quotes?

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+ - `GVariant` string representation can be parsed back into a `GVariant` reliable if we know their type
+ - `GVariant` string representation can be parsed back into a `GVariant` unreliable on type guessing by parsing
+ - `GVariant` can be serialized and deserialized, meaning it is possible to save all types binary
+ - `Elektra` offers types trough templates in its c++ api
```
markus2330: misspelled "through"

Elektra also has type definitions in C (kdbtypes.h), only the type checker (currently) is implemented in C++.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+ - save serialized `GVariant` values as binarys
+ - create a mapping between GVariants types and `spec` types and transform them when saving
+ - use metadata to always exactly know the `GVariant` type
+
+## Decision
+
+ - use GVariants string representation
+ - do not use GVariants additional type information in the string
+ - do not use type mapping in the `GSettings` backend layer
+ - keep usage of metadata as minimal as possible in favor of portability
```
markus2330: That is a constraint, not a decision. As decision we would be interested to know which meta data it should use.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+## Considered Alternatives
+
+ - save serialized `GVariant` values as binarys
+ - create a mapping between GVariants types and `spec` types and transform them when saving
+ - use metadata to always exactly know the `GVariant` type
+
+## Decision
+
+ - use GVariants string representation
+ - do not use GVariants additional type information in the string
```
markus2330: Yes, good idea. But we should store them in some meta-data, ideally directly `type`, otherwise `gvariant` with the letters encoding the gvariant-type.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Argument
+
+GVariants string representation is the most useful for all parts involved. It is immediately human readable and usable in code trough string compares.
+If a party involved knows how GVariant represents a type and its format it can use its own parser to recreate the data.
+
+GVariant offers to print additional type info included to the string but this defeats the purpose of having a
+value that is easily human readable and the output seems to be only generated in special cases e.g. if an int is unsigned.
+
+Serialized values would either create a dependency on `GVariants` deserialization or need a custom implementation in Elektra.
```
markus2330: How is this different to string serialization?

Btw. it would not matter if some plugins depend on GVariants.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+Without one of both they would not be human readable.
+
+Direct mapping to `GVariant` would need `spec` to always be up to date and changes in types could break the mapping.
```
markus2330: Which would work by mounting the schemata to spec.

The meta-data can applied to every namespace, its not specific to spec. But not every storage plugin is capable of storing meta-data, but this can be fixed with a metatostring plugin.

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+In addition it will be hard to match all possible `GVariant` types.
+
+The idea is also to keep the `GSettings` backend layer as thin as possible and handle further problems in `Elektra` itself.
```
markus2330: That is also a constraint. And I fully agree with it!

doc/decisions/gsettings_values.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Implications
+
+An application has to know what type it wants to save and what type it expects.
+It only solves the problem how to handle direct writes and reads from GSettings.
+It does not solve the problem of defaults and constrains when an application accesses GSettings values in Elekra.
+
+## Related decisions
+
+This influences the future follow up decision on how to handle defaults and constrains of GSettings in Elektra.
```
markus2330: I am afraid its not easily possible to separate the decisions, also here we already need to know where to get the type from.

markus2330:
Can I merge it?

sirblackheart:
Sorry i did not have time yet to read over your new feedback.

doc/decisions/gsettings_values.md
sirblackheart: I only realized it after writing this document. I should update the information.

sirblackheart:
I think i need time to do a rework after we are trough with the surveys issues.

markus2330:
Ok, do not worry, lets keep it open. It is not urgent.

markus2330:
I merged it in the case someone wants to pick the work on GSettings up.

Issue #766, opened 2016-06-02
Devhelp(doc) generation for (g)elektra


raetiacorvus:
This would be a nice to have for developing on Elektra in GNOME.
As far as i can tell this could be possible possible for the whole elektra api trough doxygen xml generation and XSLT transformation. Here is an example [xsl](https://cgit.freedesktop.org/dbus/dbus/tree/doc/doxygen_to_devhelp.xsl).
- [Devhelp](https://wiki.gnome.org/Apps/Devhelp)

TODO:
- [ ] doxygen xslt to devhelp for `libelektra`
- [ ] gtk-doc generation for `libgelektra`

markus2330:
This is a good idea! Is devhelp integrated in your IDE? Aren't there standard tools to convert doxygen to devhelp?

sirblackheart:
> This is a good idea! Is devhelp integrated in your IDE? 

Yes devhelp is integrated in [gnome-builder](https://wiki.gnome.org/Apps/Builder)

> Aren't there standard tools to convert doxygen to devhelp?

As far as i can tell the described way seems the default for doxygen projects. Doxygen already brings an xml generator we only need to provide the xsl and the [right commands.](https://www.slac.stanford.edu/grp/eg/minos/dist/dist_aux4/sigc++/libsigc++-2.2.4.2/build/doc-reference.am)

markus2330:
Seems like a piece of cake, maybe you even find the commands implemented in CMake.

sirblackheart:
Yes i just have to figure out how to get the xsl right, then we can use the existing doxygen html output. We also need to output xml but that is just a simple switch.

markus2330:
Closed because of low priority, will be reevaluated after 1.0.

PR #769, opened 2016-06-02
GElektra: add basic devhelp generation

raetiacorvus:
This ads devhelp generation if glib bindings are enabled and gtk-doc is found.

Currently this only contains the object hierarchy and all methods.

ref #766

### DIFF
--- a/src/bindings/glib/CMakeLists.txt
+++ b/src/bindings/glib/CMakeLists.txt
@@ -68,7 +68,7 @@ else()
 		add_subdirectory (tests)
 	endif (BUILD_TESTING)
 
-	find_package (GtkDoc 1.25)
+	find_package (GtkDoc 1.25 QUIET)
 	if (GTKDOC_FOUND)
 		include (GNUInstallDirs)
 		gtk_doc_add_module (
--- a/src/bindings/glib/CMakeLists.txt
+++ b/src/bindings/glib/CMakeLists.txt
@@ -35,6 +35,8 @@ else()
 	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC elektra-core elektra-kdb)
 	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC ${GOBJECT_LIBRARIES})
 
+	message(WARNING "GOBJECT LIBRARIES: " ${GOBJECT_LIBRARIES})
+	message(WARNING "GOBJECT LDFLAGS: " ${GOBJECT_LDFLAGS})
 	install (
 		TARGETS ${GELEKTRA_LIBRARY}
 		LIBRARY DESTINATION lib${LIB_SUFFIX}
@@ -71,13 +73,15 @@ else()
 		include (GNUInstallDirs)
 		gtk_doc_add_module (
 			gelektra SOURCE ${CMAKE_CURRENT_SOURCE_DIR}
-			XML ${CMAKE_SOURCE_DIR}/doc/glib/gelektra-docs.xml
+			XML ${CMAKE_SOURCE_DIR}/doc/devhelp/gelektra-docs.xml
 			LIBRARIES ${GELEKTRA_LIBRARY}
 		)
 		add_custom_target (
 			gelektra-doc ALL DEPENDS doc-gelektra)
 			install (DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/gelektra/html/
-		    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/gtk-doc/html/gelektra
-        )
+			DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/gtk-doc/html/gelektra
+		)
+	else()
+		message (STATUS "GtkDoc not found. Disabling devhalp generation.")
 	endif (GTKDOC_FOUND)
 endif()
--- a/src/bindings/glib/CMakeLists.txt
+++ b/src/bindings/glib/CMakeLists.txt
@@ -33,7 +33,7 @@ else()
 	target_compile_options (${GELEKTRA_LIBRARY} PUBLIC ${GOBJECT_CFLAGS_OTHER})
 
 	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC elektra-core elektra-kdb)
-	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC ${GOBJECT_LDFLAGS})
+	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC ${GOBJECT_LIBRARIES})
 
 	install (
 		TARGETS ${GELEKTRA_LIBRARY}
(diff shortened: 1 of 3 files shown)


src/bindings/glib/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ else()
 	target_compile_options (${GELEKTRA_LIBRARY} PUBLIC ${GOBJECT_CFLAGS_OTHER})
 
 	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC elektra-core elektra-kdb)
-	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC ${GOBJECT_LDFLAGS})
+	target_link_libraries (${GELEKTRA_LIBRARY} PUBLIC ${GOBJECT_LIBRARIES})
```
sirblackheart: I hope this change does not break anything, but gtk doc did put the wrong linking flags with `${GOBJECT_LDFLAGS}`

markus2330: @manuelm what do you think here? Is this a bug?

A separate PR (or at least commit) for this might be useful (if this change should be done)

src/bindings/glib/CMakeLists.txt
```diff
@@ -65,4 +65,19 @@ else()
...
+			gelektra SOURCE ${CMAKE_CURRENT_SOURCE_DIR}
+			XML ${CMAKE_SOURCE_DIR}/doc/glib/gelektra-docs.xml
+			LIBRARIES ${GELEKTRA_LIBRARY}
+		)
+		add_custom_target (
+			gelektra-doc ALL DEPENDS doc-gelektra)
+			install (DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/gelektra/html/
+		    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/gtk-doc/html/gelektra
+        )
+	endif (GTKDOC_FOUND)
```
markus2330: Do you think silent not-building of doc is a good idea? Add some warning if GtkDoc not found?

markus2330:
Thank you, looks really nice. The ${GOBJECT_LIBRARIES} issue has to be investigated though. Mac OS X build fails:
```
[ 30%] Linking C shared library ../../../lib/libgelektra-4.0.dylib

ld: library not found for -lintl

```

src/bindings/glib/CMakeLists.txt
manuelm: `GOBJECT_LIBRARIES` will be missing important linker flags, so linking might break. I've spent quite some time (read: hours) investigating into the different variables pkg-config is returning. `_LIBRARIES` will be insufficient.

sirblackheart:
Yes i noticed, i think we have to get a proper findGobject.cmake to solve this. It has to do how homebrew handles gettext installation.

src/bindings/glib/CMakeLists.txt
markus2330: Is there some docu about your findings? Is GOBJECT_LIBRARIES a subset of GOBJECT_LDFLAGS? Any explanation why "linking against ldflags" makes sense?

manuelm: No. I read the docs + looked at various .pc files on different linux distributions.
Yes, libraries is a subset of ldflags. And we're not "linking against ldflags". `target_link_libraries` is just misleading.

src/bindings/glib/CMakeLists.txt
sirblackheart: I actually added something but must have gotten lost in some copy paste

src/bindings/glib/CMakeLists.txt
markus2330: Ok, thank you for your quick and helpful response!

@sirblackheart what is the specific problem why you need to change this? Is there a way to modif gtkdoc flags independently?

sirblackheart: The difference of the two variables on my system:
```shell
-- Include Binding glib
CMake Warning at src/bindings/glib/CMakeLists.txt:38 (message):
  GOBJECT LIBRARIES: gobject-2.0glib-2.0


CMake Warning at src/bindings/glib/CMakeLists.txt:39 (message):
  GOBJECT LDFLAGS: -lgobject-2.0-lglib-2.0
```

the problem is the added -l when using ${GELEKTRA_LIBRARY} for the gtk doc as it get doubled and then not found eg `-l-lgobject-2.0`

manuelm: Tried target_link_libraries (doc-gelektra ${GELEKTRA_LIBRARY}) instead of LIBRARIES?

sirblackheart:
```
<XPREFIX>_LIBRARIES      ... only the libraries (w/o the '-l')
<XPREFIX>_LDFLAGS        ... all required linker flags
```

is there really a difference other than the -l? I mean they both just use pkg-config information.
Are you sure this was not just a workaround for OS X using brew cellar components. Look at the current output it shows both variables on osx.

```
CMake Warning at src/bindings/glib/CMakeLists.txt:38 (message):
  GOBJECT LIBRARIES: gobject-2.0glib-2.0intl
CMake Warning at src/bindings/glib/CMakeLists.txt:39 (message):
  GOBJECT LDFLAGS:
  -L/usr/local/Cellar/glib/2.46.2/lib-L/usr/local/opt/gettext/lib-lgobject-2.0-lglib-2.0-lintl
```

manuelm: Yes, there is a difference. A good example is `-lpthread`.

manuelm: What about just building the libraries yourself? `gtk_doc_add_module (... LIBRARIES elektra-core elektra-kdb ${GOBJECT_LIBRARIES})` .. and be done

manuelm: erm yeah. Forget that.
You can change `PUBLIC ${GOBJECT_LDFLAGS}` to `PRIVATE ${GOBJECT_LDFLAGS}` and additionally add $GOBJECT_LIBRARIES. But that is actually a misuse of this feature and a possible subproject linking to gelektra needs to lookup and add $GOBJECT_LDFLAGS on it's own.

sirblackheart: Thank you for the tipp.
In the long run we probably should solve this with a proper findGObject.cmake

sirblackheart: Ah no it does not solve the problem.
Lets put this on hold till we have a proper findGObject.cmake

sirblackheart: We dont have yet a `findGObject.cmake` only a `FindGObjectIntrospection.cmake` for`gi`.
The Elektra GLib bindings currently use [pkg-config module](https://cmake.org/cmake/help/v3.5/module/FindPkgConfig.html)
using
`${GOBJECT_LDFLAGS}` should be equal to `pkg-config gobject-2.0 --libs`

I really do not understand yet why this should pull in `-lpthread` or `-fPIE`, but i am no expert of cmake yet.

sirblackheart: you only get `-pthread` if you use the STATIC prefix e.g. `${GOBJECT_LDFLAGS}`

markus2330:
@manuelm can you give some help here how @sirblackheart can fix it?

manuelm:
I'm sorry but I don't have access to any OS X machine. Maybe the best is to leave gi bindings on OSX broken unless someone else fixes the cmake files.

Or we strip of `-lintl` on OS X. On FreeBSD 10 and above (afair) gettext is embedded into glibs. Maybe it's the same on OS X.

PR #773, opened 2016-06-06
fix #691

### LINKED ISSUE #691: KDB_DB_SPEC and KDB_DB_SYSTEM with ~ does not work
SPEC Expected behaviour: Should resolve to `~/.config/kdb/spec/default.ini`
SPEC Actual behaviour: Resolves to `/usr/local/~/.config/kdb/spec/default.ini`

SYSTEM Expected behaviour: Should resolve to `/home/me/...`
SYSTEM actual behaviour: resolves to `~/...`

KDB_DB_SYSTEM also needs to be STRING, otherwise it is impossible to start string with `~`.

tom-wa:


src/plugins/resolver/filename.c
```diff
@@ -406,11 +422,26 @@ static int elektraResolveSpec (resolverHandle * p, Key * warningsKey ELEKTRA_UNU
 	}
 	else
 	{
-		strcpy (p->filename, KDB_DB_SPEC);
-		strcat (p->filename, "/");
-		strcat (p->filename, p->path);
+		if (KDB_DB_SPEC[0] == '~')
```
markus2330: Did you already consider discussion at #691? ~ would not be on the first position.

tom-wa: not really, this version would only work after dropping CMAKE_INSTALL_PREFIX for pathnames starting with `~`
i don't really see an other way of doing it inside the resolver

markus2330: You need to add a new variable in kdbconfig.h, maybe KDB_DB_SPEC_RELATIVE.

markus2330:
Please update as discussed.

markus2330:
@tom-wa please ping if finished

tom-wa:
@markus2330 done

markus2330:
thank you!

PR #776, opened 2016-06-10
Fix cmake configure when BUILD_DOCUMENTATION is set to OFF

KurtMi:
I found a bug in cmake when setting `BUILD_DOCUMENTATION` to `OFF`.
This PR fixes it!


markus2330:
Thank you, well done!

PR #781, opened 2016-06-12
global plugins: remove spec keys for postgetstorage plugins from kdbGet keyset

tom-wa:


markus2330:
Does this create trouble when writing to spec keys?

tom-wa:
i don't think because that part gets only executed for plugins > STORAGE_PLUGIN 
could be wrong though, i'll do some tests

markus2330:
Which issues does it fix?

tom-wa:
the issue was that validation plugins also got the spec keys in their keyset when using global plugins and because of that, failed.

tom-wa:
ok, i made a mistake here and it doesn't really fix the problem anyway. 
do PRs get closed automatically when force pushing a branch ?

markus2330:
Nevertheless thank you for investing time. For this release, please only tackle well-known issues with an easy fix.

tom-wa:
ok, i've just noticed that i've already implemented a workaround for that issue in conditionals and that was the reason why i couldn't reproduce it. 
since the global plugin part in kdb.c will be rewritten anyway, i'll forget about the issue for now

markus2330:
It is always good to have a workflow build around the issue tracker to avoid something like that.

PR #784, opened 2016-06-13
Gsettings fix `pkgconfig paramter` and use `G_GNUC_UNUSED`

sirblackheart:


src/bindings/gsettings/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ if (CMAKE_VERSION GREATER 3.3)
 	pkg_get_variable (GIO_MODULE_DIR gio-2.0 giomoduledir)
 else ()
 	execute_process (
-		COMMAND PKG_CONFIG_EXECUTABLE --variable=giomoduledir
+		COMMAND PKG_CONFIG_EXECUTABLE --variable=giomoduledir gio-2.0
```
markus2330: I think this alone wont fix the issue, execute_process is not used correctly, see https://cmake.org/cmake/help/v3.0/command/execute_process.html.

sirblackheart: You are right it is OUTPUT_VARIABLE

src/bindings/gsettings/CMakeLists.txt
```diff
@@ -48,8 +48,8 @@ if (CMAKE_VERSION GREATER 3.3)
 	pkg_get_variable (GIO_MODULE_DIR gio-2.0 giomoduledir)
 else ()
 	execute_process (
-		COMMAND PKG_CONFIG_EXECUTABLE --variable=giomoduledir
-		OUTPUT GIO_MODULE_DIR
+		COMMAND PKG_CONFIG_EXECUTABLE --variable=giomoduledir gio-2.0
+		OUTPUT_VARIABLE GIO_MODULE_DIR
```
markus2330: Did you try it? I do not think that `PKG_CONFIG_EXECUTABLE` is a command that will be found. (Variable is not dereferenced) I think its better that we remove the if (CMAKE_VERSION). It will make it easier for you to spot such problems.

src/bindings/gsettings/elektrasettingsbackend.c
```diff
@@ -368,9 +368,9 @@ static gboolean elektra_settings_backend_get_writable (GSettingsBackend * backen
 }
 
 
-static void elektra_settings_key_changed (GDBusConnection * connection, const gchar * sender_name, const gchar * object_path,
-					  const gchar * interface_name, const gchar * signal_name, GVariant * parameters,
-					  gpointer user_data)
+static void elektra_settings_key_changed (GDBusConnection * connection G_GNUC_UNUSED, const gchar * sender_name G_GNUC_UNUSED,
+					  const gchar * object_path ELEKTRA_UNUSED, const gchar * interface_name G_GNUC_UNUSED,
```
markus2330: Why is ELEKTRA_UNUSED here at one place?

src/bindings/gsettings/CMakeLists.txt
sirblackheart: no, i just have to keep double testing with debian stable and 5 year old software

sirblackheart:
the next pushes will be so i can build it in my debian box. i will ping you once it works.

sirblackheart:
@markus2330 it now installs correctly into `/lib/x86_64-linux-gnu/gio/modules` for me. Please test if it is fixed now.

markus2330:
Looks much better now, thank you!

PR #787, opened 2016-06-13
Fix #777

### LINKED ISSUE #777: Elektra Qt Editor: error on using menu `E_dit -> Edit ...` twice on same element
`qrc:/qml/EditKeyWindow.qml:11: TypeError: Cannot read property 'metaValue' of null`

peek 2016-06-11 15-31

0003088:


markus2330:
Thank you! Maybe you can add some information in the commit message the next time, in a meeting we agreed `fix #777` should only be additional information and not be in the title.

0003088:
Since it was only one line that was changed and we already talked about it I didn't think it would be necessary to lose another word about it. But sure, I will add more info in the future.

markus2330:
Yes, it is sometimes strange while doing it, but when looking at the commit history some time later it helps.

PR #793, opened 2016-06-14
Cmake: Fix Build without bindings

KurtMi:
And again, I found a bug while playing with cmake. I tried to build without bindings and an error occurred.
This PR should fix it, but please check it carefully, because I never worked with lists in cmake.


markus2330:
Thank you for playing around with it! I used your suggestion but also fixed the whitespaces in the previous lines.

Issue #782, opened 2016-06-13
find-tools:21: 'glob' imported but unused


markus2330:
Thanks to @reox for reporting.

PR #794, opened 2016-06-14
Scripts: find-tools makes now pep and pyflakes happy

KurtMi:


markus2330:
We can discuss again if we want to change style of python scripts, but this would make the tab usage inconsistent to other python scripts, so I will delay this after release.

KurtMi:
I thought I fix all pep warnings, but I will change it back to tabs, so it is done!

markus2330:
Actually pep8 does not make much sense here, its a completely different style than we are using.

I think we should either fully use pep8 or make it similar to our existing guidelines. I would favour our own guidelines, it makes editor settings easier.

We should also consider the python code we already have.

KurtMi:
Ok then, leave this PR open I will push the file search also here, asap!

KurtMi:
@markus2330 should be fine now and fix #782 and #799. Sorry for the delay, it planned to do it before 8.17.

KurtMi:
jenkins build please

markus2330:
Thank you!

PR #797, opened 2016-06-14
Elektra Qt Editor: add optional svg component in cmake

### LINKED ISSUE #796: svg module not added to COMPONENTS
Did you try to compile Elektra if the svg component is not installed?
(I did not find a way to install it on Debian, but it might be possible at other systems and is certainly is possible if you compile Qt yourself because it is an add-on).

I think it might need to be listed in QCOMPONENTS, what do you think?

sirblackheart:
should fix #796

### DIFF
--- a/src/tools/qt-gui/CMakeLists.txt
+++ b/src/tools/qt-gui/CMakeLists.txt
@@ -3,7 +3,8 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(DISCOUNT_FIND_QUIETLY ON)
 
 set(QCOMPONENTS Core Quick Gui Qml Widgets Test)
-find_package(Qt5 COMPONENTS ${QCOMPONENTS} OPTIONAL_COMPONENTS Svg)
+find_package(Qt5 COMPONENTS ${QCOMPONENTS})
+find_package(Qt5 OPTIONAL_COMPONENTS Svg QUIET)
 find_package(Discount)
 
 if (NOT Qt5_FOUND)
--- a/src/tools/qt-gui/CMakeLists.txt
+++ b/src/tools/qt-gui/CMakeLists.txt
@@ -3,7 +3,7 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(DISCOUNT_FIND_QUIETLY ON)
 
 set(QCOMPONENTS Core Quick Gui Qml Widgets Test)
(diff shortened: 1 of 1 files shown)


markus2330:
Is it necessary at all? Did you test it?

sirblackheart:
On my system it worked without the addition from the beginning. But my system also has a qt5-svg package.

sirblackheart:
So i need you to test it if it helps on your system where you found the problems.

sirblackheart:
No this does not work

sirblackheart:
Well it kind of does, it gives a warning because FindQt5Svg was not found if qt5-svg is not installed. If it is installed it works as expected. So it probably is the best way to solve this.

sirblackheart:
This should do it. And yes it works now as expected.

markus2330:
Is the additional (manual) message then necessary?

markus2330:
I think its okay when cmake outputs a warning that a OPTIONAL_COMPONENTS was not found for us. We should not circumvent what cmake wants to do for us.

sirblackheart:
It is how you should handle such cases as it is optional. Not finding the find file is more confusing to a user.

markus2330:
I am talking about the part where its made QUIET with a manual output later.

Why not avoid the manuel output and not use QUIET?

sirblackheart:
Because of the output it produces that is more confusing to a user then the manual message.
CMake complains about not finding the Find* file if not QUIET, that is basically meaningless for the user. Module not found makes more sense to him.

markus2330:
The users are maintainers who see such messages every day.

I would rather expect that they would overlook our tiny message, because they expect warnings of missing components to be presented in the standard way.

Lets simply output both messages, then every case should be covered.

sirblackheart:
Then please do that for everything, print clear warning messages when a binding gets excluded etc... And do not just make an exception for one case

sirblackheart:
> The users are maintainers who see such messages every day.

Very bad reasoning to produce a bad warning. Only because everyday cmake users are used to this kind of warning it can be a problem for a new user.

sirblackheart:
> I would rather expect that they would overlook our tiny message, because they expect warnings of missing components to be presented in the standard way.

Then lets open an issue and rework our cmake messaging system.

markus2330:
Ok, done in 351fa1f3a9b5048599794996772a54472d66d10f

PR #802, opened 2016-06-15
GSettings Backend: fix standalone build

sirblackheart:
When building the GSettings Backend without the rest of the Elektra
source tree there is nor target elektra headers. So we move it into the
corresponding if branch of the cmake instructions.


markus2330:
Thank you!

PR #804, opened 2016-06-16
crypto-plugin: Botan backend

petermax2:
[Botan](https://botan.randombit.net/) is a crypto library written in C++ and released under the BSD license.

The main advantage over OpenSSL and gcrypt is: it does not need initialization or destruction (for now, although they reserved this option for future releases). In terms of SLOC the code is pretty handy.

The only drawback is a memory leak in the Botan internals of `get_cipher` which I could not get rid of. I added a suppression for that.


markus2330:
Nice idea, what was the effort of creating a new backend? Why not trying to write a gpg backend? It would get rid of all memory problems and would even avoid dependencies completely.

Btw. no need to ping @manuelm, he gave me access to his machines too.
I installed it on all machines except debian-unstable-mm where I got the error:
```
E: Could not get lock /var/cache/apt/archives/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/cache/apt/archives/
```
Maybe @manuelm installed something concurrently?

src/plugins/crypto/testmod_crypto.c
```diff
@@ -187,6 +196,11 @@ static void test_config_errors ()
 	test_config_errors_internal ("crypto_openssl", newWorkingConfiguration (0), 1, "kdbSet failed with valid config");
 	test_config_errors_internal ("crypto_openssl", newInvalidConfiguration (), -1, "kdbSet succeeded with invalid config");
 	test_config_errors_internal ("crypto_openssl", newIncompleteConfiguration (), -1, "kdbSet succeeded with incomplete config");
+
+	// Botan tests
+	test_config_errors_internal ("crypto_botan", newWorkingConfiguration (0), 1, "kdbSet failed with valid config");
+	test_config_errors_internal ("crypto_botan", newInvalidConfiguration (), -1, "kdbSet succeeded with invalid config");
+	test_config_errors_internal ("crypto_botan", newIncompleteConfiguration (), -1, "kdbSet succeeded with incomplete config");
```
markus2330: Very elegant way to specify tests ;)

petermax2:
> what was the effort of creating a new backend? 

about two evenings and an afternoon (like 10 hours maybe?)

> Why not trying to write a gpg backend?

This is the next thing on my agenda. I really liked the Botan API, so I checked how much effort it would be to integrate it into Elektra.

markus2330:
jenkins build stable please

markus2330:
jenkins build please

markus2330:
Nice initiative! Hopefully it is better than the others.

Let us see what the build server says. Can you also add the dep to travis?

petermax2:
jenkins build stable please

mpranj:
crypto is not built on travis/OS X currently since it is `EXPERIMENTAL`.

petermax2:
Yikes, the Debian version of Botan is full of mem-leaks :cry: valgrind reported > 800 possible leaks in `testmod_crypto` on Jessie. The Fedora version is much cleaner (maybe it got updated more often).

petermax2:
I could not reproduce the SEGFAULT from the build server.

markus2330:
Do you want access to a build server? Did you try to reproduce on debian *jessie*?

Btw. @petermax2 you can drop experimental (temporarily) to also test it with mac os x if you want.

petermax2:
> Do you want access to a build server? 

That would be helpful. Or maybe you could send me a backtrace from `testmod_crypto` (for the elektra-mergerequests-stable build or the elektra-mergerequests build).

> Did you try to reproduce on debian jessie?

Yes, definitely _Jessie_ with the lastest updates.

petermax2:
Hold on, I finally managed to reproduce the error. I forgot to rebase with the master before building in Jessie.

petermax2:
Well... that was kind of unexpected... It is an OpenSSL error.

	#0  0x00007ffff690b600 in ?? ()
	#1  0x00007ffff63ad848 in ERR_get_state () from /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0
	#2  0x00007ffff63adb65 in ?? () from /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0
	#3  0x00007ffff7005ed5 in elektraCryptoOpenSSLHandleCreate (handle=handle@entry=0x7fffffffe498, config=config@entry=0x63cbf0, errorKey=errorKey@entry=0x60ae00) at /hom
	e/pnirschl/Projekte/libelektra/src/plugins/crypto/openssl_operations.c:166
	#4  0x00007ffff7006d66 in elektraCryptoEncrypt (errorKey=0x60ae00, data=0x60a960, handle=<optimized out>) at /home/pnirschl/Projekte/libelektra/src/plugins/crypto/cryp
	to.c:102
	#5  libelektra_cryptoopenssl_openssl_LTX_elektraPluginset (handle=<optimized out>, ks=0x60a960, parentKey=0x60ae00) at /home/pnirschl/Projekte/libelektra/src/plugins/c
	rypto/crypto.c:318
	#6  0x00000000004026ba in test_config_errors_internal (pluginName=pluginName@entry=0x4041a6 "crypto_openssl", pluginConfig=0x63cbf0, expectedResult=expectedResult@entr
	y=1, message=message@entry=0x4043b8 "kdbSet failed with valid config") at /home/pnirschl/Projekte/libelektra/src/plugins/crypto/testmod_crypto.c:178
	#7  0x0000000000401e3f in test_config_errors () at /home/pnirschl/Projekte/libelektra/src/plugins/crypto/testmod_crypto.c:196
	#8  main (argc=<optimized out>, argv=<optimized out>) at /home/pnirschl/Projekte/libelektra/src/plugins/crypto/testmod_crypto.c:284

manuelm:
> Maybe @manuelm installed something concurrently?

Nope. An old aptitude was hanging. Had to kill it.

manuelm:
@petermax2 Next up libsodium (https://download.libsodium.org/libsodium/content/secret-key_cryptography/aead.html with CHACHA20_POLY1305) ? :-)

petermax2:
I managed to get a consistent segfault behaviour on both Fedora and Debian. Let's call it a success, given that I woke up at 6 am.

Fedora 23 (with debuginfos installed)

	#0  0x00007fffefe2d6d0 in _ucdcmp_decomp () from /lib64/libkrb5.so.3
	#1  0x00007fffef88144c in ERR_get_state () at err.c:1021
	#2  0x00007fffef881766 in get_error_values (inc=0, top=0, file=0x0, line=0x0, data=0x0, flags=0x0) at err.c:807
	#3  0x00007ffff052cfa8 in elektraCryptoOpenSSLHandleCreate (handle=handle@entry=0x7fffffffdce8, config=config@entry=0x660670, errorKey=errorKey@entry=0x611db0) at /home/pnirschl/Projekte/libelektra/src/plugins/crypto/openssl_operations.c
	:170

Debian Jessie

	#0  0x00007ffff690b6c0 in ?? ()
	#1  0x00007ffff63ad848 in ERR_get_state () at err.c:1021
	#2  0x00007ffff63adb65 in get_error_values (inc=inc@entry=0, top=top@entry=0, file=file@entry=0x0, line=line@entry=0x0, data=data@entry=0x0, flags=flags@entry=0x0) at err.c:807
	#3  0x00007ffff63addc3 in ERR_peek_error () at err.c:769
	#4  0x00007ffff7005f95 in elektraCryptoOpenSSLHandleCreate (handle=handle@entry=0x7fffffffe498, config=config@entry=0x65dea0, errorKey=errorKey@entry=0x65efd0) at /home/pnirschl/Projekte/libelektra/src/plugins/crypto/openssl_operations.c
	:170

petermax2:
The segfault occurs in the function `ERR_get_state` in libcrypto at

    CRYPTO_THREADID_current(&tid);

This function _should_ use the callback, that has been set up by `CRYPTO_THREADID_set_callback` but somehow the callback is set to:

    (gdb) s
    CRYPTO_THREADID_current (id=0x7fffffffd940) at cryptlib.c:486
    (gdb) p threadid_callback
    $2 = (void (*)(CRYPTO_THREADID *)) 0x7fffefe2d750 <_ucdcmp_decomp+9616>

So instead of jumping into `openssl_operations.c` / `internalThreadId(...)`, OpenSSL jumps to:

    0x00007fffefe2d750 in _ucdcmp_decomp () from /lib64/libkrb5.so.3

petermax2:
jenkins build unstable please

petermax2:
I was investigating the memory issues on Jessie (both stable and unstable) and could only find a single problem: the call of `elektraCryptoGcryInit` which initializes libgcrypt. The libgcrypt manual strongly discourages the initialization of gcrypt from within a library. I know we had this discussion before, I just wanted to mention it.

Since there are many `???` in the valgrind output  of the build server, I wonder how I should write the suppresions file? @markus2330 could you help me, please?

markus2330:
If possible you can avoid the initialization if already done before. Otherwise, please create a decision and a dedicated github issue for further discussions on global initialization.

About valgrind: You can add `set( MEMORYCHECK_COMMAND_OPTIONS "--gen-suppressions=all" )` in the PR, this should give you the suppressions embedded in the xml file downloadable from the build server. Or you can also log in and run valgrind manually with the option.

petermax2:
I generated 10 MB of valgrind output on the build server. After filtering it for duplicates and adding the suppressions, valgrind still reports errors.

markus2330:
Mmh, does this project care about memory and/or valgrind at all? It might be best to ask  upstream. Maybe they do something in the global constructor (or other non-standard place) and a different kind of suppression would be easier. (You can also give valgrind instructions in the source code.)

Anyway if you want to pursue the same way: valgrind supports a kind of globbing for suppressions with `...`, which allow you to omit many similar problems from the same library (as specified with obj) See other valgrind suppressions as example, especially libgcrypt.so.20

petermax2:
> Mmh, does this project care about memory and/or valgrind at all?

I don't know but maybe my test setup is faulty? Could you please have a look at `testmod_crypto.c` and see if I do something wrong?

>  valgrind supports a kind of globbing for suppressions with ...

I had that in 4db8757 but the errors were still reported.

markus2330:
If `fun:main` leaks its some special constructor we cannot influence. I think it will already happen at `dlopen`, you can try to reduce the testcases to find out. As said: please contact upstream for such problems, they should free what they allocate.

petermax2:
They care a lot... NOT. I just ran valgrind over `gpg` encrypting some stuff:

    ==11814== ERROR SUMMARY: 1655103 errors from 214 contexts (suppressed: 1606 from 6)

Any further questions?

markus2330:
If you find no other way you can tag your testcases with MEMLEAK if Elektra is build with the problematic libraries.

petermax2:
I found no way around it. I tried linking `testmod_crypto` against `libgcrypt` and initializing the library wihtin `testmod_crypto` and not within the plugin itself. I got the same result. Even without running any tests, memory leaks occur. As soon as the library initializes (i.e. calling `gcry_check_version`) there are 6 entries in the loss records.

    ==22525== 72 bytes in 3 blocks are still reachable in loss record 6 of 6
    ==22525==    at 0x4C2BBAD: malloc (vg_replace_malloc.c:299)
    ==22525==    by 0x525BF20: ??? (in /usr/lib64/libgcrypt.so.20.0.4)
    ==22525==    by 0x525D29B: ??? (in /usr/lib64/libgcrypt.so.20.0.4)
    ==22525==    by 0x52FA0E1: ??? (in /usr/lib64/libgcrypt.so.20.0.4)
    ==22525==    by 0x52FA15F: ??? (in /usr/lib64/libgcrypt.so.20.0.4)
    ==22525==    by 0x525BE7F: ??? (in /usr/lib64/libgcrypt.so.20.0.4)
    ==22525==    by 0x525C004: ??? (in /usr/lib64/libgcrypt.so.20.0.4)
    ==22525==    by 0x402429: main (testmod_crypto_gcrypt.c:291)

markus2330:
In the current PR the MEMLEAK is not used? Is it ready as is? Please make sure to set MEMLEAK only if botan variant was added. On debian unstable the `testmod_` currently do not compile at all. I think you need `add_plugintest` so that you can pass `LINK_PLUGIN "<no>"`.

petermax2:
I fixed the compilation issue on Debian Unstable. Why does `testkdb_allplugins` execute tests that are marked as "MEMLEAK"? Or did I misuse the MEMLEAK tag? Maybe my `src/plugins/crypto/CMakeLists.txt` is faulty??

Beside the fact that ALL our crypto libraries are full of memory leaks, the PR is ready. I think we can remove all the libraries with the GPG interface as soon as I am done with my benchmarks. What do you think?

markus2330:
`testkdb_allplugins` has its own filter directly in the source, simply add your plugin there, too.

Why do you mark all three plugins with MEMLEAK, I thought only botan has these issues?

petermax2:
> testkdb_allplugins has its own filter directly in the source, simply add your plugin there, too.

Thank you!

> Why do you mark all three plugins with MEMLEAK, I thought only botan has issues not suppressable with valgrind?

My bad, only gcrypt does the "special" initialization stuff.

Travis reports an Augeas error (wasn't me ;-) ). Let's wait for the other build jobs to finish.

petermax2:
This PR affects #731 .

markus2330:
The buildserver agent that should build the merge requests is offline. Should I merge it blindly (and then maybe create issues for memleaks) or should we rather wait until it is up again?

petermax2:
I generated suppressions for unstable and stable. In my VMs no memory leaks are reported when running `make run_memcheck`. You can send me an issue any time (but I won't respond from to Jul 16th to Jul 30th because I am on vacation). It's OK for me if you prefer to wait for the build servers to come up again.

markus2330:
Thank you!

PR #806, opened 2016-06-20
shell recorder: Add testcases to "make test"

KurtMi:
all files with *.dat ending in the shell_recorder folder will be
executed at make test.


markus2330:
Thank you! Looks good, test cases are running now. It seems like some are failing though.

KurtMi:
Yes, I have not looked at the shell recorder syntax yet, I try to fix them when I write more testcases.
For now it would be nice if @tom-wa could review the `shell_recorder.sh` just to be sure, I did not messed up something.

tests/shell/shell_recorder/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+if (ENABLE_KDB_TESTING)
+        # set kdb command
+        SET(USE_CMAKE_KDB_COMMAND "")
+        if (BUILD_FULL)
```
markus2330: can you change the order to shared, full, static (same as for linking)

markus2330:
Please rebase.

Why did you rename files to "notworking"?

tests/shell/shell_recorder/output
```diff
@@ -1,11 +0,0 @@
-set system/testmount/test/test bla
-mount /tmp/test.dump system/testmount/test dump
-set system/testmount/test teststring
```
markus2330: why remove the file?

tests/shell/shell_recorder/shell_recorder.sh
```diff
@@ -255,29 +279,36 @@ echo "protocol file: $OutFile"
...
+        printf "\n"
         EVAL=1
     else
-         echo -e "=======================================\nReplay test succeeded"
+         printf "=======================================\nReplay test succeeded\n"
     fi
 fi
 
 
 rm ./stdout 2>/dev/null
```
markus2330: this should be in temporary files, and/or in a trap exit

tests/shell/shell_recorder/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+        else()
+                message(SEND_ERROR "no kdb tool found, please enable BUILD_FULL, BUILD_STATIC or BUILD_SHARED")
+        endif ()
+        configure_file (
+                        "${CMAKE_CURRENT_SOURCE_DIR}/shell_recorder.sh"
+                        "${CMAKE_CURRENT_BINARY_DIR}/shell_recorder.sh"
+                        @ONLY
+                )
+
+        file (GLOB SCRIPT_TESTS *.dat)
```
markus2330: @tom-wa @KurtMi What is your opinion about the dat file ending?

Maybe Elektra Shell Recorder (.esr)?

tom-wa: to me .dat is like foo or bar, i just needed a fileending and didn't want to think about it. .esr seems fine to me.

KurtMi: `.esr` is it!

KurtMi:
The testcase in the notworking directory will never work in all cases (make test and ./shell_recorder.sh testcase.esr) because it mounts a ini file and the path to the ini file is on both cases different. I am not sure if this can be fixed, maybe @tom-wa has any idea.

markus2330:
Ok, then let us disable the test for now.

Why do we have `highlevel.ini` (and others) twice (they are also in `examples/spec`) if they are not working for shell recorder?

KurtMi:
I took the files from @tom-wa. For now I removed the `highlevel.ini`.

markus2330:
Ok, thank you. Please ping if ready to merge.

tests/shell/shell_recorder/shell_recorder.sh
```diff
@@ -1,12 +1,24 @@
...
+}
+
 FILE=$1
 Mountpoint=
 DBFile=
 Storage=
 MountArgs=
 DiffType=File
-OutFile=$(mktemp "${TMPDIR:-/tmp}/tmp.XXXXXXXXX")
+OutFile=$(mktemp "$(getTempDir)/tmp.XXXXXXXXX")
```
markus2330: You can leave it as is, but in the long term it might be a good idea to use include_common also for shell_recorder.

It has a mktempdir_elektra and mktempfile_elektra which does what you want here.

KurtMi: The usage of include_common would lead to a shell recorder only usable with cmake. My target is to support both (also direct execution).

KurtMi: Or a hack like:
```
if [ -z "@USE_CMAKE_KDB_COMMAND@" ]; then
        KDBCOMMAND="@KDB_COMMAND@"
else
        KDBCOMMAND="kdb"
fi
```
would be needed

sanssecours: Does the command 
```bash
mktemp -t tmp
```
work for your use case? The above command returns a file located at a path that looks something like this: `/var/folders/hx/flbncdhj4fs87095gzxvnj3h0000gn/T/tmp.RrhXNVNZ` on my computer (OS X).

KurtMi: not sure, I use now
`mktemp -t elektraenv.XXXXXXXXX 2>/dev/null || mktemp -t 'elektraenv'`
which file will be created with this command?

sanssecours: > not sure, I use now
`mktemp -t elektraenv.XXXXXXXXX 2>/dev/null || mktemp -t 'elektraenv'`
which file will be created with this command?

The command `mktemp -t elektraenv.XXXXXXXXX 2>/dev/null` creates something like:
```
/var/folders/hx/flbncdhj4fs87095gzxvnj3h0000gn/T/elektraenv.XXXXXXXXX.sTrHgctr
```
, while `mktemp -t 'elektraenv'` creates something like

```
/var/folders/hx/flbncdhj4fs87095gzxvnj3h0000gn/T/elektraenv.h0Zg7gVB
```

The overall result will be the first output, since `mktemp -t 'elektraenv'` works without problems.

markus2330: Thanks for finding out!

@KurtMi Can you please add this description as code comment in the functions?

tom-wa:
@KurtMi the listtest actually works, i just broke it when i introduced the presetcleanup placement to list for the global plugin cleanup and forgot to update the test. 
the MountArgs line should read `MountArgs: list placements=,placements/set="presetstorage presetcleanup precommit postcommit",plugins.....`

KurtMi:
@tom-wa your change does not work, but your help is really appreciated, It is not easy to write or check test cases from software someone other wrote.

KurtMi:
@tom-wa I found out that I broke the shell recorder, therefore the test cases will not work.
The problem is that mac osx does not have a `-z` option for the `grep`. Do you know any solution?

tom-wa:
you are right, for some reason i had timeofday in presetcleanup too. the testcase is ok, but the list plugin has an issue. i'll try to fix it.

it won't work without -z. -z tells grep that lines end with \0 instead of \n

KurtMi:
@tom-wa would something like this work:
```
echo "$WARNINGS" | grep -Eqz --text "($WARNINGSCMP)"

to

echo "$WARNINGS" | grep -Eq --text "($WARNINGSCMP)" | tr "\n" "\0"
```
?

tom-wa:
not really, that would replace all \n with \0. i use -z for multiline matchings, so \n gets ignored, and all "sections" read until a \0 is encountered

tom-wa:
btw i fixed the issue with the listtest in #854
the original testcase was fine, but there was an issue with the placement handling in the list plugin

markus2330:
@tom-wa there is no #854?

@KurtMi Is --null-data also not available?

KurtMi:
> @KurtMi Is --null-data also not available?
No

tom-wa:
typo, meant #845

markus2330:
Ok, #845 looks nice, but please improve commit message!

@KurtMi  should I merge it before this commit?

KurtMi:
> @KurtMi should I merge it before this commit?

I am not sure, but I think it is irrelevant, the worst thing that can happen is a broken shell recorder test case.

KurtMi:
@tom-wa I found a reddit post which says that `grep -z` should be equivalent to `zgrep`. What di you think?
https://www.reddit.com/r/commandline/comments/4jydnw/a_list_of_differences_from_gnulinux_in_os_x_in/

KurtMi:
> not really, that would replace all \n with \0. i use -z for multiline matchings, so \n gets ignored, and all "sections" read until a \0 is encountered

@tom-wa what is wrong with the following try:
`echo "$STDERR" | tr "\n" "" | tr "\0" "\n" | grep -Eq --text "$STDERRCMP"`

tom-wa:
i'm not sure if it works with multiline outputs too. you can test it with #845 and the listtest script to see if that line: `STDERR: (set.*di.*pos.*){3}` works.

KurtMi:
@tom-wa next problem mac and freebsd does not support `grep -P`, you used it in two situations:
```
WARNINGS=$(echo "$STDERR" | grep -Po "(?<=Warning number: )([[:digit:]]*)" | tr '\n' ',')
ERRORS=$(echo "$STDERR" | grep -Po "(?<=Error \(\#)([[:digit:]]*)" | tr '\n' ',')
```
and change to get rid of the `-P`?

markus2330:
If the effort is too high or the linux implementation would suffer (e.g. not handle some cases that worked before), we can also leave Mac OS X for now. It would be nice though, because hopefully the shell recorder will replace most of the other shell unit tests soon. One idea of shell recorder was that you do not have to worry about portability when writing unit tests.

tom-wa:
grep -P is needed for lookahead and lookbehind assertions.
theres a simple fix though, tell os x users to install gnu grep. afaik it's available with whatever their third party package manager is called.

markus2330:
Yes, I think a dependency to GNU grep is ok, it is much better than the -z workaround and excluding shellrecorder from mac os x.

@mpranj Can you fix travis to use GNU grep?

sanssecours:
Maybe we can just replace 
```sh
grep -Po "(?<=Warning number: )([[:digit:]]*)"
``` 
by something  like 
```
sed -nE  "s/Warning number: (\d*)/\1/p"
```
? How about 
```
sed -nE  "s/Error \(\#(\d*)/\1/p"
```
instead of
```
grep -Po "(?<=Error \(\#)([[:digit:]]*)"
```
? Do we really need the lookbehind assertion?

markus2330:
@sanssecours What is your opinion about installing GNU grep? Its not only about the lookbehind assertion, but every second flag (e.g. -z, -P) is not supported.

sanssecours:
> @sanssecours What is your opinion about installing GNU grep?

I do not think that we need to do that. As far as I know I never installed a tool that required the GNU version of `grep`.

> Its not only about the lookbehind assertion, but every second flag (e.g. -z, -P) is not supported.

I think that “every second flag“ is a little bit of an exaggeration ☺️. I took the work by Kurt and used the workaround proposed in my comment above. As a result only two of the shell tests fail on my machine:

```
        Start  66: testshell_db_changes
 66/100 Test  #66: testshell_db_changes ...............   Passed    0.18 sec
        Start  67: testshell_globaltest
 67/100 Test  #67: testshell_globaltest ...............   Passed    1.05 sec
        Start  68: testshell_hosts
 68/100 Test  #68: testshell_hosts ....................***Failed    0.39 sec
        Start  69: testshell_listtest
 69/100 Test  #69: testshell_listtest .................   Passed    0.32 sec
        Start  70: testshell_profiletest
 70/100 Test  #70: testshell_profiletest ..............***Failed    1.55 sec
        Start  71: testshell_script
 71/100 Test  #71: testshell_script ...................   Passed    0.48 sec
```

KurtMi:
> I took the work by Kurt and used the workaround proposed in my comment above. As a result only two of the shell tests fail on my machine

As far I know these two test cases fail also with the `-P`.

markus2330:
@sanssecours You are right, its only 1/3 of the used flags (EwqzPo) ;)

@KurtMi But the tests do not fail on linux? Is the PR ready to merge?

sanssecours:
> As far I know these two test cases fail also with the `-P`.

Thanks for the information. On my machine all of the shell tests failed at the start, since my build directory name [contains spaces](https://github.com/sanssecours/libelektra/commit/6338053b5eae80fb501578f57b1e9e20aa610b95). I just assumed that the tests would fail on OS X otherwise too.

markus2330:
What about adding a build job where the directories contain spaces?

sanssecours:
Currently one of the problems on OS X seems to be the line:

```sh
sed -e :a -e N -e '$!ba' -e 's/\n//g'
```

At least that seems to be the case, according to parts of the debug output of `testshell_hosts`:

```
68: + echo 127.0.0.1
68: + sed -e :a -e N -e '$!ba' -e 's/\n//g'
68: + grep -Eqa 127.0.0.1
68: + '[' 1 -ne 0 ']'
68: + echo 'STDOUT doesn'\''t match 127.0.0.1'
68: STDOUT doesn't match 127.0.0.1
68: + printf '%s\0' '=== FAILED stdout doesn'\''t match expected pattern 127.0.0.1'
68: + nbError=1
```

The following command fails on my machine:

```sh
echo 127.0.0.1 | sed -e :a -e N -e '$!ba' -e 's/\n//g' | grep -Eqa 127.0.0.1
```
. If I remove the invocation of `sed` the command returns successfully
```
echo 127.0.0.1 | grep -Eqa 127.0.0.1
```
. Maybe BSD `sed` and GNU `sed` behave differently? Does the first command work on Linux?

> What about adding a build job where the directories contain spaces?

Sounds like a great idea. I think it would be even better, if we just use something like `build directory` for all build jobs instead of `build`. Then we do not need to add any additional jobs.

KurtMi:
Thank you @sanssecours for your effort, but for me it is impossible to give Mac os x support working on a Linux machine.
I included your commits, the one with the spaces in the build folder and the one with the white spaces.
For now Mac os x will be excluded for the following reasons, first @tom-wa  and I need this PR to work on the shell recorder and second I don have a mac and it is is really a pain in the ass if you need to wait 25 minutes for the travis build job and the only information you get is that something is not working, but not exactly what.
I made a issue #858, which summarizes all things that need to be changed for the OSX support.
@markus2330 The PR is rdy!

markus2330:
Thank you all for your efforts!

@sanssecours Feel free to pick up #858 if you already have a solution. We can also review it. A shell recoder PR would be interesting.

PR #808, opened 2016-06-21
GSettings Backend: fix double free on get writable

sirblackheart:
Sorry this is a bad one that slipped us on 5b3aa15ad7b03812fb564db6c8de523bd8ee50f7


markus2330:
Thank you! Can you add some regression tests? Or do you think it would be easy to run the gsettings-testcases also on the build-server?

sirblackheart:
Sadly the GLib2 GSettings testcases do not even catch this one, therefore I overlooked it. There seems to be a glib2.0-tests package for debian, but it does not look like it contains the GSettings tests so we are probably safer of if i write some tests.

sirblackheart:
@markus2330 One problem is the tests are licensed GPL. One approach could be to create a git repository only with the existing glib gsettings tests. It if someone sets cmake do the gsettings tests it would be pulled down, compiled and run, but it would neither be directly distributed with the source or binaries of elektra.

markus2330:
Yes, it would be useful if you extend the tests with some white-box knowledge of your bindings. If they alone are not enough we can look for further options, maybe the glib people are willing change the licence for the tests or we find another way how to use them on the build-server.

PR #816, opened 2016-06-29
Use cmake GNUInstallDir for /etc/*.d files

sirblackheart:
Using the GNUInstallDir module of cmake to make the installation of this
two files honor the CMAKE_INSTALL_PREFIX variable while staying conform
to GNU and FHS.

> It does not make it more standard and as you already noted the files
> would not be used anyway, so the out-of-the-box behavior is
> identical to not installing the files at all.

That is untrue. It makes it more standard. It is correct to install to `/usr/local/etc` if it is not a package manager installation. Further it is my responsibility as user to source them then. This makes it work with other software installed the same way.

> Something that would improve usability is a script that installs and
> setups the bash completion for the specific user.

That in my opinion is wrong and overcomplicated. If i install things to `/usr/local` i should be aware of the installed files. As they are manually installed files.

> I dont know where you get this from, in the link above (FHS 3.0 and
2.3) it says "/usr/local/etc may be a symbolic link to /etc/local",
not that it is a directory that should be used for installation.

>Also I do not find your `/opt/etc`, I only find references to
`/etc/opt`.

If you read my source reference you see that i made a and switched `/opt/etc` for `/etc/opt` as a mistake.
The GNUInstallDirs variable should behave as expected from FHS. If not we should create a ticket to resolve the issue.

>And /usr/local today is basically identical to /opt, the original
usage of /usr as NFS mounted and /usr/local as locally mounted
partitions is mostly historically. In my opinion correctly packaged
software should be installed to /usr. But this discussion is not
leading to anything.

let as take a look at the stackexchange page with 121 votes for the correct answer:
___
> While both designed to contain files not belonging to the Operating System, /opt and /usr/local are not designed to contain the same set of files.

> /usr/local is a place to install files built by the administrator usually by using the make command (eg: ./configure; make; make install). The idea is to avoid clashes with files that are part of the operating systems which would either be overwritten or overwrite the local ones otherwise. eg. /usr/bin/foo is part of the OS while /usr/local/bin/foo is a local alternative.

> All files under /usr are shareable between OS instances although this is rarely done with Linux. This is a part where the FHS is weak, or at least self-contraditory, as /usr is defined to be read-only but /usr/local/bin needs to be read-write for local installation of software to succeed. The SVR4 file system standard which was the FHS main source of inspiration is recommending to avoid /usr/local and use /opt/local instead to avoid this issue.

> /usr/local is a legacy from the original BSD. At that time, the source code of /usr/bin OS commands were in /usr/src/bin and /usr/src/usr.bin while the source of commands developed locally was in /usr/local/src and their binaries in /usr/local/bin. There was no notion of packaging (outside tarballs).

> On the other hand, /opt is a directory where to install unbundled packages, each in its own subdirectory. They are already built whole packages provided by an independent third party software distributor. Unlike /usr/local stuff, these package follow the directory conventions. For example someapp would be installed in /opt/someapp with one of its command being /opt/someapp/bin/foo, its configuration file would be in /etc/opt/someapp/foo.conf, and its log files in /var/opt/someapp/logs/foo.access.

___

But anyway GNUInstallDirs should handle both `usr/local` and `opt` correctly according to the documentation.

> I am not talking about complete introduction at once, but to
completely think it through. And finding someone who is willing to
do it the full way. But I *really* think we should concentrate on
Elektra's goals, not on some beautification of the build system.

Just because you do not share the value in resolving this issue  does not mean it does not have a value to be solved. Maybe just not from your perspective. And it is not a beautification but a valid patch to solve to issues.

> Maintainers are much more pragmatic on such things. Their needs are
satisfied, so we should really listen to them.

I am the maintainer (homebrew package) in this case and I am not happy with the current solution.

> It is a major change interfering with most parts of the build
> system:
> 
> 1. You get conflicts with KDB_DB_SYSTEM

please elaborate?

> 2. configure and INSTALL documents get more complex

No they do not. This variable is neither shown in the cmake gui or documentation by default. We just honor the CMAKE_INSTALL_PREFIX.

> 3. Users have to deal with more options to set

One optional option which is set to a sane default for GNU and FHS and not visible in the cmake gui by default.

> 4. Confused users why some GNUInstallDirs work, and others not (or
> many other conflicts with the current way libraries and modules are
> installed if we fully go for GNUInstallDirs)

If someone wants to override it he must explicitly know that it is used at that location. The main reason to introduce this change is to get correct behaviour for the CMAKE_INSTALL_PREFIX.

> So you introduce a lot of complexity without eliminating the need to
still manually move files.

Honestly I still can not see all this complexity you are talking about.

> And it actually does not solve an issue, it only installs files at
> places where they are not used instead of not installing them at
> all.

It actually does, by installing it to the correct place. It is then up to the user to source it. Again when installed to with `/usr/local` as prefix you should never install something to `/etc as everything in `etc`should be only installed by the package manager…

And just to make it clear and we are not misunderstanding:

``` shell
CMAKE_INSTALL_PREFIX=/            => CMAKE_INSTALL_FULL_SYSCONFDIR=/etc
CMAKE_INSTALL_PREFIX=/opt/bla/blu => CMAKE_INSTALL_FULL_SYSCONFDIR=/etc/opt/bla/blu
CMAKE_INSTALL_PREFIX=/usr         => CMAKE_INSTALL_FULL_SYSCONFDIR=/etc
CMAKE_INSTALL_PREFIX=/usr/local   => CMAKE_INSTALL_FULL_SYSCONFDIR=/usr/local/etc
```

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -23,12 +23,13 @@ install (PROGRAMS upgrade-bootstrap DESTINATION ${TARGET_TOOL_EXEC_FOLDER})
 
 install (PROGRAMS umount-all DESTINATION ${TARGET_TOOL_EXEC_FOLDER})
 
+include (GNUInstallDirs)
 if (INSTALL_SYSTEM_FILES)
 	install (FILES kdb-bash-completion
-			DESTINATION /etc/bash_completion.d
+			DESTINATION ${CMAKE_INSTALL_FULL_SYSCONFDIR}/bash_completion.d
 			RENAME kdb)
 	install (FILES elektraenv.sh
-			DESTINATION /etc/profile.d
+			DESTINATION ${CMAKE_INSTALL_FULL_SYSCONFDIR}/profile.d
 			RENAME kdb.sh)
 endif()
 


markus2330:
Could you please state more clearly Why, What and How you want to propose? (Also long-term beyond content of PR) (Ideally with decision template). The details of the discussion should be clearly separated from the proposal outline, and it should be clear where there is a (dis)agreement.

markus2330:
Btw. the usage of the PR really biases a vote. It is not about these two arbitrary files (which are 	particularly unimportant, but if we should use GNUInstallDir. Furthermore the PR is uncomplete: docu, configure,... is missing.

And especially the head line really confuses: KDB_DB_SYSTEM is not addressed at all (which everyone hopefully thinks about when /etc is mentioned together with Elektra). Please use an issue and not an PR.

markus2330:
One small note (I am in a hurry): the idea was to write a script to install the bash completion file to the users home directory into `~/.bash_completion`.

sirblackheart:
> Could you please state more clearly Why, What and How you want to propose? (Also long-term beyond content of PR) (Ideally with decision template). The details of the discussion should be clearly separated from the proposal outline, and

for discussion see #754 
I see neither the need to add any more comments nor decision template for such a trivial change. For me this a trivial change without any long term impact. From my point of view it fixes a broken system. It is not about any further long term decisions. As clearly stated this is just about the installation of this two files. The idea of using GNUInstallDirs for something else is neither part of my intention nor goal. If so I would create the appropriate research and decision document.

>  it should be clear where there is a (dis)agreement.

Disagreement:
You say it complicates build system, whereas I do not see any direct effects on this as most of your arguments where untrue:
You: More variables for user -> fact: not directly visible and must not be exposed or mentioned
You: Does not behave correctly with /opt -> fact: the cmake module implements this correctly

Where there is still the biggest disagreement is if such files should installed inside `/opt/modulname/etc/*.d/file` or `/usr/local/*.d/file` or not. Here our opinions really differ with both having valid points.
Yours: mostly is that it wont get sourced automatically. Mine: That the user have to source them them-self for prefixes like `/usr/local`. Package manger like homebrew on OS X use this folders. e.g. `/usr/local/etc/bash_completion`

> Btw. the usage of the PR is not suitable for a vote. It is not about these two arbitrary files (which are particularly unimportant)

It is about this two files and how they are handled. e.g. absolute /etc vs a calculated etc that honors the CMAKE_INSTALL_PREFIX. Other files that are still part of the almost undocumented variable INSTALL_SYSTEM_FILES are the GIR file and the GSettings Module. As for the  GSettings Module is still WIP and I probably will look into the options we have here. 

> but if we should use GNUInstallDir.

GNUInstallDir just has a nice working solution for a conform calculation of etc with the CMAKE_INSTALL_PREFIX. It was never in my scope to introduce it in general other then for this case. As it seems to be a default module of cmake I see not downfalls of using it for this case.

For someone wanting to use it for other install targets a different PR or Issue should be created.

>Furthermore the PR is uncomplete: docu, configure,... is missing.

Then you clearly misstated what you wanted. That really is a problem when this is happening to often. Either clearly state what you want or do not expect work to be made that was never clear in this context. As of your statement:

> Please create a new PR if GNUInstallDirs is important for you, we
will have to repeat the whole discussion again there. Please do not
continue the discussion here.

None of those ideas where mentioned. This seems way out of the scope and goal of my PR.

"If there was no documentation needed before this commit then there is non needed after it." But yes someone should document this files and the INSTALL_SYSTEM_FILES variable properly but that is out of the scope of this PR. Please create the appropriate tickets as beginner friendly tasks.

> And especially the head line really confuses: KDB_DB_SYSTEM is not addressed at all (which everyone hopefully thinks about when /etc is mentioned together with Elektra). Please use an issue and not an PR.

So you know what everyone thinks about… (e.g. projecting your way of thinking on everyone) Please follow your own suggestion and do not get personal so I do not have to get personal.

Also from #754 
> I hope you reconsider and focus on non-build tasks. There is a lot of work there and not much to be gained.

Please keep in mind that different people have different values. By this sentence you are being very rude, judging and again getting personal. You are openly saying that my personal values and goals are inferior to yours.
I value to do things the way that seems right to me and I will invest time and energy if I see an opportunity to do so. If you tell me you will block this, no mater what the gain is I can leave it at that, as I have to honor your decision as project maintainer, but also have shown my interest and reasoning to do what is right in my opinion.

> One small note (I am in a hurry): the idea was to write a script to install the bash completion file to the users home directory into ~/.bash_completion.

Out of the scope and idea of this PR.

markus2330:
I clearly stated that I disliked the approach and you should not put effort into it. Maybe I was unclear about creating the decision template and about how to vote about it, the process is not yet defined (see #819).

The main issue about CMAKE_INSTALL_FULL_SYSCONFDIR is how Elektra would be able to find configuration files it installed there.  Concededly`KDB_DB_SYSTEM` is far from perfect. But installing config files without thinking about how Elektra can find them is not a good idea.

I hope we agree now on the simplest solution: Not installing anything to `/etc`, we just have to get rid of #833 and the fallback for #832.

PR #822, opened 2016-07-02
Use bash completion folder

sirblackheart:
Get correct path for bash-completion files
The new approach is to look first if there is a cmake file, it then
falls back to pkg-config. As last fallback we use the hardcoded
`/etc/bash-completion.d`.

relies on PR #821
and is an alternative approach to PR #820

Update documentation for INSTALL_SYSTEM_FILES
Explain behavior of the variable and the files affected by it in
COMPILE.md


markus2330:
Thank you for adding the docu and finding out about completionsdir. It will now be installed to `/usr/share/bash-completion/completions` instead of `/etc/bash_completion.d`, which seems to be the better choice anyway. Is it somehow possible to let pkg-config respect `CMAKE_INSTALL_PREFIX`. It would be nicer then the `BASH_COMPLETION_COMPLETIONSDIR` hack (see #832).

I think we can simply remove the elektraenv.sh, then we would avoid installing anything to /etc. (see #833)

sanssecours:
> It will now be installed to `/usr/share/bash-completion/completions` instead of `/etc/bash_completion.d`, which seems to be the better choice anyway.

Looks like `/usr/share` is protected by the “[System Integrity Protection](http://apple.stackexchange.com/questions/193368/what-is-the-rootless-feature-in-el-capitan-really)“ feature in OS X. This means even `root` is not allowed to write to this folder. How about `/usr/local/share/bash-completion/completions` instead?

markus2330:
Does this path work for the bash completion (please verify if this is true for your case)? Maybe its a bug in the bash-completion in Mac OS X: `pkg-config` should return a path that is usable.

Please open a separate issue after investigation what would be useful for Mac OS X. (Just installing for the installing sake is not useful, the completion must work out-of-the-box. To avoid installation we already have INSTALL_SYSTEM_FILES)

sanssecours:
> Does this path work for the bash completion (please verify if this is true for your case)? Maybe its a bug in the bash-completion in Mac OS X: `pkg-config` should return a path that is usable.

After I install `bash-completion` via Homebrew, the command `pkg-config --variable=completionsdir bash-completion` returns the following:

```
Package bash-completion was not found in the pkg-config search path.
Perhaps you should add the directory containing `bash-completion.pc'
to the PKG_CONFIG_PATH environment variable
No package 'bash-completion' found
```
. I then downloaded the latest release [release](https://github.com/scop/bash-completion/releases) of `bash-completion` and ran the configuration step to create the file `bash-completion.pc`. After I set `PKG_CONFIG_PATH` to the location of `bash-completion.pc`, `pkg-config --variable=completionsdir bash-completion` prints the following path:

```
/usr/local/share/bash-completion/completions
```
. Homebrew on the other hand seems to prefer `/usr/local/etc/bash_completion.d`. At least thats what line 43 of `bash-completion` tells me:

```bash
[ -n "$BASH_COMPLETION_DIR" ] || BASH_COMPLETION_DIR=/usr/local/etc/bash_completion.d```
```
And indeed if I follow the steps printed by `brew info bash-completion`:

```
Add the following lines to your ~/.bash_profile:
  if [ -f $(brew --prefix)/etc/bash_completion ]; then
    . $(brew --prefix)/etc/bash_completion
  fi

Bash completion has been installed to:
  /usr/local/etc/bash_completion.d
``` 
, then the completions inside `/usr/local/etc/bash_completion.d` work.

> Please open a separate issue after investigation what would be useful for Mac OS X. (Just installing for the installing sake is not useful, the completion must work out-of-the-box. To avoid installation we already have INSTALL_SYSTEM_FILES)

Okay. I opened a new issue [here](https://github.com/ElektraInitiative/libelektra/issues/844).

Issue #825, opened 2016-07-09
plugin: internalnotification


waht:
@markus2330 asked me to create an [internalnotification plugin](https://github.com/waht/libelektra/tree/internalnotification/src/plugins/internalnotification) which updates variables automatically when registered keys are changed on kdbGet/kdbSet.

For now the plugin allows to register integer variables. It is assumed that the key value is a string containing a number.

The plugin is located in the [internalnotification](https://github.com/waht/libelektra/tree/internalnotification)-branch of my libelektra-fork.
I also created a [testbed program](https://github.com/waht/internalnotification-test).

Unit tests and memcheck are fine, but in the testbed only every second kdbGet triggers the plugin, the first kdbSet does not trigger (but afterwards it works fine).

Due to these issues I did not create a pull request yet.
Maybe the testbed program is flawed, but the placement of the plugin could also be wrong (postgetstorage for kdbGet, postcommit for kdbSet).

Maybe someone can help.

markus2330:
Do not worry about any issues, you can still create a PR, only mark it accordingly (WIP in title or "in progress" tag). Within an PR its easier to discuss the code and we can trigger the build server.

About kdbGet/kdbSet not happening: In the API usage of kdbGet()/kdbSet() you need to consider the state diagram doc/images/state.png. kdbSet() always needs a prior kdbGet(). kdbGet() might be a NOP if it detects that the configuration files are unchanged.

postgetstorage and postcommit sound ok.

Please integrate all tests within the same repo. And please run clang-format before creating the PR.

waht:
Allright, I created a PR :)
I did not include the testbed for now, since it is interactive and only useful for manual testing. Should I move it to the repository?

Regarding the state diagram, the testbed seems fine. There is an initial kdbGet() in main and the returned KeySet is passed to subsequent kdbSet/Get calls. I constructed a case where kdbSet() succeeds (the file is changed, plugin not triggered since it is the first kdbSet()) and the next kdbGet() does not trigger the plugin:

```
Press t to print "temperature"
Press g to kdbGet and print "greeting" key
Press s to set "temperature" key to a random value

elektraInternalnotificationOpen
elektraInternalnotificationRegisterInt variable=0x7ffd6b368d20, keyName=/sw/waht/elektra-test/temperature
elektraInternalnotificationRegisterInt registered key (name=/sw/waht/elektra-test/temperature, variable=0x7ffd6b368d20)
temperature variable located at 0x7ffd6b368d20
[...]

PR #828, opened 2016-07-10
Internalnotification Plugin

waht:
PR for internalnotification plugin (see #825)


markus2330:
add to whitelist

src/plugins/README.md
```diff
@@ -165,6 +165,10 @@ Log/Send out all changes to configuration to:
 - [syslog](syslog/)
 - [logchange](logchange/) prints the change of every key on the console
 
+Notification of key changes:
+
+- [internalnotification](internalnotification/) automatically update integer variables when a key is changed from inside an application
```
markus2330: keep the short description more generic. I think "automatically update registered variables"

src/plugins/internalnotification/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+include (LibAddMacros)
+
+IF (NOT PLUGIN_INTERNALNOTIFICATION_VERBOSE)
+	SET (PLUGIN_INTERNALNOTIFICATION_VERBOSE "0" CACHE STRING "Enable verbose output for internalnotification plugin")
+ENDIF (NOT PLUGIN_INTERNALNOTIFICATION_VERBOSE)
+
+MARK_AS_ADVANCED(
+	PLUGIN_INTERNALNOTIFICATION_VERBOSE
```
markus2330: try to avoid the variable, simple use VERBOSE set in kdbconfig.h.
Improvements of Elektra's "logger" are very welcomed.

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,37 @@
...
+- infos/provides =
+- infos/recommends =
+- infos/placements = postgetstorage postcommit
+- infos/status = global libc unittest concept unfinished experimental nodoc
+- infos/metadata =
+- infos/description = Plugin for internal notification
+
+## Usage ##
+
+Allows programs to register integer variables that get automatically updated
```
markus2330: Do not mention integer here, it is already mentioned in Todo/Issues

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,37 @@
...
+when the value of the key changes.
+
+Use as global plugin:
+$ sudo kdb global-mount internalnotification
```
markus2330: Please also give a code example how to use it.

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Exported Methods ##
+
+This plugin exports the function `int elektraInternalnotificationRegisterInt (Plugin * handle, int * variable, Key * key)`.
```
markus2330: How would a user get a handle to the plugin? You should write a wrapper that takes KDB and does the lookup to your plugin's handle.

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,37 @@
...
+If the given key is contained in a KeySet on a kdbGet or kdbSet operation, its value is
+converted from string to integer and the registered variable is updated with the current value.
+This also works if the registered key is a cascading key.
+The function address is exported as `system/elektra/modules/internalnotification/exports/elektraInternalnotificationRegisterInt`.
+The plugin handle required for this plugin is exported as `system/elektra/modules/internalnotification/exports/handle`.
+Please note that the plugin API may change as this plugin is experimental.
+
+## Todo/Issues ##
+
+- Correct data storage for registered keys? elektraPlugin(Set|Get)Data?
```
markus2330: Isn't this done already?

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,37 @@
...
+- Can we get rid of exporting the plugin handle?
```
markus2330: Yes. The idea was that the user calls an initialization function, this function registers itself in _KDB->globalPlugins. That is also the place where you can lookup the handle.

So the user links against your plugin, global mounting of it is not required then.

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,37 @@
...
+- Different data types for values:
+  - integer: Before conversion: check if key is string, if key is binary, just use the value
+  - string: Update string value
```
markus2330: binary data is not so important, string would be useful though

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,263 @@
...
+	{
+		// Make new item end of list
+		listPointer->tail->next = item;
+		listPointer->tail = item;
+	}
+
+	return item;
+}
+
+static void elektraInternalnotificationUpdateRegisteredKeys (ListPointer * listPointer, KeySet * keySet)
```
markus2330: Please add comments to functions.

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,263 @@
...
+		if (listPointer == NULL)
+		{
+#if PLUGIN_INTERNALNOTIFICATION_VERBOSE
+			fprintf (stderr, "elektraInternalnotificationOpen elektraMalloc failed!\n");
+#endif
+			return -1;
+		}
+		elektraPluginSetData (handle, listPointer);
+
+		// Inizialize list pointers for registered keys
```
markus2330: typo

markus2330:
Good job, looks very nice! Tests should be fine but documentation needs some improvements! It is important that you explain from end user perspective how to use it.

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,263 @@
...
+{
+	KeyRegistration * head;
+	KeyRegistration * tail;
+};
+typedef struct _ListPointer ListPointer;
+
+
+static KeyRegistration * elektraInternalnotificationAddNewRegistration (ListPointer * listPointer)
+{
+	KeyRegistration * item = elektraMalloc (sizeof *item);
```
KurtMi: I think `sizeof (item)` should be enough. With the `*` you allocate more memory then needed.

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,263 @@
...
+	char * name;
+	int * variable;
+	struct _KeyRegistration * next;
+};
+typedef struct _KeyRegistration KeyRegistration;
+
+struct _ListPointer
+{
+	KeyRegistration * head;
+	KeyRegistration * tail;
```
KurtMi: I do not see any advantage of a tail in an single linked list :)

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,263 @@
...
+			next = current->next;
+			elektraFree (current->name);
+			elektraFree (current);
+
+			current = next;
+		}
+
+		// Free list pointer
+		elektraFree (listPointer);
+		elektraPluginSetData (handle, NULL);
```
KurtMi: I think this it not necessary.

src/plugins/internalnotification/internalnotification.c
KurtMi: The parameter `listPointer` is not necessary you could call `elektraPluginGetData` in the function.

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,263 @@
...
+	KeyRegistration * item = elektraMalloc (sizeof *item);
+	if (item == NULL)
+	{
+		return NULL;
+	}
+	item->next = NULL;
+
+	if (listPointer->head == NULL)
+	{
+		// Inizialize list
```
sanssecours: > // **Inizialize** list

Typo: **Initialize** list

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,37 @@
...
+- infos/recommends =
+- infos/placements = postgetstorage postcommit
+- infos/status = global libc unittest concept unfinished experimental nodoc
+- infos/metadata =
+- infos/description = Plugin for internal notification
+
+## Usage ##
+
+Allows programs to register integer variables that get automatically updated
+when the value of the key changes.
```
sanssecours: > when the value of **the** key changes.

Maybe better: when the value of **a** key changes.

Namoshek: That would imply that the integer gets also updated when any other key (and not the one for the integer) in the database changes. I haven't read through everything yet, but I doubt this is the case.

src/plugins/internalnotification/internalnotification.c
Namoshek: Is this even legal to reference the variable on the right side during the declaration? I wonder how this works...

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,263 @@
...
+		ksAppend (returned, contract);
+		ksDel (contract);
+
+		return 1;
+	}
+#if PLUGIN_INTERNALNOTIFICATION_VERBOSE
+	fprintf (stderr, "elektraInternalnotificationGet keyName=%s\n", parentKeyName);
+#endif
+
+	ListPointer * listPointer = elektraPluginGetData (handle);
```
Namoshek: I think you should check that the fetched data is not NULL and then cast properly (occurs several times).

src/plugins/internalnotification/internalnotification.c
tom-wa: free only frees the allocated memory, but the pointer still holds the address. i haven't read the whole code, but it's cleaner and allows you to make checks like elektraPluginGetData == NULL

src/plugins/internalnotification/internalnotification.c
tom-wa: it's legal, but better stop talking about it, otherwise you could start a war ;)

src/plugins/internalnotification/README.md
sanssecours: > That would imply that the integer gets also updated when any other key (and not the one for the integer) in the database changes.

You are right. Some might read it that way. Then maybe something like this instead: “…when the value of **a specified** key changes.”

src/plugins/internalnotification/internalnotification.c
sanssecours: A better name for `tail` might be `last`. The tail of a list is usually everything except for the first element:

listmonster

Image Source: http://learnyouahaskell.com/starting-out#an-intro-to-lists

src/plugins/internalnotification/internalnotification.c
waht: I added assertions to make sure that listPointer (now pluginState) is not NULL i.e. that the plugin state is initialized.
Normally Elekta's core will call `elektraInternalnotificationOpen` on `kdbOpen` before any other plugin functions. An application would have to call the functions directly in which case the resulting behaviour is unspecified anyway.

waht:
@markus2330 it should be ready to merge. please review

markus2330:
@haraldg Do you have some time for a code review?

haraldg:
I have little time until the cold season starts, but I'll at least try to understand the API/usage:

So at startup an application registers its int values with elektraInternalnotificationRegisterInt and later whenever kdbGet() is called from the application, the plugin updates these int values? - Note that I needed about an hour to come up with this theory; maybe the documentation not only of the plugin but also of electra itself can be improved, to give an overview of the API more easily.

In case the above theory is correct: Is kdbGet() a cheap operation, if no key has been changed? Is there some way to figure out efficiently when to call kdbGet()?

I don't quite understand why this manual attempt to dynamic linking is necessary, but since it seems to be just a temporary solution, I didn't bother too much to figure it out ... ;-)

markus2330:
@haraldg Thanks for looking into it! Feel free to also add comments to the code if you have some time.

> So at startup an application registers its int values with elektraInternalnotificationRegisterInt and later whenever kdbGet() is called from the application, the plugin updates these int values?

Yes, exactly.

> Note that I needed about an hour to come up with this theory

@waht Can you add a sentence like this in the intro, please? Giving a big picture is always important ;)

> In case the above theory is correct: Is kdbGet() a cheap operation

Yes, kdbGet() is a relatively cheap operation if nothing is to be done (then it only does `stat` on config files to check for changes). We already had kdbGet() polling in some threads/processes on an "embedded" system (I think it was an ARM Cortex-A5) without noticable overhead.

> Is there some way to figure out efficiently when to call kdbGet()?

Yes, via inter-process notification @waht will also work on that (also with event/main loop integration). We are currently thinking about dbus and redis. dbus is in openwrt (and your elektra plugin for it, too). Are you okay with dbus? Do you have some specific requirements?

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,92 @@
...
+- infos/recommends =
+- infos/placements = postgetstorage postcommit
+- infos/status = global libc unittest concept unfinished experimental nodoc
+- infos/metadata =
+- infos/description = Plugin for internal notification
+
+## Usage ##
+
+Allows applications to automatically update registered variables when the value of a specified key has changed.
+Support for callbacks that fire when a specified key has changed is planned.
```
markus2330: Move this to TODOs/Limitations. Please give some big picture in the beginning and then describe things that work.

src/plugins/internalnotification/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+Until the wrapper (see [TODOs](#todos)) is finished it may be used as global plugin:
+`$ sudo kdb global-mount internalnotification`
+
+## Exported Methods ##
+
+This plugin exports the function `int elektraInternalnotificationRegisterInt (Plugin * handle, int * variable, Key * key)`.
+If the given key is contained in a KeySet on a kdbGet or kdbSet operation, its value is
+converted from string to integer and the registered variable is updated with the current value.
+This also works if the registered key is a cascading key.
```
markus2330: What happens with cascading keys?

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,312 @@
...
+int elektraInternalnotificationOpen (Plugin * handle, Key * parentKey ELEKTRA_UNUSED)
+{
+	PluginState * pluginState = elektraPluginGetData (handle);
+	if (pluginState == NULL)
+	{
+		pluginState = elektraMalloc (sizeof *pluginState);
+		if (pluginState == NULL)
+		{
+#if PLUGIN_INTERNALNOTIFICATION_VERBOSE
+			fprintf (stderr, "elektraInternalnotificationOpen elektraMalloc failed!\n");
```
markus2330: please use logging also here and you should set an error for outside world. (Which does the logging already for you)

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,312 @@
...
+			if (*end == 0 && errno == 0 && value <= INT_MAX && value >= INT_MIN)
+			{
+				*(registeredKey->variable) = value;
+			}
+			else
+			{
+				ELEKTRA_LOG ("conversion failed! keyString=\"%s\" *end=%c, errno=%d, value=%ld", keyString (key), *end,
+					     errno, value);
+			}
+		}
```
markus2330: log if not found?

src/plugins/internalnotification/internalnotification.c
```diff
@@ -0,0 +1,312 @@
...
+			errno = 0;
+			long int value = strtol (keyString (key), &end, 10);
+			// Update variable if conversion was successful and did not exceed integer range
+			if (*end == 0 && errno == 0 && value <= INT_MAX && value >= INT_MIN)
+			{
+				*(registeredKey->variable) = value;
+			}
+			else
+			{
+				ELEKTRA_LOG ("conversion failed! keyString=\"%s\" *end=%c, errno=%d, value=%ld", keyString (key), *end,
```
markus2330: Some higher log level here (or even better, add an warning), the user did something wrong here.

haraldg:
> Yes, kdbGet() is a relatively cheap operation if nothing is to be done (then it only does stat on config files to check for changes). We already had kdbGet() polling in some threads/processes on an "embedded" system (I think it was an ARM Cortex-A5) without noticable overhead.

My systems currently don't run on battery but I like too keep the option. So I try keep the code efficient even if it isn't strictly necessary ATM. A stat every few seconds is fine of course.

>>    Is there some way to figure out efficiently when to call kdbGet()?

>Yes, via inter-process notification @waht will also work on that (also with event/main loop integration). We are currently thinking about dbus and redis. dbus is in openwrt (and your elektra plugin for it, too). Are you okay with dbus? Do you have some specific requirements?

I'd rather do some slow polling (if it only triggers one stat in the usual case) than depend on dbus. If electra is going to provide some nice way to register the files it is monitoring with an event loop (I'm using libev), this would be even better of course.

markus2330:
> If electra is going to provide some nice way to register the files it is monitoring with an event loop (I'm using libev), this would be even better of course.

Getting all file names and doing inotify on them is possible but that would be rather a workaround because you might get many events even though it is only one `kdbSet`. So better is to fire events within `kdbSet`.

Because any process may trigger changes (using `kdbSet`) and many applications might be interested, we need some kind of a bus system which is not a good idea to reinvent/reimplement. Is there any dep (providing a message bus) that would be okay for you? Usually the message busses return a file descriptor to be integrated in event loops (like libev). 0mq (zmq) would be another candidate next to dbus.

haraldg:
> Getting all file names and doing inotify on them is possible but that would be rather a workaround because you might get many events even though it is only one kdbSet.

If each event just triggers one kdbGet() and only the first kdbGet() is actually expensive, then this doesn't sound too bad. OTOH I'm not familiar with the efficiency of the inotify system, maybe it isn't that much better then polling (also given that it has lot's of limitations anyway). It definitely isn't worth your time to implement this, if I would be the only user.

> Is there any dep (providing a message bus) that would be okay for you?

I don't think so. I'm very reluctant to add new dependencies and ATM I don't see enough benefits for doing so.

markus2330:
Ok, thanks for the input! So we will also look into inotify-based and polling mechanism (without deps). They also have the advantage that they take modifications bypassing Elektra into account. I will discuss it with @waht and we will inform you when we find a generic solution (otherwise we will advise you how it can be done for your specific case).

markus2330:
Please rebase, there is a conflict in a file.

If its ready for review, please tag it accordingly.

waht:
I added the notification library (wrapper for internalnotification plugin) to this PR. Other notable changes:

- The list plugin allows to pass a plugin handle in its configuration. If a handle is present it uses the plugin instead of loading it. This enables the notification library to load the plugin itself and have a reference to it for getting the functions exported by the plugin.
- I added two fields to the `KDB` struct: `ioBinding` and `notificationPlugin`.

Since the notification library depends on the internalnotification plugin it is excluded from building it the plugin is not included.

markus2330:
Thank you for continuing the work!

> The list plugin has a new configuration key user/plugins/#/handle for passing plugin handles

Would be great if this can be done automatically. Why not use a cascading lookup to allow any namespace? Do we need to make sure that different plugins get different handles? (Some user docu would be great to better understand the purpose.)

waht:
The notification library uses the new configuration option to insert the internalnotification plugin automatically. It changes the configuration only at run-time, not permanently.

I confused the naming with the scheme used for global plugins. The option is only called `plugins/#/handle` without "user". I added the option to the readme of the list plugin. Usage of the library is  covered in the notification tutorial. But I could add a readme for the library documenting mounting.

src/include/kdbnotification.h
```diff
@@ -1,7 +1,37 @@
...
  *
- * @brief Elektra-Notification structures and declarations
+ * @brief Elektra-Notification structures and declarations for
+ * application developers
+ * 
+ * @ingroup kdbnotification
+ *
+ * @defgroup kdbnotification Notification
+ *
+ * @par Global Mounting
```
markus2330: Thanks for writing the docu.

For uniform style we should put the definition of the group in a comment block after the inclusion of header files, see e.g. in src/libs/elektra/kdb.c.

If there are reasons for your style (e.g. can the `@file` be within a group? Is this useful?) please say so, we might change the other files then.

waht: Sure, I changed that.

Yes, a file can be in a group with the `@ingroup` command. Files are then listed in a Doxygen module page before functions - I used this for the I/O and Notification pages.

waht:
I changed the docs and added notifications to the release news. 

@markus2330 I think the PR is ready for review :)

src/libs/notification/example/example_notification.c
```diff
@@ -0,0 +1,105 @@
...
+
+	int result = elektraNotificationOpen (kdb);
+	if (!result)
+	{
+		printf ("could init notification. aborting\n");
+		return -1;
+	}
+
+	int value = 0;
+	Key * valueKey = keyNew ("/sw/tests/example_notification/#0/current/value", KEY_END);
```
markus2330: valueKey is an irritating name. What about keyToWatch?

src/libs/notification/example/example_notification.c
```diff
@@ -0,0 +1,105 @@
...
+}
+
+int main (void)
+{
+	KeySet * config = ksNew (20, KS_END);
+
+	Key * key = keyNew ("/sw/tests/example_notification/#0/current", KEY_END);
+	KDB * kdb = kdbOpen (key);
+	ELEKTRA_NOT_NULL (kdb);
+	if (kdb == NULL)
```
markus2330: Remove assert because it is checked afterwards anyway. Also is one header file less.

src/libs/notification/example/example_notification.c
```diff
@@ -0,0 +1,105 @@
...
+		printf ("could not register callback. aborting!");
+		return -1;
+	}
+
+	kdbGet (kdb, config, key);
+	debugKeySet (config);
+
+	printf ("Heyo! My value is %d\n", value);
+	resetTerminalColor ();
+}
```
markus2330: I doubt that valgrind does not find any memleak here.

src/libs/notification/example/example_notification.c
```diff
@@ -0,0 +1,105 @@
...
+	printf (ANSI_COLOR_RESET "\n");
+}
+
+void debugKeySet (KeySet * ks)
+{
+	Key * cur;
+	ksRewind (ks);
+	while ((cur = ksNext (ks)) != NULL)
+	{
+		printf ("debugKeySet %s = %s\n", keyName (cur), keyString (cur));
```
markus2330: Please only print relevant two keys?

src/libs/notification/example/example_notification.c
```diff
@@ -0,0 +1,105 @@
...
+#include <kdbassert.h>
+#include <kdbhelper.h>
+#include <kdbnotification.h>
+
+#include <signal.h>
+#include <stdio.h>
+#include <string.h>
+
+// from https://en.wikipedia.org/wiki/ANSI_escape_code#Colors
+#define ANSI_COLOR_RESET "\x1b[0m"
```
markus2330: Are they used in this example?

src/libs/notification/example/example_notification.c
```diff
@@ -0,0 +1,105 @@
...
+	if (!result)
+	{
+		printf ("could not register callback. aborting!");
+		return -1;
+	}
+
+	kdbGet (kdb, config, key);
+	debugKeySet (config);
+
+	printf ("Heyo! My value is %d\n", value);
```
markus2330: The example works great but without loop this is not really visible!

src/libs/notification/example/example_notification.c
waht: Well true, I did not check the example for memory leaks, only the tests :)

src/libs/notification/example/example_notification.c
waht: Yes the are used to set the colors in `setTerminalColor()`.

src/libs/notification/example/example_notification.c
waht: Great! I will add a loop.

src/libs/notification/example/example_notification.c
markus2330: Okay, it was not clear to me how to use this then.

waht:
Great! I extended the example and added change detection to the internalnotification plugin. For string keys conversion or callbacks are now only triggered if the key's value has changed. Binary keys are triggered at every kdbGet.

markus2330:
Looks good now, is this ready to merge?

markus2330:
jenkins build all please

waht:
jenkins build all please

waht:
Yes I think so.

About the failed build checks: 
- "elektra-mergerequests-fast": seems to had an unclean cmake env (there is no typechecker plugin in my branch) 
- "elektra-multiconfig-gcc-stable": the test "testshell_markdown_blockresolver" failed - should be unrelated
- "elektra-multiconfig-gcc47-cmake-options": the test "testshell_hosts" failed - should also be unrelated 
- @sanssecours is already on the build issue on Travis/OSX (#1817)

markus2330:
jenkins build all please

markus2330:
I triggered all build jobs, let us hope they succeed now.

markus2330:
I am afraid you need to rebase so that travis succeeds again. See #1817

waht:
jenkins build all please

markus2330:
Thank you for creating this plugin!

PR #830, opened 2016-07-10
Draft: intercept open

tom-wa:
not really sure how ~/.vimrc vs /home/user/.vimrc should be handled if both are specified, any suggestions ?


markus2330:
Please write documentation (README.md) about how it is handled currently.

Btw. I think it would be better if we have the open-interceptor in bindings (we should also relocate getenv). It has the advantage that users can select if should be built or not (via BINDINGS).

Where do you check if config keys are below PRELOAD_PATH?

Does this implementation works for every application? What is the overhead? It seems to be quite expensive to do dlsym for every call?

Do you have further ideas about dynamical generation of whole configuration files?

tom-wa:
>It seems to be quite expensive to do dlsym for every call?
i've tried doing the dlsym calls in the module init phase, but that leads to segfaults in vim, emacs, and probably some other programs (nano works fine though)
doing dlsym in the open calls seems to fix the problem

markus2330:
Can you move it to `src/bindings` and create a CMakeLists.txt?

markus2330:
Please ping if ready. (Please see e-mail to arbeiten before)

tom-wa:
@markus2330 done.

markus2330:
Thank you!

Issue #812, opened 2016-06-24
augeas: update 1.4.0 to 1.5.0 breaks syntax


mpranj:
On OS X (homebrew) the augeas package was updated recently. Apparently there were some syntax changes.
Take this also as a hint that travis builds will fail form now on.

```
 30/107 Test  #30: testmod_augeas .....................***Failed    0.01 sec
AUGEAS       TESTS
==================

/Users/mpranj/workspace/libelektra/src/plugins/augeas/testmod_augeas.c:34: error in test_hostLensRead: call to kdbGet was not successful
number: 85
description: : an Augeas error occurred
ingroup: : plugin
module: : storage
at: /Users/mpranj/workspace/libelektra/src/plugins/augeas/augeas.c:414
reason: : Syntax error in lens definition
mountpoint: : user/tests/augeas-hosts
configfile: : /Users/mpranj/workspace/libelektra/src/plugins/augeas/augeas/hosts-read
/Users/mpranj/workspace/libelektra/src/plugins/augeas/testmod_augeas.c:35: error in test_hostLensRead: error in kdbGet
/Users/mpranj/workspace/libelektra/src/plugins/augeas/testmod_augeas.c:39: fatal in test_hostLensRead: ip address of localhost not found
```

@fberlakovich can you take a look if there is an easy fix?

fberlakovich:
> @fberlakovich can you take a look if there is an easy fix?

I am busy until Tuesday, but I hope to find some time to look at it next week.

markus2330:
What puzzles me is that we only use lenses (in the testcase Hosts.lns) that are distributed with augeas. Can you check if the error also happens if augeas is used directly? Maybe providing augeas version + the lens would help, too.

markus2330:
It would be great if you can look into it: it fails every build which makes it cumbersome to check what the status of a PR is.

fberlakovich:
The last time I tried I couldn't reproduce it, but I will have another look


________________________________
Von: markus2330 <notifications@github.com>
Gesendet: 11.07.2016 18:59
An: ElektraInitiative/libelektra
Cc: Felix Berlakovich; Mention
Betreff: Re: [ElektraInitiative/libelektra] augeas: update 1.4.0 to 1.5.0 breaks syntax (#812)


It would be great if you can look into it: it fails every build which makes it cumbersome to check what the status of a PR is.

markus2330:
Thank you! Maybe you can improve the test cases so that the error message is more useful?

[...]

PR #835, opened 2016-07-12
CI: workaround for augeas failing on OS X with homebrew

sirblackheart:
Tests currently fail with version 1.5 of augeas from homebrew.
Using homebrews --HEAD version works as a temporary fix so PRs do not
show up broken.

ref #812


markus2330:
Thank you, nicely done! Hopefully will make the build server happy again!

PR #838, opened 2016-07-13
zsh completion WIP

reox:
@markus2330 some things are working, but there are also others missing...
I'm not sure if the install is correct... should i have left out the /usr?


markus2330:
Thank you! Some docu what is working would be good ;)

Further improvements are very welcomed!

Issue #832, opened 2016-07-11
make zsh and bash completion installation more consistent


markus2330:
Is it possible that pkg-config respects `CMAKE_INSTALL_PREFIX`? Then we could avoid `BASH_COMPLETION_COMPLETIONSDIR` hack.

sirblackheart:
The current hack is using the `bash-completion-config.cmake` for systems  with bash-completion `>= 2.2`. For bash-completion `< 2.2` (debian still ships `2.1`) it falls back to the pkg-config file as the cmake file is not available.

It could be possible to honor `CMAKE_INSTALL_PREFIX` with some work but then we do not honer the install paths provided by cmake or pkg-config.  Those paths are generated on making the corresponding package/install based on the install "PREFIX" set for them. So if for example bash-completion got installed into `/usr/local/` `BASH_COMPLETION_COMPLETIONSDIR` would point to `/usr/local/share/bash-completion/completions`.

I think there will be no easy way to honor both Elektras `CMAKE_INSTALL_PREFIX` and the `PREFIX` that was used on the install of the other package e.g. bash-completion.

Honoring the cmake and/or pkg-config paths means we will get the appropriate paths generated on package generation that get sourced or loaded automatically. e.g. homebrews pkg-config file of bash-completion will expose the correct folder for it to be sourced.

markus2330:
With hack I mainly mean setting `BASH_COMPLETION_COMPLETIONSDIR` from outside, we already have similar occurrences such as `GOOGLETEST_ROOT` and of course `OPENSSL_MANUAL_OSX_HACK`. It would be better if we can avoid such a variable which can be influenced from outside.

Yes, and there is no guarantee that a modified path from pkg-config will work. So you are right, its better to take it as is. Maybe we can honor `CMAKE_INSTALL_PREFIX` for the fallback path, i.e. use `CMAKE_INSTALL_PREFIX + share/bash-completion/completions` as fallback.

What I do not like about this fallback is that in one case it would be a `INSTALL_SYSTEM_FILE` (the path from pkg-config), but not in the fallback case. Such logic is not good for a build system.

markus2330:
In wheezy the pkg-config variable bash-completion completionsdir could not be found. Because we use the same debian installation rules, the different fallback caused wheezy packaging to fail: http://community.markus-raab.org:8080/job/elektra-git-buildpackage-wheezy/lastFailedBuild/consoleFull

[...]

PR #841, opened 2016-07-15
scripts: hide bash-completion vars in cmake

sirblackheart:
Unset the the autogenerated cmake variable `bash-completion_DIR` so it
does not clutter the cmake cache variable list and does not make the
build system more complicated.
Also mark `BASH_COMPLETION_COMPLETIONSDIR` as INTERNAL to make it clear
that is not a user set variable.

ref #832


markus2330:
Thank you!

PR #845, opened 2016-07-20
List: refactoring, addPlugin and placement fix

tom-wa:


markus2330:
Please add some information what the placement fix is (especially the commit should have more information!) Ideally it should be split up in three commits.

src/plugins/list/list.c
```diff
@@ -47,11 +48,85 @@ typedef struct
 	KeySet * modules;
 } Placements;
 
+
+unsigned short * expects;
+
+static void listParseConfiguration (Placements * placements, KeySet * config)
```
markus2330: please comment what the function does

tom-wa:
i was about to write that it isn't finished yet, and it's just a proposal / idea on how the exporting interface could work.
i was thinking about using int function(Plugin *handle, PluginData *data) and exporting an ORed list of enums that specifies what the plugin expects.
```
  typedef struct
  {
       Key *parentKey;
       KeySet *returned;
       Key *dataKey;
       KeySet *dataKeySet;
       void *data;
  }PluginData;
 unsigned short expects = EXPECTS_PARENTKEY | EXPECTS_RETURNED | EXPECTS_DATA_KEY | EXPECTS_DATA_KEYSET | EXPECTS_DATA_STRUCT;
```
the caller can check what the plugin expects, fill the plugins `PluginData` struct, e.g.
```
if(expects & EXPECTS_PARENT_KEY)
       data->parentKey = pk;
```
that way it kinda would be possible the figure out the signature of an exported function. 


At least it seemed like a good idea at first, but i'm not so sure anymore. 
any other suggestions ?

markus2330:
I think its easier to pass all information via key/value, not via structs.

tom-wa:
i was trying to find a way to have the same signature for every function and having some kind of reflection. 
wasn't my best idea though. 

btw, it turned out that, after the refactoring, i can not only add plugins at runtime, but also edit them (remove them completely or from specific placements and add them to other placements)

markus2330:
Yes, some reflection is needed. (You need to know if the plugin exists already). Please write some docu for how to use exported functions and the reflection in general, (e.g. in doc/help/elektra-plugins-framework.md or also in a new document doc/help/elektra-plugins-symbols.md).

And please add the unit test to CMake. @waht: Does it do everything you need?

waht:
Yes, great! Seems like it does everything I need.

markus2330:
Ok, thank you!

Issue #674, opened 2016-04-29
Ninja: Warning About Multiple Rules Generating `kdberrors.h` 


sanssecours:
## Steps To Reproduce

1. Change directory to the root directory of the cloned repository
2. Execute the following commands:

    ```sh
    mkdir -p build
    cd build
    cmake .. -GNinja -DPLUGINS="dump;resolver;yajl;resolver_fm_hpu_b"
    ninja
    ```

## Result

Ninja prints the following warning multiple times:

```
ninja: warning: multiple rules generate src/include/kdberrors.h. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
```

---

This might be related to the CMake macro `add_headers`, since adding plugins also increases the number of warnings `ninja` prints. ~~If I comment the lines~~
```cmake
add_custom_command (
		OUTPUT ${BINARY_INCLUDE_DIR}/kdberrors.h
		DEPENDS elektra-export-errors
		COMMAND ${EXE_ERR_LOC}
		ARGS ${EXE_ERR_ARG} ${CMAKE_SOURCE_DIR}/src/error/specification ${BINARY_INCLUDE_DIR}/kdberrors.h
		)
```
~~in the macro, then `ninja` does not print the warning messages any more.~~

markus2330:
Thank you for reporting. The issue is well-known, see very long discussion in #258.

PR #851, opened 2016-07-23
Only Generate `kdberrors.h` Once

sanssecours:
This commit adresses issue #674:

Before this change multiple targets would generate `kdberrors.h` in parallel. Therefore it was possible that one build process would overwrite `kdberrors.h` – with exactly the same content as before – while another process was reading the file.

By the way: I do no think that the attached code is the ideal solution to the problem. However, since it looks like the commit fixes the problem,  it is – in my opinion – better than nothing 😁.

### DIFF
--- a/cmake/Modules/LibAddMacros.cmake
+++ b/cmake/Modules/LibAddMacros.cmake
@@ -204,12 +204,20 @@ macro (add_headers HDR_FILES)
 
 	find_util(elektra-export-errors EXE_ERR_LOC EXE_ERR_ARG)
 
-	add_custom_command (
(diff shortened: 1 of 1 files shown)


markus2330:
Could have been a nice and simple solution, but it seems to fail on parallel builds with `make`, see http://build.libelektra.org:8080/job/elektra-mergerequests-fast/280/console

sanssecours:
> Could have been a nice and simple solution, but it seems to fail on parallel builds with `make`, see http://build.libelektra.org:8080/job/elektra-mergerequests-fast/280/console

Thanks for the info. I guess I have to continue reading on page 42 of ”Mastering CMake“ then… 😢

PR #853, opened 2016-07-23
CONDITIONALS: updated README

tom-wa:
conditionals also works as global plugin, i've updated the readme with an example.


markus2330:
Thank you! The introduction could be improved a bit ;)

PR #857, opened 2016-07-25
Build: Only Generate `kdberrors.h` Once

sanssecours:
This commit contains the code of my second try to address issue #674:

Before this change multiple targets would generate `kdberrors.h` in parallel. Therefore it was possible that one build process would overwrite `kdberrors.h` – with exactly the same content as before – while another process was reading the file.

By the way: I think that this solution is even worse than the [first one](https://github.com/ElektraInitiative/libelektra/pull/851) I proposed 😁.


markus2330:
I do not think the solution is worse, looks like it could work. Let us see what the build server says.

Btw. this should be enough work on the build system by you for a long time! ;)

markus2330:
Thank you, looks fine!

markus2330:
(Btw. we can still review it if you cannot make a different PR, a PR with changes in code would be more interesting though)

PR #862, opened 2016-07-27
Shell Recorder: Improve Portability of Code

sanssecours:
This change adds support for OS X to Shell Recorder. The most problematic part in the attached code is probably `replace_newline_null`:

```sh
tr '\n' '⏎' | tr '\000' '\n'
```

The other changes – removal of `grep -P` – should not be problematic, unless I made a mistake when translating the regex.


KurtMi:
somehow the travis build job is missing, I am curious if it fixes the mac support :)

tom-wa:
the sed part should work, guess the only reason why i've used the lookbehind regex was to omit the "Warning Number: " and "Error " Strings.

tests/shell/shell_recorder/shell_recorder.sh
```diff
@@ -28,7 +28,8 @@ else
         KDBCOMMAND="kdb"
 fi
 
-
+shopt -s expand_aliases
```
markus2330: Avoid, use function instead.

sanssecours: Thanks for the information. I converted the alias to a function it in commit ab772e8.

markus2330:
Thank you!

PR #866, opened 2016-07-29
Shell Recorder: Bugfix & Documentation Update

sanssecours:
## Bugfix

The shell recorder tests, after commit ab772e8c, worked only because all the test outputs do not contain the `NULL` symbol.

## Documentation Update

I documented the fact that you need to use the return symbol (`⏎`) to match a newline in Shell Recorder tests.

If someone wants to know the reason why I did not choose `NULL` instead of `⏎` : Something like

```sh
echo -e "bla\0blubb" | grep -Eq --text "bla\x00blubb"
```

reports a match with BSD-grep, but not with GNU-grep. It works in GNU-grep too, if we specify the `-P` flag, which is not available in BSD's version of grep.

### DIFF
--- a/tests/shell/shell_recorder/README.md
+++ b/tests/shell/shell_recorder/README.md
@@ -9,7 +9,7 @@ e.g. `Mountpoint: user/test` tells the shell_recorder that you will be working u
 
 ### DiffType: ###
 
-Available: 
+Available:
 
 `Dump`, `Ini` exports the keys below `Mountpoint` using either the dump or ini format and diffs the changes.
 `File` does a diff on the configuration file mounted to `Mountpoint`.
@@ -24,7 +24,7 @@ If File is preset but empty a fresh database temp file will be provided for ever
 
 Posix-extended regex is used to check and validate return values and outputs.
 
-Options: 
+Options:
 
 * STDOUT:
 * STDERR:
@@ -35,7 +35,7 @@ Options:
 
 ## Commands ##
 
-A command starts with `<` followed by the usual kdb commands. 
+A command starts with `<` followed by the usual kdb commands.
 
 
 ## Examples ##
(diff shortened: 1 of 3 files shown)


tests/shell/shell_recorder/README.md
```diff
@@ -24,7 +24,9 @@ If File is preset but empty a fresh database temp file will be provided for ever
 
 Posix-extended regex is used to check and validate return values and outputs.
 
-Options: 
+**Remark:** Shell Recorder uses the `⏎` symbol as line terminator. This means that you need to use the character `⏎`  (instead of `\n`) if you want to match a line ending in a multiline output.
```
markus2330: Can you describe at which positions this line terminator is accepted, how to enter the character, and make an example how to use it?

sanssecours: > Can you describe at which positions this line terminator is accepted…

Basically we just use `⏎` instead of the `\n` symbol in a regex. This means `⏎` matches both a newline and `⏎` symbol. Just like any other normal character (e.g. `a`, `→`, `1` or `😱`) you can use `⏎` everywhere in the regex.

> …how to enter the character, …

This depends on how you insert “special characters” in your current OS. If someone does not know how to do that, one easy way is just to copy the character from the documentation or the test file `db_changes.esr` and then insert it via pasting.

> …and make an example how to use it?

I added an example and more documentation in the latest commits.

sanssecours:
jenkins build gcc-configure-debian-shared please

markus2330:
Thank you! Nice docu!

Do you think its related to #871?

sanssecours:
> Do you think its related to #871?

I might be, but I am not really sure.

sanssecours:
> Do you think its related to #871?

I might be, but I am not really sure.

PR #872, opened 2016-07-31
intercept: generate configuration

tom-wa:
```
/preload/open/~\/.vimrc =~/.vimrc-other
/preload/open/~\/.vimrc/readonly = 1
/preload/open/~\/.vimrc/generate = /other/key
/preload/open/~\/.vimrc/generate/plugin = ini
```

@markus2330 should the generated configuration be written to ~/.vimrc-other or a temporary file ?

### DIFF
--- a/src/bindings/intercept/README.md
+++ b/src/bindings/intercept/README.md
@@ -5,7 +5,7 @@
 ## DESCRIPTION
 
 When this library is loaded (either in LD_PRELOAD or ld.so) calls to `open` and `open64` will be intercepted and redirected.
-Additionaly you can force files to be read-only.
+Additionaly you can force files to be read-only, or even generated from elektra backends.
 
 
 ## CONFIGURATION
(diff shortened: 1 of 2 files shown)


markus2330:
You can also disallow such setups. Do you see a reason someone needs such a combination? If allowed, you should rather write to the specified file.

Btw. you can use O_TMPFILE to completely avoid any file in the filesystem. (if no file given, which will be the common case when someone uses `generate`)

tom-wa:
O_TMPFILE doesn't seem like a good idea,
open(2)
```
              O_TMPFILE requires support by the underlying filesystem; only
              a subset of Linux filesystems provide that support. 
```

markus2330:
Do you have any filesystem in use which does not support it? But you are right, it is not the most portable solution, but somewhat nice if you actually never need a file visible in the filesystem.

Btw. On the buildserver the kernel might be too old, its a 3.2.

markus2330:
Thank you!

PR #873, opened 2016-08-01
first logger draft

markus2330:
draft for logger, without docu (only the decision) and some missing implementation

### DIFF
--- a/cmake/ElektraCache.cmake
+++ b/cmake/ElektraCache.cmake
@@ -289,13 +289,12 @@ else (ENABLE_DEBUG)
 	set (DEBUG "0")
 endif (ENABLE_DEBUG)
 
-option (ENABLE_LOGGER "Allows Elektra to write logs (DO NOT USE, currently writes to stdout).")
+option (ENABLE_LOGGER "Allows Elektra to write logs.")
 if (ENABLE_LOGGER)
-	set (VERBOSE "1")
+	set (HAVE_LOGGER "1")
 else (ENABLE_LOGGER)
-	set (VERBOSE "0")
+	set (HAVE_LOGGER "0")
 endif (ENABLE_LOGGER)
-MARK_AS_ADVANCED(ENABLE_LOGGER)
 
 
 #
--- a/doc/CODING.md
+++ b/doc/CODING.md
@@ -60,29 +60,37 @@ Up-to-date code comments are essential to make code understandable for others.
 1. Ideally you should use assertions to state what should be true at a specific
    position in the code. Their syntax is checked and they are automatically
    verified at run-time. So they are not only useful for people reading the
-   code.
+   code:
+
    ```c
    #include <kdbassert.h>
    ELEKTRA_ASSERT (condition, "text to be printed when assert fails", ...)
    ```
+
 2. If the "comment" might be useful to be printed during execution, use logging.
    First introduce your module `modulename` in `kdblog.h`.
    Then you can use it:
+
    ```c
    #define ELEKTRA_LOG_MODULE modulename
    #include <kdblog.h>
    ELEKTRA_LOG ("text to be printed according log filters", ...)
    ```
+
    Or if multiple modules are intermixed in the same file:
+
    ```c
    #include <kdblog.h>
    ELEKTRA_LOG (modulename, "text to be printed according log filters", ...)
    ```
 
 3. Prefer to comment functions with Doxygen, see below.
+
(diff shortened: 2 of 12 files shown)


petermax2:
I like the logging interface! Should be easy to adapt to. Also I agree on using assertions more often.

markus2330:
Thanks for looking into it!

Two two more concrete questions (directed to everyone):

1. Do you want the feature that a module can be defined per file, so that `ELEKTRA_LOG` does not need the module argument? (The old-style version still exists if `ELEKTRA_LOG_MODULE` is not defined)
2. Do you want a printf like functionality for `ELEKTRA_ASSERT`, too? If the logger is enabled, assert would use the logger (with high severity), otherwise it would directly printf to stderr.

petermax2:
> Do you want the feature that a module can be defined per file, so that ELEKTRA_LOG does not need the module argument? (The old-style version still exists if ELEKTRA_LOG_MODULE is not defined)

I think `__FILE__` and `__LINE__` are more helpful than an "abstract" module name.

> Do you want a printf like functionality for ELEKTRA_ASSERT, too? If the logger is enabled, assert would use the logger (with high severity), otherwise it would directly printf to stderr.

Would the logger on high severity do something else than printf to stderr? It may be useful then. Also it would be more consistent if we used the logger.

markus2330:
> I think __FILE__ and __LINE__ are more helpful than an "abstract" module name.

Yes,  __FILE__ and __LINE__ are very useful for output (and supported). The module name is only for filtering (usually you only want to see your own stuff, not everything). Let me reformulate my question: Do you think a abbreviation (with the first parameter predefined) for `ELEKTRA_LOG` in the case that in a source file only log statements for a single module are present (which should be the common case?) outweighs the disadvantage (the ugly hack that `ELEKTRA_LOG_MODULE` needs to be defined before any header file is included). Or should modules define their own macros when they want to make log statements shorter? (there are 4 macros for the 4 severity levels)

Or if this is still unclear: look at CODING.md line 77, do you need such a feature? (or is what is on line 84 enough)

> Would the logger on high severity do something else than printf to stderr?

Yes, currently the logger can write to files and syslog.

petermax2:
ELEKTRA_LOG (modulename, "text to be printed according log filters", ...)

I think this would be enough.

markus2330:
Ok, I only thought that, with longer module names and deep indentation, it would not look nice, e.g.

```
void foo()
{
        if (x)
        {
                ELEKTRA_LOG_NOTICE (ELEKTRA_LOG_MODULE_INTERNALNOTIFICATION, "..."
```

If you think its no problem to have such multi-line log messages, I would also prefer not to have it. But what would be worse if every module defines macros in an inconsistent way.

waht:
Agree on the logging macros issue. Modules could still `#define` their module name to shorten the lines. e.g.

```
#define MODULE ELEKTRA_LOG_MODULE_INTERNALNOTIFICATION
...
ELEKTRA_LOG_NOTICE (MODULE, "..."
```

This way there is no need for defining the module name before headers.

markus2330:
Yes, the only issue there is that it is a bit dangerous to define macros without prefixes, especially with generic text that really could appear somewhere else. (I think you know the min/max macro problems.) And with all prefixes and so on appended, you end up with a long name.

Basically, we have the same situation with warnings/errors, without any problem so far. I only brought up the topic, because logging should be used much more frequent than adding warnings or setting errors.

sanssecours:
Since we currently have to specify the module (number) anyway: How about an optional way to include the plugin name in (front of) the log message? For example: If we use the following code: 

```c
ELEKTRA_LOG_DEBUG (ELEKTRA_LOG_MODULE_KDB, "Hi");
ELEKTRA_LOG_DEBUG (ELEKTRA_LOG_MODULE_AUGEAS, "Hi");
```

, then Elektra would write out something like this:

```
KDB:    Hi
Augeas: Hi
```

instead of this:

```
Hi
Hi
```
.

petermax2:
@sanssecours you would get a lot of unidentified "Hi" then, because every module will use the logger in some way. Making the identification optional would lead to chaotic log files, I'm afraid.

markus2330:
Thanks for the feedback.

The default format would be `"%s:%d:%s: %s\n", file, line, function, msg`

Then file, line and function already tells quite specifically who says "Hi". Unfortunately, it would *not* be trivial to output the module name, because we would need to maintain a mapping from enum to a string.

But let us discuss it.

PR #874, opened 2016-08-01
global-plugins: mark placements

mpranj:
Introducing only a few of the proposed global plugin placements.


markus2330:
Can you mark the placements you found to be best nevertheless and write why they are not ideal? The foreach and init placements are also missing.

src/libs/elektra/kdb.c
```diff
@@ -447,6 +447,8 @@ int kdbClose (KDB * handle, Key * errorKey)
  */
 static int elektraGetCheckUpdateNeeded (Split * split, Key * parentKey)
 {
+	// GLOBAL: getresolver [init]
```
markus2330: init could be even earlier directly in kdbGet?

src/libs/elektra/kdb.c
```diff
@@ -456,7 +458,7 @@ static int elektraGetCheckUpdateNeeded (Split * split, Key * parentKey)
 
 		if (backend->getplugins[RESOLVER_PLUGIN])
 		{
-			// GLOBAL: getresolver
+			// GLOBAL: getresolver [foreach]
```
markus2330: I think we missed at the meeting that for `[foreach]` a pre/post might be useful. (Post execute after line 465)

src/libs/elektra/kdb.c
```diff
@@ -485,6 +487,7 @@ static int elektraGetCheckUpdateNeeded (Split * split, Key * parentKey)
 			return -1;
 		}
 	}
+	// GLOBAL: getresolver [deinit]
```
markus2330: won't work here, `return -1` would bypass it.

src/libs/elektra/kdb.c
```diff
@@ -562,6 +572,10 @@ static int elektraGetDoUpdateWithGlobalHooks (KDB * handle, Split * split, KeySe
 		for (int p = start; p < end; ++p)
 		{
 			int ret = 0;
+
+			// GLOBAL: getstorage [foreach]
```
markus2330: Would be executed even if no `backend->getplugins[p]` exists.

src/libs/elektra/kdb.c
```diff
@@ -831,12 +851,28 @@ int kdbGet (KDB * handle, KeySet * ks, Key * parentKey)
 static int elektraSetPrepare (Split * split, Key * parentKey, Key ** errorKey, Plugin ** hooks)
 {
 	int any_error = 0;
+
+	// GLOBAL: setresolver [init]
+	// GLOBAL: setresolver [max once]
+
+	// GLOBAL: presetstorage [init]
+	// GLOBAL: presetstorage [max once]
```
markus2330: Seems like we need to split the loop to have presetstorage and setstorage separated.

setresolver would be the COMMIT_PLUGIN, its not within this loop.

src/libs/elektra/kdb.c
```diff
@@ -831,12 +851,28 @@ int kdbGet (KDB * handle, KeySet * ks, Key * parentKey)
...
 	{
 		for (size_t p = 0; p < COMMIT_PLUGIN; ++p)
 		{
 			int ret = 0; // last return value
 
+			// if (p == RESOLVER_PLUGIN)
+			// GLOBAL: setresolver [foreach]
+
+			// if (p == STORAGE_PLUGIN)
+			// GLOBAL: setstorage [foreach]
```
markus2330: Would not be needed if loop is split.

src/libs/elektra/kdb.c
```diff
@@ -1168,14 +1216,22 @@ int kdbSet (KDB * handle, KeySet * ks, Key * parentKey)
 
 error:
 	keySetName (parentKey, keyName (initialParent));
-	// GLOBAL: prerollback
+	// GLOBAL: prerollback [init]
+	// GLOBAL: prerollback [max once]
+	// GLOBAL: prerollback [foreach]
+	// GLOBAL: prerollback [deinit]
```
markus2330: again, a split in the loop within elektraSetRollback would be required (prerollback and rollback). The foreach must be within the loop of elektraSetRollback.

markus2330:
There are some improvements! Maybe its better to directly work on the implement instead of just comments because you need to split some loops and some control flow needs changes because of the deinit.

It would be good if you have a helper function/macro for executing the hooks so that the hooks are better visible/readable.

mpranj:
@markus2330 thanks for the feedback. Yes indeed, I think it's better to switch to concrete code now.

Did you find anything that was completely misplaced in general? I was unsure about what semantics was wanted, but it will be unambiguous with concrete code anyway.

I was already planning on having a function or macro to keep the verbosity/boilerplate code about the same as the current comments (one-liners). It would introduce **a lot** of duplicate code otherwise.

src/libs/elektra/kdb.c
```diff
@@ -1131,16 +1174,28 @@ int kdbSet (KDB * handle, KeySet * ks, Key * parentKey)
...
+	// GLOBAL: precommit [max once]
+	// GLOBAL: precommit [foreach]
+	// GLOBAL: precommit [deinit]
 	if (handle->globalPlugins[PRECOMMIT])
 	{
 		handle->globalPlugins[PRECOMMIT]->kdbSet (handle->globalPlugins[PRECOMMIT], ks, parentKey);
 	}
 
 	elektraSetCommit (split, parentKey);
+	// GLOBAL: commit [init]
```
markus2330: should be before elektraSetCommit

markus2330:
I already commented about some issues, but they are minor and it would also be trivial to fix the issues with code.

Maybe @tom-wa, @Namoshek, @waht  or @KurtMi can find further problems, they already used (or even implemented) hooks.

Another approach would be to pinpoint problems within the test cases, maybe you can write some of them upfront? The desired sequence of called hooks in certain situations should be easy to check.

src/libs/elektra/kdb.c
markus2330: or is even alias for precommit?

src/libs/elektra/kdb.c
```diff
@@ -536,6 +540,13 @@ static int elektraGetDoUpdateWithGlobalHooks (KDB * handle, Split * split, KeySe
 	const int bypassedSplits = 1;
 	int pgs_done = 0;
 	int pgc_done = 0;
+
+	// GLOBAL: getstorage [init]
+	// GLOBAL: getstorage [max once]
+
+	// GLOBAL: postgetstorage [init]
+	// GLOBAL: postgetstorage [max once]
+
```
tom-wa: just for your information: elektraGetDoUpdatewithGlobalHooks gets called twice (thats why there's the `run` parameter)

mpranj: Thanks. I did think of that since we discussed it in a meeting.

All of these are just rough markers where I thought the hooks could be placed.

mpranj:
Then I don't think that it would make sense for anyone to review it yet, so don't waste your time.
It would be appreciated when the code is ready.

mpranj:
This does not yet comply with the proposal for global plugins, but it adds some more positions which would be useful for #1072. (postgetstorage/deinit in particular)

@markus2330 would you consider merging the available part of the implementation already? Then I could continue working in another PR while @Namoshek could start using it already.

Namoshek:
Not much success so far. I used `postgetstorage/init` as well as `postgetstorage/deinit`, but it doesn't seem to work. Could also be that I'm doing something wrong or my plugin does something wrong, but I don't know how to test it (with unit tests) as global plugin.

I had to change the last function in [cmake/Modules/LibAddMacros.cmake](../blob/master/cmake/Modules/LibAddMacros.cmake#L589) by the way, it did not accept slashes for `infos/placements`. You could add this to your PR (replace `[a-zA-Z0-9 ]*)` by `[a-zA-Z0-9/ ]*`).

mpranj:
@Namoshek, you have to either:
- use kdb global-mount to mount it
- or configure it manually like it's done in `tests/shell/shell_recorder/globaltest.esr` by setting the apropriate keys. Note that after setting the keys you need to `kdbClose` and then `kdbOpen` again. Otherwise the mount procedure is not done.

Yes, `infos/placements` is very misleading w.r.t. global plugins and (currently) has nothing to do with it.

Namoshek:
> Yes, infos/placements is very misleading w.r.t. global plugins and (currently) has nothing to do with it.

Ok, that info is (obviously) very important, thanks for pointing it out. Now I also understand why you want to inform the plugin in what stage it is executed (would definitely make sense to limit the cachefilter plugin here to one state, as it could potentially increase execution time enormously, although it shouldn't matter if it is executed 20 times (from a functional perspective)).

> use kdb global-mount to mount it

That is what I did and I also confirmed it by listing the elektra keys with `kdb ls system/elektra` and getting the global plugin values (one is `spec`, the other `cachefilter`). So it should definitely be mounted. And of course I also restarted the service. I'll try once more and add some logging output to the plugin, maybe this helps.

Thank you so far!

mpranj:
If it doesn't seem to work don't waste too much time. I'll look into it on saturday. Can't do any sooner. It's possible that the placement is not executed.

mpranj:
jenkins build multiconfig-gcc47-cmake-options please

markus2330:
@mpranj Can you make this ready to merge? We now have two month to get rid of all issues.

mpranj:
@markus2330 I'm pretty busy but it's already on my asap list. I'd expect it to be ready within a week.

markus2330:
@mpranj Thank you, the code is clean and nice!

I merged it into the branch cachefilter so that we can move forward in getting the cachefilter done. @tom-wa's spec plugin won't work together with it, the MAXONCE placements were  weird.

Let us fix cachefilter's hooks first, they are straight-forward presetstorage and postgetstorage. Let us misuse MAXONCE so that list-plugin and `kdb gmount` do not need changes for now (INIT or DEINIT might fit better: we also want it to be called when `kdbSet` or `kdbGet` has nothing to do).

Unfortunately, it seems like there are some memproblems and `elektraGlobalSet (handle, ks, parentKey, PRESETSTORAGE, MAXONCE);` does not always call the hook (on multiple `kdbSet`)?

> Yes, infos/placements is very misleading w.r.t. global plugins and (currently) has nothing to do with it.

What do you mean here? The `kdb global-mount` tool reads and uses the infos/placements, so they are pretty much related? Or do you mean that MAXONCE is currently the only one that is supported?

tom-wa:
i think it's actually the list plugin that isn't working with the global placements because they are hardcoded. i wanted to wait until all the placements are done and implemented before i update it.

markus2330:
@tom-wa any Thanks for the hint! Is there any reason to not add all needed names now? doc/decisions/global_plugins.md documents the names for quite some time.

mpranj:
@markus2330 what's weird about the MAXONCE placements? Do you mean I put them in the wrong place regarding the desired semantics of the hooks?

> What do you mean here? The kdb global-mount tool reads and uses the infos/placements, so they are pretty much related? Or do you mean that MAXONCE is currently the only one that is supported?

It's possible that I'm wrong. When I tested it, it didn't mount to the specified infos/placements.

markus2330:
> @markus2330 what's weird about the MAXONCE placements? Do you mean I put them in the wrong place regarding the desired semantics of the hooks?

They are called in the middle in the loop. I would expect FOREACH there (with keyset per mountpoint) and MAXONCE should be before or after the loop (with full keyset). @tom-wa What is the reason that you call the hooks within the loop, but only once? Do we need a position that still has this semantic?

> It's possible that I'm wrong. When I tested it, it didn't mount to the specified infos/placements.

Yes, the problem seem to be the hardcoded names in the list plugin. Hopefully @tom-wa updates this soon.

mpranj:
@markus2330 I guess further work should be done on top of the cachefilter branch and this PR here can be closed as soon as I don't need any information from here?

markus2330:
@mpranj Yes, we definitely should close this PR because it is not concerned with its initial goal at all (marking the placements). It would be great if you can fix the remaining issues on the branch (memleak, sporadically not executed presetstorage MAXONCE) and add test for the cachefilter placements.

mpranj:
Thanks! I will definitely fix them.

src/plugins/cachefilter/cachefilter.c
```diff
@@ -42,7 +42,7 @@ int elektraCachefilterGet (Plugin * handle, KeySet * returned, Key * parentKey)
 	void * cache = elektraPluginGetData (handle);
 	if (cache == NULL)
 	{
-		toBeCached = ksNew (ksGetSize (returned), KS_END);
+		toBeCached = ksNew (0, KS_END);
```
markus2330: Did you benchmark this? It might waste some space but should avoid many reallocations (for larger keysets).

mpranj: No I didn't but this change is also not crucial.

The code of `ksAppend`, which is used later here, does suggest that it uses only one resize in advance.

I wanted to let the core decide about the allocation but if you're saying this is bad I don't mind undoing this. The objective was to get rid of the memleaks and move forward.

markus2330: > No I didn't but this change is also not crucial.

Thus I wondered why you changed it. Comments help in such situations.

But you are right, lets move forward for now. Before you start changing `ksNew` (to be suitable for mmap), it would be great to have benchmarks for such reallocations.

mpranj: Yep you're right. I undid the change because it was unfounded.

mpranj:
@markus2330 I'll close this PR today (I just kept it for the buildbots and some remaining infos).

Do you want me to do a PR on cachefilter or do you want to merge cachefilter to master if the memleaks are fixed now?

markus2330:
>  I'll close this PR today (I just kept it for the buildbots and some remaining infos).

Perfect, it is overdue!

> Do you want me to do a PR on cachefilter or do you want to merge cachefilter to master if the memleaks are fixed now?

I am okay with both: You can create a single PR with all changes (and close #1078); or update #1078 and create a separate PRs for different modules. In any case, please keep changes of separate modules in separate commits.

Issue #765, opened 2016-06-01
fix usage of ksCut in kdb*


markus2330:
update ksCut usage with cascading keys?

tom-wa:
ok, this one is a bit tricky.
the problem are the default keys (created by the spec plugin) which are cascading keys. when i try to remove all `spec`-keys from the keyset in `elektraGetDoUpdateWithGlobalHooks` the cascading keys are cut together with the `spec` keys because `/examples/highlevel/gps/status` is below `spec/examples/highlevel` 

i'm not even sure what exactly the expected behaviour would be. it seems like before my change to keyIsBelow default keys were simply ignored because of
```
case KEY_NS_CASCADING:
        validNS = 0;
```

so i'd say the way it currently works is actually correct and i'll either handle the default keys in `elektraGetDoUpdateWithGlobalHooks` or just tell the validation plugins to ignore the `spec` namespace
using keyRel2 in ksCut would work too, but it's too early for that i guess

@markus2330 what would you say should be done here ?

markus2330:
I think you are trying to solve an impossible problem. In the long run we need a `ksCut2` with either a filter function or an enum where you can say how the filtering should be done. Is it possible to reuse the same enum of `keyRel2` for `ksCut2`?

tom-wa:
>I think you are trying to solve an impossible problem.

i came to the same conclusion, i've tried several approaches, the best would be a workaround in ``elektraGetDoUpdateWithGlobalHooks`(i already have a working one) until we have `ksCut2` ready. 

>Is it possible to reuse the same enum of keyRel2 for ksCut2?

actually.. `ELEKTRA_REL_BELOW_SAME_NS` is exactly what you want here. with `keyRel2` `ksCut2` could be implemented in a few lines doing exactly what needs to be done. The only thing we could argue about is if we should introduce some extra enums von convenience regarding the default/cascading keys.


imho we should also review all the unittests to see if they cover everything or if some newer features (like cascading/default keys) are missing

markus2330:
> workaround in  `elektraGetDoUpdateWithGlobalHooks`

okay, hopefully it does not interfer with @mpranj work. He should have a look at the PR in any case.

> reuse of same enum

That would be really nice if both functions can use it. A filter is very powerful, but also  more complicated to use.

> some extra enums von convenience regarding the default/cascading keys

[...]

PR #882, opened 2016-08-15
Fix global hook keyset / ksCut

tom-wa:
Workaround for removing spec keys from the keyset passed to global plugins while keeping cascading/default keys. spec keys have to be removed because otherwise validation plugins might try to validate them and will most likely fail. 

currently there is no way to exclude cascading/default keys when using ksCut which causes problems with default keys created by plugins like `conditionals`. this workaround moves the removed keys back to the original keyset before passing it to the global plugins.

Edited!


markus2330:
Thank you!

@tom-wa please document what exactly you fix with this PR.

(@mpranj seems like this is the PR we discussed in #765)

mpranj:
Thanks. Doesn't look like this will be a merge conflict for me at all.

tom-wa:
@markus2330 added description to the PR

markus2330:
@mpranj Should it be merged? How is your PR progressing?

mpranj:
@markus2330 Concerning my work: yes it can be merged. I did not check the logic though, I assume tom-wa knows what he's doing ;)

Issue #475, opened 2016-02-07
Add GPG-Interface


petermax2:
# Problem Setting

The crypto plugin is quite useless at the moment as there is no way to use it with the kdb command. 

The introduction of a GPG interface enables the user to utilize her existing key-pairs for cryptographic operations in Elektra. The private key is used for encrypting a random sequence, which serves as seed for a key derivation function (KDF). This way we can safely derivate cryptographic keys for symmetric value encryption. Both OpenSSL and libgcrypt have built-in support for the PBKDF2 (see [RFC 2898](https://tools.ietf.org/html/rfc2898)). [1, 2]

The PBKDF2 needs an **iteration number** and a **salt** in order to work. Those values will be stored per Key as MetaKey.

# Solution

See discussion below to find how we came to the results.

## Procedure

During the **mount phase** a random master password _r_ is being generated. _r_ is sent to the gpg binary for encryption. The resulting encrypted master password _m_ is stored in the plugin configuration at `config/masterChallenge`.

During the **set phase** the master password _m_ is sent to the gpg binary for decryption in order to retrieve _r_. The following steps will be repeated for every Key _k_, that is supposed to be encrypted. A random salt _s(k)_ is generated. By applying the PBKDF2 (mentioned earlier) with _r_ and _s(k)_, the cryptographic key _e(k)_ and the initialization vector _i(k)_ is being derived. The value of _k_ will be encrypted using _e(k)_ and _i(k)_. The seed _s(k)_ will be encoded as prefix into the encrypted value.

During the **get phase** the master password _m_ is sent to the gpg binary for decryption in order to retrieve _r_. The following steps will be repeated for every Key _k_, that is supposed to be decrypted. The salt _s(k)_ is read from the encrypted message. By applying the PBKDF2 with _r_ and _s(k)_ the values of _e(k)_ and _i(k)_ are restored. Then the encrypted message can be decrypted.

## General notes

We do not call the gpg-agent directly. Neither KMail nor Evolution do that. Instead they both call the gpg binary directly. In `~/.gnupg/gpg.conf` the usage of the gpg-agent can be enabled by setting the `use-agent` option. The gpg binary is the best way of handeling the gpg-agent protocol.

The plugin configuration must contain at least one GPG key descriptor that can be used as **recipient** but also for decrypting messages. See `man gpg` for more information about valid GPG key descriptions.

# Implementation Tasks

- [x] implement module for calling the gpg binary
[...]

PR #889, opened 2016-08-19
GPG Interface

petermax2:
# Purpose

Implement GPG interface as described in #475 .

# TODO

- [x] Fix unit tests
- [x] lookup gpg binary in PATH
- [x] FIX segfault in gcrypt and OpenSSL variant
- [x] resolve merge conflict


doc/METADATA.ini
```diff
@@ -614,3 +614,9 @@ usedby/plugins= crypto
 type= string
 description= used to mark a key for encryption. The crypto plugin is supposed to encrypt the key. After the encryption the type of the key will be set to "binary". Its original type will be restored after decryption.
 
+[crypto/salt]
+status= implemented
+usedby/plugins= crypto
+type= string
+description= used to store the salt thas is required for deriving the cryptographic key as well as the initialization vector (IV).
```
markus2330: Can you describe the life-time of the salt? Is it a security problem if the salt leaks?

I think its best if you add some paragraphs about "security considerations" in the docu somewhere.

src/plugins/crypto/README.md
```diff
@@ -109,34 +110,35 @@ Set the CMake variables `OPENSSL_INCLUDE_DIR` and `OPENSSL_LIBRARIES` to your de
 
 The crypto plugin will encrypt and decrypt values using AES-256 in CBC mode.
 
-The key derivation is still WIP.
+The GPG binary is required for handling the cryptographic keys.
 
-### Planned Features ###
+At the moment the plugins will only run in a UNIX/Linux-like system.
```
markus2330: Maybe a bit more specific. Do you need something beyond POSIX (and the gpg binary)?

src/plugins/crypto/README.md
```diff
@@ -148,23 +150,18 @@ Shutdown is enabled in the unit tests to prevent memory leaks.
...
-
-You specify the parameters of the cryptographic operations in a KeySet together with the keys to be encrypted.
-The following parameters are required:
-
-- **Key** - the symmetric cryptographic key for encryption
-- **IV** - the initialization vector (IV) that is required by the CBC mode
+To mount a backend with the gcrypt plugin variant that uses the GPG key 859E2AD7, use:
 
-The following keys are required for metadata based encryption:
+	kdb mount test.ecf user/t crypto_gcrypt "crypto/key=859E2AD7"
```
markus2330: Shouldn't we discourage the usage of short IDs? (I think you know the story about the faked keys that collide with keys from Linux Kernel Devs. I know it does not really apply here, because the keys need to be in the ring already, but it might be better from "educational" point of view, at least in the examples.)

src/plugins/crypto/README.md
```diff
@@ -109,34 +110,35 @@ Set the CMake variables `OPENSSL_INCLUDE_DIR` and `OPENSSL_LIBRARIES` to your de
...
-## Configuration ##
+### Cryptographic Operations ###
+
+The length of the master password that protects all the other keys can be set in:
+
+	/crypto/masterpasswordlength
+
+The number of iterations that are to be performed in the PBKDF2 call can be set in:
+
+	/crypto/iterations
```
markus2330: These things seem to be very specific rarely someone wants to change. Are there use cases at all? We should at least move it after the examples.

src/plugins/crypto/README.md
```diff
@@ -109,34 +110,35 @@ Set the CMake variables `OPENSSL_INCLUDE_DIR` and `OPENSSL_LIBRARIES` to your de
...
-	system/elektra/crypto/config/key-derivation/#0 = meta
-	system/elektra/crypto/config/key-derivation/#1 = file
-	system/elektra/crypto/config/key-derivation/#2 = agent
-	system/elektra/crypto/config/key-file/path/#0 = ~/.elektra/id_aes
-	system/elektra/crypto/config/key-file/path/#1 = /etc/elektra/id_aes
+The GPG recipient keys can be specified as `/crypto/key` directly.
+If you want to use more than one key, just enumerate like:
 
-Only keys marked with a certain meta-key will be considered for encryption/decryption.
+	/crypto/key/#0
```
markus2330: Isn't the key only related to gpg? What about having a hierarchy: `/crypto/gpg/binary` (or also keeping `/crypto/gpg`) and `/crypto/gpg/key`? I like the enumeration feature! Can you explain a bit more what happens if multiple keys are given?

markus2330: Maybe `/gpg/key` is even better, because it is the main configuration, it is also nice if it is short.

markus2330:
Looks very nice! I gave some high-level feedback on the documentation, I hope thats okay even though it is "work in progress".

src/plugins/crypto/README.md
markus2330: I forgot about the benchmarks, this is certainly a usecase ;) So for now its okay to expose the details.

But what I mean is: without in-depth knowledge people have no idea which combination of length+iterations are useful. So as goal (after this PR) for standard users we should provide something like "security level".

petermax2:
The unit tests require a GPG key in order to work.

@markus2330 Do you think it's OK for the `testmod_` programs to import an "Elektra Test Key" into the key-ring of the developer/packager? Generating a new key on every run is a bad idea because GPG collects quite some entropy (in a VM it can take minutes until the key is generated).

The alternative is to pass a key ID as command line argument to the `testmod_` programs. But this breaks automatic building and requires re-configuration on the build server. Also I think it is inconvenient for the dev/packager.

Frankly I don't like either of the two solutions. I would not like a test script to "mess around" with my key-ring.

markus2330:
I think there is nothing wrong in adding a private+public key into the repo for testing purposes. If its not signed by many parties, it would be very small anyway. It is also robust to pass `--homedir` to gpg. I only hope nobody will use the key to actually encrypt something of value with it ;)

src/plugins/crypto/CMakeLists.txt
```diff
@@ -194,18 +194,18 @@ add_plugin (crypto_botan
...
 		add_plugintest (crypto_gcrypt
 				MEMLEAK
-				LINK_PLUGIN "<no>")
+				LINK_PLUGIN "crypto_gcrypt")
 	endif ()
 
 	if (HAS_BOTAN_4SURE)
 		add_plugintest (crypto_botan
-				LINK_PLUGIN "<no>")
+				LINK_PLUGIN "crypto_botan")
```
markus2330: What about splitting the tests that load the library dynamically and actually link against them?

src/plugins/crypto/test_internals.h
```diff
@@ -17,68 +18,24 @@
...
-
-	return ksNew (2, configKey, configIv, KS_END);
-}
-
-/**
- * @brief create new KeySet and add an invalid configuration to it.
- *
- * The cryptographic key in the returned KeySet has an invalid size.
- */
-static KeySet * newInvalidConfiguration ()
```
markus2330: Why not testing invalid configurations anymore?

src/plugins/crypto/test_key.h
```diff
@@ -0,0 +1,93 @@
...
+	0x45, 0x6c, 0x54, 0x44, 0x79, 0x6c, 0x58, 0x0a, 0x39, 0x6a, 0x58, 0x68, 0x62, 0x44, 0x5a, 0x68, 0x6b, 0x66, 0x69, 0x57, 0x6a, 0x6b,
+	0x69, 0x62, 0x43, 0x50, 0x73, 0x75, 0x59, 0x2b, 0x37, 0x78, 0x2f, 0x6e, 0x65, 0x4f, 0x5a, 0x6e, 0x49, 0x4f, 0x31, 0x56, 0x74, 0x50,
+	0x73, 0x32, 0x34, 0x76, 0x53, 0x36, 0x6b, 0x39, 0x34, 0x71, 0x34, 0x6c, 0x53, 0x67, 0x43, 0x2f, 0x52, 0x7a, 0x4e, 0x32, 0x66, 0x62,
+	0x79, 0x76, 0x73, 0x70, 0x51, 0x57, 0x0a, 0x6e, 0x4c, 0x51, 0x34, 0x36, 0x43, 0x36, 0x31, 0x59, 0x50, 0x67, 0x51, 0x78, 0x61, 0x51,
+	0x51, 0x54, 0x68, 0x57, 0x6b, 0x35, 0x30, 0x74, 0x2f, 0x45, 0x2f, 0x48, 0x32, 0x69, 0x41, 0x37, 0x38, 0x6b, 0x43, 0x58, 0x52, 0x62,
+	0x51, 0x3d, 0x3d, 0x0a, 0x3d, 0x78, 0x43, 0x41, 0x38, 0x0a, 0x2d, 0x2d, 0x2d, 0x2d, 0x2d, 0x45, 0x4e, 0x44, 0x20, 0x50, 0x47, 0x50,
+	0x20, 0x50, 0x52, 0x49, 0x56, 0x41, 0x54, 0x45, 0x20, 0x4b, 0x45, 0x59, 0x20, 0x42, 0x4c, 0x4f, 0x43, 0x4b, 0x2d, 0x2d, 0x2d, 0x2d,
+	0x2d, 0x0a
+};
+static const size_t test_key_asc_len = 1960;
```
markus2330: Isn't this equivalent/off-by-one to sizeof(test_key_asc)?

src/plugins/crypto/test_internals.h
petermax2: The only possible invalid configuration is one without a GPG recipient key specified. I added a test case for that. Everything else is now handled with default values.

src/plugins/crypto/test_key.h
petermax2: Yes, I took over the output of `xxd -i` but I can adapt to `sizeof(test_key_asc)` if you prefer.

petermax2:
/home/jenkins/workspace/workspace/elektra-mergerequests-unstable/src/plugins/crypto/test_internals.h:236: error in test_gpg: failed to install the GPG test key

@markus2330 are those packages installed on the build server?

- gnupg
- gnupg-agent

src/plugins/crypto/test_key.h
markus2330: Simply put a comment at the beginning that it is generated with `xxd -i`.

markus2330:
jenkins build unstable please

src/plugins/crypto/gpg.c
```diff
@@ -0,0 +1,289 @@
...
+		// fork() failed
+		ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG_FAULT, errorKey, "fork failed");
+		closePipe (pipe_stdin);
+		closePipe (pipe_stdout);
+		elektraFree (buffer);
+		return -1;
+
+	case 0:
+		// start of the forked child process
+		close (pipe_stdin[1]);
```
markus2330: O_CLOEXEC should be more robust (also handles fork/exec not done by us)

src/plugins/crypto/gpg.c
```diff
@@ -0,0 +1,289 @@
...
+
+		// redirect stdout to pipe
+		close (STDOUT_FILENO);
+		dup (pipe_stdout[1]);
+		close (pipe_stdout[1]);
+
+		// finally call the gpg executable
+		if (execv (argv[0], argv) < 0)
+		{
+			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_CRYPTO_GPG_FAULT, errorKey, "failed to start the gpg binary: %s", argv[0]);
```
markus2330: Can you improve this error?
E.g. check if the gpg binary exists and is executable.

src/plugins/crypto/gpg.c
```diff
@@ -0,0 +1,289 @@
...
+		outputLen = read (pipe_stdout[0], buffer, bufferSize);
+		keySetBinary (msgKey, buffer, outputLen);
+	}
+	close (pipe_stdout[0]);
+	elektraFree (buffer);
+
+	if (status != 0)
+	{
+		// gpg exited with an error
+		ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_CRYPTO_GPG_FAULT, errorKey, "GPG returned %d", status);
```
markus2330: Also here the error could be improved, at least handle wrong signature (return value 1) differently.

markus2330:
jenkins build unstable please

markus2330:
Seems like that on debian unstable the package "gnupg2" is necessary to have /usr/bin/gpg2. I installed it now. What about checking for both files and use what is available (by default, if the user specified something, only this should be used).

petermax2:
> Seems like that on debian unstable the package "gnupg2" is necessary to have /usr/bin/gpg2. I installed it now. What about checking for both files and use what is available (by default, if the user specified something, only this should be used).

Yeah, I also planned on checking for the executables at build time in CMake to set proper defaults in `gpg.h` . Checking for both versions is a good idea.

markus2330:
I think we should make gpg a run-time dependency only, so cmake does not need to know anything about it. Then we would check the binaries with `access` at run-time. This also leads also to better errors. (File X is missing/no access/not executable is way better than "exec" failed). What about searching `gpg` even in PATH?

Btw. it seems like installing `gnupg2` resolved the build problem.

src/plugins/crypto/test_key.h
petermax2: done

src/plugins/crypto/gpg.c
petermax2: check has been added

petermax2:
> I think we should make gpg a run-time dependency only

Agreed. Improves flexibility a lot.

> Then we would check the binaries with access at run-time.

Done.

> What about searching gpg even in PATH?

Good idea! Coming up next.

petermax2:
Found segfault in `crypto_gcrypt` and `crypto_openssl`. Steps to reproduce:

    kdb mount test.ecf user/t crypto_gcrypt "/gpg/key=DDEBEF9EE2DC931701338212DAF635B17F230E8D" 
    kdb set user/t/a "I am too long to fit into a single AES block :-("
    kdb setmeta user/t/a crypto/encrypt 1 

There seems to be a problem at the split of the value.

petermax2:
@markus2330 review pls!

doc/help/elektra-plugins-framework.md
```diff
@@ -108,6 +108,22 @@ have to `return 1` at the end of the contract to not execute the regular
 functionality of the plugin.
 
 
+## Exports
```
markus2330: Why is this reported by github as change? It is already part of master..

markus2330:
Thank you for the update!

But it looks like rebasing was only done partly? Not all of your commits are placed after any of master's commits...

This also makes the changes shown in github huge, so it would be much easier to review the PR if you can fix the rebase first.

src/plugins/crypto/gpg.c
```diff
@@ -0,0 +1,439 @@
...
+	kdb_octet_t * buffer = NULL;
+	const ssize_t bufferSize = 2 * keyGetValueSize (msgKey);
+	ssize_t outputLen;
+
+	assert (argc > 1);
+
+	// sanitize the argument vector
+	if (getGpgBinary (&argv[0], conf, errorKey) != 1)
+	{
+		return -1;
```
markus2330: No error set here, but it seems like (at least some) callers expect the error already to be set. It would be good if you can document who is responsible for setting errors.

petermax2:
> But it looks like rebasing was only done partly? Not all of your commits are placed after any of master's commits...

I messed up every rebase ever since... =( I will reopen the PR with fixed history.

PR #904, opened 2016-08-31
configure-firefox: fixed locking preferences

tom-wa:
Fixed locking values at runtime

seems like i screwed up and FF supports locking existing values


markus2330:
Thank you. So no chance that locked values can be changed? I am looking forward to your presentation today.

PR #905, opened 2016-08-31
adjusted examples/spec to new conditionals syntax

tom-wa:
updated `examples/spec` files to new `conditionals` syntax


markus2330:
Thank you!

Issue #903, opened 2016-08-31
Shell Recorder protocol formating broken


tom-wa:
# Describe what you wanted to do

read the shell recorder protocol after running `shell_recorder.sh db_changes.esr`

# Describe what you expected
```
  1 CMD: set /test/key val^@
  2 RET: 0^@
  3 ERRORS: ^@
  4 WARNINGS: ^@
  5 STDOUT: Using name user/test/key
  6 Create a new key user/test/key with string val^@
  7 STDERR: ^@
  8 DIFF: 2c2,5
  9 < ksNew 0
 10 ---
....
 ```
# Describe what actually happened
```
  1 CMD: set user/test/key val^@RET: 0^@STDERR: ^@STDOUT: Create a new key user/test/key with string val^@WARNINGS: ^@ERRORS: ^@DIFF: 2c2,22
  2 < ksNew 0
  3 ---
...
```
# System Information

- Elektra Version: master

# Further Log Files and Output
-

# Other

seems like #862 broke the formating

PR #906, opened 2016-08-31
Shell Recorder: Fix Broken Protocol Files

sanssecours:
# Purpose

Before this change Shell Recorder would not print newlines after it wrote a message to the protocol file. This commit fixes this regression introduced in commit 5d2232e.

The problem was that we replaced `echo -e` with `printf` but did not consider that `printf` does not print a newline at the end of a message.

This commit addresses issue #903.

# Checklist

- [x] I checked the commit message.

@markus2330 @tom-wa: Please review my pull request.

### DIFF
--- a/tests/shell/shell_recorder/shell_recorder.sh
+++ b/tests/shell/shell_recorder/shell_recorder.sh
@@ -77,13 +77,13 @@ execute()
 
     echo "$command"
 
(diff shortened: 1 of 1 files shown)


markus2330:
Ok, looks fine, @tom-wa Can you also take a look?

tom-wa:
seems to solve the problem.

markus2330:
Thank you!

Issue #867, opened 2016-07-29
proposal: kdbLookup


markus2330:
As discussed with @KurtMi the main purpose of run-time logging would be to inspect which `ksLookup` an application is doing. To avoid all the issues a more general logging framework in the library would have, we could extend the API with `kdbLookup` where lookup-plugins (similar to global Plugins) can intercept every lookup call.

This way we could implement the current ksLookup printer that `kdb get`, and the context-awareness features in a way accessible to every application.

The proposed API is:
`Key * kdbLookup (KDB *kdb, KeySet *haystack, Key *needle);`

What do you think?

KurtMi:
So you propose to not implement a logger and instead to add a hook to the lookup?

If the hook acts like a gobal hook, meaning that we only add one or two new positions for the globalhooks like `prelookup` and `postlookup` it would be fine.

I do not understand why we need a new Api for that? And what is context-awareness?

markus2330:
Hi! There is no plan change for the static logger. The new API and hooks is for run-time logging (and more) of `ksLookup`. Maybe we even only need a single position `lookup` (no pre/post) with the hook already present, see keyset.c line 1896. The current (internal) ksLookup plugin API is:
```
Key * elektraLookupHook (KeySet * ks, Key * key, Key * found, option_t options)
```
For examples see src/tools/kdb/get.cpp. Or do you think we can achieve more with `pre/post`?

> I do not understand why we need a new Api for that?

The API is needed because from `ksLookup` you cannot get references to plugins which live in `kdb`.

> And what is context-awareness?

Context awareness is (very briefly) explained in src/libs/getenv/README.md, section CONTEXT. More info, e.g. in ["Unanticipated Context Awareness for Software Configuration"](http://www.libelektra.org/ftp/papers/icis2016unanticipated.pdf)

markus2330:
Additionally to the main hook, pre/post lookup hooks might be useful so that is clear for a plugin which lookup hook was the first/last (e.g. to do preparations/finalizations within the plugin).

The pre/post hooks might also be useful as locking solution when applications exclusively use kdb* functions (or work with the same lock while tampering with the data structures, i.e. ks* key* methods).

markus2330:
Maybe for Elektra 2.0, I close it for now.

PR #907, opened 2016-08-31
conditionals: now with topological sort

tom-wa:
# Purpose

Added topological sorting 
# Checklist
- [ ] I added unit tests

### DIFF
--- a/src/plugins/conditionals/conditionals.c
+++ b/src/plugins/conditionals/conditionals.c
@@ -16,6 +16,8 @@
 #include <errno.h>
 #include <kdbease.h>
 #include <kdberrors.h>
+#include <kdbmeta.h>
+#include <kdbproposal.h>
 #include <math.h>
 #include <regex.h>
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


src/plugins/conditionals/conditionals.c
```diff
@@ -785,6 +787,129 @@ static CondResult evaluateKey (const Key * meta, const Key * suffixList, Key * p
 	return TRUE;
 }
 
+static KeySet * getDependencies (const char * expr, Key * cur, Key * parentKey)
+{
+	char * localExpr = strdup (expr);
```
markus2330: please use elektraStrDup.

src/plugins/conditionals/conditionals.c
```diff
@@ -785,6 +787,129 @@ static CondResult evaluateKey (const Key * meta, const Key * suffixList, Key * p
...
+	char * ptr = localExpr;
+	KeySet * deps = ksNew (0, KS_END);
+	int isEscaped = 0;
+	int hasQuote = 0;
+	for (; *ptr; ++ptr)
```
markus2330: Shouldn't this parsing code be part of libmeta (later)? Some generic link detection would be nice.

markus2330:
Please add unit tests. What else is missing?

tom-wa:
>Shouldn't this parsing code be part of libmeta (later)?

you are right. currently conditionals is the only plugin that benefits from it though, so let's keep it that way for now, i'll improve it and move it to libmeta later.

>What else is missing?

a lot of unit tests, +a few things to discuss: 
for now, should the topological sort be used by default or activated by a flag (until it's proven to work) ?
what should be done if there are cyclic dependencies or invalid dependencies ?

markus2330:
> should the topological sort be used by default

I think we should not do this for the next release anyway, but the idea is that it should be default when any calculations are involved.

In longer term we should split the plugin to a validation and a calculation plugin. The calculation should be done at lookup time (see #867). Only the calculation needs correct ordering, for validation the order only matters for better error messages, but is not needed for correctness.

markus2330:

Issue #401, opened 2015-11-23
Inconsistent set/get for cascading keys


omnidan:
I have a `test.ini` file with the following contents mounted at `/test`:

```
[test]
hello = world
```

After running the following commands:

```
dan@vagrant-ubuntu-vivid-64:~/libelektra/build/bin$ ./kdb get /test/test/hello
world
dan@vagrant-ubuntu-vivid-64:~/libelektra/build/bin$ ./kdb set /test/test/hello earth
Using name user/test/test/hello
Create a new key user/test/test/hello with string earth
dan@vagrant-ubuntu-vivid-64:~/libelektra/build/bin$ ./kdb get /test/test/hello
earth
```

The key gets written to KDB, but not to the mounted file - this seems to be an issue with the resolver.

Using `system/test` as a mountpoint works fine.

markus2330:
Be extra careful with cascading names, you might get something different than you write to. You can use `-v` to check.

omnidan:
Sorry for the late response, I kinda overlooked this issue. :sweat_smile: Output with `-v`: 

```
dan@vagrant-ubuntu-vivid-64:~/libelektra/build/bin$ ./kdb get -v /test/test/hello
got 9 keys
searching spec/test/test/hello, found: <nothing>, options: KDB_O_CALLBACK
    searching proc/test/test/hello, found: <nothing>, options: 
    searching dir/test/test/hello, found: dir/test/test/hello, options: 
The resulting keyname is dir/test/test/hello
earth
```

markus2330:
Ok, seems like you actually got the key `dir/test/test/hello` and not `user/test/test/hello` as you expected.

Writing to cascading keys is sometimes not intutive, I agree that it is not a good idea that a `get` afterwards should return what you have `set` before.

Maybe we should prohibit the writing to cascading keys at all, it causes more problems than it solves.

markus2330:
Please reopen on any idea how to improve this. Documentation got some updates and `-v` tells quite good what `kdb get` does.

markus2330:
Reopened: we should at least warn:

- if a `kdb set`is hidden by an already present key
- if `kdb set/rm` points to some other keys

markus2330:
To be solved with #3742

PR #912, opened 2016-09-01
elektra-web HTTP API

omnidan:
# Purpose

This is a proposal to allow remote configuration of (multiple) elektra instances.
- elektrad: http://docs.elektrad.apiary.io/
- clusterd: http://docs.clusterd.apiary.io/
# Checklist
- [x] split into 2 blueprints
- [x] move to doc/apiblueprints/
- [x] use `{+path}` instead of `{path}` - https://apiblueprint.org/documentation/specification.html#variable-with-reserved-characters-values
- [x] return `ls`/`value`/`meta` when accessing kdb paths via `GET`
- [x] http error/response codes
- [x] ~~output kdb errors via api~~ does this really make sense since we're not dealing with plugins?
- [x] ~~change links to point to the official repo~~


markus2330:
@Namoshek Can you take a look?

doc/webui/README.md
```diff
@@ -0,0 +1,66 @@
...
+ * **GET /kdb/:path** - get `path` configuration (similar to `kdb get path`)
+ * **POST /kdb/:path** - edit `path` configuration (similar to `kdb set path`)
+
+The cluster management server (`clusterd`) also provides a RESTful HTTP API.
+Single instances can be configured as follows:
+
+ * **GET /instances** - get a list of all instances
+ * **POST /instances** - create a new instance
+ * **GET /instances/:id** - get information about a single instance
+ * **POST /instances/:id** - edit a single instance
```
Namoshek: For updating/patching you should use **PUT** or **PATCH**.
You will also need a method for deleting instances, I guess. Use **DELETE** for this task.

doc/webui/README.md
```diff
@@ -0,0 +1,66 @@
...
+ * **POST /instances/:id/kdb/:path** - edit `path` configuration of an instance (similar to `kdb set path`)
+ * **GET /instances/:id/version** - get `elektrad` version of an instance
+
+It is also possible to create and manage groups of multiple elektra instances (clusters):
+
+ * **GET /version** - get `clusterd` version
+ * **GET /clusters** - get a list of all clusters
+ * **POST /clusters** - create a new cluster
+ * **GET /clusters/:id** - get information about a single cluster
+ * **POST /clusters/:id** - edit a single cluster
```
Namoshek: Same as for the `instance` interface.

doc/webui/README.md
markus2330: Both PUT or PATCH make sense, but we cannot use [json merge patch](https://tools.ietf.org/html/rfc7396) because in Elektra null and absent is something different.

With put we might not need DELETE, if everything missing in PUT is deleted.

doc/webui/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+_a web user interface (Web UI) to remotely manage multiple elektra instances_
+
+
+## Overview
+
+Elektra web consists of multiple components:
+
+ * (multiple) servers running an elektra daemon (`elektrad`)
+ * a single cluster management server to communicate with the elektra daemons (`clusterd`)
```
markus2330: I think its better to remove the architecture parts from here and put them into an extra document as discussed. This document should only be about the HTTP API.

doc/webui/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+## API
+
+![https://cdn.rawgit.com/ElektraInitiative/libelektra/http-api-proposal/doc/webui/daemon_structure.png](https://cdn.rawgit.com/ElektraInitiative/libelektra/http-api-proposal/doc/webui/daemon_structure.png)
+
+To access single instances, each elektra daemon (`elektrad`) provides a RESTful
+HTTP API:
+
+ * **GET /version** - get `elektrad` version
```
markus2330: Which format? I would only return Elektra's and HTTP API's version, and not invent one for the tool.

doc/webui/README.md
```diff
@@ -0,0 +1,66 @@
...
+ * **GET /kdb/:path** - get `path` configuration (similar to `kdb get path`)
+ * **POST /kdb/:path** - edit `path` configuration (similar to `kdb set path`)
```
markus2330: If it is similar, what are the differences, can you give examples?

doc/webui/README.md
markus2330: Can we generalize the API such that we do not need different API descriptions for `instances` and `clusters`?

markus2330:
I am missing how to update parts of the configuration tree and how to select different configuration formats for transfer.

Namoshek:
So far the API descriptions seems clear and well thought, except for the HTTP methods here and there. When you change them, make sure you only use either **PUT** or **PATCH**, but not both. I think they are interchangable, although some might say that **PATCH** is used for modifications and **PUT** for replacements (although this differentiation is quite hard in most scenarios). For removing resources obviously use **DELETE**. Concerning the rest of your proposal and the structure, I think I would have chosen a similar approach.

For further definition of the REST API, I have some thoughs on my mind that might be helpful for you:
* How are you handling meta data?
* How are you handling plugin operations / other common tasks (do you plan to support them at all)? 
<br/>(I think these two points are quite difficult to answer, because there are basically only two options: Use unique API endpoints for each operation, which will bloat the whole API, or encapsulate the operations within one API endpoint, which doesn't really fit the RESTful approach...)
* Your HTTP response codes for the `elektrad` daemon are quite clear, because the daemon can select them on his very own for a single operation. But how are you going to select the response codes for the `clusterd` daemon, if the response codes of the underlying instances are not all the same (e.g. one instance failed the operation)? (There is a multi-status response code (207) which could be used for the outer response and then you could send the individual response codes as part of the response message for each instance. This way you could probably also avoid having to implement transaction logic (not sure if this would be possible at all, and quite sure this would be out of scope for your work).)
* What response codes will you be using within your system at all? (Probably 200, 404, maybe 403, ...)
* When you setup an instance, you'll have to setup the `elektrad` daemon individually via SSH or something. Instead of having to put the daemon into the cluster configuration by hand after that operation, the daemon could register himself (based on a configuration file e.g.). This would also make things easier if the IP or domain of one of your machines (instances) changes for example. Of course, if your cluster management server would change his address, you would have to change the configuration for each machine by hand.
* What is your plan for authorization / security?
* Will hierarchical clusters be a thing (e.g. cluster A only encapsulates clusters B and C, which then hold real instances)?

I think that is enough to think about for the moment. :smile:

doc/webui/README.md
Namoshek: I disagree, because he references the architecture within his API description. The component description in this document isn't really extensive, so I don't see a problem at all. He could still write a lot more about it in another document, that is correct.

doc/webui/README.md
Namoshek: He didn't specify what the version is used for, so I didn't comment it so far. But you are right, version management could get easily a huge task that isn't really in scope of his work.

Namoshek:
> I am missing how to update parts of the configuration tree

Which configuration tree are you talking about? Shouldn't this be easily possible with his 
> **POST /kdb/:path** - edit `path` configuration (similar to `kdb set path`)

 operation? (For this operation I think **POST** is an appropriate HTTP method by the way.)

markus2330:
@Namoshek Thank you for the detailed feedback!

> meta data

There are configuration formats that handle both config+metadata within the same syntax, so if we have support for arbitrary syntax, we would also handle metadata without more API. It won't work with JSON though.

> plugin operations / other common tasks

Which one do you refer to? mounting and such? All of them are basically only key-value operations, but you are right that non-trivial algorithms are involved you want to avoid reimplementing in every language. @BernhardDenner is looking for a good solution here, too.

> Which configuration tree are you talking about?

The hierarchy built up within Elektra separated with `/`. If the API only supports `get`/`set` I am missing something like `import`/`export`. In any case it is not clear what "similar to get" means.

Namespace support is also unclear.

Namoshek:
> There are configuration formats that handle both config+metadata within the same syntax, so if we have support for arbitrary syntax, we would also handle metadata without more API...

... which also depends on the storage plugin you use. The old story, you know. :smile: 
But because of your comment on this, the question how he is planning to manipulate individual endpoints arises immediately. From the current description it is not clear to me how much a single operation actually can do (e.g. only write a string as value for a key or place whole snippets under a key, like I do it for my application). Answering this would also require to think about blueprints for messages already, I guess.

> Which one do you refer to? mounting and such? All of them are basically only key-value operations, but you are right that non-trivial algorithms are involved you want to avoid reimplementing in every language.

Yes, that is what i meant. Implementing these functions within his REST API a second time doesn't sound good, because it would potentially lead to unintended and unexpected behavior in later versions (if something of the plugin system changes). Of course you could leave this task open to the end-user, so he could simply use the normal interface, but I'm not sure if this is userfriendly at all.

> The hierarchy built up within Elektra separated with /. If the API only supports get/set I am missing something like import/export. In any case it is not clear what "similar to get" means.

Ah yes, that is what I'm talking about in the first paragraph. Abstracting the import/export function could be done via an additional HTTP parameter repesenting the way how to handle the supplied value (e.g. no parameter = simple value, parameter set = use storage plugin for parameter or something similar).

markus2330:
> place whole snippets under a key

This should only be a workaround for non-supported formats (actually even in this case the line-plugin would be preferable). It would not be possible to transform snippets to other formats nor to validate them if the whole snippet is within a key.

Namoshek:
> This should only be a workaround for non-supported formats (actually even in this case the line-plugin would be preferable). It would not be possible to transform snippets to other formats nor to validate them if the whole snippet is within a key.

What I actually meant was that you can place the configuration encapsulated by a snippet in the hierarchy below a certain key, not the snippet as standalone value. But it should still be possible to set single values as well, in my opinion. So the interface would actually offer two operations.

### Import of snippets

> key1 = value1
> key2 = value2

as ini-script imported for path `user/some/example/path` would result in two keys being stored in the database

> kdb set user/some/example/path/key1 "value1"
> kdb set user/some/example/path/key2 "value2"

In other words, the provided path would be the root key for the snippet. Namespaces could be done via an optional parameter as well.

### Setting of basic values

> value1

as single value set for path `user/some/example/path` would result in one key being stored in the database

> kdb set user/some/example/path "value1"

Of course there must be some way to tell the API how the provided value has to be treated.

doc/webui/README.md
omnidan: I changed it to `PUT` and added `DELETE`. I'll still keep `POST` for the `/kdb/:path` commands, because the same method will be used for creating and editing keys (could additionally add `PUT` which doesn't allow setting keys that don't exist, not sure if there's a use case for that, though)

doc/webui/README.md
omnidan: Good idea, I removed the redundant API and added a note.

doc/webui/README.md
omnidan: I agree that it would make sense to have a separate document that goes into details of the architecture. Here I'm just giving a small overview to make it easier to understand the API.

doc/webui/README.md
omnidan: I would make all responses in JSON by default. Later, we can allow adding a suffix, e.g. `.xml` to get XML output. What do you think?

doc/webui/README.md
omnidan: Right now, it actually just calls `kdb get` and `kdb set`, so it's exactly the same. I changed the text.

omnidan:
Thanks for clarifying the `set` mechanism, @Namoshek !

It would definitely be possible to save snippets under a sub-tree (specified via `:path`). How do we tell the API that we're saving a snippet and not raw data, though? (maybe a HTTP parameter, e.g. `?data=snippet`)

All additional features could be added as HTTP parameters, e.g. `?action=import`.

omnidan:
I have updated the document, please re-review.

> But how are you going to select the response codes for the clusterd daemon, if the response codes of the underlying instances are not all the same (e.g. one instance failed the operation)?
I would combine the responses from the `elektrad` daemons (would also include the HTTP error code). If everything is a success, the status code of the combined document will be 200. Otherwise it will show an error (400).

> What response codes will you be using within your system at all? (Probably 200, 404, maybe 403, ...)
200, 404 for non-existing keys, 403 for authorization failures

> When you setup an instance, you'll have to setup the elektrad daemon individually via SSH or something. Instead of having to put the daemon into the cluster configuration by hand after that operation, the daemon could register himself (based on a configuration file e.g.).
Good idea!

> What is your plan for authorization / security?
`clusterd` could provide an API key (maybe a separate one for each instance) that has to be entered when setting up an `elektrad` instance (in addition to the `clusterd` IP)

> Will hierarchical clusters be a thing (e.g. cluster A only encapsulates clusters B and C, which then hold real instances)?
I haven't considered that and I'm not sure if there's a use case for it, but it's an interesting idea.

markus2330:
I would prefer to directly review the blueprint version where the tools say the syntax is ok and which already has examples. Please make sure to respond to all concerns in the new document. If unsure you can add TODO markers.

doc/webui/README.md
markus2330: I would prefer to use something like `?format=<pluginname>`. With a suffix you never know if the key is named this way or if its a request to a different syntax.

markus2330:
> I haven't considered that and I'm not sure if there's a use case for it, but it's an interesting idea.

There is definitely a use-case to *not* have a clusterd (to connect a browser to `elektrad`), so a proper recursive variants solves both issues:

1. The browser can directly connect to either clusterd or elektrad
2. Clusterd can connect to either elektrad and clusterd

To make recursive nesting easily possible, the API for elektrad and clusterd must be identical. Then you need something like `GET topology` which is trivial for elektrad (return hostname+version), but works recursively for clusterd (return a struct containing the whole topology).

Namoshek:
For me, *hierarchical clusters* does not imply running several `clusterd` daemons, although it could make sense (but then the same interface would be required for the `clusterd` daemon and the `elektrad` daemon). I was simply talking about hierarchical/recursive clusters within *the one* cluster service.

markus2330:
> running several clusterd

The question is if clusterd is needed at all (either elektrad can manage clusters or the browser can also connect to several elektrad). But that is a different architectural decisions which should not be a part of this document (as already stated).

> but then the same interface would be required

It would be a better design to only use different APIs if it really abstracts something inherently different. It does not seem like that the abstraction differs if the API only merges some results together.

> hierarchical clusters were a thing

Up to now I do not see a reason why we should forbid them.

Btw. @omnidan Please do not squash commits (as you have in the TODO above).

Namoshek:
> The question is if clusterd is needed at all (either elektrad can manage clusters or the browser can also connect to several elektrad). But that is a different architectural decisions which should not be a part of this document (as already stated).

Although you could connect with the browser to multiple machines, you'd still need to store the configuration for your clusters somewhere. The browsers local cache doesn't seem appropriate to me, but with a configuration export & import feature, it could be interesting again. Maybe make another issue for such a discussion and brainstorming, @omnidan?

> It would be a better design to only use different APIs if it really abstracts something inherently different. It does not seem like that the abstraction differs if the API only merges some results together.

Be careful with that, I already wrote something about multi-status for a reason. Inventing a multi-status for single resources is not a good design and the other way round you would invent lots of possible bugs and inconveniences when using only a single status code for a whole batch of operations. So there is actually a good reason to use two interfaces.

markus2330:
> you'd still need to store the configuration for your clusters somewhere

One way without information stored in the browser would be that every node contains its peers. But you are right, there should be a seperate decision for this.

> [single status for many resources] invent lots of possible bugs and inconveniences when using only a single status code for a whole batch of operations

This is only the case if you do not have a proper 2phase commit. If you fail and do nothing, everything is fine. Elektra works this way already: when one plugin fail to write the configuration, the overall status is failure and nothing will be done.

doc/webui/README.md
markus2330: Yes, but we should discuss (and describe) how it should be, not like an implementation currently does.

markus2330:
@Namoshek It would also be good to see your API (maybe directly in blueprint).

Namoshek:
Yes, I'm currently writing it down as blueprint. I'll push a first draft of the first methods in a few seconds, maybe you have a look and tell me if it looks fine. The API blueprints give actually quite a few ways to go about things, there is not really the one and only way.

omnidan:
I've updated the README.md to be apiblueprint syntax. I also uploaded it via apiary.io: http://docs.elektraweb.apiary.io/

I checked the format using the apiary validation check (I think it uses [`drafter`](https://github.com/apiaryio/drafter)) and running `drafter` on the README.md file

doc/webui/README.md
```diff
@@ -24,46 +26,233 @@ single instances can be configured independently.
...
+#### GET /version
+
+get API and elektra version
+
++ Response 200 (application/json)
+
+        {
+            "api": 1,
+            "elektra": "0.8.17"
+        }
```
Namoshek: I would use an attribute list here, apiary can then generate nice doc + json schema from it. You can also describe each field with it individually.

doc/webui/README.md
```diff
@@ -24,46 +26,233 @@ single instances can be configured independently.
...
++ Parameters
+    + path: `user/hello` (string) - path to the elektra config
+
+##### get configuration [GET]
+
+this is the same as calling `kdb get {path}`
+
++ Response 200 (application/json)
+
+        "hello world"
```
Namoshek: Will this evolve into a real json response or will it be plain text?

doc/webui/README.md
```diff
@@ -24,46 +26,233 @@ single instances can be configured independently.
...
-to all instances.
 
-### Response codes
+#### /clusters
 
- * 200: success
- * 404: non-existing key
- * 403: authorization failure
+It is also possible to create and manage groups of multiple elektra instances (clusters).
+The API is the same as above, but with `/clusters` instead of `/instances`.
```
Namoshek: You'll obivously need to write this formally, plain text won't work here anymore.

Namoshek:
Have you already thought about authorization mechanism and how they will reflect on the API (e.g. 401 unauthorized responses)?

doc/webui/README.md
omnidan: it will be JSON in some cases, but in most cases it's probably just a string (enclosed in `"`, as per the JSON standard) or number

doc/webui/README.md
omnidan: I know, but I haven't found a decent solution yet, besides simply copy & pasting `/instances` and renaming it all to `/clusters` - I don't think you can re-use whole endpoints with apiblueprints

omnidan:
> Have you already thought about authorization mechanism and how they will reflect on the API (e.g. 401 unauthorized responses)?

No, I haven't. My idea would be making access to `/instances` and `/clusters` restricted by user accounts (maybe using PAM on the `clusterd` server). Then the authenticated user could generate an api key to allow `elektrad` instances to register themselves with a `clusterd` instance (by calling `/register?apikey=xxx`). When `/register`ing, `elektrad` would also have to send a generated key to the `clusterd`, which will be used to access the API of the instance later. What do you think?

doc/webui/README.md
omnidan: Good idea! I just looked into attributes and data structures and switched all parts of the api to it, much more flexible and no redundancies :ok_hand:

doc/webui/README.md
```diff
@@ -117,10 +111,9 @@ API to allow instances to register themselves with `clusterd`
 
 + Request (application/json)
 
-        {
-            "name": "auto-registered instance",
-            "host": "192.168.0.8"
-        }
+    + Attributes (Base Instance)
+        + host: 192.168.0.8
+        + name: autoregistered instance
```
Namoshek: If you specify the same fields here and in the *Base Instance* model again, you'll override the *Base Instance* model with these values. If you don't want to override anything, you should not specify them twice I think.

doc/webui/README.md
```diff
@@ -132,56 +125,37 @@ API to allow instances to register themselves with `clusterd`
...
-                "name": "test instance #2",
-                "host": "192.168.0.6"
-            }
-        ]
+    + Attributes (array[Instance])
 
 ##### create a new instance [POST]
 
++ Attributes (Base Instance)
+
```
Namoshek: Attributes above the request?

doc/webui/README.md
omnidan: Ah, so it overrides the type/required properties too? I thought it'd just override the values.

doc/webui/README.md
omnidan: That's how it was done in the [example](https://raw.githubusercontent.com/apiaryio/api-blueprint/master/examples/10.%20Data%20Structures.md)

doc/webui/README.md
Namoshek: I only tried overriding the description, so I cannot really give you a good answer on this. But why do you want to override the examples?

doc/webui/README.md
Namoshek: Interesting, maybe this would allow to re-use the attributes for multiple request types (e.g. json + xml)? Haven't seen it so far, but if it works it is ok I guess.

doc/webui/README.md
omnidan: I'll just remove it then, there's no need to override the examples

omnidan: Actually, it would be nice to do it in this case (updating a property):

```
    + Attributes (object)
        + host: 192.168.0.6
```

omnidan:
@markus2330 @Namoshek I have updated the blueprints, please let me know if everything is fine now.

About the kdb errors - is there anything besides plugins that trigger them? I'm not sure if it'll make sense to deal with them specifically in my API (I would just output the error in my frontend), I would just put the stderr result into the `message` field in case an edge case does happen.

doc/apiblueprints/elektrad.apib
```diff
@@ -0,0 +1,75 @@
...
+## elektra key database [/kdb/{+path}]
+
+access the elektra key database by specifying a `path`
+
++ Parameters
+    + path: `user/hello` (string) - path to the elektra config
+
+### get configuration [GET]
+
+this is the same as calling `kdb get {path}`
```
markus2330: Is this only the split-up or did you already update these parts? You should describe the behaviour with text or by means of C-API, not by referring to other implementations.

And we should not repeat problems already known: e.g. #401

doc/apiblueprints/clusterd.apib
```diff
@@ -1,117 +1,27 @@
 FORMAT: 1A
 
-# elektra web
+# clusterd API
 
-_an API and web user interface to remotely manage multiple elektra instances_
+the cluster management server (`clusterd`) provides a REST HTTP API to control
+multiple `elektrad` instances
```
omnidan: see https://github.com/ElektraInitiative/libelektra/pull/983#discussion_r82032831

doc/apiblueprints/elektrad.apib
omnidan: it is already updated, but I forgot to remove these descriptions (which aren't accurate anymore now)

omnidan:
@markus2330 please re-review

doc/apiblueprints/clusterd.apib
```diff
@@ -0,0 +1,604 @@
...
+
++ Response 406 (application/json)
+    + Attributes (Error)
+        + i18n (string) - a unique token representing above error description message; can be used for internationalization in frontends
+            A comprehensive list of possible errors:
+            - `REQUEST_UNSUPPORTED_CONTENT_TYPE` if the submitted Content-Type is not `application/json` or `text/plain`
+
+### delete configuration [DELETE]
+
+this is the same as calling `kdb rm {path}` on the instance
```
markus2330: Please specify behavior

omnidan: oops, I missed that - fixed now!

markus2330:
@Namoshek Can you check if everything is fine now and consistent with your API?

doc/apiblueprints/elektrad.apib
```diff
@@ -0,0 +1,134 @@
...
+    + Attributes (object)
+        + ls: user/hello, user/hello/world (array[string], required) - subkeys of the requested path, similar to `kdb ls`
+        + value: hello world (string, required)
+        + meta (object, required) - metadata of the requested path
+
++ Response 400 (application/json)
+    + Attributes (Error)
+        + i18n
+            A comprehensive list of possible errors:
+            - `APIKEY_MISSING` api key not specified
```
Namoshek: Does drafter accept this / it is rendered correctly? I thought you need an empty line before the "extended description" (i.e. add empty line between 46 and 47).

I also think you should write the type to `i18n` even if string is the default type, it makes the whole thing more clear.

doc/apiblueprints/elektrad.apib
```diff
@@ -0,0 +1,134 @@
...
+
++ Request (text/plain)
+
+        hello world
+
++ Response 204
+
++ Request (application/json)
+
+        "hello world"
```
Namoshek: Is this valid json?

doc/apiblueprints/clusterd.apib
```diff
@@ -0,0 +1,604 @@
...
+    + Attributes (Error)
+        + i18n (string) - a unique token representing above error description message; can be used for internationalization in frontends
+            A comprehensive list of possible errors:
+            - `INSTANCE_NOT_FOUND` if the requested instance does not exist in the database
+
+### edit a single instance [PUT]
+
++ Request update host of instance (application/json)
+    + Attributes (object)
+        + host: 192.168.0.6
```
Namoshek: Make type `string` explicit?

doc/apiblueprints/elektrad.apib
markus2330: It would be great to have an import/export interface where someone can directly upload/download snippets without packaging it into JSON.

The mime type should be fixed then, though.

Namoshek: You mean like I've done it where you send the plain text snippet in one field and its format in another field (of a json object)?

markus2330: No, I mean the format encoded in the URL, and the content is the configuration file as-is.

doc/apiblueprints/elektrad.apib
omnidan: I just checked and yes, it is rendered correctly. Thanks for the notice, I added the string type now!

doc/apiblueprints/elektrad.apib
omnidan: @Namoshek yes, it is valid json (simply a json string)

@markus2330 more mime types could be implemented the same way as json is implemented right now (distinguish via request header) - which formats were you thinking of? which ones should be specified in the blueprint?

doc/apiblueprints/clusterd.apib
omnidan: Nice catch - I did that now!

omnidan:
@Namoshek thanks for reviewing! let me know if everything looks fine now :grin:

doc/apiblueprints/elektrad.apib
Namoshek: I think he means something like
```
POST [...]/set/path/to/var?format=augeas%20lens=Hosts.lns
Content-Type: text/plain

Body:
127.0.0.1 localhost loopback-adapter
192.168.0.1 gateway
```
Or in other words: Give format as GET parameter + whole payload is the snippet in plain text.
But normally one would avoid GET parameters in a POST request I think...

Namoshek:
To me it looks fine now. :+1:

markus2330:
Thanks! Added #1234 for issues to be addressed later (after first version of implementation is merged).

PR #921, opened 2016-09-06
REST API: sharing of configuration snippets

Namoshek:
# Purpose

REST API description for the service that allows for sharing of configuration snippets.
# Additional info

I checked the format with [Drafter](https://github.com/apiaryio/drafter) and generated a HTML documentation with the [Apairy CLI](https://github.com/apiaryio/apiary-client).
# TODO
- [X] Authentication endpoint
- [x] Database endpoint
# Checklist


Namoshek:
So far I was unable to find a way to put schemata into a variable or something similar, so I copied it for each response. But I really don't like it doing this way. Maybe @omnidan has an idea?

markus2330:
What was the output of drafter (warnings?) and is the HTML docu available somewhere?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+
++ Response 200 (text/html; charset=utf-8)
+
+
+
+# Authentication endpoint [/auth]
+
+## Authenticate [POST]
+
+The API will try to authenticate the user by the submitted credentials. In case of success a session token (JWT) will be returned, which can be used for further requests to the API.
```
markus2330: linebreaks?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+# REST service API for sharing configuration snippets
+
+This is a description of the API provided by the REST service that can be used to share, search and download configuration snippets.
+
+
+# API Description [/]
+
+## List API methods [GET]
+
+Returns an overview of available API methods with required parameters and a description.
```
markus2330: format, example?

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Why JWT?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+# Authentication endpoint [/auth]
+
+## Authenticate [POST]
+
+The API will try to authenticate the user by the submitted credentials. In case of success a session token (JWT) will be returned, which can be used for further requests to the API.
+
++ Request (application/x-www-form-urlencoded)
+    + Attributes
+        + username: the_username (string, required)
+        + password: aSecretPassword (string, required)
```
markus2330: Plaintext password?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+            }
+
++ Response 401 (application/json)
+    + Body
+    
+            {
+                "status": 401,
+                "error": "Unauthorized",
+                "message": "You have to supply an username and a password.",
+                "loca": "AUTH_NEED_USERNAME_PASSWORD"
```
markus2330: Can you give me links to json schema or blueprint explanations where I can lookup what `loca` means?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: API Blueprints need an empty line between +Body and the example output, for example. So in order to have a more general layout, I decided to use 3 line breaks in front of each first level heading, 2 line breaks in front of each second heading, and so on.

JWT is what i decided to use, because it is very convenient to use in combination with SPA-frontends. And it actually doesn't make a difference for the user, he just has to know how to submit it.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: It is HTML and almost only plain text. I'm not sure if it makes sense to give an example here, as the whole API is described in the doc anyway?

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Longer name? And you mean internationalization and not localization? So maybe use `i18n`?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Yes, the passwords are transmitted in plain text. MITM-attacks are possible also with hashed passwords, so I don't think it makes a difference for transmission. They are stored as sha256 though.

Namoshek:
> What was the output of drafter (warnings?) and is the HTML docu available somewhere?

Drafter ends with **OK**. I uploaded the current HTML version to [my server](http://namoshek.at/api-doc.html).

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Will definitely change it to i18n, thanks for the suggestion.

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+                        "description": "HTTP status code",
+                        "type": "integer",
+                        "minimum": 0
+                    },
+                    "username": {
+                        "description": "The authenticated users name",
+                        "type": "string"
+                    },
+                    "rank": {
+                        "description": "The authenticated users rank",
```
markus2330: rank?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+            {
+                "type": "object",
+                "properties": {
+                    "status": {
+                        "description": "HTTP status code",
+                        "type": "integer",
+                        "minimum": 0
+                    },
+                    "error": {
+                        "description": "Textual description of HTTP status code",
```
markus2330: Can you describe why we send a textual description of HTTP status code, but have localization within the client?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+
+## Register [POST /register]
+
+This method can be used to register for the service. An user account will grant access to more functionality.
+
++ Request Register via form (application/x-www-form-urlencoded)
+    + Attributes
+        + username: the_username (string, required)
+        
+            The username must be between 3 and 20 signs long and may contain only letters a-z (lower- and upper-case), the numbers 0-9 and dots (.) as well as dashes (-).
```
markus2330: are users with different case and or - _ different?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+            
+        + password: somePassword (string, required) - A non-empty password
```
markus2330: No requirements?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+                "additionalProperties": false
+            }
+        
++ Response 400 (application/json)
+    + Body
+
+            {
+                "status":400,
+                "error":"Bad Request",
+                "message":"You have to supply an username, a password and an email.",
```
markus2330: Please create more specific error messages what went wrong.

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,410 @@
...
+            }
+
+
++ Response 400 (application/json)
+    + Body
+
+            {
+                "status":400,
+                "error":"Bad Request",
+                "message":"An username and a rank has to be given.",
```
markus2330: Again, we should distinguish other errors such as not allowed to modify rank.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: With some challenge/salt there should be a difference. @petermax2 Do you have some experience here?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Actually we could ommit the textual status code description completely, as every browser does translate codes anyway. I just thought it could be useful when using command line tools like cURL.

The localized status is always something more detailed, like a hint about missing credentials or wrong format or something like that. When the API is final, I would create a document containing a mapping for i18n key and meaning.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Very good question to be honest. If Elektra is case-sensitive, then yes, otherwise no. Dash and underscore are definitely not the same. Lookup basically runs the input through a regex to ensure safety of the input, then adds it to the base key for the user repository. So it really only depends on Elektra.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Currently not. I always find it very difficult to establish useful rules for passwords, because in many occasions I'm annoyed by too restrictive rules. But I'm open for suggestions!

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: There are way more specific error messages, this is only the most basic one. I wasn't sure if I should list all possible messages in the API description or not. And if yes, in what format / where. Suggestions?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: The rank is used for a very basic permission system. It allows to set a rank between 1 and 3, whereas 1 is a user, 2 something like a moderator and 3 an admin. Higher ranks will allow to delete and modify content of other users as well.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Yes, you are right. Some text for curl usage would be nice. Why not use english text as error and translate that text (without error codes)?

> hint about missing credentials or wrong format

It should not only be a hint, but when e.g. the password is too short there should be an error telling exactly this, not a list of possible problems.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: I would simply link (or show) something like https://xkcd.com/936/ and enforce some minimum length of 10. There should be no maximum length (or something like 4k if we are worried about DoS attacks.)

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: > Why not use english text as error and translate that text (without error codes)?

Because this creates limitations for other software. For example is the frontend loca plugin I use not capable of translating strings with spaces in it. It is also more likely that the spelling/wording of a detailed error message changes than a simple i18n key.

> It should not only be a hint, but when e.g. the password is too short there should be an error telling exactly this, not a list of possible problems.

There is a huge trade-off coming with this requirement. I thought a lot about it when writing the code to be honest. 
Basically there are two options: 
- Check every single constraint on its own (which would lead to many many checks for some inputs).
- Or put some checks together and see if it makes sense. 

For now I've chosen the second approach, because I don't really see where it could lead to problems. If the message says that three fields are required and the user only submitted two of them, it should not be a big deal to find out what the problem is. Other error messages state very exactly what the problem is, for example this one

> The given username is not valid. The username must be between 3 and 20 signs long and may contain only letters a-z (lower- and upper-case), the numbers 0-9 and dots (.) as well as dashes (-).

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: We are currently talking about the API, so images are not really a thing?!

Personally I find a length of 10 already very restrictive. Common constraints are length of 6-8, one lower and one upper case letter. Maybe I use this one.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: When you use the rename plugin you can make Elektra case-insensitive, otherwise it is case-sensitive.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Yes, I only pointed to the image to argue for > 10. 6-8 only makes sense if you use special characters and so on, otherwise it is really weak (Even with umlauts/special characters it is not strong and would depend that we limit the guessing rate).

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: As already said, this is the REST API, not the frontend. Of course I'll add some live-validation for the frontend. That's also a reason why I not want to waste too much time writing validation for each single constraint in the backend. It is also a lot easier to enforce a constraint by a simple regex at once in my opinion, like I did it with above mentioned username validation. So providing more info is not only more work, but also a lot less elegant...

doc/rest_api/snippet_sharing/api-description.apib
markus2330: I expected at least one per status code.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Maybe you can write some "implementation notes" for front ends?

Nevertheless, a correct status number and a single line of english text describing the problem should also be part of the API (for cURL usage)

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Thank you, can you add this to the API description, too? (Maybe as enum?)

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Ok, this will require me to add Unauthorized for all requests. But the actual errors like validation errors will all trigger Bad Request, so I would basically have to list 5-10 almost identical responses for this one. A good way to handle this is the biggest challenge currently for me.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: I'll add implementation notes on my todo list!

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Do you mean challenge because of the limitations in the blueprint specification or implementation challenge. Maybe we are talking past each other. I mean for example (in "pseudo API"):

```
> curl url/password=abc
< no user given
> curl url/user=abc,password=abc
< password too short
```

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Ok, but isn't JWT the cause of the dependency problems? Did you look into the build-in auth from cppcms? Does JWT work with `--cookie` from curl? Maybe we open a seperate issue for authentification?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: It's only about blueprint specification. It would be nice if I could list all possible error cases there, but adding a whole json response for each case is a bit much in my opinion (especially for the other API part later). Implementation is not a big deal.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: The JWT can easily be added to the Authorization header via curl, I did that before. I also added functionality so you can use it as additional GET parameter (?token=the_session_cookie) in case a separate header is not wanted.

Hence I don't really see it as dependency problem at all. I'm using the Boost library for other things as well (e.g. string splitting) and the Cryptlite "library" is only a header collection providing a more natural interface for cryptographic tasks.

But no, I didn't really look for CppCMS authentication stuff, but I also don't expect it to have some. Documentation doesn't give hints about it at all.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: > but adding a whole json response for each case is a bit much in my opinion

Yes, it is. Maybe there is possibility to specify a regex/array of possible answers or similar?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Changed in the upcoming version, ty for the catch!

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: I'll go with 8 characters and require at least one lower- + upper-case letter. I also encourage the user to use special characters as well.

doc/rest_api/snippet_sharing/api-description.apib
petermax2: @markus2330 is right. Basic plaintext authentication is not a good idea!

Please have a look at [RFC2617](https://tools.ietf.org/html/rfc2617#page-6) Section 3 "Digest Access Authentication Scheme".

Namoshek: The proposed method does not offer protection during establishment of passwords (i.e. registration of accounts) and as far as I understand, it assumes that the server knows who a certain resource belongs to. But in my scenario this is not the case, as some "resources" are just methods that several persons can access (e.g. the setrank action is open for all admins).

I'm not sure if challenge/response systems are ideal to be used with HTTP anyway...?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: I updated the docu on [my server](http://namoshek.at/api-doc.html). It does now contain much more detailed information for each field and a list of possible error codes (loca keys) for each possible response with a small description. As soon as I added all missing methods, I'll push an updated raw version.

doc/rest_api/snippet_sharing/api-description.apib
petermax2: > The proposed method does not offer protection during establishment of passwords

In this case I can only think of using HTTPS instead.

> I'm not sure if challenge/response systems are ideal to be used with HTTP anyway...?

It is certainly not ideal. HTTPS would be better. But challenge/response offers better protection of the password itself than plaintext or hashing. It does not make the authentication process more secure but many users tend to re-use their passwords across multiple service. So every countermeasure against password sniffing is important.

Namoshek: > In this case I can only think of using HTTPS instead.

This is also what I expect to be the most promising protection, but it doesn't really impact my work I think - it is only a configuration task later on.

> But challenge/response offers better protection of the password itself than plaintext or hashing.

I totally agree, but implementing a challenge/response system within HTTP is kinda akward. It may work quite well if you have an API and a frontend that takes care of, but when using the API by hand (curl), I can't imagine how it should work. My approach was also to not store states on the server (that's why I'm using stateless JWTs), because it makes development a lot more complex and error prone.

Namoshek:
I pushed a new version of the API description with a lot more details and better syntax. I also updated the rendered version on [my server](http://namoshek.at/api-doc.html). As soon as the API works, we could add the host to the blueprint, which would add console support to the rendered documentation. Maybe it would also be a good idea to return the docu instead of the current plain html API description when accessing the API root?

doc/rest_api/snippet_sharing/api-description.apib
petermax2: curl supports the digest authentication, just append `--anyauth` or `--digest`, see `man curl` or [here](https://curl.haxx.se/docs/manpage.html#--digest) .

This authentication scheme is not something I made up. RFC2617 is widely accepted and implemented by a wide range of tools.

Namoshek: If curl supports it, then maybe it is worth looking deeper into it, thanks for now.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: > dependency problem Cryptlite

Yes, the problem is also not really "ours", but its a problem for every distribution. E.g. in Debian maintainers need to write a licence file where every copied source file must be described. Thus we should avoid distributing source files of other licences/authors. Furthermore it creates huge problems if there are security bugs. cppcms provides [crypto](http://cppcms.com/cppcms_ref/latest/namespacecppcms_1_1crypto.html) and also [much more](http://cppcms.com/wikipp/en/page/secure_programming).

Boost is ok if you need more than provided by [booster](http://cppcms.com/cppcms_ref/latest/namespacebooster.html and especially "booster::locale", the author of cppcms is also author of boost::locale, where a full copy is included in booster). The only reason against boost is the bloat of the binary size, because cppcms already uses an internal version of boost and if we use one more version the binary size grows again. But you can keep it, if you do not find an  booster replacement in every case. (One problem of boost is that they do not really care about compatibility, please avoid these parts)

doc/rest_api/snippet_sharing/api-description.apib
markus2330: The API should be usable by programs, thus you should somehow specify how a program can get all available API methods (e.g. to check if what it needs is already available). If you return some unspecified plain text or html, its not really useful as API.

markus2330:
Generated API looks nice (it seems to be located on apiary.io though)

> Maybe it would also be a good idea to return the docu instead

A CSV like format (or JSON if you prefer it) with links to docu and versions for every method would be another way? Returning the whole docu seems to be hardly usable, or do you mean such a Apiary.Embed like it is done on your web server currently?

@omnidan Can you also take a look at the API?

markus2330:
Two small questions:

- What are the scopes for? (Database)
- Why is DELETE only possible for admin? Why should a user not be able to remove what he/she uploaded before?

Namoshek:
> What are the scopes for? (Database)

Basically the idea was that it is easier searchable if there is some very crudely grouping involved in the highest hierarchy level, but with tagging (which idea came a bit later) it is probably not as useful anymore. Do you think I should remove this level?

> Why is DELETE only possible for admin? Why should a user not be able to remove what he/she uploaded before?

It is only a docu mistake, of course the user should be able to delete own stuff. Thanks for the catch! :+1:

markus2330:
> Where have you seen anything about Apiary in the generated docs?

My browser tells me if content from other domains is loaded.

>  generated doc is more helpful?

Yes, if its similar to what you have on your webpage, it sounds useful. But then it is not really API anymore. There should be also some programmatic way to get some information, at least the version of the API.

> Do you think I should remove this level?

This is a difficult question and depends on how it is solved in the web ui. Basically it is a question of user experience: If it is easy to use the scopes, and it is helpful for retrieving snippets, we should have it. If users are annoyed by it and only use tags anyway, it is better if we do not have it. Maybe the following question helps you for deciding: Will it be clear for a user when to use scopes and when to use tags?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: For crypto it is required to compile CppCMS with more dependencies (e.g. OpenSSL), so it would only be a shift of dependencies in my eyes. There is also no documentation on how to use it anywhere (yes I looked this up before).

The main problem I have with getting rid of the Cryptlite lib is that using cryptographic functions requires lots of initialization and stuff that is done by this library. So I would basically have to write it on my own again, achieving nothing else than work. Why should I do that? License problems of other people are not my business, to be honest.

Namoshek:
> My browser tells me if content from other domains is loaded.

Interesting. Well, it is generated by the Apiary-CLI, so it is quite likely that it is the same sort of documentation than they use for their service. There are other tools available too, I can try some of them as well.

> Yes, if its similar to what you have on your webpage, it sounds useful. But then it is not really API anymore. There should be also some programmatic way to get some information, at least the version of the API.

To @omnidan you said version for the API is not really that important...?!

> This is a difficult question and depends on how it is solved in the web ui. Basically it is a question of user experience: If it is easy to use the scopes, and it is helpful for retrieving snippets, we should have it. If users are annoyed by it and only use tags anyway, it is better if we do not have it. Maybe the following question helps you for deciding: Will it be clear for a user when to use scopes and when to use tags?

I guess yes, but without a good set of possible scopes it will be hard to find the right one. So I will get rid of it.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: > For crypto it is required to compile CppCMS with more dependencies (e.g. OpenSSL), so it would only be a shift of dependencies in my eyes. There is also no documentation on how to use it anywhere (yes I looked this up before).

Cppcms on debian is already compiled with libgcrypt11, are there any problems with how it is compiled? It is not only a shift of dependencies. When we use libgcrypt, security bugs are libgcrypt's problems, if we distribute source, every problem is our problem. And cppcms even seems to have different security backends (similar to what @petermax2 made for Elektra's crypto plugin), i.e. if libgcrypt support end, we can switch easily.

And what do you mean with no docu?
- Tutorial: http://cppcms.com/wikipp/en/page/cppcms_1x_sessions
- Config: http://cppcms.com/wikipp/en/page/cppcms_1x_config#session

Namoshek: I read on sessions in CppCMS before, but they are cookie based only. I've never seen cookie sessions been used for any other API before, so I doubt this is good design..? Of course it would be more comfortable because it is there already, but there must be a reason why APIs use JWTs, oAuth and other mechanism in favour of cookies.

I understand your problem with the dependencies and I'll see what I can do about it, but I'm not promising anything. I was talking about crypto docu by the way, not sessions. I'm using crypto also for other taks (e.g. password hashing). The only docu I could find is the generated doxygen docu that doesn't give much info at all.

Namoshek:
>>My browser tells me if content from other domains is loaded.

>Interesting. Well, it is generated by the Apiary-CLI, so it is quite likely that it is the same sort of documentation than they use for their service. There are other tools available too, I can try some of them as well.

I tried the aglio doc generator which offers two different templates, both can be found on my server:
- [Default template](http://namoshek.at/api-doc-aglio-default.html)
- [Triple-Column Template](http://namoshek.at/api-doc-aglio-triple.html)

It works quite well too and the layout seems nice as well, but one or two things work different (e.g. many `{?optional}` parameters are not converted into `?param1&param2&param3`, but rather in to `?param1?param2?param3`. But if we make a final decision on which generated doc to use, this can be adjusted.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Maybe you can find out why they prefer other methods, just because others are doing it is not a good reason.

> I'll see what I can do about it

Thank you.

omnidan:
> So far I was unable to find a way to put schemata into a variable or something similar, so I copied it for each response. But I really don't like it doing this way. Maybe @omnidan has an idea?

I haven't done this either, but the [api-blueprint examples](https://github.com/apiaryio/api-blueprint/tree/master/examples) have helped me a lot when writing blueprints before. Maybe you can find something similar there?

> If it is easy to use the scopes, and it is helpful for retrieving snippets, we should have it. If users are annoyed by it and only use tags anyway, it is better if we do not have it.

I think it will be hard to define scopes for us if we want to make them predefined. If they're user defined, I would definitely go for tags instead (much easier to maintain, more flexible, most snippets don't have one category, e.g. `apache`, `php`, `wordpress`)

> @omnidan Can you also take a look at the API?

I just had a look at the API, here are my comments:

- [ ] Authentication endpoint could be a `/user` resource (more RESTful)
  - [ ] especially `setrank` is a little weird for a REST API (could be `PUT /user/:id` to edit a certain user or `PUT /user` to edit your own account)
  - [ ] `/auth` is fine, because it's special (and `/user/auth` might conflict with `/user/:id`), `/register` could simply be `POST /user`, though
- [ ] why does the HTTP status code need to also be part of the response body? (redundancy?)
- [ ] Database endpoint is basically the Snippets endpoint? (where you get snippets from other users) Maybe a different name would be more appropriate here?

Namoshek:
> I haven't done this either, but the api-blueprint examples have helped me a lot when writing blueprints before. Maybe you can find something similar there?

Of course I found and used them already. It is possible to replace some code (e.g. the attributes of responses) by a model, but for the different i18n descriptions I'd have to define this one attribute every time either way. And if I replace the other 3 lines for example, the i18n is afterwards always on top, because the substitution for the model is appended to the list instead of prepended... that doesn't look nice. So for now I kept it quite c&p like. Most important would be a solution to only store schemata once, because they are freaking large, but this seems impossible.

> I think it will be hard to define scopes for us if we want to make them predefined. If they're user defined, I would definitely go for tags instead...

Thanks for your opinion, I'll go with tags instead!

> - Authentication endpoint could be a /user resource (more RESTful)
 - especially setrank is a little weird for a REST API (could be PUT /user/:id to edit a certain user or PUT /user to edit your own account)
 - /auth is fine, because it's special (and /user/auth might conflict with /user/:id), /register could simply be POST /user, though

Good point there. But concept wise (and also logically) I'm not so sure if it is a good idea to open some operations for the `/user` resource to the public (e.g. POST for registration) and some not (e.g. PUT for setrank/edit). But I could imagine that a mix of both, `/auth` and `/auth/register` like now and the rest like you suggested, could work quite well. Of course I could also create a `POST /user` resource that is only applicable for admins and that is used by `/auth/register` internally...

> why does the HTTP status code need to also be part of the response body? (redundancy?)

That was part of the decision to put a textual description of the status in the response body as well. Theoretically this should be unnecessary, that is right. I also cannot really think of any reason why it should be necessary, all JS (front-end) libs that I know and use can handle headers and the response codes there. I think I'll remove them if no one can give me a good reason to not do so.

> Database endpoint is basically the Snippets endpoint? (where you get snippets from other users) Maybe a different name would be more appropriate here?

Suggestion? :smile: 
Internally I'm often using the word `repository`, which seems equally in-/appropriate to me. `snippets` doesn't really sound that well either to be honest. It is also not really the name of a resource, but rather a prefix to avoid name collission (see `/auth` and `/user`). Therefore the naming is not too important in my eyes, but of course I'm open for suggestions!

Namoshek:
Quite some changes done:
- Removed schemata descriptions because they are automatically generated (at least by apiary)
- Moved some functionality of the `/auth` endpoint to a new `/users` endpoint as suggested (also `/auth/register`, after thinking about it again it makes quite some sense)
- Added `GET /users` to retrieve a list of all users (for moderators and above)
- Removed the `status` because of redundancy
- Changed the name of the `errors` field which has the textual description of the HTTP status to `status`
- Some styling changes (e.g. empty lines for formatting in the generated doc)

Not done yet:
- Change the name of the `/database` endpoint because of no good ideas so far. But in case this changes, it should not be much work to change it in code (my design only uses the `/database` word once).

Let me know what you think about the current state @omnidan @markus2330. If you can't think of any severe issues anymore, I'd like to incorporate the changes (maybe from monday on).

markus2330:
> To @omnidan you said version for the API is not really that important...?!

Sorry for the confusion, in @omnidan  case I was referring to elektrad's version (the software implementing the API). API versioning is useful and should be done, but it might be best to simply return Elektra's version (at least as long as the API+Elektra live in the same repo and are released together).

> Therefore the naming is not too important in my eyes

Its only not too important if the name is not "wrong" and used consistently. If you use database at one place, but repository at another, it calls for trouble. `kdb` would be another option.

> but one or two things work different [in templates]

Then lets stay at the default template (unless you know that they want to change the default to the triple-column soon).

> `users` endpoint

@omnidan suggested `user`, so it seems like you have different opinions about singular/plural. I think @omnidan version is preferable, because the current Elektra API also avoids plural (except KEY_FLAGS, I think we should fix this for 1.0). Also see doc/DESIGN.md

> I'd like to incorporate the changes into my code

Would be good if @omnidan API is also "final" so that we know that both APIs are consistent.

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+## List API methods [GET]
+
+Returns an overview of available API methods with required parameters and a description.
+
++ Response 200 (text/html; charset=utf-8)
+
+
+## API Version [GET /version]
+
+Returns the currently used version of the API. A version consists of two numbers (format: a.b), where the first number (a) is the major version and the second number (b) the minor version. Minor versions may not contain breaking changes (i.e. backwards-compatibility is ensured), but major versions may. This means that minor versions may only extend the API.
```
markus2330: There are 3 numbers, see doc/VERSION.md. I do not think you need to repeat the explanation of what the numbers mean, can you link to the file?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+            - `AUTH_MISSING_PASSWORD` if no input for `password` was submitted
+
++ Response 401 (application/json)
+    + Attributes (object)
+        + status (string) - Textual description of the HTTP status
+        + message (string) - Detailed error information, e.g. hint about malformed request
+        + i18n (string) - A unique token representing above error description message; can be used for internationalization in frontends
+
+            A comprehensive list of possible errors:
+            - `AUTH_UNKNOWN_USERNAME` if there is no user registered witht he given name
```
markus2330: typo: witht he

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+
+
+# Users endpoint [/users{?rows}{?sort}{?sortby}{?filter}{?filterby}{?current}]
+
+## User list [GET]
+
+Retrieves a list of users. The result can be adjusted by additional, but optional parameters. This operation requires moderator permissions (rank 2) or higher.
+
++ Parameters
+    + rows (number, optional) - How many entries will be added to the output at max.
```
markus2330: prefer singular. Is this option without some skip (to go to the next page of users) useful?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+        + Default: `asc`
+        + Members
+            + `asc`
+            + `desc`
+    + sortby (enum[string], optional) - What criteria should be used for sorting
+        + Default: `username`
+        + Members
+            + `username` - The users name
+            + `email` - email
+            + `created_at` - Creation date
```
markus2330: Format of date? Is it needed?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+    + name (string, required) - The `username` of the user whose details should be fetched
+        
++ Response 200 (application/json)
+    + Attributes (object)
+        + username (string) - The name of the user
+        + email (string) - The email of the user
+        + rank (enum[number]) - The rank of the user (higher means more permissions)
+            + Members
+                + `1` - Normal user
+                + `2` - Moderator
```
markus2330: What is the moderator for? Can you leave some "space" between the numbers (e.g. 3..normal user, 9.. admin), so that we can come up with extensions (e.g. if we want to block a user from uploading, we give the rank 1..only downloads).

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+        + message (string) - Detailed error information, e.g. hint about malformed request
+        + i18n (string) - A unique token representing above error description message; can be used for internationalization in frontends
+
+            A comprehensive list of possible errors:
+            - `NEED_AUTHENTICATION` if no valid session token has been supplied or the user for whom the token has been created does not exist (anymore)
+        
+
+## Create user [POST]
+
+This method can be used to register for the service by creating a new user account. An user account will grant access to more functions.
```
markus2330: Its "a" user (because you say /j/ser)

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
++ Parameters
+    + name (string, optional) - The `username` of an user which is then taken as target for the operation
+
++ Request (application/json)
+    + Attributes (object)
+        + email: `a-new@email.com` (string, optional)
+            
+            The new email address which should be used for the user. It has to be a valid email.
+
+        + password: `a_new-secureP4ssw0rd` (string, optional)
```
markus2330: please use some random string such as "waePh0IuPohhi2HiEeShee8iohyi9Ai", we should not recommend to use such weak passwords. And suggest that the password should be long and different to the previous one.

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+        + rank: `3` (enum[number], optional)
+        
+            The new rank which should be set for the user. Setting this parameter does only result in a change if the currently authenticated user has the permission to change ranks.
+
+            Regex: ^[1-3]$
+        
+            + Members
+                + `1` - Normal user
+                + `2` - Moderator
+                + `3` - Admin
```
markus2330: Can you avoid this duplication?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+
+
+# Database endpoint [/database{?offset}{?rows}{?sort}{?sortby}{?filter}{?filterby}{?raw}]
+
+## List database entries [GET /database/{organization}/{application}/{scope}{?offset}{?rows}{?sort}{?sortby}{?filter}{?filterby}]
+
+Returns a list of database entries. The result can be adjusted by additional, but optional parameters.
+
++ Parameters
+    + organization (string, optional) - Only search within a given `organization` for entries
```
markus2330: Isn't this ambiguous? If I only have two parameters, is it org+app, or app+scope? Is org relevant?

Namoshek:
> API versioning is useful and should be done, but it might be best to simply return Elektra's version (at least as long as the API+Elektra live in the same repo and are released together).

If we have a version resource, it should represent the API version in my eyes. It is ok to have more information (additionally also the Elektra version), but the API of the REST service could change. I made a small proposal for this already, but if you like I'd add the Elektra version as well (although I don't really see much benefit, to be honest).

> Its only not too important if the name is not "wrong" and used consistently. If you use database at one place, but repository at another, it calls for trouble. kdb would be another option.

I'm afraid `kdb` would be confusing as the API does **not** provide a direct access to the whole key database, but only to a sub-space due to security.

> Then lets stay at the default template (unless you know that they want to change the default to the triple-column soon).

I wouldn't change to the other templates as they cut away a lot of information that the apiary-cli provides and generates...

> @omnidan suggested user, so it seems like you have different opinions about singular/plural. I think @omnidan version is preferable, because the current Elektra API also avoids plural (except KEY_FLAGS, I think we should fix this for 1.0). Also see doc/DESIGN.md

I'm fine with singular. It is an often discussed topic anyway. I just found users somewhat more appropriate for the POST operation, but I'm totally fine changing it to singular.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: As stated below, this is not the Elektra version, but an own version number for the API. But your idea of using the Elektra version could work as well if same rules apply. The question is what to do if Elektra itself is not planning to bring out a bigger new version, but the API needs a breaking change?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+
+## Get details for entry [GET /database/{organization}/{application}/{scope}/{slug}{?raw}]
+
+Returns detailed information about the entry with the specified key (path). If the optional parameter `raw` is supplied, a raw configuration is returned instead of the normal response.
+
++ Parameters
+    + organization (string, required) - The `organization` who built the `application` for which the configuration of the entry is meant
+    + application (string, required) - The `application` to which the configuration of the entry belongs to
+    + scope (string, required) - The `application` specific `scope` to which the configuration of the entry belongs to
+    + slug (string, required) - The unique identifier of the requested entry
```
markus2330: If slug is already unique, why do we need org, app and scope then?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: There should definitely be an offset, maybe I accidentally deleted it during c&p from the database resource. Thanks for the catch!

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+                + application (string) - The application to which the snippet belongs to
+                + scope (string) - The application internal scope for the configuration snippet
+                + slug (string) - The unique identification slug of the entry
+            + meta (object) - Additional information about the entry
+                + title (string) - The descriptive title of the entry
+                + description (string) - The extended description of the entry
+                + author (string) - The name of the creator of the entry
+                + created_at (number) - A timestamp of the creation date of the entry
+            + value (array[object])
+                + format (string) - A configuration format, e.g. **ini**
```
markus2330: How can I change this? What is a raw format?

markus2330: I would suggest to get the format as uploaded, but that I can also retrieve other formats by adding a parameter.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: It is a timestamp in integer format, but here it is only an enum for the parameter, so no further specification necessary. I don't think knowing the registration date is essential, but having it is often useful for statistics and other stuff (e.g. to know who has been on board for a long time and could get some more privilegues...).

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+            - `ENTRY_UNSUPPORTED_FORMAT` if an unsupported `raw` format has been supplied
+
++ Response 404 (application/json)
+    + Attributes (object)
+        + status (string) - Textual description of the HTTP status
+        + message (string) - Detailed error information, e.g. hint about malformed request
+        + i18n (string) - A unique token representing above error description message; can be used for internationalization in frontends
+
+            A comprehensive list of possible errors:
+            - `ENTRY_DOES_NOT_EXIST` if an entry with the given key (path) does not exist in the database
```
markus2330: What happens on database errors?

doc/rest_api/snippet_sharing/api-description.apib
```diff
@@ -0,0 +1,587 @@
...
+    + Attributes (object)
+        + message (string) - Detailed error information, e.g. hint about malformed request
+        + i18n (string) - A unique token representing above error description message; can be used for internationalization in frontends
+
+            A comprehensive list of possible messages:
+            - `ENTRY_CREATED_SUCCESSFULLY` if the entry has been added successfully to the database
+
++ Response 400 (application/json)
+    + Attributes (object)
+        + status (string) - Textual description of the HTTP status
```
markus2330: You should also return the kdb error in an (additional) field.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Interesting idea, but for your example I'd simply create a new account and would be fine again (only solution would be to change the default group to something with less privilegues too). But I'll change the groups to something else, then we don't have to discuss it to the end now:
- 10 - normal user
- 50 - moderator
- 100 - admin

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Thanks, I was always very unsure about it. We only learned to use "an" in front of a-e-i-o-u back in school, but it is only in front of the spoken vokals I guess. Will change.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: I'll try, but I cannot promise anything. I understand your issue though.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: It is a hierarchical database, therefore if you only provide two parameters, it will be `organization` and `application`. If you want to search the whole database, you could search the empty path with a filter instead.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Slug is only required to be unique within the namespace created by org, app and scope. Theoretically the slug could also be created from the snippet title - maybe this would be a good change?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: I had following ideas:
- When you fetch the details for a snippet, you get the snippet in all enabled configuration formats (plugins have to be enabled for use with the service first).
- Instead you can also fetch the snippet in a specific format directly by providing the name of the format. (This could be useful for use with curl and then pipeing it into a file or kdb for example.)

Although it doesn't really impact the API description, would it be better to read out all available storage plugins from kdb directly and use them instead of an individual setting? If I remember correctly, i invented this setting because the storage plugins do not behave consistently when they fail (or should fail) an import (e.g. the xml plugin).

About up- and download:
When creating a snippet, the submitted configuration snippet is run through all enabled plugins until one succeeds with the import (convertion into normal Elektra keys). It is then stored wherever the repository has been specified and in whatever format this file is mounted. 
Another solution would be to require the user to specify the format during upload and only use this plugin. Usability wise this seems not as good as the first approach to me, but definitely less error prone.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: I thought about throwing 500 errors (and internally I do s), but I doubt that API descriptions should contain 5xx errors, because they are normally not directly related to the API. What do you think?

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Is this information even necessary for the user? In my eyes he only cares about whether it fails or not, because if it is an internal error he cannot influence, then the information is of no value for him...

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Ideally we should avoid breaking changes altogether, let us discuss it in the next meeting.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: 100 different levels should be enough for a single dimension of permissions ;)

Ok, then let us discuss details of permissions later..

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Many errors are useful for the user, e.g. validation errors. And some errors would be useful for us (to debug).

doc/rest_api/snippet_sharing/api-description.apib
markus2330: What does blueprint say about this? Are there any reasons for not specifying it?

doc/rest_api/snippet_sharing/api-description.apib
markus2330: > you get the snippet in all enabled configuration formats

Is this needed? What is the separator between snippets? What happens if some formats fail?

> Instead you can also fetch the snippet in a specific format directly by providing the name of the format.

sounds great, is it possible to give a full example how this can be used (and not only "e.g. ini")

> run through all enabled plugins

The problem is we have some plugins that succeed with anything you give to them (e.g. line, ini). We would need proper auto-detection, e.g. the plugins give a score how likely it is that they should be used for a task. It will always be a somewhat messy heuristic but I have to admit that some kind of auto-detection is very useful (also for mounting).

doc/rest_api/snippet_sharing/api-description.apib
markus2330: The problem is that org+app is not always what you want, sometimes you want to search snippets from an application by any org.

What about having different views of the same database?

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Is this specified by blueprint? Otherwise you should write it as part of the specification.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: > Is this needed? What is the separator between snippets? What happens if some formats fail?

Is it needed... definitely not, but still a nice to have in my opinion. Impact of it could be quite dramatically with lots of plugins enabled though, that's worth a discussion.
Each format of the snippet has its own object in an array of the response, consisting of format name and the snippet itself (plain string, escaping is all done by the json libs). As the snippet is stored in the key database as if it was a normal configuration, it is very unlikely that it would fail the conversion into another format (in my opninion this would mean the storage plugin in question is broken?!). This already worked in an earlier version of my service by the way.

> sounds great, is it possible to give a full example how this can be used (and not only "e.g. ini")

In the API description?

> The problem is we have some plugins that succeed with anything you give to them (e.g. line, ini). We would need proper auto-detection, e.g. the plugins give a score how likely it is that they should be used for a task. It will always be a somewhat messy heuristic but I have to admit that some kind of auto-detection is very useful (also for mounting).

That's the issue. Currently I'm really limited when it comes to plugins that I can use, because some of them just always succeed. Somehow I tend to asking the user what format his snippet is in, simply to avoid problems when trying the import for all plugins. But if the users enters a wrong format with a plugin that always succeeds, the same problem would arise...

Another issue could be that some plugins export with an absolute path, but the path from the REST service will make no sense for an import, because it has a special location for the snippets. I've absolutely no idea how to fix that (except by doing a stupid string-replace that will break the import on the other end). But somehow it isn't really my problem / fault, so ...

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Blueprints does not make any suggestion for what to specify when and when not. 5xx internal errors are normally just something you can receive always, no matter what the specification of an API says, because the server itself can have issues too. But because I'm throwing them explicitely, it could make sense again. I'll sleep a night about it.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Have to see how this could work out before making a decision.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: If you want to search for an application from several organization, you can use the filter provided by the optional parameters. Although I got to say that this sounds terribly weird, like "the same application of different organizations".

> /apache/tomcat/common/some-snippet
> /org.apache/tomcat/common/some-other-snippet

You'd find both (and maybe more) by using either of the following requests:

> GET http://the-api.com/database/
> GET http://the-api.com/database/?filter=tomcat
> GET http://the-api.com/database/?filter=tomcat&filterby=key

If you provide a more specific URI (with an organization, an application and maybe even a scope), then this will only lead to a pre-filtering before the other options. If you leave the path empty, you can search all keys. The following should theoretically work as well with my design and also deliver both of above results:

> GET http://the-api.com/database/?filter=apache%2Ftomcat%2F&filterby=key
> ... which is the encoded form of ...
> GET http://the-api.com/database/?filter=apache/tomcat/&filterby=key

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: It is written in the API specification several times what the field represents and how it does that. Above snippet of the api description has nothing to do with the field itself.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: > Is it needed... definitely not, but still a nice to have in my opinion.

I do not see a benefit to requesting individual formats, but if its somehow handy in the frontend code we can keep it. (It might be more useful if it also returns the score how well the snippet matches, see below)

> In the API description?

Maybe an extra tutorial that explains how to use the API with curl?

> Somehow I tend to asking the user what format his snippet is in

The snippet service would work much better if there is some kind of auto-detection with score (give the user an option to overwrite if the auto-detection goes awry). So it would be great if you can come up with some heuristics (e.g. count number of =, #, ; and so on) so that we have an auto-detection that works most of the time.

It should be generic in following sense:

- be available as lib so that others (mounting) can use it, too (add it to src/libs/tools would be an obvious choice, then you can also implement it in C++)
- it should work based on the contracts the plugins export, not on a hardcoded logic that assumes plugins to be constant. We easily can add comment and separation characters to the contracts if it is needed by your heuristics. If multiple plugins match, the best one should be chosen (based on infos/status).
- You do not need to support the augeas plugin, it would be a large effort. But maybe you keep things in a way that supporting it does not mean to rewrite the lib.

> absolute path

No action needed on your end, that is the oldest open issue #51 and is already worked on.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Some examples: export arbitrary key/values as fstab or object/array intermixed as json. There are plugins that check fstab and json if they are valid (e.g. the plugins struct, type and spec). Currently the import/export API unforunately does not add these validation plugins (which might lead to syntactically incorrect json), it would be great if you can add support that (see src/libs/tools/include/backend.hpp ImportExportBackend for a partial implementation of the idea to use mounting logic (which is able to add validations plugins if specified in contract) for import/export).

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: >>>> you get the snippet in all enabled configuration formats

>>>Is this needed? What is the separator between snippets? What happens if some formats fail?

>> Is it needed... definitely not, but still a nice to have in my opinion.

> I do not see a benefit to requesting individual formats

Can't follow you, sorry. From my perspective the API spec is clear in this point and makes sense for both front-ends as well as CLI users. I'll add more explanation to what **raw** means and does, but that's it.

> Maybe an extra tutorial that explains how to use the API with curl?

Something that **could** be done somewhen in future, when the thing is real.

> The snippet service would work much better if there is some kind of auto-detection with score (give the user an option to overwrite if the auto-detection goes awry). So it would be great if you can come up with some heuristics (e.g. count number of =, #, ; and so on) so that we have an auto-detection that works most of the time.

Does sound like the working title of a completely different thesis to me. So long story short: Won't implement.

If there has to be a fallback option to override the format manually, I'd have to implement my currently planned approach additionally eitherway. So I don't see the benefit for me, sorry.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: Why should I need validation for export? If the export of a snippet is wrong, then the plugin is broken (or can't support the depth of the configuration, which also speaks for a broken or wrongly designed plugin).

I won't change anything here until I'm finished with my work. If I then think it would make sense because most of the exports are wrong, I will evaluate the option. But not now.

doc/rest_api/snippet_sharing/api-description.apib
markus2330: I understand you that you do not like the situation that you have already written a full implementation and I start discussing use-cases. Its best if we discuss this matter in the next meeting directly (with documents describing the scope), and not within side-discussions within API reviews. Nevertheless, two things:

1. Please make some curl examples of raw/format thus I am not yet used to the blueprint specification and might misunderstand your intend (especially because there are no written use-cases). Setting/getting configs is the central part of the whole thing, thus we should have the same understanding. It does not need to be a full-blown tutorial now, but it should be clear how you expect a user to upload/download snippets via curl.
2. Trying one plugin after the other already is a valid implementation for guessing which format it is. So even if you are not interested in providing better heuristics (which I think might be a usability problem and thus might concern your thesis), you still should separate the guessing into the library so that we have the place to start implementing better heuristics (independently of who will do it).

doc/rest_api/snippet_sharing/api-description.apib
markus2330: Same as above and please let us focus on the best API, not on implementation issues or involved efforts.

It is not about broken plugins, but about the fundamental problem that not every KeySet can be mapped to every syntax. Validation is useful to not produce syntactically invalid files. We can discuss the problem in the next meeting.

doc/rest_api/snippet_sharing/api-description.apib
Namoshek: The way I design my API also depends on these decisions, so I cannot always discuss it later. Every additional discussion takes time from me that I'm not ready to invest anymore. You know what goal I have in mind and I already see it vanishing - again.

You provided me with a title for my thesis, but not with concrete ideas or demands what the whole thing should be capable of. A title gives latitude to make own decisions and concepts. If we are now changing everything after I already made my plans, I'm not ok with it. Especially not if it takes dozens of iterations until you are satisfied or run out of ideas for new stuff. It prevents me from going on in the milestone list, I cannot implement stuff if it is likely to change again. I'm not willing to do all work two or even more times. Especially not in a programming language that I'm not using in daily business.

It is also hard for me to plan everything to the end beforehand, without writing a single line of code, as I don't know the whole library and all of its tools yet. I certainly will build my decisions upon capability and feasibility of the tools I've on my hand, even if this requires to make compromises concerning erstwhile plans and imaginations. Writing a second thesis about the recognition of configuration formats is not within the scope of my work and will never be.this format (and no meta information). It basically pipes the output of the storage export to the HTTP response. This way you could simply pipe the output of the HTTP call to a config file.

---

I made the decision to remove the guessing of the format completely as the current plugins are not stable enough to do it for me. Instead I'll offer to choose the format during upload. This will probably require a method to retrieve supported formats, but I'm fine with this. I don't want to offer override methods for this because of bad usability (everyone will first try the auto-detection only to be disappointed afterwards). 

---

As I'm no curl guru and I don't want to read up on this right now, I'll just add my existing use-cases to the PR. 

---

Finally, Blueprints are nice because they are easy to read. I understand that some things might seem weird, but the two operations in question are really easy to understand.

[The uploading of snippets](https://github.com/Namoshek/libelektra/blob/1d2c4b92405adf6e7555c2b751252f59bccac43b/doc/rest_api/snippet_sharing/api-description.apib#L370) is pretty straight forward. Within a JSON object you supply some meta information and a configuration as plain string (if this comes from curl, a textbox or a file doesn't matter at all - it is a plain string). We are then - somehow - finding the right storage plugin to use for the snippet, convert the whole snippet into a KeySet and store it with KDB. No magic happens here.

[The retrieval of snippets](https://github.com/Namoshek/libelektra/blob/1d2c4b92405adf6e7555c2b751252f59bccac43b/doc/rest_api/snippet_sharing/api-description.apib#L320) is even easier.
When the path to a snippet has been specified, a normal JSON response (the first 200 response in the list) will be returned, with the snippet already converted into all formats that the service supports.
If the additional `raw` parameter has been supplied (and contains a supported format), the converted snippet will be returned as plain text (the second 200 response). This response can then be piped into files or whatever suits the user best.

markus2330: > A title gives latitude to make own decisions and concepts.

Yes, own decisions and ideas are very welcomed, but if you you can also always ask me.

> If we are now changing everything after I already made my plans, I'm not ok with it.

The workflow is to finish the expose, use cases and some architecture/technical descriptions before anything else. I communicated this several times, and in your case it was never finished. The implementation (of the main parts) should not start before.

But do not worry, I am also interested to find a good solution for you, but we should discuss this in person. Detection of format or curl-tutorials are certainly not core parts, let us talk about this in person when you have the necessary documents. Please write me an e-mail for date&further questions.

doc/rest_api/snippet_sharing/use_cases/UC_create_snippet.txt
```diff
@@ -0,0 +1,17 @@
...
+Scope: Entry Management
+Level: User Goal
+Actors: Authenticated User
+Brief: User can create a new database entry based on a set of specified inputs, one of them being a configuration snippet.
+
+Scenarios:
+----------------------------
+Precondition: Authenticated
+Main success scenario: All inputs validate and the configuration snippet could be stored in the Elektra database successfully. The user retrieves a success response.
+Alternative scenario: Wrong inputs lead to an error message and ask the user to correct them, then he can try again. It is also possible that the supplied configuration snippet is of an unsupported format. In either case the user retrieves a response containing error information.
```
markus2330: he -> he/she or they

doc/rest_api/snippet_sharing/use_cases/UC_edit_user.txt
```diff
@@ -0,0 +1,17 @@
...
+Actors: Authenticated user, Administrator
+Brief: Users can edit their own account information the same way Administrators can do it for them. Administrators can also change permissions of users.
+
+Scenarios:
+----------------------------
+Precondition: Authenticated, sufficient permissions
+Main success scenario: Changes to the user account have been stored to the database successfully. The operating user is informed about the success through a response from the server.
+Alternative scenario: The submitted inputs could not be validated or the user entry for which changes are requested cannot be found. The server responds with an error message.
+Error scenario: Technical problems prevent the storage process from completing. The operating user is informed about the issue through an error response.
+Postcondition: Changes to the user account are stored in the database successfully and are from now on valid for further requests (e.g. authentication attempts).
```
markus2330: Are they valid immediately, or only after relogin?

doc/rest_api/snippet_sharing/use_cases/UC_edit_user.txt
```diff
@@ -0,0 +1,17 @@
+Summary:
+----------------------------
+Title: edit user
+Scope: User Management
+Level: User Goal
+Actors: Authenticated user, Administrator
+Brief: Users can edit their own account information the same way Administrators can do it for them. Administrators can also change permissions of users.
```
markus2330: "the same way" suggests that the other way was explained somewhere else. Please link to use case or change the sentence.

Do you also plan some automatical scoring for uploading many popular snippets?

doc/rest_api/snippet_sharing/use_cases/UC_edit_snippet.txt
```diff
@@ -0,0 +1,17 @@
+Summary:
+----------------------------
+Title: edit configuration snippet
+Scope: Entry Management
+Level: User Goal
+Actors: Authenticated user, Moderator
+Brief: Users can edit their own database entries the same way that moderators can edit any configuration snippet in the database.
```
markus2330: snippet -> snippets? "the same way" (see below)

doc/rest_api/snippet_sharing/use_cases/UC_details_snippet.txt
```diff
@@ -0,0 +1,17 @@
...
+Title: viel details for specific configuration snippet
+Scope: Detailed View
+Level: User Goal
+Actors: Anonymous User (+ all others)
+Brief: Any user can view all details about an entry (configuration snippet).
+
+Scenarios:
+----------------------------
+Preconditions: URI of a snippet entry is known
+Main success scenario: User retrieves a response containing all information about the entry, including tags and other meta information like a description, as well as the configuration snippet converted into all supported formats.
```
markus2330: Which "other meta information"?

doc/rest_api/snippet_sharing/use_cases/UC_delete_snippet.txt
```diff
@@ -0,0 +1,17 @@
...
+Actors: Authenticated user, Moderator
+Brief: Users can delete their own database entries the same way that moderators can delete any configuration snippet in the database.
+
+Scenarios:
+----------------------------
+Precondition: Authenticated, sufficient permissions
+Main success scenario: Configuration snippet has been deleted successfully. A success response is returned.
+Alternative scenario: The specified snippet cannot be found (wrong URI) or the user has insufficient permissions to delete it. Either way the user is informed about the error by a response.
+Error scenario: Technical problems prevent the deletion process from completing. The user is informed by an error response.
+Postcondition: Configuration snippet has been deleted from the database and can neither be found through search, nor be viewed in detail anymore.
```
markus2330: What about some "undo" feature? (Can be implemented as backup of the whole key database, e.g. in git)

doc/rest_api/snippet_sharing/use_cases/UC_edit_user.txt
Namoshek: They are valid immediately, although I'm not entirely sure how to properly update the session (token) yet (that only concerns the user who has been changed).

doc/rest_api/snippet_sharing/use_cases/UC_details_snippet.txt
Namoshek: Does the use case really have to contain all detailed information already? What exactly the response contains is specified in the API description anyway. An example what meta information can mean is already given...

doc/rest_api/snippet_sharing/use_cases/UC_delete_snippet.txt
Namoshek: Implementing it with git could be tricky if two deletes happen immediately after each other (what and how to roll back?), better would be to flag deleted resources and restore them. The problem with this approach is though that deleted resources should - after the delete - always return 404 not found for their URI. So how to properly restore them? In my opinion restoring them via create seems the best option, but I would not implement that in the API, but rather in the front-end (save the last deleted entry and automatically recreate it for undo).

doc/rest_api/snippet_sharing/use_cases/UC_edit_user.txt
Namoshek: Haven't planned scoring and comments for the current version, but it is something that could be added quite easily later on I think.

doc/rest_api/snippet_sharing/use_cases/UC_edit_user.txt
```diff
@@ -0,0 +1,17 @@
+Summary:
+----------------------------
+Title: edit user
+Scope: User Management
+Level: User Goal
+Actors: Authenticated user, Administrator
+Brief: Users can edit their own account information (e.g. change password). Administrators can change all user accounts in the database. Administrators can also change permissions of users.
```
markus2330: What I am missing here (it might be what you understand by "change permissions") some suspend-account functionality to block users which are believed to be spammers (and then to reactivate them if we find out only their account was hacked).

Also some standard maintenance tasks such as reset password seems to be missing?

Namoshek: You are right, a password reset function is missing, because currently I've no idea how to deal with emails yet. Could also be an idea to add email verification during registration (that doesn't really change the API anyway).

The problem with spammers is that they can simply create a new account and spam on. It is basically the same problem you have on any platform. Absolutely no idea how to prevent that without taking away write permissions by default.

Namoshek:
I added a technical description (quite sure I forgot some things, I'll try to think about it some more today), a distinction for use cases (what is not possible/planned), additional use cases and an updated version of the API description (added the conversion endpoint).

doc/api_blueprints/snippet-sharing.apib
```diff
@@ -16,14 +16,16 @@ Returns an overview of available API methods with required parameters and a desc
...
-        + version (string) - The currently used version in its complete format
-        + major (number) - The currently used major version
-        + minor (number) - The currently used minor version
-        + micro (number) - The currently used micro version
+        + elektra (object) - Detailed version information about the used Elektra library
+            + version (string) - The currently used version in its complete format
+            + major (number) - The currently used major version
+            + minor (number) - The currently used minor version
+            + micro (number) - The currently used micro version
+        + api (number) - The version of the API itself
```
markus2330: Please add some information here: when will you increment this number.

Namoshek:
Some notes for @omnidan:
- Empty lines between sections: The number of empty lines depends on the section level. For top-level sections (one #) use three empty lines, for the second-level sections (two #) two empty lines and for third-level sections (three #) only one empty line.
- HTTP status codes:
 - `200` if the request was successful (it would be possible to use 201, but only for POST; for PUT it is not suitable, so better stick to one common status code).
 - `400` if the request did either not contain the required data or contained data that is not suitable to be processed (e.g. data does not match the content type of the header or does not pass validation).
 - `401` if the request requires authentication, but did not contain a session token or the user lacks permission (exception: authenticate function, there it is also thrown for invalid credentials).
 - `404` if a non-existing resource is requested.
 - `406` if the request header contains a not supported content type.
 - `422` if the request tries to create an already existing resource.
- Import/Export: The configuration format has now to be supplied when creating a database entry from a configuration snippet. Only formats that the API knows are allowed. For export, simply put the `raw` GET-parameter to the desired format.

I'm not so sure about oAuth2 from GitHub yet. Will have to read up on this tomorrow a bit more in detail. But this is not relevant for @omnidan anyway.

omnidan:
@Namoshek Thanks for the summary of response codes that you used, it was really helpful!

> Empty lines between sections: The number of empty lines depends on the section level. For top-level sections (one #) use three empty lines, for the second-level sections (two #) two empty lines and for third-level sections (three #) only one empty line.

I did this when I split up the blueprints :wink:

> 200 if the request was successful (it would be possible to use 201, but only for POST; for PUT it is not suitable, so better stick to one common status code).

I use `204` for PUT/DELETE (since there is no response body) and `200` for POST, what do you think about this?

The error codes make sense and I used and documented them in my blueprints now. I also made sure to re-use your i18n codes where possible, e.g. `NEED_AUTHENTICATION`

BTW: Overriding the description of a data structure like this does not overwrite the other information (type, etc):

```
    + Attributes (Error)
        + i18n
            A comprehensive list of possible errors:
            - `APIKEY_INVALID` api key not specified or invalid
```

Namoshek:
> I use 204 for PUT/DELETE (since there is no response body) and 200 for POST, what do you think about this?

If you use 204 for PUT/DELETE, then you should definitely use 201 for POST as you are already using more specific response codes.

My PUT/DELETE requests also return a response with a i18n string when they succeed. It is not entirely necessary to be honest, but somehow I think it is better to have a common way to transport i18n keys than mixing them within frontend and backend. If I would return an empty response, I'd have to define the success key in the frontend and error keys in the backend, that doesn't seem to be a good design. What do you think?

> The error codes make sense and I used and documented them in my blueprints now. I also made sure to re-use your i18n codes where possible, e.g. NEED_AUTHENTICATION

Awesome!

> BTW: Overriding the description of a data structure like this does not overwrite the other information (type, etc):

I tried it this way, but there are several problems:
- The first part of the description is missing (overridden by an empty description).
- When using apiary the output looks aweful (there is no dividing line between i18n and the next attribute and i18n is always the first attribute, what is not intended).
- The type seems to be not overridden because `string` is the default of Blueprints.

If there is a way to solve all this problems, I'd really, really love to use a data structure for my error response, as about 80% are always the same.

omnidan:
> If you use 204 for PUT/DELETE, then you should definitely use 201 for POST as you are already using more specific response codes.

Alright, I changed that.

> My PUT/DELETE requests also return a response with a i18n string when they succeed. It is not entirely necessary to be honest, but somehow I think it is better to have a common way to transport i18n keys than mixing them within frontend and backend. If I would return an empty response, I'd have to define the success key in the frontend and error keys in the backend, that doesn't seem to be a good design. What do you think?

You're right. Although, for my API, adding an `i18n` field everywhere doesn't make much sense as I never have a "success message", I only care about the data and errors in case they happen.

> I tried it this way, but there are several problems:

Ah, I see, I thought it was inheriting the type. Re-writing only the i18n value:

```
        + i18n (string) - a unique token representing above error description message; can be used for internationalization in frontends
```

fixes all the issues except the dividing line (I don't think it's so bad that it's at the beginning), which seems to be a bug in apiary. I think that's still more readable than copy&pasting all of the attributes.

Namoshek:
Ok, I changed it as well. I hope it will not cause troubles with Dredd later on.

doc/api_blueprints/snippet-sharing.apib
```diff
@@ -1,4 +1,5 @@
 FORMAT: 1A
+HOST: api.libelektra.org
```
markus2330: Maybe "webapi.libelektra.org" or "restapi" is better?

Namoshek: I only added it as placeholder for the moment. I'll have to change it to some localhost url for testing, that's why I needed it. We can change it when the API and the deployment decisions are final.

markus2330:
Build seems to fail?

Please remove "work in progress" if it is ready to review.

Namoshek:
```
ELEKTRA CHECK FORMATTING

diff --git a/src/error/exporterrors.cpp b/src/error/exporterrors.cpp
index 63f79f9..63c9a64 100644
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
```

Must be something broken on the master branch, can't be my fault.

I changed the host, so lets see if the tests still fail.

markus2330:
And the last piece! Thanks!

PR #937, opened 2016-09-14
install zsh and bash completions

### LINKED ISSUE #844: OS X: Install BASH and ZSH Completion for User
Currently Elektra installs shell completion commands in the directories 
```
/usr/share/bash-completion/completions
``` 
and 
```
/usr/share/zsh/vendor-completions
```
.

In OS X `/usr/share` is part of the directories protected by the [System Integrity Protection](http://apple.stackexchange.com/questions/193368/what-is-the-rootless-feature-in-el-capitan-really). This means  even `root` can not write to `/usr/share`. Because of this, the installation of the completion commands under OS X fails.

This issue could be fixed by changing the default directory for the completion commands. For `bash` the directory `/usr/local/etc/bash_completion.d` looks like a good alternative ([1]).   

[1]: https://github.com/ElektraInitiative/libelektra/pull/822#issuecomment-233563951

For `zsh` the directory `/usr/local/share/zsh/site-functions` looks like a good fit, since it is part of the default `$fpath`:

```sh
> echo $fpath
/usr/local/share/zsh/site-functions /usr/share/zsh/site-functions /usr/share/zsh/5.0.8/functions
```

reox:
# Purpose

* Fixes #844 
* Adding a tool to install the required completion stuff into .rc files

# Checklist

needs to be checked.

- [ ] I added unit tests


reox:
okay, i rebased the cmake file (hopefully w/o breaking anything)
I also added some more interactiveness, the user is now asked to install zsh and bash and is asked to backup the files each time.

markus2330:
Thank you, that are a lot of questions ;) (Maybe y/n/b would be an option?)

Can someone test on Mac OS X please?

scripts/install-sh-completion
```diff
@@ -0,0 +1,84 @@
...
+    if [[ -f ~/.bashrc ]]; then
+        if ask "Create a backup of .bashrc?" Y; then
+            cp -i ~/.bashrc ~/.bashrc.bak_kdb
+        fi
+    fi
+
+cat <<HERE >>~/.bashrc
+### generated by kdb install-sh-completion
+# Linux
+if [ -d /etc/bash_completion.d ]; then
```
markus2330: Actually we do not have a clue where the file is installed to. Why not cat (with HERE document) the whole file into the home directory and source it?

reox: thats true.
That means we should copy the content of bash completion file with cmake to the installer script, because if the script is installed we can not know where the completion files are, right?
Is there any simple and nice way to do that?

markus2330: "simple and nice" are very high demands for CMake, let us say its possible with little code: see in tests/shell/CMakeLists.txt the include_common.sh is pasted at the beginning of every shell test.

markus2330: Wait, its not a ideal recommendation to do it the same way (because you likely want to paste in the middle of your file).

So you could do something like:
```
file(READ "${PATH_TO}/zsh-completion-file" CONTENT_OF_ZSH_COMP)
configure_file ( your-script.in your-script @ONLY)
```

and your-script contains:
```
cat << HERE
@CONTENT_OF_ZSH_COMP@
HERE
```

markus2330:
Are you still working on this? Otherwise I would only disable installation of the files on Mac OS X and leave this for after the release.

reox:
yes, i need to build it first but I'm missing some packages on my system to run cmake ;)

markus2330:
Basically cmake (2.8.8 or later)+build-essential should be enough to build Elektra. Plugins will automatically be excluded if you do not have their deps installed.

reox:
`ruby-ronn` was missing... It's currently building, then i will test it and push it

markus2330:
Thank you, you got an error without ruby-ronn? Actually in this case the building of docu should disable automatically.

reox:
mh yes, i don't have the log anymore but after I installed it (and `pkg-config`) it builded

reox:
Okay i tested it but it seems it is not possible to paste them via heredoc...
I get a lot of errors when running the script. I'm not sure whats wrong there :/

scripts/CMakeLists.txt
```diff
@@ -25,6 +25,11 @@ install (PROGRAMS umount-all DESTINATION ${TARGET_TOOL_EXEC_FOLDER})
 
 install (PROGRAMS install-sh-completion DESTINATION ${TARGET_TOOL_EXEC_FOLDER})
 
+file(READ "kdb_zsh_completion" CONTENT_OF_ZSH_COMP)
```
markus2330: always use full pathes, e.g. here CMAKE_CURRENT_SOURCE_DIR
see https://cmake.org/Wiki/CMake_Useful_Variables

scripts/CMakeLists.txt
```diff
@@ -25,6 +25,11 @@ install (PROGRAMS umount-all DESTINATION ${TARGET_TOOL_EXEC_FOLDER})
...
+file(READ "kdb-bash-completion" CONTENT_OF_BASH_COMP)
+configure_file ( install-sh-completion.in install-sh-completion @ONLY)
```
markus2330: also here..

markus2330:
Btw. pasting the errors would help..

markus2330:
@reox no solution here?

reox:
Not yet... 

Am 16. September 2016 15:58:49 MESZ, schrieb markus2330 <notifications@github.com>:

reox:
okay, it seems to be fixed now. single quotes are needed to not interpret the text in the heredoc section.

markus2330:
Thank you. We now have to fixes for Darwin, needs to be cleaned up later...

Why Darwin and not APPLE?

reox:
https://cmake.org/Wiki/CMake_Useful_Variables says:

>  APPLE  is TRUE on Apple systems. Note this does not imply the system is Mac OS X, only that __APPLE__ is #defined in C/C++ header files. Obtain more specific system information via CMAKE_SYSTEM_VERSION, i.e. IF(${CMAKE_SYSTEM_NAME} MATCHES "Darwin"), then it's Mac OS X. 

So i thought it would be better to use the Darwin check... But I'm not sure if this is really necessary...

markus2330:
Thanks, fixed the other place

PR #941, opened 2016-09-15
crypto-plugin: add /usr/bin/gpg as second fallback

### LINKED ISSUE #940: crypto-plugin: try using /usr/bin/gpg as fallback too
# Problem

At the moment the crypto plugin (gpg module) tries to use `/usr/bin/gpg2` as fallback value (e.g. if the `PATH` variabel is empty). If `gnupg2` is not installed, the module fails to find the executable.

# Desired Behavior

The gpg module also looks for `/usr/bin/gpg` as fallback solution, if the lookup in `PATH` was unsuccessful.

# Why is it important?

We came across this issue when certain build agents failed to run the test cases. The `PATH` environment variable is empty when the build job runs. If `gnupg2` is not installed on the build agent, the test cases fail.

petermax2:
# Purpose

If the PATH environment variable is empty we check if `/usr/bin/gpg2` or `/usr/bin/gpg` is available as a fallback.

Fixes #940.

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/src/plugins/crypto/gpg.c
+++ b/src/plugins/crypto/gpg.c
@@ -199,33 +199,28 @@ static int getGpgBinary (char ** gpgBin, KeySet * conf, Key * errorKey)
 
 	// last resort number one - check for gpg2 at /usr/bin/gpg2
 	// NOTE this might happen if the PATH variable is empty
-	*gpgBin = elektraStrDup (ELEKTRA_CRYPTO_DEFAULT_GPG2_BIN);
-	if (!(*gpgBin))
+	if (isExecutable (ELEKTRA_CRYPTO_DEFAULT_GPG2_BIN, NULL) == 0)
 	{
-		ELEKTRA_SET_ERROR (87, errorKey, "Memory allocation failed");
-		return -1;
-	}
-
-	if (isExecutable (*gpgBin, NULL) == 0)
-	{
+		*gpgBin = elektraStrDup (ELEKTRA_CRYPTO_DEFAULT_GPG2_BIN);
+		if (!(*gpgBin))
+		{
+			ELEKTRA_SET_ERROR (87, errorKey, "Memory allocation failed");
+			return -1;
+		}
 		return 1;
 	}
-	elektraFree (*gpgBin);
 
 	// last last resort - check for /usr/bin/gpg
-	*gpgBin = elektraStrDup (ELEKTRA_CRYPTO_DEFAULT_GPG1_BIN);
-	if (!(*gpgBin))
+	if (isExecutable (ELEKTRA_CRYPTO_DEFAULT_GPG1_BIN, NULL) == 0)
 	{
-		ELEKTRA_SET_ERROR (87, errorKey, "Memory allocation failed");
-		return -1;
-	}
-
-	if (isExecutable (*gpgBin, NULL) == 0)
-	{
+		*gpgBin = elektraStrDup (ELEKTRA_CRYPTO_DEFAULT_GPG1_BIN);
+		if (!(*gpgBin))
+		{
+			ELEKTRA_SET_ERROR (87, errorKey, "Memory allocation failed");
+			return -1;
+		}
 		return 1;
 	}
-	elektraFree (*gpgBin);
-	*gpgBin = NULL;
 
 	// no GPG for us :-(
 	ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "no gpg binary found. Please make sure GnuPG is installed and executable.");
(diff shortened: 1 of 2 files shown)


petermax2:
I opened an issue for the memory leak ( #942 ).

src/plugins/crypto/gpg.c
```diff
@@ -166,16 +197,39 @@ static int getGpgBinary (char ** gpgBin, KeySet * conf, Key * errorKey)
 	}
 
 
-	// last resort
-	const size_t defaultLen = strlen (ELEKTRA_CRYPTO_DEFAULT_GPG_BIN) + 1;
-	*gpgBin = elektraMalloc (defaultLen);
+	// last resort number one - check for gpg2 at /usr/bin/gpg2
+	// NOTE this might happen if the PATH variable is empty
+	*gpgBin = strdup (ELEKTRA_CRYPTO_DEFAULT_GPG2_BIN);
```
markus2330: Please use elektraStrDup here.

And why use pointer to pointer? (not needed here, maybe you can put this fallback code to separate function altogether?)

src/plugins/crypto/gpg.c
```diff
@@ -166,16 +197,39 @@ static int getGpgBinary (char ** gpgBin, KeySet * conf, Key * errorKey)
...
+	if (!(*gpgBin))
+	{
+		ELEKTRA_SET_ERROR (87, errorKey, "Memory allocation failed");
+		return -1;
+	}
+
+	if (isExecutable (*gpgBin, NULL) == 0)
+	{
+		return 1;
```
markus2330: There are memleaks (does not free gpgBin, also see below).

src/plugins/crypto/gpg.c
```diff
@@ -327,18 +381,10 @@ int elektraCryptoGpgCall (KeySet * conf, Key * errorKey, Key * msgKey, char * ar
...
 	{
-		ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "gpg binary %s not found", argv[0]);
 		elektraFree (argv[0]);
-		return -1;
-	}
-
-	if (access (argv[0], X_OK))
-	{
-		ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "gpg binary %s has no permission to execute", argv[0]);
-		elektraFree (argv[0]);
```
markus2330: Why not free `argv[0]`?

src/plugins/crypto/gpg.c
petermax2: No, `gpgBin` has to be freed by the caller. It gets deallocated only if an error occurs. Please see the doc of the method:

	/**
	 * @brief lookup the path to the gpg binary in conf.
	 * @param gpgBin holds allocated path to the gpg binary to be used or NULL in case of an error. Must bee freed by the caller.
	 * @param conf KeySet holding the plugin configuration.
	 * @param errorKey holds an error description if something goes wrong.
	 * @retval 1 on success.
	 * @retval -1 on error. In this case errorkey holds an error description.
	 */

src/plugins/crypto/gpg.c
petermax2: It is freed or am I mistaking?

	// check that the gpg binary exists and that it is executable
	if (isExecutable (argv[0], errorKey) < 0)
	{
		elektraFree (argv[0]);
		return -1; // error set by isExecutable()
	}

src/plugins/crypto/gpg.c
markus2330: I expected you checked for errors, not for success. Can you add a small comment there?

src/plugins/crypto/gpg.c
markus2330: Sorry my fault, I did not look at the upper part of the diff.

src/plugins/crypto/gpg.c
```diff
@@ -25,6 +25,37 @@ static inline void closePipe (int * pipe)
 }
 
 /**
+ * @brief checks whether or not a given file exists and is executable.
+ * @param file holds the path to the file that should be checked
+ * @param errorKey holds an error description if the file does not exist or if it is not executable. Ignored if set to NULL.
+ * @retval 1 if the file exists and is executable
```
petermax2: I can see how `0` indicating success can be confusing. I changed it to `1` to be consistent with the rest of the module.

petermax2:
Thank you for the feedback. I restructured `getGpgBinary ()` a little bit. Hopefully the fallback logic is now easier to understand.

markus2330:
Nice solution!

PR #943, opened 2016-09-15
passwd: fixed build

tom-wa:


markus2330:
Thank you!

tom-wa:
@markus2330 seems like there's still an issue: it only works after running cmake twice. any idea why ?

markus2330:
Yes, I was a bit sceptical if this would work..

You need to build up a string in DEPENDENCY_PHASE and pass the whole string. Then either everything or nothing is passed to add_plugin.

Btw. please always paste error messages. E.g. in this case: what happens when running cmake only once. In this case it was easy to spot, but you should make a habit out of always pasting error messages.

PR #946, opened 2016-09-15
fix openwrt warnings

tom-wa:

### DIFF
--- a/src/include/kdbtypes.h
+++ b/src/include/kdbtypes.h
@@ -98,10 +98,12 @@ typedef long double kdb_long_double_t;
 typedef long double kdb_long_double_t;
 #endif
 
-#if SIZEOF_TV_USEC == SIZEOF_INT
-#define ELEKTRA_TIME_USEC_F "%d"
-#else
+#if SIZEOF_TV_USEC == SIZEOF_LONG
 #define ELEKTRA_TIME_USEC_F "%ld"
+#elif SIZEOF_TV_USEC == SIZEOF_LONG_LONG
+#define ELEKTRA_TIME_USEC_F "%lli"
+#else
+#define ELEKTRA_TIME_USEC_F "%d"
 #endif
 
 #endif


src/include/kdbtypes.h
```diff
@@ -98,10 +98,12 @@ typedef long double kdb_long_double_t;
...
 #endif
 
-#if SIZEOF_TV_USEC == SIZEOF_INT
-#define ELEKTRA_TIME_USEC_F "%d"
-#else
+#if SIZEOF_TV_USEC == SIZEOF_LONG
 #define ELEKTRA_TIME_USEC_F "%ld"
+#elif SIZEOF_TV_USEC == SIZEOF_LONG_LONG
+#define ELEKTRA_TIME_USEC_F "%lli"
+#else
```
markus2330: What about also adding == SIZEOF_INT and return an warning if an unhandled size occurs?

Is this ready to merge then?

tom-wa: >What about also adding == SIZEOF_INT and return an warning if an unhandled size occurs?

i was just wondering: does defining format strings that way even work properly ? for me sizeof(long long) == sizeof(long), and afaik there are systems where sizeof(long) == sizeof(int) and sizeof(int) == sizeof(short) 

what should it default to if an unhandled size occurs ?

src/plugins/tcl/CMakeLists.txt
```diff
@@ -2,13 +2,13 @@ if (DEPENDENCY_PHASE)
 	find_package (Boost)
 	if (NOT Boost_FOUND)
 		remove_plugin (tcl "boost not found, but needed for tcl plugin")
+	else ()
+		include_directories(SYSTEM ${Boost_INCLUDE_DIR})
```
markus2330: Can you remove this commit? Your solution includes the directory globally, target_include_directory is not yet available in cmake 2.8.8, thus I made a monstrous workaround in 97654f5e293f5c0de61041f3febc3ad5ebdbb1a6.

tom-wa:
removed

what should `ELEKTRA_TIME_USEC_F` default to when it's an unhandled size ?

markus2330:
Thank you! Hard to tell what the best option on such systems is, maybe use the most portable solution? But there is not much benefit of adding it, because you get a warning on the wrong printf specifier anyway, so I merged the PR.

PR #950, opened 2016-09-15
ENABLE_DEBUG: no RTLD_NODELETE for OpenBSD

tom-wa:
# Purpose

don't set RTLD_NODELETE when building with ENABLE_DEBUG on OpenBSD

### DIFF
--- a/src/libs/loader/CMakeLists.txt
+++ b/src/libs/loader/CMakeLists.txt
@@ -5,7 +5,9 @@ add_sources (elektra-full "static.c")
 if (BUILD_SHARED)
 	add_sources (elektra-shared "dl.c")
 	add_includes (elektra-shared)
-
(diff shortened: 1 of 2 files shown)


markus2330:
Do this systems have RTLD_NODELETE at all? What about checking if RTLD_NODELETE is available? (Similar to what is done in passwd)

markus2330:
Can you make a check for RTLD_NODELETE? (It should work with check_symbol_exists)

markus2330:
Any progress here? Would be good to fix ENABLE_DEBUG for *BSD.

tom-wa:
@markus2330 done

markus2330:
Thank you!

I removed the cmake check 6b8e40957ba5971657816a87850d00775d58e8dc

PR #970, opened 2016-09-19
Use cases: sharing of configuration snippets

Namoshek:
# Purpose

Use cases for sharing of configuration snippets.
# Checklist

### DIFF
--- a/doc/usecases/snippet_sharing/UC_convert_snippet.md
+++ b/doc/usecases/snippet_sharing/UC_convert_snippet.md
@@ -1,0 +1,19 @@
+# Use Case: Convert configuration snippet
+
+## Summary
+
+Title: Convert configuration snippet
+Scope: Configuration conversion
+Level: User Goal
+Actors: Anonymous user (+ all others)
+Brief: User can input a configuration snippet in any supported format (e.g. ini, xml) and convert it into another supported format (e.g. json).
+
+## Scenarios
+
+Precondition: Service supports the in- and output configuration format
+Main success scenario: All inputs validate and the configuration snippet could be converted into the target format. The user is then shown the result of the conversion - the input snippet in another format.
+Alternative scenario: Wrong inputs lead to an error message and ask the user to correct them, then they can try again. It is also possible that the supplied configuration snippet is of an unsupported format. In either case the user retrieves a response containing error information.
+Error scenario: Technical problems prevent the conversion from completing. The user is informed about the issue.
+Postcondition: No changes have been made to the database.
+Non-functional Constraints: 
+	- Show-case functionality of Elektra
--- a/doc/usecases/snippet_sharing/UC_duplicate_snippet.md
+++ b/doc/usecases/snippet_sharing/UC_duplicate_snippet.md
@@ -15,5 +15,5 @@ Main success scenario: All inputs validate and the configuration snippet could b
 Alternative scenario: Wrong inputs lead to an error message and ask the user to correct them, then they can try again. It is also possible that the supplied configuration snippet is of an unsupported format. In either case the user retrieves a response containing error information.
 Error scenario: Technical problems prevent the storage process from completing. The user is informed about the issue.
 Postcondition: Configuration snippet is stored in the database successfully and can be found through search and direct link.
-Non-functional Constraints: 
-	- Essential functionality of the service
+Non-functional Constraints:
+	- Can be used to create snippet histories
--- a/doc/usecases/snippet_sharing/UC_duplicate_snippet.md
+++ b/doc/usecases/snippet_sharing/UC_duplicate_snippet.md
@@ -1,0 +1,19 @@
+# Use Case: Create configuration snippet from existing snippet
+
+## Summary
+
+Title: Create configuration snippet from existing snippet
+Scope: Entry Management
+Level: User Goal
+Actors: Authenticated user
+Brief: User can create a new configuration snippet (database entry) based on an existing snippet. The existing snippet will be filled in the mask for the creation of a new snippet.
+
+## Scenarios
+
+Precondition: Authenticated, sufficient permissions
+Main success scenario: All inputs validate and the configuration snippet could be parsed and stored in the Elektra database successfully. The user is then redirected to the detailed view of the created snippet.
+Alternative scenario: Wrong inputs lead to an error message and ask the user to correct them, then they can try again. It is also possible that the supplied configuration snippet is of an unsupported format. In either case the user retrieves a response containing error information.
+Error scenario: Technical problems prevent the storage process from completing. The user is informed about the issue.
+Postcondition: Configuration snippet is stored in the database successfully and can be found through search and direct link.
+Non-functional Constraints: 
+	- Essential functionality of the service
--- a/doc/usecases/snippet_sharing/UC_edit_snippet.md
+++ b/doc/usecases/snippet_sharing/UC_edit_snippet.md
@@ -1,0 +1,19 @@
+# Use Case: Edit configuration snippet
+
+## Summary
+
+Title: Edit configuration snippet
+Scope: Entry Management
+Level: User Goal
+Actors: Authenticated user (snippet owner), Moderator
+Brief: Users can edit their own database entries. Moderators can edit all configuration snippets in the database, also the ones of other users.
+
+## Scenarios
+
+Precondition: Authenticated, sufficient permissions
+Main success scenario: Changes to the configuration snippet have been stored to the database successfully. The user is informed about the success.
+Alternative scenario: The submitted inputs could not be validated or the entry for which changes are requested cannot be found. The user is informed about the error.
+Error scenario: Technical problems prevent the storage process from completing. The user is informed about the issue.
+Postcondition: Changes to the configuration snippet are stored in the database successfully and can be immediately seen in new requests of the snippet.
+Non-functional Constraints:
+	- Moderative feature
--- a/doc/usecases/TEMPLATE.md
+++ b/doc/usecases/TEMPLATE.md
@@ -1,0 +1,20 @@
+# Use Case: <Title>
+
+## Summary
+
+Title: <Title, e.g. Authenticate>
+Scope: <Scope, e.g. Authentication>
+Level: <Level, e.g. User Goal>
+Actors: <Actors, e.g. Anonymous User>
+Brief: <Short explanation, e.g. User authenticates against the service to gain more privilegues>
+
+## Scenarios
+
+Precondition: <What has to be satisfied before the use case can work?, e.g. Not authenticated>
+Main success scenario: <What is the main scenario, i.e. success scenario?, e.g. Is authenticated>
+Alternative scenario: <What is the alternative scenario, i.e. simple error handling?, e.g. Wrong credentials>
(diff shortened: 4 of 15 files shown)


doc/usecases/snippet_sharing/UC_duplicate_snippet.md
```diff
@@ -0,0 +1,19 @@
+# Use Case: Create configuration snippet from existing snippet
```
markus2330: create vs. duplicate?

doc/usecases/snippet_sharing/UC_duplicate_snippet.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Scenarios
+
+Precondition: Authenticated, sufficient permissions
+Main success scenario: All inputs validate and the configuration snippet could be parsed and stored in the Elektra database successfully. The user is then redirected to the detailed view of the created snippet.
+Alternative scenario: Wrong inputs lead to an error message and ask the user to correct them, then they can try again. It is also possible that the supplied configuration snippet is of an unsupported format. In either case the user retrieves a response containing error information.
+Error scenario: Technical problems prevent the storage process from completing. The user is informed about the issue.
+Postcondition: Configuration snippet is stored in the database successfully and can be found through search and direct link.
+Non-functional Constraints: 
+	- Essential functionality of the service
```
markus2330: I do not think it is essential.

doc/usecases/snippet_sharing/UC_edit_snippet.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Scenarios
+
+Precondition: Authenticated, sufficient permissions
+Main success scenario: Changes to the configuration snippet have been stored to the database successfully. The user is informed about the success.
+Alternative scenario: The submitted inputs could not be validated or the entry for which changes are requested cannot be found. The user is informed about the error.
+Error scenario: Technical problems prevent the storage process from completing. The user is informed about the issue.
+Postcondition: Changes to the configuration snippet are stored in the database successfully and can be immediately seen in new requests of the snippet.
+Non-functional Constraints:
+	- Moderative feature
```
markus2330: Why only moderative?

doc/usecases/snippet_sharing/UC_convert_snippet.md
```diff
@@ -0,0 +1,19 @@
+# Use Case: Convert configuration snippet
+
+## Summary
+
+Title: Convert configuration snippet
+Scope: Configuration conversion
+Level: User Goal
+Actors: Anonymous user (+ all others)
+Brief: User can input a configuration snippet in any supported format (e.g. ini, xml) and convert it into another supported format (e.g. json).
```
markus2330: As discussed: Maybe its better to have this as front-end feature and not as API feature?

markus2330:
Btw. it was mentioned last meeting but not discussed: I think it might be useful to download parts of snippets, e.g. when the whole apache config is uploaded, one might be interested in the config of a module.

Namoshek:
That is a really tough one. What do you have in mind how it could/should work? I see a few options here, but I'm not sure if I like either of these. Basically, the most straight forward option would be to specify a snippet specific config path additionally to the snippet path to download only the configuration below this path. So you could only get the config below one key, but never the config of two siblings at once for example (absolutely no idea how the latter could work and still make sense to a user).

I'm also not too sure how it could look in the UI or how the user could specify what he wants to download...

markus2330:
> That is a really tough one. What do you have in mind how it could/should work?

I thought it would be rather easy: If you import the snippets into Elektra you have all subtrees available anyway.

> additionally to the snippet path

You only make the path longer, so apache/vhosts instead of only apache.

> two siblings at once for example

I do not think it will be used (so better to not add this feature). But it could be done, e.g. by using XPath.

> look in the UI

If you provide all subkeys as "menu" it should be ok?

Namoshek:
> You only make the path longer, so apache/vhosts instead of only apache.

That sounds like the way to go and rather easy, but...

> If you provide all subkeys as "menu" it should be ok?

Not sure if this is as trivial as it sounds, because this would require some sort of own format implementation. Currently I'm just passing the converted snippets as plain string from the backend to the frontend. But for this to work, I'd need a better and direct integration into the API responses, so I know what paths to query...?! And what about special configuration stuff like overrides?

markus2330:
> Not sure if this is as trivial as it sounds, because this would require some sort of own format implementation.

What do you mean? You can simply store them in dump (or anything else that is feature complete with meta data).

> Currently I'm just passing the converted snippets as plain string from the backend to the frontend.

See what kdb import/export or qt-gui does, it is nothing complicated. Improvements in the libtools API, however, would be highly appreciated. (using plugin assembling logic from mounting to determine how to import+export)

> But for this to work, I'd need a better and direct integration into the API responses, so I know what paths to query...?!

You mean the "ls" feature? @omnidan actually should have done this by yesterday.

> And what about special configuration stuff like overrides?

That is irrelevant for you, you assume the configuration to be within a single namespace without any spec.

Namoshek:
> You mean the "ls" feature? @omnidan actually should have done this by yesterday.

Well, sort of. But I'm not sure if showing paths without actual configuration (values) besides it is enough? I mean it would be nice if I could show a nicely formatted tree of configuration key-value pairs.

An almost perfect solution would be something like GitHub does with code reviews. You select a line and it gives you the output for everything below this key. I don't think it is too hard to implement in the backend, but the frontend.

markus2330:
I think this goes too far and is too much effort. Even cutting files by ranges (from .. to ..) would be easier to implement. Because it is likely a rarely used feature, appending to URL or via REST should be enough (unless you find a nice frontend solution).

Namoshek:
Well, cutting files from line to line could also be done via `sed` (from a users perspective). To know the line numbers you'd have to know the output in a certain format too. But I'm afraid basic cutting could result in undesired behavior (think about cutting a dump file at the wrong places).

markus2330:
Yes, it is not useful to provide such a feature. Would be problematic for XML and JSON, too. Config files are not always line based, so a feature "just like github" would also sometimes be inconvenient. To specify a part of snippet via URL would be interesting, though.

doc/usecases/snippet_sharing/UC_edit_snippet.md
```diff
@@ -0,0 +1,19 @@
+# Use Case: Edit configuration snippet
+
+## Summary
+
+Title: Edit configuration snippet
+Scope: Entry Management
+Level: User Goal
+Actors: Authenticated user (snippet owner), Moderator
+Brief: Users can edit their own database entries. Moderators can edit all configuration snippets in the database, also the ones of other users.
```
markus2330: What about storing different versions of snippets, is there some undo? And is there a help system for explanations? (E.g. capabilities of formats)

Namoshek: No storing of different versions. But you can create a new snippet from the old one (in the GUI), which you could use to build a history if you really need it.

Help system in what sense? Some explanations of what is what and how it works should obviously be part of the web GUI later for sure. But that's not worth being mentioned in the use cases.

markus2330: > Help system in what sense?

I mean beyond small hover texts but descriptions of configuration format syntax, limitations and so on. Plugins should document this as part of their README.md so maybe it is easiest to simply show the rendered README.md (if someone wants to understand what a configuration file format is capable of doing).

Namoshek: To do so, I would need to retrieve a mapping between plugins and formats from the API, but I think that's easily possible, so why not.

markus2330: Thank you! Did you have time to investigate the github API? Can we use github for authentification (should be noted in the use cases, too)?

What might be great for Elektra development is some easy way to report issues that consideres context, e.g. if something is wrong with a plugin (README.md wrong, import fails,...) that a user already can create a new issue with the boring stuff already already filled out (Which plugin was used, what was the snippet that should be imported, what was the error message, and so on...).

Namoshek: Yes, I had a look at the GitHub API and I don't think there would be a problem in using it, but I'm not sure how much effort it takes to implement the whole registration process in C++ (not really the language that is normally used with oAuth systems). I think there could be some helpful libraries, but that also means more dependencies. Some library for HTTP requests is the minimal requirement though.

For the second point: No idea if GitHub does allow something like this, but that is definitely a total-minor on my list.

doc/usecases/snippet_sharing/UC_duplicate_snippet.md
Namoshek: I think my wording describes better what is happening. "Duplicate" sounds to me like an atomic action that does not allow editing in the same step, but my implementation will allow editing before the snippet gets actually stored (frontend-only feature). So there is never really a duplicate created (except when the user wants to do so).

doc/usecases/snippet_sharing/UC_edit_snippet.md
Namoshek: What else should it be in your opinion? Personally I don't think it is essential, as you can achieve the same with create + delete.

doc/usecases/snippet_sharing/UC_convert_snippet.md
Namoshek: It is bad design to create a dummy snippet only for conversion and delete it again afterwards. The additional endpoint is no effort and already done.

doc/usecases/snippet_sharing/UC_edit_snippet.md
markus2330: Is it possible to do OAUTH on the client side? Maybe its possible to store some information at github and use them for registration/login?

> No idea if GitHub does allow something like this, but that is definitely a total-minor on my list.

Seems [to be available](https://developer.github.com/v3/issues/#create-an-issue) and could be really nice for manual testing of plugins. But sending an e-mail or simply logging errors is okay, too.

Namoshek: Most of the oAuth process would be frontend stuff, which is not the problem (there is an [awesome angular library](https://github.com/sahat/satellizer) for that), but it also requires the server to validate the access token by doing calls to GitHub (obviously). I'm going to do other stuff first before investing time into that.

Your link for the issue creation is an API call that creates an issue directly; it does not open the GitHub form with pre-filled fields for you. Implementing a second GitHub frontend doesn't seem to be the best idea...

Namoshek:
I think we can merge this PR too. The open points oAuth and downloading of config-parts are very specific and can be implemented later if necessary as well, I will not do them / get them done before our release anyway.

doc/usecases/snippet_sharing/UC_edit_snippet.md
markus2330: I was confused because you put it below "constraint". I would not say it is constrainted to be a moderative feature. The actors above made it clear for me that you also think that way (they mention "Authenticated user (snippet owner)"). But thus the "Moderative feature" is consistent with the other use cases, let us keep it that way.

markus2330:
Thank you, great job!

PR #986, opened 2016-09-28
lib-tools: new functions to retrieve list/map of providers

Namoshek:
# Purpose

Two new functions for lib-tools that allow to lookup a whole bunch of suitable plugins (e.g. a list/map of all storage plugins). The new functions work basically the same as `PluginDatabase::lookupProvides`, except that they return all found plugins and not only the one with highest status rating.

I refactored the code of `ModulesPluginDatabase::lookupProvides` into the new function `ModulesPluginDatabase::lookupAllProvidesWithStatus`, which is used by the old and the new function.

# Checklist

- [X] reformat-source script was run
- [ ] I added unit tests

### DIFF
--- a/src/libs/tools/include/plugindatabase.hpp
+++ b/src/libs/tools/include/plugindatabase.hpp
@@ -105,6 +105,33 @@ public:
 	virtual PluginSpec lookupProvides (std::string const & provides) const = 0;
 
 	/**
+	 * @brief looks up all plugins which are a suitable provider
+	 *
+	 * @note in case a plugin name is provided, the plugin with the name will also be
+	 *	 part of the result. But if there are other plugins providing the requirement,
+	 *	 then they will also be part of the result.
+	 *
+	 * @param provides is the provider to find
+	 *
+	 * @return a map of plugins with their status offering the requirement or are named after it
+	 */
+	virtual std::map<int, PluginSpec> lookupAllProvidesWithStatus (std::string const & provides) const = 0;
+
+	/**
+	 * @brief looks up all plugins which are a suitable provider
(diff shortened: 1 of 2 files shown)


markus2330:
Thanks for adding stuff to libtools!

The code is good, but I am afraid that we bloat the API here. What about having only `virtual std::map<int, PluginSpec> lookupProvides (std::string const & provides) const = 0;` (the most generic version) and the other helpers are only free-standing functions (working with every *PluginDatabase) taking `std::map<int, PluginSpec>`(which is a typedef for convenience) as argument. Disadvantage: breaks API next to ABI.

Then I would also suggest to pimpl ModulesPluginDatabase (so that we do not have to break ABI again soon). And I am missing a MockPluginDatabase implementation and some tests.

Btw. it would be great to know what your other plans with libtools are so that I can think of a plan how we release the ABI incompatible libtools. Other important tasks would be registering plugins with configuration for PluginDatabase (e.g. for augeas with different lenses) and a ImportExportBackend. I plan to do lazy mounting soon (new class LazyBackend, might also break ABI).

Namoshek:
I decided to go this way to not break the current API and I did the refactoring to not bloat the API completely (basically it is just two small new functions, that could also easily be set as `inline`). As we did not discuss this before, I thought it would be the easiest approach to get the PR done.

Currently I don't know if I'll need more (new) functionality of libtools. It probably depends on our decision how to detect storage plugins that actually work with my REST service. I hope I can finish my prototype today (looks good so far), then I'll open the issue for the discussion.

markus2330:
Ok, so let us discuss this later.

Issue #944, opened 2016-09-15
Shell Recorder Problems


markus2330:
# Overview
- [ ] prefix with tmp if absolute path  #1139
- [ ] prefix with constant (tests?) and unique id on any path (file path) #1753
- [ ] check if only allowed Elektra paths are used (like /examples, /tests, and /tutorials?) #1753 
- [ ] check for left over files (and fail if there are)
- [ ] STDERR hidden
- [ ] diffs on exported symbols
- [ ] might add keys (global plugins)
- [ ] shell recorder should have a way to specify (plugin) dependencies needed to run the test case
- [ ] Cannot iterate over multiple similar plugins (if available) e.g. kdb mv test cases for git resolver and other resolver.
- [x] support HEREDOC
- [x] cleanup tmp file (protocol)
- [x] should fail if backup (export in line 52) could not done be successfully
- [x] should check if KDB is left with any changes that were not reverted by backup+restore (like tests/shell/run_all.sh does)
- [x] rename "Mountpoint:" to "Backup-and-restore:"
- [x] testshell_* testcases do not execute anything, protocol file is empty.
- [x] might remove keys
- [x] protocol files not removed
- [x] sed command might not be terminated correctly (sed: -e Expression #1, Character 86: `s'-command not terminated) caused by data in system/tests/dump when running testshell_db_changes

# Describe what you wanted to do

Run shell recorder tests without necessary plugins.
# Describe what you expected

Get some clear info that plugin is missing
# Describe what actually happened

Output did not show the error at all (STDERR only in protocol).
Ideally we should have some way to specify which plugins a shell-recorder tests and jump over the testcase.

In the protocol files there are many differences of the exported function pointers shown. Why do they not lead to errors?
# System Information
- Elektra Version: master
# Further Log Files and Output

```
CMD: get /sw/org/myapp/#0/current/testkey
RET: 1
STDERR: Did not find key3 Warnings were issued:
 Warning number: 1
        Description: could not load module, dlopen failed
        Ingroup: modules
        Module: dl
        At: /home/jenkins/workspace/workspace/elektra-gcc-configure-xdg/src/libs/loader/dl.c:86
        Reason: of module: libelektra-list.so, because: libelektra-list.so: cannot open shared object file: No such file or directory
        Mountpoint: system/elektra
        Configfile: kdbOpen(): mountGlobals
 Warning number: 64
        Description: could not load plugin in process plugin
        Ingroup: modules
        Module: kdb
[...]

PR #990, opened 2016-09-29
Shell Recorder: Add Plugins syntax for test cases.

KurtMi:


KurtMi:
Draft for #944 
shell recorder should have a way to specify (plugin) dependencies needed to run the test case

markus2330:
Thanks! What is missing in the PR? (It says Draft) Or do you mean that only one subpoint is implemented?

Some further useful features in this context:
- What about if storage already implies Plugins?
- Or even more comfortable, what if we get the plugins from import/export/mount statements, too?

I can move this to #944 if you see them out of scope for this PR.

KurtMi:
I mentioned draft because I was not sure if you are happy with such a solution and the used plugins in the other test cases are missing.

> What about if storage already implies Plugins?
> Or even more comfortable, what if we get the plugins from import/export/mount statements, too?

In my opinion the usage of the storage is not appropriate because it is only a variable which is used later in the test case and it is used to define the storage plugins and we have a bunch of others too.
I think it is possible to parse the import/export/mount statements but very hard to do with cmake syntax.

Do I need to specify the resolver also?

markus2330:
You can assume a "resolver" and "storage" to be present (should checked by cmake at compile time).

Why do you need cmake syntax to parse it? Why not parse within shell recorder?

KurtMi:
As discussed, this problem will be solved in a other way, so i close this PR.

PR #999, opened 2016-10-01
Use cases: elektra-web

omnidan:
# Purpose

Use cases for elektra-web.
# Checklist


doc/usecases/elektra_web/UC_add_instance.md
```diff
@@ -0,0 +1,20 @@
+# Use Case: Add instance
```
markus2330: Where will it be added?

doc/usecases/elektra_web/UC_add_instance.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Summary
+
+Title: Add instance
+Scope: Setup
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User sets up elektrad on the instance and connects it to clusterd according to the corresponding README.
```
markus2330: Try to be a bit more high-level, do not mention elektrad or similar.

Actually "add instance" and "add cluster" should be merged to "describe network topology"

doc/usecases/elektra_web/UC_add_cluster.md
```diff
@@ -0,0 +1,20 @@
...
+Title: Add cluster
+Scope: Setup
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User groups multiple instances into a cluster.
+
+## Scenarios
+
+Precondition: Add instance (multiple times)
+Main success scenario: User successfully creates a cluster of multiple instances.
```
markus2330: Recursion missing.

doc/usecases/elektra_web/UC_cluster_config.md
```diff
@@ -0,0 +1,19 @@
...
+Scope: Configuration
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User configures a cluster (multiple instances at once) via the client.
+
+## Scenarios
+
+Precondition: Add cluster
+Main success scenario: User successfully changes configuration of the cluster.
+Alternative scenario: One of the instances is not online. The user is informed about the issue.
```
markus2330: Does this mean nothing happens here except a popup?

doc/usecases/elektra_web/UC_cluster_config.md
```diff
@@ -0,0 +1,19 @@
...
+Error scenario: Technical problems while persisting configuration to one of the instances. The user is informed about the issue.
```
markus2330: Again: What happens, are there any changes anywhere?

You could describe the behaviour also in "Postcondition".

doc/usecases/elektra_web/UC_multiple_config.md
```diff
@@ -0,0 +1,19 @@
...
+Scope: Configuration
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User configures multiple (standalone) instances via the client.
+
+## Scenarios
+
+Precondition: Add instance (multiple times)
+Main success scenario: User successfully changes configuration of multiple instances.
+Alternative scenario: Instance not online. The user is informed about the issue.
```
markus2330: Why is it possible to select offline nodes?

Better say: "A node got offline ..."

doc/usecases/elektra_web/UC_remove_cluster.md
```diff
@@ -0,0 +1,19 @@
+# Use Case: Remove cluster
```
markus2330: I thought the browser is directly connected to a clusterd? Maybe you need an alternative scenario if the user tries to do that?

doc/usecases/elektra_web/UC_remove_instance.md
```diff
@@ -0,0 +1,19 @@
+# Use Case: Remove instance
+
+## Summary
+
+Title: Remove instance
+Scope: Setup
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User removes an instance via the client.
```
markus2330: Generally remove instance or remove instance from cluster?

doc/usecases/elektra_web/UC_single_config.md
```diff
@@ -0,0 +1,19 @@
+# Use Case: Remote configuration of a single instance
```
markus2330: Is this the use case when browser is also connected to single instance?

doc/usecases/elektra_web/UC_add_cluster.md
markus2330: They could be interesting to:

1. define multiple groups if you use a cluster per application (all mail, web servers ...)
2. for failover (two clusterd for endpoints refer to other clusters, if one machine is gone, the system is still reachable)

Also only connecting to elektrad is essential, nodejs is a too heavy dependency for many systems.

Namoshek:
During review of the PR, I came across this:

>> omnidan:
>> the client actually doesn't access elektrad directly, clusterd acts as a kind of router that maintains instances (this makes it easy to have a client that can be accessed from various devices, otherwise I'd have to store the connection info for clusterd/elektrad instances in localstorage, cookies or something similar). this way all you have to provide in the client is the host of the clusterd instance.

> markus2330:
> Should also be discussed in use cases. If you also provide a lightweight implementation of clusterd, the dependency might be ok.

I think @omnidan's approach is good. If he opens `elektrad` servers to the web client directly, he also has to give authorization data to the web client (which is currently not the case - the client only holds authorization for the `clusterd` server). I also think that the hierarchical design allows for more security through firewalls. For example could the `clusterd` server be reachable for the public, but all other servers (`elektrad`) only through the intranet.

omnidan:
- merged "add cluster" and "add instance" into "describe network topology"
- described adding instances/clusters in a more high-level way
- clarified what happens when an instance in a cluster is not available
- removed "multiple config" because it just meant executing "single config" on multiple instances one after another (not a separate use case)
- added "undo/redo" use case

> I thought the browser is directly connected to a clusterd? Maybe you need an alternative scenario if the user tries to do that?
clusterd actually manages all clusters, I don't think it makes much sense to have recursion of clusterd's, because there are many issues with that approach (see @Namoshek's comment) It would also make the setup and maintainance a lot harder.

> Generally remove instance or remove instance from cluster?
generally removing instances

> Is this the use case when browser is also connected to single instance?
no, this is also done via clusterd

> add keys, change keys, find keys, arrays
this is part of the "single config" and "cluster config" use cases, maybe I should make a new "configuration" use case that further describes these operations?

> import keys copy&paste
what do you mean? there will be an interactive tree view where you can create, delete and modify keys (this is the single/cluster config use case)

> help system
not sure what is meant by that, the UI should be intuitive, if something is unclear I will add short info blocks or tooltips, there shouldn't be a need for a separate help system

markus2330:
> import keys copy&paste

I mean that existing keys can be copied around via shortcuts (ctrl+c, ctrl+v) or drag&drop.

> help system

I also mean help about the configuration itself. So metadata like "description" should be nicely displayed (e.g. in tool tips when hovering keys).

omnidan:
> I mean that existing keys can be copied around via shortcuts (ctrl+c, ctrl+v) or drag&drop.

Key contents can already be copied around via shortcuts, or do you mean whole sets of keys? Drag & drop for moving keys around should be possible.

> I also mean help about the configuration itself. So metadata like "description" should be nicely displayed (e.g. in tool tips when hovering keys).

Ah, yes I was going to use metadata to adjust the UI (displaying number fields in a special way, validating regex constraints, adding descriptions as tooltips, etc)

---

What is the best way to document these features? Should I make a "configuration" use case that "single config" and "cluster config" both make use of? The same configuration interface would be used for configuring single instances and clusters.

omnidan:
@markus2330 I updated the use cases, they should now describe the focus of elektra-web more (configuration through an interactive tree view).

markus2330:
There are still two comments without response.

doc/usecases/elektra_web/UC_remove_cluster.md
omnidan: clusterd actually manages all clusters, I don't think it makes much sense to have recursion of clusterd's, because there are many issues with that approach (see @Namoshek's comment) It would also make the setup and maintainance a lot harder.

doc/usecases/elektra_web/UC_single_config.md
omnidan: no, this is also done via clusterd

omnidan:
@markus2330 I already responded to them in the comment above: https://github.com/ElektraInitiative/libelektra/pull/999#issuecomment-260984107 - but for a better overview I also pasted the answers into the comment threads on the diff. Let me know if anything is still unclear.

markus2330:
I am still not convinced about problems with recursion. The main problem is if the same node is present multiple times, and this can happen regardless of recursion. Can you write a decision about why you decide to do the clustering the way you do and how to deal with conflicts and the same node added to multiple clusters? (Separate PR)

And I also would like to see the single node use case to be supported better. It is okay if two node instances need to be running, but nevertheless there should be a really easy way to set this up. There should be a dedicated use case for this thus I think it will be the first contact for everyone trying out your tool. "Setup instance" is really vague. Please describe what the user must do.

And please self-review the use cases once more to make them more clear. E.g. why is there remove cluster if we have a use case "describe network"? Many questions are unanswered: What can we describe? What does the user get from describing it? (Why would he/she want to do it? Most other use cases refer to "Add instance/cluster".) These are only examples, please check them thoroughly for consistency.

omnidan:
@markus2330 I updated the use cases and added a decision for the structure (#1173)

omnidan:
@markus2330 is there still anything missing here? I have created decisions for the (recursive) structure: #1173 #1354

markus2330:
Thus you (understandably and well argued) use simple techniques for conflict resolution and disallow inserting the same node multiple times as well as recursion; and use nodejs as only language for implementation (also understandable); please add more focus for usability. For example, there is still no use case about validation. (It is only very vaguely mentioned).

So for example, when `check/validation` defines a regex, you should also define it as input validation, without round-trips to the backend. We could also check in a study with and without this feature if users experience differences.

markus2330:
Maybe a clarification: Of course you should not write a use case to avoid round-trips (maybe the round-trip does not matter at all, then simply forget my implementation suggested). Instead the use cases should be something like "users tries to input an erroneous string; no character is added in the input field and user gets feedback..."

omnidan:
@markus2330 yes, I agree that validation should be done on the client (e.g. it would mark the input field as red and not send it to the backend at all)

I added a use case explaining validation.

doc/usecases/elektra_web/UC_key_validation.md
```diff
@@ -0,0 +1,25 @@
...
+
+Precondition: Modifying keys
+Main success scenario: User edits a key's contents in a field that is
+  specifically formatted according to its metadata (e.g. number/date fields vs
+  string fields with regex validation). If the entered data passes the
+  validation, it is sent to the backend and saved to the key database.
+Alternative scenario: User enters data that fails the validation, the input
+  field is marked as invalid (e.g. with a red color and some information text)
+  and the data is NOT sent to the backend.
+Error scenario: Technical problems while persisting to the key database. The
```
markus2330: Is this a copy&paste problem? For me it is not clear when this use case applies. Isn't it before modifying keys and/or persisting keys? (So the user only typing into input fields without pressing ok).

Usually there is not only an invalid state but also a state where validation might pass if the user continues typing.

omnidan: in a way it's part of the "modifying keys" use case. my idea was that it will keep validating and persist only when the input becomes valid.

markus2330: Yes, the idea is good, but this is not whats written in this use case. In the use case you have "modifying keys" as precondition. Actually it is the other way round: modification needs the key validation as precondition.

Ideally you think through the steps for a user again and compare it carefully with what currently is written.

omnidan: @markus2330 I swapped out the preconditions now.

markus2330: The issue is not only there. A general line in what the user can do when seems to be missing. For example, why is there no validation when "Adding keys"? Maybe you need a Use Case Diagram or some overview description to get this straight.

Please rethink the whole use cases from a users perspective, rebase, and let us finally merge this PR!

omnidan: @markus2330 I think the issue is that validation is supposed to be everywhere, it's not a separate use case, it's integrated in all other use cases.

omnidan: So maybe the best way to go about this is removing the validation use case and adapting all other use cases to describe the validation there?

markus2330: No, use cases will always somehow interrelate, you cannot copy everything (you copied too much, hence the inconsistency issues now). For example undo/redo also depends on most other use cases. You should rethink how they depend on each other and fix currently wrong precondition. Maybe a behavioral UML diagram helps you to do so. Or read a bit more about how to create use cases.

omnidan: Maybe I'm misunderstanding something, but while undo/redo is definitely its own use case (it's a separate feature that is triggered with certain buttons), validation is a feature that is integrated in all use cases that include text fields.

That's also what makes it hard to have a definitive order of preconditions - because validation happens until the field is valid, in which case it is stored on the server.

If I was gonna make it as a separate use case, I would have to copy & paste the same validation use case for various text fields, with little differences. So the best way to go about this would be explaining validation in the modifying keys use case, but also in the adding keys use case, and others.

markus2330: Maybe you are trying to say too much with the use cases? I only noticed several times that it is very easy to find inconsistencies and points where clarifications are needed.

For example, now that you say it: undo is supported after every other *use case* (e.g. finding keys)? (Rather it only applies to actions that changed the state of the system.)

In the same way the validation only applies when the user edits something in the tree view. It shouldn't be too hard to say that explicitly.

Basically I am okay with every description that:

1. clearly describes what the user can achieve with your proposed system
2. are internally consistent

omnidan: So you are proposing to make a generic validation use case with multiple use cases listed as precondition? but in that case, what would be the postcondition? would it be "continue with the post condition of the use case"? that's what confuses me, validation is neither before or after editing a field, it's something inbetween/an extension of use cases.

markus2330: Extension (and inclusion) is something that is quite well defined in UML use case diagram, thus I suggested you could use that.

There is more than one correct way to describe it. It would be better if you can stick to the standard, it should have more than enough possibilities to say what you want.

You do not have to be complete. But it should be clear and consistent.

markus2330: (You can also use UML features in the textual variant, if you do not want to draw something.)

omnidan: Thanks for the hints, I will look into the UML standard more and then rework all use cases accordingly.

markus2330: Thanks! It shouldn't take long, but you cannot fix it without going through the use cases completely.

markus2330:
Please remove "work in progress" when updated.

omnidan:
@markus2330 I reworked all the use-cases and used "extends" for key validation. I also removed everything related to clusters, it is still possible to have multiple instances, just not in clusters. I also split up some use-cases in the configuration tree view, as this is the focus of the application (not cluster configuration).

omnidan:
The path through the use cases is as follows:

- Setup
- Setup instance
- View config

Now we are in the main part of the application, and we can do the following use cases:

- Adding keys
- Modifying keys
- Duplicating keys
- Drag & Drop keys
- Finding keys

Any of these use cases (except for finding because no change is made), are extended by the *Key validation* use case. After doing any of these use cases, we can do the *Undo/Redo* use case to revert them.

doc/usecases/elektra_web/UC_setup.md
```diff
@@ -0,0 +1,21 @@
+# Use Case: Setup
```
markus2330: Please write a README.md containing the info of the path trough the use cases.

doc/usecases/elektra_web/UC_undo_redo.md
```diff
@@ -0,0 +1,22 @@
...
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User undos/redos changes to a configuration.
+
+## Scenarios
+
+Precondition: Modifying keys, Adding keys, Duplicating keys, Drag & Drop keys
+Main success scenario: User successfully undos/redos a configuration change.
+Alternative scenario: Instance not online. The user is informed about the issue.
+  If a cluster, changes will be written when the affected instance comes back
```
markus2330: Still clusters?

doc/usecases/elektra_web/UC_undo_redo.md
```diff
@@ -0,0 +1,22 @@
...
+	online.
+Error scenario: Technical problems while configuring the instance. The user is
+  informed about the issue.
+Postcondition: The updated configuration is persisted to the instance/s.
+Non-functional Constraints:
+	- Ideally every operation should be undoable
```
markus2330: Ideally?

doc/usecases/elektra_web/UC_duplicating_keys.md
```diff
@@ -0,0 +1,20 @@
+# Use Case: Duplicating keys
```
markus2330: Do we really need this as extra usecase? Isn't this copy&paste or ctrl+drag&drop?

doc/usecases/elektra_web/UC_duplicating_keys.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Scenarios
+
+Precondition: View configuration of an instance
+Main success scenario: User duplicates an existing key (and its subtree).
+Alternative scenario: User enters data that violates a validation rule, an error
+  message is shown.
+Error scenario: Technical problems while persisting to the key database. The
+  user is informed about the issue.
+Postcondition: The new key is persisted to the database.
```
markus2330: new key (and its subtree)

doc/usecases/elektra_web/UC_finding_keys.md
```diff
@@ -0,0 +1,20 @@
...
+Scope: Configuration
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User searches for a key by filtering the key database.
+
+## Scenarios
+
+Precondition: View configuration of an instance
+Main success scenario: User searches for a key by entering information (its path
+  or data) in a field. The whole key database is filtered according to the
```
markus2330: data -> value.

Search for key is much faster, what about first show by key and continue searching for values in background?

doc/usecases/elektra_web/UC_finding_keys.md
```diff
@@ -0,0 +1,20 @@
...
+  contents of the search field.
```
markus2330: No description what happens if (multiple) key(s) is/are found. Will it/they be shown in the tree?

doc/usecases/elektra_web/UC_setup.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Summary
+
+Title: Setup
+Scope: Setup
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User sets up elektra-web according to the corresponding README.
```
markus2330: Why not a script? Too detailed for a use case.

doc/usecases/elektra_web/UC_setup_instance.md
```diff
@@ -0,0 +1,25 @@
...
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User sets up a daemon on the instance and connects it to elektra-web.
+
+## Scenarios
+
+Precondition: Setup
+Main success scenario: User successfully sets up elektrad on the instance (by
+	installing the dependencies and starting elektrad) and connects it to
+	elektra-web (clusterd) by specifying the host and an API key in a config file.
```
markus2330: rename clusterd?

doc/usecases/elektra_web/UC_duplicating_keys.md
omnidan: @markus2330 I was going to add an icon to duplicate keys, because this will be a pretty common use case (e.g. creating a new student with the same fields)

doc/usecases/elektra_web/UC_setup_instance.md
omnidan: @markus2330 I have already implemented cluster configuration in the backend, just decided to remove it in the frontend to keep the focus on the editor. I would keep the name because it still provides that functionality, but if you have a better idea I can also rename it :smile:

doc/usecases/elektra_web/UC_finding_keys.md
omnidan: As described, the search field works like a filter, so all keys that match the search input will be shown, the other keys will be hidden. I described this more clearly now.

omnidan:
@markus2330 should be fine now!

omnidan:
I just realized that the markdown renderer does not put line breaks unless you make two in the file. This is also a problem with the other use cases, e.g. https://github.com/ElektraInitiative/libelektra/blob/master/doc/usecases/snippet_sharing/UC_convert_snippet.md

We could either append `<br>` after the lines, or use two line breaks instead of one.

doc/usecases/elektra_web/UC_undo_redo.md
```diff
@@ -0,0 +1,20 @@
...
+Scope: Configuration
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User undos/redos changes to a configuration.
+
+## Scenarios
+
+Precondition: Modifying keys, Adding keys, Duplicating keys, Drag & Drop keys
+Main success scenario: User successfully undos/redos a configuration change.
+Alternative scenario: Instance not online. The user is informed about the issue.
```
markus2330: Seems to be a copy&paste error from earlier versions? Instances going offline would be an error scenario?

doc/usecases/elektra_web/UC_setup_instance.md
```diff
@@ -0,0 +1,25 @@
...
+
+Title: Setup instance
+Scope: Setup
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User sets up a daemon on the instance and connects it to elektra-web.
+
+## Scenarios
+
+Precondition: elektra-web is set up
```
markus2330: We are already finished before starting?

doc/usecases/elektra_web/UC_setup_instance.md
```diff
@@ -0,0 +1,25 @@
...
+Main success scenario: User successfully sets up elektrad on the instance (by
+	installing the dependencies and starting elektrad) and connects it to
+	elektra-web (clusterd) by specifying the host and an API key in a config file.
```
markus2330: Better to drop the name `clusterd` and simply call it elektra-web everywhere.

doc/usecases/elektra_web/UC_setup_instance.md
```diff
@@ -0,0 +1,25 @@
...
+Title: Setup instance
+Scope: Setup
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User sets up a daemon on the instance and connects it to elektra-web.
+
+## Scenarios
+
+Precondition: elektra-web is set up
+Main success scenario: User successfully sets up elektrad on the instance (by
```
markus2330: As already said earlier: too many details, keep the use cases high-level. Remove whats written in () and in the following sentence (API key, ...).

doc/usecases/elektra_web/UC_modifying_keys.md
```diff
@@ -0,0 +1,25 @@
...
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User modifies a key's contents.
+
+## Scenarios
+
+Precondition: View configuration of an instance
+Main success scenario: User edits a key's contents in a field that is
+  specifically formatted according to its metadata (e.g. number/date fields vs
+	string fields with regex validation). Arrays are specifically formatted as
```
markus2330: tab space mixture?

doc/usecases/elektra_web/UC_modifying_keys.md
```diff
@@ -0,0 +1,25 @@
...
+	lists of fields under a key. A description of the key is also shown.
+Alternative scenario: User enters data that fails the field validation, the
+  input field is marked as invalid (e.g. with a red color and some information
+  text) and the data is NOT sent to the backend.
```
markus2330: Use markdown formatting instead of all-caps?

doc/usecases/elektra_web/UC_modifying_keys.md
```diff
@@ -0,0 +1,25 @@
...
+Error scenario: Technical problems while persisting to the key database. The
+  user is informed about the issue.
```
markus2330: The word "problem" seems to fit better than "issue" here.

omnidan:
@markus2330 sorry it took so long to rework again - I fixed the formatting issues and reworked the texts again.

I hope everything is fine now :grin:

doc/usecases/elektra_web/UC_duplicating_keys.md
markus2330: I somehow doubt that it is a common operation outside of the student use case (which is not relevant anymore). I rather think that common use cases are:

- add array member (`#N+1`)
- add a key within a specific namespace for which the specification is given (but no key exists)

doc/usecases/elektra_web/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Now we are in the main part of the application, and we can do the following use cases:
+
+- [Adding keys](./UC_adding_keys.md)
+- [Modifying keys](./UC_modifying_keys.md)
+- [Duplicating keys](./UC_duplicating_keys.md)
+- [Drag & Drop keys](./UC_drag_n_drop.md)
+- [Finding keys](./UC_finding_keys.md)
+
+Any of these use cases (except for finding because no change is made), are extended by the [Key validation](./UC_key_validation.md) use case. After doing any of these use cases, we can do the [Undo/Redo](./UC_undo_redo.md) use case to revert them.
```
markus2330: Good intro!

doc/usecases/elektra_web/UC_adding_keys.md
```diff
@@ -0,0 +1,21 @@
+# Use Case: Adding keys
+
+## Summary
+
+Title: Adding keys
```
markus2330: Formatting still broken. An itemization with `-` might be a good solution.

doc/usecases/elektra_web/UC_adding_keys.md
```diff
@@ -0,0 +1,21 @@
...
+Scope: Configuration
+Level: User Goal
+Actors: User (usually a sysadmin)
+Brief: User adds a key (to a subtree).
+
+## Scenarios
+
+Precondition: View configuration of an instance
```
markus2330: Sentences should have an dot `.` at the end.

doc/usecases/elektra_web/UC_finding_keys.md
```diff
@@ -0,0 +1,24 @@
...
+Actors: User (usually a sysadmin)
+Brief: User searches for a key by filtering the key database.
+
+## Scenarios
+
+Precondition: View configuration of an instance
+Main success scenario: User searches for a key by entering information (its path
+  or value) in a field. The whole key database is filtered according to the
+  contents of the search field. At first, only keys that match the path are
+  shown. Then, the application will continue searching for values that contain
```
markus2330: enumeration/itemize would make it easier to read

doc/usecases/elektra_web/UC_modifying_keys.md
```diff
@@ -0,0 +1,25 @@
...
+  string fields with regex validation). Arrays are specifically formatted as
+  lists of fields under a key. A description of the key is also shown.
+Alternative scenario: User enters data that fails the field validation, the
+  input field is marked as invalid (e.g. with a red color and some information
+  text) and the data is *not* sent to the backend.
+Error scenario: Technical problems while persisting to the key database. The
+  user is informed about the problem.
+Postcondition: The updated key is persisted to the database.
+Non-functional Constraints:
+  - Essential functionality
```
markus2330: is this a non-functional constraint?

doc/usecases/elektra_web/UC_setup_instance.md
```diff
@@ -0,0 +1,23 @@
...
+  connects it to elektra-web.
+Alternative scenario: Authentication issues when connecting the services,
+  e.g. wrong API key.
+Alternative scenario: The instance is already connected to elektra-web. A
+  message informs the user about this.
+Error scenario: Technical problems - the user is informed about the problem.
+Postcondition: User can now access the instance via the client.
+Non-functional Constraints:
+  - Security mechanism
+  - Is precondition to use configuration functions
```
markus2330: is this a non-functional constraint? (or a precondition of the others)

omnidan:
I fixed the issues mentioned in the comments and reformatted all usecases. See #1550 for the PR with the updated template file.

markus2330:
Looks much better! Please add "ready to merge" if ready to merge.

omnidan:
@markus2330 done! :ok_hand:

Issue #682, opened 2016-04-30
Translation of qt-gui


markus2330:
Currently we only have a german translation:
`src/tools/qt-gui/qml/i18n/lang_de_DE.ts`
It would be great to have other languages, too.

sirblackheart:
Probably we should use [zanata](https://translate.zanata.org/) once it [supports](https://zanata.atlassian.net/browse/ZNTA-689) ts files.

markus2330:
Do you think it could attract currently uninvolved people? Then lets go for it ;) 

Otherwise, there are many offline tools for translation tasks.

sirblackheart:
I do not think that it will get more developers directly involved with elektra, but i do think that the barrier of offline translation tools will keep possible translators away. Here comes the question if we assume that a possible translator of the project would do more then translate and get involved in other ways with the project and if we want to keep the offline tool barrier for that reason. On the other hand offering a web interface for simple translation does lower the barrier for people with basic technical skills but good translation skill to translate the application.
Zanata is just one proposal for going that way and it sill does not support ts files yet. [Transifex](https://www.transifex.com/) for example would also be free for open source projects and already would support [qt .ts](http://docs.transifex.com/formats/qt-ts/).

markus2330:
I mean involvement in any way, such as translating it or also using it. If it lowers the barrier, it would be good to have it!

markus2330:
We now have some translations. It is low priority to add more translations.

PR #1018, opened 2016-10-11
Added Spanish Translation of qt-gui #682

pixelead0:
# Purpose

Added Spanish Translation of qt-gui #682 
# Checklist

needs to be checked.
- [ x ] i18n spanish

### DIFF
--- a/src/tools/qt-gui/qml/i18n/lang_es.ts
+++ b/src/tools/qt-gui/qml/i18n/lang_es.ts
@@ -1,36 +1,36 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="de_DE">
+<TS version="2.1" language="es">
 <context>
     <name>AboutWindow</name>
     <message>
         <source>About Elektra Editor</source>
-        <translation type="vanished">Über Elektra Editor</translation>
+        <translation type="vanished">Acerca de Elektra Editor</translation>
     </message>
     <message>
         <location filename="../AboutWindow.qml" line="8"/>
         <source>About Elektra Qt Editor</source>
-        <translation>Über Elektra Qt Editor</translation>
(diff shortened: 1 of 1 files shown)


markus2330:
Thank you for creating the translation! That is a very nice contribution!

Did you try to run qt-gui with the translation enabled? I think we also need to add the qm file and add it to i18n.qrc. @0003088 is this correct?

@0003088 Is there any specific reason why the german translation is called "de_DE" and not simply "de"? I think @pixelead0 is right here and we should use the most general locale name as long as we do not distinguish.

0003088:
@markus2330 As far as I can remember you are correct. I do not think that I chose 'de_DE' "for fun" so to speak, but that I followed a convention by doing so. @pixelead0 I really can recomment qt's Translator tool, it doesn't get any easier than that.

mpranj:
As for the locale, the 2-letter code is sufficient. The .qm file definitely needs to be produced with lrelease and then added to the .qrc resource file.

In fact, I [think](http://doc.qt.io/qt-5/qtranslator.html#load) the de_DE locale would not be loaded for environments with locale de_AT.

markus2330:
Yes, I think @mpranj is right here. For me (de_AT locale) the (de_DE) also was not found. It seems like that QTranslator::load as we do it is not recommended anymore. In the long run we also should generate the qm files and not have them checked in.

@pixelead0 is there anything we can help you with? Are there any issues with creating the qm file and add it to i18n.qrc? Or do you have troubles starting up the gui?

markus2330:
@pixelead0 Hi, any news here? You are getting late for the Hacktoberfest!

markus2330:
@pixelead0 Can I merge it for the upcoming release?

markus2330:
@pixelead0 Would be great if I can add it to 0.8.20 release.

markus2330:
Thank you!

Issue #1021, opened 2016-10-13
Plugins: add storage format and clarify capabilities in status


Namoshek:
# Situation

Currently `storage` plugins are either named by their format (e.g. `ini` plugin) or have the format as part of their `info/provides` list. Because this list is not consistently structured, it is impossible to know which provides key represents the actual format supported by the plugin.

Furthermore, some plugins cause troubles when being fed with arbitrarily structured data because they can only operate on a well-defined structure (e.g. `fstab`, `hosts`).

This is the implementation task following the discussion in #994.

# Requirements

It should be possible
- to find plugins based on formats (already possible).
- to enumerate all plugins with their respective format.
- to know whether a plugin is limited in certain terms (e.g. cannot represent arbitrary data, can only read or only write).

# Implementation

- `storage` providers that do not support an actual format will be renamed to `storage/info`.
- `storage` providers that support a format will get the format as value for `storage`, e.g. `storage/ini` (separator to be discussed in #1010).
 - with help of CMake also `storage` and `ini` will then be added to the provides list to keep backwards-compatibility.
- additionally to the format, `storage` plugins will get
 - the status `limited` if they cannot consume arbitrarily formatted data (e.g. `hosts` plugin).
 - the status `readonly` if they can only read files.
 - the status `writeonly` if they can only write files.

markus2330:
Very clearly written, please also add this text to the documentation of infos/status and infos/provides (in doc/CONTRACT.ini).

One point is missing: in the PluginDatabase (src/libs/tools/src/plugindatabase.cpp) you need to do the split-up using the seperator, too. (in the same way as done in CMake).

Namoshek:
If I already need to modify the CMake function, it might be easier doing both there. 

Did you btw. already decide whether we should use a list or an array for the provides list? (An array would probably require quite some changes in `plugindatabase.cpp` too?)

markus2330:
> If I already need to modify the CMake function, it might be easier doing both there. 

I am not sure about that ;)

> Did you btw. already decide whether we should use a list or an array for the provides list? (An array would probably require quite some changes in plugindatabase.cpp too?)

provides? It is not a list, nor an array, it is a space-separated entry in the contract. Do you mean examples? Please create separate issues for separate questions.

Namoshek:
> I am not sure about that ;)

[...]

PR #1022, opened 2016-10-13
improvement: format for storage plugins, status for limited plugins

Namoshek:
# Purpose

Description of PR content in #1021.

# Checklist


Namoshek:
Do we also need changes for `infos/status` in [CONTRACT.ini](https://github.com/ElektraInitiative/libelektra/blob/master/doc/CONTRACT.ini) (and thus also in [plugindatabase.cpp](https://github.com/ElektraInitiative/libelektra/blob/master/src/libs/tools/src/plugindatabase.cpp))?

Namoshek:
Ok, failing tests say we need to add the new statuses to the contract. Should we use negative values for `limited`, `readonly` and `writeonly` or, because they are normally only used on very distinct plugins, a zero value?

markus2330:
Intuitively I would use negative values (if a plugin providing the same format does not have the restriction, it should be preferred). The values are currently not fine-tuned, so if you have ideas how to get a better order of plugins (you also should use this order for the drop-down in the interface), please propose it.

Namoshek:
The thing with `readonly` and `writeonly` is that the plugin can be quite nice, but just do not offer this functionality. If a plugin supports a certain task or not has to be checked anyway, so I wouldn't give them a negative value.

`limited` on the other hand makes a plugin definitely worse, so I would give it a value somewhere between -500 and -2000. (The only status that seems somewhat equal would be `unfinished` with -1000 for important functionality that is still missing.)

markus2330:
> limited, readonly and writeonly

Yeah, it is quite subjective. I would argue the other way round (limited is the purpose of what it needs to fulfil, thus zero; *only is a implemention not yet done, thus minus points).

In the end these points are minor things, so maybe "maintained", "concept", "orphan" and "obsolete" should be rated much higher.

Namoshek:
Well, `limited` says you that it cannot represent arbitrary data, so I would consider it a bad storage plugin to use for random data and thus give minus points. `*only` does certainly also prevent a plugin from being a good common storage plugin, but I wouldn't argue with "not yet implemented", because e.g. `constants` is read-only on purpose.

For the moment I've chosen -750 for `limited` and 0 for `*only`, to get the test errors gone.

Any idea why `testscr_check_real_world` fails? I've not changed anything that would explain it to me...

Namoshek:
Any idea why the test fails, @markus2330?

markus2330:
Yes: because uname is not storage anymore, the mount tool adds another storage (dump) which fails badly. Seems like we did not consider this case.

What about using `storage/info` for info plugins? Then mounting will be happy, and you can filter for info plugins. (Btw. they are quite nice in the snippet converter because they generate configuration without anything to type). The alternative would be a teaching `kdb mount` that `info` also satisfies the `storage` requirement.

Namoshek:
Oh, okay. Yeah, seems like using `storage/info` is a good solution then. But currently they don't give any output when used within the converter (they always fail). Maybe because I give some (wrong) input to them? Or do they always need to be mounted to work?

markus2330:
Did you change the port of the frontend (namoshek.at:9000)? I can only reach the REST service, but the web page does not load.

> But currently they don't give any output when used within the converter (they always fail).

For me it always did work. On the left side you could select a info storage (e.g. uname or constants) and on the right side which format you want the output to be displayed in.

>  Maybe because I give some (wrong) input to them?

input should be irrelevant. (graying out the left box for info plugins might help users though). Info plugins obviously only work on the left side, but this should be fixed once you support readonly and writeonly. (Then readonly can only be selected left, writeonly only on the right side)

> Or do they always need to be mounted to work?

No, there is nothing special about mounting, mounting is only a description how the plugin should be instanciated and what its parent key is.

Namoshek:
> For me it always did work. On the left side you could select a info storage (e.g. uname or constants) and on the right side which format you want the output to be displayed in.

I thought the other way round, but of course you are right, the `info` plugins are read-only.

> Did you change the port of the frontend (namoshek.at:9000)? I can only reach the REST service, but the web page does not load.

No, it works for me. Maybe an issue with JavaScript?

> input should be irrelevant. (graying out the left box for info plugins might help users though). Info plugins obviously only work on the left side, but this should be fixed once you support readonly and writeonly. (Then readonly can only be selected left, writeonly only on the right side)

Good point, this requires a change of the API description, backend and frontend. I'll see what I can do.

Namoshek:
can you please have a look and merge if it's ok @markus2330?

markus2330:
Yes, looks fine! Is it okay to keep it open until thursday? Maybe someone wants to review it.

Namoshek:
Yes, it is ok from my side. But then I'll not push the changes to the rest-backend until then, because I'd like to rebase first.

markus2330:
Cmake code is not so interesting to review anyway, the review should not hinder you in your work!

Thanks for the fix!

PR #1027, opened 2016-10-16
csvstorage: fixed tests overwriting testfiles + bogus testcase

tom-wa:
forgot to change the parent key value after reading the file, before calling kdbSet


markus2330:
Thank you!

PR #1078, opened 2016-11-11
cachefilter plugin update

Namoshek:
# Purpose

Show problems with cachefilter plugin.

# Test Case

```
$> kdb mount users.dump dir/users dump
$> kdb mount configs.dump dir/configs dump
$> kdb global-mount cachefilter

$> kdb shell
> kdbGet dir/users
return value: 0
> kdbGet dir/configs
return value: 0
> ksOutput
> keySetName dir/users/Namoshek
> keySetString password
> ksAppendKey 
> ksOutput
dir/users/Namoshek string: password
> kdbSet dir/users/Namoshek
return value: 1
> ksOutput
dir/users/Namoshek string: password
> ksCut dir/users
> ksOutput
> keySetName dir/configs/some/test/config/path
> keySetString some-conf-value
> ksAppendKey
> ksOutput
dir/configs/some/test/config/path string: some-conf-value
> kdbSet dir/configs/some/test/config/path
return value: 1
> ksCut dir/configs
> ksOutput
> kdbGet dir/users
return value: 0
> ksOutput
dir/users/Namoshek string: password
> ksCut dir/users
> kdbGet dir/configs
return value: 0
> ksOutput
dir/configs/some/test/config/path string: some-conf-value
> ksCut dir/configs
> ksOutput

$> kdb ls dir
dir/configs/some/test/config/path
dir/users/Namoshek
1 Warning was issued:
 Warning number: 79
	Description: Postcondition of backend was violated
	Ingroup: kdb
	Module: 
	At: /home/marvin/Schreibtisch/libelektra-improvements/src/libs/elektra/split.c:501
	Reason: drop key dir/users/Namoshek not belonging to "dir/users" with name "dir/users" but instead to "dir/configs" with name "dir/configs" because it is hidden by other mountpoint
	Mountpoint: dir
	Configfile: /home/marvin/Schreibtisch/libelektra-improvements/build/.dir/users.dump
```

The test case basically reflects what my REST service is doing, with the difference that successive kdbGet calls do not return anything in the latter (probably because the plugin is not called... for whatever reason).

As you can see above, keys are also written to the wrong mountpoint when the whole cache is appended during kdbSet. I also noticed that my global plugin list gets always reset when working with mount commands (maybe a side-effect of my plugin?).

### DIFF
--- a/src/libs/elektra/kdb.c
+++ b/src/libs/elektra/kdb.c
@@ -753,6 +753,10 @@ int kdbGet (KDB * handle, KeySet * ks, Key * parentKey)
 	{
 	case 0: // We don't need an update so let's do nothing
 		keySetName (parentKey, keyName (initialParent));
+		if (handle && handle->globalPlugins[POSTGETSTORAGE])
+		{
+			handle->globalPlugins[POSTGETSTORAGE]->kdbGet (handle->globalPlugins[POSTGETSTORAGE], ks, parentKey);
+		}
 		splitUpdateFileName (split, handle, parentKey);
 		keyDel (initialParent);
 		splitDel (split);
--- a/src/plugins/cachefilter/cachefilter.c
+++ b/src/plugins/cachefilter/cachefilter.c
@@ -34,44 +34,54 @@ int elektraCachefilterGet (Plugin * handle, KeySet * returned, Key * parentKey)
 
 	// initialize working KeySets
 	KeySet * toBeCached = 0;
-	KeySet * temp = 0;
 
 	// get old cache data
-	void * cachedData = elektraPluginGetData (handle);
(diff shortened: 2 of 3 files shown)


markus2330:
POSTGETSTORAGE has weird semantics: @tom-wa what was the idea here?. They are executed for in line 600 for only one backend (but similar to foreach semantics), thus keys are injected in wrong backends. But in other places (error condition) it has deinit semantics. This mixture makes it unusable.

For this PR you could remove the line 600, and fix the hook for your needs. This would also be valuable input for @mpranj which requirements the hook positions have.

markus2330:
I pushed this together with  #874 in the new branch cachefilter. The testcase above should work, but their are issues with #874

PR #1098, opened 2016-11-15
Gitignore improvements

e1528532:
# Purpose

I had the issue that i had a lot of cmake generated files in my git, and some mac specific files, so i've used http://www.gitignore.io to ignore some common build or OS artifacts which are not useful in a repository.
I deleted two autogenerated entries, namely *.map and MAKEFILE, as the project contains some occurences of such files so those could probably have caused some issues.

# Checklist

needs to be checked.

- [X] commit messages are fine (with references to issues) (there are no issues yet)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,7 +103,7 @@ Temporary Items
 
 # Linker output
 *.ilk
-*.map
+# *.map
 *.exp
 
 # Precompiled Headers
@@ -169,7 +169,6 @@ dkms.conf
 CMakeCache.txt
 CMakeFiles
 CMakeScripts
-Makefile
 cmake_install.cmake
 install_manifest.txt
 CTestTestfile.cmake
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,166 @@ build*
 
 #ignore vim swp files
 *.swp
+
+
+# The entries until the end of the file are created by https://www.gitignore.io/api/windows,linux,macos,c,c++,cmake,java
+
+### Windows ###
+# Windows image file caches
+Thumbs.db
+ehthumbs.db
+
+# Folder config file
+Desktop.ini
+
+# Recycle Bin used on file shares
+$RECYCLE.BIN/
+
+# Windows Installer files
+*.cab
+*.msi
+*.msm
+*.msp
+
+# Windows shortcuts
+*.lnk
+
+
+### Linux ###
+*~
+
+# temporary files which can be created if a process still has a handle open of a deleted file
+.fuse_hidden*
+
+# KDE directory preferences
+.directory
+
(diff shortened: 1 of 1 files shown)


Namoshek:
Are you using a build directory?

e1528532:
Yes i am, its called simply build and is directly in the elektra folder.
I definately have some mac specific files (.DS_Store) in my elektra folder, these are auto-generated by mac so i think that we could benefit from the OS-specific entries at least. I also had some java class files lying around because i compiled the jna bindings example like described on the bindings page, so i think ignoring these should also be fine.
The things i'm most concerned of are the C/C++/Cmake parts because i have limited knowledge yet of which files are required by our build process.

markus2330:
Thank you for creating this PR!

I happily accept every .gitignore entry someone needs (and wanting to add them is unlikely), but this PR seems to contain files for which valid reasons to add them exist. In particular I could imagine someone wants to commit one of these many preferences files listed in the PR. Having gitignore of files that you actually wanted to commit is also a bit of danger (oversight). There also seem to be some very generic entries, like `._*` I do not really understand what they are for.

Adding something like `.DS_Store` and `*.class` goes without saying. For the others I would prefer if we had some rationale.

> of which files are required by our build process

The build process does not modify or create any file of/in the source directory. So we only need gitignore for files tools create.

markus2330:
Btw. which is the license of the output from gitignore.io?

sanssecours:
> Adding something like `.DS_Store` and `*.class` goes without saying.

In my opinion these files should be part of a users [global `.gitignore` file](http://stackoverflow.com/questions/7335420/global-git-ignore).

markus2330:
Yes, they can be added in other ways (there is also .git/info/exclude) but having them in .gitignore might lead to better out-of-the-box experience for new developers.

And is there any reason for adding `.DS_Store` and `*.class` dirs/files?

e1528532:
i think sanssecours opinion makes sense, i can't think of a reason why you'd want to add .DS_Store and *.class files to a project, so globally ignoring them might makes sense (though everyone has to do this on his own) and avoids potential issues. So i guess we can drop this?

markus2330:
In the last meeting we discussed it a bit and we agreed that it is useful to add files to .gitignore someone is annoyed by. So you can add .DS_Store and *.class.

> (though everyone has to do this on his own)

Yes exactly. So its better to have it here, it provides a better out-of-the-box experience.

Note that `.gitignore` can also be used in subdirectories, so `.class` might be better placed where the java binding lives.

Please rebase (or recreate the branch/PR, might be easier for this simple change) and add only parts you are actually affected by.

e1528532:
created a new branch for this, so closing this pull request here

Issue #666, opened 2016-04-28
finalize infos/status


markus2330:
## add in CONTRACT.ini

- [ ] finish #4758
- [ ] example .. plugin is only an example how plugins should be (doc+template), see  #1012
- [ ] remove nodep and use dependencies/ instead

## also add automatic check

- [ ] documentation
- [ ] environment

## check features of plugins manually

- [ ] read
- [ ] write
- [ ] spec/*/*
- [ ] preserves/order
- [ ] preserves/comment
- [ ] preserves/comment/empty
- [ ] preserves/indentation
- [ ] limited
- [ ] arbitrary/keyset
- [ ] arbitrary/metadata
- [ ] directory/value
- [ ] directory/holes
- [ ] type
- [ ] array
- [ ] format/array
- [ ] format/hierarchy
- [ ] format/type
- [ ] format/human/read
- [ ] format/human/write

markus2330:
Maybe we also can find automated ways here?

markus2330:
@Namoshek Would nocomments and noorder be useful for the web frontend?

Namoshek:
Are not (almost) all plugins `noorder`? I thought during import, when keys are put into a keyset, they get automatically sorted in alphabetical order?

I think they are not really useful/necessary, because on the one hand, the fact that comments are missing is immediatly visible and on the other hand, it should be clear that not all formats do even support comments (e.g. json). So I wouldn't really use this information.

But you brought the idea up to me to just pass all providers through the REST API to the user. Then additions to the plugin information would only require a change in the frontend in future, not to the API description and the backend anymore.

markus2330:
> Are not (almost) all plugins noorder? I thought during import, when keys are put into a keyset, they get automatically sorted in alphabetical order?

Yes, most plugins are noorder and nocomment. Exceptions are hosts and ini.

> not all formats do even support comments (e.g. json)

Yes, that is similar to readonly/writeonly/limited: sometimes only implementation is missing, sometimes it is a wanted limitation.

> But you brought the idea up to me to just pass all statuses through the REST API to the user. Then additions to the plugin information would only require a change in the frontend in future, not to the API description and the backend anymore.

Good idea!

KurtMi:
Should I add those?

markus2330:
Could you first investigate which of ìnfos/status`(also of those we already have) we can determine automatically easily (within your tool)?

KurtMi:
[...]

PR #1107, opened 2016-11-17
scripts: update-info-status: nodoc

KurtMi:
Does the nodoc part of the devils issue #666


markus2330:
Thank you! We should consider to have a policy so that update-info-status only adds or removes flags if it is quite sure, see #1110

Issue #1106, opened 2016-11-17
improve readme of python and lua plugins


Namoshek:
The readmes of the [python](../blob/master/src/plugins/python/README.md) and the [lua](../blob/master/src/plugins/lua/README.md) plugins could need some love. The tests help in understanding how they work, but a comprehensive description with a good example is missing.

markus2330:
Thanks for reporting! At least they are correctly labeled as "nodoc".

PR #1109, opened 2016-11-17
readme improvements for lua & python plugin

Namoshek:
# Purpose

Improve plugin readme of lua and python plugin. This PR should solve #1106.

### DIFF
--- a/src/plugins/lua/README.md
+++ b/src/plugins/lua/README.md
@@ -48,6 +48,35 @@ Access to **kdb** can be retrieved using the Lua import
 
     require("kdb")
 
+## Example ##
+
+An example script that prints some information for each method call would be:
+
+    function elektraOpen(config, errorKey)
+	    print("Lua script method 'elektraOpen' called")
+	    return 0
+    end
+
+    function elektraGet(returned, parentKey)
+	    print("Lua script method 'elektraGet' called")
+	    return 1
+    end
+
+    function elektraSet(returned, parentKey)
+	    print("Lua script method 'elektraSet' called")
+	    return 1
+    end
+
+    function elektraError(returned, parentKey)
+	    print("Lua script method 'elektraError' called")
+	    return 1
+    end
+
+    function elektraClose(errorKey)
+	    print("Lua script method 'elektraClose' called")
+	    return 0
+    end
+
 ## Disclaimer ##
 
 Note, this is a technical preview. It might have severe bugs
--- a/src/plugins/lua/README.md
+++ b/src/plugins/lua/README.md
@@ -77,10 +77,13 @@ An example script that prints some information for each method call would be:
 	    return 0
     end
 
+Further examples can be found in the [lua](lua/) directory.
+
 ## Disclaimer ##
 
 Note, this is a technical preview. It might have severe bugs
 and the API might change in the future.
 
 Be also aware that a Lua script will never be as performant as a native C/C++ plugin.
+Spinning up the interpreter takes additional time and resources.
 
--- a/src/plugins/lua/README.md
+++ b/src/plugins/lua/README.md
@@ -4,22 +4,53 @@
 - infos/provides =
 - infos/needs =
 - infos/placements =
-- infos/status = maintained unittest configurable global memleak nodoc
-- infos/description = magic things require magic plugins
+- infos/status = maintained unittest configurable global memleak
+- infos/description = proxy that calls other plugins (scripts) written in lua
 
 ## Introduction ##
 
-The plugin uses Lua to do magic things.
+The plugin uses Lua to do magic things. It basically allows to call plugins written in Lua.
+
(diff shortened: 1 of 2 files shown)


src/plugins/lua/README.md
```diff
@@ -5,21 +5,82 @@
...
-The plugin uses Lua to do magic things.
+The plugin uses Lua to do magic things. It basically allows to call plugins written in Lua.
+
+What a Lua script can do is not really limited by design, so any kind of plugin may be
+implemented. The lua plugin is especially useful to write filter and logging scripts.
 
 ## Usage ##
 
-    kdb mount file.ini /lua ini lua script=/path/to/lua/lua_filter.lua
+The lua plugin accepts only the **script** configuration parameter holding the path to a Lua
```
markus2330: Can you mention here that tree.libelektra.org/src/plugins/lua/lua/ contains examples?

src/plugins/lua/README.md
```diff
@@ -5,21 +5,82 @@
...
+script. The mount command would look like
+
+    kdb mount file.ini /lua ini lua script=/path/to/filter_script.lua
```
markus2330: `kdb check -c script=/path/to/filter_script.lua lua` might also be worth mentioning

src/plugins/lua/README.md
```diff
@@ -5,21 +5,82 @@
...
+- elektraOpen(config, errorKey)
+- elektraGet(returned, parentKey)
+- elektraSet(returned, parentKey)
+- elektraError(returned, parentKey)
+- elektraClose(errorKey)
+
+where *config* & *returned* are KeySets and *errorKey* & *parentKey* are Keys.
+For the return codes of the functions, the same rules as for normal plugins apply.
+
+If a function is not available, it simply is not called. A script does not have to
```
markus2330: Is this also true for get?

src/plugins/python/README.md
```diff
@@ -5,22 +5,86 @@
...
+            print("Python script method 'error' called")
+            return 1
+
+        def close(self, errorKey):
+            print("Python script method 'close' called")
+            return 0
+
+## Disclaimer ##
+
+Be aware that a Python script will never be as performant as a native C/C++ plugin.
```
markus2330: (Especially spinning up the interpreter is quite expensive) Also noteworthy is that the python interpreter does not clean up and is not restartable.

src/plugins/lua/README.md
Namoshek: As far as I've seen all functions use the same helper that does return normally when the function is not implemented, but I'll give it another look. Maybe I've overlooked something, because one test that does only implement open is supposed to fail.

src/plugins/lua/README.md
Namoshek: Will add it. I think these plugins could also be worth being mentioned in a plugin tutorial as "first step to get familiar".

src/plugins/python/README.md
Namoshek: Thanks for that info, will definitely add it.

src/plugins/lua/README.md
Namoshek: I already thought about doing so, but wasn't sure if they are self-explanatory enough (as they are meant to be test scripts). I can add a link, yes.

src/plugins/lua/README.md
Namoshek: Why is `kdb check` important here in particular?

src/plugins/lua/README.md
Namoshek: `kdb check` exits successfully on an empty script file, so I guess it works also without `elektraGet`?

markus2330:
@Namoshek Could you please rebase?

Namoshek:
Yes, I'll do so right away. My VM crashed yesterday after installing some updates and now my GUI is missing - have to fix that first.

Namoshek:
Can you answer me the question I asked in the review above, @markus2330 (think it got lost in other notifications)? Then I can finalize the PR!

src/plugins/lua/README.md
markus2330: The mounting command sometimes produces confusing error messages, more parts are involved. If you want to check if kdbOpen works and the yielded contract makes somewhat sense, `kdb check` is much easier to use!

src/plugins/lua/README.md
markus2330: Not sure about this, might be a bug.

markus2330:
Sorry, yes I only saw your first response. I am not sure about your second question, you can leave it out for the tutorial if you cannot find it out.

Namoshek:
I think I'll leave it out as it seems a very generic hint/trick. From my side you can merge if you've no further comments.

markus2330:
Thank you, very nicely done! Good to have some documentation there now.

Issue #887, opened 2016-08-17
The Order Preserving Minimal Perfect Hash Map Build Integration


KurtMi:
The Order Preserving Minimal Perfect Hash Map (`OPMPHM`) is a very special kind of hash map. In Order to make a KeySet ready for lookups a build process is need. This build process in mainly divided in three steps. This Issue describes the needed tasks to Integrate the `OPMPHM` build process in Elektra.

### General
- [x] think about reusability of code

### Mapping
- [x] add ratio param
- [ ] add ksLookup flag for explicit hash build
- [x] add random function
- [x] add hash function
- [x] add graph debug tool graphviz as soft dept
- [ ] init a fast and nice bipartite graph data structure
- [ ] test cases
- [x] check heap function typdef
- [ ] try to find a more platform independent random function
- [ ] try to find a more platform independent seed init function
- [ ] fix  endian at hashfunction
### Order
- [ ] how to bring data to searching phase without malloc
- [ ] add points here!
### Searching
- [ ] how to bring data to searching phase without malloc
- [ ] add points here!

markus2330:
Thank you, very nice explanation! I am looking forward to the visualizations of the mappings!

Btw. please do not add any configuration for `OPMPHM` to cmake, it only confuses package maintainers and people with the needs to modify such internals, will patch the source anyway. Simply add it as constant to the beginning of kdbprivate.h.

KurtMi:
> Thank you, very nice explanation! I am looking forward to the visualizations of the mappings!

The visualization is more for debug purpose later on in the ordering and searching. The mapping constructs only this bipartite graph and I found out that drawing a bipartite graph with 16+ vertices over and over is very annoying.

KurtMi:
Due to plan change, this issue will be closed. The new one is #1487.

PR #1131, opened 2016-11-23
OPMPHM: add random and hash function

KurtMi:
Should do #887 add {random,hash} function

### DIFF
--- a/src/include/kdbopmphm.h
+++ b/src/include/kdbopmphm.h
@@ -1,54 +1,59 @@
 #ifndef OPMPHM_H
 #define OPMPHM_H
 
+#include <stdlib.h>
+
 /* Hash function
  * By Bob Jenkins, May 2006
  * http://burtleburtle.net/bob/c/lookup3.c
  */
 
-#define rot(x, k) (((x) << (k)) | ((x) >> (32 - (k))))
+#define OPMPHM_HASHFUNCTION_ROT(x, k) (((x) << (k)) | ((x) >> (32 - (k))))
 
-#define final(a, b, c)                                                                                                                     \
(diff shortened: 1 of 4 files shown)


src/include/kdbopmphm.h
```diff
@@ -1,6 +1,55 @@
 #ifndef OPMPHM_H
 #define OPMPHM_H
 
+/* Hash function
+ * By Bob Jenkins, May 2006
+ * http://burtleburtle.net/bob/c/lookup3.c
+ */
+
+#define rot(x, k) (((x) << (k)) | ((x) >> (32 - (k))))
```
KurtMi: I copied only the code, I will investigate and change it.

src/libs/elektra/opmphm.c
```diff
@@ -0,0 +1,47 @@
+/**
+ * @file
+ *
+ * @brief The Order Preserving Minimal Perfect Hash Map.
+ *
+ * @copyright BSD License (see doc/LICENSE.md or http://www.libelektra.org)
+ */
+
+
+#include "kdbprivate.h"
```
KurtMi: I have some questions regarding the reusability of my code, I would leave it so for now and discuss my questions the next time we meet.

KurtMi:
> Is it really necessary to have macros here? Did you benchmark that (inline) functions are slower?

I made some benchmarks and found out that inlining is in my opinion to slow.

Open things raised by this PR are noted in #887.

KurtMi:
@markus2330 this PR would be ready, if you don't have any objection.

markus2330:
Thanks great job! Please also set "ready to merge" label next time, there are plenty of ways to misunderstand "PR would be ready" (ready to review? OR: it would be, but it is not)

PR #1141, opened 2016-11-29
OPMPHM: remove kdbprivate include, refactor vstack vheap

KurtMi:
Trie removed from doxygen, due to be internal and no doc.


markus2330:
Thank you! Please prefer separate commits if possible (removal of kdbprivate and refactoring, not separate commits for reformating source)

PR #1143, opened 2016-11-29
OPMPHM: Add data structs + debug graph draw + Mapping phase + TestCases

KurtMi:
This PR includes:
* bipartite graph
* bipartite graph debug
* reusability
* ratio
* Mapping Phase
* Missing Vheap typedef usage
* TestCases Mapping

One question arises: who to test the Mapping?


scripts/opmphm-draw-graph
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
markus2330: Can you rename the file to draw-opmphm(-graph)?

src/include/kdbopmphm.h
```diff
@@ -4,6 +4,55 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+/* Vertex represents the vertex of the bipartite graph.
```
markus2330: Is there a specific reason not to use doxygen syntax (extra `*`) ?

src/include/kdbopmphm.h
```diff
@@ -4,6 +4,55 @@
...
+typedef struct
+{
+	uint32_t h[3];   /* the hash function value of each key */
+	int nextEdge[2]; /* the index of the next edge in the edge list of the vertex */
+	size_t order;    /* the desired index of the key in the opmphm hash map */
+} Edge;
+
+/* The vertices and edges represent the bipartite graph.
+ *
+ * Each Vertex has a list, each Edge is exactly in two list.
```
markus2330: two list**s**?

src/libs/elektra/opmphm.c
```diff
@@ -8,8 +8,90 @@
...
+/* Prints the bipartite graph to a dot file, to draw it, for debug purpose. */
+void opmphmPrintGraph (Edge * edges, void ** data, opmphmGetString fpOpmhpmGetString, size_t n)
+{
+	FILE * f = fopen ("opmphmGraph.dot", "wr");
+	fprintf (f, "graph graphname {\n");
+	for (size_t i = 0; i < n; ++i)
+	{
+		uint32_t h1 = edges[i].h[1];
+		uint32_t h2 = edges[i].h[2];
+		fprintf (f, "%u -- %u [label=\"%s\"]\n", h1, h2, fpOpmhpmGetString (data[i]));
```
markus2330: start with tab?

src/libs/elektra/opmphm.c
```diff
@@ -8,8 +8,90 @@
...
+	}
+	fprintf (f, "}\n");
+	fclose (f);
+}
+
+// The ratio defines the number of vertices on one side of the biparithegraph, denoted as r.
+// Double this value to get to the Fox et al. ratio.
+const double ratio = 0.75;
```
markus2330: static? (Do keep it local to file)

src/libs/elektra/opmphm.c
```diff
@@ -8,8 +8,90 @@
...
+/* The first phase of the order preserving minimal perfect hash map build. */
+int opmphmMapping (OPMPHM * opmphm, Vertex * vertices, Edge * edges, void ** data, opmphmGetString fpOpmhpmGetString, size_t n)
+{
+	unsigned int seed = time (NULL);
+	// set the seeds, for the hash function
+	for (int i = 0; i < 3; ++i)
+	{
+		opmphm->opmphmHashFunctionSeeds[i] = opmphmRandom (&seed);
+	}
+	size_t r = opmphmR (n);
```
markus2330: Why not simply multiply here? The function seems to be overkill!

src/libs/elektra/opmphm.c
```diff
@@ -8,8 +8,90 @@
...
+	ELEKTRA_LOG_DEBUG ("OPMPHM: Mapping: r=%lu seed[0]=%u seed[1]=%u seed[2]=%u\n", r, opmphm->opmphmHashFunctionSeeds[0],
```
markus2330: Isn't this more important then the logging of the individual keys? (Maybe simply use `ELEKTRA_LOG`)?

src/include/kdbopmphm.h
```diff
@@ -4,6 +4,55 @@
...
+ * nextEdge[2] is for the lists in the other side of the bipartite graph (r to 2*r-1)
+ */
+
+/* OPMPHM represents the final order preserving hash map.
+ * This struct is needed for the lookup.
+ */
+typedef struct
+{
+	uint32_t opmphmHashFunctionSeeds[3]; /* the seeds for the tree hash function calls */
+	unsigned int * g;		     /* saves the g () function need for the lookup and filled while the search phase */
```
markus2330: g () function?

KurtMi:
@markus2330 I haven't found a way to test this function, do you think a define and some rare ifdefs are the only possibility?

markus2330:
> I haven't found a way to test this function, do you think a define and some rare ifdefs are the only possibility?

Sorry, which function do you refer to?

KurtMi:
> Sorry, which function do you refer to?

opmphmMapping

markus2330:
Yes, you need to extract the `time()` call and add it as parameter, then you should be able to do regression tests? There should be no need for `#ifdefs`.

Btw. there is not much use in calling `time()` in a loop, you would nearly always get the same number (seconds tick really slow compared to execution of such functions).

markus2330:
One more hint: While a `#ifdef` within opmphmRandom would be acceptable a "closure"-like construct (that combines the state and a pointer to opmphmRandom) which is passed around is more elegant.

In the testcases the "closure"-like construct you pass around is a simple counter, otherwise it is some platform-dependent random number generator.

KurtMi:
I saw no other way than adding an `#ifndef` and making the `opmphmRatio` extern, to ensure a proper testing, now and in the future.

markus2330:
Thank you!

PR #1148, opened 2016-11-30
fix rest-backend linking + conf-spec improvements

### LINKED ISSUE #1147: rest-backend links only against elektra-full
(and it also seems to link against many libs that it does not directly depend on?)

expected behaviour: similar to other tools, please link against shared version: it is the recommended version. (FULL and STATIC is disabled by default)

Namoshek:
# Purpose

Fixes #1147 and improves config-specification.ini (added missing key + default values for API description).

# Checklist

needs to be checked.

- [ ] I added unit tests


doc/tutorials/snippet-sharing-rest-service.md
```diff
@@ -117,21 +117,15 @@ use the command `kdb mount`. You should see a list of mountpoints, of which one
...
+> kdb set system/sw/elektra/restbackend/#0/current/backend/jwt/encryption/secret "use a secret key here"
 ```
 
 To generate a secure key, you can also use `pwgen` (install via `apt-get install pwgen`).
-Simply swap the `use_a_secret_key_here!` with `$(pwgen -1sy 20)` to generate a secret.
-
-In case you don't want to publish the API description, you can also set the keys
-to other links, e.g., your main page. It is good practice to have the description
-published though. Information on how to generate the API description can be found below.
+Simply swap the key value (`use a secret key here`) with `$(pwgen -1sy 20)` to generate a secret.
```
markus2330: Why not directly put it there? Btw. the `-1sy` option might also put in quotes `' and "` which will cause problems.

src/tools/rest-backend/CMakeLists.txt
```diff
@@ -18,7 +18,12 @@ else ()
     set (config_default_profile "current")
 
     set (install_directory ${TARGET_TOOL_DATA_FOLDER}/${tool})
-    file (READ config-specification.ini configuration-specification)
+
+    configure_file (
+        "${CMAKE_CURRENT_SOURCE_DIR}/config-specification.ini"
```
markus2330: Please call it `.in`, otherwise in source builds will have troubles.

markus2330:
one more thing: you need to remove your tool when BUILD_SHARED is not active (unless you also add full and static variants as done in `kdb`. Not sure if there is any use for it.)

markus2330:
Thank you!

Issue #1152, opened 2016-11-30
kdbGet with cascading key yields too many results


Namoshek:
I think it is counter-intuitive that `kdbGet()` yields several equal keys from different namespaces when using a cascading key as parentKey. E.g. for `/sw/elektra/restbackend` I get
```
dir/sw/elektra/restbackend/something/here
spec/sw/elektra/restbackend/something/here
system/sw/elektra/restbackend/something/here
user/sw/elektra/restbackend/something/here
```
If you know what key you need, `ksLookup` solves the problem for you. But if you want to iterate a keyset and do something with it, you get way too many keys. Because they are not sorted by namespace, you'll also end up with overrides from lower namespaces if you do not filter them in some way (see above).

There seems to be some sort of solution for the issue with [keysetget.hpp](../blob/master/src/bindings/cpp/include/keysetget.hpp), but I think that it is not sufficient, because it does not solve the issue for the `spec` namespace; or in other words it seems to not create keys for default values in spec (or something like that).

Maybe the issue could be solved together with the "default-values from spec problem"?

markus2330:
Thank you for reporting the issue here!

> several equal keys from different namespaces

You are right, that is an important concept from Elektra that is not well explained (physical vs. logical view of configuration). http://tree.libelektra.org/doc/help/elektra-cascading.md does not really mention it neither are there any hints in the API description (in particular not where iteration is explained).

>  But if you want to iterate a keyset

Yes, currently the iteration API is really low-level and gives you every physical key.

If your application defines every key via spec you can simply *only* use spec keys.
Otherwise, you need quite some code to handle every case.

An API that would solve the problem quite elegantly would be a hierarchical nesting of the logical keys as done for python in http://tree.libelektra.org/src/tools/gen/support/nested.py. But we do not have it yet.

> they are not sorted by namespace

What do you mean here? They are sorted by namespace.

> does not solve the issue for the `spec`

As long default keys do not get added by `kdbGet` this is a bug.

> Maybe the issue could be solved together with the "default-values from spec problem"?

We could add every logical key within the keyset, which would allow you to iterate over the logical keys. The problem there would be that logical keys never get transferred back to the key database (to which file should they? It is the namespaces purpose to tell!) Additionally, you could easily create mismatches between what a cascading lookup should yield and the value of the logical key.

But we certainly need to clarify these topics before releasing 1.0. Thanks again for writing this up!

Namoshek:
[...]

PR #1156, opened 2016-12-01
rest-backend: add config transform tests + read default values

Namoshek:
# Purpose

This PR improves the whole configuration transformation by adding an additional test and using default values in case no key value is present. It can also use spec keys if no other configuration could be found and it adds keys for which no spec key is available as well to the output.

# Checklist

needs to be checked.

- [x] I added unit tests


src/tools/rest-backend/service_config.cpp
```diff
@@ -36,12 +39,73 @@ cppcms::json::value ConfigEngine::loadApplicationConfiguration (const std::strin
...
+	kdb::KeySet ksToVisit = ksDup.cut (kdb::Key ("spec/", KEY_END));
+
+	// all spec keys
+	for (auto elem : ksToVisit)
 	{
-		std::string key = elem.getName ().substr (elem.getName ().find (conf_root) + conf_root.length () + 1);
-		std::replace (key.begin (), key.end (), '/', '.');
+		kdb::Key elemCascading = elem.dup ();
+		kdb::tools::helper::removeNamespace (elemCascading);
+		kdb::Key key = ksDup.lookup (elemCascading.getName ());
```
markus2330: Can you elaborate why you chose to handle spec in a special way? I do not think this is necessary or can yield correct results.

By looking up in ksDup (which does not have spec keys anymore) the lookup cannot honor the specification anymore.

src/tools/rest-backend/tests/test_service_configengine.cpp
```diff
@@ -0,0 +1,93 @@
...
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12.with.trailing.array.#0.and.path", regex));
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12.with.trailing.array.#_12.and.path", regex));
+}
+
+TEST (kdbrestServiceConfigengineTest, TransformKeysetToJsonValue)
+{
+
+	std::string conf_root = "/sw/elektra/restbackend/#0/current";
+
+	kdb::Key k1 (conf_root + "/nested/object/var", KEY_CASCADING_NAME, KEY_VALUE, "simple string", KEY_END);
```
markus2330: KEY_CASCADING_NAME is obsolete, it is not needed anymore.

src/tools/rest-backend/tests/test_service_configengine.cpp
```diff
@@ -0,0 +1,93 @@
...
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12", regex));
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#0.with.trailing.path", regex));
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12.with.trailing.path", regex));
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12.with.trailing.array.#0", regex));
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12.with.trailing.array.#_12", regex));
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12.with.trailing.array.#0.and.path", regex));
+	ASSERT_TRUE (std::regex_match ("firstpart.second.third.#_12.with.trailing.array.#_12.and.path", regex));
+}
+
+TEST (kdbrestServiceConfigengineTest, TransformKeysetToJsonValue)
```
markus2330: Can you add another example which has spec keys? (and default values, overrides and so on)

By exporting with the c plugin can you can get code that represents the live config you use. It would be good to have such a test case.

src/tools/rest-backend/tests/test_service_configengine.cpp
```diff
@@ -0,0 +1,93 @@
...
+	expected["object"]["array3"][0]["var"][0] = "arr0-in-arr0";
+	expected["object"]["array3"][0]["var"][1] = "arr1-in-arr0";
+	expected["object"]["array3"][1]["var"][0] = "arr0-in-arr1";
+
+	cppcms::json::value result = kdbrest::service::ConfigEngine::instance ().transformKeysetToJsonValue (testKeySet, conf_root);
+
+	ASSERT_TRUE (result == expected);
+}
+
+int main (int argc, char * argv[])
```
markus2330: main not needed here, you do not do something special here.

src/tools/rest-backend/service_config.cpp
```diff
@@ -36,12 +39,73 @@ cppcms::json::value ConfigEngine::loadApplicationConfiguration (const std::strin
...
+		{
+			key = elem;
+		}
 
-		this->setValue (result, key, elem);
+		std::string keyName = key.getName ().substr (key.getName ().find (conf_root) + conf_root.length () + 1);
+		std::replace (keyName.begin (), keyName.end (), '/', '.');
+
+		// set default value in case elem has no value
+		if ((key.getBinarySize () == 0 || key.getString () == "") && key.hasMeta ("default"))
```
markus2330: This would be done by ksLookup if you had not removed the spec keys,

src/tools/rest-backend/service_config.cpp
Namoshek: That's why spec is used as fallback two lines below.

src/tools/rest-backend/tests/test_service_configengine.cpp
Namoshek: The cmake function to add tests is the same for all files. If I omit the main function in one test, I would have to add NO_MAIN to the test, which I don't want to do manually. And in case the test file gets more tests in future, we are prepared. :)

src/tools/rest-backend/service_config.cpp
markus2330: They key has to be inside the keyset so that `ksLookup` works.

src/tools/rest-backend/service_config.cpp
markus2330: It is not implemented within `kdbGet` as discussed in #1152.

But `ksLookup` of course implements all these specifications as described in `doc/help/elektra-cascading.md` or in `ksLookup`'s API docu. (This might add new default keys to the keyset, which is not described in the API docu because we hopefully can get rid of this behavior.).

I updated the docu to better describe this.

Namoshek: In that case it can be written a lot easier for sure.

src/tools/rest-backend/service_config.cpp
Namoshek: You mean that the default-value retrieval works? Makes sense.

src/tools/rest-backend/tests/test_service_configengine.cpp
markus2330: I could simply remove it and it still worked? But not really important..

src/tools/rest-backend/service_config.cpp
markus2330: Also so that symlinks and different namespaces work ;)

src/tools/rest-backend/tests/test_service_configengine.cpp
Namoshek: What is the NO_MAIN flag for the gtest function/macro in cmake then? Not that I ever understood it, but I hate it when things magically work for no obvious reason...

src/tools/rest-backend/service_config.cpp
markus2330: That would be great!

src/tools/rest-backend/tests/test_service_configengine.cpp
markus2330: In this case its quite simple: static linkers take only symbols that are needed.
So even you did not specify NO_MAIN, it still works: the other main is not added.

But we trust here on a very specific behaviour (which even fails when we swap the order of gcc arguments).

So you are right: its better if you add NO_MAIN and add main everywhere (so keep it here).

markus2330:
Thanks, looks much better now!

Issue #1151, opened 2016-11-30
website issues


markus2330:
### Frontend

- [ ] The rest docu seems like it needs data from apiary.io to be able to display anything (which again pulls in google analytics and more).
- [ ]  If snippet import was successful, but the export to the various formats failed both "success" and "error" is shown.
- [ ] plugin sections expandable

### Questions

- [x] Why is a plugin required for RSS? Do you generate atom or RSS?
---

### Frontend

- [x] fix snippets without author
- [X] create issue for privacy badger: https://github.com/EFForg/privacybadger/issues/1103
- [x] Mozilla Firefox 45.5.1: Webpage still is initially black (only at main page, others work as before, so it might be related to embedded asciinema). When scrolling down and then up, it gets white.
- [x] Every individual snippet exported separately in the repo
- [x] have asciinema delievered by main page
- [x] give some hints for people that do not know Elektra but want to upload a snippet: e.g. which plugins to use when uploading snippets (line if unsupported, ini for ini, not xmltool for xml,..), that some parts are unchangeable after upload..
- [x] search for all snippets
- [x] button for listing all snippets created by yourself
- [x] login fails with misleading error messages (connection failed)
- [x] have links to plugins (open in new tab) when selecting a plugin (share/convert)
- [x] firefox is not able to autocomplete login (user+password)
- [x] rename "additional plugin arguments:" + question mark explaining what happens with additional conf for plugin variants
- [x] add "additional plugin configuration" to "report issue on github" script
- [x] github button on main page
- [x] move rss feed to /news/feed.rss
- [x] link AUTHORS for "ElektraInitiative" bottom right
- [x] time zone for "Last update:"
- [x] blinking should stop when finally is there (or even remove bar, too)
- [x] "Examples" button should be "Tutorials" instead, the examples are not really interesting (or we *only* have "Getting Started"?)
- [x] root url http://namoshek.at:9000/ should not produce 404 but redirect to /home
- [x] use some other plugin per default in snippet conversion (not dump), maybe line is best
- [x] please add icons for the orientation, too
- [x] registration page: I enter first both passwords, then user, then mail, suddenly it says that second password is wrong
- [x] when entering: "hosts fstab system iso9600" tags get red and cannot submit tags/snippet
- [x] scope and organization should have defaults, description should not be enforced
- [x] when submitting errornous snippet, it says github button below should be pressed, but there is no such button
- [x] green tick in snippet window overlaps with scroll bar
[...]

PR #1158, opened 2016-12-02
rest-backend: add conversion tests + fix issues

Namoshek:
# Purpose

This PR adds some tests for the conversion (they only run if the required plugins are available). It does also fix the import / conversion issue mentioned in #1151.

# Checklist

needs to be checked.

- [x] I added unit tests

### DIFF
--- a/src/tools/rest-backend/benchmarks/benchmark_kdbrest_mysql.cpp
+++ b/src/tools/rest-backend/benchmarks/benchmark_kdbrest_mysql.cpp
@@ -176,8 +176,8 @@ std::vector<model::Entry> rsToEntryVector (sql::ResultSet * rs)
 		entry.setUploadPlugin (rs->getString ("plugin").asStdString ());
 		model::ImportedConfig cfg = service::ConvertEngine::instance ().import (rs->getString ("configuration").asStdString (),
 											entry.getUploadPlugin (), entry);
-		auto subkeys = cfg.getKeySet ();
-		entry.addSubkeys (subkeys.begin (), subkeys.end ());
+		auto ks = cfg.getKeySet ();
+		entry.addSubkeys (ks);
 		entry.setAuthor (rs->getString ("author").asStdString ());
 		entry.setViews (rs->getInt ("views"));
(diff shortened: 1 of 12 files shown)


markus2330:
Thanks for adding the tests, they are great!

But I do not think that overloading `addSubkeys` with quite different behaviour is a good idea! Better you pick different names if you really need both of these behaviors. Are you sure the initial implementation of `addSubkeys` is correct (in the case of having subkeys in different namespaces)?

Namoshek:
> if you really need both of these behaviors

I don't really *need* the behavior of the other (old) implementation, but it makes a huge difference when it comes to bootstrapping the cache. The very initial implementation was like
```
for (auto elem : ks)
{
    if (elem.isEntryRoot()) // actually done by regex, but for sake of simplicity...
    {
        model::Entry entry (elem);
        entry.addSubkeys (ks);
        // and add to list of entries here
    }
}
```
Obviously it is completely unnecesary to try adding every single key of the keyset as subkey because it is ordered; so the moment you leave the area where a key **can** be a subkey, you can quit. That's what the other implementation does.

But I agree, somehow this should be improved by taking the logic from the model to the service implementing the cache. This will also make it easier to swap it for the cachefilter implementation. I'll do so later, so keep the PR open pls.

Namoshek:
I removed the obsolete iterator-based `addSubkeys` method and added an optional flag to the `addSubkey` function, because the subkey comparison might be too expensive to do it twice for every key (inavoidable in some places otherwise).

markus2330:
Thank you, nice job!

Maybe you can come up with a better name than "secure" for this optimization flag and I am also interested in the benchmarks that show that this optimization actually is measureable.

Namoshek:
It is only measurable implicitely by using the cache (pre-load vs. live-load). Not sure how much of a difference it actually makes, but I guess true/false comparison is much cheaper than `key.isBelow(otherkey)`, so it should give an advantage here and there.

markus2330:
Of course there is a difference, the question is if it matters (is it measurable at all, is it 0.001%?). It definitely makes the API more difficult to use and obscures the responsibilities.

PR #1162, opened 2016-12-02
rest-frontend: improvements

Namoshek:
# Purpose

Fixes multiple issues as described in #1151.

- rss feed link needs `target="_self"` + icon
- readme explains links better

# Checklist

needs to be checked.

- [ ] I added unit tests


Namoshek:
I leave it to you if you want to merge it right away, maybe we collect some more minor stuff to fix right away too.

src/tools/rest-frontend/README.md
```diff
@@ -420,11 +420,22 @@ Angular binding (i.e. use `<span translate="LOCA_KEY"></span>` in favor of
...
+
 For internal links (that are links that lead to another sub-page of the website)
-the special router directive `ui-sref` should be used, as a normal link would
-trigger a page reload. The `ui-sref` directive works on state names and not on links,
+two options are available. It is possible to use the normal `href` HTML attribute
+or to use the special `ui-sref` attribute defined by the frontend router.
+The `ui-sref` directive works on state names and not on links,
 so if a sub-page like `<website-url>/docs/tutorials` exists, one cannot use
-`<a ui-sref="/docs/tutorials"></a>`. The state name for the tutorials page has to
+`<a ui-sref="/docs/tutorials"></a>`; the state name for the tutorials page has to
```
markus2330: Why would I want to use ui-sref? Isn't it better to only use the other variant?

src/tools/rest-frontend/README.md
```diff
@@ -420,11 +420,22 @@ Angular binding (i.e. use `<span translate="LOCA_KEY"></span>` in favor of
...
 be used, which is most likely `main.dyn.tutorials` if the tutorials section is based
 on the `structure.json.in`. The link (with a simple loca key) would look like
 `<a ui-sref="main.dyn.tutorials" translate="MENU.LINK.TUTORIALS"></a>` therefore.
+The `ui-sref` variant requires the HTML to be specially compiled though, what makes
+the usage of a normal `href` attribute easier in most scenarios.
+The following link does exactly the same as the last example with `ui-sref`:
+`<a href="/docs/tutorials" translate="MENU.LINK.TUTORIALS"></a>`.
```
markus2330: Why the `translate`? Is this automatically done by ui-sref?

Namoshek: No, absolutely not - the ui.router and angular-translate are entirely different packages and do not interact with each other at all. The reasons I added it are: 
1) There is probably no (serious) angular app out there at all that does not use the translate module. Everyone who had angular in their hands for more than 5 minutes knows what I'm talking about.
2) Writing a HTML link into the documentation without link text would be strange and maybe more confusing. For sake of simplicity I change it to three points though, that leaves space for own thoughs.

Namoshek:
Thanks for the feedback, I improved it again (hopefully). Let us keep the PR open, so we can use it as test PR if the build script is ready.

markus2330:
We can keep it open. But note that you can also trigger build jobs without PRs, simply press "Build Now" in [the jenkins user interface](http://build.libelektra.org:8080/job/elektra-homepage/): master will be built then.

src/tools/rest-frontend/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ else ()
     # install files to destination
     install (DIRECTORY public resources DESTINATION ${install_directory})
     install (FILES package.json README.md ${CMAKE_CURRENT_BINARY_DIR}/Gruntfile.js DESTINATION ${install_directory})
-    if (NOT EXISTS ${install_directory}/application-config.json)
+    if (NOT EXISTS ${CMAKE_INSTALL_PREFIX}/${install_directory}/application-config.json)
```
markus2330: I am afraid this will not work as you expect it: The test is already done at cmake time, which is too early. (You want it at installation time)

Two config files would be a much better way.

Namoshek: I'll add integration later, for the moment I want to keep it simple. Enumerating options is not really simple, my approach (default config + custom config) seems more interesting tbh.

Or to stay with the current approach, I could also execute a copy process during the installation that does not override.

markus2330: > my approach (default config + custom config) seems more interesting tbh. Or to stay with the current approach, I could also execute a copy process during the installation that does not override.

Yes, default config + custom config (where custom does not get installed but gets mounted for modifications) is a good idea. Complications in the installation procedure are better to be avoided (there are some obstacle, some we can deal with, like "DESTDIR", but others we cannot deal with: the approach simply won't work when Elektra is installed as package).

markus2330:
Thank you! Default config + custom config for a later PR then.

src/tools/rest-frontend/CMakeLists.txt
Namoshek: Then I'll change the mount script to:
- ensure a copy of the application-config.json exists (default > custom)
- mounts the copy of the config file
- and add a hint in the readme of course

Issue #1160, opened 2016-12-02
tutorial issues


markus2330:
- [x] please add the npm version requirements  #1150 to the tutorial (nodejs is not even mentioned in the apt-get command).

- [x]  the tutorial also needs updates regarding `kdb mount` (it always needs to be mounted manually, as you already suggested scripts would help).
- [x] "mountpoint" metadata and spec-mount seem to be still missing (thus type checking wont work)? You need:
```
[current]
mountpoint = rest-backend.ini
```
- [x] then (in the tutorial/with script) `kdb spec-mount /sw/elektra/restbackend/#0`, `kdb global-mount` and validation works if cascading names are used, e.g. `sudo kdb set -N system /sw/elektra/restbackend/#0/current/backend/jwt/validity blah` (-N system would be used as default because of sudo, but you never know if people remove sudo)

- [x] The tutorial still uses encryptionkey?

Issue #886, opened 2016-08-15
kdb global-mount: overwrites mounted plugins


mpranj:
Mounting multiple plugins globally with `kdb global-mount` results in only the most recently mounted plugin being executed.

I expected multiple global mounts not to overwrite previous mounts.

When I run:

```
kdb global-mount dbus
kdb global-mount tracer
```

the dbus plugin is lost/overwritten by tracer.
- Elektra Version: master

markus2330:
Thanks for reporting! Can you also report expected behavior? (maybe also in a non-trivial setup when you have 2 plugins and want add 2 others)

tom-wa:
the issue is probably the global-mount tools description running `kdb global-mount dbus tracer` instead seems to work fine

markus2330:
The docu doc/help/kdb-global-mount.md says "Note that previous calls of `kdb global-mount` will be overwritten.". Any suggestions how it can be improved?

But I would like to discuss what @mpranj expected, I agree that the overwrite semantics are maybe not what a user expects.

markus2330:
Btw. and it seems like even @tom-wa runned in the trap that `kdb global-mount` will overwrite plugins, so there is definitely a gap in user expectations here.

PR #1163, opened 2016-12-03
snippet-sharing: fixes + tutorial improvements

Namoshek:
# Purpose

This PR solves the issues in #1160.

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/doc/tutorials/snippet-sharing-rest-service.md
+++ b/doc/tutorials/snippet-sharing-rest-service.md
@@ -131,14 +131,19 @@ After that you need to set an additional configuration parameter that has no def
 It is recommended to set it for the system namespace if you will use a tool like
 `systemctl` to manage the services.
 ```
-> kdb set system/sw/elektra/restbackend/#0/current/backend/jwt/encryption/secret "use a secret key here"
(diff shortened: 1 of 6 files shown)


doc/tutorials/snippet-sharing-rest-service.md
```diff
@@ -146,14 +161,21 @@ configuration:
 > kdb set system/sw/elektra/restbackend/#0/current/cppcms/http/script_names/#0 "/"
 ```
 
+Note: here we have not used the option `-N system` because the CppCMS configuration is
```
markus2330: Any reason to do here differently? Maybe validation will be added later? Or we should always do validation (even when setting per namespace).

markus2330:
Thank you! The question was too general to address in this PR, should we open an extra issue for this?

Namoshek:
You are actually right, I could have written it with some view into the future.

Doing validation always seems a good idea and shouldn't be difficult to implement for the kdb tool.

markus2330:
Thanks, you are right, I opened #1164.

Btw. there is also #886 which would allow you to always issue kdb global-mount (if other plugins are already mounted, it would be a NOP)

PR #1166, opened 2016-12-05
rest-backend: benchmark fixes and improvements

Namoshek:
# Purpose

This PR fixes some issues with the rest-backend benchmarks and adds some speed improvements (in the prepare-test-data section). Details can be found in the commit messages.

# Checklist

needs to be checked.

### DIFF
--- a/src/tools/rest-backend/benchmarks/benchmark_kdbrest_elektra.cpp
+++ b/src/tools/rest-backend/benchmarks/benchmark_kdbrest_elektra.cpp
@@ -125,7 +125,14 @@ void benchmarkLookupSingleByKey (int numUsers, int numEntriesPerUser, int numTag
 	timer.start ();
 
 	// the stuff to benchmark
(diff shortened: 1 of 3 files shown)


markus2330:
Thank you! The `if (numUsers <= 0) return;` "optimization" looks dangerous, please make sure you actually always execute the benchmarks with data. Why not aborting the benchmark in such cases?

Namoshek:
This change is only because it resulted in an SQL error as we had an incomplete statement like `INSERT INTO table (field1, field2) VALUES ` before (without a set of values at the end). If we do not have to insert anything, we also do not need to execute the statement. It is not part of the measured part, but does only belong to the test data generation.

Btw. the test for zero entries was your suggestion, I wouldn't have added it in the first place.

markus2330:
> Btw. the test for zero entries was your suggestion, I wouldn't have added it in the first place.

My remark was mainly about that it should be visible where "zero" is in the diagram.

PR #1169, opened 2016-12-06
rest-frontend: improve configuration handling

Namoshek:
# Purpose

The PR changes the way configuration is managed for the rest-frontend. By default only a configuration template is installed, which needs to be copied and mounted by a script.

I know that new configuration values would not be visible for the user, but without full integration this is hardly achievable (at least not easily).

@markus2330 please let me hear what you think of the approach


markus2330:
Thank you, it is an improvement! But why not having no config, and then simply do `kdb set` as needed? (Only as question, not to be changed.) On updates the users can be instructed to do more `kdb set`.

The solution in this PR has an advantage: if the user keeps the original templates, he/she can also do a 3-way merge with `kdb import` when the template changes.

I would say let us leave it for now, the solution is good enough.

Issue #1033, opened 2016-10-20
Homepage Essentials


markus2330:
Following essentials should be done before the webpage can be redirected to www.libelektra.org:
- [x] PRs merged
- [x] installed on libelektra.org server (+ tutorial how to install)
- [x] nice main page with main points present
- [x] Registration stable (no discarding of accounts)
- [x] URLs stable (+ ability to link to all documents)
- [x] entry points of docu are nice and proof-read
- [x] documentation mostly fixed, so it looks good on the website and has a uniform style
- [x] news/changelog page
- [x] equal stability also for the backend
- [x] full automatic pipeline to deploy frontend and backend
- [x] improve security #1153 (at least add security flags to homepage pipeline)
- [x] reliable start/stop (see  #1181)
- [x] write release notes for the new snippet service/web page

Is something else missing?

markus2330:
Thanks for your input!

I added your points except that it says "documentation *mostly* fixed" and I excluded that #1030 needs to be done completely. Would of course be nice to have it, but it is not essential. Instead I added "entry points of docu are nice and proof-read".

markus2330:
I added deployment pipeline and security

Namoshek:
Deployment through build should actually work already.

markus2330:
Thank you! Deployment works! But please get rid of the sleep(s) during deployment. This adds a lot of downtime.

Namoshek:
I removed the sleep between the backend stop and start, but the other sleep should stay for a good reason: 
The start script runs the backend in the background and because the cache needs to be loaded first, we cannot be sure the backend is reachable already (necessary for sitemap generation and this time will increase with the amount of snippets). But the 60 seconds sleep is no problem because the last frontend version is still online and reachable the whole time - we just override the frontend files without taking it down at all.

markus2330:
Perfect, is resolved then!

markus2330:
I also added "write release notes for the new snippet service/web page". It should be nothing new and I can help. Can we also mention the homebrew and other things (not directly related to the software)?

Btw. what is the easiest way to create preliminary news that should not yet be displayed on the website?

Namoshek:
A PR or new branch should be the easiest way to avoid early publishing.

markus2330:
What about not having the correct file extension? Or do you use every file from the news directory?

Namoshek:
The news script fails on inappropriate filenames, but this could be changed easily so that it ignores them.

[...]

PR #1174, opened 2016-12-07
rest-frontend: some fixes and improvements

Namoshek:
# Purpose

Some fixes and improvements as requested in #1151 and #1175. For details see commit messages.


markus2330:
Thank you! Is there really nothing useful we can do when javascript is disabled? (E.g. at least show the content of the mainpage?)

Namoshek:
The home page itself is pretty much ugly without the whole frame around it. But we can keep it as improvement for the future (with low priority and after I'm done with all other stuff): create a static copy of the website or only the home page.

markus2330:
Yes, you are right. I started a list with low priority homepage issues in doc/todo/TOOLS

Let us focus on #1033!

markus2330:
jenkins build-homepage please

markus2330:
jenkins build homepage please

Namoshek:
Doesn't seem to build the homepage. But you can also merge it and we run it manually.

markus2330:
> Doesn't seem to build the homepage. But you can also merge it and we run it manually.

It worked, my first command had a typo (is deleted now).

So we can leave it open, but I would recommend to merge it soon (maybe after fixing links?)

Namoshek:
Should be fixed now

markus2330:
jenkins build homepage please

markus2330:
Great! Thank you!

A very minor thing: is it possible to jump to the anchor at storage plugins on the plugin's page?

Namoshek:
Even though the markdown parser adds some IDs to the elements, it doesn't seem to be possible because the markdown is rendered after the page is loaded. So the browser will look for an anchor element when it is not available yet.

Namoshek:
Btw. maybe it would look better if we'd use wide-screen images for the Qt-GUI and "other applications" as well, because the current ones look quite large in their hight (caused by the little text on the side).

markus2330:
Can you give some hints which resolution you want in the appropriate ticket?

PR #1184, opened 2016-12-12
Build: Various Small Fixes

sanssecours:
# Purpose

This pull request fixes various problems, mostly concerning the build system.

# Checklist

- [x] I checked the commit messages. The only issue reference I used (see commit 0fd64e8) should be correct.
- [x] I ran all tests. Everything works fine, if I set `ENABLE_LOGGER` to`OFF`. If I enable the logger, then a segmentation fault occurs in `test_opmphm_mapping` . As far as I can tell, the current state of the master branch shows the same behaviour.
- [x] I added code comments in commit 0fd64e8. All other changes should be well documented in the respective commit messages.
- [x] I checked a lot of checkboxes ✅.

@markus2330 Please review my pull request.


markus2330:
jenkins build gcc-configure-debian-log please

.travis.yml
```diff
@@ -33,10 +34,13 @@ before_install:
       brew install yajl;
       brew install augeas --HEAD;
       brew install dbus;
-      brew install qt5;
+      # Building Qt5 currently takes too much time.
```
markus2330: Why are you trying to build qt5, aren't there any binaries (with dbus)?

If they do not exist, we should simply opt-out dbus (see like it is done for Svg + a ifdef in src/treeviewmodel.cpp to not include the header file and body of TreeViewModel::connectDBus).

src/include/kdblogger.h
```diff
@@ -86,10 +86,10 @@ int elektraLog (int level, const char * function, const char * file, const int l
 
 #ifdef HAVE_LOGGER
 
-#define ELEKTRA_LOG_WARNING(...) elektraLog (ELEKTRA_LOG_LEVEL_WARNING, __func__, __FILE__, __LINE__, ##__VA_ARGS__)
-#define ELEKTRA_LOG_NOTICE(...) elektraLog (ELEKTRA_LOG_LEVEL_NOTICE, __func__, __FILE__, __LINE__, ##__VA_ARGS__)
-#define ELEKTRA_LOG(...) elektraLog (ELEKTRA_LOG_LEVEL, __func__, __FILE__, __LINE__, ##__VA_ARGS__)
-#define ELEKTRA_LOG_DEBUG(...) elektraLog (ELEKTRA_LOG_LEVEL_DEBUG, __func__, __FILE__, __LINE__, ##__VA_ARGS__)
+#define ELEKTRA_LOG_WARNING(...) elektraLog (ELEKTRA_LOG_LEVEL_WARNING, __func__, __FILE__, __LINE__, __VA_ARGS__)
```
markus2330: Thanks for removing this portability issue!

.travis.yml
sanssecours: > Why are you trying to build qt5, aren't there any binaries (with dbus)?

It looks like the current binary version of the Homebrew formula (a.k.a. ”bottle”) does not include D-Bus support. 

The [Qt download page](https://www.qt.io/download-open-source) lists disk image most likely containing binary versions of Qt. The current disk image is quite large though (about 800MiB). It will take me some time to download the image. After  the download finishes I will check if we can use the disk image to install Qt in the Travis build.

> If they do not exist, we should simply opt-out dbus…

Not requiring D-Bus support when building the Qt-GUI sounds like a good idea.

> Btw. is it possible to build mac os X packages on travis: `.dmg`, homebrew or something similar?

In theory [yes](http://sleepycoders.blogspot.co.at/2013/03/sharing-travis-ci-generated-files.html), but as far as I know there exists no official support for something like that.

markus2330: > After the download finishes I will check if we can use the disk image to install Qt in the Travis build.

Please do not waste too much time on this, I made dbus optional: 4bf21f4d400de168e9e8e3edf969ade79f600e8d

A homebrew formula would be much more wanted ;)

> In theory yes

This seems to be similar to https://stackoverflow.com/questions/29587581/how-to-build-macosx-executables-on-travis-ci

sanssecours: > Please do not waste too much time on this, I made dbus optional: 4bf21f4

Great! I undid my changes to the Travis configuration and rebased this branch on `master`. After that I successfully built the Qt-GUI on my machine, using the bottled version of Qt 5. As far as I can tell this pull request is ready to be merged \o/. 

By the way: It looks like the build jobs did not trigger correctly after my rebase. Do  you know a method to manually start the Travis build job?

markus2330:
Thank you, great job!

.travis.yml
sanssecours: > If the bottled version of Qt 5 is better than the homebrew one we can still change.

The bottled version of Qt 5 is the one installed via Homebrew! Homebrew uses analogies from beer brewing:

| Usual Term             | Homebrew Term |
| ---------------------- | ------------- |
| Package (Description)  | Formula       |
| Prebuilt Package       | Bottle        |
| Repository             | Tap           | 

. Popular Homebrew formulas also provide bottles. So does the Qt 5 formula. However, the Qt 5 formula only provides bottles for Qt without D-Bus support.

> Did the qt-gui work for you?

Yes.

> Even with dbus? (`kdb global-mount dbus`)

Currently `brew install qt5 --with-dbus` does not work on my machine, since I use the latest version of macOS (macOS 10.12.1). I am afraid I can not test this, unless I install Qt from the official [disk image](http://download.qt.io/official_releases/qt/5.7/5.7.0/qt-opensource-mac-x64-clang-5.7.0.dmg).

> When you are logged in you can restart it in the travis GUI. I think we had the discussion once in #160. (I also did not find the button)

Thank you for the information.

Issue #1172, opened 2016-12-07
memleak: test_service_convertengine


markus2330:
The test test_service_convertengine has a memleak [0]. If you cannot solve the memleak itself, please exclude jni. And please try to reduce the number of hardcoded excluded plugins in your tests. Maybe you can even share code with tests/kdb/testkdb_allplugins.cpp (see getAllPlugins()). A good place to share code between tests is tests/gtest-framework/gtest/gtest-elektra.h.

[0]
```
==6365==    at 0x4C28C20: malloc (vg_replace_malloc.c:296)
==6365==    by 0x52D4185: elektraMalloc (internal.c:241)
==6365==    by 0x15AEF1AB: elektraJniOpen (jni.c:156)
==6365==    by 0x50C88F4: elektraPluginOpen (plugin.c:295)
==6365==    by 0x4E54FB9: kdb::tools::Plugin::Plugin(kdb::tools::PluginSpec const&amp;amp;, kdb::KeySet&amp;amp;) (plugin.cpp:39)
==6365==    by 0x4E860AD: kdb::tools::Modules::load(kdb::tools::PluginSpec const&amp;amp;) (modules.cpp:49)
==6365==    by 0x4E86162: kdb::tools::Modules::load(std::string const&amp;amp;, kdb::KeySet const&amp;amp;) (modules.cpp:44)
==6365==    by 0x4E7E9DD: kdb::tools::ModulesPluginDatabase::lookupInfo(kdb::tools::PluginSpec const&amp;amp;, std::string const&amp;amp;) const (plugindatabase.cpp:217)
==6365==    by 0x4E812EE: kdb::tools::ModulesPluginDatabase::lookupAllProvidesWithStatus(std::string const&amp;amp;) const (plugindatabase.cpp:330)
==6365==    by 0x4E7FD2E: kdb::tools::ModulesPluginDatabase::lookupAllProvides(std::string const&amp;amp;) const (plugindatabase.cpp:363)
==6365==    by 0x477861: kdbrest::service::ConvertEngine::loadEnabledFormats() (service_convert.cpp:81)
==6365==    by 0x4381E4: ConvertEngine (service.hpp:133)
==6365==    by 0x4381E4: instance (singleton.hpp:18)
==6365==    by 0x4381E4: kdbrestServicesConvertengineTest_FindSuitablePluginCheck_Test::TestBody() (test_service_convertengine.cpp:34)
[...]

PR #1187, opened 2016-12-12
generate plugin variants + rest-backend/rest-frontend allow for plugin configuration

Namoshek:
# Purpose

Adds a function to the PluginDatabase that allows to generate plugin variants (currently only based on genconf, `system/elektra/plugins/variants/...` is not considered).

It does also add all plugin variants to the format list of the rest-backend and allows for configuration of plugins (both backend + frontend).

There seems to be an issue with augeas, as it does crash the backend on almost every import. Conversion to an augeas format seems to work though (tested with properties lens). Interestingly augeas does not accept the output it delivered as input, it still crashes. The simpleini plugin does work well with configuration (import and export).

Btw. I don't really see an option to override plugin infos, maybe you have an idea?

# Checklist

needs to be checked.

- [x] commit messages are fine
- [x] I added unit tests


markus2330:
jenkins build homepage please

markus2330:
jenkins build homepage please

markus2330:
Thank you, looks nice! The issue of the PR is that it again breaks ABI of libtools without any benefit to the PluginDatabase user (no structural or API improvements). I think the easiest way for you would be to create a subclass PluginVariantDatabase which adds the method(s) you need. KeySet for `system/elektra/plugins` can be constructor argument. (Alternative: free-standing function with PluginDatabase and KeySet as parameter.) 
See https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B if you need details for what you can do without breaking ABI. If you restructure PluginDatabase and reduce its bloat (and also avoid reloading every plugin again and again) I am fine with an API/ABI incompatibility.

> augeas crash

Can you please open a separate issue? Do you have a backtrace?

> Btw. I don't really see an option to override plugin infos, maybe you have an idea?

You are right, the current Plugin class does not support it. Let us keep this feature for later. So let us only define variants with different configs or suppress plugin/variants which do not work.

Btw. I installed libaugeas-dev libxml2-dev on the homepage server!

markus2330:
jenkins build homepage please

Namoshek:
> I think the easiest way for you would be to create a subclass PluginVariantDatabase which adds the method(s) you need. KeySet for system/elektra/plugins can be constructor argument.

That sounds like a good idea, I'll see what I can do.

> So let us only define variants with different configs or suppress plugin/variants which do not work.

So I can assume that an existing key like `system/elektra/plugins/simpleini/variants/spacesep` suppresses a genconf variant starting with `user/spacesep`?

markus2330:
>  That sounds like a good idea, I'll see what I can do.

Thanks!

> So I can assume that an existing key like system/elektra/plugins/simpleini/variants/spacesep suppresses a genconf variant starting with user/spacesep?

I made it more explicit with a `disable` and `override` flag. You can also update the decision if needed, or even better: make docu out of it.

Namoshek:
The new `listAllPlugins()` implementation could potentially also be used to fix the memleak #1172, but it could be a bit dirty to set the suppressing keys in the live database.

src/libs/tools/include/plugindatabase.hpp
```diff
@@ -165,6 +165,60 @@ class ModulesPluginDatabase : public PluginDatabase
 	std::vector<PluginSpec> lookupAllProvides (std::string const & provides) const;
 };
 
+class PluginVariantsDatabase : public ModulesPluginDatabase
```
markus2330: Please prefer singular (PluginVariantDatabase)

see doc/DESIGN.md

src/libs/tools/include/plugindatabase.hpp
```diff
@@ -165,6 +165,60 @@ class ModulesPluginDatabase : public PluginDatabase
...
+	 * @note if no plugin variant could be found, an empty vector will be returned.
+	 *
+	 * @param whichplugin is the plugin for which we want all variants
+	 *
+	 * @return a vector of plugin variants for the given plugin
+	 */
+	std::vector<PluginSpec> getPluginVariants (PluginSpec const & whichplugin) const;
+
+private:
+	KeySet pluginconf;
```
markus2330: The private data must be inside the impl pointer, otherwise pimpl idiom is rendered useless. (If we want to remove pluginconf)

src/libs/tools/src/plugindatabase.cpp
```diff
@@ -379,6 +380,108 @@ std::vector<PluginSpec> ModulesPluginDatabase::lookupAllProvides (std::string co
...
+class PluginVariantsDatabase::Impl
+{
+public:
+	Impl ()
+	{
+	}
+	~Impl ()
+	{
+	}
+	Modules modules;
```
markus2330: We already have Modules within  ModulesPluginDatabase we inherit from?

Make impl protected or make them friends to access impl (should not affect ABI).

src/libs/tools/src/plugindatabase.cpp
```diff
@@ -379,6 +380,108 @@ std::vector<PluginSpec> ModulesPluginDatabase::lookupAllProvides (std::string co
...
+
+PluginVariantsDatabase::~PluginVariantsDatabase ()
+{
+}
+
+std::vector<std::string> PluginVariantsDatabase::listAllPlugins () const
+{
+	std::vector<std::string> plugins (ModulesPluginDatabase::listAllPlugins ());
+	plugins.erase (std::remove_if (plugins.begin (), plugins.end (),
+				       [this](const std::string & elem) {
```
markus2330: Nice implementation

src/libs/tools/src/plugindatabase.cpp
```diff
@@ -379,6 +380,108 @@ std::vector<PluginSpec> ModulesPluginDatabase::lookupAllProvides (std::string co
...
+				       }),
+		       plugins.end ());
+	return plugins;
+}
+
+std::vector<PluginSpec> PluginVariantsDatabase::getPluginVariants (PluginSpec const & whichplugin) const
+{
+	PluginPtr plugin = impl->modules.load (whichplugin);
+
+	// TODO variant overrides from kdb
```
markus2330: TODO

Please also refactor then, method is getting too long.

src/libs/tools/tests/testtool_plugindatabase.cpp
```diff
@@ -0,0 +1,88 @@
+/**
+ * @file
+ *
+ * @brief Tests for the pluginspec class
```
markus2330: copy&paste error (pluginspec -> plugindatabase)

src/tools/rest-backend/model_pluginformat.hpp
```diff
@@ -97,10 +99,21 @@ class PluginFormat
 		return m_pluginstatuses;
 	}
 
+	/**
+	 * @brief getter for the plugin configuration
+	 * 
+	 * @returen A keyset containing the plugin config
```
markus2330: typo: return

src/tools/rest-backend/model_pluginformat.hpp
```diff
@@ -97,10 +99,21 @@ class PluginFormat
...
+	 */
+	kdb::KeySet getConfig () const
```
markus2330: Return a const object? (Otherwise values of the keyset could be changed affecting m_config)

src/tools/rest-backend/service_convert.cpp
```diff
@@ -77,7 +96,10 @@ std::vector<model::PluginFormat> ConvertEngine::loadEnabledFormats ()
 	using namespace kdb;
 	using namespace kdb::tools;
 
-	ModulesPluginDatabase db;
+	KDB kdb;
+	KeySet conf;
+	kdb.get (conf, "system/elektra/plugins");
+	PluginVariantsDatabase db (conf);
```
markus2330: Can we avoid having it twice?

src/libs/tools/src/plugindatabase.cpp
Namoshek: How can I add an additional member (`KeySet pluginconf`) in the inherited `Impl` class? Adding an inherited implementation of it for `PluginVariantDatabase` yields a redefinition error.

markus2330: You do not have to inherit the Impl class? The `impl` needs a different name then..

markus2330: I am not sure what you tried and what the error was. Please update the PR with the issue you have (you can rebase the last commit later)

> But how to add the member without inheriting the class?

You have two different `impl` and can access both as needed.

> Can I just add the KeySet pluginconf to ModulesPluginDatabase::Impl?

From ABI point of view: yes. But it would be strange because ModulesPluginDatabase does not need it.

Namoshek:
jenkins build homepage please

markus2330:
Augeas plugin variants seem to be missing?

When I tried to create a new snippet with plugin parameters (btw. please prefer the word configuration here) I got an error and success and the same time. When trying to save again it says: "An entry with the entered details (organization, application, scope and slug) does already exist. Please choose different values."

Namoshek:
> Augeas plugin variants seem to be missing?

Weird, they are shown on my system properly.

> When I tried to create a new snippet with plugin parameters (btw. please prefer the word configuration here) I got an error and success and the same time. When trying to save again it says: "An entry with the entered details (organization, application, scope and slug) does already exist. Please choose different values."

In this case the snippet import was successful, but the export to the various formats failed.

markus2330:
I created #1192, conversion with augeas did not work.

The format for augeas is also always "none", is this because the contract cannot be changed?

What happens if configuration is changed for a plugin variant? There should be some clarifications in the user interface.

Namoshek:
> The format for augeas is also always "none", is this because the contract cannot be changed?

Yes.

> What happens if configuration is changed for a plugin variant? There should be some clarifications in the user interface.

What do you mean with that?

markus2330:
> What happens if configuration is changed for a plugin variant? There should be some clarifications in the user interface.

When uploading or converting a snippet there is a field "Additional plugin parameters: " field (which should be renamed to "Additional plugin configuration:"). My question is what if I have text in this field and select a plugin variant that already sets plugin configuration. How is this configuration merged?

For example: I select augeas Hosts.lns and additionally set some unrelated configuration in the field, is the lens configuration of the variant passed to the plugin or not?

Namoshek:
Yes, it is simply added to the plugin name (which is the plugin name + variant config for a plugin variant). So you will get:
```
Plugin: augeas lens=Hosts.lns
Config: pretty=1
------------------------------------------
Result: augeas lens=Hosts.lns pretty=1
```
What happens in case of an override (e.g. custom config defines `lens` as well), depends on the `parsePluginArguents()` function from libtools I used. Haven't tried this so far.

I can add a question mark with a hint.

markus2330:
> I can add a question mark with a hint.

Thank you!

Namoshek:
jenkins build homepage please

Namoshek:
I think I found the reason: The augeas plugin seems to require a new-line at the end of the file to import.

Namoshek:
jenkins build homepage please

markus2330:
Thank you, great job!

Unfortunately testtool_plugindatabase seems to fail at MacOS X

Namoshek:
```
[ RUN      ] PluginVariantsDatabase.getPluginVariants
/Users/travis/build/ElektraInitiative/libelektra/src/libs/tools/tests/testtool_plugindatabase.cpp:86: Failure
Value of: augeas_variants.size () > 0
  Actual: false
Expected: true
[ FAILED   ] PluginVariantsDatabase.getPluginVariants (2 ms)
```
I would interprete this as an error of the augeas genconf, which doesn't seem to generate variants on mac osx. The test itself is secured so that it only executes the test on augeas if it was found. A dirty fix would be to remove the check, but I think a genconf check for augeas is quite useful / almost mandatory.

markus2330:
Let us add an ifdef to exclude this testcase for APPLE as workaround.

The [FAQ of Augeas](http://augeas.net/faq.html) says: 
"File format descriptions are contained in files with the extension .aug. In the source distribution, they are in the lenses directory, which, by default, is installed into /usr/share/augeas/lenses."

Which does not really give an hint where it can be installed to in other cases (and how to find out). @sanssecours Can you tell us where the lenses are installed with homebrew? Does `pkg-config --variable=prefix augeas` give the correct prefix?

sanssecours:
> @sanssecours Can you tell us where the lenses are installed with homebrew?

Of course, the folder is `/usr/local/share/augeas/lenses/dist`. The prefix (`/usr/local`) depends on the install location of Homebrew, as you can see [here](https://github.com/Homebrew/homebrew-core/blob/master/Formula/augeas.rb#L40).

> Does `pkg-config --variable=prefix augeas` give the correct prefix?

Yes, the command returns 

    /usr/local/Cellar/augeas/1.7.0

. If we add the postfix 

    share/augeas/lenses/dist

, then we have 

    /usr/local/Cellar/augeas/1.7.0/share/augeas/lenses/dist

, which is the correct, since `/usr/local/share/augeas` points to `/usr/local/Cellar/augeas/1.7.0/share/augeas`.

markus2330:
Thank you to you both! I will merge this and lets fix the augeas-lenses directory issue separately.

Namoshek:
Sounds good, thank you!

Issue #1019, opened 2016-10-12
Crash on exporting with tcl plugin


markus2330:
# Describe what you wanted to do

```
$ cd src/plugins/xmltool/xmltool
$ kdb import user/tests/dump xmltool < dump.xml
$ kdb export user/tests/dump tcl
{
        {
*** Error in `kdb': munmap_chunk(): invalid pointer: 0x000000000119c490 ***
zsh: abort (core dumped)  noglob kdb export user/tests/dump tcl
```

Exporting with ni plugin (or xmltool or dump) works without issues.
Seems like tcl plugin creates broken KeySets refering to deleted or non-existing keys?
# System Information
- Elektra Version: master
# Backtrace

```
#0  keyDecRef (key=key@entry=0x40) at /home/markus/Projekte/Elektra/current/src/libs/elektra/key.c:583
#1  0x00007f65b0ca7e58 in ksClose (ks=ks@entry=0x1068380) at /home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c:2393
#2  0x00007f65b0ca7eb3 in ksDel (ks=0x1068380) at /home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c:356
#3  0x0000000000445af7 in ~KeySet (this=0x7ffd9efb2300, __in_chrg=<optimized out>) at /home/markus/Projekte/Elektra/current/src/bindings/cpp/include/keyset.hpp:568
#4  ExportCommand::execute (this=<optimized out>, cl=...) at /home/markus/Projekte/Elektra/current/src/tools/kdb/export.cpp:39
#5  0x0000000000417427 in main (argc=4, argv=0x7ffd9efb26a8) at /home/markus/Projekte/Elektra/current/src/tools/kdb/main.cpp:14
```
# Valgrind

``````
{
        {
==23667== Invalid write of size 8
==23667==    at 0x4E3A835: ksRewind (keyset.c:1219)
==23667==    by 0x4E3BE4D: ksClose (keyset.c:2390)
==23667==    by 0x4E3BEB2: ksDel (keyset.c:356)
==23667==    by 0x445AF6: ~KeySet (keyset.hpp:568)
==23667==    by 0x445AF6: ExportCommand::execute(Cmdline const&) (export.cpp:39)
==23667==    by 0x417426: main (main.cpp:145)
==23667==  Address 0xce41688 is 24 bytes inside a block of size 48 free'd
[...]

PR #1194, opened 2016-12-13
TCL Plugin: Fix Bug in Serialization Function

sanssecours:
# Purpose

This pull request fixes a bug in the TCL plugin.

# Checklist

- [x] Commit messages are fine (References to issues are correct).

@markus2330 Please review my pull request.

### DIFF
--- a/src/plugins/tcl/README.md
+++ b/src/plugins/tcl/README.md
@@ -9,7 +9,7 @@
 
 ## Introduction ##
 
-This plugin is a storage plugin which write keys to lists *in the style of*
+This plugin is a storage plugin which writes keys to lists *in the style of*
 the Tcl programming language.
 
 ## Format ##
--- a/src/plugins/tcl/action.cpp
+++ b/src/plugins/tcl/action.cpp
@@ -34,7 +34,7 @@ void serialise (ostream & ofs, KeySet & output)
 	while (Key k = output.next ())
 	{
 		ofs << "\t{" << endl;
-		ofs << "\t\t" << k.getName () << " = " << k.getString () << endl;
+		ofs << "\t\t" << k.getName () << " = " << (k.isString () ? k.getString () : k.getBinary ()) << endl;
 		k.rewindMeta ();
 		while (const Key m = k.nextMeta ())
 		{


markus2330:
Hi, thanks for the PR!

I am afraid the solution is unsatisfactory because:

1. it would make tcl-files binary without any escaping
2. in the current implementation it gets cut off at the first null byte?

The proper solution is to require a encoding plugin that encodes the binary values to string values.

Currently we only have the "null" plugin, which only handles null keys but not other binary keys. base64 would be a solution, but I am afraid it is not yet tested for this scenario.

Btw.  most storage plugins have this problem.

sanssecours:
> I am afraid the solution is unsatisfactory…

Thank you for the information, I will take another look at it.

markus2330:
Btw. you can simply push the typo you found. Thanks for that!

sanssecours:
> The proper solution is to require a encoding plugin that encodes the binary values to string values.

As far as I can tell I need to add `null` and `base64` to the list `infos/needs` in the ReadMe of the TCL plugin:

```md
- infos/needs = code null base64
```

. I did that and mounted the TCL plugin at `user/tests` with the following command:

```sh
kdb mount test.tcl user/tests tcl
```

. After that I imported `dump.xml` and looked at the resulting file using the commands below.

```sh
kdb import user/tests/dump xmltool < src/plugins/xmltool/xmltool/dump.xml
kdb file user/tests | xargs "$EDITOR"
```

The content of `test.tcl` looks more or less correct:

```tcl
{
	{
		user/tests/dump/.HiddenBinaryKey = @BASE64QmluYXJ5VmFsdWUA
		{
			comment = Binary key with hidden name
		}
	}
	{
		user/tests/dump/.HiddenDirectoryKey = DirectoryValue
		{
			comment = Directory key with hidden name
		}
		{
			mode = 700
		}
	}
	{
		user/tests/dump/.HiddenStringKey = StringValue
		{
			comment = String key with hidden name
		}
	}
	{
		user/tests/dump/PerfectBinaryKey = @BASE64QmluYXJ5VmFsdWUA
		{
			comment = Binary key with standard name
		}
	}
	{
		user/tests/dump/PerfectDirectoryKey = DirectoryValue
		{
			comment = Directory key with standard name
		}
		{
			mode = 700
		}
	}
	{
		user/tests/dump/PerfectStringKey = StringValue
		{
			comment = String key with standard name
		}
	}
	{
		user/tests/dump/Ug.ly:Bin@a€ry Key = @@NULL
		{
			comment = Binary key with ugly name
		}
	}
	{
		user/tests/dump/Ug.ly:Dir@ect€ory Key = @@EMPTY
		{
			comment = Directory with ugly name
		}
		{
			mode = 700
		}
	}
	{
		user/tests/dump/Ug.ly:St@ri€n.g Key = With a string value
		{
			comment = string key with ugly name
		}
	}
}
```

. If I export these values via

```sh
kdb export user/tests/dump tcl
```

, then all base64 encoded values will be converted back to their original binary form before they arrive at the `set` function of the TCL plugin. This is of course not ideal since this way the TCL plugin tries to write binary values using `key.getString()` again, and fails with the error described in issue #1019. It also seems kind of weird, that the plugin tries to convert the values again, although the version on disk is already correct?

Another problem is, that if I **do not mount** the TCL plugin to `user/tests`, import `dump.xml` and export `user/tests/dump` then the `base64` plugin is not invoked at all.

As you can see in pull request #1246 I re-read large parts of the documentation. However, I was not able to find a definite answer to my problem. It looks like the  [list plugin](/src/plugins/list/README.md) could help here, but I am not really sure how I can add the `null` and `base64` plugin at the appropriate places. 

Any help would be appreciated.

markus2330:
> As far as I can tell I need to add null and base64 to the list infos/needs in the ReadMe of the TCL plugin:

For long term it would be better if we have some provider for binary escaping, and base64 is one way to do the escaping. So we would have:

     - infos/needs = code null binary

and base64 would have

      - infos/provides = binary

@petermax2 I assume the filefilter in provides is a copy&paste error from fcrypt?

`kdb import user/tests/dump xmltool` for me this crashes already... When running again I get error `#61`.

markus2330:
> `kdb export` then all base64 encoded values will be converted back to their original binary form before they arrive at the set function of the TCL plugin

This is a different issue: kdb export currently only allows a single plugin, but would need base64. See  #1229 for the proposal to add an improved import/export (which considers `infos/needs`).

>  do not mount [...] export user/tests/dump

I assume you mean exporting with tcl? Then it is also #1229.

> I re-read large parts of the documentation

If you are working on issues not solved up to now there is unfortunately not much documentation around. And it is also not a good idea to write documentation of what actually does not work (except of `doc/decisions`). So it is better if you ask here.

The greater goal is to have storage plugins exclusively concerned with parsing and writing out configuration. Limitations of the parser should be handled by dedicated plugins that handle exactly that. So in this case here, we want to help parsers (that cannot handle binary data) by always giving them strings where the problematic bytes are already encoded. The idea is old, but it was never fully implemented. There are some plugins flying around that do some things in this direction (like ccode, hexcode, and base64) but they actually nowhere near actually fulfilling the greater goal.

If it helps you, we can split the issues into smaller steps, so instead of only talking about the "greater goal", we could write many small steps towards the goal. `base64` was not written for exactly this use case, so it might fail in some cases.

It might also help you to get started by writing your own parser, then you better know the limitations of your parser and know what other plugins need to recode first.

I would also recommend to you to write a tutorial that explains what is going on behind the scenes how plugins work together. If you write such a tutorial (like you nearly did in your post above) I can give you better feedback if we are talking about the same goals.

It might also help you to look at the test cases for plugin functionality (there are some already in `kdb test`, see `src/tools/kdb/test.cpp`; or also tests/shell/check_import.sh) that test if every possible key/value can be written and read out again. Roundtripping properties are something quite essential for configuration file parsers, so you should acquaint yourself with it.

sanssecours:
Thank you for your detailed response.

> `kdb import user/tests/dump xmltool` for me this crashes already... When running again I get error `#61`.

I remember having a similar issue too. However, currently I am unable to reproduce this problem.

> I assume you mean exporting with tcl?

Yes.

petermax2:
> I assume the filefilter in provides is a copy&paste error from fcrypt?

Very likely!

sanssecours:
Looks like fixing this issue will take more time than I assumed. I think I will look into this later, when I have more experience with the Elektra code base.

markus2330:
Yes, as said it is better to start with your own storage plugin. This should give you faster progress and more confidence to throughly fix the underlying issues surfacing in tcl (and nearly all others). In the next meeting we will look a bit more into plugin dependencies.

PR #1203, opened 2016-12-18
Date

tom-wa:
# Purpose

Date validation plugin

# Checklist

- [x] I added unit tests

### DIFF
--- a/src/plugins/date/README.md
+++ b/src/plugins/date/README.md
@@ -1,0 +1,46 @@
+- infos = Information about the date plugin is in keys below
+- infos/author = Name <name@libelektra.org>
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/recommends =
+- infos/placements = presetstorage postgetstorage
+- infos/status = recommended productive maintained reviewed conformant compatible coverage specific unittest tested libc final
+- infos/metadata =
+- infos/description = one-line description of date
+
+## Introduction ##
+
+Copy this date if you want to start a new
+plugin written in C.
+
+## Usage ##
+
+You can use `scripts/copy-date`
+to automatically rename everything to your
+plugin name:
+
+	cd src/plugins
+	../../scripts/copy-date yourplugin
+
+Then update the README.md of your newly created plugin:
+
+- enter your name+email in `infos/author`
+- make sure `status` and other clauses conform to
+  descriptions in `doc/CONTRACT.ini`
+- update the one-line description above
+- add your plugin in `src/plugins/README.md`
+- and rewrite the rest of this `README.md` to give a great
+  explanation of what your plugin does
+
+## Dependencies ##
+
+None.
+
+## Examples ##
+
+TBD
+
+## Limitations ##
+
+None.
--- a/src/plugins/date/README.md
+++ b/src/plugins/date/README.md
@@ -7,7 +7,7 @@
 - infos/placements = presetstorage postgetstorage
 - infos/status = recommended productive maintained reviewed conformant compatible coverage specific unittest tested libc final
 - infos/metadata =
-- infos/description = one-line description of date
+- infos/description = validates date and time strings
 
 ## Validation options ##
 
--- a/src/plugins/date/README.md
+++ b/src/plugins/date/README.md
@@ -11,40 +11,51 @@
 
 ## Validation options ##
 
-### Format string ###
+The following representation standards of dates and times are currently supported and can be use by setting `check/date` to:
 
-`check/date/format`
-	a format string compatible to `STRPTIME(3)`,
+* `POSIX`
+ 
+   see `STRPTIME(3)` for more information. a valid format has to be specified in `check/date/format` 
 
-the date plugin tries to parse the date/time value base on the format string
+* `ISO8601`
+  
+   see [ISO8601](https://en.wikipedia.org/wiki/ISO_8601). possible format strings specified in `check/date/format`:
+ 
+ * Date:
+ 
+  * `<calendardate>`
+   
+   `YYYY-MM-DD`, `YYYYMMDD`, `YYYY-MM`, `--MM-DD`, `--MMDD`
+   
+  * `<weekdate>`
+   
+   `YYYY-Www`, `YYYYWww`, `YYYY-Www-D`, `YYYYWwwD`
+   
+   * `<ordinaldate>`
+   
+   `YYYY-DDD`, `YYYYDDD`
+   
+   * `<date>`
+   
+   `<calendardate>`, `<weekdate>`, `<ordinaldate>` 
+ 
+ * Time:
+  
+  * `<time>`
+  
+   `hh:mm:ss`, `hhmmss`, `hh:mm`, `hhmm`, `hh`
+  
+  * `<tz>`
+  
+   `<time>Z`, `<time>±hh:mm`, `<time>±hhmm`, `<time>±hh`
+ 
+ * Combined:
+ 
+   `<date>T<time>`
 
-### ISO8601 ###
-
-`check/date/iso8601`
-	a combination of the basic formats specified in iso8601 (see [wikipedia.org/ISO8601](https://en.wikipedia.org/wiki/ISO_8601) )
-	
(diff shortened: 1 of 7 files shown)


tom-wa:
@markus2330 can you take a look at [README.md](https://github.com/tom-wa/libelektra/blob/date/src/plugins/date/README.md) and [testmod_date](https://github.com/tom-wa/libelektra/blob/date/src/plugins/date/testmod_date.c) regarding the possible validation options ? `check/date`

markus2330:
Thanks for the PR!

Is the combination of different standards with different formats for a date useful?

If not, you should instead use a `check/date` which specifies the standard (ISO8601, POSIX (for STRPTIME), or RFC2822) and an optional `check/date/format` for further format specifications.

> i'm planing to use (just) check/date for a lax validation which just tests if it's valid date/time based on the locales and 2-3 commonly used representations

Sounds rather surprising, better to add more standards where people know what they will get.

src/plugins/date/README.md
```diff
@@ -0,0 +1,70 @@
+- infos = Information about the date plugin is in keys below
+- infos/author = Name <name@libelektra.org>
```
markus2330: Please add name

src/plugins/date/README.md
```diff
@@ -0,0 +1,70 @@
...
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/recommends =
+- infos/placements = presetstorage postgetstorage
+- infos/status = recommended productive maintained reviewed conformant compatible coverage specific unittest tested libc final
+- infos/metadata =
+- infos/description = one-line description of date
```
markus2330: Please add short description

src/plugins/date/README.md
```diff
@@ -0,0 +1,70 @@
...
+ 
+  a set of possible format strings derived from rfc2822 3.3, no format string needed.
+
+## Dependencies ##
+
+None.
+
+## Examples ##
+
+TBD
```
markus2330: Please use shell recorder tutorial syntax and add README to be executed.

src/plugins/date/date.c
```diff
@@ -0,0 +1,347 @@
...
+#include <kdbhelper.h>
+#include <langinfo.h>
+#include <locale.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+
+static int stricmp (const char * s1, const char * s2)
```
markus2330: How does this differ from strcasecmp? Please document all functions.

src/plugins/date/date.c
```diff
@@ -0,0 +1,347 @@
...
+	}
+}
+
+static int doIsoStringValidation (const char * date, const char * isoString)
+{
+	int hasDate = 0;
+	int hasT = 0;
+	int hasTime = 0;
+	int hasTZ = 0;
+	if (strstr (isoString, "<date>") || strstr (isoString, "<calendardate>") || strstr (isoString, "<weekdate>") ||
```
markus2330: Why not use %X identifiers here, too?

src/plugins/date/README.md
```diff
@@ -0,0 +1,70 @@
...
+  
+   `<time>Z`, `<time>±hh:mm`, `<time>±hhmm`, `<time>±hh`
+ 
+ * Combined:
+ 
+   `<date>T<time>`
+
+* `RFC2822`
+ 
+  a set of possible format strings derived from rfc2822 3.3, no format string needed.
```
markus2330: How to specify a specific one?

src/plugins/date/date.c
```diff
@@ -0,0 +1,347 @@
...
+	if (strchr (isoString, 'T')) hasT = 1;
+	if (strstr (isoString, "<time>")) hasTime = 1;
+	if (strstr (isoString, "<tz>")) hasTZ = 1;
+	if (!hasT && ((hasDate && hasTime) || (hasDate && hasTZ))) return 0;
+	if (hasT && (!hasDate || !hasTime)) return 0;
+	struct tm tm;
+	memset (&tm, 0, sizeof (struct tm));
+	if (hasDate)
+	{
+		const char ** fmtStrings;
```
markus2330: please refactor

src/plugins/date/testmod_date.c
```diff
@@ -0,0 +1,87 @@
...
+}
+
+int main (int argc, char ** argv)
+{
+	printf ("DATE     TESTS\n");
+	printf ("==================\n\n");
+
+	init (argc, argv);
+
+	testFmt ("20:15:00", "%H:%M:%S", 1);
```
markus2330: Tests are really nicely specified!

src/plugins/date/date.c
tom-wa: i'd like to say the reason for not using strcasecmp is that it's nonstandard and not portable... but the actual reason: i had no idea strcasecmp actually existed.

src/plugins/date/date.c
tom-wa: i don't think it's guaranteed that %X is valid because it's locale dependent.

src/plugins/date/date.c
markus2330: Use elektraStrCaseCmp if you are worried about portability issues.

src/plugins/date/date.c
markus2330: Thanks for investigating!

tom-wa:
@markus2330 
i'm having some troubles on how to (let the user) specify iso8601 formats. the iso specifies:

* `dates`:
  - `calendar date`
  - `ordinal date`
  - `week date`
* `time`:
  - `time of day`
  - `universal time`
* `date and time`:
  - rules for combinations of date + time

each of the date and time representations has 3 different "sub"-representations with a basic and an extended representation

so `calendar date` e.g. consists of 1 `complete`, 3 `reduced` and 6 `truncated` representations with a basic representations each (e.g. `YYYYMMDD`) and an extended representation (e.g. `YYYY-MM-DD`) for half of them

any suggestions on a naming scheme or syntax ?

src/plugins/date/README.md
tom-wa: is it necessary to do that ? 
to be rfcX conform the date/time has to match one of the possible derivations

markus2330: Ok, if they specifically require every parser needs to handle any of the derivations, it should be okay.

markus2330:
Because they even have rules how to combine date+time, you can simply give a list of all possible variants. So check/date/format would be (in the case of ISO) an enum with the possibilities you think are useful to implement.

Testing, docu and comments are more important than a complete implementation. If someone needs a specific variant, we can add it.

tom-wa:
I've already implemented everything (possible combinations are generated based on rules). 
i just need a simple way to allow the user to say that the date should be validated as e.g. a "complete calendar date" or just a "time" which would allow complete, reduced and truncated time formats

markus2330:
What is wrong with the enum?

tom-wa:
oh, you mean like literally having e.g."complete calendar date" as possible enum value ? 
that would work, but there would be many implementation decisions regarding defaults (e.g. if not specified otherwise both `basic` and `extended` representations are valid)

markus2330:
> oh, you mean like literally having e.g."complete calendar date" as possible enum value

Yes. If you think names without spaces are more fancy, we can use them instead.

> many implementation decisions regarding defaults

They are very important, but I cannot really help you there. You already know the standard better. Ideally, you should follow suggestions from the standard.

@beku Do you have an opinion how to specify dates?

I think we can give you much better feedback once you wrote a README.md explaining all details and all defaults.

src/plugins/date/README.md
```diff
@@ -0,0 +1,101 @@
+- infos = Information about the date plugin is in keys below
+- infos/author = Thomas Waser <thomas.waser@libelektra.org>
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/recommends =
+- infos/placements = presetstorage postgetstorage
+- infos/status = recommended productive maintained reviewed conformant compatible coverage specific unittest tested libc final
```
markus2330: A lot of promises here ;)

markus2330:
Thank you, great job!

PR #1204, opened 2016-12-18
Curlget put

tom-wa:
# Purpose

added set/upload support + turned curlget into a resolver

# Checklist

- [ ] I added unit tests

### DIFF
--- a/src/error/specification
+++ b/src/error/specification
@@ -1019,3 +1019,10 @@ severity:error
 ingroup:plugin
 module:curlget
 macro:CURL_FETCH_FAILED
+
+number:167
+description:Invalid plugin configuration.
+severity:error
+ingroup:plugin
+module:curlget
+macro:CURL_CONFIG_INVALID
--- a/src/error/specification
+++ b/src/error/specification
@@ -1005,3 +1005,17 @@ severity:error
 ingroup:plugin
 module:curlget
 macro:CURL_UPLOAD_FAILED
+
+number:165
+description:Couldn't fetch file from remote host, try falling back to local copy.
+severity:warning
+ingroup:plugin
+module:curlget
+macro:CURL_LOCAL_FALLBACK
+
+number:166
+description:Failed to fetch configuration file.
+severity:error
+ingroup:plugin
+module:curlget
+macro:CURL_FETCH_FAILED
--- a/src/error/specification
+++ b/src/error/specification
@@ -1000,28 +1000,28 @@ module:fcrypt
 macro:FCRYPT_FUTIMENS
 
 number:164
-description:Failed to upload config file.
+description:failed to upload config file.
 severity:error
 ingroup:plugin
 module:curlget
 macro:CURL_UPLOAD_FAILED
 
 number:165
-description:Couldn't fetch file from remote host, try falling back to local copy.
+description:could not fetch file from remote host, try falling back to local copy.
 severity:warning
 ingroup:plugin
 module:curlget
 macro:CURL_LOCAL_FALLBACK
 
 number:166
-description:Failed to fetch configuration file.
+description:failed to fetch configuration file.
 severity:error
 ingroup:plugin
 module:curlget
 macro:CURL_FETCH_FAILED
 
 number:167
-description:Invalid plugin configuration.
+description:invalid plugin configuration.
 severity:error
 ingroup:plugin
 module:curlget
--- a/src/error/specification
+++ b/src/error/specification
@@ -998,3 +998,10 @@ severity:warning
 ingroup:plugin
 module:fcrypt
 macro:FCRYPT_FUTIMENS
+
+number:164
+description:Failed to upload config file.
+severity:error
+ingroup:plugin
+module:curlget
+macro:CURL_UPLOAD_FAILED
--- a/src/plugins/curlget/CMakeLists.txt
+++ b/src/plugins/curlget/CMakeLists.txt
@@ -1,9 +1,9 @@
 if (DEPENDENCY_PHASE)
 	find_package(CURL)
-
 	if(NOT CURL_FOUND)
 		remove_plugin(curlget "Curl-dev not found")
 	endif ()
+	find_package(OpenSSL)
 endif ()
 		
 add_plugin (curlget
@@ -12,7 +12,9 @@ add_plugin (curlget
 		curlget.c
 	INCLUDE_DIRECTORIES 
 		${CURL_INCLUDE_DIR}
+		${OPENSSL_INCLUDE_DIR}
 	LINK_LIBRARIES
 		${CURL_LIBRARIES}
+		${OPENSSL_LIBRARIES}
 	ADD_TEST
 	)
--- a/src/plugins/curlget/CMakeLists.txt
+++ b/src/plugins/curlget/CMakeLists.txt
@@ -4,6 +4,13 @@ if (DEPENDENCY_PHASE)
 		remove_plugin(curlget "Curl-dev not found")
 	endif ()
 	find_package(OpenSSL)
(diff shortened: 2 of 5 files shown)


tom-wa:
for some reason i can't link any packages against openssl as long as the crypto plugin is present, not sure what i'm doing wrong here. after removing the crypto plugin everything builds fine

src/error/specification
```diff
@@ -998,3 +998,24 @@ severity:warning
 ingroup:plugin
 module:fcrypt
 macro:FCRYPT_FUTIMENS
+
+number:164
+description:Failed to upload config file.
```
markus2330: please start with small letters.

src/error/specification
```diff
@@ -998,3 +998,24 @@ severity:warning
...
+severity:error
+ingroup:plugin
+module:curlget
+macro:CURL_UPLOAD_FAILED
+
+number:165
+description:Couldn't fetch file from remote host, try falling back to local copy.
```
markus2330: could not

src/plugins/curlget/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 if (DEPENDENCY_PHASE)
 	find_package(CURL)
-
 	if(NOT CURL_FOUND)
 		remove_plugin(curlget "Curl-dev not found")
 	endif ()
+	find_package(OpenSSL)
```
markus2330: what if OpenSSL was not found?

Why is it needed? Why use md5?

To fix the OpenSSL problem you need to pack all arguments you need later together here.
E.g. set (CURLGET_INCLUDE_DIRS ${CURL_INCLUDE_DIR}	${OPENSSL_INCLUDE_DIR})

src/plugins/curlget/CMakeLists.txt
```diff
@@ -12,7 +12,9 @@ add_plugin (curlget
 		curlget.c
 	INCLUDE_DIRECTORIES 
 		${CURL_INCLUDE_DIR}
```
markus2330: and then use ${CURLGET_INCLUDE_DIRS} here

src/plugins/curlget/CMakeLists.txt
```diff
@@ -12,7 +12,9 @@ add_plugin (curlget
...
+		${OPENSSL_INCLUDE_DIR}
 	LINK_LIBRARIES
 		${CURL_LIBRARIES}
```
markus2330: Same for libraries..

src/plugins/curlget/README.md
```diff
@@ -4,33 +4,18 @@
...
 - infos/description = mount remote config files via curl
 
 ## Example ##
 
-    cat /tmp/curltest.ini
-    cat: /tmp/curltest.ini: No such file or directory
+### GET + POST ###
 
-    kdb mount -R noresolver /tmp/curltest.ini system/curltest ini curlget url="https://raw.githubusercontent.com/ElektraInitiative/libelektra/master/src/plugins/ini/ini/plainini"
+kdb mount -R curlget -c get="http://127.0.0.1:8000/curltest.ini",upload="http://127.0.0.1:8000",user="thomas",password="pass",upload/method="POST",upload/postfield="file" /tmp/curltest.ini system/curl ini
```
markus2330: I think the old example (without upload) is still relevant, please describe them both

src/plugins/curlget/curlget.c
```diff
@@ -8,12 +8,196 @@
  */
 
 #include "curlget.h"
-
+#define _XOPEN_SOURCE
```
markus2330: Please add comments why you need the includes + _XOPEN_SOURCE

src/plugins/curlget/curlget.c
```diff
@@ -8,12 +8,196 @@
...
+		else if (!(strcmp (keyString (key), "FTP")))
+		{
+			data->uploadMethod = FTP;
+		}
+		else
+		{
+			data->uploadMethod = NA;
+		}
+	}
+	key = ksLookupByName (config, "/prefer", KDB_O_NONE);
```
markus2330: Please explain config in the README.md

src/plugins/curlget/curlget.c
```diff
@@ -32,46 +219,245 @@ int elektraCurlgetGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA
...
-	if (!rkey)
+	Data * data = elektraPluginGetData (handle);
+	if (!data)
 	{
-		return 0;
+		return -1;
 	}
-	const char * url = keyString (rkey);
+	int fd = 0;
+	/*	if (!*(data->lastHash))
```
markus2330: remove it?

src/plugins/curlget/curlget.c
```diff
@@ -8,12 +8,196 @@
...
+#include <openssl/md5.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <time.h>
+#include <unistd.h>
+
+#define TMP_NAME "/tmp/elektraCurlTempXXXXXXXXXXXX"
```
markus2330: Is this used?

tom-wa:
currently the plugin keeps a local copy of the file which is read as a "fallback" if the remote file can't be fetched and i'm about the add an option that only operates on temporary files an doesn't keep any local version of the file. 
what should be the default behaviour ?

markus2330:
> what should be the default behaviour ?

If a URL is given for mounting (via /path), I suppose there should be no corresponding local file (only tmp to be downloaded every time).

Otherwise, I suppose the file should exist and be used.

markus2330:
@tom-wa any progress?

src/plugins/curlget/CMakeLists.txt
tom-wa: works fine for crypto/fcrypt, but i still got issues with base64
```
CMake Error at cmake/Modules/LibAddPlugin.cmake:185 (message):
   Internally inconsistency, plugin elektra-crypto_openssl got different
   values for variable ARG_INCLUDE_DIRECTORIES:
   '/usr/include;/home/thomas/Dev/libelektra/src/plugins/base64/' !=
   '/usr/include'
   cmake/Modules/LibAddPlugin.cmake:244 (restore_variable)
   src/plugins/crypto/CMakeLists.txt:162 (add_plugin)
```

markus2330: Same pattern: you have to make sure that only a single variable is passed to every argument of `addPlugin`. Or any problems by doing so?

tom-wa:
```
 10         if(OPENSSL_FOUND AND CURL_FOUND)
 11             set (CURLGET_INCLUDE_DIRS ${CURL_INCLUDE_DIR} ${OPENSSL_INCLUDE_DIR})
 12             set (CURLGET_LIBRARY_DIRS ${CURL_LIBRARIES} ${OPENSSL_LIBRARIES})
 13         endif ()
...
 20         INCLUDE_DIRECTORIES
 21                 ${CURLGET_INCLUDE_DIRS}
 22         LINK_LIBRARIES
 23                 ${CURLGET_LIBRARY_DIRS}
```

anyway seems like it's a different issue. running "configure" in ccmake with all plugins yields the error mentioned before, but curlget and crypto_openssl + base64 are added to the plugins, and running "configure" again doesn't yield the error anymore + all plugins build file. 
i'm sure whats going on here though, because i'm still not familiar with cmake

tom-wa:
i guess the plugin needs a new name now.
would `curlresolver` be ok  ?

markus2330:
jenkins build fast please

PR #1205, opened 2016-12-18
Multifile

tom-wa:
# Purpose

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/src/plugins/multifile/multifile.c
+++ b/src/plugins/multifile/multifile.c
@@ -163,17 +163,29 @@ static MultiConfig * initialize (Plugin * handle)
 	Key * stayAliveKey = ksLookupByName (config, "/stayalive", 0);
 	MultiConfig * mc = elektraCalloc (sizeof (MultiConfig));
 	if (resolverKey)
-		mc->resolver = elektraStrDup (keyString (resolverKey));
+	{
+    	    mc->resolver = elektraStrDup (keyString (resolverKey));
+	}
 	else
+	{
 		mc->resolver = elektraStrDup (DEFAULT_RESOLVER);
+	}
 	if (storageKey)
+	{
 		mc->storage = elektraStrDup (keyString (storageKey));
+	}
 	else
+	{
 		mc->storage = elektraStrDup (DEFAULT_STORAGE);
+	}
 	if (patternKey)
+	{
 		mc->pattern = elektraStrDup (keyString (patternKey));
+	}
 	else
+	{
 		mc->pattern = elektraStrDup (DEFAULT_PATTERN);
+	}
 	if (stayAliveKey) mc->stayAlive = 1;
 	mc->directory = elektraStrDup (path);
 	mc->childBackends = ksNew (0, KS_END);
@@ -247,6 +259,18 @@ static Codes updateFiles (MultiConfig * mc, KeySet * returned, Key * parentKey)
 	ret = glob(pattern, 0, NULL, &results);
 	if(ret != 0)
 	{
+	    if(ret == GLOB_NOSPACE)
+	    {
+		ELEKTRA_SET_ERRORF(ELEKTRA_ERROR_MULTI_GLOB_FAILED, parentKey, "glob(%s) ran out of memory", pattern);
+	    }
+	    else if(ret == GLOB_ABORTED)
+	    {
+		ELEKTRA_SET_ERRORF(ELEKTRA_ERROR_MULTI_GLOB_FAILED, parentKey, "glob(%s) failed with a read error", pattern);
+	    }
+	    else if(ret == GLOB_NOMATCH)
+	    {
+		ELEKTRA_SET_ERRORF(ELEKTRA_ERROR_MULTI_GLOB_FAILED, parentKey, "glob(%s) failed with no matches", pattern);
+	    }
 	    return ERROR;
 	}
 	struct stat sb;
@@ -271,6 +295,7 @@ static Codes updateFiles (MultiConfig * mc, KeySet * returned, Key * parentKey)
 			{
 				if (!mc->stayAlive)
 				{
+				    	globfree(&results);
 					return ERROR;
 				}
 				else
@@ -291,6 +316,7 @@ static Codes updateFiles (MultiConfig * mc, KeySet * returned, Key * parentKey)
 	
 	    }
 	}
+	globfree(&results);
 
 	ksRewind (mc->childBackends);
 	ksRewind (found);
@@ -362,12 +388,12 @@ static Codes doGetStorage (MultiConfig * mc, Key * parentKey)
 		{
 			if (s->ks) ksDel (s->ks);
 			s->ks = ksDup (readKS);
-			;
 			rc = SUCCESS;
-			;
 		}
 		else
+		{
 			rc = ERROR;
+		}
 		ksDel (readKS);
 	}
 	keySetName (parentKey, keyName (initialParent));
@@ -442,11 +468,17 @@ int elektraMultifileGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned, Key 
 	}
 	elektraPluginSetData (handle, mc);
 	if (rc == SUCCESS)
+	{
 		return 1;
+	}
 	else if (rc == NOUPDATE)
+	{
 		return 0;
+	}
 	else
+	{
 		return -1;
+	}
 }
 
 static Codes resolverSet (MultiConfig * mc, Key * parentKey)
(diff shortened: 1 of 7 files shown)


markus2330:
Great to see progress here!

Please make sure to have a description in README.md so that I can check the goal.

tom-wa:
currently it just dispatches to the resolver and storage plugins and updates the backends on filesystem changes. nothing really worth mentioning. 

what kind of globbing pattern are supported ? deriving the mountpoint/path from the filename + pattern would get quite complicated for everything more complex than `*.ext`

 
is there any way to get a list of all keys with a `sync` flag without going through the whole keyset ?

markus2330:
> currently it just dispatches to the resolver and storage plugins and updates the backends on filesystem changes. nothing really worth mentioning.

I am talking about the interface to the user (plugin config, how it processes the files and expected limitations).

> what kind of globbing pattern are supported ? deriving the mountpoint/path from the filename + pattern would get quite complicated for everything more complex than *.ext

`*.ext` should be okay. That said: glob(3) allows you to iterate over the matches it gets, so why should another globbing expression be more complicated?

> is there any way to get a list of all keys with a sync flag without going through the whole keyset ?

Do not worry, iterating over keys is very cheap, and checking the flag, too.

tom-wa:
>*.ext should be okay. That said: glob(3) allows you to iterate over the matches it gets, so why should another globbing expression be more complicated?

the problem isn't really the globbing, but the mountpoint e.g. getting `local` from `local.conf` and `*.conf`
while it's easy for `*.ext` pattens with multiple wildcards can get quite complicates (e.g. files in `/sys` or `/proc`)

markus2330:
Exactly such decisions belong to the README.md!

The problem is you are trying to solve two problems at once. The first problem is how to specify which files to include. The second problem is how to assign the mountpoints.

For pk it is trivial: The glob is `*` and the file names are the mountpoints. So it would be `local.conf` as mountpoint. When using keyAddBaseName `/` in mountpoints will be correctly encoded.

I would not introduce further complications (such as removing .conf) because most software actually want all files to be merged into the same tree. Then we would need meta data to remember where each setting came from. But let us handle this issue later..

tom-wa:
should the parentkey of each child-backend also be added to the keyset ? i could think of some scenarios where it would be practical 

should the multifile resolver be allowed to create new configuration files if necessary ?

markus2330:
> should the parentkey of each child-backend also be added to the keyset ? i could think of some scenarios where it would be practical

In Elektra the parentKey of every backend belongs to the responsibility of the respective backend. It should only be added if the child-backend adds it.

> should the multifile resolver be allowed to create new configuration files if necessary ?

Yes, please. It should also delete configuration files if no keys are passed to a child-backend. But both actions should already be done by the resolver, so hopefully you do not need any extra code for that.

tom-wa:
>. But both actions should already be done by the resolver,

i currently have a flag to switch between calling and not call child-backends if the key passed keyset would be empty and manually remove the file at the commit flag (actually to not even call child backends if no changes have happened. could that cause any problems ?

markus2330:
> i currently have a flag to switch between calling and not call child-backends if the key passed keyset would be empty and manually remove the file at the commit flag (actually to not even call child backends if no changes have happened. could that cause any problems ?

This is the behavior of Elektra's core, too. It is expected that plugins are not called if nothing is to do. This is especially important if there are many config files with changes in only one of them. This is a common situation and should be supported well, especially in multifile resolvers.

I think you can remove the flag for different behavior, it makes testing much more complicated.

tom-wa:
ok, i was just a little bit confused by the ksGetSize == 0 check in resolver kdbSet

markus2330:
The `ksGetSize == 0` is for removal of the configuration file when there are no settings. But it is already done by the resolver, no code duplication needed.

tom-wa:
what should happen if only one of multiple child-backends fails ? currently i have a flag which either fails everything or tries to continue. it seems useful to me for kdbGet.

markus2330:
> what should happen if only one of multiple child-backends fails ? currently i have a flag which either fails everything or tries to continue. it seems useful to me for kdbGet.

kdbSet: If something fails before commit, nothing should be done (call `kdbError` of all resolvers).

kdbGet: Tricky question, maybe we should deliver as much as possible even on failures. Discussion see #1291. Let us discuss it in the next meeting.

src/plugins/multifile/multifile.c
```diff
@@ -0,0 +1,693 @@
...
+		}
+		else if (s->rcStorage != 1)
+			continue;
+		
+		keySetName (parentKey, s->parentString);
+		keySetString (parentKey, s->fullPath);
+		Plugin * resolver = s->resolver;
+		int r = resolver->kdbSet (resolver, s->ks, parentKey);
+		rc = rvToRc (r);
+		if (rc == ERROR) break;
```
e1528532: i think in that case it would be clearer if we add this condition to the loop condition

src/plugins/multifile/multifile.c
```diff
@@ -0,0 +1,693 @@
...
+
+static int diffOrNeedSync (KeySet * ks, KeySet * checkKS)
+{
+	if (ksGetSize (ks) != ksGetSize (checkKS)) return 1;
+	ksRewind (ks);
+	ksRewind (checkKS);
+	Key * key = NULL;
+	Key * check = NULL;
+	int done = 0;
+	while (!done)
```
e1528532: i'd rewrite this loop a bit, having a !done variable in a loop condition which gets never set in the loop body is not the best style i think, since it looks like the result will be either 0 or 1 in the end, i'd do it like
int ret = -1;
while (ret == -1) {
// assign to ret instead of return statement
}
return ret;

src/plugins/multifile/multifile.c
```diff
@@ -0,0 +1,693 @@
...
+		}
+		keyDel (cutKey);
+		ksDel (cutKS);
+	}
+}
+
+int elektraMultifileSet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_UNUSED, Key * parentKey ELEKTRA_UNUSED)
+{
+	// set all keys
+	// this function is optional
```
e1528532: i guess when can remove these template comments once we implemented them

src/plugins/multifile/multifile.c
```diff
@@ -0,0 +1,693 @@
...
+}
+
+static void closeBackend (SingleConfig * s)
+{
+	if (!s) return;
+	if (s->fullPath) elektraFree (s->fullPath);
+	if (s->filename) elektraFree (s->filename);
+	if (s->parentString) elektraFree (s->parentString);
+	if (s->tmpFilename) elektraFree (s->tmpFilename);
+	if (s->resolver) elektraPluginClose (s->resolver, NULL);
```
e1528532: most of the other plugins use elektraPluginClose (<plugin>, 0) instead of NULL, not sure whats the right thing to do. And by the way, are all those null checks necessary? I think for plugins they might be fine, if plugin developers have no checks, but i guess elektraFree simply does nothing if its already null?

markus2330: Exactly, I clarified it in e1eb25e06e02747d6c4d543c98c7e98542ad82b3

e1528532:
Ok i managed to make it compile, there was a cmake option missing which seems to be required by clang, heres a patch:

```
diff --git a/src/plugins/multifile/CMakeLists.txt b/src/plugins/multifile/CMakeLists.txt
index e1af1ac7a..44b412593 100644
--- a/src/plugins/multifile/CMakeLists.txt
+++ b/src/plugins/multifile/CMakeLists.txt
@@ -4,5 +4,7 @@ add_plugin (multifile
 	SOURCES
 		multifile.h
 		multifile.c
+	LINK_ELEKTRA
+		elektra-kdb
 	ADD_TEST
 	)
```

i then tried to follow the example but unfortunately the keys didn't show up in kdb ls then (and sudo is required for mounting)

```
mbp-armin% sudo kdb mount -R multifile -c storage="ini",pattern="*.ini",resolver="resolver" /tmp/multitest user/multi
Note that absolute paths are still relative to their namespace (see `kdb info resolver`).
Only system+spec mountpoints are actually absolute.
Use `kdb file user/multi` to determine where the file(s) are.
Use `-q` or use `kdb set /sw/elektra/kdb/#0/current/quiet 1` to suppress infos.
mbp-armin% kdb file user/multi
mbp-armin% kdb ls user/multi
mbp-armin%
```

any idea? created the files exactly like stated in the examples

markus2330:
Did you also try if the examples are executable with shell recorder?

@e1528532 thanks for finding out!

@tom-wa do you actually need functionality from `kdb`? Such reverse dependencies are a bit ugly.

@tom-wa please update docu, "one-line description of multifile" and so on should be replaced ;)

tom-wa:
i guess the issue is that you use /tmp with the user namespace.
```
Note that absolute paths are still relative to their namespace (see `kdb info resolver`).
Only system+spec mountpoints are actually absolute.
```

markus2330:
Thank you! infos/status is exaggerated

PR #1226, opened 2016-12-23
rest-service: improvements

Namoshek:
# Purpose

Fixes a minor issue with the rest-backend and adds some nice features/improvements to the rest-frontend. Details as always in the commit messages.

# Checklist


Namoshek:
jenkins build homepage please

markus2330:
Thank you, great job!

Btw. when pressing the raw button I had more the expectation I will get it in the format as uploaded, but it seems to be always in json?

But it is nice to have something where middle button works (open in background tabs)

Namoshek:
That's a good point actually, I can change it to this behavior (and/or add a button *API* that shows the current one?), but it requires an addition to the API (currently the upload format is not added to the output of listings) - it is fully compatible though.

markus2330:
Yes, upload format would be an nice info in the search interface. I would rather change the behavior, getting every snippet in one specific (non-selectable format) does not sound too useful. I added it in #1228

Namoshek:
Hm, I reduced the search interface because it was bloated (slug removed). Adding the format would make it crowded again (e.g. think of `augeas lens=Hosts.lns`, which is a short lens name).

markus2330:
That is a good point. Maybe we should have org, app, scope, and slug as a single column and collapse too long formats? Or show it by hovering somewhere?

Btw. if you want you can self-assign you to #1228 and of course you are free to do the tasks there ;)

Namoshek:
org, app and scope are splitted so you can sort by any of these, otherwise I would only display the key of course. I think making a hover somewhere is reasonable, but also intransparent (you'll only find this by luck). Gonna have to figure a better way out. :)

Issue #1083, opened 2016-11-12
Installing Java Bindings


e1528532:
I wanted to try out the java bindings for my further work i inteded to do on the java bindings and related things. These are my experiences with the bindings, some things i was able to resolve, other things are still unclear. I will then try to update some of the documentations of jni and jna to help other users to get started easier.

# Getting Started

So first i had no real idea about how to install the java bindings. I quickly figured out that its probably a binding, and searched for bindings on the main page of libelektra. This lead me to the bindings folder which actually included the jna folder. On the tutorial main page the jna bindings are not listed though, i think we should add it there https://github.com/ElektraInitiative/libelektra/tree/master/doc/tutorials too.

So then i checked the documentation on https://github.com/ElektraInitiative/libelektra/tree/master/src/bindings/jna . Having to have jna.jar in the classpath was clear for me, but i wondered about the mentioned libelektra.jar . I wasn't really able to locate it anywhere in the system and that lead me to thinking that i have to compile the jni plugin which i thought will install this libelektra.jar in my system. I know that the jna documentation states

>or this directory which contains the elektra subdirectory

but i didn't figure out initially that this is the source of the libelektra.jar which i was searching for. 

# Stuggling with compiling the JNI plugin

## Setting the Java Path
So since i initially though that i need the jni plugin for the libelektra.jar i tried to compile elektra with this plugin. I headed over to the plugin documentation on https://github.com/ElektraInitiative/libelektra/tree/master/src/plugins/jni and then tried to compile it. I saw all these instructions with default-java and using ldconfig, but since i am using macOS Sierra these things are not available for me. So i decided to ignore that first and try to compile it right away using the stated command:

> cmake -DPLUGINS="...;jni" /path/to/libelektra

As this was the first plugin i tried to compile apart from just compiling elektra with the default plugins i was really confused about the correct cmake parameters (yes i haven't extensively read the documentation yet at that point...). So i actually ended up trying to compile elektra with 

>cmake -DPLUGINS="...;jni" ..

being in the build directory i created on the initial installation. Yes i actually kept the ... and then the plugin kept failing to compile. I always checked out the CMakeError.log file but it was totally non-obvious to figure out whats missing. Since the documentation was about changing some system paths before which i haven't done yet i figured that thats whats missing. But using macOS i had no idea where it searches for the files and what i have to change, until i saw in the CMakeError.log that it searches the java headers at the path 

[...]

PR #1238, opened 2016-12-28
jna-bindings: maven support

e1528532:
restructure to a more maven-ish project layout
fix some potential memory leak in Key and KeySet finalizer
add simple KDB test
make it possible to build it via maven
a lot of small code cleanups / convention improvements suggested by
sonarlint
update readmes to work with new project layout
update cmake to work with new project layout

# Purpose

Mavenized the jna bindings along with several code improvements

# Checklist

- [X] I added unit tests

@markus2330 please review my pull request. There is a small problem yet as the cmake files assumes the /usr/share/java directory as the target for the built libelektra.jar, and also searches for the required jna library there. However this directory is protected in macOS (via the system integrity protection) so its not possible to use that directory. I haven't yet found out which directory is intended in macOS for shared java libraries, maybe /Library/Java? For now i specified paths manually. Then we would have separate cmake file for mac and linux. Apart from that i think its working fine, imho much easier to use maven than manually compiling around.


src/bindings/jna/README.md
```diff
@@ -17,27 +20,48 @@ then you can compile and run [HelloElektra](HelloElektra.java):
 
     javac HelloElektra.java && java HelloElektra
 
-For plugin development, see [plugins](elektra/plugin)
+Alternatively you can also specify the classpath directly, both during compilation and execution. Note that in this example, we have the elektra and jna libraries in different directories, as mac doesn't allow writing to /usr/share/java in never versions. Also note its important in that case to include this directory containing the compiled HelloElektra.class when executing it, otherwise it will not be found:
```
markus2330: Please add some line breaks in between of longer text. It makes reading diffs easier.

src/bindings/jna/README.md
```diff
@@ -17,27 +20,48 @@ then you can compile and run [HelloElektra](HelloElektra.java):
...
+	javac -cp ./libelektra4j.jar:./libelektra4j/jna.jar HelloElektra.java
+	java -cp .:./libelektra4j.jar:./libelektra4j/jna.jar HelloElektra
+
+For plugin development, see [plugins](libelektra4j/plugin)
 and also [here](/src/plugins/jni) for more information.
 
+### Maven ###
+
+In case you want to use maven as your build system, simply go into the libelektra4j subfolder and execute mvn install. This will install the plugin into your local maven repository. Afterwards you can include it as a dependency into your maven project, and given that libelektra is actually installed on your system you can use it by including the maven dependency:
+	<groupId>org.libelektra</groupId>
```
markus2330: Some markdown renderer need a newline between text and code blocks.

src/bindings/jna/README.md
```diff
@@ -17,27 +20,48 @@ then you can compile and run [HelloElektra](HelloElektra.java):
...
+	<artifactId>libelektra4j</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
```
markus2330: Should we really use a version number so low?

What about simply using Elektra's version number?

What does the `-SNAPSHOT` mean?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
```diff
@@ -333,120 +397,131 @@ public Key nextMeta() {
...
+	public int rel(final Key other) {
 		return Elektra.INSTANCE.keyRel(get(), other.get());
 	}
 
 	/**
 	 * Helper function to check if synchronization is necessary
-	 * @return 1 if needs sync, 0 if no change done and -1 in case of
-	 * 		a null pointer
+	 *
+	 * @return 1 if needs sync, 0 if no change done and -1 in case of a null pointer
```
markus2330: doxygen supports `@retval` for such cases.

src/bindings/jna/libelektra4j/src/test/java/org/libelektra/README.md
```diff
@@ -0,0 +1,22 @@
+Unit tests can be run after importing jUnit, JNA and the libelektra4j java library into a project (eclipse, netbeans, intelliJ, ...). Alternatively, you can use maven to both build the project or just run the unit tests.
```
markus2330: Is it necessary to have this duplicated?

src/bindings/jna/README.md
e1528532: 0.0.1-SNAPSHOT the default for new maven projects. I agree that a higher version number makes sense, but I'm not sure if it really makes that much sense to use the same as elektra because this introduces the problem of having to synchronize this version number with elektra's one somehow to stay consistent in future releases. But if there are no changes to the jna binding for a new elektra version, does it make sense to update the version number nevertheless? Since it looks like some things are not really perfect yet with the jna bindings (see the readme) i'd use something like 0.9.0 or similar for now.

SNAPSHOT is some maven convention, usually you'd use -SNAPSHOT for development versions and then switch to -RELEASE for, well, release versions. It is not required to call it like that though, so I could drop it completely if you prefer.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
e1528532: In Javadoc you'd use @return, are the javadocs even parsed with doxygen? Because if not, i'm not sure if using retval makes much sense currently.

src/bindings/jna/libelektra4j/src/test/java/org/libelektra/README.md
e1528532: No, it was like that when i started working on it so i left it in there for now, but the "outermost" readme in the jna directory already contains test instructions. So I'd drop those here.

src/bindings/jna/README.md
e1528532: done

src/bindings/jna/README.md
markus2330: > But if there are no changes to the jna binding for a new elektra version, does it make sense to update the version number nevertheless?

In many Elektra releases also the C API has no changes, the number increments nevertheless. If there are changes in the C API we will add the new functionality in all bindings, too. So it makes sense to expose the version to the user. If you think we need an extra number to indicate incompatibilities of the bindings, we can add such a (single) number. But the easiest way is to use exactly the same number from the build system (use `KDB_VERSION` with `configure_file`, see `kdb.h.in`). It also would be consistent with the other bindings.

> Since it looks like some things are not really perfect yet with the jna bindings (see the readme) i'd use something like 0.9.0 or similar for now.

We should definitely avoid a number higher than Elektra's.

e1528532:
just noticed the jna binding site looks bad in our homepage https://www.libelektra.org/bindings/jna so i did some further markdown improvements to make it look better (although i'm not a markdown expert yet)

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
markus2330: No, the c++ bindings are the only ones included in Doxygen. But it would be great to have other bindings included, too. Can you check if we simply can add the the other bindings to doxygen?

markus2330:
Thank you for the markdown improvements! We need a much better README anyway, maybe a small tutorial like it was done for Ruby by @BernhardDenner

markus2330:
jenkins build homepage please

markus2330:
Lets see how the new homepage looks like... (with your improvements)

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
e1528532: I checked out doxygen and it supports javadocs basically, though not as nicely but good enough for now i think. It does parse the @return correctly, so i think theres no need to change that.

In my last commit i added the jna folder to the Doxyfile. But i had to set this EXTRACT_ALL option to true in addition, otherwise it seems not to extract the javadocs from the source files. However, this might has some bad implications on the other documentation, do you have any idea why it didn't worked without changing EXTRACT_ALL? I also set the optimize for c output to false, as it mentioned in the config file you should only set this to true if you only have c files, but we have c, c++, java, and there are also some python files now in my documentation (i suspect those are also extracted due to EXTRACT_ALL?)

markus2330: In general I like the inclusion of the java API in doxygen. It is only included in Classes/Namespaces below "org", which is nice and easily distinguishable to C++ docu.

> otherwise it seems not to extract the javadocs from the source files

Very weird, is this a doxygen bug? EXTRACT_ALL would require us to put `@internal` to every undocumented function. Otherwise it pulls in a lot of unrelated things, like python_*, Edge,.. Adding @internal to all functions does not seem feasible?

For me (doxygen 1.8.8) all documented classes/interfaces are also added without EXTRACT_ALL. The namespaces (=packages?), interfaces, and some Java classes are not, but this is not surprising as they are currently not documented. If I add some docu to org.libelektra.Elektra it is added to doxygen.

> optimize for c output to false, as it mentioned in the config file you should only set this to true if you only have c files

Yes, this should be okay. Did you find any difference at all when changing it?

src/bindings/jna/README.md
markus2330: I mean "empty line" instead of newline.

e1528532:
As discussed in https://github.com/ElektraInitiative/libelektra/issues/1258, i implemented the kdb test in a way it complies to the current kdbGet behavior for now.
Furthermore, as it was useful for this test, i implemented the delete method for the keyset iterator. So functionality wise it should be good again for now.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
e1528532: Doxygen version 1.8.13 here. The issue was that doxygen only includes classes which have a javadocs on the class themselve. In the jna bindings, most of the documentation was on methods, but not on the actual class, therefore it didn't include that file at all. So i added some basic class documentation on the important classes in order to make doxygen recognize them.

The removal of OPTIMIZE_C seems to affect how some components on the documentation are named, e.g. classes are called data structures as in C there are no classes. As we have some c++ documentation anyway, which supports classes, and also java, i think it makes sense to leave the optimize_XYZ options of doxygen disabled to keep it as generic as possible. E.g. OPTIMIZE_JAVA would rename the namespaces item in the html to packages which is also not what we want as c++ does not have packages. There is no real way though to have perfect naming for each language used.

markus2330:
Did you already try to upload this to maven central repository? I could not find it.

e1528532:
No, not yet. We had the discussion that it probably doesn't make much sense to upload it as it currently is, because elektra has to be installed on the system. If we upload it to the central repository, one might be tempted to use it directly without having elektra installed, which will not work then and thus discourages possible users. 
Also as I've never uploaded something to the central repository yet and it looks like there is quite some overhead associated with uploading libaries as seen in https://maven.apache.org/guides/mini/guide-central-repository-upload.html .
Looks like I'd have to get a maven repository on one of maven central's approved hosters (http://central.sonatype.org/pages/ossrh-guide.html) and then release it there, then it will be found via the central repository.

I'd estimate a few more hours to do that. Or we leave it as it currently is, everyone interested in using the jna binding can use the pom file in the elektra sources for an easy setup of a java project using elektra via jna, so its not like this PR is useless.

markus2330:
> No, not yet. We had the discussion that it probably doesn't make much sense to upload it as it currently is, because elektra has to be installed on the system.

There are plenty of [native libs](https://search.maven.org/#search|ga|1|g%3A%22com.nativelibs4java%22) in the repository. So it seems to make sense.

> If we upload it to the central repository, one might be tempted to use it directly without having elektra installed, which will not work then and thus discourages possible users.

The main reason of having it in the repository would be visibility. And if it is more visible, more people might install Elektra, and it would work out-of-the-box in more cases ;)

> Also as I've never uploaded something to the central repository yet and it looks like there is quite some overhead associated with uploading libaries as seen in https://maven.apache.org/guides/mini/guide-central-repository-upload.html .
Looks like I'd have to get a maven repository on one of maven central's approved hosters (http://central.sonatype.org/pages/ossrh-guide.html) and then release it there, then it will be found via the central repository. I'd estimate a few more hours to do that.

The main question is if *you* want to upload something there. If you want, this would be a good opportunity.

> Or we leave it as it currently is, everyone interested in using the jna binding can use the pom file in the elektra sources for an easy setup of a java project using elektra via jna, so its not like this PR is useless.

No, I did not want to imply that the PR is useless, even if we decide against uploading!

e1528532:
ok, as elektra will really profit from more (potential) users i will look into it, how i can bundle or require libelektra with maven and how the upload procedure would work.

e1528532:
I'll handle the maven repository uploading in another PR, i think that task is large enough to justify that. There is already some value with being able to use maven to make projects from our sources. 

So i think we can merge this, but we probably want to wait for @BernhardDenner to review this first.

markus2330:
Yes, let us keep this open for reviews.

markus2330:
Two small things:

- can you rename the test folder to tests (is called like that everywhere else)
- there is still a merge commit in the history?

e1528532:
in maven projects its usually called test, thats why i did pick that name. But if you want to rename it to tests for the sake of having uniform names, i can do that of course.
There merge commit is there as this is a rather old branch where i was not aware of the rebasing yet. I can make a new branch and cherry-pick the commits if you wish.

markus2330:
> in maven projects its usually called test, thats why i did pick that name. But if you want to rename it to tests for the sake of having uniform names, i can do that of course.

Yes, seems like the [standard directory layout](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) mandates to do so. Then let us better leave it that way.

> There merge commit is there as this is a rather old branch where i was not aware of the rebasing yet. I can make a new branch and cherry-pick the commits if you wish.

You can also rebase afterwards.

src/bindings/jna/libelektra4j/pom.xml
```diff
@@ -0,0 +1,43 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>org.libelektra</groupId>
+	<artifactId>libelektra4j</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+
+	<properties>
+		<javac.target>1.8</javac.target>
+        <java.version>1.8</java.version>
```
BernhardDenner: Whitespace issue

src/bindings/jna/README.md
BernhardDenner: From my point of view, the main issue is the whole build process.

Now we've two different ways to build the jna-bindings jar, with two different outcomes:

- the cmake way
- and the maven way

cmake builds a plain jar, containing only the compiled byte code / class files and maven, building a 'maven' aware jar file, containing the pom in addition.

I think this is one build process too much. Which one to use when? Are the two artefacts really the same? I would prefer the maven way, as this is currently the default within the Java world. Also an upload to maven central would be very nice. 
Suppose you have installed elektra already, e.g. through apt-get, yum or similar. Then a Java developer simply can start by adding the 'libelektra4j' dependency into his pom.xml and run 'mvn install'.

Also, currently the Unit tests aren't integrated in the cmake build process (as far as I've seen). So basically wrapping the 'mvn' calls in cmake would be the best solution from my point of view.

Then we should definitely use the elektra version as pom version. So an update to the jna bindings issues a new elektra version, whereas a new elektra version issues a new jna-bindings version.

What do you think?
@e1528532 @markus2330

src/bindings/jna/libelektra4j/src/test/java/org/libelektra/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Tested library versions are:
+	JNA: 4.2.2
+	jUnit: 4.12 [jUnit 3 is not supported]
+	hamcrest-core: 1.13 (required by newer jUnit versions)
+	Both libraries should work on version 4.0 too, though.
+
```
BernhardDenner: I think we could add a few lines about the tested Java version. I've used openJDK 1.8.0 x86_64

markus2330:
> From my point of view, the main issue is the whole build process.
> Now we've two different ways to build the jna-bindings jar, with two different outcomes
>
>    the cmake way
>    and the maven way
> make builds a plain jar, containing only the compiled byte code / class files and maven, building a 'maven' aware jar file, containing the pom in addition.
>
> I think this is one build process too much. Which one to use when? Are the two artefacts really the same? I would prefer the maven way, as this is currently the default within the Java world. Also an upload to maven central would be very nice.

I fully agree that an upload to maven central would be nice, so if mvn build is prerequisite for it we need maven anyway.

> Also, currently the Unit tests aren't integrated in the cmake build process (as far as I've seen). So basically wrapping the 'mvn' calls in cmake would be the best solution from my point of view.

The cmake-java has quite many limitations. I open to call maven from cmake. Let us discuss it in the next meeting.

src/bindings/jna/libelektra4j/pom.xml
e1528532: fixed

e1528532:
Ok so as we agreed in the meeting i've finally managed to make cmake compile the jna bindings using maven (cmake is not easy to tame given you don't know it...). Hard part was to separate building it to check for possible errors and installing it due to my lack of cmake knowledge, but make without install shouldn't install anything IMO.

It seems to work quite well now, except a last hassle: maven repositories are per user, but i usually use sudo make install as some files need root access to be installed successfully. However, in that case maven will install the jna bindings to the root's maven repository and the normal user won't be able to use it.

When you use make install as a normal user (switching this one build flag off so that that works), the jna bindings are in your repository as expected. There is not really a global local repository in maven, usually you'd have a (remote) company repository where everyone gets it dependencies from and install it there in a maven environment. So do you think it is fine that way, or not?

Apart from that, i have to add a call to maven test and adjust the documentation, then this PR can finally be closed, uploading it to maven central will be a different one.

markus2330:
> in that case maven will install the jna bindings to the root's maven repository and the normal user won't be able to use it.

Can you elaborate? This seems to be really weird? What is the purpose of this root's maven repository? Or do you mean it is installed in the home directory of the root's user? Isn't there a way to globally install java libraries (for all users on a system)?

> So do you think it is fine that way, or not?

Why not install the files as they were installed before? Is maven unable to install and find files installed in `/usr/share/java`?

The cmake integration is mainly used by distributions, and for them a installation into a home directory would be completely useless.

> Oh and i should add that cmake fills in the correct kdb version now for the pom file, so it'll be synced to the elektra version as we wanted.

Perfect!

e1528532:
>Can you elaborate? This seems to be really weird? What is the purpose of this root's maven repository? >Or do you mean it is installed in the home directory of the root's user? Isn't there a way to globally install java libraries (for all users on a system)?

I meant the latter, it would install it in the current user's repository (which is the root when using sudo). As far as i know there is no such possibility in maven to install it globally on a single machine easily. 

> Why not install the files as they were installed before? Is maven unable to install and find files installed in /usr/share/java?

Right, that is also still possible of course. maven doesn't pick up stuff in /usr/share/java per default, it searches things in its local repository (usually in <user>/.m2/repository unless configured otherwise) and loads things it doesn't find from remote repositories.

Maven central is included as a remote repository by defaut, so uploading it there would mean that everyone can get it without further hassle. So i guess for manual build its really the best to just put the jar file in /usr/share and maven users will get it via maven central without further configuration. So we should really aim to upload them there, probably automatically by the build server everytime we release a new version.

e1528532:
So my last improvements:

* Building a jar file and placing it in /usr/share/java as it was before. Except the jar file is built via maven and contains the pom file and other stuff required by maven. 
* Adding a test to the testing phase, which will be run when issuing make run_all while having the jna bindings active. It basically calls maven test which then executes all the tests of the bindings.
* Adjusted the readme, incorporating some notes i've made in #1083 and explaining the maven changes. Also some markdown improvements (at least it looks fine now in my sublime markdown preview plugin).
* Explain how you can install the dependency into your local maven repository easily while its not added to maven central yet (or if you don't want to use maven central for whatsoever reason) with a simple command from /usr/share/java. If you just want the jar, then you can skip this step (for the package maintainers). I think this is the best solution for the maven repository issue.

IMO this is pretty much finished now, except the maven central upload which i'll do in another PR.

src/bindings/jna/CMakeLists.txt
```diff
@@ -1,50 +1,52 @@
...
-		install_jar(libelektra share/java)
-
-	else(jna)
-		remove_binding(jna "jna jar file not found, searched in standard paths and /usr/share/java")
-	endif(jna)
-
-else()
-	remove_binding(jna "Java is not greater version 8, but ${Java_VERSION} was found")
-endif()
+find_package (Java 1.8 REQUIRED)
```
markus2330: Please also do not abort if no java found (but disable plugin, like it is done with maven.

Do we need this package at all? (What if java is installed private to maven, undetectable by CMake?)

src/bindings/jna/CMakeLists.txt
```diff
@@ -1,50 +1,52 @@
...
+find_program (MAVEN_EXECUTABLE mvn)
+
+if (NOT MAVEN_EXECUTABLE STREQUAL "MAVEN_EXECUTABLE-NOTFOUND")
```
markus2330: The `STREQUAL` shouldn't be necessary, according to cmake semantics strings ending with `-NOTFOUND` should evaluate to false.

src/bindings/jna/README.md
```diff
@@ -1,43 +1,121 @@
-A full java binding using JNA.
+A full java binding using JNA. This binding requires elektra to be installed on 
+the system to work.
 
 ## Usage ##
 
+To use the bindings in a java project, we have to include the jar file 
+libelektra-0.8.19.jar in the project. This jar is created upon build, if you
```
markus2330: Please keep this name generic with globbing chars (`*`) so that we do not have to update the text every version.

(Several more occurrences below)

markus2330: Alternatively we could provide a script to source doing these things below. (The script would have the version number when we use `configure_file`.

markus2330:
And please rebase, there is a file conflicting: `src/bindings/jna/README.md`

src/bindings/jna/README.md
markus2330: Not really changing much for the implementation, but maybe the style of `elektrify-*` is better suited. Then one can use `kdb run-java javaapp` (or similar for setting the CLASSPATH and so on).

src/bindings/jna/CMakeLists.txt
BernhardDenner: hmm, agree on that, maven required JAVA itself, so if Java isn't found maven will not work too.
From my point of view, the best way is to check the premisses for maven: exported JAVA_HOME and javac in PATH. This way cmake will disable the plugin in dependency phase, rather than failing hard at build time.

src/bindings/jna/README.md
BernhardDenner: yea, just remove the version number. This is common and KISS

BernhardDenner:
Looks good to me, however cmake build has some issues
- calls maven build several times
- development/maintenance experience will be bad, since source files will be copied, no dependencies to build added ...

But these are minor isses, which requires some better Cmake foo, which I also do not have ;)
I give +1 for this PR now, for the build process

e1528532:
@BernhardDenner do you mean that the jar file generated by cmake in /usr/share/java should include all dependencies? The only dependency is jna anyway, and its fairly easy to include them in the maven build as far as i know. Then the jar file placed in /usr/share/java would then already contain the jna.jar file and could be used right away.
The only issue i see for that is, that if you use my described command to import the jar file in your local maven repository, it shouldn't really have the dependency bundled so you can let maven manage it, eventually overwrite dependency versions, stuff like that. So then we'd have to generate both the version with dependencies included, and one without for a possible installation in a local repository. Not sure if thats that good. Or cmake gives you the version with dependencies bundled and later on when its uploaded to maven central it doesn't really matter for maven users anyway?

Also theoretically maven is only run once with the package goal, and if you run the tests, then its executed again for the tests. So for builds without tests its ran only once? Or do i miss something here?

src/bindings/jna/CMakeLists.txt
e1528532: The original behavior basically already did that, so i restored that. I think Java doesn't have to be found if maven would have its private jvm somehow and thus work, but i've never heard of such setup and i think its safer and more user friendly to check for its presence as bernhard also suggests.

src/bindings/jna/README.md
e1528532: ok i've done the kiss solution and just call it *version*, i guess people will get it. To be absolutely precise i've even added a small sentence that its equivalent to the elektra version.

src/bindings/jna/CMakeLists.txt
e1528532: simplified.

src/bindings/jna/README.md
markus2330: Note that in your solution it sometimes is *version*, and sometimes `*version*` (when within code), so users might be confused if they see markdown-only.

BernhardDenner:
no no. I mean the build dependencies for the jar, thus all *.java files in src/main/... normally if you add these source deps to the build artefact (jar) cmake detects changes to the src files and rebuilds the jar. As far as I know, this is not not the case.
But as I already sad: this is a bonus. Many someone with more cmake experience can help here.

The runtime deps can be resolved by maven. This is what it is used for ;)

e1528532:
ah thats what you mean. Yes you are right but i think maven does only recompile new files. Anyway i guess this is not critical and rather a small optimization which we can do if its required at some point.

btw i had to merge, rebasing gave me conflicts and more conflicts so i decided against it.

src/bindings/jna/README.md
e1528532: Yes i'm aware of that as far as i've seen there is no clean possibility to use the * inside code. I considered it to be easy to understand to use the appropriate version number as most developers know markdown anyway so i just did i that way. any better idea? If not, i think we can finally merge it.

markus2330:
Thank you great job!

I opened #1349 for a java detection issue (Or a leftover, I already suggested that Java is not needed here before)

Issue #920, opened 2016-09-05
Unify Spelling


petermax2:
# Problem

At the moment ( 2e0b9d518a22a65ef3713db67d9f2eeac1580a4c ) there is no consistent usage of the word "metakey" within the repo.
- "meta key" is used 103 times
- "meta-key" is used 20 times
- "metakey" is used 298 times

So "metakey" is the most popular variant.
# Desired State

Consistent usage of the "metakey".

NOTE: Maybe other words and phrases are affected too. Please comment below.

Namoshek:
I think `mountpoint` / `mount point` is also affected (I never know how to write it, so I simply assume that).

markus2330:
Thank you for writing the proposal! I think we should go for the one-word variant, in long term popular words tend to this variant.

Note that doc/help/elektra-glossary.md is a good place to define how a word should be written and how it is meant.

Other words:
metaname
metavalue
metadata
basename
dirname
command-line
timestamp
startup
lookup
cleanup

markus2330:
Thanks for reporting! Should be fixed, the words are now written together.

markus2330:
What to do about key name and key value?

There are also constructs like metakey name.

e1528532:
for keyname a quick n dirty check gave me the following results:

```
grep -r -i 'keyname' ./* | wc -l
    3878
grep -r -i 'key-name' ./* | wc -l
     237
grep -r -i 'key name' ./* | wc -l
    1275
```

and the results for key value

```
grep -r -i 'key value' ./* | wc -l
     435
grep -r -i 'keyvalue' ./* | wc -l
    1093
grep -r -i 'key-value' ./* | wc -l
      25
```

@markus you said in a pullrequest "Please use key-?name consistently" so i assume we use the key-name variant in that case? or use keyname as in this issue you suggested using the variant where the words are written together (seems to be used very often throughout all files,  but not that much in the documentation, i guess thats because of variables being called keyname or keyName similar?)

markus2330:
Thanks for looking at it! For the PR pick anything, but keep consistent within the PR. Let us discuss it in the next meeting which name we choose for whole Elektra. Maybe we should even use `key` instead of `keyname` which is much more standard.

markus2330:
"key name" because "keyname" has another meaning, so we will also use "key value" or "value"

markus2330:
datatype, data-type -> "data type"

sanssecours:
[...]

Issue #531, opened 2016-02-18
Add Homebrew Formula for Elektra


sanssecours:
[Homebrew](http://brew.sh) is one of the most popular package managers for OS X. It would be nice if we  provide an official Homebrew formula – a.k.a. package – for Elektra. Someone already made a basic formula [here](https://github.com/fish2000/homebrew-praxa/blob/master/elektra.rb). Maybe we can base our work on that.

sirblackheart:
@tryge if you are ok with it I would take this over

tryge:
Please, go ahead!

markus2330:
As discussed in #1184, it would be also nice to build mac os x binaries using travis.

But adding Homebrew Formula has the highest priority, I hope someone with a Mac can pick this up!

sanssecours:
I created a Homebrew formula and [tap](http://github.com/sanssecours/homebrew-elektra) for Elektra.

@markus2330 I transferred the repo to you Markus, since I would have needed admin status to move the repo to the ElektraInitiative organization. Can you please move the repository from your personal account to the ElektraInitiative organization?

markus2330:
> I created a Homebrew formula and a tap for Elektra.

Thank you, that is really great!

Does it also includes binary packages for the latest release?

> I transferred the repo to you Markus, since I would have needed admin status to move the repo to the ElektraInitiative organization. Can you please move the repository from your personal account to the ElektraInitiative organization?

Sorry, where can I find it? In https://github.com/sanssecours/homebrew-elektra I do not have the button "settings" (which would be required for transfering).

Why not simply cloning it? (Or recreating one by pushing the same commits into it)

Btw. Is it possible to bring this Formula also upstream to brew?

markus2330:
@omnidan Can you test if the tap/bottle works for you, too?

Should we create issues about the tap/bottle here or at
https://github.com/ElektraInitiative/homebrew-elektra?

sanssecours:
> Does it also includes binary packages for the latest release?

Not yet, creating a binary package does [not look to be that hard](https://github.com/Homebrew/brew/blob/master/docs/Bottles.md) though. I will take a look.

> Sorry, where can I find it? In https://github.com/sanssecours/homebrew-elektra I do not have the button "settings" (which would be required for transfering).

Thanks you for the admin access. I just transferred the [repo](https://github.com/ElektraInitiative/homebrew-elektra).

> Btw. Is it possible to bring this Formula also upstream to brew?

[...]

PR #1240, opened 2016-12-28
Documentation improvements

e1528532:
# Purpose

Started improving some general documentation things i encountered when first trying elektra. Still WIP.

# Checklist

@markus2330 since you encouraged me also to make pull requests for WIP things, here it is ;). I yet have to go over the usage of key-path / key-name etc. once again, with my better understanding of the difference now. Since i struggled a lot more with getting the java parts to run, improvements on the java documentation will follow in a separate branch/pull requests.


README.md
```diff
@@ -212,7 +212,7 @@ For [OpenSUSE, CentOS, Fedora, RHEL and SLE](https://build.opensuse.org/package/
 Kai-Uwe Behrmann kindly provides packages [for download](http://software.opensuse.org/download.html?project=home%3Abekun%3Adevel&package=libelektra4).
 For Debian wheezy and jessie amd64 we provide latest builds. See build server below.
 
-If there are no packages available for your distribution, see the
+If there are no packages available for your distribution, or if you are using a different operating system like macOS, see the
```
markus2330: macOS now has homebrew packages, I think we can recommend that now.

Please also try it and give feedback.

doc/help/kdb-editor.md
```diff
@@ -89,3 +89,4 @@ Or set a new editor as default using:
 ## SEE ALSO
 
 - [elektra-merge-strategy(7)](elektra-merge-strategy.md)
+- [elektra-key-names(7)](elektra-key-names.md) for an explanation of key names.
```
markus2330: Please use key-?name consistently, discussion about it in #920

src/tools/kdb/cmdline.cpp
```diff
@@ -517,7 +517,11 @@ kdb::Key Cmdline::createKey (int pos) const
 
 	if (!root.isValid ())
 	{
-		throw invalid_argument (name + " is not a valid keyname");
+		throw invalid_argument (name + " is not a valid keyname" +
+								"\n\n"
+								"Please make sure that the key's namespace is valid and not missing (see 'man elektra-namespaces').\n"
```
markus2330: "not missing" is misleading, in cascading keys it is kind of "missing"?

README.md
e1528532: ah yeah, i added it to the package list. 
Encountered a small issue with the brew package https://github.com/ElektraInitiative/libelektra/issues/1243 which shouldnt be a problem though for normal users.

I was not sure how to control which optional components of elektra shall be built using the homebrew package but after a small study on the homebrew issue https://github.com/ElektraInitiative/libelektra/issues/531 i found out that i can see the parameters via brew info elektra. Is it worth mentioning this in the readme of our tap?

README.md
```diff
@@ -212,7 +212,10 @@ For [OpenSUSE, CentOS, Fedora, RHEL and SLE](https://build.opensuse.org/package/
 Kai-Uwe Behrmann kindly provides packages [for download](http://software.opensuse.org/download.html?project=home%3Abekun%3Adevel&package=libelektra4).
 For Debian wheezy and jessie amd64 we provide latest builds. See build server below.
 
-If there are no packages available for your distribution, or if you are using a different operating system like macOS, see the
+For macOS there is a homebrew tap available:
+  - [macOS](https://github.com/ElektraInitiative/homebrew-elektra)
+
+If there are no packages available for your distribution see the
```
markus2330: Shouldn't we link to doc/COMPILE.md?

src/tools/kdb/cmdline.cpp
e1528532: true. i adjusted the wording a bit, the problem is basically when you try to use a non-cascaded key, as i initally kept wondering why creating a key test/key did not work (> as there is most likely no test namespace globally). user/test would have worked instead, as the user namespaces does exist. 

is it better to understand now? any special situations i'm not yet aware about?

src/tools/kdb/cmdline.cpp
```diff
@@ -518,8 +518,8 @@ kdb::Key Cmdline::createKey (int pos) const
 	if (!root.isValid ())
 	{
 		throw invalid_argument (name + " is not a valid keyname" +
-								"\n\n"
-								"Please make sure that the key's namespace is valid and not missing (see 'man elektra-namespaces').\n"
+								"\n\n" +
+								"For absolute keys (starting without /), please note that only one of the predefined namespaces can be used (see 'man elektra-namespaces').\n" +
```
markus2330: What about giving an example of an valid keyname?

e1528532: i added both examples for a cascading and a non-cascading key

e1528532:
So I've revised the naming again. key name should be used consistently now, and keeping the explanation of key name vs path again and hopefully used it properly now. As a key name sort of represents a path its not always the easiest thing to decide what to use, but i try to use key name only when it actually points to a specific key and not to paths/hierarchies (so multiple keys might be affected by the operation or no actual key gets created like for mountpoints).

So if its ok now, i think this is pretty much ready for merge.

markus2330:
Thank you! Please avoid so many merges in future and rebase instead.

PR #1261, opened 2017-01-08
Documentation: Fix PDF Manual

sanssecours:
# Purpose

Before this commit building the PDF manual would fail, since some of the included Markdown documents contain the character `⏎`. This character was not setup for use with LaTeX. We now replace the symbol `⏎` with the similar looking leftwards arrow with hook symbol (`↩︎`).

# Checklist

### DIFF
--- a/doc/markdownlinkconverter/elektraSpecialCharacters.sty
+++ b/doc/markdownlinkconverter/elektraSpecialCharacters.sty
@@ -1,1 +1,4 @@
+\usepackage{newunicodechar}
+
 \DeclareUnicodeCharacter{0101}{\=a}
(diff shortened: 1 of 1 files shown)


markus2330:
Thanks, if you know some other letters that are available with every latex installation, please go ahead and add them too. It is quite a trap to add some text somewhere in the docu and suddenly building the pdf fails.

sanssecours:
> It is quite a trap to add some text somewhere in the docu and suddenly building the pdf fails.

Yeah, it would be great if Doxygen supported [XeLaTeX](https://en.wikipedia.org/wiki/XeTeX) or [LuaLaTeX](http://www.luatex.org) 🙁.

Issue #1277, opened 2017-01-13
jni plugin fails on macOS


e1528532:
# Describe what you wanted to do

I wanted to install the jni plugin. I have opened a pull request regarding updated documentation about installing jni on macOS. Then i wanted to execute kdb-full as suggested in the jni README to see if it actually works.

# Describe what you expected

I expected that everything will work.

# Describe what actually happened

After having solved some dylib related issues, which you see in my pull request, when opening kdb-full as suggested in the jni readme to test it, it fails as it tells me to install the legacy java 6 from apple. Which is weird as i had quite some effort to get jni to use the java 8 headers.

I did a quick research and found this https://bugs.openjdk.java.net/browse/JDK-7131356 which indicates this might be related to a jdk bug? The fix seems to be targeted for the jdk9, which will be released in a few months. However i'm not sure if its really related to that bug, but i wonder if its worth spending more time trying to make it run or simply wait for jdk9 and hope it will fix the issue?

# System Information

- Elektra Version: master
- jdk 1.8.112

markus2330:
Yes, sounds sensible to not use even more time for your java setup. Such an effort is unacceptable for potential users anyway.

PR #1278, opened 2017-01-13
Jni documentation

e1528532:
# Purpose

I documented the steps i took when i initially tried to make the jna bindings run (where i first assumed that jni is required too for them to work). However, there is still an issue described in #1277 which i haven't encountered back then as i found out jni is not required for jna, so after i made elektra compile with the jni plugin included, i never tried running kdb-full then.

# Checklist

- [ ] I added unit tests

@markus2330 So i think the updated readme is good now with information how to compile elektra on macOS, but how should we deal with #1277 ?


src/plugins/jni/README.md
```diff
@@ -47,6 +48,25 @@ and run:
 
     sudo ldconfig
 
+### Java prerequisites on macOS ###
+
+macOS includes an old apple specific version of java, based on 1.6. However, for the jni plugin version 1.8 of Java is required, so either the openjdk or the oracle jdk has to be installed.
```
markus2330: Please use linebreaks.

Did you try if java 1.6 works? As far as I can remember the constant `JNI_VERSION_1_8` is the only bit actually requiring 1.8.

src/plugins/jni/README.md
```diff
@@ -47,6 +48,25 @@ and run:
...
+
+Please install oracle's jdk8 via their provided installer. After that, you have to set the JAVA_HOME environment variable to the folder where the jdk is installed, usually like
+
+	export JAVA_HOME="/Library/Java/JavaVirtualMachines/jdk1.8.0_112.jdk/Contents/Home/"
+
+If it should still not find the correct jni version, or says the jni version is not 1.8, then it most likely still searches in the wrong directory for the jni header file. You can try clearing your build folder and configuring it again from scratch.
```
markus2330: Is it enough to remove CMakeCache.txt?

Please first write how it should be done (set `JAVA_HOME` and then call cmake) and later go to trouble shooting (ideally with an error message you get in such a situation).

src/plugins/jni/README.md
```diff
@@ -47,6 +48,25 @@ and run:
...
+It has been experienced that if the project has been built already without this environment variable set, the java location is cached and will be resolved wrong in future builds, even though the environment variable is set.
+
+As macOS handles linked libraries differently, there is no ldconfig command. Instead you can export an environment variable to tell elektra the location of the java dynamic libraries.
```
markus2330: Does every Java Application needs to set that? What about using `/etc/ld.so.conf.d` as done for Linux?

e1528532: It is only required for compiling JNI plugins, not for normal java applications (usually packaged as a jar).
There is not /etc/ld.so.conf.d in macOS, it seems to handle dynamic libraries a bit differently, i haven't looked at it in details yet as the environment variable was the fastest "fix".

src/plugins/jni/README.md
e1528532: i haven't tried as java 1.6 is not bundled with macOS anymore since 10.11, and as seen on https://www.java.com/en/download/faq/java_mac.xml#yosemitejava they don't support it anymore and recommend to use newer java versions by oracle instead:

> Apple has posted notice that Mac OS X 10.11 (El Capitan) will be the last OS X release that supports Java 6, and as such, recommends developers whose applications depend on Java 6 to migrate their apps to a newer Java version provided by Oracle.

should i try it anyway if it would work?

src/plugins/jni/README.md
e1528532: yes, this is enough, updated documentation according to that. Note that this will also reset any changes you have make with e.g. ccmake instead. 
also restructured this and moved to a troubleshooting section

markus2330:
Thanks for writing!

markus2330:
Ooops, this was not rebased, the history is not nice. Please always mark PRs with "ready to merge" if ready to merge.

markus2330:
> should i try it anyway if it would work?

No, let us continue with more important issues.
Please do not forget to wrap up the other PRs, in particular #1238.
It is not good if PRs are lying around for too long.

PR #1279, opened 2017-01-13
Virtualization with Vagrant or Docker

krit0n:
# Purpose
I think it is quite useful to be able to build Elektra with Vagrant or Docker.

When I for instance review a pull request I just start a dockercontainer and build Elektra from the hash of the pullrequests last commit and can try everything out in the sandboxenvironment of the dockercontainer.

Maybe some of you will find these files useful.


doc/docker/share/packelektra.sh
```diff
@@ -0,0 +1,20 @@
...
+cd /share
+if [ ! -d "${ELEKTRADIR}" ]; then
+    curl -sS ${SOURCE} | tar xz --transform "s/libelektra-${IDENTIFIER}/${ELEKTRADIR}/"
+fi
+if [ ! -d "{ELEKTRADIR}/build" ]; then
+  mkdir "${ELEKTRADIR}/build"
+fi
+cd "${ELEKTRADIR}/build"
+cmake -DCMAKE_INSTALL_PREFIX=/usr .. && make
+checkinstall -D -y --pkgname "${PKGNAME}" --pkgversion "${PKGVERSION}" --pkgrelease "${RELEASE}" --pkgsource "${SOURCE}" --pakdir /share
```
krit0n: Takes more time than just `make install` but I wanted to demonstrate that you can create customized packages.

doc/vagrant/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+When the machine is ready and running, access it with
+```bash
+$ vagrant ssh
+```
+In this SSH session you can interact with the machine.
+
+You could for instance change to `/vagrant`, which contains the contents of `share` from the host machine.
+
+Therefore it should contain an exemplary shell script that pulls a specified snapshot of Elektra from Github, installs it in the VM and builds a .deb package from it.
```
markus2330: Doesn't it first build the .deb and then installs it?

doc/vagrant/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Go ahead and give it a try:
```
markus2330: :smile:

doc/docker/share/packelektra.sh
```diff
@@ -0,0 +1,20 @@
...
+SOURCE="https://codeload.github.com/ElektraInitiative/libelektra/tar.gz/${IDENTIFIER}"
+ELEKTRADIR="libelektra-${IDENTIFIER}"
+
+# install elektra and create deb package
+cd /share
+if [ ! -d "${ELEKTRADIR}" ]; then
+    curl -sS ${SOURCE} | tar xz --transform "s/libelektra-${IDENTIFIER}/${ELEKTRADIR}/"
+fi
+if [ ! -d "${ELEKTRADIR}/build" ]; then
+  mkdir "${ELEKTRADIR}/build"
```
markus2330: please use tabs (currently number of spaces different)

doc/docker/README.md
```diff
@@ -0,0 +1,25 @@
+# Trying out Elektra with Docker
+
+This Dockerfile builds a dockerimage based on ubuntu:xenial including all requirements to build Elektra.
+
+If [Docker](https://www.docker.com/) is available on your machine change to the directory containing the file you are currently reading and build the image with
+```bash
```
markus2330: Does this really use bash?

doc/docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+This runs bash in a container based on the buildelektra image and mounts the `./share` folder into the container.
+```bash
+$ docker run --rm -v ${PWD}/share:/share -it buildelektra bash
+```
+
+In share there is an exemplary shell script that pulls a specified snapshot of Elektra from Github, installs it in the container and builds a .deb package from it.
+
+Try it out from within the container with
+```bash
+$ cd /share
```
markus2330: Why using it here? The scripts does so, too? (see also below)

doc/docker/share/packelektra.sh
```diff
@@ -0,0 +1,20 @@
...
+fi
+cd "${ELEKTRADIR}/build"
+cmake -DCMAKE_INSTALL_PREFIX=/usr .. && make
```
markus2330: Some `-j` would be useful in most cases?

doc/docker/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    libaugeas-dev \
+    libyajl-dev \
+    libgit2-dev \
+    libboost-all-dev \
+    libssl-dev \
+    libdbus-1-dev \
+    libpcre3-dev \
+    libpcre++-dev \
+    checkinstall
+    && gem install ronn
```
krit0n: There was a missing backslash. Should work now.

doc/docker/share/packelektra.sh
krit0n: Everything is 'tabified' now.

doc/docker/share/packelektra.sh
krit0n: Is added in the vagrant and docker versions of `packelektra.sh`

doc/docker/README.md
krit0n: I've replaced it with sh. I always use _bash_ for highlighting out of a habit, but it should be [synoymous](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3950).

doc/vagrant/share/packelektra.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
krit0n: I think I will remove this file anyway. This file was meant as a showcase for how one can use containers/VMs. I think its better if I incorporate and describe the functionality of this file into the README.

markus2330: Its better to execute the script than doing copy and paste.

Or do you have an idea how we can have docker/vagrant with Elektra already installed? (Both is sometimes needed)

krit0n: It would be easy to expand the Vagrant-/Dockerfile to build environments that have Elektra installed.
If I add these two lines to the Dockerfile it generates an image with elektra 0.8.19:
```
ADD https://codeload.github.com/ElektraInitiative/libelektra/tar.gz/0.8.19 /tmp/elektra.tar.gz
RUN cd /tmp && tar xzf elektra.tar.gz && cd libelektra-0.8.19 \
  && mkdir build && cd build && cmake <your desired flags> .. && make -j && make install
```
I could add similar lines to the Vagrantfile.
(This approach has the drawback, that the resulting VM/image has unnecessary packages installed, that you only need to build Elektron but not to run it. As far as I know the best approach to tackle this is to build two containers: One that builds the package and another that just installs it).
I think if you provide such containers its best to distribute them via [dockerhub](https://hub.docker.com/) or [atlas](https://atlas.hashicorp.com/boxes/search) (but I think you have to pay for that).

I didn't do this because I had these two usecases in mind:

- Build your own customized package

- Try out a certain Elektra commit (useful if you want to review a pullrequest)

doc/vagrant/README.md
krit0n: I'm not sure, but my guess from the wording in the manpage

> DESCRIPTION
> checkinstall is a program that monitors an installation procedure (such as make install,
> install.sh ),  and creates  a  standard package for your distribution (currently deb, rpm
> and tgz packages are supported) that you can install through your distribution's
> package management system (dpkg, rpm or installpkg).

is that checkinstall is that it first installs (otherwise it couldn't monitor it) and than builds the deb based on that.

doc/vagrant/share/packelektra.sh
markus2330: Yes, but for these two use cases it will be good to have a script that already does everything? (Maybe installing is not always necessary, we could add command-line arguments for the script)

doc/vagrant/README.md
markus2330: Yes, it does some temporary install to track everything, but why mention this in the README.md? Important for the reader of the readme is that after running the shell script, Elektra will be installed (maybe unless some command-line option is provided?)

Thus I would not mention the temporary install (implementation detail of checkinstall), but I would mention that the deb gets installed.

krit0n:
@markus2330 I think I implemented all that you wanted me to change.

`buildelektra.sh` is now a script with commandline options to install elektra in the VM/container or build a .deb package.

For building a .deb package I use `checkinstall`. I'm not sure if that is the best way to do that, because I'm rather inexperienced in building debian packages.

.gitignore
```diff
@@ -11,7 +11,7 @@
 /debian/patches/
 
 #ignore all build directories of cmake
-build*
+build*/
```
krit0n: I oversaw commiting buildelektra.sh because of this.

markus2330:
@petermax2 Thanks for the timely review!

@krit0n Looks great! Please add "ready to merge" if you do not see a way to merge the scripts.

petermax2:
EDIT: Now I managed to use your Vagrantfile for a fully automated provisioning. It works really well! Thx again.

markus2330:
Thank you!

Issue #1167, opened 2016-12-05
zsh completion requirements: better kdb ls


reox:
for a better zsh completion a tool would be nice, that gives more information than `kdb ls`.
It would be nice to have information if a key is leaf or a node and if it is a node, if subkeys (nodes or leafes) exists.
each item can be null seperated or seperated by newlines.

maybe such a command can be useful for someone else as well, so maybe null seperated values can be triggered using a cmdline option?

the output could look like this:
```
kdb super-ls /
system/foo node 1
system/foo/bla node 1
system/foo/bla/something node 4
system/foo/bla/something/a leaf
system/foo/bla/something/b leaf
system/foo/bla/something/c leaf
system/foo/bla/something/d leaf
system/bla node 1
system/bla/xx node 0
```

markus2330:
@e1528532 maybe this is an easy first step? Please use null-separation, otherwise keys with news-lines will make trouble.

And we need a better name than super-ls ;)

markus2330:
One more not-so-obvious note: the list should also contain mountpoints which are leaf. This is because after mounting you often do not have keys (non-existing file) in the mountpoint but want nevertheless get completion to the mountpoint.

Issue #1248, opened 2017-01-02
complexity analysis with pmccabe


markus2330:
Some functions are too complex. Please add tests and refactor them.

```
pmccabe `find . -name '*.c'` | sort -nr | head -20
"./src/tools/gen/util/util.c", line 74: expected ';' got EOF
121     121     243     101     367     ./src/plugins/ini/inih-r29/inih.c(101): ini_parse_file
76      82      46      930     101     ./src/plugins/ni/nickel-1.1.0/src/io.c(930): PutUtf8Char
52      56      78      17      212     ./src/tools/gen/template/genopt.c(17): header
48      48      174     592     291     ./src/plugins/curlget/curlget.c(592): elektraCurlgetSet
48      48      128     1174    224     ./src/plugins/ini/ini.c(1174): iniWriteKeySet
45      45      129     32      251     ./src/plugins/xmltool/kdbtools.c(32): consumeKeyNode
43      51      106     188     154     ./src/plugins/conditionals/conditionals.c(188): evalCondition
42      51      130     92      314     ./src/plugins/ni/nickel-1.1.0/src/io.c(92): GetNextIdentifier
35      46      68      287     71      ./src/libs/proposal/proposal.c(287): keyRel2
35      35      90      231     154     ./src/plugins/curlget/curlget.c(231): elektraCurlgetOpen
32      32      89      217     149     ./doc/markdownlinkconverter/markdownlinkconverter.c(217): convertLinks
32      32      31      842     84      ./src/plugins/ni/nickel-1.1.0/src/io.c(842): PutString
31      32      123     1192    200     ./src/libs/meta/meta.c(1192): elektraSortTopology
31      31      70      150     160     ./src/plugins/xmltool/stream.c(150): keyToStreamBasename
30      30      100     36      137     ./src/plugins/validation/validation.c(36): elektraValidationSet
26      26      51      29      105     ./src/plugins/csvstorage/csvstorage.c(29): parseLine
25      33      111     196     181     ./src/plugins/mathcheck/mathcheck.c(196): parsePrefixString
24      32      98      983     150     ./src/libs/elektra/keyset.c(983): ksCut
24      27      51      102     94      ./src/plugins/mathcheck/mathcheck.c(102): doPrefixCalculation
24      24      60      216     89      ./src/libs/elektra/keytest.c(216): keyIsBelow
```

e1528532:
whats the target complexity level here? I seems like 10 is the original recommendation by mc cabe

[...]

Issue #788, opened 2016-06-13
Fix sanitizer issues


markus2330:
Hello newcomer

Since 702cd0cfd025c313e5e165af5394a83fcf01959a one can easily enable [undefinied behavior sanitizer](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html).

Simply use clang as toolchain and use debug/asan (e.g. by adding `-DENABLE_DEBUG=ON -DENABLE_ASAN=ON -DCMAKE_C_COMPILER=clang-3.8 -DCMAKE_CXX_COMPILER=clang++-3.8` to your cmake invocation). It might work for gcc, too.

Then try to run all tests (see tree.libelektra.org/doc/TESTING.md) and fix some issues that are reported. Do not try to fix all at once, but create a PR as early as possible.

See also:
https://build.libelektra.org/jenkins/job/elektra-clang-asan/lastFailedBuild/console
https://build.libelektra.org/jenkins/job/elektra-gcc-asan/lastFailedBuild/console

sirblackheart:
Should we open separate bugs per found undefined behavior or just add them here?

sirblackheart:
This one in `keyset.c`
`/src/libs/elektra/keyset.c:922:11: runtime error: unsigned integer overflow: 42 + 18446744073709551592 cannot be represented in type 'unsigned long'`

the used nickel library seems to have quite some in `hash.c` that should probably just suppressed as it is not really part of elektra
`/elektra/src/plugins/ni/nickel-1.1.0/src/hash.c:`

and one in the `resolver.c`
`/src/elektra/src/plugins/resolver/resolver.c:95:16: runtime error: member access within null pointer of type 'resolverHandles' (aka 'struct _resolverHandles')`

markus2330:
I updated the issue to use `-DENABLE_ASAN=ON`. `ENABLE_DEBUG` does not activate asan by default anymore.

sanssecours:
Since the ASAN build jobs do not build all plugins, here are some other tests that currently also fail on my machine after I enable the address sanitizer:

- [ ] `testjna_maven`
- [ ] `test_kdb.lua`
- [ ] `test_key.lua`
- [ ] `test_keyset.lua`

.

markus2330:
The error messages would be interesting. If lua/java cause the problems we might have to exclude the tests (at least for the version/OS where it does not work).

I get now a segmentation fault I had not earlier (only with ASAN):
```
[...]

Issue #265, opened 2015-08-27
Improve Naming Schemes for KDB Commands


iandonnelly:
You are completely right!
This was outlined in [issue #260](https://github.com/ElektraInitiative/libelektra/issues/260) by Markus but I wanted to separate this into it's own issue for clarity.

From Markus:

> I would suggest to use a hierarchical naming schema:  
> kdb-plugin (no real command, just general info about plugins)
> kdb-plugin-check
> kdb-plugin-info
> kdb-plugin-list (or kdb-plugin-ls?)
> 
> To be discussed if this also applies to meta and/or mounting:    
> 
> kdb-meta (info about meta data)
> kdb-meta-ls
> kdb-meta-set
> 
> and for future spec, global mounts:    
> 
> kdb-global-mount
> kdb-global-unmount
> 
> kdb-spec-mount
> kdb-spec-unmount
> 
> I agree we should have aliases to not break scripts.

markus2330:
Thanks for opening a new issue! I removed the text from #260. One line was missing:

We should also hide some less important commands from --help (and give them to list-tools), e.g. fstab, test, vset is hardly used.

markus2330:
Any progress here?

markus2330:
As mentioned in #1188 we should also get rid of `kdb fstab` but use a generic way to set multiple keys at once.

markus2330:
As mentioned in #876, `mountpoints-info` might be renamed to `info`, after `info` is renamed to `plugin-info`.

markus2330:
As discussed in #2150 `kdb check` should be a check for the KDB, and for plugins there should be `kdb plugin-check`.

markus2330:
See #2388, @kodebach suggested:

> Regarding improving kdb for array manipulation and in general: Once kdb has better support for sub-commands (using elektraGetOpts?) I think we should move to a hierarchical system would be a good idea. So kdb would for example have the subcommands array and meta, while each of those have subcommands called get, set and rm (and more), meaning kdb array set, kdb meta get and kdb array rm would all be commands.

Yes, that would be very nice.

We already have something like this with the external executor, for example `kdb elektrify-getenv getenv`.

Ideally, we would also unify how the external commands work. So `kdb umount all` instead of `kdb umount-all`. 

Let us see if @ulschaef wants to work on this.

PhilippGackstatter:
@markus2330 
This seems like a very broad issue. Is there something concrete I can work on?

[...]

PR #1281, opened 2017-01-15
super-ls

e1528532:
# Purpose

This branch is used to develop an advantaged version of the ls tool to include further information e.g. wheter if a key is a node or a leaf, and the number of nodes/leaves inside a given node as request in #1167 .

# Checklist

needs to be checked.

- [ ] I added unit tests

@markus2330 Please note this is heavy WIP currently. You don't seem to like the name super-ls, but i think it suits fine as a working title for now until we find a proper name. Another idea i had is that we could just add an option to the original ls tool to display node/leaf/count information instead of handling this as a new tool. In that case we could also add an option to switch between newline or NULL separators. Or lets keep it a new tool?

And consider the following case of keys:
user/test/key = "test1"
user/test/key/2 = "test2"

should the output be like
user/test node 1
user/test/key node 1
user/test/key/2 leaf

or simply (so not listing the user/test node which does not actually exist in elektra in the example)
user/test/key node 1
user/test/key/2 leaf

the latter would be easier i think as it allows me to simply iterate over the keyset i get via kdbGet like i'm doing it now.
Looking forward to some feedback as its the first actual c++ code i'm doing for elektra

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,10 @@ Go to the [website](https://www.libelektra.org), see [News](doc/news/), and its 
 The preferred way to install Elektra is by using packages provided for
 your distribution, see [INSTALL](/doc/INSTALL.md) for available packages.
 
-If there are no packages available for your distribution, see the
+For macOS there is a homebrew tap available:
+  - [macOS](https://github.com/ElektraInitiative/homebrew-elektra)
+
+If there are no packages available for your distribution see the
 [installation document](doc/INSTALL.md).
 
 ### Download ###
@@ -203,7 +206,8 @@ following commands to compile it:
 
  * `mkdir build`
  * `cd build`
- * `cmake ..` or `ccmake ..`
+ * `cmake ..`
+ * `ccmake ..` (optional, provides a console based GUI to give an overview of the available compilation options and settings)
  * `make`
 
 Then you can use `sudo make install` to install it.
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ applications' configurations, leveraging easy application integration.
 
 ## Often used links ##
 
-- [build server](http://build.libelektra.org:8080/)
-- [tutorials](http://git.libelektra.org/blob/master/doc/tutorials/)
-- [API docu](http://doc.libelektra.org/api/latest/html/)
+- [build server](https://build.libelektra.org/)
+- [tutorials](/doc/tutorials/)
+- [API docu](https://doc.libelektra.org/api/latest/html/)
 
 ## Overview ##
 
@@ -58,7 +58,7 @@ implemented in C, works cross-platform, and does not need any external dependenc
 ## Contact ##
 
 Do not hesitate to ask any question on
-[GitHub issue tracker](https://github.com/ElektraInitiative/libelektra/issues),
+[GitHub issue tracker](https://issues.libelektra.org/),
 [Mailing List](https://lists.sourceforge.net/lists/listinfo/registry-list)
 or directly to one of the [authors](doc/AUTHORS.md).
 
@@ -109,7 +109,7 @@ To get an idea of Elektra, you can take a look at the
 
 The full documentation, including:
 
-- [tutorials](http://libelektra.org/blob/master/doc/tutorials/),
+- [tutorials](/doc/tutorials/),
 - [faq](/doc/help/elektra-faq.md),
 - [glossary](/doc/help/elektra-glossary.md), and
 - [concepts and man pages](/doc/help/elektra-introduction.md)
@@ -119,10 +119,10 @@ is available in the GitHub repository.
 You can read the documentation for the kdb tool, either
 
 - [on the Website](https://www.libelektra.org)
-- [in the API docu](http://doc.libelektra.org/api/latest/html/md_doc_help_kdb.html)
+- [in the API docu](https://doc.libelektra.org/api/latest/html/md_doc_help_kdb.html)
 - by using `man kdb`
 - by using `kdb --help` or `kdb help <command>`
-- [on GitHub](http://libelektra.org/blob/master/doc/help/kdb.md)
+- [on GitHub](https://master.libelektra.org/doc/help/kdb.md)
 
 
 
@@ -172,20 +172,7 @@ And in terms of quality, we want:
 
 ## News ##
 
- - [22 Nov 2016 0.8.19](http://doc.libelektra.org/news/8e05231a-4f3d-488b-8dc2-5f0d5c474c39.html) tutorials, ruby, cleanup
- - [16 Sep 2016 0.8.18](http://doc.libelektra.org/news/190576e0-9fef-486e-b8da-c4e75be08329.html) adds intercept open
- - [14 Jun 2016 0.8.17](http://doc.libelektra.org/news/e6153a39-c4bd-41c3-bc86-785d451eb6c5.html) with survey, usability improvements
- - [29 Apr 2016 0.8.16](http://doc.libelektra.org/news/9c9247ee-ee9c-4f4a-a68e-76959def9b82.html) with stability improvements
- - [16 Feb 2016 0.8.15](http://doc.libelektra.org/news/1ab4a560-c286-46d2-a058-1a8e7e208fe8.html) with lib split, improved mount
- - [19 Nov 2015 0.8.14](http://doc.libelektra.org/news/519cbfac-6db5-4594-8a38-dec4c84b134f.html) adds docu and plugins
- - [17 Sep 2015 0.8.13](http://doc.libelektra.org/news/3c00a5f1-c017-4555-92b5-a2cf6e0803e3.html) adds elektrify-getenv
- - [12 Jul 2015 0.8.12](http://doc.libelektra.org/news/98770541-32a1-486a-98a1-d02f26afc81a.html) adds dir namespace
- - [03 Apr 2015 0.8.11](http://doc.libelektra.org/news/7d4647d4-4131-411e-9c2a-2aca39446e18.html) adds spec namespace
- - [02 Dec 2014 0.8.10](http://doc.libelektra.org/news/6ce57ecf-420a-4a31-821e-1c5fe5532eb4.html) adds XDG/OpenICC compatibility
- - [04 Nov 2014 0.8.9](http://doc.libelektra.org/news/38640673-3e07-4cff-9647-f6bdd89b1b08.html) adds qt-gui
- - [02 Sep 2014 0.8.8](http://doc.libelektra.org/news/eca69e19-5ddb-438c-ac06-57c20b1a9160.html) adds 3-way merging
-
-Also see [News](doc/news/) and its [RSS feed](http://www.libelektra.org/news/feed.rss).
+Go to the [website](https://www.libelektra.org), see [News](doc/news/), and its [RSS feed](https://www.libelektra.org/news/feed.rss).
 
 
 ## Sources ##
@@ -206,7 +193,7 @@ You can clone the latest version of Elektra by running:
 
          git clone https://github.com/ElektraInitiative/libelektra.git
 
-Releases can be downloaded from [http](http://www.libelektra.org/ftp/elektra/releases/) and
+Releases can be downloaded from [http](https://www.libelektra.org/ftp/elektra/releases/) and
 `ftp://ftp.libelektra.org/elektra/releases/`
 
 ### Compiling ###
--- a/src/tools/kdb/superls.cpp
+++ b/src/tools/kdb/superls.cpp
@@ -31,7 +31,8 @@ int SuperLsCommand::execute (Cmdline const & cl)
 	}
 
 	printWarnings (cerr, root);
-	root = cl.createKey (0);
+	Key originalRoot = cl.createKey (0);
+	root = originalRoot;
 	kdb.get (ks, root);
 	
 	cout.setf (std::ios_base::unitbuf);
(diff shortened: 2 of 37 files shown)


markus2330:
> You don't seem to like the name super-ls

Yes, you are right ;) Maybe `completion-helper`? We should take it out from the list of commands, maybe even develop as external command?

>  Another idea i had is that we could just add an option to the original ls tool to display node/leaf/count information instead of handling this as a new tool.

Only if the feature is useful for humans, too. And this might be not the case, especially if we add the feature to complete to mountpoints.

We should dedicate the tool to the goal: "given a key name, give me possibilities how to complete the key name". This is something quite different to what `kdb ls` does.

> In that case we could also add an option to switch between newline or NULL separators.

`-0` already does that.

@reox

reox:
> We should dedicate the tool to the goal: "given a key name, give me possibilities how to complete the key name". This is something quite different to what kdb ls does.
Yes, that would be exactly what the completion file does. I get the current cmdline like `/sw/pk/`, then i query for all keys that are there and get for example `/sw/pk/foo/sdf/`, `/sw/pk/foo/sdfsdg`, `/sw/pk/bla/sdf` and  then i strip all the compontents from the front and back to have only the next level.

Maybe have it like `find` with --maxdepth and --mindepth to look up keys:
`super-ls --maxdepth=1 /sw/pk` would return `/sw/pk/foo` and `/sw/pk/bla/` (+all the other information like leaf and node)

e1528532:
ok, i have created a prototype (yeah still called super-ls for now, will migrate it to an own tool or rename it to completion-helper once the functionality is more or less done). There is plenty of space for refactoring and optimizations but for now we should get the functionality straight before spending time on that.

As its still based on the kdb tool right now, i just reused the -r parameter (recursive) to display all possible completions, and otherwise its just the next level, so like --maxdepth=1 would be. Internally there is already a support for having a max and mindepth.

I think the easiest way to clarify the requirements is to show some examples how it currently works, based on this test dataset:
```
mbp-armin:build admin$ kdb ls user/
user/kdb_test/1/key_name
user/kdb_test/1/key_name/2
user/kdb_test/1/other
user/test
user/test/bla
user/test/blu
```

```
mbp-armin:build admin$ kdb super-ls user/
user/kdb_test node 1
user/test node 2
```

```
mbp-armin:build admin$ kdb super-ls user/kd
user/kdb_test node 1
```

```
mbp-armin:build admin$ kdb super-ls user/kdb_test
user/kdb_test/1 node 2
```

```
mbp-armin:build admin$ kdb super-ls -r user/kd
user/kdb_test node 1
user/kdb_test/1 node 2
user/kdb_test/1/key_name node 1
user/kdb_test/1/key_name/2 leaf
user/kdb_test/1/other leaf
```

For cascading keys, i return the absolute keynames so you clearly see to which key it would resolve which I think is the more helpful approach:
```
mbp-armin:build admin$ kdb super-ls -r /kd
user/kdb_test node 1
user/kdb_test/1 node 2
user/kdb_test/1/key_name node 1
user/kdb_test/1/key_name/2 leaf
user/kdb_test/1/other leaf
```

```
mbp-armin:build admin$ kdb super-ls /
system/elektra node 2
user/kdb_test node 1
user/test node 2
```

@reox is this going in the right direction?

reox:
Yes, i think this looks good!

e1528532:
@markus2330 ok since it looks like i've hit the main requirements successfully, now its time to decide wheter it should be an external command or not.

I hope its ok that its done in c++ like the other kdb tools are, as i think i would have been much slower coding this in plain c. It still took me quite some time as i had to get familiar with c++ again and exploring how the elektra api can be used for a real world problem, handling cascading keys and "non existent keys" was also not too easy (with non existant keys i mean if there is a key "user/this/is/a/key", and i want to complete "user/this/is" which does not correspond to an actual existing key, and "user/this/is/a" being no existent key either, that it still suggests "user/this/is/a" which would be a node then with the child key).

Leaving it as a kdb tool would have the benefit that its not much more work, but if all the kdb tools are intended for humans, its probably not the best idea.

If we want to make an own tool out of it, how would it be called, kdb-completion-helper? kdb-complete? kdb-suggest? I guess i could reuse most of the command line parsing from kdb, and i think thats all i need apart from my super-ls files.

Ah and a last thing, I yet have to do the mountpoint completion. Do i get it right that i can fetch a current list of mountpoints from system/elektra/mountpoints which I'd parse and add to the suggestion processing then? If so, i think its not too difficult to implement it. Should it somehow be marked as a mountpoint in the output? Its very nice to see how elektra uses itself to store parts of its configuration, clearly demonstrates its capabilities!

markus2330:
> Leaving it as a kdb tool would have the benefit that its not much more work

 :+1:  Sounds like a good idea then!

> but if all the kdb tools are intended for humans, its probably not the best idea.

We already have this issue with `list-commands` which is not really useful for users, too. (Any mistyped command gives you the list.) So I would suggest we add a feature in the factory to hide internal/renamed commands or even aliases (gmount vs. global-mount) See #265 to have an idea about the long-term goal.

Then we would simply hide it from the list.

markus2330:
Implementation note: better to not mess around with the factory, but directly improve the `Command` or `Instancer` class. `getLongHelpText` actually is no longer needed. Instead we would need a `isHidden` and `isAlias` or similar. Would also be nice to make the instantiation cheaper if `execute` is not needed (maybe bound statically with `Cnstancer` instead of the instantiation). Can you make an suggestion which does not take too much effort?

But let us better do this in a different PR, otherwise this gets never merged. So here you can keep it as internal command, simply rename it to `completion-helper` or `list-completions`.

markus2330:
Btw. if all of you really want `super-ls` when can continue to use the name...

@e1528532 It is your decision

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+SuperLsCommand::SuperLsCommand () : kdb (root)
+{
+}
+
+int SuperLsCommand::execute (Cmdline const & cl)
+{
+	if (cl.arguments.size () != 1)
+	{
+		throw invalid_argument ("1 argument required");
```
markus2330: Keep it consistent with others: `wrong number of arguments, 1 needed`.

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+	
+	cout.setf (std::ios_base::unitbuf);
+	if (cl.null)
+	{
+		cout.unsetf (std::ios_base::skipws);
+	}
+	
+	function<bool(string)> pred = [] (string) { return true; };
+	bool isExistent = ks.lookup(root) || root.getBaseName().empty();
+	if (!isExistent) {
```
markus2330: please run code formatter.

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+		string fullName = cl.arguments[0];
+		string newKeyName = getParentKey(root).getFullName();
+		cout << "new name is " << newKeyName << endl;
+		root = Key (newKeyName, KEY_END);
+		cout << " building predicate " << root.isCascading() << " " << fullName << " " << fullName.find("/") << endl;
```
markus2330: is this a verbose output? Please only if verbosity is set.

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+		pred = [=] (string test) { 
+			return fullName.size() <= test.size() && std::equal(fullName.begin(), fullName.end(), test.begin() + (root.isCascading() ? test.find("/") : 0));
```
markus2330: There is no requirement that lambdas fit in a single line.

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+		};
+	}
+	
+	cout << "cutting at " << root << endl;
```
markus2330: should definitely be only in verbose

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+	ks.rewind();
+	
+	KeySet hierarchialKeys;
+	Key parent;
+	Key last;
+	Key tmp;
+	
+	stack<Key> keyStack;
+	if (!(tmp = ks.next())) {
+		// return?
```
markus2330: ?

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+		return 0;
+	}
+	
+	int curDepth = 0;
```
markus2330: function gets too long, please add more comments.

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+	}
+	
+	// Adjust the output offset, in case the given string exists in the hierarchy but not in the original ks
+	int offset = !isExistent && hierarchialKeys.lookup(originalRoot);
+	int minDepth = 0 + offset;
+	int maxDepth = cl.recursive ? std::numeric_limits<int>::max() : (1 + offset);
+	cout << "max is " << maxDepth << " and min is " << minDepth << endl;
+	
+	cout << "dumping all results" << endl;
+	for (auto it = hierarchy.begin(); it != hierarchy.end(); it++) {
```
markus2330: there are even shorter for loops.

src/tools/kdb/superls.cpp
markus2330: maybe you can even create some reusable functions here?

src/tools/kdb/superls.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+	return 0;
+}
+
+Key SuperLsCommand::getParentKey(Key key) {
+	return Key (key.getFullName().erase(key.getFullName().size() - key.getBaseName().size()), KEY_END);
+}
+
+void SuperLsCommand::increaseCount(map<Key, pair<int, int>> & hierarchy, Key key, function<int(int)> depthIncreaser) {
+	cout << "increasing for " << key << endl;
```
markus2330: again, should be only in verbose.

markus2330:
jenkins build fast please

markus2330:
jenkins build please

e1528532:
I think I failed to emphasize enough that its WIP and thus still contains some debugging output (i think most of the stuff i print right now wouldn't be wanted in a -v output and is really just for debugging). Also there is plenty of room for refactorings, and no standardized formatting yet (just using notepad++ right now), as i wanted to implement the functionality first and would have dealt with that afterwards.

However, it looks like overall you are more happy with my code than i currently am, and that using small functional abstractions is ok :) I had the feeling that larger adjustments are necessary.

markus2330:
> I think I failed to emphasize enough that its WIP

There is a "work in progress" label just for that. But its better to give more feedback early than to have too many review rounds.

> and no standardized formatting yet (just using notepad++ right now), as i wanted to implement the functionality first and would have dealt with that afterwards.

Do not worry about editor or formatting. Formatting is completely left to `scripts/reformat-source`. But for the next time please run the script before creating the PR, it makes it easier to read the code and the build jobs will not fail (they check if the formatter was used).

> However, it looks like overall you are more happy with my code than i currently am, and that using small functional abstractions is ok :) I had the feeling that larger adjustments are necessary.

Yes, thus early PRs are important! It is just a helper tool, some shortcuts are okay here.

sanssecours:
> @sanssecours any opinion?

If you want to merge this pull request immediately, then you do not need to wait for my review. I am currently sick 🤒. Therefore it might take me some time to come back to this.

As for the name: I do not like `super-ls` either. How about `complete`, which I think is both descriptive and short?

markus2330:
> If you want to merge this pull request immediately

In general it is better if PRs can be merged fast (less rebasing and forgetting about what the PR was about). In this case merging would be helpful so that @reox could continue his work on the auto completion.

> How about complete, which I think is both descriptive and short?

Sounds nice! In #265 we basically came to the conclusion that the names should contain what to do and where it should be done. E.g., not `list` but `list-plugins`. But in this case it directly works on keys, so `complete` actually would be fine. If we add completion for plugins, it would be `complete-plugins`.

Thanks for the input and get well soon!

e1528532:
Ok I think I'm getting pretty close to finishing this.

The most important change i've made is that i decided to call the tool simply "complete" as sansseccours suggested and markus approved, because its our general completion tool. I agree that this name is better, as if I'd look at the list of commands kdb offers, I would have no idea what super-ls does at a first sight compared to ls, but complete already gives a good indication about its purpose.

I have also changed the interface now to my proposal:
-v to display verbose output (main difference: shows you the min/max output depths it will actually use and shows you if the path you want to complete does not exist or what else it uses as the reference path used for determining the depth)
-d to display debugging output (shows more details how it processed/expands keys and will give you the whole analyzation result prior to filtering it by name/depth
--min-depth to specifiy the minimum output depth, exclusive. 0 by default.
--max-depth to specify the maximum output depth, inclusive, 1 by default as i think when using a complete command you'd usually just want to see suggestions for the current level. -1 to show all levels.
-0 to have nulls as whitespace (stolen from kdb ls)

Some feedback would be appreciated if the depth parameters and defaults make sense.

The last big addition is that the command will also complete to mountpoints now, even if there are no keys in the mountpoint yet as requested by markus.

Furthermore i have refactored the code to separate the logic in (hopefully) meaningful functions and reviewed the output, displaying some debugging and verbose things only when the corresponding parameters are set. I also used our code formatter and tried to keep all your suggestions in mind.

I have a question about the getopt stuff though, i provided short options for my minimum and maximum depth (m for minimum and M for maximum), but i failed to use them. It works fine when i specify the options in the long format, like kdb complete -vd --min-depth=2 --max-depth=5 user/ , but fails when i try it in the short format like kdb complete -vd -m 2 -M 5 user/, -m2 or -m=2 doesn't work either. Is it not possible to have arguments for short options, and if so, can i hide the short option somehow?

Also, do we have some code linter doing some code analysis and monitoring things like unfreed references or complexity violations or similar I could/should use?
And do we maintain unit tests for the kdb tools? There are some tests ins libelektra/tests but they seem to be related to the main elektra library and not to any tools.

reox:
For min/max I would use the defaults of `find`: min 0 and max inf. 
I think the default should be to show everything, where other tools that might only need some subpart, can explicitly ask for it.

Am 21. Jänner 2017 17:07:34 MEZ schrieb e1528532 <notifications@github.com>:

README.md
```diff
@@ -13,9 +13,9 @@ applications' configurations, leveraging easy application integration.
 
 ## Often used links ##
 
-- [build server](http://build.libelektra.org:8080/)
-- [tutorials](http://git.libelektra.org/blob/master/doc/tutorials/)
-- [API docu](http://doc.libelektra.org/api/latest/html/)
+- [build server](https://build.libelektra.org/)
```
markus2330: It seems like you have severe rebasing/merging issues here!

markus2330:
I would avoid having min/max for now. Is it really needed? It would be more useful for the actual `ls` (@tom-wa requested it for `ls` already).

> Also, do we have some code linter doing some code analysis and monitoring things like unfreed references or complexity violations or similar I could/should use?

See doc/TESTING.md. For memory issues we use valgrind and ASAN, complexity see #1248

> And do we maintain unit tests for the kdb tools?
> There are some tests ins libelektra/tests but they seem to be related to the main elektra library and not to any tools.

We have tests for libtools (the more complex stuff of what the tools do is implemented there, if the complete stuff gets too complicated we might even move it there) and we have the shellrecorder.

Btw. man page for the command is missing. (doc/help)

README.md
e1528532: Looks like it. I've never used merge for this branch, and tried rebasing by using "git rebase ElektraInitiative/master" while i'm in this branch, but that somehow lead to this merge commit b970a1f ... isn't this the way to rebase things? It hasn't told me about any conflicts and hasn't asked me to resolve anything either so i wonder where this commit comes from.

markus2330: `git pull --rebase` should do the right thing. Now its the easiest to do a `cherry-pick` of the correct commits in a new branch.

e1528532: i slowly suspect that the github desktop app did all kinds of messy things. I've switched to sourcetree now and it think its much better, gives you a much better indication of what you are about to do and its very easy to rebase or work with branches. I just created a new branch based on the last good commit of super-ls, so no cherry picking needed. Will open a new pull request after dealing with the last few things here.

sanssecours: > I just created a new branch based on the last good commit of super-ls, so no cherry picking needed.

Thank you for providing the new branch. Currently the code in `kdb-complete` does not compile on my machine. It works after I remove the files`superls.cpp` and `superls.hpp`, and replace 

```cpp
#include <superls.hpp>
```

in `complete.cpp` with

```cpp
#include <complete.hpp>
```

.

e1528532: ah yeah i've already fixed that but haven't comitted sorry, now it should be in the branch

e1528532:
@markus2330 IMHO having min and maxdepth is not a bad thing, as the completion tool is primarily intended for internal use / shell completion and things like that, so those two parameters give you some flexibility in how you can use it. 

manpage is added

valgrind seems unsupported on osx, running it with ASAN enabled gives me the following output:
```
/Users/admin/git/libelektra/src/libs/elektra/keyset.c:914:11: runtime error: unsigned integer overflow: 19 + 18446744073709551597 cannot be represented in type 'unsigned long'
SUMMARY: AddressSanitizer: undefined-behavior /Users/admin/git/libelektra/src/libs/elektra/keyset.c:914:11 in 
```

... That doesn't seem to be related to my code right?

And I've also checked the complexity, the analyze method has 13, but 3 are caused by output switches (if cl.verbose etc), so i'd say thats okay, the same for printResults (12, where 2 are caused by output switches)

so the last big thing to do is to check out how to create tests for it

markus2330:
> IMHO having min and maxdepth is not a bad thing, as the completion tool is primarily intended for internal use / shell completion and things like that, so those two parameters give you some flexibility in how you can use it.

Flexibility is not necessarily a good thing if nobody needs it. But you can leave it for now if we need the same arguments for `ls` anyway.

> manpage is added

How is it called? I cannot find it.

> valgrind seems unsupported on osx

iirc people reported that it worked on some versions of osx. It is the best tool, so you might want to consider to have some virtual box/chroot/..

> running it with ASAN enabled gives me the following output:

Please also fix other ASAN issues if possible. You are even assigned for #788 ;)

> And I've also checked the complexity, the analyze method has 13, but 3 are caused by output switches (if cl.verbose etc), so i'd say thats okay, the same for printResults (12, where 2 are caused by output switches)

Please add this information as source code comment.

> so the last big thing to do is to check out how to create tests for it

If you already tested it manually I would prefer to get it merged quickly and get your concentration back to more important things. Do not get me wrong, auto-completion is important, but I am afraid this PR already does way too much (e.g. min/maxdepth).

e1528532:
Ok so i'll leave the parameters for now.

manpage is called 'kdb-complete' (like the one for kdb ls is called kdb-ls), just tried it out it should work.

I'll fix the ASAN issues in a different pull request, as they don't seem to be directly related to this.

I added the mc cabe info to the source code for the two exceeding methods, the others are fine

And yeah i already did quite a lot of manual testing during development, so i've opened a new PR #1294 as in this branch there are these merge issues, and in the other not. Guess its ready to be merged then.

Issue #1286, opened 2017-01-21
reformat-source script does not work on macOS


e1528532:
# Describe what you wanted to do

I wanted to format my c++ code for comitting and thus executed the reformat-source script on macOS from the libelektra root directory.

# Describe what you expected

I expected that it will successfully reformat the sourcecode.

# Describe what actually happened

First, clang-format-3.8 was missing, which i installed using homebrew "brew install clang-format". However, this provides the clang-format command, not clang-format-3.8, i simply changed the script locally for now, alternatively i could have made an alias for clang-format-3.8 to clang-format.

Then it complained that readlink does not support the -f option. The funny thing is that the man pages says it should support this option, but upon execution it tells you
```
readlink: illegal option -- f
usage: readlink [-n] [file ...]
```
 I just stripped the option then it stopped complaining.

After these steps the script worked successfully and did reformat the code. Is there any way to make this work for both platforms, or would it be better to offer a slightly adjusted script for mac? Or is there another solution?

# System Information

- Elektra Version: master
- clang-format
- macOS Sierra

markus2330:
Thanks for reporting! I thought it works nicely thus @sanssecours was the one who initiated using `clang-format` after all and never complained ;)

`clang-format` should be simply to fix by detecting it with `which`. (Prefer `-3.8` if available, otherwise use it without postfix, maybe with version check).

The `-f` option is used nearly everywhere, maybe we should have an `include-common` (like we have `configure-common` now) which fixes these incompatibilities and provides common functions/variables for shellscripts?

sanssecours:
> I thought it works nicely thus @sanssecours was the one who initiated using `clang-format` after all and never complained ;)

As far as I can remember I never used `reformat-source`. I just use the [”Reformat Document”](https://github.com/textmate/c.tmbundle/blob/master/Commands/Reformat%20Document.tmCommand#L9-L12) command of [TextMate](https://github.com/textmate/textmate)’s C-Bundle, which reformats the current file after you press <kbd>Ctrl</kbd> + <kbd>⇧</kbd> + <kbd>H</kbd>.

PR #1295, opened 2017-01-28
macOS-reformat-source

e1528532:
# Purpose

fix the issue with reformat-source not working on macOS (#1286) by detecting clang-format via which, and by providing a common include file "include-common" which provides an adjusted readlink function, as readlink -f is not available on macOS.

On macOS it tries to use greadlink, which is the same as readlink on linux given the user installed coreutils via homebrew. Otherwise i've found a shellscript which provides the required functionality to emulate at least readlink -f (we only use that at the moment, no other arguments for readlink in our scripts anywhere, so it should be good for now and avoids enforcing brewing coreutils).

# Checklist

- [ ] I added unit tests

@markus2330 what do you think about this approach? It fixes the issue with the script on my mac, both with coreutils brewed and without. If its okay that way, i'd include it in all other scripts using readlink -f.

### DIFF
--- a/scripts/cbmc
+++ b/scripts/cbmc
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-SELF=$(readlink -f "$0")
+. "$(dirname "$0")/include-common"
 
 if [ "$#" -ne 1 ] || ! [ -d "$1" ]
 then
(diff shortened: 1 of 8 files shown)


e1528532:
Updated all the scripts to use the common includes where it made sense to do so. Primarily i wanted to ensure that every readlink -f is redirected properly on osx. I think this can be merged.

markus2330:
> The licence

The license text says so, so let us better do it.

e1528532:
done

PR #1300, opened 2017-01-30
spec: ns fix

tom-wa:
# Purpose

in some cases spec treated spec-keys as normal keys when globbing - fixed


markus2330:
Thank you!

Issue #701, opened 2016-05-07
Property Plugin


markus2330:
Hello newcomer,

Thus simpleini uses glibc specific features a replacement should be written.

As initial version it is enough if it writes out/parses every configuration option in the format `path/to/key = value`. Later you can add support for comments, ordering and configurable separators (instead of `/` and `=`).

To achieve this, you have to create a new plugin (see https://github.com/ElektraInitiative/libelektra/blob/master/doc/tutorials/plugins.md ). You only have to implement kdbGet and kdbSet.

To complete the task, please:
- [ ] implement as described above
- [ ] fully fill out README.md. For tricky parts (provides, placements) you can peek into the READMEs of other storage plugins.
- [ ] handle all errors (also peek in other storage plugins what to do on I/O errors)
- [ ] add unit tests demonstrating that a file read and written again stays the same (you guessed it, other storage plugins do the same)
- [ ] remove not needed parts of template (open, close, error)

You do not need to escape the key or value, this should be handled by the rename and code plugins.

Please write something below before starting the task. You can always ask here for help.

markus2330:
mini was already implemented by @sanssecours

PR #1301, opened 2017-01-30
MSR: replace grep -P with sed

tom-wa:
# Purpose

replaced `grep -P` with `sed`

### DIFF
--- a/src/plugins/conditionals/README.md
+++ b/src/plugins/conditionals/README.md
@@ -101,7 +101,7 @@ sudo kdb mount main.ini /examples/conditionals ni
 sudo kdb mount sub.ini /examples/conditionals/sub ini
 
 # mount conditionals as global plugin
-sudo kdb global-mount conditionals
+sudo kdb global-mount conditionals || $(exit 0)
 
 # create testfiles
 cat > `kdb file /examples/conditionals` << EOF \
@@ -113,9 +113,9 @@ EOF
 echo "key = false" > `kdb file /examples/conditionals/sub`
 
 # should fail and yield an error
-kdb export /examples/conditionals ini
-#> sub/key = false
+kdb export /examples/conditionals simpleini
 #> key1 = val1
+#> sub/key = false
 # ERRORS:135
 # Error (#135) occurred!
 # Description: Validation failed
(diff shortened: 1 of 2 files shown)


src/plugins/conditionals/README.md
```diff
@@ -101,7 +101,7 @@ sudo kdb mount main.ini /examples/conditionals ni
 sudo kdb mount sub.ini /examples/conditionals/sub ini
 
 # mount conditionals as global plugin
-sudo kdb global-mount conditionals
+sudo kdb global-mount conditionals || $(exit 0)
```
markus2330: I added #1306 to revert this later

src/plugins/conditionals/README.md
```diff
@@ -113,9 +113,9 @@ EOF
 echo "key = false" > `kdb file /examples/conditionals/sub`
 
 # should fail and yield an error
-kdb export /examples/conditionals ini
-#> sub/key = false
+kdb export /examples/conditionals simpleini
```
markus2330: We should avoid simpleini, it currently does not work on macOS. (In particular such a change is strange in a PR attempting to do macOS fixes :laughing:) Any reason why you used it here?

src/plugins/conditionals/README.md
```diff
@@ -129,9 +129,9 @@ kdb export /examples/conditionals ini
 kdb set /examples/conditionals/sub/key true
 
 # should succeed 
-kdb export /examples/conditionals ini
-#> sub/key = true
+kdb export /examples/conditionals simpleini
```
markus2330: again

src/plugins/conditionals/README.md
tom-wa: oh, didn't know that.
i changed it to simpleini because it's easier to read in tutorial after ini started to support metakeys (the metakey would be printed to stdout too)
i could add a `nometa` flag to ini though.

markus2330: @sanssecours said he will add a simpleini replacement for MacOS.  #701

We should not make `ini` even more complicated.

markus2330:
Thank you!

Already merged as part of  #1456

Issue #1176, opened 2016-12-08
Rate Tutorials


markus2330:
Extending on #280 we should summarize the current state of the tutorials. For which tutorials do you think we need improvements, which need to be newly written?

- [ ] give a rating for current tutorials

sanssecours:
# Overview

| Tutorial                  | Rating |
| ------------------------- | ------ |
| Application Integration   |   +    |
| Cascading                 |   ++   |
| Compilation Variants      |   ◦    |
| Elektra Merge Integration |   ?    |
| Export                    |   ++   |
| Import                    |   +    |
| Merge                     |   ◦    |
| Mounting                  |   +    |
| Namespaces                |   +    |
| Plugins                   |   +    |
| Python                    |   +    |
| REST Service              |   +    |
| Validation                |   ◦    |

The rating above apply if we merge pull request #1259.

## Application Integration

More concrete examples would be nice. Maybe we could add a  step by step tutorial , that describes how to create a very simple application (e.g. “Hello Elektra”) that uses Elektra in Linux/macOS.

## Compilation Variants

I am not really sure how helpful this tutorial is. I guess I will see if I ever need to define compilation variants. For now I think the tutorial seems rather short, for a description of a process that might be complicated. However, I think we can leave it as is, and add additional information later if someone discovers important data that might be missing currently.

## Elektra Merge Integration

I did not know what “ufc” is before I read this tutorial. Reading the tutorial did not change this fact. It might make sense to add a link to the homepage of “ufc” or an article that describes this tool, if such a page exists.

## Import

I think we should add some concrete examples a user can try by himself here.

## Merge

I think we should add `kdb` commands to the tutorial that generate the keysets `system/base`, `system/ours`, `system/base`.

The description of the strategies makes sense, when we only look at `key4`. However I do not understand how the strategies determine some of the other keys in the merged keyset. For example, why do we lose `key2` if we use the strategy `ours`?

## Validation

### Customized Schemas

The examples in the section “Customized Schemas” do not work on my computer. The command

```sh
kdb spec-mount /tutorial/schema
```

fails, displaying the following error message:

```
The command kdb spec-mount terminated unsuccessfully with the info:
No plugin that provides conflict/set/missing could be found
```

. I also quite dislike the use of here document syntax, since my favorite shell does not, and quite possibly will never, support this [feature](https://github.com/fish-shell/fish-shell/issues/540).

markus2330:
Thanks for the great feedback!

[...]

PR #1311, opened 2017-02-01
tutorial-improvements

e1528532:
# Purpose

In the course of #1176 i've checked the text via a linter (https://github.com/btford/write-good) and saw that there is probably some room for improvement regarding words which don't really add much value to a sentence like only or using too much passive form. Therefore i reworked some parts which i think can be improved.

# Checklist

@markus2330 do you think this way we make the tutorials easier to read? If you do I will also try to improve the other tutorials as well.

### DIFF
--- a/doc/tutorials/application-integration.md
+++ b/doc/tutorials/application-integration.md
@@ -69,8 +69,8 @@ or cross-cutting concerns.
 To sum up, if a developer wants to **elektrify** software, he or she can
 do that without any need for changes to the outside world regarding the
 format and semantics of the configuration. In the interconnected
-world it's a matter of time until other software also wants to
-access the configuration, and with elektrified software it's possible
+world it is a matter of time until other software also wants to
+access the configuration, and with elektrified software it is possible
 for every application to do so.
 
 
@@ -107,7 +107,7 @@ Key *parentKey = keyNew("/sw/org/myapp/#0/current", KEY_END);
   - `current` is the [profile](/src/plugins/profile/README.md)
       to use. Administrators need it
       if they want to start up applications with different configurations.
-- `KEY_END` is required by C needs a proper termination of variable
+- `KEY_END` as C needs a proper termination of variable
     length arguments.
 
 The key name is standardized to make it easier to locate configuration.
@@ -146,7 +146,7 @@ Now we have everything ready to fetch the latest configuration:
 kdbGet(repo, conf, parentKey);
 ```
 
-Note it's important for applications that the parentKey starts with a slash `/`.
+Note it is important for applications that the parentKey starts with a slash `/`.
 This ensures pulling in all keys of the so-called [namespace](/doc/help/elektra-namespaces.md).
 Such a name cannot physically exist in configuration files, but they are
 the most important key names to actually work with configuration within
@@ -170,7 +170,7 @@ in configuration files, but are "representatives", "proxies" or
 "logical placeholders" for keys from any other [namespace](/doc/help/elektra-namespaces.md).
 
 So that every tool has a consistent view to the key database
-it's vital that every application does a `ksLookup` for every
+it is vital that every application does a `ksLookup` for every
 key it uses. So even if your application iterates over keys,
 always remember to do a [cascading](cascading.md) lookup for every single key!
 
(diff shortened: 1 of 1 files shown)


doc/tutorials/application-integration.md
```diff
@@ -185,12 +180,12 @@ Thus we are interested in the value we use:
 char *val = keyString(k);
 ```
 
-Finally, we need to convert the configuration value to the datatype we
+Now we need to convert the configuration value to the data-type we
```
markus2330: strange recommendations: "now" is not a good style, "finally" is not necessary but okay.

data-type also seems very non standard: it seems to be "data type" see #920

doc/tutorials/application-integration.md
```diff
@@ -205,39 +200,39 @@ For more information about that, continue reading
 
 ## Specification ##
 
-Now, we have a fully working configuration system without any hard-coded
-information (such as configuration files). So we already gained something.
+Now, we have a fully working configuration system without any hardcoded
```
markus2330: All forms seem to be acceptable: https://en.wikipedia.org/wiki/Hard_coding

Why change it?

Does the tool allow automatic transformation so that we have it easier in future when we change it?

doc/tutorials/application-integration.md
```diff
@@ -205,39 +200,39 @@ For more information about that, continue reading
...
 
 ```sh
 kdb set /overrides/test "example override"
 sudo kdb setmeta spec/test override/#0 /overrides/test
 ```
 
-This technique provides complete transparency how a program will fetch a configuration
-value. In practice that means that:
+This technique provides complete transparency how a program will fetch a
+configuration value. In practice that means that:
```
markus2330: Please separate formatting changes with other changes in future. It is difficult to read this PR.

(At least different commits!)

doc/tutorials/application-integration.md
```diff
@@ -30,57 +30,54 @@ additional benefits arise:
...
+But they can also use self-written plugins for adding needed behavior
 or cross-cutting concerns.
 
 To sum up, if a developer wants to **elektrify** software, he or she can
 do that without any need for changes to the outside world regarding the
-format and semantics of the configuration.  But in the interconnected
-world it is only a matter of time until other software also wants to
-access the configuration, and with elektrified software it is possible
+format and semantics of the configuration. In the interconnected
+world it's a matter of time until other software also wants to
```
markus2330: "it is" is preferable in more formal English. For tutorials both are okay.

doc/tutorials/application-integration.md
e1528532: Right, so we should still decide for a common style for our usage and not mix it up. a quick grep indicates that apart from some news that probably don't matter hardcode or its variations are only used in the application integration tutorial, so its hard to find a majority to follow here.

The tool is a linter, so all it does is complaining :)

doc/tutorials/application-integration.md
e1528532: you are right, will do in future commits

doc/tutorials/application-integration.md
e1528532: it doesn't display recommendations, it only complains that finally isn't really necessary. but you are right now isn't really necessary either imo its perfectly fine to start with We.
fixed data type.

doc/tutorials/application-integration.md
e1528532: i guess thats a matter of preference. i think it's is a bit easier to read and fine to use unless writing a paper. But grep reveiled that it is is much more common throughout our documents than it's though, so ill stick to that.

doc/tutorials/application-integration.md
markus2330: Maybe we should write a script that automatically does the replacements to a canonical form. Discussions for that in #920

markus2330:
Thank you!

Is this ready to merge? The flag shows still "work in progress"? Its better if we keep PRs small and merge them soon.

e1528532:
i'm currently checking and rating the other tutorials too, but i can do a separate PR for each of them if i do some changes. In that case, it is ready to merge.
Regarding formatting, do we have some standards for that? I currently try to aim for a default width of ~80 chars. To preview my changes, i simply check how it looks in my forked repository after comitting, i hope this ensures that it will also look good on the website afterwards.

markus2330:
> Regarding formatting, do we have some standards for that?

I always used the unix tool `fmt`. If there are no objections, we could make it standard in `doc/CODING` "Markdown Guidelines" section.

markus2330:
Thank you! That is a nice PR! Looking forward to improvements in the other tutorials!

Issue #1291, opened 2017-01-23
improved KDB logic


markus2330:
- [x] kdbGet should never fail by default (to allow admins to fix anything) #2955
- [x] on request programs should be able to fail `kdbGet`, maybe using #868
- [ ] should check if `system/elektra` is updated and remount everything if needed
- [ ] get everything (mmap) so that all relevant keys are available for links

PR #1321, opened 2017-02-05
Prepare types

tom-wa:
# Purpose

prepare type plugins for type dispatcher

# Checklist

- [ ] I added unit tests

### DIFF
--- a/src/plugins/date/date.c
+++ b/src/plugins/date/date.c
@@ -399,7 +399,7 @@ static int rfc822StringValidation (const char * date)
 	return -1;
 }
 
-static int validateDate (Key * key, Key * parentKey)
+static int validateKey (Key * key, Key * parentKey)
 {
 	const Key * standard = keyGetMeta (key, "check/date");
 	const Key * formatStringMeta = keyGetMeta (key, "check/date/format");
@@ -480,7 +480,7 @@ int elektraDateGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_UN
 		const Key * meta = keyGetMeta (cur, "check/date");
 		if (meta)
 		{
-			int r = validateDate (cur, parentKey);
+			int r = validateKey (cur, parentKey);
 			if (r == 0)
 			{
 				rc = -1;
@@ -501,7 +501,7 @@ int elektraDateSet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_UN
 		const Key * meta = keyGetMeta (cur, "check/date");
 		if (meta)
 		{
-			int r = validateDate (cur, parentKey);
+			int r = validateKey (cur, parentKey);
 			if (r == 0)
 			{
 				rc = -1;
--- a/src/plugins/date/date.c
+++ b/src/plugins/date/date.c
@@ -413,6 +413,7 @@ static int validateDate (Key * key, Key * parentKey)
 		if (rc == -1)
 		{
 			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_D_T_FMT, parentKey, "%s doesn't match format string %s", date, formatString);
+			rc = 0;
 		}
 	}
 	else if (!strcasecmp (stdString, "ISO8601"))
@@ -425,11 +426,11 @@ static int validateDate (Key * key, Key * parentKey)
 						    formatString);
 			else
 				ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_D_T_FMT, parentKey, "%s is not a valid ISO8601 date", date);
+			rc = 0;
 		}
 		else if (rc == 0)
 		{
 			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_D_T_FMT, parentKey, "syntax error in ISO8601 format string '%s'", formatString);
-			rc = -1;
 		}
 	}
 	else if (!strcasecmp (stdString, "RFC2822"))
@@ -438,6 +439,7 @@ static int validateDate (Key * key, Key * parentKey)
 		if (rc == -1)
 		{
 			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_D_T_FMT, parentKey, "%s doesn't match rfc2822 specification", date);
+			rc = 0;
 		}
 	}
 	else if (!strcasecmp (stdString, "RFC822"))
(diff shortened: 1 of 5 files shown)


src/plugins/path/path.c
```diff
@@ -12,6 +12,60 @@
 #include "kdbconfig.h"
 #endif
 
+static int validateKey(Key *key, Key *parentKey)
```
markus2330: Seems like code is not formatted yet. And parentKey should be renamed to errorKey?

Don't you want to export this function? (Why is it static?)

tom-wa:
yeah, i'm still working on it because the current signature doesn't really work for the boolean checks.

markus2330:
Please try to keep the interface as simple as possible. For boolean: Is canonicalization and the validation separated? Why isn't the interface enough?

tom-wa:
boolean has some hardcoded default values + to plugin configuration

tom-wa:
how should keys that fail to validate on kdbGet be handled ? 
should kdbGet fail with an error, or just delete all keys that fail validation (+ maybe a warning) ?

markus2330:
> how should keys that fail to validate on kdbGet be handled ? should kdbGet fail with an error, or just delete all keys that fail validation (+ maybe a warning) ?

Remove them **with** a warning and let us write a plugin that allows us to turn this warning into an error. (For applications that want to fail.) For admins and so on there should always be a way with Elektra to fix the situation, and if `kdbGet` would fail, they could not fix it. See #1291

tom-wa:
should the validation plugins remove the key or only the type dispatcher ? 
does that mean validation plugins should only set warnings but never errors ?

markus2330:
> should the validation plugins remove the key or only the type dispatcher ?

I would say the type dispatcher. Maybe we can even configure the type dispatcher for different behaviour?

> does that mean validation plugins should only set warnings but never errors ?

On writing configuration they certainly set errors and fail!

tom-wa:
>I would say the type dispatcher. Maybe we can even configure the type dispatcher for different behaviour?

if it's done by the type dispatcher we can use `validateX(Key * key, Key * errorKey)` as signature. The validation plugins would use the parentKey as errorKey, the type dispatcher a new key and then sets a warning with the description/text of the error. 

what do you think about generating both SET_ERROR and ADD_WARNING macros for every entry in `error/specification` ?

markus2330:
> if it's done by the type dispatcher we can use validateX(Key * key, Key * errorKey) as signature.

Sounds good!

> The validation plugins would use the parentKey as errorKey, the type dispatcher a new key and then sets a warning with the description/text of the error.

Okay!

> what do you think about generating both SET_ERROR and ADD_WARNING macros for every entry in error/specification ?

I am not in favour of it. It should be a rare exception that errors/warnings are up/downgraded.

Maybe you are right and we should let the validation plugins do nothing in `kdbGet` (neither iterate, nor check, no warning). Then only the type dispatcher allows us to validate within `kdbGet`, which should be okay.

We need an architecture document for these decision, can you write such a decision about validation? (With complete rationale and all pro/cons)

src/plugins/date/date.c
```diff
@@ -465,6 +465,7 @@ int elektraDateGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_UN
 			       keyNew ("system/elektra/modules/date/exports", KEY_END),
 			       keyNew ("system/elektra/modules/date/exports/get", KEY_FUNC, elektraDateGet, KEY_END),
 			       keyNew ("system/elektra/modules/date/exports/set", KEY_FUNC, elektraDateSet, KEY_END),
+			       keyNew ("system/elektra/modules/date/export/validateKey", KEY_FUNC, validateKey, KEY_END),
```
markus2330: Its a Elektra key, we should avoid camelcase here. So either `validate` (should be enough?) or `validate/key` (if we might need further validation methods but seems unlikely: validating of keyset is already `kdbSet`). And there is a typo: `export` -> `exports`.

tom-wa:
@markus2330 i have no idea how to export the validation function from the types plugin, could you do that ?

markus2330:
Yes, should be easy.

Basically we only need to wrap `bool check (Key & k)`.
Can you please document the API of the validation function? In particular:

- signature
- will open/close be called as expected?
- Are there some restrictions which errors needs to be set?
- return values?

tom-wa:
`int validateKey(const Key *key, const Key *errorKey)`
returns 1 if validation succeeded, 0 if validation failed
not really sure if an additional return value for invalid syntax is needed.
if an error occurred it's set on errorKey

open/close is not called, is it needed somewhere ?

markus2330:
In the current version of the plugin `type`, `open` builds up a data structure (containing all types) which is used during checking keys. I could build up the data structure in the wrapper as well but I think it is somewhat expected that a plugin gets opened before anything is done with it? Would it be a problem to introduce this behavior?

And can you please add the docu within the source, please?

And we should follow the plugin convention and fail with -1? Or were you thinking about booleans as return value?

tom-wa:
i forgot that elektraPluginOpen and elektraPluginClose call open/close by default, so open/close do get called.

initially i wanted to use 1 for success, 0 for error and -1 for syntax errors in the check/X metakeys. but i'm not so sure about the last party anymore. what do you think ?

markus2330:
> initially i wanted to use 1 for success, 0 for error and -1 for syntax errors in the check/X metakeys. but i'm not so sure about the last party anymore. what do you think ?

I also was thinking about some intermediate state (like "could be okay, depends on other keys"), but it seems like that we do not need it. A simpler design is better, so lets stay with a boolean as return value.

tom-wa:
now that you mention "depends on other keys" i realized i've forgotten about `conditionals` and `math` which both need the whole keyset + a parentkey if keynames relative to the parentkey are used. so `int validateKey(const Key *key, const Key *errorKey)` won't work for them. 
it should be `int validateKey( const Key *key, KeySet *ks, Key *errorKey)`

markus2330:
The question is if the do not better fit into the transformation plugin (especially with the assignment feature).

I think it would be better if you wrap up the type checker for now. Let us merge it, test it a bit, and progress from there.

So please fix docu, comments and so on and set the PRs to ready to merge.

markus2330:
The type plugin now exports validateKey, too.

Would be great to get the validation for lcdproc merged, soon!

markus2330:
Thank you, great work!

PR #1325, opened 2017-02-06
Add KDB Completions for Fish

sanssecours:
# Purpose

This pull request adds [Fish completions](http://fishshell.com/docs/current/index.html#completion) for the command line tool `kdb`. I tested the code in this pull request with the **changes of PR #1310 already applied**. It might not make sense to test this pull request without merging #1310 first, since the namespace completion will fail otherwise.

## Installation

The steps below show one way to test the completions included in this pull request.

1. Install [Fish](http://fishshell.com)
2. Create the directory `~/.config/fish/completions`
3. Change the current working to `~/.config/fish/completions`

   ```sh
   cd ~/.config/fish/completions
   ```

4. Create a symlink to `kdb.fish`

   ```sh
   ln -s "Location of Elektra Repository"/scripts/kdb.fish kdb.fish
   ```

# Checklist

- [x] As far as I can tell the commit messages should be correct.


scripts/kdb.fish
```diff
@@ -0,0 +1,492 @@
...
+    set metanames $metanames 'check/condition' 'deprecated' 'internal/<plugin>/*' 'source' 'dependency/control' 'dependency/value'
+    set metanames $metanames 'application/name' 'application/version' 'restrict/write' 'restrict/null' 'restrict/binary' 'restrict/remove'
+    set metanames $metanames 'evaluate/<language>' 'uid' 'gid' 'mode' 'atime' 'mtime' 'ctime' 'spec' 'proc' 'dir' 'user' 'system'
+    set metanames $metanames 'comment/#' 'comment/#/start' 'comment/#/space' 'csv/order' 'crypto/encrypt' 'crypto/salt'
+    printf '%s\n' $metanames
+end
+
+function __fish_kdb_print_namespaces -d 'Print a list of possible namespace completions'
+    set -l namespace (commandline -ct)
+    kdb complete --max-depth=1 -- "$namespace" | string match -vr '(dir|proc|spec|user)$'
```
markus2330: Is system missing in here?

sanssecours: I just assumed the namespace `system` does always exist. Is this not the case?

markus2330:
Impressive work!

Please readd "ready to merge" to confirm that "system" is not missing.

Are you sure you want to wait for #1310, it does not seem to me that there is an actual code dependency but instead it is only a bug that can be triggered by your code?

@reox Do you want to take a look?

scripts/CMakeLists.txt
```diff
@@ -52,9 +52,18 @@ if (INSTALL_SYSTEM_FILES)
 			DESTINATION ${BASH_COMPLETION_COMPLETIONSDIR}
 			RENAME kdb)
 
+	set (FISH_COMPLETION_COMPLETIONSDIR "/usr/share/fish/vendor_completions.d")
+	find_package (PkgConfig)
+	if (PKG_CONFIG_FOUND)
+		include (LibFindMacros)
+		pkg_get_variable (FISH_COMPLETION_COMPLETIONSDIR fish completionsdir)
```
markus2330: Please check first if the module fish exists to avoid the error:

```
Package fish was not found in the pkg-config search path.
Perhaps you should add the directory containing `fish.pc'
to the PKG_CONFIG_PATH environment variable
No package 'fish' found
CMake Error at scripts/CMakeLists.txt:61 (install):
  install FILES given no DESTINATION!
```

scripts/CMakeLists.txt
```diff
@@ -52,9 +52,18 @@ if (INSTALL_SYSTEM_FILES)
...
+	endif ()
+	install (FILES kdb.fish
```
markus2330: Please only install if `INSTALL_SYSTEM_FILES` is set.

scripts/kdb.fish
markus2330: All namespaces do always exist. In the system namespace you are very likely to always find some keys because the built-in version keys do currently live there.

But why is such an assumption valuable?

scripts/CMakeLists.txt
sanssecours: The problem should be fixed in commit [ef38dbd0](https://github.com/sanssecours/libelektra/commit/ef38dbd0e489039bec4bd20c3b8f411c70c90bfd).

scripts/CMakeLists.txt
sanssecours: Is this not already the case? The whole code is in inside the conditional `if (INSTALL_SYSTEM_FILES)` after all.

scripts/kdb.fish
sanssecours: > But why is such an assumption valuable?

It is not that valuable. It allows us to save 6 characters of code though 😀.

sanssecours:
> Are you sure you want to wait for #1310, it does not seem to me that there is an actual code dependency but instead it is only a bug that can be triggered by your code?

I do not think that the code triggers a bug, but rather that it relies on features, such as “empty namespace” completion, that are not part of the current version of `kdb complete` yet.

Most of the completion code should work even if #1310 is not merged yet. If someone wants to test the completions I would **strongly recommend**, that they wait until someone merges pull request #1310 though. 

In conclusion: If you want to merge the code now, that is fine with me. However, it would make sense to wait until the code in pull request #1310 is part of the repository.

Issue #1323, opened 2017-02-05
scripts/include-common  [[: not found


tom-wa:
```
./reformat-source: 12: ./include-common: [[: not found
```
```
 % ../scripts/configure-debian-musl                                                                                                                                             [128]
../scripts/configure-debian-musl: 12: ../scripts/include-common: [[: not found
../scripts/configure-debian-musl: 12: ../scripts/configure-common: [[: not found
CMake Error: The source directory "/home/thomas/Dev/libelektra/build/-KDB_DEFAULT_STORAGE=ini" does not exist.
```

markus2330:
@e1528532 is this a regression from #1295?

e1528532:
Looks like it, i think the issue is that [[ is only available in bash, zsh and similar. I assume you use debian? It looks like debian uses /bin/dash per default which is quite limited compared to bash.

Since its probably better to avoid such bash specific extensions and try to use whatever the original shell specification gives you due to its wider compatibility, does the following patch fix your problem @tom-wa?
https://gist.github.com/e1528532/f73243d211ea5611c279eebe544c9f5b

I've tried it and it works in bash, zsh and dash both on a arch linux and macOS now, so basically as long as the shell supports the sh specification.

markus2330:
> It looks like debian uses /bin/dash per default which is quite limited compared to bash.

In Debian `/bin/sh` is a symlink, in newer installations usually to `dash`. But it could be to another shell. If you state in the shebang `/bin/sh` you claim you exclusively use POSIX shell features. Thus `dash` does not support much more than POSIX, it is a very good test.

> git.github.com

Please always prefer PRs, then the build server can build it and it is much easier for us to apply the patch.

> I've tried it and it works in bash, zsh and dash both on a arch linux and macOS now

Then there no need to let it other people try it, too ;) Thanks for the extensive tests! I am looking forward to the PR.

Can you please also add a unit test of something that includes scripts/include-common to avoid future regressions?

e1528532:
done and closed in #1326

markus2330:
You can also write `close #xxx` in the commits, then the issues will be closed automatically.

PR #1326, opened 2017-02-06
macOS-reformat-source: fix include-common on dash and fix test for mac

e1528532:
# Purpose

fix the include-common script for dash (it used some bash specific extensions while it should use posix sh instead)

# Checklist

- [X] I added unit tests

@markus2330 here is the fix for #1323 . There was already a test in tests/shell/check_formatting.sh which calls the reformat-source script and thus also checks the include-common, i only fixed it to support clang-format instead of clang-format-3.8 too.

### DIFF
--- a/scripts/include-common
+++ b/scripts/include-common
@@ -6,16 +6,19 @@
 # load our readlink replacement for osx
 . "$(dirname "$0")/realpath"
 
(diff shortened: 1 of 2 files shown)


markus2330:
Thank you!

PR #1337, opened 2017-02-12
Format Project with ClangFormat 4

sanssecours:
# Purpose

This pull request reformats the C and C++ files in the repository with the current version of ClangFormat.

# Checklist

- [x] I double-checked all commit messages.

### DIFF
--- a/scripts/reformat-source
+++ b/scripts/reformat-source
@@ -12,6 +12,6 @@ if [ -z "${CLANG_FORMAT}" ]; then
 	CLANG_FORMAT=$(which clang-format)
 fi
 
-cd $SOURCE
+cd "$SOURCE"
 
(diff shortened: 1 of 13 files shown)


markus2330:
Thanks for creating the PR, I would love to merge it right now but unfortunately it involves quite some efforts on the build agents (we would need to upgrade llvm on every agent) and also on every other computer.

So I would prefer if we can delay this until @BernhardDenner reinstalls the build agents.

The issue with folders containing spaces, however, could be merged right now. Do you want to reduce this PR to contain only the fix?

sanssecours:
> The issue with folders containing spaces, however, could be merged right now. Do you want to reduce this PR to contain only the fix?

I cherry-picked the first commit in this pull request and applied it to master myself. Hope that is okay with you.

markus2330:
Thank you, that is perfect!

PR #1034, opened 2016-10-20
Documentation markdown improvements

Namoshek:
# Purpose

Improve markdown syntax of documentation files so that they work well on the website as described in #1030. Also unification of documentation styles.
# Changed files
- README.md of almost all plugins
- src/libs/README.md (absolute paths of individual libs changed to relative paths)
# Other documentation

All documentation files could need some overhaul, but it is really a lot of work, so I'd like to share it with others.

@markus2330 please review my pull request

PR #1338, opened 2017-02-12
Improve formatting

domhof:
# Purpose

(Long descriptions should be in documentation, not here)

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/doc/decisions/high_level_api.md
+++ b/doc/decisions/high_level_api.md
@@ -20,7 +20,7 @@ Limitations:
 ## Assumptions
 
 - Thread-safety: a handle is the accepted better solution than having to
-  care about whether it is reentrant, thread-safe,..
+  care about whether it is reentrant, thread-safe, ...
 - assumes that spec is available and installed correctly (fail in kdbhlOpen otherwise)
 - lookups for non-specified keys yield errors (in particular if they are not present)
 - many projects do not care about some limitations (no binary, no meta-data)
@@ -42,7 +42,7 @@ First draft of API:
 
 #### Basic
 
-```
+```c
 // might fail, you need to check for error afterwards!
 KDBHL * kdbhlOpen (const char * application);
 
@@ -66,7 +66,7 @@ void kdbhlClose (KDBHL * handle);
 
 #### Needed
 
-```
+```c
 // might fail, you need to check for error afterwards!
 void kdbhlReload (KDBHL * handle);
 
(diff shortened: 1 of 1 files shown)


markus2330:
Thanks for creating the PR!

It is a good idea to add fences, but why not saying which language is used? This improves syntax highlighting.

And in `doc/CODING.md` (Markdown Guidelines) we decided to use the `#` of titles on the left and right side, please revert these changes.

domhof:
Thanks for the feedback! I've added the language specifiers. What is the benefit of having the `#` on both sides of the title?

markus2330:
> What is the benefit of having the # on both sides of the title?

It is only a matter of taste, I also had it mostly done as you propose. But we decided to do it that way as documented and reformatted everything consistently thereafter. So the benefit now is consistency.

domhof:
Okay, I've added the '#' as well. However, since markdown neither enforces it nor does it check if it is the same number of '#' on both sides, I would argue that duplicating the '#' on the right side is only a potential source of error in terms of consistency. So for this reason I would be happy if we could rethink this decision in order to keep things simple and tidy (after all, it will have an impact also on future documents for a long time to come).

doc/decisions/high_level_api.md
```diff
@@ -102,11 +134,13 @@ What is not so nice:
 ## Argument ##
 
 1. Very easy to get started with, to get a key needs 3 lines of codes:
-   ```c
-   KDBHL *handle = kdbhlOpen ("/sw/elektra/kdb/#0/current");
-   printf ("number /mykey is %d\n", kdbhlGetInt (handle, "/mykey"));
-   kdbhlClose (handle);
-   ```
+   
+```c
```
markus2330: I am not sure if every markdown renderer will handle this as expected. Some might start counting again from `1.` below? Not really important, so I will merge anyway.

markus2330:
Thank you! For future commits, please use `module: ...` syntax in the commit messages, where module in this case would be `doc:`.

> However, since markdown neither enforces it nor does it check if it is the same number of '#' on both sides, I would argue that duplicating the '#' on the right side is only a potential source of error in terms of consistency. So for this reason I would be happy if we could rethink this decision in order to keep things simple and tidy (after all, it will have an impact also on future documents for a long time to come).

When we had the requirement the other way round, it was inconsistent, too (much more than it is now). I never saw the case that the number of `#` on the left and right side were differently. So I would argue it really does not matter at all.

But if it is important to you, please create a proposal for a "better" markdown style. But please be aware that reformatting the current base is a considerable effort. Hopefully we will find a style-formatter for Markdown soon, that would be the easiest solution. (We tried some, see #1034, but they were not configurable and we did not find time to investigate them further.)

Issue #1237, opened 2016-12-27
add depth/levels option to kdb ls


tom-wa:
there should be a way to limit the depth of kdb ls (and probably kdb export too)

markus2330:
Thanks for the input! Can you please also describe the usecase you think of?

Its obvious why we need a list of keys that are direct below (so to say "non-recursive"), but why do we need other depths?

Btw. the issue might be related to #1167

tom-wa:
keys direct below would the the primary usecase, other levels would be needed for special usecases (and most likely used together with / parsed for some other option like the actual keybasename) 

kinda related: list/output only keynames having a metakey or keybasename specified as a ls or export option

markus2330:
> other levels would be needed for special usecases (and most likely used together with / parsed for some other option like the actual keybasename)

Can you name one of the special usecases? If there are none, we should rather go for recursive/non-recursive which is a more convenient interface than a `--max-depth=<number>`.

> kinda related: list/output only keynames having a metakey or keybasename specified as a ls or export option

Yes, it is related: `max-depth` is something which I would expect in such a query interface, but not so in a ls/list interface.

Your use case can be implemented quite easily as plugin, we would only need a way to specify an arbitrary number of plugins to `kdb export`, see suggestion in #1229.

Btw.  #867 is related: it suggests that we allow to have plugins as filter, also on API level. But #867 is actually not needed to export using a filter.

markus2330:
@e1528532 should be a very easy task thus `kdb complete` already implements this feature.

e1528532:
You are right, it was not that much effort anymore.

I was not sure if you want a max-depth or a recursive option now, but since kdb complete already makes use of max-depth and min-depth, i've tried to archieve a similar behavior for kdb ls to have a more or less uniform meaning of those two parameters.

This might seem a bit strange though for kdb ls. Consider `kdb ls user/test --max-depth=1`, it would give you user/test as a result (its just showing the current level that way. If you want to show the next level, you would type `kdb ls user/test/` as the last slash indicates that we want to see whats under this key. But thats how it has to be if we want the same behavior as for kdb complete.

If you make no use of the options, the output should be exactly the same as it was before, so there is no depth limitation.

See the PR #1345

e1528532:
i think this is done already, closing before we forget this issue completely

Issue #1328, opened 2017-02-08
kdb complete: hangs


markus2330:
# Describe what you wanted to do

run `kdb complete` in a non-trivial mount setup with many keys.

# Describe what you expected

That it shows the completion.

# Describe what actually happened

Does not terminate, here 3 back traces of somewhere within this infinite calculation:

```
#0  0x00007ffff79b33cc in keyGetFullNameSize (key=0xdaa8d1) at /home/markus/Projekte/Elektra/current/src/libs/elektra/keyname.c:609
#1  0x0000000000463efa in getFullNameSize (this=0x7fffffffdca0) at /home/markus/Projekte/Elektra/current/src/bindings/cpp/include/key.hpp:920
#2  getFullName (this=0x7fffffffdca0) at /home/markus/Projekte/Elektra/current/src/bindings/cpp/include/key.hpp:930
#3  CompleteCommand::getParentKey (this=this@entry=0x69a3f0, key=...) at /home/markus/Projekte/Elektra/current/src/tools/kdb/complete.cpp:230
#4  0x0000000000466602 in CompleteCommand::analyze (this=this@entry=0x69a3f0, ks=..., cl=...) at /home/markus/Projekte/Elektra/current/src/tools/kdb/complete.cpp:174
#5  0x0000000000467c09 in CompleteCommand::complete (this=this@entry=0x69a3f0, argument="", cl=...) at /home/markus/Projekte/Elektra/current/src/tools/kdb/complete.cpp:63
#6  0x00000000004686be in CompleteCommand::execute (this=0x69a3f0, cl=...) at /home/markus/Projekte/Elektra/current/src/tools/kdb/complete.cpp:50
#7  0x000000000041a70e in main (argc=2, argv=0x7fffffffe438) at /home/markus/Projekte/Elektra/current/src/tools/kdb/main.cpp:198
```

```
#0  0x00007ffff79b54eb in keyCompareByNameOwner (p1=p1@entry=0x7fffffffdbc8, p2=p2@entry=0x7fffffffdbc0)
    at /home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c:521
#1  0x00007ffff79b567c in keyCmp (k1=0x762890, k2=0xdaadf0) at /home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c:614
[...]

PR #1345, opened 2017-02-15
kdb-ls-depth: Implement depth parameter for kdb ls

e1528532:
# Purpose

Like said, the purpose of this PR is to provide a depth parameter for kdb ls, like we have in kdb complete as requested in #1237

# Checklist

- [x] I added unit tests

@markus2330 please review my pull request, and keep my comments in #1237 in mind.


markus2330:
@tom-wa can you check if you like the behavior?

src/tools/kdb/ls.cpp
```diff
@@ -21,12 +21,9 @@ LsCommand::LsCommand () : kdb (root)
 {
 }
 
-int LsCommand::execute (Cmdline const & cl)
+int LsCommand::execute (const Cmdline & cl)
```
markus2330: Please keep it `<obj> const & cl`, otherwise we have a mix of style.

e1528532: is there any difference between Cmdline const and const Cmdline? I used const before the class type all the time, also in kdb complete. Should i change it too in there?

and what about something like `const Key key` (without &), should it also be `Key const key`?

markus2330: > I used const before the class type all the time, also in kdb complete. Should i change it too in there?

It is not urgent, please not within this PR. Maybe together with #1328 or even seperately.

>and what about something like const Key key (without &), should it also be Key const key?

It actually does not make much sense to do that, at least not like you did it. It will increase/decrease the ref counter unnecessarily.

Also `const int` as formal parameter is of little use, the value is copied anyway, why protect it from writing to it?

The position of const seems to be inconsistent at some other places, e.g. `const char *`, maybe one of the llvm tools can change it automatically?

Btw. it is good that you use `const` so frequently.

e1528532: Ok i have to admit i haven't thought about the ref counter of the Keys yet. You are right in that cases it probably makes more sense to pass them around by reference if applicable to avoid that.

But i meant the general cases where i pass a const object around, for example a string. Should i use `const string foo` or `string const foo` in the future?

Doesn't the const int indicate that this integer will not change in this method? I've seen it more from the perspective of immutability. If i want to express that some parameter should not change in a method, i mark it as const, regardless of wheter its getting passed by copy so modifications wouldn't make a difference outside the method.

markus2330: Okay, its a very strict sense of immutability, which is a bit unusual, e.g. https://isocpp.org/wiki/faq/const-correctness does not recommend it. It is a bit misleading from the interface perspective. But I am okay with it if you use it. (For APIs (like libtools) we should check if all compilers are ABI compatible when `const` needs to be removed).

About the placement of `const`: Anything is okay as long as it is used consistently. It is not so often used in the code base, so even if llvm cannot do it, you can also change it by hand if you prefer one over the other. Otherwise simply use, what is already used.

tests/shell/shell_recorder/kdb_ls.esr
```diff
@@ -0,0 +1,19 @@
+Storage: dump
+Mountpoint: system/tmount/kdb_ls
+DiffType: File
+
+< kdb mount $File $Mountpoint $Storage
+
+< kdb set $Mountpoint/test bla
```
markus2330: You can and should use the tutorial style syntax or did you have any problems with it?

e1528532: Ahh you've already seen it, i was still messing around with the shell recorder. What do you mean with tutorial syntax? I haven't found anything about that in the Readme. Do you mean i should write a tutorial how to use kdb ls and then use this tutorial wrapper?

I'm not really sure how the shellrecorder has to be used correctly yet. I've created my esr script which contains some commands that will be done on a temporary mountpoint to avoid polluting the host system. I think the extra mount was totally unnecessary as it seems the shell recorder already takes care about that, so i removed it. Also i moved the mountpoint to user where everyone can write to without having root rights.

Then i execute this script via the shellrecorder and grab the protocol it produces (kdb_lsprotocol). After checking if the output of each command was as expected in the protocol, the test can be executed by running the shellrecorder script again and comparing the output to the stored protocol file, which should match (and did match when i execute the test directly via `./shell_recorder.sh kdb_ls.esr kdb_lsprotocol`).

But does the test execution (make_run) even compare the output against the protocol file? Inspecting the CMakeFile made me doubt that. Do i have to copy the expected output back in the script file (kdb_ls.esr) and check it everywhere with the STDOUT command? That would be much more cumbersome than simply comparing against the protocol.

And in all the .esr files it uses the kdb commands directly, but if i don't use e.g. kdb ls instead of just ls, it will fail as it actually tries to use my system's ls command and not the one of kdb.

e1528532:
checked out the tutorial syntax and used the example section in the documentation of kdb-ls to implement the test cases, so that they serve as examples for users at the same time. The syntax looks not perfect, but still a win/win i think.

tom-wa:
i can't get the short options to work
the only thing thats a little bit annoying is the difference between having and not having a / at the end 

```
kdb ls system/info/elektra --max-depth=1
kdb ls system/info/elektra/ --max-depth=1
system/info/elektra/constants
system/info/elektra/uname
```

e1528532:
fixed the short option argument issue, depends on PR #1510  now.
that behavior with the slash at the end is intended as explained in #1237 to be consistent with how min and max depth work with kdb complete. If we want that consistency.

e1528532:
if we are fine with that, we can merge this i think

markus2330:
jenkins build fast please

markus2330:
Looks good, thank you. But seems like we have a conflict here 😢 please rebase.

sanssecours:
Sorry, I should not have resolved the conflict 😳. I just wanted to try GitHub’s online editor. Anyway, as far as I can tell, rebasing should work the same as always. I already rebased the PR locally. Sorry again, if commit cc0afe6 caused any problems.

markus2330:
Seems like you merged master into this branch?

Merging now would work. If rebasing can be easily done I would prefer rebasing nevertheless.

@e1528532 Simple readd "ready to merge" if done.

sanssecours:
> Seems like you merged master into this branch?

Yes, I assume commit cc0afe6 is the result of the command `git merge master`.

> If rebasing can be easily done I would prefer rebasing nevertheless.

If you want, I can just push a rebased version of this pull request.

src/tools/kdb/ls.cpp
```diff
@@ -45,13 +42,71 @@ int LsCommand::execute (Cmdline const & cl)
...
+	{
+		throw invalid_argument ("the minimum depth has to be a positive number");
+	}
+}
+
+void LsCommand::printResults (KeySet const & part, const int rootDepth, Cmdline const & cl)
+{
+	const int offset = root.getBaseName ().empty () || shallShowNextLevel (cl.arguments[0]) ? 1 : 0;
+	const int relativeMinDepth = rootDepth + cl.minDepth + offset;
+	const int relativeMaxDepth = std::max (cl.maxDepth, rootDepth + cl.maxDepth + offset);
```
sanssecours: Looks like this line may cause an integer overflow:

```
../src/tools/kdb/ls.cpp:76:64: runtime error: signed integer overflow: 1 + 2147483647 cannot be represented in type 'int'
SUMMARY: AddressSanitizer: undefined-behavior ../src/tools/kdb/ls.cpp:76:64
```

.

PR #1347, opened 2017-02-16
Fix kdb set base name

e1528532:
# Purpose

Fix the bug that setting the basename of a cascading key like /cascading to NULL will also delete the namespace, so the / that marks the key as cascading.

# Checklist

- [X] I added unit tests

@markus2330 IMO this is a special case that only happens when deleting a cascading key with depth 1 (so something like /cascading) as the setBaseName method simply strips everything up to including the last / character.
For everything with a namespace this is fine (e.g. user/key gets stripped to user), but for a cascading namespace, the last / has to be preserved. This only seems to be relevant upon basename deletion, therefore i've put the appropriate adjustment in the if branch. Also added unit tests for such cases.

### DIFF
--- a/src/libs/elektra/keyname.c
+++ b/src/libs/elektra/keyname.c
@@ -1081,7 +1081,7 @@ ssize_t keySetBaseName (Key * key, const char * baseName)
 	if (!baseName)
 	{
 		// Avoid deleting the last / of a cascading key by increasing the size by one again
-		key->keySize += 1 == key->keySize && KEY_NS_CASCADING == keyGetNamespace (key);
+		key->keySize += (1 == key->keySize) && (KEY_NS_CASCADING == keyGetNamespace (key));
 
 		// just remove base name, so we are finished
 		elektraFinalizeName (key);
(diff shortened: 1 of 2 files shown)


markus2330:
Very nice fix!

Thank you, yes seems like we need a new case here. Adding parentheses to the expression would not harm, though.

I am missing at least three test cases (it is modification in the core):

- removing /foo/bar down to nothing (multiple removals with the same key)
- removing /X with single character X
- removing /\Y with Y the characters that need to be escaped like /.

markus2330:
jenkins build please

markus2330:
jenkins build stable please

markus2330:
jenkins build unstable please

e1528532:
You are right, parantheses make it even more clear whats going on. Furthermore i added some more test cases as requested. Is it enough to have one test case for escaped characters or should i make extra tests for the other characters too (afaik . .. %, maybe we should try an array too like #arr?). IMO more test cases can't hurt, especially in that case, but it could also be that you want to clutter the testcode up.

markus2330:
> should i make extra tests for the other characters too

Yes, that would be nice!

> but it could also be that you want to clutter the testcode up.

Dont worry, if the function gets too long you can put these tests to a separate one. As long as the tests are not exact duplicates it is better to have more than less, they execute in no time anyway.

e1528532:
here they are

markus2330:
jenkins build please

markus2330:
Thank you! travis seems to have some problems unrelated to this PR.

PR #1352, opened 2017-02-17
fix-bashisms

### LINKED ISSUE #1256: fix bashism
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=772254

add it to build job where scripts are installed?

e1528532:
# Purpose

Fixing some possible bashism cases in our scripts to resolve #1256 

# Checklist

- [ ] I added unit tests

@markus2330 here are my first changes. Some things may need a bit more discussion, as you'll see in the issue. I think most of the fixes are so simple that they most likely won't break anything or make things worse on some systems,


scripts/configure-common
```diff
@@ -1,3 +1,5 @@
+#!/bin/sh
```
markus2330: The file is only sourced, so this is not really necessary.

markus2330:
Seems everything okay for me, where is discussion needed? Maybe let us first do the easy things, and create a second PR for the rest.

e1528532:
Now i've added the remaining things as discussed in #1256 . Should be good to merge i think.

```
mbp-armin% checkbashisms -px scripts/*
script scripts/CMakeLists.txt does not appear to have a #! interpreter line;
you may get strange results
script scripts/README.md does not appear to have a #! interpreter line;
you may get strange results
script scripts/filter-coverage.awk.in does not appear to be a /bin/sh script; skipping
script scripts/find-tools does not appear to be a /bin/sh script; skipping
script scripts/kdb-zsh-noglob does not appear to be a /bin/sh script; skipping
script scripts/kdb.fish does not appear to be a /bin/sh script; skipping
script scripts/kdb_zsh_completion does not appear to be a /bin/sh script; skipping
possible bashism in scripts/realpath line 38 (local foo):
    local dir_context
possible bashism in scripts/realpath line 39 (local foo):
    local path
possible bashism in scripts/realpath line 65 (local foo):
    local target
possible bashism in scripts/realpath line 66 (local foo):
    local path
possible bashism in scripts/realpath line 91 (local foo):
    local dir
possible bashism in scripts/realpath line 92 (local foo):
    local file
script scripts/update-infos-status does not appear to be a /bin/sh script; skipping
script scripts/zsh does not appear to be a /bin/sh script; skipping
```

As you can see, no more relevant warnings except for the local variable thing, which seems to be no problem when having only 1 variable definition per local keyword as seen in https://wiki.ubuntu.com/DashAsBinSh#local. If we execute it without -px, it doesn't complain about it.

scripts/configure-common
e1528532: yes but in #1256 we agreed to add it because i think it doesn't hurt and supresses the checkbashisms warning.

markus2330:
Thank you great job!

Can we also integrate the `checkbashism` easily into the regression tests so that @pinotree does not see any further regressions of bashisms in the future?

I added the task to #160. If it returns a proper exit code something trivial like `tests/shell/check_formatting.sh` would already do.

Issue #1353, opened 2017-02-17
JNA bindings test fails on Build Server


e1528532:
Looks like the JNA bindings test fails on the build server because the Elektra library cannot be loaded. I guess the build server has no actual elektra installation, so it doesn't find the dynamic library. I have to check if there is a possibility to point it to the right path in such environment. I guess its better to point it to load dynamic libraries from the build directory anyway and not rely on a system installation. If required we can exclude the test for now.

markus2330:
Please let the test pass if no Elektra installation was found. (You can do this even within Java I suppose?) Other external tests deactivate themselves, too.

markus2330:
Thank you for putting the reference here. Please always try to add a link to the build server or ideally even copy&past the error. When talking about the build server, we usually mean https://build.libelektra.org/, and not travis.

e1528532:
So I've tried to make the tests work on macOS when no elektra is installed, but without success yet (see #1360 ). Tried updating jna version, setting some flags for dlopen in jna, setting LD_LIBRARY_PATH/DYLD_LIBRARY_PATH both in the shell and in the pom file for surefire (the maven test plugin), and a bunch of other things i've already forgotten about.

I guess the issue boils down due to macOS 10.11 and onwards not allowing to load shared libraries from other locations than the default one for restricted applications.

>  Spawning children processes of processes restricted by System Integrity Protection, such as by launching a helper process in a bundle with NSTask or calling the exec(2) command, resets the Mach special ports of that child process. Any dynamic linker (dyld) environment variables, such as DYLD_LIBRARY_PATH, are purged when launching protected processes.

The problem is that java is usually marked as restricted (as a part of macOS's System Integrity Protection), and maven is java based, so this seems to apply to maven too. Therefore it doesn't pass the adjusted dynamic library path for the test, thus it fails. And thats also the reason why it works on my arch where i tried it too, and on the old macOS 10.10 on travis.

You could now argue why it loads libelektra.dylib successfully. To my understanding, thats because it is loaded by JNA where you can customize the location with the jna.library.path argument. But then when using KDB, it fails, as using KDB seems to imply elektra loading the libelektra-resolver.so shared library. But this happens outside of java/jna, so it looks it up in the DYLD_LIBRARY_PATH or uses the RPATH mechanism (not totally sure about that yet), and as this does only search in the system locations, it fails unless elektra is installed.

[...]

PR #1360, opened 2017-02-20
jna-maven-fix-tests

e1528532:
# Purpose

Fix the failing jna bindings test as the native library is not yet installed to the system, so use the one from the build environment for our tests

# Checklist

- [ ] I added unit tests

### DIFF
--- a/src/bindings/jna/CMakeLists.txt
+++ b/src/bindings/jna/CMakeLists.txt
@@ -43,7 +43,7 @@ if (MAVEN_EXECUTABLE) # set by find_program
 	# and add maven test to execute with ctest in the testing phase
 	add_test (
 		NAME testjna_maven
-		COMMAND ${MAVEN_EXECUTABLE} -q -B -P env-test test
+		COMMAND ${MAVEN_EXECUTABLE} -q -B -e -P env-test test
 		WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/libelektra4j"
 	)
 else (MAVEN_EXECUTABLE)
--- a/src/bindings/jna/CMakeLists.txt
+++ b/src/bindings/jna/CMakeLists.txt
@@ -43,7 +43,7 @@ if (MAVEN_EXECUTABLE) # set by find_program
 	# and add maven test to execute with ctest in the testing phase
 	add_test (
 		NAME testjna_maven
-		COMMAND ${MAVEN_EXECUTABLE} -B test
+		COMMAND ${MAVEN_EXECUTABLE} -q -B -P env-test test
 		WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/libelektra4j"
 	)
 else (MAVEN_EXECUTABLE)
--- a/src/bindings/jna/CMakeLists.txt
+++ b/src/bindings/jna/CMakeLists.txt
@@ -41,10 +41,29 @@ if (MAVEN_EXECUTABLE) # set by find_program
 	)
 
 	# and add maven test to execute with ctest in the testing phase
+	if ( (CMAKE_SYSTEM_NAME STREQUAL "Darwin") AND NOT (CMAKE_SYSTEM_VERSION VERSION_LESS 15) )
+		# we cannot set DYLD_LIBRARY_PATH on new macOS versions, making the kdb tests
+		# fail if its not installed in the system
+		set (MAVEN_TEST_ENV "env-test-without-kdb")
+	else ()
+		set (MAVEN_TEST_ENV "env-test")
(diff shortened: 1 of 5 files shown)


markus2330:
jenkins build please

markus2330:
jenkins build stable please

markus2330:
jenkins build unstable please

markus2330:
jenkins build log please

markus2330:
Please always start the build server jobs when creating PRs. It seems like the build server does not start automatically anymore.

e1528532:
Ok it finds the native library now. The remaining issue is that it looks like the kdb tests fail if Elektra is not actually installed (so those that write to kdb and expect the keys to be still present when reopening kdb). Is this supposed to work on travis (persisting stuff in kdb)?

markus2330:
Why shouldn't it work to persist keys? Many other tests do so, too.

Is it maybe a concurrency issue? One of the 3 builds succeeded: https://travis-ci.org/ElektraInitiative/libelektra/jobs/203455791

Btw. the errors are really vague, can you add more infos in the exceptions? (Everything you get from KDB)

e1528532:
Yes the errors are indeed really vague, the bindings needed something like the printWarnings from c so i ported it over. Now we have better information:

```
Caused by: java.lang.Throwable:
failure in I/O to KDB: 3 Warnings were issued:
 Warning number: 1
        Description: could not load module, dlopen failed
        Ingroup: modules
        Module: dl
        At: /Users/admin/git/libelektra/src/libs/loader/dl.c:88
        Reason: of module: libelektra-resolver.so, because: dlopen(libelektra-resolver.so, 2): image not found
        Mountpoint:
        Configfile:
 Warning number: 1
        Description: could not load module, dlopen failed
        Ingroup: modules
        Module: dl
        At: /Users/admin/git/libelektra/src/libs/loader/dl.c:88
        Reason: of module: libelektra-resolver.so, because: dlopen(libelektra-resolver.so, 2): image not found
        Mountpoint:
        Configfile:
 Warning number: 40
        Description: failed to open default backend (see warnings for more information)
        Ingroup: kdb
        Module:
        At: /Users/admin/git/libelektra/src/libs/elektra/kdb.c:287
        Reason: could not open default backend
        Mountpoint:
        Configfile:

        at org.libelektra.KDBTest.removeCreatedKeys(KDBTest.java:57)
```

So it looks like the issue happens outside of java? Java loads the libelektra.dylib from the build/lib folder now (thats what my previous commit fixed). I think this dynamic library tries to locate the resolver then, which seems to fail. Shouldn't it detect it given that the libelektra-resolver.so sits in the same folder?

markus2330:
Doesn't a "failure in I/O" imply that there also have been an error (and not only warnings?)

> Shouldn't it detect it given that the libelektra-resolver.so sits in the same folder?

It might using RPATH but this seems to be a fragile thing if multiple languages are involved. Try to set `LD_LIBRARY_PATH`. Lua and python test cases do so already.

Btw. seems like someone wants to work with the java binding to improve the website.
Very good timing, thus the java binding just got completed :-)

Anyway, I would ask you to not do the website tasks. There would be too much effort involved for you to learn the technology stack of the website and your time is better invested in solving other issues.

e1528532:
Well thats how the error was called before so i left it that way. As far as i've seen it just throws this error message if some kdb method goes wrong, regardless of the actual reason. So I've added the warnings in order to see the actual reasons.  To my understanding, printWarnings already includes the "errors" too right? Or is there another metadata for it? In c++ i've only seen printWarnings so far.

But i can rename it to something more general like Failure when communicating with KDB: Warnings: <...>.

To the RPATH issue: I've tried it on my arch today and there it works without any further hassle, every time i tried it. The one build that succeeded on travis seems to be on an older version of macOS, specifically 10.10 `instance: aadc8bfa-b7c8-4c0c-a664-d2f8059138be:travis-ci-osx10.10-xcode6.4-1479248236`. Since 10.11 its no longer possible to pass LD_LIBRARY_PATH (or DYLD_LIBRARY_PATH as its called in mac) to child processes. But more about that in #1353 .

I think this PR can be merged after we have decided about the error message. It already contains some improvements (warning port and correct jna library path, which makes it work at least on linux and older macs).

markus2330:
> To my understanding, printWarnings already includes the "errors" too right?

No, there also should be a `printError`.

> I think this PR can be merged after we have decided about the error message.

Yes, please add the error message and disable the tests on broken MacOSX versions.

e1528532:
look at that beautiful green checkmark travis will produce :)

e1528532:
jenkins build please

markus2330:
Thank you!

Issue #1228, opened 2016-12-23
website: Next Steps for Snippet Sharing


markus2330:
From feedback:

- [ ] improve performance!
- [ ] make use case more clear!
- [ ] make formats more clear (what to do on any file)
- [ ] Simplify login
- [ ] Improve ways to give feedback

Here are some ideas how we can improve snippet sharing:

- [ ] upload more snippets!
- [ ] better description when to use `line`
- [ ] github auth/register
- [ ] markdown for description field
- [ ] allow curl put (curlget plugin) to upload snippets
- [ ] command-line tool to upload/download snippets
- [ ] introduce rating system
- [ ] commenting on snippets
- [ ] allow to upload snippets from the snippet converter
- [ ] upload format as column in search and "Raw" uses upload format
- [ ] combine multiple plugins, e.g. validation, transformations (see #1229)
- [ ] allow code to be inserted for transformation plugins
- [ ] "rename" snippets key (e.g. when changing organization, frontend creates a new snippet and removes the old one)

### Frontend Fixes

- [ ] github link if js does not work
- [ ] when creating snippets: clicking on "?" does not do anything (for hover you need patience, please show it immediately)
- [ ] support for compilation variants, see [HTML version of the crypto documentation](https://www.libelektra.org/plugins/crypto) (ifdefs should be removed)

### Setup Improvements

- [ ] make sure backend does not go down
- [ ] enable core dumps
- [ ] backup into private repo
- [ ] install apache log analyzer
- [ ] user of rest-backend should have minimal rights (not jenkins)

### Backend

- [ ] automated API Testing #931

###

Which of them do you find useful?

e1528532:
I encountered two issues:
* You cannot change organization after you created a snippet and want to edit it because you forgot to specify the organization
* XML plugin needs improvements, wasn't able to recognize my maven configuration snippets

markus2330:
Thanks for your feedback! I think the changing of organization would require some "rename" functionality, I added it in the list above.

The XML plugin would be a new completely new plugin. It might be not too hard to write, though (when using a XML parser with a nice API, e.g. look at `ni` plugin). xmltool is very old code, I can not recommend to base a new plugin on that code. Please open a new issue if you see an easy (~3h) way to do it.

@e1528532 I also added the other "better description when to use `line`" you mentioned at the meeting. From the docu it should be clear that `line` is the way to go if no other plugin fits. The file plugin #1047 would be a good alternative to line,though.

[...]

PR #1371, opened 2017-02-23
crypto: Display available GPG keys if config option is missing.

### LINKED ISSUE #895: crypto plugin: list available keys in case of missing config
# Situation / Problem

If no gpg key is specified at `/gpg/key` at mount time, the following error message is displayed:

```
Error (#130) occurred!
Description: The configuration is invalid or incomplete.
Ingroup: plugin
Module: crypto
At: /root/libelektra/src/plugins/crypto/gpg.c:219
Reason: Missing GPG key (specified as /gpg/key) in plugin configuration.
Mountpoint: 
Configfile:
```
# Expected Behaviour (Proposal)

Additionally a list of available gpg keys should be presented to the user.

petermax2:
# Purpose

If /gpg/key is not provided in the plugin configuration a list of available GPG key IDs is shown in the error description.

Fixes #895 .

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/src/plugins/crypto/README.md
+++ b/src/plugins/crypto/README.md
@@ -150,7 +150,40 @@ If more than one key is defined, every owner of the corresponding private key ca
 This might be useful if applications run with their own user but the administrator has to update the configuration.
 The administrator then only needs the public key of the application user in her keyring, set the values and the application will be able to decrypt the values.
 
+If you are not sure which keys are available to you, the `kdb` program will give you suggestions in the error description.
+For example you can type:
+
+    kdb mount test.ecf user/t crypto_gcrypt
+
+In the error description you should see something like:
+
+    The command ./bin/kdb mount terminated unsuccessfully with the info:
+    The provided plugin configuration is not valid!
+    Errors/Warnings during the check were:
+    Sorry, the error (#130) occurred!
+    Description: the configuration is invalid or incomplete.
+    Ingroup: plugin
+    Module: crypto
+    At: /Users/pnirschl/Projects/libelektra/src/plugins/crypto/gpg.c:512
+    Reason: Missing GPG key (specified as /gpg/key) in plugin configuration. Available key IDs are: B815F1334CF4F830187A784256CFA3A5C54DF8E4,847378ABCF0A552B48082A80C52E8E92F785163F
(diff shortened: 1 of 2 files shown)


markus2330:
Good to see activity from you again!

markus2330:
Thank you, can you add some documentation of the new feature? (See also other requests about docu)

@sanssecours do you have some time to give this a try?

markus2330:
jenkins build please

sanssecours:
> @sanssecours do you have some time to give this a try?

I tried it and everything worked as described. I just saved my very first secret key 🙌. The change should improve usability of the `crypto` plugin for novice users like me quite considerably.

I noticed a **small bug** while I read the well-written and comprehensive `ReadMe` of the crypto plugin. It looks like the [HTML version of the documentation](https://www.libelektra.org/plugins/crypto) still contains `#ifdef` directives.

markus2330:
> small bug

Its done correctly by @petermax2 (even though some more newlines around the `#ifdefs` would make the Markdown a bit more pretty). If you use `kdb info` the ifdefs should be gone.

Fixing it for the homepage is not so trivial, as it would need to understand the compilation variants. I added it to #1228.

sanssecours:
> Fixing it for the homepage is not so trivial, as it would need to understand the compilation variants.

I just used the homepage as an example. I think this is a problem of the Doxygen documentation, rather than a problem of the website. Both, the HTML version of the documentation, and the PDF manual, contain `#ifdef` directives. 

> If you use `kdb info` the ifdefs should be gone.

That is correct. Thanks you for this useful information. For now I use the command

```sh
kdb info crypto_gcrypt > /tmp/help.md | open -a Marked /tmp/help.md
```

to read the documentation of the `crypto` plugin.

markus2330:
> this is a problem of the Doxygen documentation, rather than a problem of the website. Both, the HTML version of the documentation, and the PDF manual, contain #ifdef directives.

Doxygen would be relatively easy to fix, we could run the preprocessor as filter. Nevertheless, I think its a minor issue thus we do not have many plugins that use compilation variants.

GitHub itself obviously always will show the `ifdefs`.

PR #1334, opened 2017-02-09
range plugin

tom-wa:
# Purpose

range plugin preview

# Checklist

- [x] commit messages are fine (with references to issues)
- [x] I ran all tests and everything went fine
- [x] I added unit tests
- [x] affected documentation is fixed
- [x] I added code comments, logging, and assertions
- [x] meta data is updated (e.g. README.md of plugins)

PR #1373, opened 2017-02-24
typedispatcher

tom-wa:
# Type system preview only

currently code is a mess - will be refactored and cleaned later 

# Purpose

demo: scoping, sub types and recursive (depth-first) type hierarchy traversing, subtype checking

depends on #1321, #1334 (and some ugly hack in `types`)

@markus2330 

- any preferences on how to handle cyclic type dependencies ?
- would you be ok with having a max-depth for type hierarchy levels for now ?
- mapping metadata to plugins: i guess the easiest way would be parsing `METADATA.ini` during the build process and create a static struct. any other suggestions ?
- i'd say abbreviations should be handled by the validation plugins like 
    ```
          int validateSimple(key, configString, errorKey)
           
           type = enum('a', 'b', 'c')
               => enum::validateSimple(key, "'a', 'b', 'c'", errorKey)
           type = date("ISO|datetime complete")
               => date::validateSimple(key, "ISO|datetime complete", errorKey)
    ```
     what do you think ?

# Checklist

- [ ] I added unit tests


markus2330:
Please write a README.md so that we know what you tried to achieve. Please start with very simple examples. Please remember that the log output makes only sense to you.

Why not use Elektra's logger, are there any issues with it?

> any preferences on how to handle cyclic type dependencies ?

Please abort with an error unless we find a real use case where we need it.

> would you be ok with having a max-depth for type hierarchy levels for now ?

It is a bit weird, especially because you use dynamic memory allocations anyway.

> mapping metadata to plugins: i guess the easiest way would be parsing METADATA.ini
> during the build process and create a static struct. any other suggestions ?

I created the issue #1374 for discussions.

> i'd say abbreviations should be handled by the validation plugins like 

What are abbreviations?

tom-wa:
>Why not use Elektra's logger, are there any issues with it?

vim folding for #if , i initially had addition code within them. will look into to logger again

>Please abort with an error unless we find a real use case where we need it.

i was more worried about the costs of checking for cycles, thats why i also asked about a max-depth

>It is a bit weird, especially because you use dynamic memory allocations anyway.

the type traversal is recursive, thats why i thought setting a max depth might be a good idea 

>What are abbreviations?

the examples below, using `type = long` or `type = enum (...` instead of adding multiple `check/X metakeys`




should i also implement a `undef` for types ?

markus2330:
> the examples below

The example was the part that was unclear. What does it show? A function declaration and output from printf?

> should i also implement a undef for types ?

I do not see a necessity for it. Better keep the features as is (there are enough already?) and let us add something if it is really needed.

tom-wa:
@markus2330 
the typedispatcher has 2 jobs: read and build a "type tree" with all defined types, and validate the keys, but i'm currently stuck at the decision when the "type tree" should be build and updated. 
there are several options: 
- only read it once and the beginning of kdbGet and never update it, 
- read it every time the plugin is called and overwrite the whole tree, 
- or read it once at kdbGet and if a key has a sync flag check for type data and update the type tree if needed

the first 2 options would be the easiest to implement, but the last probably the most useful ?

markus2330:
Sorry, I did not see this earlier. As discussed: always discard and rebuild.

tom-wa:
@markus2330 i think we have to split usedby/plugin into checkby and usedby in METADATA.ini.
`check/type` is currently used by 2 plugins: `type` and `range` but i can't distinguish if `type` or `range` should validate `check/type`
+ i need a way to figure out which plugin needs which metadata

markus2330:
> i think we have to split usedby/plugin into checkby and usedby in METADATA.ini.

Yes, good idea!

tom-wa:
using `check/type` instead of `check/range/type` currently wont work because typedispatcher only checks keys below `check/range` to create it's internal datatypes so i changed it back until i find a proper solution.

markus2330:
Hi! Is it nevertheless ready for review? @e1528532 Can you take a look?

tom-wa:
yes. it depends on #1321

markus2330:
Okay, I merged #1321.

@e1528532 do you have time to take a look?

markus2330:
Did you use a different version than 3.8 of llvm for reformat?

tom-wa:
>Did you use a different version than 3.8 of llvm for reformat?

i formated testmod_typedispatcher by hand because with clang-format it becomes unreadable.
i'll reformat it before it gets merged

markus2330:
@e1528532 could you please take a look at this PR?

markus2330:
@tom-wa Could you please document how the scoping (shadowing of `define/type`) works? Then @domhof could implement it in the same way in the code generator.

e1528532:
@tom-wa i have a small issue with the test cases. Judging from the log the typedispatcher seems to do its thing, but then ultimately it tells me it can't load the range or enum plugin. For instance the output for test_simple:

```
60: Getting type definitions from user/tests/typedispatcher
60:      Key define/type/ace is 1 levels below define/type
60:      Key define/type/ace/check/range is 3 levels below define/type
60:      Key define/type/bar is 1 levels below define/type
60:      Key define/type/bar/check/range is 3 levels below define/type
60:      Key define/type/foo is 1 levels below define/type
60:      Key define/type/foo/check/range is 3 levels below define/type
60:      Key define/type/ace is 1 levels below define/type
60:      Key define/type/bar is 1 levels below define/type
60:      Key define/type/foo is 1 levels below define/type
60:     defines type ace
60:     defines type bar
60:     defines type foo
60: populating ace
60: datatype ace found in config->types
60:      Key define/type/ace/check/range is 1 levels below define/type/ace/check
60:     define/type/ace/check/range:(0-4,6-9)
60:             adding Metadata check/range:(0-4,6-9) to key range
60: populating bar
60: datatype bar found in config->types
60:      Key define/type/bar/check/range is 1 levels below define/type/bar/check
60:     define/type/bar/check/range:(3-5,8-20)
60:             adding Metadata check/range:(3-5,8-20) to key range
60: populating foo
60: datatype foo found in config->types
60:      Key define/type/foo/check/range is 1 levels below define/type/foo/check
60:     define/type/foo/check/range:(4-20)
60:             adding Metadata check/range:(4-20) to key range
60: Getting type definitions from user/tests/typedispatcher/all
60: validating user/tests/typedispatcher/all:(4)
60:      Key type/#0 is 1 levels below type
60:      Key type/#1 is 1 levels below type
60:      Key type/#2 is 1 levels below type
60:     with type ace
60:     parseTypeString ace returned 0x0, typeName: ace
60:             user/tests/typedispatcher/all has SUBTYPE
60:     range-validation
60: replacing parameters in 0-4,6-9
60: result: 0-4,6-9
60:             check/range:(0-4,6-9)
60: plugin providing range not loaded
60: ERROR: failed to load plugin providing range
60:                     checkKey(config, user/tests/typedispatcher/all:(4), range) returned -1
60: Key user/tests/typedispatcher/all failed to validate
```

The range and the enum plugin are both compiled though i think, and running the enum and the range test both succeed. Anything else to keep in mind?

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+						strncat (preparedString, keyString (lookup), elektraStrLen (keyString (lookup)) - 1);
+						dstPtr = preparedString + newLen - 1;
+						preparedStringLen = newLen;
+					}
+					else
+					{
+						ssize_t newLen = elektraStrLen (preparedString) + sizeof (tmp) + 2;
+						elektraRealloc ((void **)&preparedString, newLen);
+						strcat (preparedString, "%");
+						strncat (preparedString, tmp, sizeof (tmp));
```
e1528532: The compiler gives me this warning. shouldn't be strlen be used instead of sizeof here?
helpers.c:274:37: warning: size argument in 'strncat'
      call appears to be size of the source [-Wstrncat-size]
                                                strncat (preparedString, tmp, sizeof (tmp));

src/plugins/typedispatcher/typereader.c
```diff
@@ -0,0 +1,422 @@
+#include "typehelper.h"
```
e1528532: file header

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+		{
+			const char * endPtr = ptr + 1;
+			if (lastPtr == ptr) ++lastPtr;
+			while (*endPtr)
+			{
+				if (*endPtr == '%')
+				{
+					// copy string from lastPtr to ptr
+					preparedStringLen += (ptr - lastPtr) + 1;
+					elektraRealloc ((void **)&preparedString, preparedStringLen);
```
e1528532: i think we should always include a failure handling for realloc (if (elektraRealloc(..) < 0) ...)

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+					memcpy (tmp, ptr + 1, len);
+					Key * searchKey = keyNew (tmp, KEY_META_NAME, KEY_END);
+					Key * lookup = ksLookup (args, searchKey, KDB_O_NONE);
+					keyDel (searchKey);
+					if (lookup)
+					{
+
+						ssize_t newLen = preparedStringLen + elektraStrLen (keyString (lookup)) -
+								 1; // no need for 2 nullbytes
+						elektraRealloc ((void **)&preparedString, newLen);
```
e1528532: failure handling for elektraRealloc

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+						strncat (preparedString, keyString (lookup), elektraStrLen (keyString (lookup)) - 1);
+						dstPtr = preparedString + newLen - 1;
+						preparedStringLen = newLen;
+					}
+					else
+					{
+						ssize_t newLen = elektraStrLen (preparedString) + sizeof (tmp) + 2;
+						elektraRealloc ((void **)&preparedString, newLen);
```
e1528532: failure handling for elektraRealloc

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+				}
+				++endPtr;
+			}
+		}
+		++ptr;
+	}
+	if (!(*ptr) && lastPtr && preparedString)
+	{
+		ssize_t newLen = elektraStrLen (preparedString) + (ptr - lastPtr) + 1;
+		elektraRealloc ((void **)&preparedString, newLen);
```
e1528532: failure handling for elektraRealloc

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+	ssize_t typeNameLen = (ptr - typeString) + 1;
+	if (typeNameLen <= 0) return NULL;
+
+	while (*ptr && (*ptr != '('))
+		++ptr;
+	if (!*ptr) return NULL;
+
+	ArgumentConfig * argConfig = NULL;
+	argConfig = elektraCalloc (sizeof (ArgumentConfig));
+	argConfig->type = elektraCalloc (typeNameLen);
```
e1528532: i think the failure handling for two callocs here is missing

src/plugins/typedispatcher/pluginhelper.c
```diff
@@ -0,0 +1,123 @@
...
+		fprintf (stderr, "mountpoint: : %s\n", keyString (keyGetMeta (errorKey, "error/mountpoint")));
+		fprintf (stderr, "configfile: : %s\n", keyString (keyGetMeta (errorKey, "error/configfile")));
+#endif
+		keyDel (errorKey);
+		return NULL;
+	}
+	keyDel (errorKey);
+
+	PluginConfig * pc = NULL;
+	pc = elektraCalloc (sizeof (PluginConfig));
```
e1528532: i think the failure handling for calloc is missing

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+#include <kdbmodule.h>   //elektraModulesInit, elektraModulesClose
+#include <kdbos.h>       //elektraNamespace;
+#include <kdbprivate.h>  //elektraPluginClose. elektraPluginOpen, elektraStrNDup
+#include <kdbproposal.h> //keyRel2, KeyRelType
+#include <stdio.h>
+#include <string.h>
+#include <strings.h>
+
+
+// hekoer fir freeTyoes
```
e1528532: looks like you have a little typo there

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
+#include "typehelper.h"
+#include <ctype.h>
+#include <kdbease.h>     //elektraArrayIncName
+#include <kdbmodule.h>   //elektraModulesInit, elektraModulesClose
+#include <kdbos.h>       //elektraNamespace;
+#include <kdbprivate.h>  //elektraPluginClose. elektraPluginOpen, elektraStrNDup
```
e1528532: ok thats really a minor thing but maybe consistent about using , ; and . or nothing at all, otherwise one could maybe believe that there is some meaning behind it

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+{
+	const Key * scope = config->scope;
+	if (keyCmp (scope, key))
+	{
+		if (keyRel2 (scope, key, ELEKTRA_REL_BELOW_SAME_NS) <= 0)
+		{
+			elektraNamespace scopeNS = keyGetNamespace (scope);
+			if (scopeNS == KEY_NS_SPEC)
+			{
+				if (keyRel2 (scope, key, ELEKTRA_REL_BELOW_IGNORE_NS) <= 0)
```
e1528532: in other parts of the code you don't use braces for single line return statements though cause of the nested structure here it might be appropriate. Maybe even simplify this as
return keyRel2 (scope, key, ELEKTRA_REL_BELOW_IGNORE_NS) > 0; ?

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+
+TypeType getTypeType (TypeConfig * tc)
+{
+	return tc->type;
+}
+
+Key * getTypeKey (DispatchConfig * config, const char * type)
+{
+	Key * lookup = ksLookupByName (config->types, type, KDB_O_NONE);
+	return lookup;
```
e1528532: i prefer directly returning the method's result in such cases

src/plugins/typedispatcher/helpers.c
```diff
@@ -0,0 +1,437 @@
...
+
+
+// parses "type (arg1, arg2, ...)" strings into ArgumentConfig
+// TODO: rewrite
+ArgumentConfig * parseTypeString (DispatchConfig * config ELEKTRA_UNUSED, const char * typeString)
+{
+	const char * ptr = typeString;
+	while (*ptr)
+	{
+		if (*ptr == '(' || *ptr == ' ') break;
```
e1528532: directly below in line 326 you have such check in your loop condition, which i also prefer.

tom-wa:
@e1528532 
>it tells me it can't load the range or enum plugin

is it the same for the `type`-plugin ?

src/plugins/typedispatcher/helpers.c
tom-wa: what compiler are you using ? neither gcc 6.3 locally nor the buildserver are showing any warnings. 
it's valid but by far not the best solution

src/plugins/typedispatcher/typereader.c
```diff
@@ -0,0 +1,422 @@
...
+#include <stdio.h>
+
+
+// creates skeletons(name+type+scope) for new types and fails if type already exist
+static int readTypeNames (DispatchConfig * config, const Key * scope, const Key * typeKey, Key * parentKey ELEKTRA_UNUSED)
+{
+	if (!typeKey) return ERROR;
+
+	const char * typeName = keyBaseName (typeKey);
+#ifdef DEVBUILD
```
tom-wa: i'll drop both of them when i'm done, they are just for internal debugging (+ vim hides/folds them nicely)

tom-wa:
is it necessary to set an error if we run out of memory or would returning with -1 or calling exit be enough ? i mean, it realloc/calloc actually fails there would be a high chance that setting the error message will fail too

markus2330:
> is it necessary to set an error if we run out of memory or would returning with -1 or calling exit be enough ? i mean, it realloc/calloc actually fails there would be a high chance that setting the error message will fail too

`return -1` is okay for now, it is done at many places. If you have the parentKey, please set the error, it will save time if we support systems where malloc fails under normal circumstances. I like @sanssecours suggestion to have a macro explicitly for this case, then we need to implement the special case for "we need an error message without allocating anything" only there. But its not a goal now, testing out-of-memory situations is a beyond 1.0 goal.

markus2330:

PR #1380, opened 2017-02-24
xerces-plugin

e1528532:
# Purpose

This storage plugin serializes and deserializes to XML.

# Checklist

- [X] I added unit tests


markus2330:
jenkins build please

markus2330:
jenkins build unstable please

src/plugins/xerces/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+include (LibAddMacros)
+
+if (DEPENDENCY_PHASE)
+	find_package(XercesC)
+
+	if (NOT XercesC_FOUND)
+		remove_plugin (xerces "XercesC library not found")
```
markus2330: do we need a specific version of it?

src/plugins/xerces/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+	CPP
+	SOURCES
+		xerces.hpp xerces.cpp
+		deserializer.hpp deserializer.cpp
+	INCLUDE_DIRECTORIES
+		${XercesC_INCLUDE_DIRS}
+	LINK_LIBRARIES
+		${XercesC_LIBRARIES}
+	ADD_TEST
+)
```
markus2330: please add new lines at end

src/plugins/xerces/deserializer.cpp
```diff
@@ -0,0 +1,141 @@
+/**
+ * @file
+ *
+ * @brief deseialization implementation for xerces plugin
```
markus2330: typo

src/plugins/xerces/deserializer.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+XERCES_CPP_NAMESPACE_USE
+using namespace std;
+using namespace kdb;
+
+/*
+ + Smart pointer helper functions to ease the pain of dealing between XMLCh and std::string
+ */
+
+// no need to expand this, just keep it in one line
```
markus2330: Why not? Does the formatter fail in doing so?

src/plugins/xerces/deserializer.cpp
```diff
@@ -0,0 +1,141 @@
...
+// TODO move to external file if also used for serializer
+// clang-format off
+struct XmlChDeleter { void operator() (XMLCh * ptr) { XMLString::release (&ptr); } };
```
markus2330: Please have a look at the RAII pattern, it looks like you overuse smart pointers.

src/plugins/xerces/deserializer.cpp
```diff
@@ -0,0 +1,141 @@
...
+	return unique_ptr<std::string, StringDeleter> (new string (cStr), StringDeleter{ cStr });
+}
+
+/*
+ * Actual Xerces logic
+ */
+
+static unique_ptr<XercesDOMParser> doc2dom (const std::string src)
+{
+	unique_ptr<XercesDOMParser> parser (new XercesDOMParser);
```
markus2330: The C++ style of their docu seems to be quite outdated. Your style using unique_ptr is definitely better. But why not simply allocate XercesDOMParser as value on the stack, for example:

```
XercesDOMParser parser;
parser.setValidationScheme (XercesDOMParser::Val_Auto);
...
```

src/plugins/xerces/deserializer.hpp
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+#ifndef ELEKTRA_PLUGIN_XERCES_DESERIALIZER_H
+#define ELEKTRA_PLUGIN_XERCES_DESERIALIZER_H
+
+#include <keyset.hpp>
+
+void deserialize (const std::string src, kdb::KeySet & ks);
+
+#endif
```
markus2330: No newline.

src/plugins/xerces/deserializer.cpp
e1528532: Thats not possible as the copy constructor of XercesDOMParser is private, looks like its not intended to be done that way.

src/plugins/xerces/deserializer.cpp
e1528532: no, but i thought its unecessary to expand this simple custom deleter to several lines

src/plugins/xerces/deserializer.cpp
e1528532: i think its fine that way, as xerces uses this XMLCh type for strings internally, we always have to transform them to either c or c++ strings using the XMLString::transform function. And the result of the transform function has to be released afterwards with XMLString::release. So IMO the easiest way is to use a smart pointer, which deals with calling release so i can't forget it and i don't have to do it everywhere i use a string.

e1528532:
Basic serialization, deserialization (including value & metadata) and error handling should be implemented now.
Is the current xml structure good? the root element is always called <namespace> now, which contains the namespace elements like <user> or <cascading>, and then the rest of the key follows. You see the structure in the simple.xml test file. This way we can export keys of different namespaces in a single xml file.

I guess up next we have to take care about more complicated use cases, like name escaping. E.g. having a key "user/\\\/key" should be fine but i think that would not be a valid XML element name. So i have to think about how to convert between key names and xml element names.

Also the xerces parser would offer a lot of possibilities (e.g. turn on/off pretty printing the xml, turn on some xml validation, ...). So we should think if/how we allow to configure the plugin.

markus2330:
The PR needs a rebase!

> Basic serialization, deserialization (including value & metadata) and error handling should be implemented now.

Very good!

> Relative path handling has to be done yet.

As commented, it should be the only option!

> The current xml structure is hard to decide IMO. On the one hand, if we want to export keys of various namespaces, we need some root element like namespace.

No, we do not need documents with multiple root elements. This will be done by kdb export ( #51 ).

> And if we want to be able to export keys which do not yet exist in any of the actual namespaces but is only cascading, how would we do that? Would we even do that? (otherwise i can leave the whole cascading namespace handling away).

Dont worry, everything is relative and below the parent key. Making sure that this is the case should be done

> On the other hand, we also want to support relative exports, As a basis, the approach of keyname == xml element, attribute == metadata and text in xml element == value, is useful i think, there are definately xml files out there that are structured like that.

Yes exactly!

> But what if an xml configuration file is structured differently, so would be mapped wrong to elektra?

The parentKey should be the root key, so it is mapped like all other keys.

> And another problem that arises, while there should be no real issue parsing special characters from xml files as the parser handles that (assuming these are well formed xml files), there are problems the other way around. In elektra its fine to have keys like /% or /\\/slashes, but xml does not allow element names like <%> or <\\/>. So that would speak against the previous approach. In xmltool this is no problem as we have a key element, where the name is an attribute, thus we can escape everything for xml. But with that we are not able to mount arbitrary xml files what we also want.

Yes, some mapping of key names is required. Doesn't xerces offer something like that?

> I only have the idea of a dual approach right know, so we can choose the output style. If you just want to import an existing arbitrary xml file or export a non-problematic part of the keys we can use this general format, and if we want to export special characters we'd fall back to the xmltool approach?

No, fall back does not sound good. If xerces does not offer a solution, we can improve the rename plugin so that we properly escape key names before.

> Also the xerces parser would offer a lot of possibilities (e.g. turn on/off pretty printing the xml, turn on some xml validation, ...). So we should think if/how we allow to configure the plugin.

I think we can turn all these on, performance shouldn't be an issue?

src/plugins/xerces/util.hpp
```diff
@@ -0,0 +1,83 @@
...
+struct XercesDeleter<char>
+{
+	void operator() (char * ptr) const
+	{
+		XERCES_CPP_NAMESPACE::XMLString::release (&ptr);
+	}
+};
+
+template <typename T>
+using xerces_unique_ptr = std::unique_ptr<T, XercesDeleter<T>>;
```
e1528532: so XercesUniquePtr? I usually use camelcase but as the std::unique_ptr is also written with an underline and xerces_unique_ptr is just a specialization of that, i sticked to that naming scheme

markus2330: Yes, or even XercesPtr is okay! There are plenty of such definitions in libtool, e.g. PluginPtr.

src/plugins/xerces/CMakeLists.txt
e1528532: i tested it with 3.0.0, worked, so i use 3.0.0 as a minimum version. Version 2 is no longer maintained, so i think its good to avoid that version (its from 2008). Maybe it works though, i haven't checked, but can do if you think it makes sense.

src/plugins/xerces/xerces.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+int elektraXercesOpen (Plugin * handle, Key * errorKey)
+{
+	XMLPlatformUtils::Initialize ();
+	return 1;
+}
+
+int elektraXercesClose (Plugin * handle, Key * errorKey)
+{
+	XMLPlatformUtils::Terminate ();
```
e1528532: According to https://xerces.apache.org/xerces-c/faq-parse-3.html#faq-4 it is. Only restriction is that it should be from the same thread (i haven't really cared about that for now, i guess thats usually the case anyway?)

markus2330: Mmh, surprising how many library developers were unable to program without global variables. Unfortunately, we cannot provide any guarantee from which threads we are called, applications decide that for us. There is not much we can do about it; except to warn applications that they should not call kdbOpen/kdbClose with their own xerces initialization concurrently. We could only use semlock to protect us from concurrent kdbOpen/kdbClose, but that would only protect applications not using xerces themselves. I added this requirement in #1399

src/plugins/xerces/CMakeLists.txt
markus2330: Even Debian Jessie has 3.1.1-5.1, so we can require 3.0.0.

e1528532:
rebased. Also now its always working with relative paths. Fixed serialization for that too. And added tests which use all kind of special characters and also those which are not allowed in xml, xerces handles everything properly.

The only issue that remains is that xerces doesn't handle illegal element & attribute names. I haven't really found a lot about that, only microsoft seems to offer something for that case: https://msdn.microsoft.com/en-us/library/35577sxd(v=vs.80).aspx
I think their approach should be fine basically, and if we can extend the rename plugin so it changes key names (also of metakeys) to this escaped version (and vice versa) we can also handle forbidden characters in xml element names. The only problem i can think of is that if someone calls a key like _x00010000_ for a reason (seems to be rather unlikely for me, but everything is possible ;) ) as it would also be unescaped then.

markus2330:
> rebased. Also now its always working with relative paths.

Perfect, than we can finish everything up, please set ready to merge if ready to merge ;)

> The only issue that remains is that xerces doesn't handle illegal element & attribute names.

We also have this issue for JSON and other plugins, so you should not solve this issue within this plugin. I opened  #1400 for further discussions.

Btw. version check and README still missing, did you push the latest version?

e1528532:
Last additions:
- Completed README, also with examples and limitations (> suggesting rename plugin already though we yet have to do that, but otherwise we might easily to forget to add the rename plugin hint)
- Use anonymous namespaces instead of static as recommended by our coding guidelines, xerces namespace for the plugin (similar to the dump plugin uses the dump namespace)
- Removed unnecessary header files
- added version check
- uses cppcheck as a linter and fixed possible issues
- renamed testdata folder to xerces too, as otherwise INSTALL_TEST_DATA option of cmake fails.

Unless there are still larger issues, i guess its good to merge for now. We can extend it in further PRs if required. There is one thing though, clang complains about several unused parameters, as i don't need every parameter the plugin API specifies currently. Should we supress those warnings somehow, or is there a good way to deal with that?

markus2330:
jenkins build please

markus2330:
jenkins build stable please

markus2330:
jenkins build unstable please

markus2330:
> supress those warnings somehow

Yes, simply but ELEKTRA_UNUSED there.

Where does memleak refer do? Do the test cases leak?

src/plugins/xerces/README.md
```diff
@@ -0,0 +1,61 @@
+- infos = Information about the template plugin is in keys below
+- infos/author = e1528532 <e1528532@libelektra.org>
+- infos/licence = BSD
+- infos/provides = storage/xml
+- infos/placements = getstorage setstorage
+- infos/status = recommended experimental unittest memleak
```
markus2330: memleak?

src/plugins/xerces/deserializer.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+	for (auto child = parent->getFirstChild (); child != NULL; child = child->getNextSibling ())
+	{
+		if (DOMNode::TEXT_NODE == child->getNodeType () || DOMNode::CDATA_SECTION_NODE == child->getNodeType ())
+		{
+			DOMText * data = dynamic_cast<DOMText *> (child);
+			if (!data->getIsElementContentWhitespace ()) str += toStr (data->getData ());
+		}
+	}
+	// trim whitespace
```
markus2330: Why trim whitespace here? (Please write a code comment why this is done)

Isn't this a limitation? (And makes it impossible to store whitespaces?)

src/plugins/xerces/testmod_xerces.c
```diff
@@ -0,0 +1,194 @@
...
+	KeySet * conf = ksNew (0, KS_END);
+	PLUGIN_OPEN ("xerces");
+
+	KeySet * ks = ksNew (0, KS_END);
+
+	succeed_if (plugin->kdbOpen (plugin, parentKey) == 1, "call to kdbOpen was not successful");
+
+	succeed_if (plugin->kdbGet (plugin, ks, parentKey) == 1, "call to kdbGet was not successful");
+
+	// succeed_if (plugin->kdbSet (plugin, ks, parentKey) == 1, "call to kdbSet was not successful");
```
markus2330: Why is this commented out? (Please write a code comment why this is done)

src/plugins/xerces/README.md
```diff
@@ -0,0 +1,61 @@
...
+- Maps key values to textual content of XML elements
+- Maps metakeys to XML attributes. Metakey name = attribute name, Metakey value 
+ 	= attribute value
+- Ignores XML comments
+
+## Usage
+
+To mount an XML file we use:
+
+    kdb mount file.xml user/test xerces
```
markus2330: Please use shell recorder syntax + add regression test for this example

src/plugins/xerces/deserializer.cpp
e1528532: you never fail to point out erroneous parts :)

Now that i've looked closer at it, i found that this is quite a problem where i have no general idea of how to fix it. 
The issue is that if there is no schema or DTD provided, xerces does not now wheter certain whitespace can be safely ignored (e.g. produced by pretty printing) or if it is indeed part of the value. Consider the following XML fragment, which is obviously pretty printed
```
<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
<xerces>  value of<!-- Comment inside text --> xerces  
	<!-- this is an attribute -->
	<fizz buzz="fizzBuzz" />
	<userKey user="key">withValue</userKey>
	<!-- those need escaping: " ' < > &  -->
	<späciöl_-keüs1 attr='"' attr2="$%(){}``äüö²[/\'>&quot;&lt;&apos;&amp;">>"&amp;&lt;'</späciöl_-keüs1>
	<cdata more-s_päcials="1 &amp; 2 are &lt; &#51; &#x0A;"><![CDATA[this is some cdata text "'<>&ä "]]></cdata>
</xerces>
```

Now the line of the xerces element contains the text `  value of xerces  ` (note the two intended spaces). But actually there are much more text nodes under the xerces element than only the first one, most of them containing only the tabs and newlines for pretty printing. So when appending all the text nodes together, we have a lot of newlines and tabs caused by pretty printing which we don't want to have in our output in that example.

So i though about just picking the first text node, but this would fail if we put the text after the other child elements, or have multiple text nodes between other element nodes (which is allowed in xml).

My other idea was the trimming which i did before, but you are right it may lead to unwanted behavior too. I could ignore text nodes which only contain whitespace as those are most likely caused by pretty printing, but this may result in some false positives. And the way you can pretty print an xml file may vary, so its really hard to make assumptions here.

If there is some XML schema or DTD provided, xerces can deduce from the schema if a given text node is just unnecessary whitespace or actually a relevant node and this should prevent such situations. But as a lot of XML documents probably have no schema or DTD, we need a good default behavior for this. Pretty printed documents are fairly common, but the way they are pretty printed may vary.

markus2330: > you never fail to point out erroneous parts :)

I wish I would see all of them, then we would have 200 issues less ;)

Thanks for the detailed analysis! Please document this in the README.md.
Ignoring whitespace-only-nodes looks like a good compromise, doesn't it?
It should cover the pretty-printing issues. Other spaces should be taken as-is, you never know if they are significant.

BernhardDenner:
If tested the current state of your plugin, but the serialization seems to have a problem with the parentkey.

What I've done:
create a simple xml file:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
<root>
      <elem1 attr="hello">value</elem1>
</root>
```
```shell
kdb mount /tmp/xmlfile.xml system/myxml xml
kdb set system/myxml/root/elem2 "hello"
```
this is what I get:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
<myxml>

  <root>
    <elem1 attr="hello">value</elem1>
    <elem2>hello</elem2>
  </root>

</myxml>
```
after a second set
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
<myxml>

  <myxml>
    <root>
      <elem1 attr="hello">value</elem1>
      <elem2>hello</elem2>
    </root>
  </myxml>

  <root>
    <elem3>world</elem3>
  </root>

</myxml>
```
and so on

BernhardDenner:
and the "infos/needs" is missing in the contract (README.md)

e1528532:
Thanks for pointing out the issue @BernhardDenner . It is due to a misassumption of mine how the mounting behaves, or how kdb set behaves. Apparently when setting a key it first deserializes relative to system/myxml , which will result in keys like system/myxml/root/elem1 . Then it serializes it back relative to system/myxml again, this time it uses myxml as the root element in order to ensure that there will be only 1 root element, which is sort of required to be done that way.

While the latter is fine IMO to prevent duplicate possible roots, i think the semantic of deserialization has to be changed slightly so this can work, specifically i think i have to enforce that the mount point's name and the xml root element matches. So you would mount the plugin like kdb mount system/myxml/root then instead to ensure it fits to the serialization procedure. At least i have no quick other idea how this could be implemented.

Furthermore i noticed that after a rebase i don't see the already contained keys of the xml file (like elem1 in your case) after mounting. Judging from the logs it looks like it never tries to deserialize it initially? As far as i can remember this used to work, did the mounting semantics change in the meanwhile or do i have a bug somewhere?

markus2330:
You need to make sure that you only work with names relative to the parentKey, see `elektraKeyGetRelativeName`. Then it does not matter where the root is, you exclusively work relatively to the parentKey.

> i think i have to enforce that the mount point's name and the xml root element matches

Do not do this please. This was done in earlier versions and it was a huge mistake and it is very difficult to fix all these places that try to do something like this. See #51

> Furthermore i noticed that after a rebase i don't see the already contained keys of the xml file (like elem1 in your case) after mounting. Judging from the logs it looks like it never tries to deserialize it initially? As far as i can remember this used to work, did the mounting semantics change in the meanwhile or do i have a bug somewhere?

No nothing changed. I do not understand your issue, though. First your `get` part of the plugin is called, you try to parse the file and insert keys relative to the parentKey. When writing out (`set` part), you simply cut of the common parentKey and serialize the relative keys back to XML.

Can you please update docu where relative keys should have been mentioned?

e1528532:
Ok i will try to avoid it. But maybe i should give an example of where i see an issue. Maybe it is very easy to bypass. It will also explain the mounting issue. Assume this xml file lets call it file.xml:

`<root><elem1>test</elem1></root>`

Now i mount it with `kdb mount file.xml user/test/file xerces`. But when i do a `kdb ls user/test` then i don't see the elem1 key. The issue seems to be that i looks for the file in <home>/.config/file.xml which does not exist after mounting. If i copy it over to that location manually the elem1 key will be successfully displayed.

So much to that, if i make new keys with `kdb set user/test/file/root main` and `kdb set user/test/file/root/key2 value2` it does everything correctly, the file.xml in the <home>/.config/ directory has the appropriate content:

`<root>main<elem1>test</elem1><key2>value2</key2></root>`

The issue with the relative keys arises now if i create a key with `kdb set user/test/file/root2 r2`. Suddenly i have two keys below the mountpoint, but in xml i can only have one root key. A simple way would now be not doing that. I cannot take the mountpoint itself into account if we don't want to enfore that its called the same as the xml root elem. Am I just having a missassumption somewhere or is this really not that easy to resolve (or is it fine this way?)

> Can you please update docu where relative keys should have been mentioned? (Different PR please.)

of course i can check but basically i was working with relative keys all the time now. To me it was clear to use relative keys but i'm unsure about the issue explained above

markus2330:
> Maybe something wrong with my build configuration?

No, looks like some kind of misunderstanding of namespaces and how the resolver works. `doc/tutorials/mount.md` should clear everything up. If you specify `file.xml` you do not refer to the `file.xml` in your current working directory. (Only `/file.xml` in the `dir` namespace does.)

But I agree we should emit a warning when there is a file in the same folder with exactly the same name. You are certainly not the only one with this expectation.

> but i'm unsure about the issue explained above

The *parentKey* itself already belongs to the KeySet to persist, and there is only one of it. So you could simply use the baseName of the parentKey as XML-rootkey.

But I see your point, this might not be wanted. If you mount the XML file to `/sw/elektra/dq/#0/current` you might not want the root to be named `current`.

Our options are:

1. have a plugin config if the user wants a root named differently than the `baseName`
2. store the name of the rootKey as metadata in the parentKey
3. derive the rootName from a schemata

BernhardDenner:
What about simply aborting the write process and issuing an error message if someone tries to add an additional root key?

e.g. xml file `<root><elem1>value</elem1></root>` mounted under `user/xml`
```sh
kdb ls user/xml
$> user/xml/root
$> user/xml/root/elem1
kdb set user/xml/root2
$> error, found 2 root elements but only one root element allowed in xml files
```

markus2330:
> What about simply aborting the write process and issuing an error message if someone tries to add an additional root key?

It would be a severe limitations about which KeySets can be written out.

By design the parent should be the root key. Root key with a name in the configuration file is quite specific to XML.

But it would be okay for a first demonstration in the web interface, and if it would be okay for you ( @BernhardDenner would be good if we can use it for Jenkins Config, then we have Elektra's enumerations and so on ) we can finish up the PR with this limitation.

e1528532:
I am already implementing the metakey approach which stores the original root name as this is most likely the expected default behavior. if not, the metakey can still be deleted manually so it doesn't remap anything then. will finish this approach today (i hope)

src/plugins/xerces/README.md
```diff
@@ -2,10 +2,11 @@
 - infos/author = e1528532 <e1528532@libelektra.org>
 - infos/licence = BSD
 - infos/provides = storage/xml
+- infos/needs =
 - infos/placements = getstorage setstorage
 - infos/status = recommended experimental unittest
+- infos/metadata = elektraXercesOriginalRootName
```
markus2330: Please no camelcase for metadata name. Maybe `rootname`? Can you please add it to `doc/METADATA.ini`?

e1528532: wouldn't something like elektraxercesrootname or xercesrootname or elektrarootname be better to have a lower chance of collisions with existing attributes?

markus2330: In `METADATA.ini` we define the meaning of meta data. There are no collisions, if another plugin chooses to use the same name, the other plugin has to adhere the same specifications and must use the meta data in an interchangable way.

So if another plugin has some sort of rootname, one can import/export between these plugins without loosing the rootname.

The name you have currently chosen (`xerces/rootname`) defines the metadata private to the plugin. So there can be no "collision", but also no interchange between plugins.

e1528532: ah i see. But the issue i was thinking about is that given the following xml file:
<root rootname="attribute"></root> would cause it to be serialized as an attribute then, since rootname would be both the name of the xml attribute and the metadata for storing the original name. With it being xerces specific this would be no issue. I thought that rootname isn't such an uncommon name that this would never be the case.

markus2330: Yeah, there is a more general issue of directly exposing meta data within a configuration file involved here. (The same issue exists with any meta data., e.g. comments)

We would need a way to specify if attributes are meta data to be interpreted by Elektra. But this is something for the next PR!

Btw. github cuts away XML.

e1528532: So should i leave it to xerces/rootname for now or just use rootname and ignore the possibility of conflicts for now?

markus2330: Yes, you can leave it with `xerces/rootname` for now. We can discuss it next meeting.

sanssecours: > Btw. github cuts away XML.

```xml
<that>
  <is>
    <not>
      <necessarily>
        <true>
          👋
        </true>
      </necessarily>
    </not>
  </is>
</that>
```
However, since Markdown supports [inline HTML](https://daringfireball.net/projects/markdown/syntax#html) I would assume the GitHub Markdown converter just throws away unknown tags inside Markdown files.

markus2330: Luckily it does not throw away but only does not display it. So I was able to read it in the email, and by pressing `Edit comment` you see it, too.

e1528532:
ok then finally i think this should be ready to merge. shellrecorder tests were also added with the latest commit. I'd still leave the experimental flag until we clarified the rootname metakey name, as this implies the contract of the plugin might still change a bit. If i've overseen anything please let me know, this PR got kind of large already but we encountered some issues that had to be discussed about.

markus2330:
Thank you, great job!

Having experimental for at least one release is standard, anyway.

PR #1403, opened 2017-03-05
Conditionals: multiple conditions

tom-wa:
# Purpose

Better evaluation of multiple conditions

# Checklist

- [x] I added unit tests

### DIFF
--- a/doc/METADATA.ini
+++ b/doc/METADATA.ini
@@ -509,17 +509,23 @@ usedby/plugin= conditionals
 type= string
 description= A unit of measure style prefix
 
-[check/condition/#]
+[check/condition/any/#]
 status= implemented
 usedby/plugin= conditionals
 type= string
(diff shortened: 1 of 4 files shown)


markus2330:
I am inclined to accept it as is, but thus it is not urgent I want to share a thought:

The mixing of array and other data might be problematic. Thus check/condition/eval only has a small number of possibilities, the alternative would be to have two arrays:

check/condition/any/# and check/condition/all/#

where the overall condition would be true iff any of any and all of all are true.

The problem here is that the change would be incompatible with previous arrays, so your solution is a good alternative.

markus2330:
Can you change it then to the new name please?

markus2330:
Thank you, looks great!

Can you please explain in the README what `assign/condition/#` is for?

Then you can set to "ready to merge".

markus2330:
Thanks!

Issue #1388, opened 2017-02-26
Hex Support


domhof:
In view of the kdb-types defined in kdbtypes.h we can implement the following getters in KDBHL:

```c
int kdbhlGetInt (KDBHL * kdbhl, const char * name);
const * char kdbhlGetString (KDBHL * kdbhl, const char * name);
kdb_boolean_t kdbhlGetBoolean (KDBHL * kdbhl, const char * name);
kdb_char_t kdbhlGetChar (KDBHL * kdbhl, const char * name);
kdb_octet_t kdbhlGetOctet (KDBHL * kdbhl, const char * name);
kdb_short_t kdbhlGetShort (KDBHL * kdbhl, const char * name);
kdb_unsigned_short_t kdbhlGetUnsignedShort (KDBHL * kdbhl, const char * name);
kdb_long_t kdbhlGetLong (KDBHL * kdbhl, const char * name);
kdb_unsigned_long_t kdbhlGetUnsignedLong (KDBHL * kdbhl, const char * name);
kdb_long_long_t kdbhlGetLongLong (KDBHL * kdbhl, const char * name);
kdb_unsigned_long_long_t kdbhlGetUnsignedLongLong (KDBHL * kdbhl, const char * name);
kdb_float_t kdbhlGetFloat (KDBHL * kdbhl, const char * name);
kdb_double_t kdbhlGetDouble (KDBHL * kdbhl, const char * name);
kdb_long_double_t kdbhlGetLongDouble (KDBHL * kdbhl, const char * name);
```

@tom-wa Is there anything else you would need or like to have?

markus2330:
If we keep using CORBA I would argue we should also use them exactly as specified and not introduce one or two extras ;)

@tom-wa usually does not see when he is mentioned, so you might need to send him an extra email. (He is subscribed to all changes.)

Btw. did you already look into `doc/todo/FUTURE`? Names ending with `_t` seem to be reserved by POSIX. We should clarify this before letting the types leak into the API.

And if we use them, please make a pedantic review of the types if they really conform to what they claim to be.

tom-wa:
i'll update my email filters.. 
i'm missing 2 things (they are not corba though so i'm not sure if they actually belong here): void * and unsigned char (because: compilers)

markus2330:
As already mentioned I would say we should take a deeper look into the CORBA spec.
I am all for making the types more strict, with known ranges and sign.

[...]

PR #1412, opened 2017-03-13
Elektra High Level API

domhof:
# Purpose

First draft of Elektra High Level API.

# Checklist

needs to be checked.

- [ ] I added unit tests


markus2330:
Thank you!

Can you put elektra.c into a separate library? It won't compile (and does not make sense) to put it into the core.

@tom-wa Can you do a review, too?

src/include/elektra.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+#ifndef ELEKTRA_H
+#define ELEKTRA_H
+
+#include "kdb.h"
+#include "kdbhelper.h"
+#include "kdbtypes.h"
+
+struct _ElektraError
```
markus2330: Structs should be hidden!

src/include/elektra.h
```diff
@@ -0,0 +1,56 @@
...
+{
+    int errorNr;
```
markus2330: Why is this needed? Do you want to change elektraHasError to return a kdb_long_t?

src/include/elektra.h
```diff
@@ -0,0 +1,56 @@
...
+    const char * msg;
```
markus2330: Shouldn't this be an array if you can hold errorNr errors?

src/libs/elektra/elektra.c
```diff
@@ -0,0 +1,88 @@
...
+
+const char * elektraErrorMessage (const Elektra * elektra)
+{
+    return elektra->error->msg;
+}
+
+void elektraClearError (Elektra * elektra)
+{
+    elektraFree (elektra->error);
+    elektra->error = NULL;
```
markus2330: Nice and simple!

src/libs/elektra/elektra.c
```diff
@@ -0,0 +1,88 @@
...
+{
+    Key * const nameKey = keyDup (elektra->parentKey);
+    keyAddName (nameKey, name);
+
+    Key * const resultKey = ksLookup (elektra->config, nameKey, 0);
+    const char * string = keyString (resultKey);
+
+    keyDel (nameKey);
+
+    const kdb_long_long_t value = ELEKTRA_LONG_LONG_S (string, NULL, 10);
```
markus2330: use keyGetLongLong here?

src/libs/elektra/elektra.c
```diff
@@ -0,0 +1,88 @@
...
+ * @brief Elektra High Level API.
+ *
+ * @copyright BSD License (see doc/LICENSE.md or http://www.libelektra.org)
+ */
+
+#include "elektra.h"
+#include <stdlib.h>
+#include <memory.h>
+
+Elektra * elektraOpen (const char * application)
```
markus2330: Please add error handling here!

It is important to understand which guarantees we can give to libraries.

src/include/elektra.h
domhof: Yes, we could do something like that (maybe with another function). The idea is that the developer can identify an error-type without parsing the string. Maybe it is not needed, I would say we will see which error-types we will have and can then decide upon that if we keep it or remove it.

src/include/elektra.h
domhof: errorNr just identifies the error-type. msg is a readable string that describes it.

src/libs/elektra/elektra.c
domhof: As far as I know we only have getString and getValue in the layer below. (we decided to remove all typed getters)

src/include/elektra.h
markus2330: > The idea is that the developer can identify an error-type without parsing the string.

Yes, sounds useful. For the code generator option we could even provide an enum.

> I would say we will see which error-types we will have

They are already defined in `src/error/specification`.

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * @param elektra The elektra instance initialized with the parent key.
+ * @param name The keyname to look up. The keyname is appended to the parent key.
+ */
+kdb_boolean_t elektraGetBoolean (Elektra * elektra, const char * name)
+{
+    READ_KEY
+
+    kdb_boolean_t value = 0;
+    if (!strcmp(string, "true") || !strcmp(string, "1") || !strcmp(string, "on")) {
```
markus2330: Here you introduce the same issue as discussed with hex values in #1388

We already have defined booleans to be `0` or `1`.

Furthermore, you need to check here if `check/type` isn't something else other than `boolean`.

src/include/elektra.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef ELEKTRA_H
+#define ELEKTRA_H
+
+#include "kdbtypes.h"
+
+typedef struct _ElektraError ElektraError;
+typedef struct _Elektra Elektra;
+
+Elektra * elektraOpen (const char * application);
```
markus2330: error parameter missing

src/include/elektra.h
```diff
@@ -0,0 +1,38 @@
...
+void elektraClose (Elektra * elektra);
+
+kdb_boolean_t elektraHasError (const Elektra * elektra);
```
markus2330: old API

src/include/elektraprivate.h
```diff
@@ -0,0 +1,29 @@
...
+    const char * msg;
+};
+
+struct _Elektra
+{
+    KDB * kdb;
+    KeySet * config;
+    Key * parentKey;
+
+    struct _ElektraError * error;
```
markus2330: not needed

src/libs/CMakeLists.txt
```diff
@@ -15,3 +15,5 @@ add_subdirectory (plugin)
 add_subdirectory (elektra)
 
 generate_manpage (elektra-libs FILENAME ${CMAKE_CURRENT_SOURCE_DIR}/README.md SECTION 7)
+
+add_subdirectory (highlevel)
```
markus2330: please add newline

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,245 @@
...
+Key * const nameKey = keyDup (elektra->parentKey); \
+keyAddName (nameKey, name); \
+\
+Key * const resultKey = ksLookup (elektra->config, nameKey, 0); \
+const char * string = keyString (resultKey); \
+\
+keyDel (nameKey); \
+
+#define RELEASE_KEY_AND_RETURN_VALUE \
+keyDel (resultKey); \
```
markus2330: is this macro really needed?

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,245 @@
...
+
+#include "elektra.h"
+#include <stdlib.h>
+#include <memory.h>
+#include "kdb.h"
+#include "kdbhelper.h"
+#include "elektraprivate.h"
+
+#define READ_KEY \
+Key * const nameKey = keyDup (elektra->parentKey); \
```
markus2330: Why duplicate? What about having a searchKey in `Elektra` and only modify its name?

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * @param elektra The elektra instance initialized with the parent key.
+ * @param name The keyname to look up. The keyname is appended to the parent key.
+ */
+const char * elektraGetString (Elektra * elektra, const char * name)
+{
+    READ_KEY
+
+    char * value = elektraMalloc (keyGetValueSize(resultKey));
```
markus2330: Do not return allocated memory to the user.

src/tools/gen/support/c.py
```diff
@@ -13,13 +13,13 @@ def funcname(self, key):
 			raise Exception("invalid keyname " + key)
 
 	def getfuncname(self, key):
-		return "get_"+self.funcname(key)
+		return "elektraGet"+self.funcname(key)
```
markus2330: This change would need a fix in the tests. It might be better to have a new py file which inherits from `c.py`.

Please cut off basename here, or fix specification of your user study example to not contain `Sw`.

markus2330:
Thanks for working on it!

Seems you need to rebase (conflicing file .gitignore)

Please remove "work in progress" when I should do a review.

tests/ctest/test_elektra.c
```diff
@@ -0,0 +1,82 @@
...
+    int result = kdbSet (handle, config, parentKey);
+    printf ("SET: result is %d\n", result);
+    if (result == -1) {
+        Key * problemKey = ksCurrent (config);
+        printf ("SET: problemKey is %s\n", keyName (problemKey));
+    }
+
+    // Close
+    kdbClose (handle, parentKey);
+    keyDel (parentKey);
```
markus2330: not everything cleaned up, e.g. config

tests/ctest/test_elektra.c
```diff
@@ -0,0 +1,82 @@
...
+    printf ("CHECK: value is %s\n", keyString(checkKey));
+
+    // Close
+    kdbClose (handle, parentKey);
+    keyDel (parentKey);
+}
+
+static void test_development ()
+{
+    const char * parentKey = "user/sw/elektra/kdb/#0/current";
```
markus2330: Only write in `/test` for test cases, e.g. `user/test/...`

markus2330: Btw. the test cases redirect `user` to `/tmp`, so you cannot use `kdb get` to check the key afterwards.

tests/ctest/test_elektra.c
```diff
@@ -0,0 +1,82 @@
...
+#include <elektra.h>
+
+static void set (const char * parentKeyName, const char * name, const char * value) {
+    // Open
+    KeySet * config = ksNew (0, KS_END);
+    Key * parentKey = keyNew (parentKeyName, KEY_END);
+    KDB * handle = kdbOpen (parentKey);
+    kdbGet (handle, config, parentKey);
+
+    printf ("Error occurred: %s\n", keyString (keyGetMeta (parentKey, "error/description")));
```
domhof: lldb tells me that the handle is NULL ;-)

tests/ctest/test_elektra.c
```diff
@@ -0,0 +1,82 @@
...
+
+    // Set
+    Key *key = keyNew(parentKeyName, KEY_END);
+    keyAddName(key, name);
+    keySetString(key, value);
+    ksAppendKey(config, key);
+
+    // Check
+    Key *checkKey = ksLookupByName (config, name, 0);
```
domhof: This file was never meant to be added to git, I tried different things until I found out that the KDB could not be loaded in the first place. I'm aware of the inconsistencies, please ignore them, the code will get removed anyway after we have found out why the resolver cannot be loaded (or about the real cause).

tests/ctest/test_elektra.c
domhof: see comment above

tests/ctest/test_elektra.c
```diff
@@ -0,0 +1,82 @@
...
+
+static void check (const char * parentKeyName, const char * name) {
+    // Open
+    KeySet * config = ksNew (0, KS_END);
+    Key * parentKey = keyNew (parentKeyName, KEY_END);
+    KDB * handle = kdbOpen (parentKey);
+    kdbGet (handle, config, parentKey);
+
+    // Check
+    Key *checkKey = ksLookupByName (config, name, 0);
```
domhof: see comment above

tests/ctest/test_elektra.c
domhof: Thank you! I didn't know about user being redirected to tmp. Is there a documentation for writing tests etc.?

tests/ctest/test_elektra.c
```diff
@@ -0,0 +1,82 @@
...
+    const char * name = "test";
+    const char * value = "1";
+
+    set (parentKey, name, value);
+    check (parentKey, name);
+}
+
+int main (int argc, char ** argv)
+{
+	printf ("KEY      TESTS\n");
```
domhof: See comment above. I initially copied test_key.c. This is just a leftover.

tests/ctest/test_elektra.c
markus2330: Okay, something seems to be very broken, then.

But in this case the error 37 with the description `handle or ks null pointer` should be set?

tests/ctest/test_elektra.c
markus2330: Okay, for me it worked except of this.

tests/ctest/test_elektra.c
markus2330: Only `doc/TESTING.md`. For non-ABI tests it is okay to use gtest, which is much more comfortable and better documented. (And according to @fberlakovich it has some CLion support.)

We can also consider to write ABI tests with gtest. Until now nobody checked what would happen in the case of gtest upgrades. Maybe it will work without issues anyway. (Because gtest is compiled in after all.)

We can also consider a gtest upgrade, we still have 1.7.0 but 1.8.0 was already released.

markus2330: Improvements of `doc/TESTING.md` and how we use gtest is of course very welcomed!

src/error/exporterrors.cpp
```diff
@@ -113,7 +113,29 @@ ostream & operator<< (ostream & os, parse_t & p)
 		os << p[i]["macro"] << " " << i << endl;
 	}
 
+    os << endl << endl;
+
+	/* Temporarily disable enum generation.
```
markus2330: Please revert that change, there is little use of passing all enums to the user.

src/include/kdbprivate.h
```diff
@@ -24,6 +24,32 @@
 
 #include <limits.h>
 
+#define KDB_STRING_TO_BOOLEAN(string) (kdb_boolean_t) !strcmp (string, "1")
```
markus2330: Maybe better in kdbmacros.h? Is this useful without any error handling? Prefer to prefix with ELEKTRA_

src/include/elektra_error_private.h
```diff
@@ -0,0 +1,26 @@
+/**
```
markus2330: One private header should be enough?

src/include/elektra_error_codes.h
```diff
@@ -0,0 +1,58 @@
+/**
```
markus2330: So many errors are not needed.

src/include/elektra_error.h
```diff
@@ -0,0 +1,33 @@
...
+typedef enum {
+    ELEKTRA_ERROR_SEVERITY_FATAL = 0,
+    ELEKTRA_ERROR_SEVERITY_ERROR,
+    ELEKTRA_ERROR_SEVERITY_WARNING
+} ElektraErrorSeverity;
+
+typedef const char * ElektraErrorGroup;
+typedef const char * ElektraErrorModule;
+
+ElektraErrorCode elektraErrorCode (ElektraError * error);
```
markus2330: rename to elektraErrorGetCode?

src/include/elektra.h
```diff
@@ -0,0 +1,393 @@
...
+#include "kdb.h"
+#include "kdbtypes.h"
+#include "elektra_error.h"
+
+typedef struct _Elektra Elektra;
+
+Elektra * elektraOpen (const char * application, ElektraError ** error);
+void elektraClose (Elektra * elektra);
+
+/**
```
markus2330: docu please in source files. Add more docu, use copyDoc for same content.

doc/decisions/high_level_api.md
```diff
@@ -75,13 +75,18 @@ sure to use pass a `KeySet`, that has your whole specification included
...
+ElektraErrorGroup elektraErrorGroup (ElektraError * error);
+ElektraErrorModule elektraErrorModule (ElektraError * error);
+
+void elektraErrorFree (ElektraError * error);
+
 ```
 
-`elektraErrorAbort` tells you if you need to quit your application because of severe
-issues that are permanent. Otherwise, developers can just print the message and
+`elektraErrorSeverity` tells you if you need to quit your application because of severe
```
markus2330: Also add this (or move this) parts to the official docu (tutorial)

.gitignore
```diff
@@ -32,3 +32,7 @@ nbproject/
 #ignore macOS folder attribute files
 .DS_Store
 .directory
+
+#ignore CLion project files
+.idea/
+cmake-build-debug/
```
markus2330: also add cmake-build-release/?

src/error/specification
```diff
@@ -24,6 +24,7 @@ module:dl
 severity:warning
 class:FailedLoadModuleError
 extends:ModuleError
+enum:DLOPEN_FAILED
```
markus2330: Great to see all these enum values!

src/error/specification
```diff
@@ -436,650 +506,746 @@ module:dl
...
 module:simpleini
+enum:INVALID_LINE_FORMAT_PATTERN
 
 number:174
 description:general xml error
 severity:error
 ingroup:plugin
 module:xerces
-
-number:175
```
markus2330: Please rebase, you remove something by accident here.

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,473 @@
...
+    {
+        *error = elektraErrorCreateFromKey(parentKey);
+        return NULL;
+    }
+
+    Elektra * const elektra = elektraCalloc (sizeof (struct _Elektra));
+    elektra->kdb = kdb;
+    elektra->parentKey = parentKey;
+    elektra->config = config;
+    elektra->lookupKey = keyNew(NULL);
```
markus2330: Please use KEY_END

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,473 @@
...
+
+
+Elektra * elektraOpen (const char * application, ElektraError ** error)
+{
+    Key * const parentKey = keyNew (application, KEY_END);
+    KDB * const kdb = kdbOpen (parentKey);
+
+    if (kdb == NULL)
+    {
+        *error = elektraErrorCreateFromKey(parentKey);
```
markus2330: error fatal?

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,473 @@
...
+
+        keySetString (key, value);
+
+        ret = kdbSet (elektra->kdb, elektra->config, elektra->parentKey);
+        if (ret == -1)
+        {
+            Key * problemKey = ksCurrent (elektra->config);
+            if (problemKey != NULL)
+            {
+//                ELEKTRA_LOG_DEBUG("problemKey: %s\n", keyName (problemKey));
```
markus2330: Why comment out?

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,473 @@
...
+            }
+
+            kdbGet (elektra->kdb, elektra->config, elektra->parentKey);
+        }
+    } while (ret == -1);
```
markus2330: Might not terminate.

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,473 @@
...
+        ret = kdbSet (elektra->kdb, elektra->config, elektra->parentKey);
+        if (ret == -1)
+        {
+            Key * problemKey = ksCurrent (elektra->config);
+            if (problemKey != NULL)
+            {
+//                ELEKTRA_LOG_DEBUG("problemKey: %s\n", keyName (problemKey));
+            }
+
+            kdbGet (elektra->kdb, elektra->config, elektra->parentKey);
```
markus2330: Unnecessary kdbGet if kdbSet was successful.

Please take a look at examples/kdbset.c or qt-gui for how to do this correctly. You also need to do a 3-way merge considering who changed the value (for the one key to set).

src/libs/highlevel/elektra.c
```diff
@@ -0,0 +1,473 @@
...
+
+        keySetString (key, value);
+
+        ret = kdbSet (elektra->kdb, elektra->config, elektra->parentKey);
+        if (ret == -1)
+        {
+            Key * problemKey = ksCurrent (elektra->config);
+            if (problemKey != NULL)
+            {
+//                ELEKTRA_LOG_DEBUG("problemKey: %s\n", keyName (problemKey));
```
markus2330: Why comment out? Seems to be duplication of code here?

src/libs/highlevel/README.md
```diff
@@ -0,0 +1,154 @@
...
+{
+  ElektraError * error = NULL;
+  Elektra * elektra = elektraOpen ("/sw/yourApplication", &error);
+
+  if (error != NULL) 
+  {
+    printf ("Sorry, there seems to be an error with your Elektra setup: %s\n", elektraErrorDescription (error));
+    printf ("Will exit now...\n");
+    exit (EXIT_FAILURE);
+    elektraErrorFree (error);
```
domhof: An error occured and the handle is NULL so elektraClose is not necessary. elektraErrorFree really makes no sense here, swapped the two lines.

src/libs/highlevel/README.md
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  const char * welcomeMessage = elektraGetString (elektra, "welcomeMessage");
```
domhof: Changed it to "message"

src/libs/highlevel/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Example
+```c
+#include <stdio.h>
+#include "elektra.h"
+
+int main(int argc, char ** argv)
+{
+  ElektraError * error = NULL;
+  Elektra * elektra = elektraOpen ("/sw/yourApplication", &error);
```
domhof: Changed it to "/sw/org/myapp/#0/current".

src/libs/highlevel/README.md
```diff
@@ -0,0 +1,154 @@
...
+### Struct `ElektraError`
+The library is designed to hide as many problems a developer can encounter when usign KDB directly. However it is not possible to hide all those issues. As with every library, things can go wrong and there needs to be a way react to errors once they have occurred at runtime. Therefore the High-Level API introduces a simplified struct called `ElektraError`, which encapsulates all information neccessary for the developer to handle runtime-errors appropriately in the application.
+
+Functions that can produce errors accept an ElektraError pointer as parameter, for example:
+
+```c
+Elektra * elektraOpen (const char * application, ElektraError ** error);
+```
+
+In most cases you'll want to set the error variable to NULL before passing it to the function. You can do this either by declaring and initializing a new variable with `ElektraError * error = NULL` or by reusing an already existing error variable by resetting it with `elektraErrorFree (error)`.
```
domhof: Thank you for pointing this out! In fact it was necessary. I changed the name to elektraErrorReset and the function now in addition resets the passed error variable to NULL. Seems to me more intuitive this way.

src/libs/highlevel/README.md
```diff
@@ -0,0 +1,154 @@
...
+First you have to add `elektra-highlevel` to the linked libraries of your application. To be able to use it in your source file, just include the main header with `#include "elektra.h"` at the top of your file.
+
+## Core Concepts
+
+### Struct `Elektra`
+`Elektra` is the handle you use to access the underlying KDB. All key-value read and write operations expect this handle to be passed as in as a parameter. To create the handle, you can simply write:
+
+```c
+ElektraError * error = NULL;
+Elektra * elektra = elektraOpen ("/sw/yourApplication", &error);
```
domhof: I think the user gets the point. Otherwise he would ask himself what the NULL parameter could be.

src/bindings/cpp/benchmarks/benchmark_thread.cpp
```diff
@@ -67,8 +67,9 @@ class Environment : public ThreadBoolean
 	: ThreadBoolean (ks, context_, Key ("/test/%layer1%", KEY_CASCADING_NAME, KEY_META, "default", "1", KEY_END)),
 	  nested (ks, context_), person (ks, context_),
 	  profile (ks, context_, Key ("/%layer1%/profile", KEY_CASCADING_NAME, KEY_META, "default", "default", KEY_END)),
-	  bm (ks, context_, Key ("/%layer1%/%layer2%/%layer3%/%layer4%/%layer5%/%layer6%/%layer7%/%layer8%/%layer9%/", KEY_CASCADING_NAME,
```
markus2330: Did you use clang-format-3.8? Maybe @sanssecours can give you a tip how to get it.

If we would use different versions of `clang-format` we would keep reverting our changes.

sanssecours: > Did you use clang-format-3.8? Maybe @sanssecours can give you a tip how to get it (on Mac).

You can install `clang-3.8` via `brew install clang-format@3.8`.

> If we would use different versions of `clang-format` we would keep reverting our changes.

We could also just use the current version (`5.0.0`) of the software.

markus2330: Thanks for giving the tip!

> We could also just use the current version (5.0.0) of the software.

Installing clang on all build agents take some time, so I would prefer to wait for @BernhardDenner's build server case study. Furthermore, it needs to be coordinated with all developers. We can discuss it in the next meeting, but for now @domhof please use 3.8.

domhof: Thanks, I used the current version (5.0.0). Reverted back to 3.8 for now.

markus2330:
Btw. before merging please rebase. And it would be great if you can put the module you are working on as short message before the commit messages. (kdbhl or similar)

And as long term perspective we should get rid of the python dep for the simple parts of the code generator (generating the key names and default config).

markus2330:
What is the status of this PR? Still seems to have a lot of wrong reformatting? Please check the PR at github.

markus2330:
Btw. last time we discussed you will start a new PR. What is the status? Please also make sure to work in all feedback you got so far and self-review what you have done.

PR #1419, opened 2017-03-18
elektra-web update

omnidan:
- better error handling
- implemented metadata
- usage of metadata for client-side validation and description
- bugfixes and small UI improvements
- implemented single-instance mode (pass host via `INSTANCE` env var)
- added information about auth and single instance usage in README


src/tools/web/clusterd/src/config.js
```diff
@@ -8,3 +8,5 @@
 
 export const PORT = process.env.PORT || 33334
 export const ROOT_PATH = process.env.ROOT_PATH || 'user/sw/elektra/web/#0/current'
+
+export const INSTANCE = process.env.INSTANCE || false
```
markus2330: Please also use Elektra here (to get INSTANCE) and document it.

src/tools/web/README.md
```diff
@@ -23,6 +23,7 @@ Elektra-web requires:
    * `cd clusterd`
    * `npm install`
    * `npm start`
+   * or `INSTANCE="http://localhost:33333" npm start` to connect to a single instance only
```
markus2330: More explanation what this mode does, please.

src/tools/web/client/src/components/Configuration.jsx
```diff
@@ -118,7 +127,7 @@ const Configuration = ({
     getKey: getCorrectKey,
     setKey: setCorrectKey,
     deleteKey: deleteCorrectKey,
-  }, tree)
+  }, { 'user': (tree && tree.user) || {} }) // only show user/ namespace
```
markus2330: Can you explain that change please?

Why would you only show one namespace?

src/tools/web/client/src/components/CreateKey.jsx
```diff
@@ -20,6 +20,11 @@ export default class CreateKey extends React.Component {
   render () {
     const { setKey } = this.props
 
+    // remove / prefix/suffix
```
markus2330: Does this round-trip nicely with escaped slashes?

src/tools/web/client/src/components/TreeItem.jsx
```diff
@@ -14,48 +14,139 @@ import React from 'react'
 import TextField from 'material-ui/TextField'
 import IconButton from 'material-ui/IconButton'
 import ActionDelete from 'material-ui/svg-icons/action/delete'
+import ContentCreate from 'material-ui/svg-icons/content/create'
 
 import TreeView from './TreeView.jsx'
 
+// TODO: set min/max appropriately for these types
```
markus2330: Yes, would be good to have it!

src/tools/web/client/src/components/TreeItem.jsx
```diff
@@ -14,48 +14,139 @@ import React from 'react'
...
+const INTEGER_TYPES = [
+  'short', 'unsigned_short', 'long', 'unsigned_long', 'long_long',
+  'unsigned_long_long',
+]
+
+const FLOAT_TYPES = [ 'float', 'double' ]
+
+const isNumber = (value) => !isNaN(value)
+const isInt = (value) => /^(-|\+)?[0-9]+$/.test(value)
```
markus2330: Isn't there a better way to check this? (try to convert to int?)

It seems to allow 0 prefix.

src/tools/web/client/src/components/Configuration.jsx
omnidan: I decided that elektra-web should focus on the editing of user configuration values.

I had an issue where validation wouldn't run, because I was running everything on the namespace (and not with `/` prefix). That's why I decided to only show the `user/` namespace and then prefix all set operations with `/` to make the validation run.

I also don't think that editing metadata/specifications should be the focus of elektra-web. The idea is to have an existing configuration/specification and edit values there.

While editing of metadata could be easily added later (the API is still there, I'm just only using `user/` and `/`), the focus of this project should be the interactive configuration tree.

src/tools/web/client/src/components/CreateKey.jsx
omnidan: slashes are escaped on a much lower level (in the `kdb.js` "library"), so this shouldn't cause an issue. it is only cleaning user input from unnecessary `/` prefixes/suffixes

omnidan:
Thanks for the review @markus2330! I fixed the issues and commented on your questions.

src/tools/web/client/src/components/Configuration.jsx
markus2330: > I decided that elektra-web should focus on the editing of user configuration values.

Yes, the focus is okay, but this `user` restriction seems to be arbitrary and not a focus? Most configuration admins care about lives in `system`. Or is this only as discussed with @tom-wa for PK? Let us discuss it via email.

When limiting scope, we should rather think about removing the cluster features?

> I had an issue where validation wouldn't run, because I was running everything on the namespace (and not with / prefix). That's why I decided to only show the user/ namespace and then prefix all set operations with / to make the validation run.

You can prefix everything with /. Then you do not need to care in which namespace the config is (except for creating new keys; here it would suffice for now that you can select the parameter of `-N`).

> I also don't think that editing metadata/specifications should be the focus of elektra-web. The idea is to have an existing configuration/specification and edit values there.

That is quite a limitation, then admins could not even modify descriptions or comments? For this PR meta data editing is not needed.

> While editing of metadata could be easily added later (the API is still there, I'm just only using user/ and /), the focus of this project should be the interactive configuration tree.

Metadata is an important part of the interactive configuration tree. You should update the use cases to better reflect your focus.

markus2330:
We will try to install it.

PR #1426, opened 2017-03-20
Various Small Fixes & Utility String Functions

sanssecours:
# Purpose

This branch contains various small fixes and three new private utility functions 

1. `lskip`: Returns a pointer to the first non-whitespace character in a string
2. `rstrip`: Removes whitespace from the end of a string
3. `strip`: Combines the first two functions to remove leading and trailing whitespace

. The repository already contains at least two versions of the first two functions in [`inih.c`](https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/ini/inih-r29/inih.c#L25-L43) and [`mozprefs.c`](https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/mozprefs/mozprefs.c#L32-L45). Since I also wanted to use the functionality provided by these functions, I concluded that it might be a good idea to actually write a third version 😀 of `lskip` and `rstrip`. The include utility functions have the added benefit,  that they are at least partially tested, reasonably well documented and can be used by just including `kdbinternal.h`. I am not really sure about the current interface of `rstrip` though. Providing a pointer to the end of the modified string might be overkill.

# Checklist

- [x] I added unit tests.
- [x] I documented all included utility functions.
- [x] I added code comments and assertions

@markus2330: 👋 Please review my pull request.


sanssecours:
Currently Jenkins reports undefined references to `lskip`, `rstrip` and `strip`:

```
CMakeFiles/test_internal.dir/test_internal.c.o: In function `test_lskip':
tests/ctest/test_internal.c:257: undefined reference to `lskip'
tests/ctest/test_internal.c:258: undefined reference to `lskip'
tests/ctest/test_internal.c:259: undefined reference to `lskip'
tests/ctest/test_internal.c:260: undefined reference to `lskip'
tests/ctest/test_internal.c:261: undefined reference to `lskip'
CMakeFiles/test_internal.dir/test_internal.c.o: In function `test_rstrip':
tests/ctest/test_internal.c:272: undefined reference to `rstrip'
tests/ctest/test_internal.c:273: undefined reference to `rstrip'
tests/ctest/test_internal.c:277: undefined reference to `rstrip'
tests/ctest/test_internal.c:279: undefined reference to `rstrip'
tests/ctest/test_internal.c:283: undefined reference to `rstrip'
CMakeFiles/test_internal.dir/test_internal.c.o:tests/ctest/test_internal.c:285: more undefined references to `rstrip' follow
CMakeFiles/test_internal.dir/test_internal.c.o: In function `test_strip':
tests/ctest/test_internal.c:321: undefined reference to `strip'
tests/ctest/test_internal.c:324: undefined reference to `strip'
```
. Since the build process works on my machine: Might I ask where I need to put the definitions of these functions for the compilation to also works on Linux?

markus2330:
Thank you for the many fixes and your proposal to add these 3 useful utility functions!

Thus only plugins use (their own implementations of) the function, there is no reason to put them into the core. (Only if required in the core, it should be in the core.) Instead we can put them into one of the other libraries, either `ease` or we start with a new `helper`/`utility` library.

Please also note that we need to prefix every name with `elektra`, otherwise applications might get symbol clashes if they have defined the same name.

And can you also fix the plugins to actually use the functions?

> May I ask where I need to put the definitions of these functions for the compilation to also works on Linux?

Names within the core not starting with elektra are not exported.

sanssecours:
> Thus only plugins use (their own implementations of) the function, there is no reason to put them into the core. (Only if required in the core, it should be in the core.) Instead we can put them into one of the other libraries, either ease or we start with a new helper/utility library.

It was never my intention to put the functions in a place where anyone besides the Elektra developers can use them. In one of my latest commits I put the functions into a new `utility` library as you suggested.

> Please also note that we need to prefix every name with elektra, otherwise applications might get symbol clashes if they have defined the same name.

I prefixed the functions with the name `elektra` in one of the latest comments. I have to admit I never thought namespaces were an essential programming language feature until recently…

> And can you also fix the plugins to actually use the functions?

I replaced `rstrip` and `lskip` in the `mozprefs` plugin with the new functions. The tests of the `mozprefs` plugin still seem to work. The only logical conclusion is that the new functions are a perfect replacement 😀 for the old ones. In all seriousness: I hope that the functions do not cause any new problems.

markus2330:
Thank you, looks great!

@tom-wa Can you take a look if everything is okay in mozpref+inih?

I would suggest that we quickly take a look at the PR tomorrow so that the new utility library is more known.

Any reason why the PR is not ready to merge?

sanssecours:
> Any reason why the PR is not ready to merge?

I just waited for Jenkins to finish all tests.

markus2330:
Thank you!

Issue #1424, opened 2017-03-20
JSON error in webinterface


reox:
I get this error message popping up:
```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

is there a way to debug this more?

i added the single instance fix both with env variable and in kdb but it seems not to work...
Using now nodejs 6.10.0

omnidan:
@reox I am `kdb export`ing to a temporary buffer file, because there are no official node.js bindings yet, and there was an issue with exporting/importing through stdout on linux. This buffer file should be in `/tmp/elektra-web.buffer.json`. You can try checking the contents of this file, or deleting it and then restarting elektra-web.

Sometimes there is an issue with it not being fully written, although this very rarely happened to me in developer mode, and never happened in production mode.

If it persistently happens, please let me know your specs (OS + version, elektra version, ...) so I can try to reproduce it.

reox:
hi!
The file looks like a empty json file to me:
```
sebastian@pcs:~$ cat /tmp/elektra-web.buffer.json
{

}
sebastian@pcs:~$ hd /tmp/elektra-web.buffer.json
00000000  7b 0a 0a 7d 0a                                    |{..}.|
00000005
```

Maybe because there is not much configuration else?

omnidan:
@reox should be fine then - does this error happen every time and persist through refreshes? which OS and elektra version is it running on?

markus2330:
@omnidan Are you sure you do not run `kdb` concurrently with the same buffer? Why not use fresh names for every buffer and unlink the file afterwards?

reox:
@omnidan added the information in the last comment. No it comes on every action. Maybe this is a problem because of the nginx reverse proxy? Do i need to pass some extra arguments? I just did `proxy_pass`, nothing else.

markus2330:
@omnidan And please improve the error messages, "SyntaxError: Unexpected token < in JSON at position 0" isn't something a user should read.

reox:
When I start it with the INSTANCE set, should there be a button Add Instance? I also set the corresponding elektra key, but there is no effect... Is it possible this bug is due to some missing settings?

markus2330:
In any case, the error message is really troublesome. @omnidan Please add some checks if everything is as expected and give better error messages.

PR #1430, opened 2017-03-22
elektra-web buffer file fix

omnidan:
@reox unfortunately I could not reproduce the issue in #1424 

I think there is a race condition with the buffer file - please test this potential fix for the bug and let me know if it works now


markus2330:
@reox Can you try to reinstall please?

reox:
same error as before :/

Issue #1049, opened 2016-10-23
Crypto: Utilize libinvoke


petermax2:
# Problem

The ~~latest~~ compile error ( #1043 ) revealed a design flaw in shared code from the plugins:
- crypto
- fcrypt
- base64

These plugins share two common functionalities:
- GPG interface (`crypto/gpg.c`)
- Base64 encoding / decoding (`base64/base64_functions.c`)

Statically linked builds caused troubles with function name resolution, as the shared code is compiled per plugin.

Utilizing `libinvoke` solves the problem, if the `base64` plugin exports its encoding functions properly.

~~Moving these features to a library (or maybe two libraries) would reduce compile time and lead towards a cleaner API. Also readability in affected `CMakeLists.txt` files would increase, because instead of~~
- ~~including the source files from other plugins~~
- ~~adding the plugin source folder to the include path~~

~~a single line to the call of `add_plugin` is required instead: `LINK_ELETRA elektra-libname`.~~

# Proposal

- The `base64` plugin exports its encoding functions.
- The `crypto` plugin utilizes `libinvoke` to encode/decode Base64 strings.
- The `crypto` plugin no longer compiles the source files of the `base64` plugin into its code, thus eliminating the need for complicated `ELEKTRA_PLUGIN_FUNCTION` macro uses.

EDIT: update proposal, edit problem statement

markus2330:
Some quick thoughts:

- What about a library that does not only execute gpg, but can be used to invoke any process? (gpg only as convenience wrapper)
- I am not conviced that encoding functionality should be moved to a library, but we should rather fix the plugin system by either:
  1. allow plugin configuration to be encoded (this seems the reason why crypto depends on base64). A quite easy way to do so would be to add an encoding and the boolean plugin in `elektraBackendOpenDefault`.
  2. Load plugins from other plugins: Actually you could easily introduce a soft dependency from crypto to base64 which only uses base64 if it was found at runtime (keep binary key, otherwise).

petermax2:
> What about a library that does not only execute gpg, but can be used to invoke any process? 

Sounds good. Especially if we want to support OS that do not provide `fork` (thinking about Windows).

> Load plugins from other plugins: Actually you could easily introduce a soft dependency from crypto to base64 which only uses base64 if it was found at runtime (keep binary key, otherwise).

[...]

PR #1431, opened 2017-03-23
resolver updates

tom-wa:
# Purpose

fixed user and system ns resolving

currently duplicated files for common resolver functions, i'd like to move them to a library for resolvers, not sure where i should put it though. it's just a slightly modification from `resolver/filename.c`


markus2330:
Thanks for the fix! But yes, we really should avoid the duplication.

Wouldn't it be best to load the `resolver` module (which can be configured via symlink, or might be built-in in the FULL/STATIC variant) and use a `resolveFile` function from the module? Then it even would work with `wresolver` or others OS-specific resolver yet to write.

One issue is that `gitresolver`, `blockresolver` and so on never could be default resolver (they would try to get `resolveFile` from themselves, which obviously cannot work and should be terminated with an error), but that shouldn't be much of an issue?

tom-wa:
yeah, that would work too. i initially wanted to use the filename function of the resolver, but i only need the resolved path, no temporary filenames, or the whole resolveHandle structure. so for now i've copied the functions and changed them so that they only need the path + namespace and return the resolved path

markus2330:
> yeah, that would work too.

That would be much better! It is hard to get rid of such code duplication after we introduced it.

> i initially wanted to use the filename function of the resolver, but i only need the resolved path, no temporary filenames, or the whole resolveHandle structure.

Do not be shy and restructure `src/plugins/resolver/filename.c` so that you can get exactly the information you need. There are some tests for this functionality already.

> so for now i've copied the functions and changed them so that they only need the path + namespace and return the resolved path

It would be better if the functions with the handle call the functions you now copied.

tom-wa:
what about other functions needed by resolvers (like moving a file, hashing of a file, etc..) ?

markus2330:
> what about other functions needed by resolvers

Only export what is needed elsewhere. What else other than resolving filenames is needed?

> (like moving a file,

That is trivial. Or do you mean a version that also works on Windows?

> hashing of a file, etc..) ?

This is currently not implemented.

tom-wa:
i know it's trivial, but it's still code that probably needs to be used in most resolvers (i currently use it in 2 places since rename doesn't work if source and target are on different devices), creating a path, etc.. that would be a lot of duplicated code which could be shared

markus2330:
> i know it's trivial, but it's still code that probably needs to be used in most resolvers (i currently use it in 2 places since rename doesn't work if source and target are on different devices),

Yes, but this code is not needed for the standard resolver because it only renames within one directory.

> creating a path, etc.. that would be a lot of duplicated code which could be shared

Ok, agreed. But system abstraction should not be in `utility` because of the POSIX dependence. 

What about creating `libelektra-fs` for file system functionality for POSIX systems?

Or because of #1433 maybe we should call it `libelektra-system`. Then we could also include stuff like `getpwnam_r` to get home directories.

markus2330:
Btw.: API first: please first think of a consistent API before starting moving stuff to the library.

tom-wa:
sounds like a good idea, yeah it was something i only became aware of while updating them
hashing a file or buffer is actually needed in some cases. curl- and gitresolver won't work without it, and probably other resolvers to come where it's possible that the file doesn't exist locally.

markus2330:
Yes, but signing/verifying would be something along the lines  #1049, and rather not be part of the system library. Maybe @petermax2 finds some time to give us a good crypto API for such hashes/signatures.

So it is definitely out-of-scope for this PR.

markus2330:
Thanks for updating the PR! The API looks promising!

libproposal, however, is not a good place for this functionality. (because of dep to kdb) What about a lib that is used to invoke functions in plugin? Maybe `libelektra-invoke`?

tom-wa:
@markus2330 would something like this be fine for `libelektra-invoke` ? 
it's basically
```
void *handle = elektraInvokeInitialize(const char * elektraPluginName);  // loads plugin
const void *functionPtr = elektraInvokeGetFunction(void * handle, const char * elektraPluginFunctionName); 
(void) elektraInvokeClose( void * handle);  // closes plugin
```

markus2330:
The API has quite some overlap with the modules API. What about:

`const void *functionPtr = elektraInvokeGetFunction(KeySet * modules, const char * plugin, const char * elektraPluginFunctionName);`

For easy use, you can internally create a new `modules` if null pointer is passed. Or can you do a benchmark if the modules impl is faster than doing `dlopen` again? (If not we can leave the modules pointer at all).

Handles should be done by typedef, not with `void*`.

And it would be better to return a function pointer, because casting data to function pointers is not portable.

tom-wa:
>What about:
const void *functionPtr = elektraInvokeGetFunction(KeySet * modules, const char * plugin, const char * elektraPluginFunctionName);
For easy use, you can internally create a new modules if null pointer is passed.

i don't really understand, do you mean like calling elektraInvokeGetFunction(NULL, "resolver", "filename") ? wouldn't that result in a memleak because i won't be able to close the plugin afterwards ?

markus2330:
> wouldn't that result in a memleak because i won't be able to close the plugin afterwards ?

Yes, you are right. Forget the sentence with "For easy use". It would memleak or need global variables.

tom-wa:
so, keep the current implementation and expose the handler typedef ?
it's split in init(plugin) and get(handle, function) instead of init(), get(modules, plugin, function) because i think most of the time you'd want multiple functions from one plugin instead of multiple plugins with one or more functions

markus2330:
> so, keep the current implementation and expose the handler typedef ?

Okay.

tom-wa:
ok. 

the multifile resolver now depends on fts(3) which was introduced in 4.4BSD. i could make it a compile variant, but seems like every *nix system supports that function for at least 10 years so i guess it's not much of a problem ? or should we start reimplementing some functions (e.g. fgetpwent, putpwent, fts, ...) to keep everything posix conform?

markus2330:
> but seems like every *nix system supports that function for at least 10 years so i guess it's not much of a problem ?

Sounds good!

> or should we start reimplementing some functions (e.g. fgetpwent, putpwent, fts, ...) to keep everything posix conform?

Seems to be unnecessary overkill!

tom-wa:
i think we can merge this on. the only thing missing is a documentation for libinvoke, but sure where it belongs though, `doc/decisions/` ?

markus2330:
For some sentences the space in `src/libs/README.md` is a good place, for longer docu  `src/libs/invoke/README.md` fits.

Mac OS X has some linker error with this PR.

tom-wa:
>/Users/travis/build/ElektraInitiative/libelektra/src/plugins/resolver/filename.c:134:29: error: use of undeclared identifier '_SC_GETPW_R_SIZE_MAX'

>        ssize_t bufSize = sysconf (_SC_GETPW_R_SIZE_MAX);

i have no idea where that error comes from, because that line isn't new

sanssecours:
> i have no idea where that error comes from, because that line isn't new

I do not know if it helps, but the very first commit “helper function for resolvers” already produces  messages that look very similar to the one you mentioned above:

```
[14/308] Building C object src/plugins/gitresolver/CMakeFiles/elektra-gitresolver-objects.dir/resolve.c.o
FAILED: src/plugins/gitresolver/CMakeFiles/elektra-gitresolver-objects.dir/resolve.c.o
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -DELEKTRA_STATIC -DHAVE_KDBCONFIG_H -DPLUGIN_SHORT_NAME=gitresolver -DREADME=readme_gitresolver.c -Isrc/include -I../src/include -I/usr/local/include -I../src/plugins/base64 -Isrc/plugins/gitresolver -std=gnu99  -Wno-deprecated-declarations -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer -fsanitize=integer   -Wno-long-long -pedantic -Wno-variadic-macros -Wall -Wextra -Wno-overlength-strings -Wsign-compare -Wfloat-equal -Wformat-security -Wshadow -Wcomments -Wtrigraphs -Wundef -Wuninitialized -Winit-self -Wsign-compare -Wfloat-equal -Wformat-security -O2 -g -DNDEBUG   -fPIC -MD -MT src/plugins/gitresolver/CMakeFiles/elektra-gitresolver-objects.dir/resolve.c.o -MF src/plugins/gitresolver/CMakeFiles/elektra-gitresolver-objects.dir/resolve.c.o.d -o src/plugins/gitresolver/CMakeFiles/elektra-gitresolver-objects.dir/resolve.c.o   -c ../src/plugins/gitresolver/resolve.c
../src/plugins/gitresolver/resolve.c:25:5: warning: implicitly declaring library function 'snprintf' with type 'int (char *, unsigned long, const char *, ...)' [-Wimplicit-function-declaration]
    snprintf(dirname, dirnameSize, "%s", keyName(canonify) + 4);
    ^
../src/plugins/gitresolver/resolve.c:25:5: note: include the header <stdio.h> or explicitly provide a declaration for 'snprintf'
../src/plugins/gitresolver/resolve.c:36:32: error: use of undeclared identifier '_SC_GETPW_R_SIZE_MAX'
    ssize_t bufSize = sysconf (_SC_GETPW_R_SIZE_MAX);
                               ^
../src/plugins/gitresolver/resolve.c:253:18: warning: implicit declaration of function 'strtok_r' is invalid in C99 [-Wimplicit-function-declaration]
        char * result = strtok_r (str, ":", &saveptr);
                        ^
../src/plugins/gitresolver/resolve.c:253:9: warning: incompatible integer to pointer conversion initializing 'char *' with an expression of type 'int' [-Wint-conversion]
        char * result = strtok_r (str, ":", &saveptr);
               ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/plugins/gitresolver/resolve.c:261:11: warning: incompatible integer to pointer conversion assigning to 'char *' from 'int' [-Wint-conversion]
                        result = strtok_r (0, ":", &saveptr);
                               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/plugins/gitresolver/resolve.c:275:10: warning: incompatible integer to pointer conversion assigning to 'char *' from 'int' [-Wint-conversion]
                result = strtok_r (0, ":", &saveptr);
                       ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
5 warnings and 1 error generated.
```
.

src/plugins/resolver/filename.c
```diff
@@ -5,23 +5,27 @@
  *
  * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
  */
+#define _POSIX_SOURCE
```
sanssecours: Looks like you also need to define the minimum POSIX level you need on macOS:

```c
#define _POSIX_SOURCE
#define _POSIX_C_SOURCE 200409L 
```

.

src/libs/invoke/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+file (GLOB SOURCES *.c)
+add_lib (invoke SOURCES ${SOURCES})
```
sanssecours: As far as I can tell, we also need to link `elektra` here, since this lib contains the code for `elektraPluginOpen` and `elektraPluginClose`:

```cmake
add_lib (invoke SOURCES ${SOURCES} LINK_ELEKTRA elektra)
```

.

sanssecours:
The code changes mentioned in the two comments above should be enough to fix the compilation and linker problems on macOS:

```diff
--- a/src/libs/invoke/CMakeLists.txt
+++ b/src/libs/invoke/CMakeLists.txt
@@ -1,2 +1,2 @@
 file (GLOB SOURCES *.c)
-add_lib (invoke SOURCES ${SOURCES})
+add_lib (invoke SOURCES ${SOURCES} LINK_ELEKTRA elektra)
diff --git a/src/plugins/resolver/filename.c b/src/plugins/resolver/filename.c
index 722697269..f8be9aac8 100644
--- a/src/plugins/resolver/filename.c
+++ b/src/plugins/resolver/filename.c
@@ -6,6 +6,7 @@
  * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
  */
 #define _POSIX_SOURCE
+#define _POSIX_C_SOURCE 200409L

 #include "resolver.h"
```

. Hope the information above is helpful, since it took me quite some time to locate the first issue 😢.

PR #1445, opened 2017-03-30
fix for running elektra-web with nginx reverse proxy and PAM auth

### LINKED ISSUE #1424: JSON error in webinterface

omnidan:
Fixes #1424

The problem was that the `fetch()` requests in the app did not send the `Authorization` header, thus nginx rejected the requests. This was fixed by adding `{ credentials: 'same-origin' }` to the fetch options, as suggested in: https://bugs.chromium.org/p/chromium/issues/detail?id=504344

This PR also shows invalid server responses as "Error: invalid response from server" instead of showing the JSON error.


markus2330:
Thank you!

PR #1448, opened 2017-04-02
Fix Warning and Various Runtime Errors

sanssecours:
# Purpose

This pull request fixes various run-time errors reported by `kdb ls` after we enable address sanitization.

# Checklist

@markus2330 @tom-wa 👋 Please review my pull request.

### DIFF
--- a/doc/CODING.md
+++ b/doc/CODING.md
@@ -76,7 +76,7 @@ Thus please use following techniques (in order of preference):
    `ELEKTRA_LOG ("formatted text to be printed according log filters", ...)`
 
    There are four log levels (ERROR is reserved for aborts within `ELEKTRA_ASSERT`):
-   - ELEKTRA_LOG_WARNING, something critical that should be shown the user (e.g. API misuse), see #ELEKTRA_LOG_LEVEL_WARNING
+   - ELEKTRA_LOG_WARNING, something critical that should be shown to the user (e.g. API misuse), see #ELEKTRA_LOG_LEVEL_WARNING
    - ELEKTRA_LOG_NOTICE, something important developers are likely interested in, see #ELEKTRA_LOG_LEVEL_NOTICE
    - ELEKTRA_LOG, standard level gives information what the code is doing without flooding the log, see #ELEKTRA_LOG_LEVEL_INFO
    - ELEKTRA_LOG_DEBUG, for less important logs, see #ELEKTRA_LOG_LEVEL_DEBUG
--- a/src/libs/elektra/keyset.c
+++ b/src/libs/elektra/keyset.c
@@ -911,7 +911,7 @@ ssize_t ksCopyInternal (KeySet * ks, size_t to, size_t from)
 	ELEKTRA_ASSERT (length >= 0, "length %zu too small", length);
 	ELEKTRA_ASSERT (ks->size >= to, "ks->size %zu smaller than %zu", ks->size, to);
 
-	ks->size += sizediff;
+	ks->size = ssize + sizediff;
 
 	if (length != 0)
 	{
--- a/src/libs/elektra/log.c
+++ b/src/libs/elektra/log.c
@@ -149,7 +149,9 @@ int elektraVLog (int level ELEKTRA_UNUSED, const char * function ELEKTRA_UNUSED,
 	ret |= elektraLogFile (level, function, file, line, msg);
 #endif
 
+#ifndef NO_FILTER
 end:
+#endif
 	elektraFree (str);
 	elektraFree (msg);
 	return ret;
--- a/src/libs/elektra/log.c
+++ b/src/libs/elektra/log.c
@@ -109,7 +109,8 @@ static void replaceChars (char * str)
 			str[lenOfMsg] = '@';
 		else if (str[lenOfMsg] == '\f')
 			str[lenOfMsg] = '@';
-	} while (lenOfMsg--);
+		lenOfMsg--;
+	} while (lenOfMsg);
 	str[last] = '\n';
 }
 
--- a/src/plugins/list/list.c
+++ b/src/plugins/list/list.c
@@ -362,7 +362,7 @@ int elektraListGet (Plugin * handle, KeySet * returned, Key * parentKey)
 	ksRewind (pluginKS);
 	int ret = runPlugins (pluginKS, placements->modules, placements->plugins, ksDup (config), returned, parentKey, GET, ksNext);
 	placements->getCurrent = ((++currentPlacement) % getEnd);
-	while (!placements->getPlacements[currentPlacement])
+	while (!placements->getPlacements[placements->getCurrent])
 	{
 		placements->getCurrent = ((++currentPlacement) % getEnd);
 	}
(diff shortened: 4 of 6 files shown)


markus2330:
Thank you! I like the changes except the cast for elektraFree. Does your compiler emit a warning there? Is there some way to suppress this warning for different pointers passed to elektraFree?

sanssecours:
>  I like the changes except the cast for elektraFree. Does your compiler emit a warning there? 

Yes. The build server reports the same [warning](https://build.libelektra.org/job/elektra-clang/lastBuild/warnings9Result/package.1645469987):

```
testmod_range.c:182: passing 'const char *' to parameter of type 'void *' discards qualifiers
```

. 

> Is there some way to suppress this warning for different pointers passed to elektraFree?

I do not know of any. I have to admit though, that I did not look too much into this problem. 

We could also fix the warning by removing the `const` qualifier from the variable `old_locale`, if you prefer that. Another option would be a more specific cast to `char *` instead of `void *`.

markus2330:
Ahh okay, then removing const is the correct fix.

sanssecours:
> Ahh okay, then removing const is the correct fix.

Thank you for the information. I updated the last commit accordingly.

markus2330:
I think this is ready to merge. Do you want to wait for @tom-wa to review?

sanssecours:
> I think this is ready to merge. Do you want to wait for @tom-wa to review?

If the changes in commit c2e4a39b are correct from your point of view, then I think it is alright if we merge this without waiting for Tom’s review.

markus2330:
Ok, then lets have @tom-wa a look.

sanssecours:
> Ok, then lets have @tom-wa a look.

Since this probably will take some time, I fixed another runtime error in the latest commit. I am not sure if the code changes address the problem properly though. Do we assume that `elektraPluginGetData` in [line 76 of `resolver.c`](https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/resolver/resolver.c#L76):

```c
resolverHandles * pks = elektraPluginGetData (handle);
```

never returns a null pointer?

markus2330:
> never returns a null pointer?

Yes, we assume it (because open always sets the handle on success, thus it must be available in get). But this would be a perfect place for `ELEKTRA_ASSERT`.

sanssecours:
> Yes, we assume it (because open always sets the handle on success, thus it must be available in get). But this would be a perfect place for `ELEKTRA_ASSERT`.

I replaced the `if` with an assertion in the latest commit. The text below shows the backtrace of the crash after I run `lldb -- kdb ls /` starting with a clean configuration:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff9da47d42 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff9db355bf libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fff9d9ad420 libsystem_c.dylib`abort + 129
    frame #3: 0x0000000100fc90c7 libelektra-core.4.dylib`elektraAbort(expression=<unavailable>, function=<unavailable>, file=<unavailable>, line=<unavailable>, msg=<unavailable>) at nolog.c:33 [opt]
    frame #4: 0x000000010454301f libelektra-resolver.so`elektraGetResolverHandle(handle=<unavailable>, parentKey=0x0000606000000d40) at resolver.c:78 [opt]
    frame #5: 0x0000000104541fc2 libelektra-resolver.so`libelektra_resolver_fm_hpu_b_LTX_elektraPluginget(handle=0x000060700000cc00, returned=0x0000604000009410, parentKey=0x0000606000000d40) at resolver.c:441 [opt]
    frame #6: 0x0000000100f26414 libelektra-kdb.4.dylib`kdbGet at kdb.c:468 [opt]
    frame #7: 0x0000000100f26066 libelektra-kdb.4.dylib`kdbGet(handle=0x000061600000f680, ks=0x000060400000d010, parentKey=<unavailable>) at kdb.c:804 [opt]
    frame #8: 0x0000000100026d61 kdb`kdb::KDB::get(this=<unavailable>, returned=<unavailable>, parentKey=0x000060400000f198) at kdb.hpp:183 [opt]
    frame #9: 0x0000000100188a41 kdb`LsCommand::execute(this=<unavailable>, cl=<unavailable>) at ls.cpp:35 [opt]
    frame #10: 0x0000000100198f79 kdb`main(argc=<unavailable>, argv=0x00007fff5fbff830) at main.cpp:198 [opt]
    frame #11: 0x00007fff9d919235 libdyld.dylib`start + 1
```

. It looks like the problematic code might somehow be related to the `system` namespace, since the following commands do not crash:

- `kdb ls user`
- `kdb ls spec`
- `kdb ls dir`

. The command `kdb ls system` on the other hand fails reporting the following error:

```
Elektra/src/plugins/resolver/resolver.c:78:elektraGetResolverHandle: Assertion `pks != ((void*)0)' failed: Unable to retrieve plugin data

Sorry, I crashed by the signal SIGABRT
This should not have happened!

fish: 'kdb ls system' terminated by signal SIGABRT (Abort)
```

. By the way: Do you know why the build job `elektra-mergerequests` always fails?

sanssecours:
Looks like after adding the assertion, additional tests fail on my machine:

```
	  4 - testtool_backendbuilder (OTHER_FAULT)
	 32 - test_kdb.lua (OTHER_FAULT)
	 85 - testscr_check_distribution (Failed)
	 86 - testscr_check_error (Failed)
	 92 - testscr_check_kdb_internal_check (Failed)
	 96 - testscr_check_mount (Failed)
	 99 - testscr_check_real_world (Failed)
	100 - testscr_check_resolver (Failed)
	101 - testscr_check_spec (Failed)
	151 - testkdb_allplugins (OTHER_FAULT)
```

. The files below contain detailed debug output for some of these tests. If someone wants to reproduce these failures: I used the following CMake command to generate the build system:

```sh
cmake ..                                                  \
  -GNinja                                                 \
  -DENABLE_ASAN=OFF                                       \
  -DENABLE_LOGGER=OFF                                     \
  -DENABLE_DEBUG=ON                                       \ 
  -DENABLE_TESTING=ON                                     \
  -DBUILD_PDF=OFF                                         \
  -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5                  \
  -DTOOLS=ALL                                             \
  -DBINDINGS="ALL;-swig_ruby;-swig_python;-swig_python2"  \
  -DPLUGINS="ALL"
```

.

## Logs 
- [testscr_check_distribution.txt](https://github.com/ElektraInitiative/libelektra/files/892943/testscr_check_distribution.txt)
- [testscr_check_error.txt](https://github.com/ElektraInitiative/libelektra/files/892945/testscr_check_error.txt)
- [testtool_backendbuilder.txt](https://github.com/ElektraInitiative/libelektra/files/892944/testtool_backendbuilder.txt)

markus2330:
Thank you, great job! You found a bug! The initializer of the pk handle was done too early, which made no sense in the situation if the resolver is below `system/elektra/modules/resolver_*`. By printing the key name in the assertion it was trivial to find.

PR #1450, opened 2017-04-03
Xerces plugin arrays

e1528532:
# Purpose

The Xerces Plugin had no support for arrays / multiple elements yet (an xml element can contain multiple children with the same name, which gets mapped to elektra arrays with # numbering)

# Checklist

- [X] I added unit tests

@markus2330 this is probably the reason for a lot of failures, totally forgot about that possibility... I've added a new test which uses some arbitrary maven pom file (in xml) i had lying around, to see some "real world" tests, it gets parsed and serialized fine now.
Also in array.c of kdbease there was a mistake IMO, it interpreted -1 as true (as c does it as its not 0) but -1 actually means invalid array key name, so it definately shouldn't be included in the resulting keyset i think. I hope this wasn't there for a reason and actually a bug as i perceived it.

i yet want to try out some more test files tomorrow, like a jenkins configuration file as you mentioned and other snippets - therefore WIP. I'd really love to see it in a better shape for the next release.


markus2330:
jenkins build homepage please

e1528532:
jenkins build homepage please

e1528532:
so i guess since the other issues were not directly related to bugs in the plugin but due to incompatible formats (single root element, internal attributes with invalid names, ...) we can merge this now.

markus2330:
Thank you, the PR looks great!

Can you please rebase? `src/plugins/xerces/xerces/jenkins.xml ` has a conflict.

Btw. (for later!) you can also use gtest like already used for other plugins.

e1528532:
ah you added it to the master and i didn't see it and also added it, therefore the conflict. rebase it now it should be fine.

markus2330:
Thank you!

PR #1453, opened 2017-04-05
Plugins: Add Status Values for Plugin Functions

sanssecours:
# Purpose

This pull request adds the following macros for the return values of plugin functions:

- `ELEKTRA_PLUGIN_STATUS_ERROR`: An error occurred inside the plugin function
- `ELEKTRA_PLUGIN_STATUS_SUCCESS`: Everything went fine and the function **did not** change the given keyset
- `ELEKTRA_PLUGIN_STATUS_SUCCESS_KEYSET_MODIFIED`: Everything went fine and the function did modify the given keyset

.

# Checklist

- [x] I updated the documentation.

### DIFF
--- a/doc/CODING.md
+++ b/doc/CODING.md
@@ -76,7 +76,7 @@ Thus please use following techniques (in order of preference):
    `ELEKTRA_LOG ("formatted text to be printed according log filters", ...)`
 
    There are four log levels (ERROR is reserved for aborts within `ELEKTRA_ASSERT`):
-   - ELEKTRA_LOG_WARNING, something critical that should be shown the user (e.g. API misuse), see #ELEKTRA_LOG_LEVEL_WARNING
+   - ELEKTRA_LOG_WARNING, something critical that should be shown to the user (e.g. API misuse), see #ELEKTRA_LOG_LEVEL_WARNING
    - ELEKTRA_LOG_NOTICE, something important developers are likely interested in, see #ELEKTRA_LOG_LEVEL_NOTICE
    - ELEKTRA_LOG, standard level gives information what the code is doing without flooding the log, see #ELEKTRA_LOG_LEVEL_INFO
    - ELEKTRA_LOG_DEBUG, for less important logs, see #ELEKTRA_LOG_LEVEL_DEBUG
--- a/doc/help/elektra-algorithm.md
+++ b/doc/help/elektra-algorithm.md
@@ -117,10 +117,10 @@ The synopsis of the function is:
 
 The user passes a key set, called `returned`.
 If the user invokes `kdbGet()` the first time, he or she will usually
-pass an empty key set.	If the user wants to update the application's
+pass an empty key set. If the user wants to update the application's
 settings, `returned` will typically contain the configuration of the
 previous `kdbGet()` request.  The `parentKey` holds the information
-below which key the configuration should be retrieved.	The `handle`
+below which key the configuration should be retrieved. The `handle`
 contains the data structures needed for the algorithm, like the `Split`
 and the `Trie` objects.
 
@@ -135,7 +135,7 @@ A backend may yield keys that it is not responsible for.
 It is not possible for a backend to know that another backend has been
 mounted below and the other backend is now responsible for some of the
 keys that are still in the storage.  Additionally, plugins are not able
-to determine if they are responsible for a key or not.	Consequently, it
+to determine if they are responsible for a key or not. Consequently, it
 can happen that more than one backend delivers a key with the same name.
 
 `kdbGet()` ensures that a key is uniquely identified by its name.
@@ -197,7 +197,7 @@ and returns zero.
 
 Now we know which backends do not need an update.  For these backends, the
 previous configuration from `returned` is appointed from to the key sets
-of the `Split` object.	The algorithm will not set the *syncbits*
+of the `Split` object. The algorithm will not set the *syncbits*
 of the `Split` object for these backends because the storage of the
 backends already contains up-to-date configuration.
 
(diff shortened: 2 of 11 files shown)


markus2330:
`ELEKTRA_PLUGIN_STATUS_SUCCESS` not changing the keyset sounds unintuitive to me. What about `ELEKTRA_PLUGIN_STATUS_NOP` instead? (And `ELEKTRA_PLUGIN_STATUS_SUCCESS` for the real success return value?)

And please use doxygen docu for the macros.

markus2330:
And thanks for the PR :smile:

sanssecours:
> `ELEKTRA_PLUGIN_STATUS_SUCCESS` not changing the keyset sounds unintuitive to me. What about `ELEKTRA_PLUGIN_STATUS_NOP` instead? (And `ELEKTRA_PLUGIN_STATUS_SUCCESS` for the real success return value?)

That would mean that a function like `elektraPluginOpen` would always return `ELEKTRA_PLUGIN_STATUS_NOP` if it is successful (the function does not change a keyset after all)? In my opinion `NOP` reminds me too much of the assembler directive with the same name. The only code I would accept returning `ELEKTRA_PLUGIN_STATUS_NOP` would look something like this:

```c
int elektraPluginOpen (Plugin * handle ELEKTRA_UNUSED, Key * errorKey ELEKTRA_UNUSED)
{
  asm ("nop");
  asm ("nop");
  asm ("nop");
  asm ("nop");
  return ELEKTRA_PLUGIN_STATUS_NOP; // 😊
}
```

.

> And please use doxygen docu for the macros.

I already did that:

doxygen

, or do you mean something different?

markus2330:
> That would mean that a function like elektraPluginOpen would always return ELEKTRA_PLUGIN_STATUS_NOP if it is successful

No, actually the `SUCCESS` is standard everywhere. Only for `get` and `set`, and currently only for the resolver, the `NOP` is relevant.

You are right that this should be documented better and your macros are a good place for this docu.

> In my opinion NOP reminds me too much of the assembler directive with the same name

Other suggestions are welcomed, `NOP` is short.

What about `STOP` or `BREAK`? It only stops/breaks at certain positions, though. So it might be misleading. Have a look in `kdb.c` and see `doc/help/elektra-algorithm.md`, especially the parts about resolver.

The `KEYSET_MODIFIED` is misleading, too---at least for resolvers. (Maybe the docu is misleading and the name is perfectly consistent with the misleading docu.) 

> I already did that

Ahh nice, I am so used to multi-line doxygen comments that I overlooked the two stars.

You could also add pointers to these new macros in the docu for the return values of `elektraDocGet`,`elektraDocSet`,...

sanssecours:
> No, actually the `SUCCESS` is standard everywhere. Only for `get` and `set`, and currently only for the resolver, the `NOP` is relevant.

Thank you for the information.

> You are right that this should be documented better…

I agree 😀.

> Other suggestions are welcomed, `NOP` is short.

> What about `STOP` or `BREAK`? It only stops/breaks at certain positions, though. So it might be misleading. 

The latest commit uses `ELEKTRA_PLUGIN_STATUS_SUCCESS` for the return value `1` and `ELEKTRA_PLUGIN_STATUS_SUCCESS_NO_CHANGE` for the return value `0`. Hope these names are okay from your point of view. While `ELEKTRA_PLUGIN_STATUS_SUCCESS_NO_CHANGE` might be quite long, I like it, since it is quite descriptive in my opinion.

> Have a look in `kdb.c` and see `doc/help/elektra-algorithm.md`, especially the parts about resolver.

I looked at (the documentation) in both files. Unfortunately I was unable to find anything about the return values of resolver plugins in anyone of them. Could you please write down a single sentence or a few words that describes the return value `0` from the perspective of a resolver plugin? I would then use this information to update the current documentation of `ELEKTRA_PLUGIN_STATUS_SUCCESS_NO_CHANGE`:


```c
/** Everything went fine and the function **did not** modify the given keyset */
#define ELEKTRA_PLUGIN_STATUS_SUCCESS_NO_CHANGE 0
```

> You could also add pointers to these new macros in the docu for the return values of `elektraDocGet`,`elektraDocSet`,...

For now I think it is enough if we specify the information in `plugins.md` and most importantly in `template.c`. Since you mentioned `elektraDocGet` and `elektraDocSet` I checked the documentation of these functions in `kdb.h`. Looks like the documentation of the return value is partially incorrect too? As far as I can tell from your description every plugin function should return `1` on success unless it did not change anything significant, in which case it should return `0`? I updated the documentation (according to my current state of knowledge) in commit 4be4d708.

markus2330:
Yes, the name now is much better! Not so ideal is that it has a `NO`, and that it starts with the same name `SUCCESS` (problematic for completion). But I have to admit that `NOP` and `BREAK` are also not so ideal.

The issue we are having is that 0 and 1 has different meanings in different scenarios. For `elektraTemplateCheckConfig` your name+description fits perfectly.

In the resolver `kdbGet` case , however, the description is not so ideal: the resolver checks if the configuration files were updated. So `NO_CHANGES` is still okay, but the changes are about the configuration files, not about the KeySet.

Please retick "ready to merge" if everything was addressed. From my point of view we can merge it now. The other docu issues can be fixed in a separate PR. The many small fixes are really nice!

sanssecours:
> Yes, the name now is much better! Not so ideal is that it has a `NO`, and that it starts with the same name `SUCCESS` (problematic for completion). But I have to admit that `NOP` and `BREAK` are also not so ideal.

I changed the name to `ELEKTRA_PLUGIN_STATUS_NO_UPDATE` in the latest commit. There is still a `NO` in there though.

> In the resolver `kdbGet` case , however, the description is not so ideal: the resolver checks if the configuration files were updated. So `NO_CHANGES` is still okay, but the changes are about the configuration files, not about the KeySet.

The latest name change should address this problem. I also updated the description. Hope the new name reflects both situation well enough.

markus2330:
Thank you for your improvement!

PR #1464, opened 2017-04-10
Address Sanitizer: Fix Warnings

sanssecours:
# Purpose

This pull request fixes some warnings reported by the address sanitizer. 

# Checklist

### DIFF
--- a/src/plugins/list/list.c
+++ b/src/plugins/list/list.c
@@ -362,7 +362,7 @@ int elektraListGet (Plugin * handle, KeySet * returned, Key * parentKey)
 	ksRewind (pluginKS);
 	int ret = runPlugins (pluginKS, placements->modules, placements->plugins, ksDup (config), returned, parentKey, GET, ksNext);
 	placements->getCurrent = ((++currentPlacement) % getEnd);
-	while (!placements->getPlacements[currentPlacement])
+	while (currentPlacement < getEnd && !placements->getPlacements[currentPlacement])
 	{
 		placements->getCurrent = ((++currentPlacement) % getEnd);
 	}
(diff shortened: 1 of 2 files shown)


sanssecours:
This update should be ready. If I should change anything else, then please just comment below. Thank you.

markus2330:
Thank you for taking care of such issues!

@tom-wa @e1528532 Can you take a look?

markus2330:
@tom-wa Can you take a look?

tom-wa:
thanks for finding/fixing that bug.
would that also work if kdbGet gets called twice ?

markus2330:
@tom-wa Thanks for looking at it!

Which problem do you think could occur if `kdbGet` is called twice?

tom-wa:
i'm currently trying to wrap my head around it. the loop runs until the next placement is found. if the last placement is reached it would go straight back the the first placement (% getEnd) but with the placement < getEnd condition that might never happen. 

maybe `while (!placements->getPlacements[currentPlacement % getEnd])` would be a better solution ? an incorrectly configured list plugin would result in an endless-loop though. 

what do you think ?

markus2330:
Actually it is enough if you confirm that it does not break anything that worked before.

Does `while (!placements->getPlacements[currentPlacement % getEnd])` also yield no sanitizer problems? What would be the difference in behavior?

> an incorrectly configured list plugin would result in an endless-loop though.

Better to avoid that (if otherwise there is no difference).

tom-wa:
>Actually it is enough if you confirm that it does not break anything that worked before. 

thats what i'm currently trying to confirm. in my original implementation (which i have to admit had some bugs) it would loop % lastplacement, so after the last placement is reached, it would begin with the first one again. and i'm not sure if behaviour is still the same

markus2330:
Take your time. Can you add some unit tests so that we can be more confident for the next changes?

tom-wa:
the problem with the endless loops with a misconfigured list plugin is present anyway, i only noticed it now
i think something like 
```
	placements->errCurrent = ((++currentPlacement) % errEnd);
        ErrPlacements counter = 0;
	while (!placements->errPlacements[currentPlacement % errEnd] && counter++ < errEnd)
	{
		placements->errCurrent = ((++currentPlacement) % errEnd);
	}
```
would fix both issues for now until someone (me ? ) adds a check if the plugin actually has any placements for kdb{Get,Set,Error} when List{Get,Set,Error} are called.

markus2330:
@tom-wa Thanks for looking into it! Do you have time to prepare a PR, ideally with test cases?

PR #1468, opened 2017-04-12
rename examples: cut -> ksCut, set -> set_key

BernhardDenner:
# Purpose

rename examples, since they result in 'cut' and 'set' binaries
(conflict with unix tool cut and shell builtin set)

# Checklist

needs to be checked.

- [ ] I added unit tests


markus2330:
Thanks for creating the PR!

Did you use `git mv` for ksCut? GitHub does not display it as renamed.

Maybe we should even make the camel case vs. underscores between the examples consistent, too?

And we could write a shell recorder test which calls all of the examples.

BernhardDenner:
Yes, I've used git mv for both source files, but I think, since I also fixed the intention in ksCut, git doesn't track this change as simple rename, instead of a "full" change. Good to know, that such things should go in separate commits ;)

I've used ksCut, since it demonstrates the functionality of the `ksCut` function. For the `set_key`, I didn't actually know, how to rename it, since there is already a `kdbSet` example and both do basically the same (kdbset also does proper error checking).

Should I rename all examples to a camel case variant?

markus2330:
Thank you!

> Should I rename all examples to a camel case variant?

No, the camel case is used less, so we should get rather rid of it. (And all camel case written applications seem to be examples).

Issue #1400, opened 2017-03-03
escaping in (meta)keynames


markus2330:
Issue: some storage plugins are not able to write every keyname (or metakeyname).

(Only brain storming at the moment)

Writte in #1380 by @e1528532:
> The only issue that remains is that xerces doesn't handle illegal element & attribute names. I haven't really found a lot about that, only microsoft seems to offer something for that case: https://msdn.microsoft.com/en-us/library/35577sxd(v=vs.80).aspx
I think their approach should be fine basically, and if we can extend the rename plugin so it changes key names (also of metakeys) to this escaped version (and vice versa) we can also handle forbidden characters in xml element names. The only problem i can think of is that if someone calls a key like _x00010000_ for a reason (seems to be rather unlikely for me, but everything is possible ;) ) as it would also be unescaped then.

We also have this issue for JSON and other plugins. There is already the hexcode plugin which could be configured to use `_` as escaping character, but it only applies to values. The rename plugin supports renaming of key names, but has no escaping strategies.

TODO: Design general strategy for how to handle renaming in (meta)keynames.

Should we group together by escaping algorithm or by what is renamed (key, value, metakey)?

e1528532:
Regarding the grouping:

For simplicity, we might just specify we want to escape for xml and then it escapes the key and metakey names automatically, leaving the rest like it is. But this kind of escaping is only useful for the xerces plugin which uses keynames as element names, it wouldn't be necessary for the old xmltool plugin which also maps keynames as attributes. So i don't know if its a good idea to call the escaping algorithmn "xml" then as its only required if we serialize keysets to xml with the xerces plugin.

If we group both by escape algorithmn and what is renamed we would get the most flexible startegy. For instance, we can say something like escape key names as xml and escape metakey names as xml too. Now imagine another plugin or use case which maybe also works with xml where you want to use key values as element names for whatsoever reason. If we offer both grouping possibilities in those cases you could still use the rename plugin and tell it to escape key values as xml now. 

So i guess its rather a question of simplicity (fixed xml escaping algorithm always only escaping key names and metakey names) vs flexibility (both grouping options possible).

markus2330:
Yes, having an extra plugin that is only useful for xerces does not make much sense.

The ccode plugin allows you to specify which character to use for escaping and what to escape. simpleini makes use of this escaping by stating in the contract which characters should be escaped.

[...]

PR #1469, opened 2017-04-14
Plugin: “Minimal” INI Storage Plugin

sanssecours:
# Purpose

This pull request adds a very simple INI plugin, mainly intended as replacement for SimpleINI on platforms that do not use `glibc`, such as macOS.

This pull request also contains

- the ASAN fixes from pull request #1464,
- a simplistic function for string replacement called `elektraReplace`,
- various fixes for the (Markdown) Shell Recorder,
- and other minor fixes.

# Checklist

- [x] I double checked all commit messages.
- [x] I added unit tests.
- [x] The plugin contains a ReadMe file describing the current capabilities. I also added Doxygen documentation for the function `elektraReplace`. 
- [x] I added code comments, logging, and assertions.
- [x] The meta data of the plugin is up to date.


src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+	ksRewind (keySet);
+	while ((key = ksNext (keySet)) != 0)
+	{
+		const char * name = elektraKeyGetRelativeName (key, parentKey);
+		ELEKTRA_LOG_DEBUG ("Write mapping “%s=%s”", name, keyString (key));
+
+		char * escapedName = escape (name);
+		char * escapedKey = escape (keyString (key));
+		fprintf (file, "%s=%s\n", escapedName, escapedKey);
+		free (escapedName);
```
markus2330: Also use `elektraFree`.

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+	if (!destination)
+	{
+		ELEKTRA_LOG_WARNING ("Could not open file “%s” for writing: %s", keyString (parentKey), strerror (errno));
+		ELEKTRA_SET_ERROR_GET (parentKey);
+		errno = errorNumber;
+		return ELEKTRA_PLUGIN_STATUS_ERROR;
+	}
+	writeFile (destination, returned, parentKey);
+
+	fclose (destination);
```
markus2330: Again, `fclose` can fail.

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+	FILE * destination = fopen (keyString (parentKey), "w");
+
+	if (!destination)
+	{
+		ELEKTRA_LOG_WARNING ("Could not open file “%s” for writing: %s", keyString (parentKey), strerror (errno));
+		ELEKTRA_SET_ERROR_GET (parentKey);
+		errno = errorNumber;
+		return ELEKTRA_PLUGIN_STATUS_ERROR;
+	}
+	writeFile (destination, returned, parentKey);
```
markus2330: Writing files can fail (disk full), no return value here?

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+	Key * key;
+	ksRewind (keySet);
+	while ((key = ksNext (keySet)) != 0)
+	{
+		const char * name = elektraKeyGetRelativeName (key, parentKey);
+		ELEKTRA_LOG_DEBUG ("Write mapping “%s=%s”", name, keyString (key));
+
+		char * escapedName = escape (name);
+		char * escapedKey = escape (keyString (key));
+		fprintf (file, "%s=%s\n", escapedName, escapedKey);
```
markus2330: fprintf to file can fail (disk full)

src/plugins/mini/Examples/read.ini
```diff
@@ -0,0 +1,15 @@
+keyWithoutLeadingWhitespace = 	valueWithLeadingWhiteSpace
```
markus2330: Please do not name the folder `Examples` but `mini` (like the plugin is named). Otherwise we have confusing directories or even get a clash when installing these files.

tests/ctest/test_utility.c
```diff
@@ -96,6 +122,7 @@ int main (int argc, char ** argv)
 	test_elektraLskip ();
 	test_elektraRstrip ();
 	test_ElektraStrip ();
+	test_ElektraReplace ();
```
markus2330: uniform small `e`? (As the function is also called like this?)

sanssecours: Oh, I did not even notice this inconsistency. I already fixed it in my local copy of the repository.

src/plugins/mini/Examples/read.ini
sanssecours: I moved the files to the main plugin folder. I hope I never have to name a directory just like its parent folder, since I think using the same name for a subfolder is `/pretty useless/pretty useless/pretty useless`.

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+
+	*equals = '\0';
+
+	char * name = unescape (elektraRstrip (pair, NULL));
+	char * value = unescape (elektraLskip (equals + 1));
+
+	Key * key = keyNew (keyName (parentKey), KEY_END);
+	keyAddName (key, name);
+	keySetString (key, value);
+	ELEKTRA_LOG_DEBUG ("Name:  “%s”", keyName (key));
```
sanssecours: I think that mainly depends on the preference of the user and the length of key and value. If both are very long, then the current solution should be better. If they are very short, it might not make sense to split them into two lines. 

Anyway, I like it this way, so I will leave it as is for now. If you have a strong preference for a one line solution, then please leave a comment below and tell me why the current solution is horrible 😊.

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+
+	if (*pair == '\0')
+	{
+		return;
+	}
+
+	char * equals = findUnescapedEquals (pair);
+	if (*equals == '\0' || equals == pair)
+	{
+		fprintf (stderr, "Ignored line %lu since “%s” does not contain a valid key value pair\n", lineNumber, pair);
```
sanssecours: I replaced `fprintf` with `ELEKTRA_LOG_WARNING`.

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+	while (*current != '\0' && ((*current != '#' && *current != ';') || (before && *before == '\\')))
+	{
+		before = current;
+		current++;
+	}
+	*current = '\0';
+	return line;
+}
+
+static inline char * findUnescapedEquals (char * text)
```
sanssecours: Do I really have to document **private** functions? This function contains about 5 lines of code! All of the other functions are also relatively short too and should be more or less readable as far as I can tell. At least [OClint](http://oclint.org) does not complain about any of them.

> What happens if no unescaped equal was found?

The function will return a pointer to the `\0` character at the end of the string.

src/plugins/mini/Examples/read.ini
markus2330: > I moved the files to the main plugin folder.

The installation of the files won't work then.

> using the same name for a subfolder is /pretty useless

We could also rename all of these subfolders to `examples` (please lower case for consistency) and make the installation routine `cmake/Modules/LibAddPlugin.cmake` line 91 rename the configs.

But please in a different PR, for now please name it according the convention.

Please also add `INSTALL_TEST_DATA` so that your test can be executed after Elektra was installed.

src/plugins/mini/README.md
```diff
@@ -0,0 +1,122 @@
...
+#> 🦄
+
+# Undo modifications to the key database
+kdb rm -r /examples/mini
+kdb umount /examples/mini
+```
+
+### Escaping
+
+As with most configuration file formats, some characters carry special meaning. In the case of the `mini` plugin that character is the `=` sign, which separates keys from values. In most cases you do not need to care about the special meaning of the `=` sign though, since the plugin handles escaping and unescaping of `=` for you. The following example shows this behaviour.
```
sanssecours: > A truly minimalistic plugin would let other plugins escape the `=`.

Okay. Which plugin can I use for that purpose and how can I load it every time the `mini` plugin is used, without any interaction from the user?

src/plugins/mini/mini.c
markus2330: If the function does anything that the name does not already say, it would be better if we have it documented.

Log output and assertions also count as docu.

For private functions you obviously do not need to be as verbose as in `elektraReplace`.

> The function will return a pointer to the \0 character at the end of the string.

Please simply add this as your documentation.

src/plugins/mini/README.md
markus2330: For example, `src/plugins/simpleini/simpleini.c` demonstrates how it can be done: You need the `config/needs/escape`, `config/needs/chars/3D` and `infos/needs = code` in your contract.

Thus the code plugins are quite untested right now, a separate PR might be better. Then you can also fix the TCL issue #1194.

> without any interaction from the user

`kdb mount` will automatically add needed plugins. In the test cases though it won't be done thus the interface used is too low-level. (The tests in `tests/kdb/` use the mount interface, here needed plugins would be added, too.)

markus2330: Another way to test it would be `src/tools/kdb/test.cpp` which is executed for every plugin in `tests/shell/check_kdb_internal_suite.sh`.

markus2330: It would be great if these places are documented in `doc/TESTING.md` :smile:

src/plugins/mini/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+[INI]: https://en.wikipedia.org/wiki/INI_file
+
+## Examples
+
+### Basic Usage
+
+The following example shows basic usage of the `mini` plugin.
+
+```sh
```
sanssecours: I updated the documentation in commit 4cc4cc3e.

src/plugins/README.md
```diff
@@ -101,6 +101,7 @@ Only suited for import/export:
 - [xmltool](xmltool/) uses XML (in Elektra's XML schema).
 - [simpleini](simpleini/) line-based key-value pairs with configurable
   format (without sections)
+- [mini](mini/) minimalistic dependency free plugin for the INI  file format.
```
sanssecours: > But shouldn't we call such plugins `properties` (as in https://en.wikipedia.org/wiki/.properties) instead?

That would make sense, although I suspect that more people know about the INI format, than about `.properties` files. The syntax of `.properties` looks a bit different compared to the standard INI format too. As far as I can tell, `mini` does not support any of these two formats properly.

> Or do you have in mind to add sections later?

No.

src/libs/utility/text.c
```diff
@@ -87,11 +89,84 @@ char * elektraRstrip (char * const start, char ** end)
...
+	char * result = malloc ((textLength + numberOccurrences * (replacementLength - patternLength) + 1) * sizeof (char));
+	if (result == NULL)
+	{
+		return NULL;
+	}
+
+	char const * current = text;
+	char const * before = text;
+	char * destination = result;
+	size_t length;
```
sanssecours: Oh, I totally forgot about this variable after I simplified the code… Thanks for the hint.

src/libs/utility/text.c
```diff
@@ -87,11 +89,84 @@ char * elektraRstrip (char * const start, char ** end)
...
+			"The Parameter `pattern` contains an empty string. Please provide a string with a minimum length of 1 character.");
+	ELEKTRA_ASSERT (replacement != NULL, "The Parameter `replacement` contains `NULL` instead of a valid string.");
+
+	ssize_t patternLength = strlen (pattern);
+	ssize_t numberOccurrences = occurrences (text, pattern, patternLength);
+	ELEKTRA_LOG_DEBUG ("Found “%lu” occurrences of “%s” in “%s”", numberOccurrences, pattern, text);
+
+	ssize_t replacementLength = strlen (replacement);
+	ssize_t textLength = strlen (text);
+	char * result = malloc ((textLength + numberOccurrences * (replacementLength - patternLength) + 1) * sizeof (char));
```
sanssecours: I just updated the code accordingly in commits dec47a42 and b88c034e.

scripts/kdb.fish
```diff
@@ -241,8 +241,8 @@ function __fish_kdb_print_resolver_plugins -d 'Print a list of available resolve
 end
 
 function __fish_kdb_print_storage_plugins -d 'Print a list of available storage plugins'
-    set -l formats constants crypto_botan crypto_gcrypt crypto_openssl desktop dpkg dump fcrypt hosts line ini json ni passwd regexstore
-    set -l formats $formats simpleini simplespeclang tcl xmltool uname
+    set -l formats constants crypto_botan crypto_gcrypt crypto_openssl desktop dpkg dump fcrypt hosts line ini json mini ni passwd
+    set -l formats $formats regexstore simpleini simplespeclang tcl xmltool uname
```
sanssecours: That sounds like a good idea, although I would prefer we just add the possibility to list only storage plugins to `kdb list`. We could for example add an optional argument (`storage`, `resolver`, …) to the command. This way it would still be possible to use `kdb.fish` directly.

markus2330: In `tests/shell/include_common.sh.in` there is also the code to get all storage plugins. Basically you iterate over all PLUGINS and check if its a storage using:

```
test ! "x`"$KDB" info $PLUGIN provides 2>/dev/null`" = "xstorage"
```

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+		ELEKTRA_LOG_DEBUG ("Read Line %lu: %s", lineNumber, line);
+		parseLine (line, lineNumber, keySet, parentKey);
+	}
+
+	elektraFree (line);
+
+	if (!feof (file))
+	{
+		ELEKTRA_LOG_WARNING ("Did not reach end of configuration file “%s”", keyString (parentKey));
+		ELEKTRA_SET_ERROR (ELEKTRA_ERROR_NOEOF, parentKey, strerror (errno));
```
sanssecours: I just fixed this issue in commit de1d7e29 and 1ff1bca1.

src/plugins/mini/mini.c
sanssecours: I addressed  this problem in commit b88c034e.

markus2330: It is enough if you do :+1: for comments that were addressed and/or mark as "ready to merge" if all issues were addressed (or created as separate issues to be resolved later).

scripts/kdb.fish
sanssecours: I updated the code in commit cf1a3af9. The new code looks nicer:

```fish
for plugin in (kdb list)
  kdb info $plugin provides ^/dev/null | string match -qr -- storage
  and echo $plugin
end
```

. However, the runtime of `__fish_kdb_print_storage_plugins` is even worse than before (~ 200ms). Running `kdb list` takes about 200ms on my machine. Calling `kdb info` for all of the 83 plugins takes about 600ms. Altogether it takes approximately 800ms to display all available storage plugins 😴.

src/plugins/mini/mini.c
sanssecours: I addressed this issue in commit fb33c61f.

src/plugins/mini/mini.c
sanssecours: This problem should be fixed with commit fb33c61f.

src/plugins/mini/mini.c
sanssecours: I added an error check in commit 724100cf.

src/plugins/mini/mini.c
```diff
@@ -0,0 +1,218 @@
...
+	{
+		ELEKTRA_LOG_WARNING ("Could not open file “%s” for reading: %s", keyString (parentKey), strerror (errno));
+		ELEKTRA_SET_ERROR_GET (parentKey);
+		errno = errorNumber;
+		return ELEKTRA_PLUGIN_STATUS_ERROR;
+	}
+
+	int status = parseINI (source, returned, parentKey);
+
+	fclose (source);
```
sanssecours: Commit fb33c61f should fix this problem.

src/plugins/mini/mini.c
sanssecours: I addressed this issue in commit ebdf963b.

src/plugins/mini/mini.c
```diff
@@ -60,6 +60,8 @@ static inline char * findUnescapedEquals (char * text)
 	{
 		before = equals++;
 	}
+	/* The function will return a pointer to the `\0` character at the end of the string
+	   if it did not find any unescaped `=` character. */
```
markus2330: Thank you, looks nice! Please put it above the function. Please also document (via asserts) preconditions (`text != 0`).

src/plugins/mini/Examples/read.ini
sanssecours: > The installation of the files won't work then.

Oh, I did not know that.

> Please also add `INSTALL_TEST_DATA` so that your test can be executed after Elektra was installed.

Okay, I added the attribute in commit 6ca5e40d.

scripts/kdb.fish
markus2330: Sorry for giving a bad tip, I didn't know that performance was an issue.
With 478c7c3c71de076ff2b5032de454c4487dbd4c1d you can use `kdb list storage` which does what you want. On my computer it takes 0,04s (measured with time(1)).

sanssecours: Thank you for adding the new parameter to `kdb list`. I already used it in commits 675f02c3 and 63d20b36. It looks like your machine is 5 times faster than mine. Can you think of a reason why there is such a big difference in performance? I already disabled the address sanitizer, debugging and logging…

markus2330: We would need to profile it to find out where the bottleneck is.

By design `kdb list` needs to load every plugin (even if it does not output all of them). On my computer about 2/3 of `kdb list storage` is spent loading all modules (kdb::tools::Modules::load, and 1/3 in `dlopen`) and afaik module loading is implemented completely different on mac os x. But that is a wild guess, we would need profile data from your machine to find out.

Is it still a usability problem? 200ms (5x 40ms) should still be barely noticeable?

src/plugins/mini/mini.c
sanssecours: I added Doxygen documentation and assertions in commit 7c987de3.

markus2330: Thank you, very nicely done! mini is now Elektra's best documented plugin!

scripts/kdb.fish
sanssecours: > Is it still a usability problem? 200ms (5x 40ms) should still be barely noticeable?

I would say 200ms is noticeable but still okay. However, as I would not consider my machine especially slow (i7-4960HQ, 1TB SSD, 16GB RAM), the user experience of other Elektra users on macOS might be considerably worse.

src/plugins/mini/README.md
sanssecours: I’m all for moving complexity out of `mini` into other plugins. Especially since I realized that the current version of mINI does not handle unescaped comment characters inside key names and values properly. I also forgot to check if the memory allocation inside `elektraReplace` was successful.

> `kdb mount` will automatically add needed plugins. In the unit test cases though it won't be done thus the plugin interface used in `PLUGIN_OPEN` is too low-level.

That is not a big problem. I will just test the escaping mechanism in Markdown Shell Recorder tests then.

> You need the `config/needs/escape`, `config/needs/chars/3D` and `infos/needs = code` in your contract.

I recently started to add this changes to the plugin. One thing that I noticed was that the `ccode` plugin currently only escapes key values, while mINI should also escape key names. Do you know an easy solution to fix this problem?

markus2330: > Do you know an easy solution to fix this problem?

Unfortunately, there is no easy solution. Escaping of keys is currently not implemented (#1400). But it affects many plugins (simpleini, tcl, later also fstab, yajl, xml) thus it is an investment beyond mini.

But I think we should first finish up this PR for now (simply document the limitation) and extend ccode later to also handle key names.

scripts/kdb.fish
markus2330: We could improve it by caching the contracts of the plugins (or even the results of `kdb list`). But it is quite an effort for tab completion of plugin names? The other parts of tab completion shouldn't be effected, or are they?

For slower computers the workaround is to install fewer plugins.

sanssecours: > But it is quite an effort for tab completion of plugin names?

Yes, I think so too. We should not make things more complicated if we do not need to. In my opinion the run-time of `kdb list` on Linux looks already good. Would be nice if we could have similar performance on macOS too though.

> The other parts of tab completion shouldn't be effected, or are they?

Not as far as I can tell.

src/plugins/mini/README.md
sanssecours: > But I think we should first finish up this PR for now (simply document the limitation)…

I changed the escaping code in the commit eb758190. The plugin now also escapes and unescapes  `;`, `#` and `\` characters in key values as well. I described the limitations about unusable characters in key names [here](https://github.com/ElektraInitiative/libelektra/pull/1469/commits/35341f5a9243d540b9f65b6675a47832497238e7#diff-0c1742553c4f791d84d0b20d5dba2052R102). As far as I can tell, the pull request should be ready now.

markus2330:
Thank you, great job!

Together with the improvements in escaping using code plugins, we should also find a generic solution to strip whitespaces.

Btw. we should also recommend the rename plugin with `tolower` transformation thus INI is usually defined to be case-insensitive.

Issue #1406, opened 2017-03-06
Documentation Compilation broken


e1528532:
# Describe what you wanted to do

I wanted to compile the current version including documentation.

# Describe what you expected
I have all necessary tools installed and it always worked in the past, so i expected that this will be the case again.

# Describe what actually happened

ronn complains about invalid byte sequences in US-ASCII. On a closer look it seems to dislike the new way we write Elektra's as "Elektra´s". It doesn't like the ´ character. But we could use the ` or the ' character instead which is valid ascii, and convert the files back to ascii instead of utf-8. Then it worked again.

# System Information

- Elektra Version: master
- ronn 0.7.3
- macOS Sierra

# Further Log Files and Output

```
[  1%] Generating elektra-faq.7
/Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:330:in `split': invalid byte sequence in US-ASCII (ArgumentError)
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:330:in `markdown_filter_heading_anchors'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:299:in `process_markdown!'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:210:in `markdown'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:290:in `input_html'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:305:in `process_html!'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:215:in `html'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:251:in `to_html_fragment'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:227:in `to_roff'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/lib/ronn/document.rb:221:in `convert'
        from /Library/Ruby/Gems/2.0.0/gems/ronn-0.7.3/bin/ronn:202:in `block (2 levels) in <top (required)>'
[...]

PR #1477, opened 2017-04-24
Ronn: Use UTF-8 as Input Encoding

sanssecours:
# Purpose

This commit should finally fix the issue mentioned in commit 214898ed:

> Before this change building the man pages would not work if the current locale only allowed ASCII characters. This commit addresses issue #1406.

. The problem with the old fix was that CMake only sets environment variables at configuration time. Now the encoding is properly set at execution time, just before `ronn` generates the man pages.

This fix will only work on Unix like systems though. We could use `cmake -E env` instead  of `export` to address this problem. However, as  far as I can tell, `cmake -E env` is [only available since CMake 3.2](http://stackoverflow.com/questions/35029277/how-to-modify-environment-variables-passed-to-custom-cmake-target#35032051).

# Checklist

- [x] I double checked the commit message

### DIFF
--- a/cmake/Modules/LibAddMacros.cmake
+++ b/cmake/Modules/LibAddMacros.cmake
@@ -529,7 +529,7 @@ if (BUILD_DOCUMENTATION AND RONN_LOC) # disable function when RONN_LOC is not se
 	add_custom_command(
 		OUTPUT ${OUTFILE}
 		DEPENDS ${MDFILE}
-		COMMAND ${RONN_LOC}
+		COMMAND export RUBYOPT="-Eutf-8" && ${RONN_LOC}
 		ARGS -r --pipe ${MDFILE} > ${OUTFILE}
 		)
 	add_custom_target(man-${NAME} ALL DEPENDS ${OUTFILE})
--- a/doc/help/CMakeLists.txt
+++ b/doc/help/CMakeLists.txt
@@ -1,8 +1,6 @@
 find_program (RONN_LOC ronn)
 
 if (RONN_LOC)
-	SET (ENV{RUBYOPT} "-E utf-8")
-
 	add_custom_target (man ALL)
 
 


cmake/Modules/LibAddMacros.cmake
```diff
@@ -529,7 +529,7 @@ if (BUILD_DOCUMENTATION AND RONN_LOC) # disable function when RONN_LOC is not se
 	add_custom_command(
 		OUTPUT ${OUTFILE}
 		DEPENDS ${MDFILE}
-		COMMAND ${RONN_LOC}
+		COMMAND export RUBYOPT="-Eutf-8" && ${RONN_LOC}
```
markus2330: Is maybe `RUBYOPT="-Eutf-8" ${RONN_LOC}` more portable? Or even better would be`${CMAKE_COMMAND} -E env RUBYOPT="-Eutf-8" ${RONN_LOC}` but it is not available in cmake 3.

sanssecours: > Is maybe `RUBYOPT="-Eutf-8" ${RONN_LOC}` more portable?

I already tried that. Setting the environment variable only for the command does not work.

> Or even better would be `${CMAKE_COMMAND} -E env RUBYOPT="-Eutf-8" ${RONN_LOC}` but it is not available in cmake 3.

This works, but as far as I know only in [**CMake 3.2.**](http://public.kitware.com/Bug/view.php?id=5145#c36770) or newer. Since some of the build servers use old versions of CMake (2.x) I decided against using `cmake -E env`.

markus2330: Ok, thank you! I think we can merge this. Or do you want to wait for @KurtMi to try?

sanssecours:
> Tested it. It works!

Thank you!

cmake/Modules/LibAddMacros.cmake
sanssecours: > I think we can merge this.

Since Kurt already successfully tested the pull request, I think so too.

markus2330:
Thanks for fixing and testing!

PR #1479, opened 2017-04-25
Utility: Use Standard C String Copy Function

sanssecours:
# Purpose

Before this change Elektra would not compile on MinGW, since `stpncpy` is not part of the ISO C standard.

# Checklist

### DIFF
--- a/src/libs/utility/text.c
+++ b/src/libs/utility/text.c
@@ -159,8 +159,12 @@ char * elektraReplace (char const * const text, char const * const pattern, char
 	char * destination = result;
 	while ((current = strstr (current, pattern)) != NULL)
 	{
-		destination = stpncpy (destination, before, current - before);       //! OCLint (False Positive: Constant Conditional OP)
-		destination = stpncpy (destination, replacement, replacementLength); //! OCLint (False Positive: Constant Conditional OP)
+		size_t length = current - before;
+		strncpy (destination, before, length); //! OCLint (False Positive: Constant Conditional OP)
+		destination += length;
+
+		strncpy (destination, replacement, replacementLength); //! OCLint (False Positive: Constant Conditional OP)
+		destination += replacementLength;
 
 		current += patternLength;
 		before = current;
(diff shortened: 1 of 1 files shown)


sanssecours:
jenkins build mingw64 please

sanssecours:
@markus2330 👋 Can you please start the `mingw64` build job for this pull request? Thank you.

src/libs/utility/text.c
```diff
@@ -147,7 +147,7 @@ char * elektraReplace (char const * const text, char const * const pattern, char
 	ELEKTRA_LOG_DEBUG ("Found “%lu” occurrences of “%s” in “%s”", numberOccurrences, pattern, text);
 
 	ssize_t replacementLength = strlen (replacement);
-	ssize_t textLength = elektraStrLen (text);
+	ssize_t textLength = strlen (text) + 1;
```
markus2330: Can you please explain that change?

markus2330:
jenkins build mingw64 please

markus2330:
add to whitelist

markus2330:
Build is started! I added you to whitelist, so I hope you can now start it, too. I hope @BernhardDenner finds a solution for central management of people allowed to trigger build jobs. Currently it is chaotic (defined per job).

src/libs/utility/text.c
sanssecours: Honestly, I do not know why Clang Analyzer reports the warnings and why using `strlen` instead of `elektraStrLen` fixes the issue. I could also wrap the first and last call of `strncpy` in  

```c
#ifndef __clang_analyzer__
#endif
```

, to disable the static analysis partially if you prefer that. Adding an additional null check after determining the size of `text`:

```c
if (text == NULL)
{
    return NULL;
}
```

works too, but does needlessly increase the cyclomatic complexity to `11`. Another possible fix would be to just remove `elektraReplace` altogether.

markus2330:
Thank you for the fixes!

PR #1481, opened 2017-04-27
(Markdown) Shell Recorder: Minor Improvements

sanssecours:
# Purpose

This pull request fixes minor problems concerning (Markdown) Shell Recorder tests, including the issues below.

1. All Markdown Shell Recorder (MSR) tests  should now properly clean up modifications to the database.
2. The MSR now tests output following `STDOUT-REGEX:` properly.
3. Markdown Shell Recorder tests now supports whitespace after directives such as `RET:` or `STDOUT-REGEX:`.

# Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
       brew install yajl;
       brew install augeas;
       brew install dbus;
-      brew install qt5;
+      brew install qt;
       brew install discount;
       brew install openssl botan libgcrypt;
       brew install libgit2;
--- a/doc/help/kdb-complete.md
+++ b/doc/help/kdb-complete.md
@@ -64,13 +64,13 @@ kdb complete user --max-depth=1
 # list suggestions for the namespace user, only the next level as it ends with /
 # note the difference to the previous example, which uses no trailing /
 kdb complete user/ --max-depth=1
-#>RET: 0
-#>STDOUT: .*
+# RET:0
+# STDOUT-REGEX: .*
 
 # list all possible namespaces or mountpoints, only the current level
 kdb complete --max-depth=1
-#>RET: 0
-#>STDOUT: .*
+# RET:0
+# STDOUT-REGEX: .*
 
 # list suggestions for /sw/elektra/examples/kdb-complete, only the current level
 kdb complete /sw/elektra/examples/kdb-complete --max-depth=1
--- a/doc/help/kdb-complete.md
+++ b/doc/help/kdb-complete.md
@@ -64,12 +64,10 @@ kdb complete user --max-depth=1
 # list suggestions for the namespace user, only the next level as it ends with /
 # note the difference to the previous example, which uses no trailing /
 kdb complete user/ --max-depth=1
-# RET:0
 # STDOUT-REGEX: .*
 
 # list all possible namespaces or mountpoints, only the current level
 kdb complete --max-depth=1
-# RET:0
 # STDOUT-REGEX: .*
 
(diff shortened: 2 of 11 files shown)


tests/shell/shell_recorder/listtest.esr
```diff
@@ -16,4 +16,5 @@ RET: 0
 
 < kdb ls $Mountpoint
 
+< kdb rm /sw/kdb/current/plugins
```
markus2330: This is dangerous, it will remove user data.

Ideally tests should only manipulate `/test` or at least restore other changes.

markus2330:
Thank you, many nice changes!

Please do not remove keys from outside of `/test`.

tests/shell/shell_recorder/listtest.esr
sanssecours: I tried to backup and restore the data in commit f3b9900. However, it looks like `kdb cp` does not work correctly. The example below shows what I mean.

```sh
# 🐣 Create new test data
kdb set /sw/kdb/current/plugins 🦄
#> Using name user/sw/kdb/current/plugins
#> Create a new key user/sw/kdb/current/plugins with string 🦄
kdb get /sw/kdb/current/plugins
#> 🦄

# 💾 Copy the data using a cascading namespace
kdb cp /sw/kdb/current/plugins /backup/plugins

# 😢 Looks like this does not work
kdb get /backup/plugins
# RET: 1
# STDERR: Did not find key

# 😭 The command `kdb cp` did not create new data, although
# it exited successfully.
kdb ls user
#> user/sw/kdb/current/plugins

# 🤕 Let us try again — this time using an absolute namespace.
kdb cp /sw/kdb/current/plugins user/backup/plugins
# 🤔 Looks like this does not work either.
kdb get user/backup/plugins
# RET: 1
# STDERR: Did not find key

# 😱 Oh, there is our copy
kdb ls user
#> user/backup/pluginsgins
#> user/sw/kdb/current/plugins
kdb get user/backup/pluginsgins
#> 🦄
```

markus2330: Please create an issue or add it to #401.

I am afraid you have to do import/export (currently not working with all plugins, see #51) or copy all namespaces separately (using `kdb cp`). Did you also try `kdb mv`?

sanssecours: > Please create an issue or add it to #401.

Since I do not think that issue #401 describes the same problem, I just opened a new issue [here](https://github.com/ElektraInitiative/libelektra/issues/1483).

> Did you also try `kdb mv`?

Yes, but I already stopped after `kdb mv` reported the following problem:

```
First key found user/sw/kdb/current/plugins does not exactly match given key
/sw/kdb/current/plugins, aborting (use -r to move hierarchy)
```
after I entered

```sh
kdb mv /sw/kdb/current/plugins user/backup
```

. 

> I am afraid you have to do import/export (cascading currently not working with all plugins, see #51) or copy all namespaces separately (using `kdb cp`).

Honestly, that sounds like too much effort for a simple test. I just removed the two commits that changed `listtest.esr`. That should get rid of the possibility that the test removes user data.

markus2330:
Thank you for the improvements! Great job!

tests/shell/shell_recorder/listtest.esr
markus2330: > Honestly, that sounds like too much effort for a simple test.

Yes, but I think this functionality is also needed elsewhere. Let us discuss further discuss expected behavior in #1483.

Issue #1490, opened 2017-05-08
testshell_markdown_mini test case fails


markus2330:
The mini testcase fails on my computer. Latest Elektra installed via debs from build server.

```
139/176 Test #139: testshell_markdown_mini .................***Failed    2.58 sec
protocol file: /tmp/elektraenv.tPrszXf4O
kdb mount mini.ini /examples/mini mini
Return value “134” doesn't match “0”
kdb set /examples/mini/key value
kdb set /examples/mini/mi/mi/mr beaker
kdb export /examples/mini mini
Return value “6” doesn't match “0”

ERROR - STDOUT:
“”
doesn't match “key=value⏎mi/mi/mr=beaker”

echo "🔑 = 🦄"           >> `kdb file /examples/mini`
echo "level1/level2 = 👾" >> `kdb file /examples/mini`
kdb ls /examples/mini

ERROR - STDOUT:
“user/examples/mini/key
user/examples/mini/mi/mi/mr”
doesn't match “user/examples/mini/key⏎user/examples/mini/level1/level2⏎user/examples/mini/mi/mi/mr⏎user/examples/mini/🔑”

kdb get "/examples/mini/🔑"
Return value “1” doesn't match “0”

ERROR - STDOUT:
“”
doesn't match “🦄”

kdb rm -r /examples/mini
kdb umount /examples/mini
Return value “1” doesn't match “0”
kdb mount mini.ini /examples/mini mini
Return value “134” doesn't match “0”
kdb set /examples/mini/key ';#=\'
kdb file /examples/mini | xargs cat

ERROR - STDOUT:
“kdbOpen 1
ksNew 204
keyNew 17 10
user/MyApp/mykeynew_value
keyMeta 5 3

[... huge keyset]

user/valueable_data1234_important_data
keyMeta 14 3
trigger/error10
keyEnd
ksEnd”
doesn't match “key=\\;\\#\\=\\\\”

kdb get /examples/mini/key
echo 'background = \#0F0F0F ; Background color' >> `kdb file /examples/mini`
echo 'foreground = \#FFFFFF # Foreground color' >> `kdb file /examples/mini`
kdb get /examples/mini/background
Return value “1” doesn't match “0”

ERROR - STDOUT:
“”
doesn't match “#0F0F0F”

kdb get /examples/mini/foreground
Return value “1” doesn't match “0”

ERROR - STDOUT:
“”
doesn't match “#FFFFFF”

kdb rm -r /examples/mini
[...]

PR #1491, opened 2017-05-09
Travis: Also Test Installed Version of Elektra

sanssecours:
# Purpose

This change makes sure that the installed version of Elektra also works without problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,3 +64,5 @@ before_script:
 script:
   - ninja
   - ninja run_all
+  - ninja install
+  - ninja run_all


markus2330:
Thank you for creating the PR, good idea to add this! Why is it work in progress?

Btw. Github has some troubles on my side, I cannot click merge pull request and the markdown rendering is broken. Is this a local issue or does Github has problems right now?

sanssecours:
> Thank you for creating the PR, good idea to add this! Why is it work in progress?

Mainly because I wanted to check if the build server had the same problems that I reported [here](https://github.com/ElektraInitiative/libelektra/issues/1490#issuecomment-300063334).

> Btw. Github has some troubles on my side, I cannot click merge pull request and the markdown rendering is broken. Is this a local issue or does Github has problems right now?

Everything works fine on my side. Maybe you need to enable JavaScript?

markus2330:
Thank you! Good to have more tests!

> Everything works fine on my side. Maybe you need to enable JavaScript?

Seems to be a problem with my Firefox (45.9.0).

Issue #1493, opened 2017-05-13
Crypto: improve plugin performance


petermax2:
I discovered a design flaw in the crypto plugin while debugging the benchmark  `src/libs/tools/benchmarks/benchmark_crypto_comparison.cpp`.

We fork into the gpg process for every key in a keyset in both methods: kdb get and kdb set. This makes the plugin unusable for larger keysets. Encrypting 100 keys takes between 3 and 7 seconds!

Instead we should derive the cryptographic key and the initialization vector only once for every keyset that is encrypted/decrypted.

markus2330:
Thank you for reporting the issue!

Is there no simple way to keep the process open and keep sending data to it?

petermax2:
> Is there no simple way to keep the process open and keep sending data to it?

As far as I can remember gpg quits after the request has been processed, but I will look into it anyway.

petermax2:
> Is there no simple way to keep the process open and keep sending data to it?

Luckily this is not necessary. It is enough to pull the `fork()` into GPG a level up. Now only one GPG call takes place for the whole `KeySet` during `kdb get` and `kdb set`.

markus2330:
I do not think we should over-optimize it: The idea of this feature is to encrypt single keys, not everything. (For everything, `fcrypt` is the better solution.)  Security and usability is more important, let us better concentrate on that. Thus a case study where you also discuss the other aspects would be a really important addition to the benchmarks.

Of course, if you find a simple and secure solution to make it faster, go for it. But avoiding  PBKDF2 sounds like a redesign where security would be traded with performance.

petermax2:
> avoiding PBKDF2 sounds like a redesign where security would be traded with performance.

I agree on that and do not have a secure alternative. Let's close the issue.

PR #1494, opened 2017-05-14
crypto: improve plugin performance

petermax2:
# Purpose

Instead of reading the master password for every Key in a KeySet during en/decryption we now read and decrypt the master password once for the whole KeySet. 

See #1493 for full discussion.

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/src/plugins/crypto/crypto.c
+++ b/src/plugins/crypto/crypto.c
@@ -21,6 +21,7 @@
 #include "botan_operations.h"
 #endif
 #include "gpg.h"
+#include "helper.h"
 #include <kdb.h>
 #include <kdberrors.h>
 #include <kdbtypes.h>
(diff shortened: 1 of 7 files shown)


src/plugins/crypto/crypto.c
```diff
@@ -145,6 +146,28 @@ static int elektraCryptoCreateRandomString (Key * errorKey ELEKTRA_UNUSED, char
...
+ * @brief overwrites the value of the key with zeroes and then releases the Key.
+ * @param key to be overwritten and released
+ */
+static void elektraCryptoSafelyReleaseKey (Key * key)
+{
+	if (key)
+	{
+		// overwrite key content with zeroes
+		ssize_t length = keyGetValueSize (key);
+		kdb_octet_t * newVal = calloc (length, 1);
```
markus2330: Please use elektraCalloc and friends for memory management.

src/plugins/crypto/crypto.c
```diff
@@ -145,6 +146,28 @@ static int elektraCryptoCreateRandomString (Key * errorKey ELEKTRA_UNUSED, char
...
+		if (newVal)
+		{
+			keySetBinary (key, newVal, length);
```
markus2330: I am afraid in the current implementation there is no guarantee this line will overwrite anything.
(realloc might return a new pointer despite of requesting the same size). We could change keySetRaw to get the guarantee (is trivial and an optimization, too. The code in line 529 should use elektraRealloc btw.).

Or what about simply doing a `memset` of what you get from `keyValue`?

petermax2:
> Is this a security relevant change?

No, it's not. Everything works as before, only the GPG forks are reduced.

petermax2:
@markus2330 I'm ready to merge for now. But we have to further discuss the key derivation in #1493 .

markus2330:
Thank you!

Issue #1370, opened 2017-02-23
cbmc detects errors


markus2330:
@e1528532 another quality issue: as cbmc is unable to verify the assertions in `kdbSet`:

```
> scripts/cbmc ~build
...
Counterexample:

State 3 file <built-in-additions> line 22 thread 0
----------------------------------------------------
  __CPROVER_thread_id=0 (0000000000000000000000000000000000000000000000000000000000000000)

State 4 file <built-in-additions> line 23 thread 0
----------------------------------------------------
  __CPROVER_threads_exited=__CPROVER_threads_exited#1 (?)

State 5 file <built-in-additions> line 24 thread 0
----------------------------------------------------
  __CPROVER_next_thread_id=0 (0000000000000000000000000000000000000000000000000000000000000000)

State 6 file <built-in-additions> line 31 thread 0
----------------------------------------------------
  __CPROVER_deallocated=NULL (0000000000000000000000000000000000000000000000000000000000000000)

State 7 file <built-in-additions> line 32 thread 0
----------------------------------------------------
  __CPROVER_dead_object=NULL (0000000000000000000000000000000000000000000000000000000000000000)

State 8 file <built-in-additions> line 33 thread 0
----------------------------------------------------
  __CPROVER_malloc_object=NULL (0000000000000000000000000000000000000000000000000000000000000000)

State 9 file <built-in-additions> line 34 thread 0
----------------------------------------------------
  __CPROVER_malloc_size=0 (0000000000000000000000000000000000000000000000000000000000000000)

State 10 file <built-in-additions> line 35 thread 0
----------------------------------------------------
  __CPROVER_malloc_is_new_array=FALSE (0)

State 11 file <built-in-additions> line 36 thread 0
----------------------------------------------------
  __CPROVER_memory_leak=NULL (0000000000000000000000000000000000000000000000000000000000000000)

State 12 file <built-in-additions> line 58 thread 0
----------------------------------------------------
  __CPROVER_rounding_mode=0 (00000000000000000000000000000000)

State 13 file <built-in-additions> line 79 thread 0
----------------------------------------------------
  __CPROVER_pipe_count=0 (00000000000000000000000000000000)

State 17 file kdb.c line 1154 thread 0
----------------------------------------------------
  handle=1 + ((struct _KDB *)NULL) (0000000000000000000000000000000000000000000000000000000000000001)

State 18 file kdb.c line 1154 thread 0
----------------------------------------------------
[...]

PR #1508, opened 2017-06-15
Quality improvements

e1528532:
# Purpose

A bunch of small quality improvements.

# Checklist

needs to be checked.

- [X] I added unit tests

@markus2330 some considerations:

ASAN: I've had two places it complained about during the tests. one was in the range.c file (see changed files), but as it only gets multiplied by 1 or -1 it should not matter. i've added an assertion to guarantee that. it still complains though, doesn't seem to get the assertion

The other place it complains about is in keyvalue.c  on line 541 (`memcpy (key->data.v, newBinary, key->dataSize);`). The exact error is

```
95: ==63583==ERROR: AddressSanitizer: global-buffer-overflow on address 0x00010a8aa749 at pc 0x00010a9da1e8 bp 0x7fff553cc500 sp 0x7fff553cbcb0
95: READ of size 10 at 0x00010a8aa749 thread T0
95:     #0 0x10a9da1e7 in __asan_memcpy (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x4a1e7)
95:     #1 0x10a9550ed in keySetRaw keyvalue.c:541
```

I don't get this issue though, after checking the affected code, it seems fine to me. It either uses realloc or elektraMalloc to allocate key->dataSize bytes, then it copies that many bytes with memcpy, so in both cases enough should be allocated and no overflow. The only thing i can possibly think of is that it could maybe copy "too much" if the argument dataSize is larger than the size of newBinary maybe? Or could it be that this is some bug in this __asan_memcpy call itself, of which we don't have much influence over?

CBMC: after fixing the paths on mac, i get a totally different result than in the initial issue #1370 . My output looks like
```
size of program expression: 14356 steps
simple slicing removed 105 assignments
Generated 8 VCC(s), 6 remaining after simplification
Passing problem to propositional reduction
converting SSA
Running propositional reduction
Post-processing
Solving with MiniSAT 2.2.1 with simplifier
245089 variables, 557781 clauses
SAT checker: instance is SATISFIABLE
Solving with MiniSAT 2.2.1 with simplifier
245089 variables, 45210 clauses
SAT checker inconsistent: instance is UNSATISFIABLE
Runtime decision procedure: 3.66s

** Results:
[] free called for new[] object: SUCCESS
[free.assertion.1] free argument is dynamic object: FAILURE
[free.assertion.2] free argument has offset zero: FAILURE
[free.assertion.3] double free: SUCCESS
[free.assertion.4] free called for new[] object: SUCCESS

** 2 of 5 failed (2 iterations)
VERIFICATION FAILED
```
without a specific hint where to search for the issues.

rest should be ok i think.

### DIFF
--- a/src/plugins/xerces/README.md
+++ b/src/plugins/xerces/README.md
@@ -82,7 +82,7 @@ XSD transformations, schemas or DTDs are not supported yet.
 ```sh
 # Backup-and-Restore:/examples/xercesfile
 
-sudo kdb mount /xerces.xml /examples/xercesfile xerces
+sudo kdb mount xerces.xml /examples/xercesfile xerces
 
 kdb set /examples/xercesfile foo
 kdb setmeta /examples/xercesfile xerces/rootname xerces
--- a/src/plugins/xerces/README.md
+++ b/src/plugins/xerces/README.md
@@ -82,7 +82,7 @@ XSD transformations, schemas or DTDs are not supported yet.
 ```sh
 # Backup-and-Restore:/examples/xercesfile
 
-sudo kdb mount xerces.xml /examples/xercesfile xerces
+sudo kdb mount /xerces.xml /examples/xercesfile xerces
 
 kdb set /examples/xercesfile foo
 kdb setmeta /examples/xercesfile xerces/rootname xerces
--- a/scripts/cbmc
+++ b/scripts/cbmc
@@ -22,7 +22,7 @@ ARG="--depth 200 -D ELEKTRA_BMC"
 EINC="-I $EDIR/src/include -I $BDIR/src/include"
 if stringContain "Darwin" "$(uname)"; then
 	YINC="-I /usr/local/include/yajl -I $BDIR/src/plugins/yajl"
-yelse
+else
 	YINC="-I /user/include/yajl -I $BDIR/src/plugins/yajl"
 fi
 
--- a/scripts/cbmc
+++ b/scripts/cbmc
@@ -20,7 +20,11 @@ set -ex
 
 ARG="--depth 200 -D ELEKTRA_BMC"
 EINC="-I $EDIR/src/include -I $BDIR/src/include"
-YINC="-I /usr/include/yajl -I $BDIR/src/plugins/yajl"
+if stringContain "Darwin" "$(uname)"; then
+	YINC="-I /usr/local/include/yajl -I $BDIR/src/plugins/yajl"
+yelse
+	YINC="-I /user/include/yajl -I $BDIR/src/plugins/yajl"
+fi
 
 cd $EDIR/src/plugins/yajl
 cbmc --function elektraYajlSet $ARG $YINC $EINC yajl_gen.c
--- a/src/libs/ease/array.c
+++ b/src/libs/ease/array.c
@@ -60,7 +60,7 @@ int elektraArrayValidateName (const Key * key)
 			underscores++;
 		}
 
-		for (; isdigit ((unsigned char) *current); current++)
+		for (; isdigit ((unsigned char)*current); current++)
 		{
 			digits++;
 		}
--- a/src/libs/ease/array.c
+++ b/src/libs/ease/array.c
@@ -60,7 +60,7 @@ int elektraArrayValidateName (const Key * key)
 			underscores++;
 		}
 
-		for (; isdigit (*current); current++)
+		for (; isdigit ((unsigned char) *current); current++)
 		{
 			digits++;
 		}
--- a/src/libs/ease/keyname.c
+++ b/src/libs/ease/keyname.c
@@ -13,10 +13,17 @@
 /**
  * @brief get relative position of key based on parentKey
  *
+ * @pre  parentKey is either the same key as cur, or one of its parents
+ * @post a pointer to the relevant part of the parent key's name, the full
+ * name if there is no relation to the parentKey
+ *
+ * If the parentKey does not fulfill the precondition, the result won't be the
+ * correct relative key of cur.
+ *
  * @param cur the key below parentKey we want to get the relative basename of
  * @param parentKey the key that defines the root/base
  *
- * @return a pointer to the name of the key cur
+ * @return a pointer to the relative part name of the key cur
  */
 const char * elektraKeyGetRelativeName (Key const * cur, Key const * parentKey)
 {
@@ -29,6 +36,14 @@ const char * elektraKeyGetRelativeName (Key const * cur, Key const * parentKey)
 		{
 			offset += strstr (keyName (cur), keyName (parentKey)) - keyName (cur);
 		}
+	}
+	if (offset == keyGetNameSize (cur))
+	{
+		offset = keyGetNameSize (cur) - 1; // equality of the keys
+	}
+	else if (offset > keyGetNameSize (cur))
+	{
+		offset = 0; // no relation or invalid arguments, return full name
 	}
 	return keyName (cur) + offset;
 }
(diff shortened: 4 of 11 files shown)


markus2330:
> ASAN

Yes, ASAN is a run-time check independent of the assertions. Seems like 1 or -1 already triggers the undefined behavior (if you are sure that the assertion is not triggered).

> keyvalue.c on line 541

Maybe we need a memmove here? In the case the realloc gives back the same pointer, we have overlapping regions. We could "optimize" to use memmove only in one branch, memcopy in the other. We should use elektraMemcpy/elektraMemmove here.

Furthermore, we should use elektraRealloc in this function.

> without a specific hint where to search for the issues.

Strange. Did you use the same version? Do you get the same output on linux?

e1528532:
Making some progress here, though the issues were tricky. Those two things were indeed errors, just not obvious to see on a first glance.

The memcpy issue: Actually, the issues wasn't directly with our way of using realloc and memcopy, but we had a test that creates a binary key with a value e.g. "asdf" and a fixed size. But the size didn't match the length of the string. Therefore, memcopy actually copied more than just the string "asdf", and ASAN detected this as an error. I've corrected the size of the affected keys in the test so it never reads over the bounds of our test data to avoid reading arbitrary things. There were a few such cases. Now ASAN is happy.

I have still reorganized keySetRaw a bit so it uses elektraRealloc now. If the memory regions overlap, i use memmove now to be safe in all possible cases, otherwise i stick to memcpy. I think I can't use elektraMemcpy in this case, as this requires two keys as parameters, however in that function we only copy possibly arbitrary data to the value of a single key which seems to be a different use case to me.

I have also identified a few issues in range.c. It was indeed an overflow caused by multiplying e.g. (-1) * 3 in the scope of being an unsigned long long. Therefore it interpreted the (-1) as ULLONG_MAX, and multiplied that by 3. Obviously, that is over the bounds. I have adjusted the function to handle this case correctly. Furthermore, errno has to be reset to 0 before calling a function like strtoull, this isn't done automatically. I had an issue with the testcases because of that. Last, i use strtoull now instead of using strtoll to read in the correct bounds. As far as I've seen, negative signs get handled beforehand, so it always reads an unsigned string. I also added test cases that check the edge cases of LLONG_MIN, LLONG_MAX and ULLONG_MAX.

markus2330:
Thank you, great job!

markus2330:
jenkins build please

e1528532:
reproduced the cbmc issue but it seems strange to me, especially considering the following:

First i have slightly modified the code like this (with line numbers):

```
1207	int syncstate = 0;
1208	syncstate = splitDivide (split, handle, ks);
1209	ELEKTRA_ASSERT (syncstate == -1 || syncstate == 0 || syncstate == 1, "syncstate not -1 or 0 or 1, but %d", syncstate);
```
and i get this output
```
State 154 file kdb.c line 1207 function kdbSet thread 0
----------------------------------------------------
  syncstate=0 (00000000000000000000000000000000)

State 158 file kdb.c line 1208 function kdbSet thread 0
----------------------------------------------------
  syncstate=3 (00000000000000000000000000000011)

Violated property:
  file kdb.c line 1209 function kdbSet
  assertion syncstate == -1 || syncstate == 0 || syncstate == 1
  !((signed long int)!(syncstate == -1 || syncstate == 0 || syncstate == 1) != 0l)
```

How is it possible that syncstate  = 3 after line 1208, so it hurts my assertion at 1209? I have checked the splitDivide function. It defines an int needsSync which it initializes to 0. If it encounters an error, it returns -1. And it may set needsSync to 1 via assignment. At the end it returns needsSync. Doesn't seem to me that there is an obvious possibility how this can be 3, maybe i oversee something, do you have any idea?

markus2330:
> reproduced the cbmc issue

What did you do differently?

> Doesn't seem to me that there is an obvious possibility how this can be 3, maybe i oversee something, do you have any idea?

Maybe you can help cbmc by adding postconditions in splitDivide/splitSync?

e1528532:
> What did you do differently?

On linux it works anyway, on mac i simply disabled the first cbmc test that tests yajl for now so the second one runs :) 

> Maybe you can help cbmc by adding postconditions in splitDivide/splitSync?

I added the postcondition `ELEKTRA_ASSERT (needsSync == 0 || needsSync == 1, "needsSync not 0 or 1, but %d", needsSync);` now before the return statement in splitDivide and even ran cbmc on the splitDivide method then with `cbmc --trace -D HAVE_KDBCONFIG_H --function splitDivide $ARG $EINC split.c` but this assertion holds there. the other return statement returns -1; so i see no issue either. return type is int which should match with the variable in kdb.c so also no conversation error. But somehow it is always 3 when running cbmc on kdb.c after the call to splitDivide. Tried a bunch of things already but not really getting further with that issue. Maybe we use cbmc wrong or some cbmc bug?

markus2330:
> Maybe we use cbmc wrong or some cbmc bug?

Could be, can you report the issue?

e1528532:
Ok, i think this should be treated separately then. I think this PR should be fine to merge then, before blowing it up with further stuff?

markus2330:
Yes, good idea! Let us wait for the build server before merging this.

markus2330:
jenkins build please

markus2330:
Seems like we now have a "Uninitialized Memory Conditional - 36":
https://build.libelektra.org/jenkins/job/elektra-mergerequests-unstable/1148/console

e1528532:
jenkins build please

e1528532:
looks better now but since when do we have this encoding issue with the xerces tests? didn't this used to work?
Judging from the output it seems that the shellrecorder seems to read the expected output wrong, as there it displays the line endings as a ⏎ character?

`<\?xml version=\"1\.0\" encoding=\"UTF-8\" standalone=\"no\" \?>⏎<xerces>⏎⏎  <bar meta=\"da_ta\">bar</bar>foo⏎⏎</xerces>`

wasn't able to reproduce this locally, works both on my osx and arch machines.

sanssecours:
> looks better now but since when do we have this encoding issue with the xerces tests?

I guess commit 0dc865d introduced the problem, since Travis did [not report any problems](https://travis-ci.org/ElektraInitiative/libelektra/builds/250836466) with your most recent pull request (#1530). As far as I can tell the [actual output](https://travis-ci.org/ElektraInitiative/libelektra/jobs/250873032#L2501) of the `xerces` plugin contains an additional `foo` after the tag `<xerces>`

```
<xerces>foo
```

that is not part of the expected output specified via the [ReadMe](https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/xerces/README.md#mounting-setting-a-key-and-exporting).

> Judging from the output it seems that the shellrecorder seems to read the expected output wrong, as there it displays the line endings as a ⏎ character?

This is [intentional](https://github.com/ElektraInitiative/libelektra/blob/master/tests/shell/shell_recorder/README.md#checks). We replace newline characters with `⏎`, to compare multiline strings with `grep`.

src/plugins/xerces/README.md
```diff
@@ -82,7 +82,7 @@ XSD transformations, schemas or DTDs are not supported yet.
 ```sh
 # Backup-and-Restore:/examples/xercesfile
 
-sudo kdb mount xerces.xml /examples/xercesfile xerces
+sudo kdb mount /xerces.xml /examples/xercesfile xerces
```
sanssecours: Looks like the leading `/` before `xerces.xml` causes the problem on the Travis build server. I can reproduce the problem locally after I add the slash character.

e1528532:
seems good now. That one failed build is not related to my PR right?

markus2330:
Thank you, great job!

Yes, it is unrelated. Seems like the ssh-setup is not complete and artifacts cannot be pushed to doc.libelektra.org. @BernhardDenner: Can you take a look?

Issue #1290, opened 2017-01-23
Unclear what was mounted


markus2330:
As discussed with @krit0n 

`kdb list-plugins <key name>`

returns the list of plugins that would be used for key name.

- [ ] `kdb mount -v`

list plugins per mountpoint

- [ ] `kdb gmount -v`

list plugins for global mount

- [x] `kdb global-umount` is missing, see #1132

sanssecours:
I really like the idea of listing the plugins for a specific mountpoint via `kdb mount -v` (or even just `kdb mount`). It would be really awesome if someone implements this functionality 🙏.

markus2330:
`kdb mountpoint-info <mountpoint>` lists plugins, for example:

```
> kdb mountpoint-info /hosts                                                                                                                                                                                                     Dienstag 23:24:41
Version: 0.8.19
Default resolver: resolver_fm_hpu_b
Default storage: dump
Mountpoint: /hosts
File: hosts
        config:
        glob/set/#0 = /ipv4/*
        glob/set/#1 = /ipv6/*
        glob/set/#2 = /ipv4/*/*
        glob/set/#3 = /ipv6/*/*
        glob/set/#4/ = /*
        path = hosts
getplugins:
        #0#resolver
        #5#hosts#hosts#
        #6#glob#glob#
setplugins:
        #0#resolver
        #1#glob
        #2#error#error#
        #3#network#network#
        #5#hosts
        #6#sync#sync#
        #7#resolver
errorplugins:
        #5#resolver_fm_hpu_b#resolver#
```

Merged with #1364. The output may be improved by removing the `#`, though.

PR #1509, opened 2017-06-15
kdb-global-umount: first draft #1290

e1528532:
# Purpose

Adds global-umount tool

# Checklist

- [ ] I added unit tests

@markus2330 Honestly this bugged me quite a lot because to me it seems to be too easy the way it currently is. So i was trying to find out what i'm missing. But judging from the kdb tools library (which is not really easy to understand imo) it seems that global plugins are always mounted to this particular path and just removing the keys from there seems to be sufficient. There is no need to reorder the array of global plugins or anything as the next time you mount a new global plugin, the mounting code reorders it anyway.

But shouldn't global plugins also have an entry in system/elektra/modules so i could call their corresponding close function before or is that not necessary?

### DIFF
--- a/doc/help/kdb-global-umount.md
+++ b/doc/help/kdb-global-umount.md
@@ -9,6 +9,27 @@ kdb-global-umount(1) - Unmount a global plugin from the key database
 
 Unmount a global plugin from the key database.
 
+## EXAMPLES
+
+```sh
+
+sudo kdb global-mount tracer
+
+sudo kdb global-mount
+# STDOUT-REGEX: .*tracer\nspec.*
+
+sudo kdb global-umount tracer
+
+# spec is always mounted by default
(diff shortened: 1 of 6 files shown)


e1528532:
@markus2330 may i ask for your opinion?

markus2330:
Thank you, looks okay!

Extending libtools API GlobalPluginsBuilder and using remPlugin would have been nice (especially once other tools want to do the same). But it is not really important, other cleanups of the libtools are more urgent. (Like the import/export interface).

Some shellrecorder tests would be great. And it would be good if you can unmount what other shellrecorder tests globally mount at the moment.

markus2330:
Thank you for adding this features! For now its best to concentrate on the release blockers so that we can get the next release done.

PR #1510, opened 2017-06-15
fix-getopt

e1528532:
# Purpose

Apparently for the kdb tools options that have arguments don't work in the short format. 
This is due to an error in our getopt string. As seen on e.g. https://linux.die.net/man/3/getopt_long

> optstring is a string containing the legitimate option characters.
> If such a character is followed by a colon, the option requires an argument, so getopt() places a pointer to the following text in the same argv-element, or the text of the following argv-element, in optarg.
> Two colons mean an option takes an optional arg;

# Checklist

needs to be checked.

- [ ] I added unit tests

@markus2330 yet another small quality improvement but i thought it deserves its own branch since its rather a bugfix and so we can quickly merge it as #1345 now depends on it.

### DIFF
--- a/src/tools/kdb/check.hpp
+++ b/src/tools/kdb/check.hpp
@@ -22,7 +22,7 @@ public:
 
 	virtual std::string getShortOptions () override
 	{
-		return "vcfC";
+		return "vcfC::";
 	}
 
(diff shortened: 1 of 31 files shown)


markus2330:
Great job, thank you!

PR #1514, opened 2017-06-18
fix-getopt

e1528532:
# Purpose

some forgotten option colons which apparently caused sigsevs

# Checklist

@markus2330 lets see what travis says now, i only had some issues with shellrecorder tests now but those don't seem to be related to that at a first glance


e1528532:
looks good

markus2330:
Thank you for the quick fix!

Did you also consider to automate the adding of the colons? Doesn't `long_options` (in `cmdline.cpp`) contain all the necessary information to fix the short options?

PR #1527, opened 2017-07-05
Benchmarks: Generate KeySet & Opmphm benchmark demo

KurtMi:
# Purpose

As promised the generateKeySet function and a small demo.
Includes also a change of the benchmarks executable names, which where non intuitive.

# Checklist


benchmarks/CMakeLists.txt
```diff
@@ -5,17 +5,17 @@ add_headers (HDR_FILES)
...
 
-	set_target_properties (${source} PROPERTIES
+	set_target_properties (benchmark_${source} PROPERTIES
 			COMPILE_DEFINITIONS HAVE_KDBCONFIG_H)
 
-	get_target_property (command ${source} LOCATION)
+	get_target_property (command benchmark_${source} LOCATION)
 
-	add_custom_target (benchmark_${source}
+	add_custom_target (benchmark_${source}_callgrind
```
markus2330: Is there a need to give the target a different name?

Btw. I like the rename of the executables, we need to document it in the release notes, though.

benchmarks/benchmarks.c
```diff
@@ -55,3 +54,188 @@ void benchmarkFillup ()
 		}
 	}
 }
+
+#ifdef ELEKTRA_ENABLE_OPTIMIZATIONS
+
+// needed for the opmphmRandom ()
+#include "../src/libs/elektra/opmphm.c"
```
markus2330: A bit hackish (also the whole approach with the `ifdef`), but okay for a benchmark.

benchmarks/benchmarks.c
```diff
@@ -55,3 +54,188 @@ void benchmarkFillup ()
...
+ *
+ */
+static void fillUpWithRandomChars (char * name, int length, uint32_t * seed)
+{
+	int i = 0;
+	char * c = name;
+	while (i < length)
+	{
+
+		if ((opmphmRandom (seed) % 10) == 0)
```
markus2330: This breaks uniformity of the random numbers. Maybe we can change the signature so that we can tell between which numbers we want a number?

We should also reconsider it to be more than 32bit, because this limits the maximum length of a random number sequence.

benchmarks/benchmarks.c
```diff
@@ -55,3 +54,188 @@ void benchmarkFillup ()
...
+
+/**
+ * @brief Fills a string up with random chars and null terminates it.
+ *
+ * @param name the string
+ * @param length the length of the string (excluding the '\0')
+ * @param seed to generate random data
+ *
+ */
+static void fillUpWithRandomChars (char * name, int length, uint32_t * seed)
```
markus2330: size_t for length?

benchmarks/benchmarks.c
```diff
@@ -55,3 +54,188 @@ void benchmarkFillup ()
...
+	int i = 0;
+	char * c = name;
+	while (i < length)
+	{
+
+		if ((opmphmRandom (seed) % 10) == 0)
+		{
+			/* SPECIAL CHARS:
+			 *
+			 * do not use '/' = 47 for generation
```
markus2330: You could use everything if you use `keyAddBasename`. Isn't that simpler?

benchmarks/benchmarks.h
```diff
@@ -39,4 +39,37 @@ extern int num_dir; // default = NUM_DIR;
...
+#ifdef ELEKTRA_ENABLE_OPTIMIZATIONS
+
+/**
+ * KeySetShape
+ * Set minWordLength < maxWordLength and parent > 0.
+ * Not all Keys must have a direct parent,
+ * with the parent parameter the occurrence of direct parents can be influenced,
+ * the probability of having a direct parent is (1/parent).
+ * The shapef is used to determine the number of sub Keys.
+ * The parameter of the shapef are the following (in order):
```
markus2330: Can you give one example for these parameters?

benchmarks/opmphm.c
```diff
@@ -0,0 +1,60 @@
...
+	uint32_t seed;
+	if (getRandomSeed (&seed) != &seed)
+	{
+		fprintf (stderr, "FATAL: Seed Parsing Error or feed me more seeds\n");
+		return EXIT_FAILURE;
+	}
+	KeySetShape shape;
+	shape.parent = 1;
+	shape.minWordLength = 3;
+	shape.maxWordLength = 7;
```
markus2330: What about command-line arguments for the seed and the shape? (If given, otherwise use some default)

scripts/benchmark_opmphm
```diff
@@ -0,0 +1,17 @@
...
+numberOfSeeds=2
+
+if [ ! -f "CMakeCache.txt" ];
+then
+	echo "Execute $0 in build dir"
+	exit 1
+fi
+
+make benchmark_opmphm
+R --slave --no-save <<< "cat(runif($numberOfSeeds,0,4294967295), sep=\"\n\")" | bin/benchmark_opmphm
```
markus2330: bin/benchmark has bin/ hard coded.

Can you shortly document what the script does and how to run it?

scripts/find-tools
```diff
@@ -15,6 +15,7 @@ tags = {	'configure': "This script is used for the build configuration",
 		'env': "This script does some env stuff",
 		'mount': "This script mounts things",
 		'reformat': "This script reformats things",
+		'benchmark': "This script runs benchmarks",
```
markus2330: Do you refer to benchmark_opmphm here?

benchmarks/CMakeLists.txt
KurtMi: If I would not rename the custom target it would have the same name as the executable and would lead to a cmake error.

benchmarks/benchmarks.c
KurtMi: I am also not happy with this solution, I could rename `opmphmRandom ()` to `elektraRandom ()` then the function gets exported. But I think nobody except me needs a random function in the elektra core.

scripts/find-tools
KurtMi: I wanted to add a proper header in my script, so I need a new category since it does not fit in any other listed here.

benchmarks/benchmarks.c
KurtMi: I am not happy with the while function, I think it would be better to define somewhere a string with all possible chars in it and pick randomly from there.

The 32 bit limit for that type comes from the limitation of the `opmphmHashFunction ()` but the random numbers can surely be 64 bit.

benchmarks/benchmarks.c
markus2330: Yes, it would be used only in tests. Maybe you can at least separate the random function in a separate file?

Btw. should we really use randr_r? It is only in POSIX 2001 and declared obsolete in POSIX 2008. Maybe we use directly the algorithm of the paper you have found? (drand48 is not thread-safe).

benchmarks/benchmarks.c
markus2330: More bits are mostly needed for the sequence length. The lrand48 (which we can not use, see above) also only returns values from 0 and 2^31.

KurtMi:
@markus2330 This would be it.
The build seems broken:
```
SSH: Connecting from host [debian-stable.mm.build.libelektra.org]
SSH: Connecting with configuration [doc.libelektra.org] ...
ERROR: Exception when publishing, exception message [Failed to connect session for config [doc.libelektra.org]. Message [java.net.ConnectException: Connection refused (Connection refused)]]
```

markus2330:
Thank you!

Issue #1506, opened 2017-05-26
Try out randoop


markus2330:
With the Java binding we can try https://randoop.github.io/randoop/

e1528532:
I can try it out, is that ok?

markus2330:
Yes, that would be great!

e1528532:
After messing around a bit to make it work (I'll document it soon) with our bindings, i managed to make it produce some errors. Unfortunately errors in the sense that this creates a SIGSEGV in our native code thus terminating the JVM the hard way. The bad thing: randoop didn't log the sequence then that caused this errors in case of a JVM error, as it seems to be built to log after each test case.

So I finally came up with the idea of logging whats going on in the bindings by including a logging framework and write it to a file (in the bindings), so i can see the sequence of commands invoked by randoop independent of the jvm crash (by having the logger flush after each invocation, and always logging the method before each call to the native function of elektra).

Turned out the issue was by creating a Keyset like this (so no KS_END specified either, as randoop doesn't care about such details, but it also happens when you specify it at the end of the varargs):

`KeySet * ks = ksNew (10, -1, -2);`

Its probably not intended to be called that way i know, but crashing with a sigsegv is not nice.

As randoop always runs in the same issue again and again then we should maybe think about a way to fix this. While it would normally resume upon errors, it doesn't seem to be able to deal with JVM crashes caused by native code. I bet there are much more similar errors that it will discover by invoking methods with random input then.

markus2330:
Thanks for checking!

> Its probably not intended to be called that way i know, but crashing with a sigsegv is not nice. How should we handle such cases?

It cannot be fixed in C, but we can fix it in the Java Code. For example, always appending `KS_END` and `KEY_END`, or/and by checking if what we got is valid and throwing an exception if it is not.

> While it would normally resume upon errors to find more, it doesn't seem to be able to deal with JVM crashes caused by native code.

Maybe there is an option to spawn a new JVM for the tested code? Hopefully no other crashes will happen once you validate the vaargs, so maybe simply try that first.

e1528532:
I see. So the arguments given to keyset are interpreted as Key pointers. And if i give a random pointer like randoop does it is going to fail. There is no real way to check if the pointer is an actual elektra key right?

It is possible to add KEY_END automatically though at least to the java bindings, if we want such behavior.

I think the best way to resume is to simply exclude this creation method with varargs from the randoop tests so it cannot create keysets with arbitrary key pointers.

[...]

PR #1530, opened 2017-07-06
Randoop

e1528532:
# Purpose

Adds a script you can use to execute randoop, including instructions. Furthermore adds a note for randoop to docs/TESTING.md.
The other bigger thing is i refactored the jna bindings a bit as discussed in #1506 to only allow "safe" operations in the public binding api (so nothing that crashes the jvm). I also provided appropriate factory methods so you can really use the bindings in a style you'd expect from java without having to deal with details like the varchar separators used in C. It is still possible to use the raw operations directly if one really wishes to do so (e.g. by extending the Key class to expose the pointer), but I've made sure this is not encouraged by the default API.
Some other improvements: use primitives types for getShort and so on to prevent unnecessary boxing of values. Also added better null handling to some methods where passing null would have just caused a NPE and some minor polishing.

# Checklist

- [X] I added unit tests


markus2330:
Looks great, many improvements in the API! So using randoop was successful after all.

@vanessaNWK can you do a review?

markus2330:
Does this fix #1652?

@e1528532 If yes, can we finish this for the release? (this weekend?)

markus2330:
> Possibly, i have to check. Not sure if i find time tomorrow but i think i find time on monday or tuesday if thats still ok

It is okay but we will likely release without this, then. If you can fix something today, it would be ideal ;)

e1528532:
today i'm already occupied but i try to shift some time tomorrow to check it out earlier then

e1528532:
In case you agree with the memleak issue as stated in #1652 i think we can merge this to have better jna bindings for the upcoming release. In any case, even if there is still this one leak left where i am not sure if it is related to the bindings this version works better as before there were way more leak issues.

markus2330:
Do we really need a copy of all those CMake files? Especially FindPackageHandleStandardArgs.cmake?

e1528532:
These all were dependencies of FindJava.cmake (find_package_handle_standard_args is actually used in there), so if we want to use that module as it is (got it from cmake https://github.com/Kitware/CMake/tree/master/Modules) i think we do need them.

markus2330:
We use `find_package_handle_standard_args` at several places already.

What is wrong with the upstream `FindJava.cmake` which must be a part of every CMake installation? It worked quite well up to now.

sanssecours:
The package `FindPackageHandleStandardArgs.cmake` is already part of a standard CMake install at least on my machine:

```sh
locate FindPackageHandleStandardArgs.cmake
#> /usr/local/Cellar/cmake/3.9.4_1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake
```

. The same is true for `FindJava.cmake`, `CMakeFindJavaCommon.cmake` and `FindPackageMessage.cmake`. Please remove these files.

markus2330:
And bit more docu of how to execute the script would be nice. The script is also missing meta data.

Btw. is it on purpose that:

- CLASSPATH from outside is not considered
- CLASSPATH is not exported

e1528532:
Ok fixed the java detection i think. on ubuntu the name of the module in find_package is case sensitive, on osx its not, which falsely led me to assume that FindJava is not included on ubuntu. 
There is a little mixup between java version strings between the jdk versions (oracle/openjdk, java8/java9) which i think have caught with my workaround.

> And bit more docu of how to execute the script would be nice. 

Slightly improved docu and added metadata, but could you tell me what is unclear? To me it seems pretty understandable, but that can easily be because i am familiar with it. Also bumped the default libelektra jar to 0.8.20 for the relase. 

I usually just pass the absolute paths like `./randoop /Users/admin/Downloads/randoop-3.1.5/randoop-all-3.1.5.jar /usr/local/share/java/libelektra4j-0.8.19.jar`

> CLASSPATH from outside is not considered
> CLASSPATH is not exported

Well yes it is intended yes. This variable gets submitted to the jvm via the -classpath argument and only has to contain libelektra4j-0.8.19.jar and randoop-all-3.1.5.jar in order to make randoop work.  This gets set up by the script, so i think there is little use to consider the system-wide classpath additionally (it may even break things if it is configured in a non-usual way i fear). Therefore there is no need to export it either, as it gets submitted as a parameter. At least on my systems the $CLASSPATH environment variable is empty anyway, so its probably the case for most of the systems.

Ultimately, my initial intention for the script was to have some kind of blueprint how randoop can be used to check our bindings using a standard configuration, and if someone wants to do more special things it would act as a start to get it running.

markus2330:
Thank you for the update!

It seems like we have to wait with the release because the homepage is not ready. So if you can fix the other things till tomorrow, it would be fabulous. In particular #1666 should be fixed.

I saw that you hard code version numbers from Elektra, please use KDB_VERSION and friends instead.

> Slightly improved docu and added metadata, but could you tell me what is unclear?

It is clear together with what you say here. It should be the other way round: a more detailed explanation in the README, shorter descriptions in the PR. In particular examples are always good, the text after "I usually just pass the absolute paths like" exactly is what is missing in the description. You already improved it a lot, though.

> Ultimately, my initial intention for the script was ... it would act as a start to get it running.

Nothing wrong with that in general! For tests, however, it must be really simple so that anybody executes them at all.

e1528532:
added my example from before, and use cmake to configure the script now to get the version number dynamically. Wasn't aware that we do preprocess our scripts via cmake too :)

markus2330:
It seems like there is a conflict in this PR, please rebase.

e1528532:
rebased

markus2330:
Travis reports an error:

```
-- Found Java: /usr/lib/jvm/java-8-oracle/bin/java (found version "1.8.0.144") 

CMake Error at src/bindings/jna/CMakeLists.txt:11 (if):

  if given arguments:

    "Java_VERSION" "VERSION_GREATER_EQUAL" "1.8.0"

  Unknown arguments specified
```

markus2330:
Also here the build takes endless... (not a single macOS build so far). Will have to wait for tomorrow.

e1528532:
Unfortunately the macOS builds are very slow currently. I've cancelled the previous randoop builds and the haskell dependency builds (already merged, had my own build) which were still in the queue to free up some instances. Also i've triggered my own build at https://travis-ci.org/e1528532/libelektra/builds/295085816 like i did for the haskell dependency builds to hopefully speed up the process.

markus2330:
Thank you, your build went through!

PR #1531, opened 2017-07-08
Minor Enhancements

sanssecours:
# Purpose

This pull request contains various minor enhancements. I am not really sure if the reformatting of the code in commit 0614e57a is correct. It would be nice if some could check this particular update.

# Checklist

- [x] I did not run all tests, but assume everything will be fine anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,6 @@ before_script:
 
 script:
   - ninja
-  - travis_wait ninja run_all
+  - travis_wait 30 ninja run_all
   - ninja install
-  - travis_wait ninja run_all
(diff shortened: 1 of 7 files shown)


markus2330:
Thank you, these are great fixes!

About doc/tutorials/logging.md I was thinking about having it in doc/dev, but I can do that. Great to have something to build on now.

0614e57 is correct but I doubt that the example is a good one: Not too many know the C++ semantics for exceptions, so a different explanation/example would be better. You do not have to preserve everything in the docu as is: if you have other ideas how to explain something, please try it out!

PR #1532, opened 2017-07-11
Static code checker fixes

e1528532:
# Purpose

Fix issues reported by static code checkers

# Checklist

- [ ] I added unit tests

@markus2330 most of the issues have been minor things like unnecessary dead stores which i've removed, but there were also some possible null dereferences and obvious mistakes in some tests. If you find a change too drastic i can also just make it ignore some of the dead stores instead of removing them.
Furthermore i've made scan-build aware of our ELEKTRA_ASSERT to avoid some false positives and improved the documentation.
The only 4 reports left are in python.cpp of the python bindings, where it is not aware that SWIG_NewPointerObj takes care about freeing the allocated memory. I can supress the warnings though that requires some defines around the affected code. Is it even necessary?

### DIFF
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -225,10 +225,15 @@ of the issues found in the whole project.
 with LLVM/clang and is also primarily intended for C and C++ code. On macOS you have to install the 
 package `llvm` with homebrew, then you'll find the tool in the folder `/usr/local/opt/llvm/bin/`.
 
-To use it, prefix the compilation make command with `scan-build` in the build folder and specify an
-output folder for the results:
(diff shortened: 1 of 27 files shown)


e1528532:
jenkins build please

markus2330:
Thank you, great work! Looks nice.

I think we can accept all parts of the PR. But it seems like the build jobs have problems with _has_feature.

> I can supress the warnings though that requires some defines around the affected code. Is it even necessary?

Only if we integrate it in the build server.

e1528532:
looks good i think, apart from this non related issue

markus2330:
Thank you, great work!

PR #1533, opened 2017-07-12
Opmphm Api Finalization + small fixes

KurtMi:
# Purpose

- Fix small stufff at `generateKeySet ()` and draw-opmphm.
- Opmphm Api Finalization

# Checklist

- [x] I added unit tests

@markus2330 please review my pull request and please answer the following question in the comments,
my next PR will be the integration of the OPMPHM in elektra, but this is a big step and leaving a PR, which is concerning the core of elektra, open for long time is not a good idea, so is it alright when I mark the feature as non functioning at this time and make sure it can not be used accidentally?

@mpranj please have a look at the `Opmphm` struct in `kdbopmphm.h` and let me know if it does not suits you!

### DIFF
--- a/doc/dev/data-structures.md
+++ b/doc/dev/data-structures.md
@@ -362,6 +362,37 @@ synchronised.
 
 Continue reading [with the error handling](error-handling.md).
 
-## Order Preserving Minimal Perfect Hash Map
+## Order Preserving Minimal Perfect Hash Map (aka OPMPHM)
 
 All structs are defined in [opmphm.h](/src/include/kdbopmphm.h).
+
+Generally the OPMPHM is a arbitrary function.
+
+The desired return value aka the order is set in `OpmphmOrder->index.p`,
+additionally a minimum and maximum order is needed in `OpmphmInit->minOrder`
+and `OpmphmInit->maxOrder`.
+
+Due to the nature of the OPMPHM all elements need to be known at build time.
+
+The Build consists of three steps.
+
+### The Initialization
+
+Set the data `OpmphmInit->data` and the String extraction function `OpmphmInit->getString`.
+Provide a good seed in `OpmphmInit->initSeed`, needed in the next step. Call the `opmphmNewOrder ()`
+and use the default order or your own order. Min and Max order also need to be set respectively.
+
+The `opmphmInit ()`, transforms the orders in a internal representation, this step
+is irreversible.
+
+### The Mapping
+
+The `opmphmMapping ()` uses your seed (the OpmphmInit->seed will be changed) and tries to map your elements
+to a fixed space, this mapping might not succeed, on duplicate just call it again.
+
+### The Build
+
+The `opmphmBuild ()` function calculates the final OPMPHM.
+
+After the build the OpmphmInit and OpmphmOrders can be securely freed.
+The OPMPHM is now ready for constant lookups with the `opmphmLookup ()`.
--- a/src/include/kdbopmphm.h
+++ b/src/include/kdbopmphm.h
@@ -22,9 +22,14 @@
  * opmphmRatio should never be less than 1.
  *
  */
-#define OPMPHMTUPLE 7
+#define OPMPHMTUPLE 5
 extern double opmphmRatio;
 
+/**
+ * Typedefs used to reduce the memory footprint
+ */
+
+typedef size_t opmphmTranssition_t;
 
 /**
  * Saves a k-tuple in h[] and tells the OPMPHM the return value of each element, only needed during build.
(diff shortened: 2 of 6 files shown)


mpranj:
@KurtMi the size `OPMPHMTUPLE ` will be known at compile time like in the example or will this change to be dynamic? The KeySets will then each simply have a pointer to the `Opmphm` struct or how will it affect Key or KeySet?

Either way I don't think this will introduce any kind of problem for my mmap-storage implementation.

KurtMi:
> @KurtMi the size OPMPHMTUPLE will be known at compile time like in the example or will this change to be dynamic? The KeySets will then each simply have a pointer to the Opmphm struct or how will it affect Key or KeySet?

`OPMPHMTUPLE` is a constant known at compile time and each KeySet gets a pointer to the Opmphm struct.

markus2330:
> my next PR will be the integration of the OPMPHM in elektra, but this is a big step and leaving a PR, which is concerning the core of elektra, open for long time is not a good idea, so is it alright when I mark the feature as non functioning at this time and make sure it can not be used accidentally?

Yes, it should be enough to disable ENABLE_OPTIMIZATIONS by default, mark it as advanced option, and write a clear docu that it currently does not work (as long not all unit tests pass). But please do not change anything in `kdb.h` (if you need flags for ksLookup, please put them to `kdbproposal.h` first.

If you want you can setup a build job where ENABLE_OPTIMIZATIONS is activated.

src/include/kdbopmphm.h
```diff
@@ -22,17 +22,25 @@
  * opmphmRatio should never be less than 1.
```
markus2330: Could be checked via macro.

src/include/kdbopmphm.h
```diff
@@ -22,17 +22,25 @@
...
  *
  */
-#define OPMPHMTUPLE 7
+#define OPMPHMTUPLE 5
```
markus2330: Could you please give rationales for how you came up with the number. The description says what it does, but not why you chose the specific number.

doc/dev/data-structures.md
```diff
@@ -362,6 +362,37 @@ synchronised.
...
 
 All structs are defined in [opmphm.h](/src/include/kdbopmphm.h).
+
+Generally the OPMPHM is a arbitrary function.
+
+The desired return value aka the order is set in `OpmphmOrder->index.p`,
+additionally a minimum and maximum order is needed in `OpmphmInit->minOrder`
+and `OpmphmInit->maxOrder`.
+
+Due to the nature of the OPMPHM all elements need to be known at build time.
```
markus2330: Define what "build time" is, maybe give all phases a name or simply call them like the functions are named.

doc/dev/data-structures.md
```diff
@@ -362,6 +362,37 @@ synchronised.
 
 Continue reading [with the error handling](error-handling.md).
 
-## Order Preserving Minimal Perfect Hash Map
+## Order Preserving Minimal Perfect Hash Map (aka OPMPHM)
 
 All structs are defined in [opmphm.h](/src/include/kdbopmphm.h).
+
+Generally the OPMPHM is a arbitrary function.
```
markus2330: Do you mean "OPMPHM is a general function"? The sentence currently sounds like it would be okay if OPMPHM calculates Fibonacci numbers.

doc/dev/data-structures.md
```diff
@@ -362,6 +362,37 @@ synchronised.
...
+
+The desired return value aka the order is set in `OpmphmOrder->index.p`,
```
markus2330: The text goes quickly quite deep, maybe a bit more introduction?

doc/dev/data-structures.md
```diff
@@ -362,6 +362,37 @@ synchronised.
...
+### The Mapping
+
+The `opmphmMapping ()` uses your seed (the OpmphmInit->seed will be changed) and tries to map your elements
+to a fixed space, this mapping might not succeed, on duplicate just call it again.
+
+### The Build
+
+The `opmphmBuild ()` function calculates the final OPMPHM.
+
+After the build the OpmphmInit and OpmphmOrders can be securely freed.
```
markus2330: "securely freed" -> "should be freed"

src/include/kdbopmphm.h
KurtMi: For now the constant `OPMPHMTUPLE` is only a random number, after the evaluation of the benchmark results it will be final.

src/include/kdbopmphm.h
KurtMi: I am not sure how to check this via macro.

markus2330:
```
#if opmphmRatio < 1
# error "opmphmRatio should never be less than 1."
#endif
```

KurtMi: thank you!

KurtMi: `opmphmRatio` is not a `#define`. It is a c variable. I think the preprocessor has no knowledge of c vars.

markus2330: Seems like the C language is quite limited with its static asserts (In C++ with static_assert it works). The check is not really important (its highly unlikely that anyone will ever touch your internal constants), the statement was rather a general recommendation to use formal checks if possible. In general please avoid non-const global variables and defines. Please consider if OPMPHMTUPLE could be `const int` and `opmphmRatio` `const double`?

markus2330: Btw. of course is getting the code correct, fast, and elegant much more important than these little things. This is why I merged it immediately. Simply continue your work on the main parts...

PR #1536, opened 2017-07-13
Cppcheck fixes

e1528532:
# Purpose

Aims to resolve most of the warnings raised by cppcheck. Seems to contain further possible null check, resource leaks and similar not detected by scan-build, so its good to use both.

# Checklist

needs to be checked.

- [X] I added unit tests

@markus2330 WIP, still some outstanding issues i have to check. Also depends on static-code-checker-fixes PR, as some issues where detected by both to avoid overlapping.
Also looks like it somehow included the commits of the other PR too? (Initially i branched from master but then realized it makes more sense to be on top of the other PR so i rebased it. Shouldn't matter though)


markus2330:
Yes, seems to contain duplication of the other commits. Please rebase, otherwise it is hard to do the code review.

e1528532:
there are quite some cases where the scope of a variable can be reduced (usually if its only used in a block).

I see very mixed styles regarding variable declarations thought. In some files they are declared as late as possible as its also written in our coding guidelines (https://www.libelektra.org/devgettingstarted/coding), but in other files it seems to use the style where it declares them on top.
I'd move them inside blocks at least where cppcheck says its safe to reduce the scope, also for methods that use a declaration-first approach, is that ok? 

Also, for c++ it suggests to declare the one-argument constructors of classes as static, to avoid implicit conversations. I guess in most of our use cases this would be the cleaner approach (for exception classes and so on)?

src/bindings/cpp/examples/cpp_example_hierarchy.cpp
```diff
@@ -70,7 +70,7 @@ unsigned long name_depth (Key k)
 class Visitor
 {
 public:
-	virtual void visit (std::string name, unsigned long depth, Key k) = 0;
+	virtual void visit (std::string const & name, unsigned long depth, Key k) = 0;
```
markus2330: Does cppcheck already know about [C++11](https://juanchopanzacpp.wordpress.com/2014/05/11/want-speed-dont-always-pass-by-value/)?

markus2330:
> I'd move them inside blocks at least where cppcheck says its safe to reduce the scope, also for methods that use a declaration-first approach, is that ok?

Yes, this would be perfect! (and seems to be already done)

> Also, for c++ it suggests to declare the one-argument constructors of classes as static, to avoid implicit conversations. I guess in most of our use cases this would be the cleaner approach (for exception classes and so on)?

Do you mean declare them as `explicit`? Yes, this is an excellent idea. Can you add this to the coding guidelines?

e1528532:
> Do you mean declare them as explicit? Yes, this is an excellent idea. Can you add this to the coding guidelines?

It's basically already contained indirectly, as its mentioned in the Cpp Core Guidelines we linked in the CODING.md . I'm not sure if it is necessary to explicitly mention it again, as its probably not that a big deal.

src/bindings/cpp/examples/cpp_example_hierarchy.cpp
e1528532: Yes they do support some c++11 things too, though maybe not this one. But as the article mentions it seems to depend on the use case anyway which variant is better, so i think sticking to the usual const reference should be fine in general.

markus2330: Exactly this "in general" is not true anymore. So better not blindly replace the one with the other. You can keep the current changes if it is work to undo them.

e1528532: ok then lets keep them

markus2330:
Thank you, these are great quality improvements!

PR #1539, opened 2017-07-15
Canonical: Preview

tom-wa:
# Purpose

Preview: Canonicalization plugin - matches keyvalue against either a list of values, posix extended regex or fnmatch and on a match brings the value to a canonical form

Some examples are in README.md

Pending decisions:
- move stringToArray helper function to a library ? 
- compilation variants for regex, lists and fnmatch ?
- per mountpoint configurations ? 

TODO:
- [ ] error handling
- [ ] cleanup
- [ ] Documentation

# Checklist

- [x] I added unit tests

### DIFF
--- a/src/plugins/canonical/canonical.c
+++ b/src/plugins/canonical/canonical.c
@@ -1,0 +1,365 @@
+/**
+ * @file
+ *
+ * @brief Source for canonical plugin
+ *
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ *
+ */
+
+#include "canonical.h"
+
+#include <fnmatch.h>
+#include <kdbhelper.h>
+#include <kdbmeta.h>
+#include <kdbutility.h>
+#include <regex.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+
+
+int elektraCanonicalOpen (Plugin * handle ELEKTRA_UNUSED, Key * errorKey ELEKTRA_UNUSED)
+{
+	// plugin initialization logic
+	// this function is optional
+
+	return ELEKTRA_PLUGIN_STATUS_SUCCESS;
+}
+
+int elektraCanonicalClose (Plugin * handle ELEKTRA_UNUSED, Key * errorKey ELEKTRA_UNUSED)
+{
+	// free all plugin resources and shut it down
+	// this function is optional
+
+	return ELEKTRA_PLUGIN_STATUS_SUCCESS;
+}
+
+static char ** stringToArray (const char * string, const char * delim)
+{
+	if (!string || !delim) return NULL;
+
+	char ** array = NULL;
+	int elems = 1;
+	const char * ptr = string;
+
+	while (*ptr)
+	{
+		if (*ptr == *delim)
+		{
+			++elems;
+		}
+		++ptr;
+	}
+	++elems; // terminating element
+	array = elektraCalloc (elems * sizeof (char *));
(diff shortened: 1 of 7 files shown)


src/plugins/canonical/canonical.c
```diff
@@ -0,0 +1,337 @@
...
+		{
+			++elems;
+		}
+		++ptr;
+	}
+	++elems; // terminating element
+	array = elektraCalloc (elems * sizeof (char *));
+	char * localString = NULL;
+	if (string[0] == '[' || string[0] == '(' || string[0] == '{')
+		localString = elektraStrDup (string + 1);
```
markus2330: minor: you could avoid the else branch by calculating an offset (which is either 0 or 1)

Please prefer to use `{}`.

src/plugins/canonical/canonical.c
```diff
@@ -0,0 +1,337 @@
...
+	if (lastChar == ']' || lastChar == ')' || lastChar == '}') localString[strlen (localString) - 1] = '\0';
+
+	int current = 0;
+	ptr = strtok (localString, delim);
+	while (ptr)
+	{
+		if (*ptr == '\0')
+		{
+			ptr = strtok (NULL, delim);
+			continue;
```
markus2330: minor: you could easily avoid continue by putting the rest in the else branch.

src/plugins/canonical/canonical.c
```diff
@@ -0,0 +1,337 @@
...
+		}
+		char * tmp = elektraStrDup (ptr);
+		char * start = elektraLskip (tmp);
+		(void)elektraRstrip (start, NULL);
+		if (start[0] == '\'') ++start;
+		if (start[strlen (start) - 1] == '\'') start[strlen (start) - 1] = '\0';
```
markus2330: Please use `{}`, its easy to get confused here where `[]()` start/stop.

tom-wa:
>What about having a preconfigured way to get canonicalization like done in the plugin bool? 

i'm currently trying to figure out which approach of preconfiguring works best: plugin configuration or the way typedispatcher currently does it (mounting a configuration and letting the plugin read it)
the latter one seems easier for everything thats more than 3 configuration keys to set.
should we give plugins a way to automatically fetch configurations below some specified mountpoint on kdbOpen ?

markus2330:
tom-wa wrote:
> i'm currently trying to figure out which approach of preconfiguring works
> best: plugin configuration or the way typedispatcher currently does it
> (mounting a configuration and letting the plugin read it) the latter one
> seems easier for everything thats more than 3 configuration keys to set.
> should we give plugins a way to automatically fetch configurations below
> some specified mountpoint on kdbOpen ?

It would be a great idea to allow plugin configs to be mounted. But it should 
not be done using kdbOpen within the plugins---we would need framework 
extensions to do this properly. And I am afraid that is out of scope for 
lcdproc or even 1.0.

For now the best solutions would be:

1. having commonly used options hard coded (such as the suggested bool 
   canonicalization, which is likely the one which is used at most by far (for 
   the tristate bool)). Your current canonical plugin config would allow such
   hard-coded plugin settings well (as option next to the regex, fnmatch, 
   list..)
2. having a type/spec-dispatcher-like plugin which copies metadata from one 
   place to all other places where needed

If the type dispatcher is already capable to copy the canonical metadata to 
all places which use the same type it would be a nice solution. Otherwise I 
would refain from bloating the type plugin even more, the review is quite hard 
because it is already a monster.

markus2330:

PR #1544, opened 2017-07-19
YAML Plugin: First Version

sanssecours:
# Purpose

This pull request adds a **very basic** YAML plugin. I wrote the parsing code myself, since I wanted to learn how to “properly” write a recursive descent parser. While the code should not be too hard to extend (see commit commit 17aa7a6 for an example), I do not think that it makes sense to continue with the development of this parser, since it currently has multiple issues that I listed below.

- **No Scanner:**  While I prefer a scanner-less parser (no distinction between lexing and parsing), this decision might be problematic. I assume that some advanced YAML features require a lookahead of one symbol and a one character (byte) lookahead is not enough. 
- **Ignores Whitespace:** This is problematic since YAML’s block style uses additional leading spaces to denote child nodes.
- **Error Handling:** A lot of the code currently consists of checks for errors. Unfortunately C does not support exceptions which would make the code much cleaner.

Anyway, I guess I will remove most of the parsing code later and just use [libyaml](http://pyyaml.org/wiki/LibYAML) or something similar for my example of a hand-written parser.

# Checklist

- [x] Not a single one of the included commits closes an open issue :o).
- [x] I ran all tests and only the usual test cases:
   - `testjna_maven`,
   - `test_kdb.lua`,
   - `test_key.lua`,
   - `test_keyset.lua`
  failed, since I enabled the address sanitizer.
- [x] I added unit tests.
- [x] I added code comments, logging, and assertions.
- [x] Commit 2a0f94b includes a minor update to the specification.

### DIFF
--- a/.oclint
+++ b/.oclint
@@ -1,3 +1,5 @@
 rule-configurations:
   - key: LONG_LINE
     value: 140
+disable-rules:
+  - GotoStatement
--- a/doc/decisions/validation.md
+++ b/doc/decisions/validation.md
@@ -2,15 +2,15 @@
 
 ## Issue
 
-Validation plugins operate as indenpendent blackboxes. 
-For every backend each mounted validation plugin iterates 
-over the whole keyset, checks every key for its trigger metakey, 
+Validation plugins operate as indenpendent blackboxes.
+For every backend each mounted validation plugin iterates
+over the whole keyset, checks every key for its trigger metakey,
 and validates the key.
 
-Currently all needed validation plugins need to be specified at 
-mount-time - if additional validation is required, the backend 
-has to be remounted with the required plugins and plugin 
-configuration. 
+Currently all needed validation plugins need to be specified at
+mount-time - if additional validation is required, the backend
+has to be remounted with the required plugins and plugin
+configuration.
 
 If validation of a key fails, each plugin decides on its own
 how to handle the issue and proceed in ways that might be
@@ -21,7 +21,7 @@ different than what is expected or desired.
 
 ## Assumptions
 
-While plugins should always fail and return an error if validation 
+While plugins should always fail and return an error if validation
 fails on kdbSet, there are be can be several different requirements
 for what should happen on kdbGet and handle problems e.g.
 
@@ -36,7 +36,7 @@ for what should happen on kdbGet and handle problems e.g.
 
   we want to read the whole configuration, but drop invalid keys
 
-  invalid keys might be replaced by default values, requested 
+  invalid keys might be replaced by default values, requested
   from the user, ...
 
 - fail with error
@@ -47,15 +47,15 @@ for what should happen on kdbGet and handle problems e.g.
 
 ## Considered Alternatives
 
-- Extend validation plugins to allow us to specify what should happen 
+- Extend validation plugins to allow us to specify what should happen
   if a key fails to validate
-- Export a validation function that allows us to use an additional plugin 
+- Export a validation function that allows us to use an additional plugin
   to decide what should be done
 
 
 ## Decision
 
-Use a wapper plugin to iterate over the keyset and delegate the validation 
+Use a wapper plugin to iterate over the keyset and delegate the validation
 of each key to the corresponding validation plugin.
 
 ## Argument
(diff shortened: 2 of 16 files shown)


markus2330:
I really love your clean and simplistic style!

I fully agree that there is not much sense to continue this development, better to use some finished library for further comparison.

> Anyway, I guess I will remove most of the parsing code later and just use libyaml or something similar for my example of a hand-written parser.

It would be okay to have different plugins here. You rated this one low enough (via `infos/status`) that users will not come in contact with it. (We should implement a "hide" feature for too low rated plugins in `kdb list`.) The plugin might be nevertheless interesting for people who dislike the dep of `libyaml`. If we find out it is slow and buggy, we can remove it later anyway.

> I ran all tests and only the usual test cases: 

Are all of these already reported?

> setErrorErrno

We should integrate such functionality directly into `ELEKTRA_SET_ERROR*` `ELEKTRA_ADD_WARING*` (errno number+text as extra fields). So before you try to reuse your setErrorErrno between plugins, I would prefer to fix it for all code. Please create an issue if you are confronted with the problem and do not know how to extend the code generator.

Some other macros you have maybe should be shared, too. (I am looking at `ASSERT_NOT_NULL` and `SET_ERROR_MALLOC`, it would need a rename, though)

Otherwise I found in the review only a maybe-unnecessary optimization improvement: In yaml.c:470 you could avoid reallocation if there is already enough allocated space.

sanssecours:
>> I ran all tests and only the usual test cases: 

> Are all of these already reported?

I added [a comment](https://github.com/ElektraInitiative/libelektra/issues/788#issuecomment-316726141) about the failing tests to issue #788.

> We should integrate such functionality directly into `ELEKTRA_SET_ERROR*` `ELEKTRA_ADD_WARING*` (errno number+text as extra fields).

Sounds like a good idea.

> Some other macros you have maybe should be shared, too. (I am looking at `ASSERT_NOT_NULL` and `SET_ERROR_MALLOC`, it would need a rename, though)

How about `ELEKTRA_ASSERT_NOT_NULL` and `ELEKTRA_SET_ERROR_MALLOC`?

> In yaml.c:470 you could avoid reallocation if there is already enough allocated space.

Yeah, I know :o). However handling both `malloc` and `realloc` would increase the code complexity, at least as far as I can tell. When I am confronted with a decision about cleaner or faster code I usually prefer cleaner code. Although I also wrote a lot of code in my life that is neither 😄.

markus2330:
> `ELEKTRA_ASSERT_NOT_NULL` and `ELEKTRA_SET_ERROR_MALLOC`?

Yes, I like the suggestions. Inversing it `ELEKTRA_NULL_ASSERT` (`NOT` might be not needed?) and `ELEKTRA_MALLOC_ERROR` (`SET` might be not needed if reversed?) might be more completion friendly, but is less consistent.

> However handling both malloc and realloc would increase the code complexity

Yes, it does and without benchmarking we do not know if this optimization is relevant.

sanssecours:
`ELEKTRA_MALLOC_ERROR` sound fine. I am not a fan of `ELEKTRA_NULL_ASSERT` though. To me that sounds like we want to assert that a variable contains `NULL` instead of the opposite. 

Another option, instead of `ELEKTRA_NOT_NULL`, would be `ELEKTRA_ASSERT_TRUE`, which I do not like at all. `ELEKTRA_ASSERT_DEFINED` sounds better, but I still prefer `ELEKTRA_NOT_NULL`, since the term “defined” is not quite the opposite of “null“.

markus2330:
René Schwaiger wrote:
> `ELEKTRA_MALLOC_ERROR` sound fine.

Ok!

Another idea would be to pass parentKey to elektraMalloc etc. and set the 
error there at one place. But as said, this is a post-1.0 goal.

> I am not a fan of `ELEKTRA_NULL_ASSERT`
> though. To me that sounds like we want to assert that a variable contains
> `NULL` instead of the opposite. 
> 
> Another option, instead of `ELEKTRA_NOT_NULL`, would be
> `ELEKTRA_ASSERT_TRUE`, which I do not like at all. `ELEKTRA_ASSERT_DEFINED`
> sounds better, but I still prefer `ELEKTRA_NOT_NULL`, since the term
> “defined” is not quite the opposite of “null“.

I agree, `ELEKTRA_ASSERT_TRUE` is more of what you would expect in test 
frameworks. Then the error message would be too vague. (Because we want to 
assert against null pointers, not if something is true).

I am fine with `ELEKTRA_NOT_NULL`. Its internal, so we can search&replace it 
if someone comes up with a better name.

PR #1551, opened 2017-07-26
Run All: Use Non-Buffered Output

sanssecours:
# Purpose

Currently the output of `ninja run_all` is buffered. Consequently we do not see which sub-command causes problems if we terminated the script before it finishes. This behaviour is especially problematic, if Travis kills the script cause it runs for too long.

# Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,6 @@ before_script:
 
 script:
   - ninja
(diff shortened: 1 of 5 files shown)


markus2330:
Thank you for taking care of these problems!

PR #1558, opened 2017-07-27
asan-fixes: further issues resolved #788

### LINKED ISSUE #788: Fix sanitizer issues

e1528532:
# Purpose

Fixes 2 further ASAN issues

# Checklist

@markus2330 There are still some remaining on our asan build. I am currently trying to hunt down a global buffer overflow in the testmod_ni. 

Furthermore there are issues with xerces and botan, where i am not sure if these are actual issues. It boils down to a lot of messages like

```
/home/jenkins/workspace/workspace/elektra-clang-asan/src/plugins/crypto/botan_operations.cpp:125:25: runtime error: member call on address 0x60200001d8d0 which does not point to an object of type 'Botan::PBKDF'
0x60200001d8d0: note: object is of type 'Botan::PKCS5_PBKDF2'
 0b 00 00 00  90 84 d1 a2 b1 7f 00 00  a0 7c 01 00 80 60 00 00  03 00 00 00 00 00 00 02  04 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'Botan::PKCS5_PBKDF2'
```

Looks like it has an issue with polymorphic calls (PBKDF is the base type, the other is the subclass of it), i am not sure what we can do against that because judging from the examples this seems to be the way this library is used. 

Xerces has the same issue, they also have an object oriented design with interfaces and implementations and asan complains when you call methods against the interface and reports a leak then.

I tried quite a lot but in the end without doing very ugly casting hacks which can't be the way those libs are designed to be used i don't get around the leaks.

### DIFF
--- a/src/libs/loader/dl.c
+++ b/src/libs/loader/dl.c
@@ -53,9 +53,9 @@ int elektraModulesInit (KeySet * modules, Key * error ELEKTRA_UNUSED)
 elektraPluginFactory elektraModulesLoad (KeySet * modules, const char * name, Key * errorKey)
 {
 #ifdef _WIN32
-	const char elektraPluginPostfix[] = ".dll";
+	static const char elektraPluginPostfix[] = ".dll";
 #else
-	const char elektraPluginPostfix[] = ".so";
+	static const char elektraPluginPostfix[] = ".so";
 #endif
 
 	Key * moduleKey = keyNew ("system/elektra/modules", KEY_END);
--- a/src/plugins/crypto/botan_operations.cpp
+++ b/src/plugins/crypto/botan_operations.cpp
@@ -9,10 +9,12 @@
 
 #include <botan/auto_rng.h>
 #include <botan/filters.h>
+#include <botan/hmac.h>
 #include <botan/init.h>
 #include <botan/lookup.h>
 #include <botan/pbkdf2.h>
 #include <botan/pipe.h>
+#include <botan/sha2_32.h>
 #include <botan/symkey.h>
 #include <kdbplugin.h>
 #include <memory>
@@ -70,8 +72,8 @@ static int getKeyIvForEncryption (KeySet * config, Key * errorKey, Key * masterK
 		const kdb_unsigned_long_t iterations = CRYPTO_PLUGIN_FUNCTION (getIterationCount) (errorKey, config);
 
 		// generate/derive the cryptographic key and the IV
-		unique_ptr<PBKDF> pbkdf = unique_ptr<PBKDF> (get_pbkdf ("PBKDF2(SHA-256)"));
-		OctetString derived = pbkdf->derive_key (
+		PKCS5_PBKDF2 pbkdf (new HMAC (new SHA_256));
+		OctetString derived = pbkdf.derive_key (
 			requiredKeyBytes, std::string (reinterpret_cast<const char *> (keyValue (masterKey)), keyGetValueSize (masterKey)),
 			salt, sizeof (salt), iterations);
 
@@ -121,8 +123,8 @@ static int getKeyIvForDecryption (KeySet * config, Key * errorKey, Key * masterK
 	try
 	{
 		// derive the cryptographic key and the IV
-		unique_ptr<PBKDF> pbkdf = unique_ptr<PBKDF> (get_pbkdf ("PBKDF2(SHA-256)"));
-		OctetString derived = pbkdf->derive_key (
+		PKCS5_PBKDF2 pbkdf (new HMAC (new SHA_256));
+		OctetString derived = pbkdf.derive_key (
 			requiredKeyBytes, std::string (reinterpret_cast<const char *> (keyValue (masterKey)), keyGetValueSize (masterKey)),
 			saltBuffer, saltBufferLen, iterations);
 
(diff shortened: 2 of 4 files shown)


markus2330:
Thank you for looking into it: @petermax2 what do you think?

petermax2:
>  judging from the examples this seems to be the way this library is used.
Exactly. I could not find another way of using the library in Botan's documentation (also checked their doxygen doc when I wrote the plugin).

markus2330:
Do we have to add even more casts to get rid of ASAN problems?

Btw. ASAN is for us mostly because of undefined behavior, the leak check is already done by valgrind. So maybe its best to turn the leak check off in ASAN? Or can we suppress the wrongly states leaks?

petermax2:
I just browsed through the source code of Botan and found that they use PBKDF like

    49     PKCS5_PBKDF2 pbkdf(new HMAC(new SHA_512));                                                                 
      1                                                                                                               
      2    OctetString master_key = pbkdf.derive_key(                                                                 
      3       PBKDF_OUTPUT_LEN,                                                                                       
      4       passphrase,                                                                                             
      5       pbkdf_salt.data(),                                                                                      
      6       pbkdf_salt.size(),                                                                                      
      7       PBKDF_ITERATIONS);                                                                                      
 
instead of `get_pbkdf("<algorithm>")` as demonstrated in the example.

Maybe this helps us in getting rid of the problems?

petermax2:
#1561 did not fix the issues with Botan. I can't reproduce the errors in `crypto_botan` locally with openSUSE and libbotan-1.10.10.

markus2330:
Okay, I think we should go for suppression or completely deactivate the memory checker  (afaik they cannot find anything valgrind does not find already).

sanssecours:
> Okay, I think we should go for suppression…

I think this is the way to go.

> …or completely deactivate the memory checker (afaik they cannot find anything valgrind does not find already).

Since `valgrind` only partially works under macOS I would prefer we use MemorySanitizer too. Although MemorySanitizer currently [only works on Linux](https://github.com/google/sanitizers/wiki/MemorySanitizer#getting-memorysanitizer) I would assume that, if someone decides to port the code to macOS, the Mac support would be much better than the one provided by Valgrind.

markus2330:
> the Mac support would be much better than the one provided by Valgrind.

That is speculative, maybe they run into the same troubles.

Then @e1528532 can decide. If the suppression is too much effort, from my side it is okay to deactivate the memory checker. Once MacOSX support is available we will obviously reconsider and can activate it again.

e1528532:
We can use the already existing blacklist in tests/sanitizer.blacklist to exclude the xerces and botan source files from asan.
This way the memory checking is still active for the remaining code.
However, it is up to us obviously to guarantee xerces/botans correctness then. According to valgrind they seem fine / no leaks. But if we modify the library we should maybe recheck to see if some new issue got introduced by the changes. And i guess one would easily overlook that task.

Blacklisted the xerces and the botan issues in the latest commit. Also i used the fix in #1561 because otherwise asan would report a leak in c++'s unique_ptr.h file due to botan and i didn't want to blacklist that file as all other ways of supression didn't work.
Locally these went fine now obviously. There are probably still some more issues to fix, @markus2330 can we rerun the asan build job on this branch to see whats left?

markus2330:
jenkins build clang-asan please
jenkins build gcc-asan please

markus2330:
jenkins build clang-asan please

markus2330:
jenkins build gcc-asan please

e1528532:
So xerces and botan seem happy now. There are 2 issues remaining with testshell_markdown_base6 and testshell_markdown_tutorial_validation which i can't reproduce locally.
Could it be that its due to the gcc version? 
Judging from the line
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/4.9/../../../../include/c++/4.9/bits/stl_tree.h:741:25: runtime error: downcast of address 0x7ffef93e67b8 with insufficient space for an object of type '_Rb_tree_node<std::pair<const std::basic_string<char>, std::basic_string<char> > >'
0x7ffef93e67b8: note: pointer points here
 54 7f 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  b8 67 3e f9 fe 7f 00 00  b8 67 3e f9
```
it looks like gcc 4.9 is used. I have 7.2.0 locally, could it be that a newer gcc version fixes it?

sanssecours:
> Could it be that its due to the gcc version?

I would assume that those errors are caused by problems in external software. In the Linux Travis build jobs the ASAN problem you mentioned above usually occurs if we use `printf` in conjunction with `sed`. I already [disabled some of the tests](https://github.com/ElektraInitiative/libelektra/blob/master/tests/shell/shell_recorder/tutorial_wrapper/CMakeLists.txt#L39-L46) on ASAN enabled Linux systems.

Issue #1555, opened 2017-07-27
CMake: Fix Deprecated Code and Warnings


sanssecours:
# Steps to Reproduce the Problem

1. Clone the repository
2. Run CMake (Version `3.9.0` or later)

# Expected Result

CMake creates the build system without any warnings.

# Actual Result

CMake prints a warning about policy CMP0026,

```
CMake Deprecation Warning at CMakeLists.txt:16 (cmake_policy):
  The OLD behavior for policy CMP0026 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
```

about [policy CMP0053](https://travis-ci.org/ElektraInitiative/libelektra/jobs/257825569#L1410-L1455), if we include the Qt GUI:

```
-- Include Tool qt-gui
CMake Warning (dev) at /usr/local/opt/qt5/lib/cmake/Qt5/Qt5ModuleLocation.cmake:4 (set):
  Policy CMP0053 is not set: Simplify variable reference and escape sequence
  evaluation.  Run "cmake --help-policy CMP0053" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.
  For input:
    '${_qt5_install_prefix}/Qt5@module@/Qt5@module@Config.cmake'
  the old evaluation rules produce:
    '/usr/local/opt/qt5/lib/cmake/Qt5/Qt5Config.cmake'
  but the new evaluation rules produce:
    '/usr/local/opt/qt5/lib/cmake/Qt5@module@/Qt5@module@Config.cmake'
  Using the old result for compatibility since the policy is not set.
Call Stack (most recent call first):
  /usr/local/opt/qt5/lib/cmake/Qt5/Qt5Config.cmake:25 (include)
  src/tools/qt-gui/CMakeLists.txt:7 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
CMake Warning (dev) at /usr/local/opt/qt5/lib/cmake/Qt5/Qt5ModuleLocation.cmake:4 (set):
  Policy CMP0053 is not set: Simplify variable reference and escape sequence
  evaluation.  Run "cmake --help-policy CMP0053" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.
  For input:
[...]

PR #1559, opened 2017-07-27
CMake: Fix Deprecated Code

sanssecours:
# Purpose

This pull request tries to fix some of the warnings reported in issue #1555.

# Checklist

- [x] I did not run any tests and opened this pull request to test the updates instead ☺️.


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8.8)
+cmake_minimum_required (VERSION 3.0)
```
markus2330: I will think about if we do the cmake 3 version push this release or next one. For Debian it should be actually okay because it only effects oldoldstable.

CMakeLists.txt
```diff
@@ -11,21 +11,6 @@ set(CMAKE_MACOSX_RPATH 1)
...
-	cmake_policy(SET CMP0026 OLD) # reinvestigate at 2.8.11 or later
-endif()
-if (POLICY CMP0043)
-	cmake_policy(SET CMP0043 OLD) # to have consistent behavior for pre 2.8.10
-	#also reinvestigate when they do not need to be supported anymore
-endif()
-if (POLICY CMP0046)
-	cmake_policy(SET CMP0046 OLD) # new behaviour seems stupid, of course
-	# dependency is not yet available if it's generated later?
-endif()
```
markus2330: Good to see this gone, thanks for your effort.

cmake/Modules/LibAddPlugin.cmake
```diff
@@ -351,8 +351,6 @@ function (add_plugin PLUGIN_SHORT_NAME)
 	add_library (${PLUGIN_OBJS} OBJECT ${ARG_SOURCES})
 	add_dependencies (${PLUGIN_OBJS} kdberrors_generated)
 
-	add_dependencies(${PLUGIN_OBJS} readme_${PLUGIN_SHORT_NAME}.c)
```
markus2330: Are you sure this dep is not needed or did you remove it because it did not work?

I think the idea was to retrigger plugin compilation after `README.md` is changed.

doc/CMakeLists.txt
```diff
@@ -63,6 +63,10 @@ if (DOXYGEN_FOUND)
 		"${CMAKE_CURRENT_BINARY_DIR}/Doxyfile"
 		@ONLY
 	)
+	file(GENERATE
```
markus2330: Why is `configure_file` still needed?

CMakeLists.txt
sanssecours: > For Debian it should be actually okay because it only effects oldoldstable.

There is a Debian version that is even more ancient than `oldstable` 😱… Anyway, even the [official Debian release page](https://wiki.debian.org/DebianReleases) does not list `oldoldstable` under current releases. For now I just removed the commit that includes the version bump. I guess the current CMake files should still work, even if we use CMake 2.8 to generate the build system.

markus2330:
I removed "ready to merge" so that it is clear when I should merge it. Without the version pump we can merge it without further considerations (if it does not break something, lets see what the buildserver says). Please readd "ready to merge" when you are ready.

markus2330:

markus2330:

CMakeLists.txt
sanssecours: Looks like removing the version bump triggers warnings about policy `CMP0043`:

```
-- Configuring done
CMake Warning (dev) in src/tools/qt-gui/CMakeLists.txt:
  Policy CMP0043 is not set: Ignore COMPILE_DEFINITIONS_<Config> properties.
  Run "cmake --help-policy CMP0043" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) in src/tools/qt-gui/CMakeLists.txt:
  Policy CMP0043 is not set: Ignore COMPILE_DEFINITIONS_<Config> properties.
  Run "cmake --help-policy CMP0043" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) in src/tools/qt-gui/CMakeLists.txt:
  Policy CMP0043 is not set: Ignore COMPILE_DEFINITIONS_<Config> properties.
  Run "cmake --help-policy CMP0043" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

. No direct support for you then, [outdated ancient OS](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819703) 🙃.

cmake/Modules/LibAddPlugin.cmake
sanssecours: > Are you sure this dep is not needed or did you remove it because it did not work?

I deleted this line since it triggered warnings after I removed the setting for policy `CMP0046`.

> I think the idea was to retrigger plugin compilation after `README.md` is changed.

As far as I can tell, rebuilding works fine even after I delete the dependency. I tried the following:

1. Remove one of the commas in the example of the YAML ReadMe
2. Invoke the build system
3. The Markdown Shell Recorder correctly reports the syntax problem.

Can you think of an example where the rebuilding process does not work?

CMakeLists.txt
markus2330: Debian Wheezy uses [cmake 2.8.9-1](https://packages.debian.org/wheezy/cmake).

Seems testmod_date even failed before this PR on wheezy. Otherwise Elektra compiles and all test cases pass, so we should consider to do the version pump after the release. (Then we can also remove the wheezy build agents)

The warnings you get are likely only because of problems within Qt-CMake---nothing we should worry about. We do not use `COMPILE_DEFINITIONS_*`. Workaround: set the policy within `qt-gui`.

doc/CMakeLists.txt
sanssecours: We need the command, since `file(GENERATE …)` [**does not** expand variables](https://cmake.org/Bug/view.php?id=16056):

> file(GENERATE) does not document that it expands variables. It doesn't expand them in either case. In fact it is not well-defined to do so because the generation does not occur until after configuration is complete when the variable definitions may no longer be available.

.

cmake/Modules/LibAddPlugin.cmake
markus2330: > As far as I can tell, rebuilding works fine even after I delete the dependency.

It did not not even work before this PR, I reported the problem in #1560

Thus this PR does not change the behavior (it did not work before nor after), you do not have to worry about it.

CMakeLists.txt
markus2330: Just saw that we already announced in 2016-04-29_0.8.16.md

> This might be the last release supporting wheezy, because
it gets more and more time-intensive to find workarounds
for the old compiler. The C++11 regex do not work at all.

So actually you are right, we should not worry and finally drop the support (and pump the cmake version number).

sanssecours: > The warnings you get are likely only because of problems within Qt-CMake---nothing we should worry about. We do not use `COMPILE_DEFINITIONS_*`. Workaround: set the policy within `qt-gui`.

Oh, I guess I will remove the version bump then once again ☺️.

markus2330: Sorry, doing cmake stuff is thankless work. Its hard to know on which systems you break something and usually somebody is unhappy ;)

markus2330:
Thank you!

PR #1570, opened 2017-08-10
Opmphm Integration Part 1

KurtMi:
# Checklist

- [ ] I added unit tests


KurtMi:

KurtMi:
@markus2330  which cmake version is installed on `debian-stable-mm`, because it does not know  `continue()`.
```
CMake Error at tests/ctest/CMakeLists.txt:5 (continue):
  Unknown CMake command "continue".
```

KurtMi:

markus2330:
As you see in top-level "CMakeLists.txt" we want to support cmake 3.0 (this was increased recently in #1559).

So "continue" seems to be way too new (around cmake 3.2).

To answer your question nevertheless, on the build server it is cmake version 3.0.2, so it tests exactly what we want to support.

src/libs/elektra/keyset.c
```diff
@@ -2379,6 +2400,10 @@ int ksClose (KeySet * ks)
 
 	ks->size = 0;
 
+#ifdef ELEKTRA_ENABLE_OPTIMIZATIONS
+	opmphmDel (ks->opmphm);
```
markus2330: Better to check for null pointer in ks->opmphm. (opmphmDel has a non-documented precondition)

src/libs/elektra/keyset.c
```diff
@@ -2350,7 +2363,15 @@ int ksInit (KeySet * ks)
 
 	ksRewind (ks);
 
-	return 1;
+#ifdef ELEKTRA_ENABLE_OPTIMIZATIONS
+	ks->opmphm = opmphmNew ();
+	if (!ks->opmphm)
+	{
+		return -1;
```
markus2330: Also here: what about not aborting here?

src/libs/elektra/keyset.c
```diff
@@ -380,7 +384,15 @@ int ksDel (KeySet * ks)
...
 {
 	ksClose (ks);
-	// ks->array empty now
+// ks->array empty now
+
+#ifdef ELEKTRA_ENABLE_OPTIMIZATIONS
+	ks->opmphm = opmphmNew ();
+	if (!ks->opmphm)
+	{
+		return -1;
```
markus2330: Is it sensible to continue here? (If there was enough memory for array, but not for opmphm? Array should be allocated first, though)

Btw. the current implementation of opmphmNew uses Malloc although Calloc would be better suitable?

KurtMi:

src/include/kdbrand.h
```diff
@@ -20,6 +20,12 @@ extern "C" {
 #define ELEKTRARANDMAX 2147483647
 void elektraRand (int32_t * seed);
 
+#ifdef KDBRAND_BENCHMARK
+extern int32_t elektraRandBenchmarkInitSeed;
```
markus2330: Please avoid global variable.

src/libs/elektra/keyset.c
KurtMi: Currently I defined that the `opmphm` is empty when all `size[]` are zero, I see no reason to use `calloc()`, since the `opmphmNew()` sets the sizes to zero. All other fields of the Opmphm get initialized during build.

src/include/kdbrand.h
KurtMi: I know, but I see no better way to influence the initial seed when the `kslookup ()` makes the Opmphm.

KurtMi:

src/include/kdbrand.h
markus2330: What about storing it in the Opmphm struct? Do you seed every Opmphm differently?

markus2330:
Thank you, looking forward to see Part 2 soon!

Issue #1395, opened 2017-03-01
lcdproc: size


markus2330:
As @tom-wa found, lcdproc intensively uses `Size = WIDTHxHEIGHT` (and also some other `*Size`), i.e. structure within individual values.

This causes in lcdproc a substantial amount of `sscanf` parsing with many checks if individual parts are within bounds. This leads to postponed errors within the modules and might not be checked at all if a module is currently inactive.

We could write a simple transformation plugin that splits these `WIDTHxHEIGHT` types into `size/width` and `size/height` respectively and check the sizes as with other ints.

So something like:
```
[CFontz/Size]
transform/size

[CFontz/Size/Width]
default = 20
range = ...
type = long

[CFontz/Size/Height]
default = 4
range = ...
type = long
```

Within the code the user would use:
```
elektraGetLong (ELEKTRA_INSTANCE, "CFontz/Size/Width")
elektraGetLong (ELEKTRA_INSTANCE, "CFontz/Size/Height");
```

or if the user prefers the code generator:

```
elektraGetCFontzSizeWidth(ELEKTRA_INSTANCE);
elektraGetCFontzSizeHeight(ELEKTRA_INSTANCE);
```

What do you think?

(Discussions about ELEKTRA_INSTANCE in #1378)

markus2330:
not relevant

PR #1574, opened 2017-08-16
size plugin

tom-wa:
# Purpose

size plugin  #1395

# Checklist

- [x] I added unit tests

### DIFF
--- a/src/plugins/size/README.md
+++ b/src/plugins/size/README.md
@@ -1,0 +1,41 @@
+- infos = Information about the size plugin is in keys below
+- infos/author = Thomas Waser <thomas.waser@libelektra.org>
+- infos/licence = BSD
+- infos/needs =
+- infos/provides = conv
+- infos/recommends =
+- infos/placements = postgetstorage presetstorage
+- infos/status = maintained conformant compatible coverage specific unittest shelltest tested nodep libc configurable experimental
+- infos/metadata = transform/size
+- infos/description = transforms WIDTHxHEIGHT values into separate keys
+
+## Introduction
+
+The `size` plugin transforms values with the format `WIDTH<separator>HEIGHT` into a separate `Width` and `Height` key. The metakey `transform/size` triggers the conversion.
+
+## Configuration
+
+By default `x` is assumed as separator. If the plugin configuration `/separator` is set, its value is used instead. If set, `transform/size`'s value takes precedence.
+
+## Examples
+
+```sh
+# Backup-and-Restore:/examples/size
+
+sudo kdb mount size.ecf /examples/size size dump
+
+kdb set /examples/size/SizeKey 20x4
+kdb setmeta user/examples/size/SizeKey transform/size ""
+
+kdb get /examples/size/SizeKey/Width
+#> 20
+kdb get /examples/size/SizeKey/Height
+#> 4
+
+kdb rm -r /examples/size
+sudo kdb umount /examples/size
+```
+
+## Limitations
+
+None.
--- a/src/plugins/size/size.c
+++ b/src/plugins/size/size.c
@@ -1,0 +1,140 @@
+/**
+ * @file
+ *
+ * @brief Source for size plugin
+ *
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ *
+ */
+
+#include "size.h"
+
+#include <kdbhelper.h>
+#include <stdio.h>
+#include <string.h>
+
+
+#define DEFAULT_SEPARATOR "x"
+
+static int transform (KeySet * appendKS, const Key * key, const char * sep)
+{
+	char * value = elektraStrDup (keyString (key));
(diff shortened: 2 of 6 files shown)


src/plugins/size/size.c
```diff
@@ -0,0 +1,140 @@
...
+
+
+#define DEFAULT_SEPARATOR "x"
+
+static int transform (KeySet * appendKS, const Key * key, const char * sep)
+{
+	char * value = elektraStrDup (keyString (key));
+	if (!value) return 0;
+	char * dptr = strstr (value, sep);
+	if (!dptr) return 0;
```
markus2330: Isn't this a validation error? I would expect an error if a transformation is requested but no separator is available.

src/plugins/size/size.c
```diff
@@ -0,0 +1,140 @@
...
+
+int elektraSizeSet (Plugin * handle, KeySet * returned, Key * parentKey ELEKTRA_UNUSED)
+{
+	Key * cur = NULL;
+	KeySet * iterKS = elektraPluginGetData (handle);
+	if (!iterKS) return 0;
+	ksRewind (iterKS);
+	while ((cur = ksNext (iterKS)) != NULL)
+	{
+		keyDel (ksLookup (returned, cur, KDB_O_POP));
```
markus2330: I do not understand this, please add source-code comment here.

How does this code transform the width/height keys back to the original value widthxheight?
Or document the limitation if it is one.

src/plugins/size/README.md
```diff
@@ -0,0 +1,41 @@
...
+- infos/provides = conv
+- infos/recommends =
+- infos/placements = postgetstorage presetstorage
+- infos/status = maintained conformant compatible coverage specific unittest shelltest tested nodep libc configurable experimental
+- infos/metadata = transform/size
+- infos/description = transforms WIDTHxHEIGHT values into separate keys
+
+## Introduction
+
+The `size` plugin transforms values with the format `WIDTH<separator>HEIGHT` into a separate `Width` and `Height` key. The metakey `transform/size` triggers the conversion.
```
tom-wa: >Does it work togehter with the rename plugin?

i don't see any reason why shouldn't, i'll add some shelltests for it

src/plugins/size/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Configuration
+
+By default `x` is assumed as separator. If the plugin configuration `/separator` is set, its value is used instead. If set, `transform/size`'s value takes precedence.
```
tom-wa: the `transform/size` value (per key) overrides `/separator` (plugin configuration), neither is needed and it defaults to `x` (hardcoded).
LCDproc only uses `x`, but it could be useful for similar things like aspect ratios which are usually separated by a `:`
but we can drop it if there's no need for it

src/plugins/size/README.md
markus2330: Thanks! If rename is done earlier than size, different capitalizations of width and height would not work. Its only relevant if rename and this plugin use the same placements. If they do, a simple `infos/ordering = rename` in the contract of this plugin here should fix the potential problem.

tom-wa:
ok, it doesn't really transform the key, but just adds `width` and `height` and drops them `presetstorage`. i don't think theres any reason to remove the original key, but i forgot to change the value of the original key if `height` or `width` gets changed, will add it now

src/plugins/size/size.c
tom-wa: i'm not sure if it should fail. using your example from #1395 
```
[CFontz/Size]
transform/size

[CFontz/Size/Width]
default = 20
range = ...
type = long

[CFontz/Size/Height]
default = 4
range = ...
type = long
```
the value of `Size` would be assumed as invalid/broken/whatever and the default values used instead. imho a warning would be a better

where should we draw the line between a validation and a transformation plugin though ?

src/plugins/size/README.md
markus2330: Okay, if you already found a use case its okay. Maybe give aspect ratios as example for how to change the separator.

src/plugins/size/size.c
markus2330: Yes, it would only be a validation error if on writing out size or height is missing (and no default exists). When parsing, emitting a warning would be needed though, we should not silently ignore metadata.

markus2330:
Ok, if it does not transform we should not name it like that. So it basically only duplicates/copies so that the API finds what it needs, but the keys are actually read-only.

What about calling it `copy/size` then? And we should tag the resulting keys with `readonly` and yield an error if the user tried to change them.

Maybe is doing a proper transformation less work and more user friendly? (Where the original key is hidden and the transformed keys can be changed.)

tom-wa:
already changed it so that the values will be written back if either `height` or `width` are changed. i'll hide the original key too and update the PR

markus2330:

Issue #1579, opened 2017-08-18
Warnings about Ignored Return Values


sanssecours:
# Steps to Reproduce

1. Clone the repository
2. Build Elektra in Ubuntu 14.04 using the steps provided in [`.travis.yml`](https://github.com/sanssecours/libelektra/blob/🏗/.travis.yml)

# Expected Result

Elektra builds without any warnings

# Actual Result

The [Travis build](https://travis-ci.org/sanssecours/libelektra/jobs/265902181#L835-L1166) reports the following warnings:

```
src/plugins/crypto/gpg.c: In function ‘libelektra_ELEKTRA_PLUGIN_NAME_C_LTX_elektraPlugingpgCall’:
src/plugins/crypto/gpg.c:787:8: warning: ignoring return value of ‘dup’, declared with attribute warn_unused_result [-Wunused-result]
    dup (pipe_stdin[0]);
        ^
src/plugins/crypto/gpg.c:793:7: warning: ignoring return value of ‘dup’, declared with attribute warn_unused_result [-Wunused-result]
   dup (pipe_stdout[1]);
       ^
src/plugins/crypto/gpg.c:798:7: warning: ignoring return value of ‘dup’, declared with attribute warn_unused_result [-Wunused-result]
   dup (pipe_stderr[1]);
       ^
src/plugins/crypto/gpg.c:818:9: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
   write (pipe_stdin[1], keyValue (msgKey), keyGetValueSize (msgKey));
         ^
[27/777] Building C object src/plugins...ktra-crypto_gcrypt-objects.dir/gpg.c.o
src/plugins/crypto/gpg.c: In function ‘libelektra_crypto_gcrypt_LTX_elektraPlugingpgCall’:
src/plugins/crypto/gpg.c:787:8: warning: ignoring return value of ‘dup’, declared with attribute warn_unused_result [-Wunused-result]
    dup (pipe_stdin[0]);
        ^
src/plugins/crypto/gpg.c:793:7: warning: ignoring return value of ‘dup’, declared with attribute warn_unused_result [-Wunused-result]
   dup (pipe_stdout[1]);
       ^
src/plugins/crypto/gpg.c:798:7: warning: ignoring return value of ‘dup’, declared with attribute warn_unused_result [-Wunused-result]
   dup (pipe_stderr[1]);
       ^
[...]

Issue #1505, opened 2017-05-26
libelektra.org compromised


markus2330:
libelektra.org (all domains including build server+webpage) is down because the server is compromised.

Hopefully we can use @BernhardDenner setup soon, or can someone provide a temporary setup to host the webpage done by @Namoshek?

tom-wa:
what exactly happened and what does that mean for the debian repository and releases hosted on the server ?

markus2330:
It is yet unclear how exactly the breach happened. We only know that the container hosting libelektra.org suddenly started massive network scans.

Everything hosted on the server currently is not available. Backups exist of course. For releases the backups are also available online at https://github.com/ElektraInitiative/ftp/tree/master/releases

http://www.libelektra.org temporarily redirects to github (https unfortunately is not available). I updated the main README.md to give a notice about the current problem we have.

Sorry for all inconveniences. Please ask here if there are any further questions.

tom-wa:
is there any chance that the packages hosted in the debian repository or release tarballs got compromised ?

markus2330:
The tar balls on the server: certainly. But they are currently not provided.

Did you recently downloaded something and noticed something strange?

petermax2:
> Hopefully we can use @BernhardDenner setup soon, or can someone provide a temporary setup to host the webpage done by @Namoshek?

I could provide a droplet on DigitalOcean as temporary solution. Just send me an SSH key per e-mail and I'll send you the login IP.

markus2330:
Thank you for the great offer! @BernhardDenner can choose, but maybe to directly deploy the website to its final place is less work overall with little difference in downtime.

tr37ion:
You might use redirect to Github pages in the meantime, just to inform people. https://pages.github.com/

BernhardDenner:
Sorry for the very long delay...

The Site is now back again. Hopefully, if found all subdomains and restored them ;)
Jenkins is also online again however, didn't check if it is still building stuff.
The REST backend/frontend are still offline, I'll check that in the next days.

I'm sure, there are lot of things I've overseen, so if you find broken links that worked before, please let me know.

petermax2:
@BernhardDenner Thank you for fixing the server!

The "jenkins build please" commands do not work yet. Probably a new auth token has to be created or maybe it's a Github setting.

markus2330:
Thanks for all your provided help! Looks much better now with the website up again!

BernhardDenner:
[...]

PR #1580, opened 2017-08-19
Travis: Add Basic Linux Build

sanssecours:
# Purpose

This update adds a basic Linux build for Travis. While we already [test multiple variations of Elektra on Linux](https://build.libelektra.org/jenkins), adding an additional build still has some merits:

1. Travis also [tests updates on forks](https://travis-ci.org/sanssecours/libelektra) of this repo
2. Since Travis uses Ubuntu instead of Debian there might be slight differences in the OS, which allow us to [discover new problems](https://github.com/ElektraInitiative/libelektra/issues/1579)
3. Travis still works if someone [compromises the build server](https://github.com/ElektraInitiative/libelektra/issues/1505)

.

# Checklist

- [x] I ran all tests and [everything went fine](https://travis-ci.org/sanssecours/libelektra/builds/266210248).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,8 @@ before_install:
 #
 before_script:
   - cd $TRAVIS_BUILD_DIR/..
+  - INSTALL_DIR="$PWD/install"
+  - SYSTEM_DIR="$PWD/kdbsystem"
   - mkdir build && cd build
   - CMAKE_OPT=()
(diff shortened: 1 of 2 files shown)


markus2330:
Thank you, great to see more build server functionality!

Another advantage is that we test building with ninja more (currently not implemented on our Jenkins, see #160).

PR #1583, opened 2017-08-19
crypto: add payload version to the contract

### LINKED ISSUE #1582: crypto: put payload version information in the contract
I relocated @markus2330  's message from #1567 to here:

A minor implementation note: It would be great if a user can introspect which values a plugin uses by default by returning relevant information via the contract. (Similar to the resolver, e.g. `system/elektra/modules/crypto` ELEKTRA_PLUGIN_NAME `/constants/TMPDIR`)

You should also consider returning the payload version, so something like (`system/elektra/modules/crypto` ELEKTRA_PLUGIN_NAME `/constants/PAYLOAD_VERSION`). Then users can easily introspect which payload version their plugin is able to handle.

Obviously you can also return any other information relevant to the user, such as the used crypto algorithm, password length and so on...

petermax2:
Closes #1582 .

# Purpose

(Long descriptions should be in documentation, not here)

# Checklist

needs to be checked.

- [ ] I added unit tests


petermax2:
@markus2330 I added the `crypto` paylaod version to the contract.

markus2330:
Yes, this is exactly what I am talking about. Is there some reason the other macro constants are not added?

Seems like `scripts/reformat-source` was not applied, can you do it please?

petermax2:
I fixed the remaining problems and added some more constants. I am not sure about what is important for the user tough.

markus2330:
Thank you! Should now contain everything someone wants to know ;)

PR #1589, opened 2017-08-26
Fix compiler warnings (#1579 )

petermax2:
# Purpose

Fix compiler warnings from #1579 in the `crypto` and `fcrypt` plugin.

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/src/plugins/crypto/botan_operations.cpp
+++ b/src/plugins/crypto/botan_operations.cpp
@@ -275,7 +275,11 @@ int elektraCryptoBotanDecrypt (KeySet * pluginConfig, Key * k, Key * errorKey, K
 		if (decryptor.remaining () > 0)
 		{
 			// decode the "header" flags
-			decryptor.read (static_cast<byte *> (&flags), sizeof (kdb_octet_t));
+			const size_t flagLength = decryptor.read (static_cast<byte *> (&flags), sizeof (kdb_octet_t));
+			if (flagLength != sizeof (kdb_octet_t))
+			{
+				throw std::length_error ("Failed to restore the original data type of the value.");
+			}
 		}
 
 		const size_t msgLength = decryptor.remaining ();
--- a/src/plugins/crypto/gpg.c
+++ b/src/plugins/crypto/gpg.c
@@ -787,7 +787,7 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
 			if (dup (pipe_stdin[0]) < 0)
 			{
 				ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "failed to redirect stdin.");
-				return -2;
+				exit (42);
 			}
 		}
 		close (pipe_stdin[0]);
@@ -797,7 +797,7 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
 		if (dup (pipe_stdout[1]) < 0)
 		{
 			ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "failed to redirect the stdout.");
-			return -2;
+			exit (42);
 		}
 		close (pipe_stdout[1]);
 
@@ -806,7 +806,7 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
 		if (dup (pipe_stderr[1]) < 0)
 		{
 			ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "failed to redirect stderr.");
-			return -2;
+			exit (42);
 		}
 		close (pipe_stderr[1]);
 
@@ -814,7 +814,7 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
 		if (execv (argv[0], argv) < 0)
 		{
 			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "failed to start the gpg binary: %s", argv[0]);
-			return -2;
+			exit (42);
 		}
 		// end of the child process
 	}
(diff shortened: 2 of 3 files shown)


src/plugins/crypto/botan_operations.cpp
```diff
@@ -275,7 +275,11 @@ int elektraCryptoBotanDecrypt (KeySet * pluginConfig, Key * k, Key * errorKey, K
 		if (decryptor.remaining () > 0)
 		{
 			// decode the "header" flags
-			decryptor.read (static_cast<byte *> (&flags), sizeof (kdb_octet_t));
+			const size_t flagLength = decryptor.read (static_cast<byte *> (&flags), sizeof (kdb_octet_t));
+			if (flagLength != sizeof (kdb_octet_t))
+			{
+				throw std::length_error ("Failed to restore the original data type of the value.");
```
markus2330: who is catching the exception?

src/plugins/crypto/gpg.c
```diff
@@ -784,25 +784,37 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
 		if (msgKey)
 		{
 			close (STDIN_FILENO);
-			dup (pipe_stdin[0]);
+			if (dup (pipe_stdin[0]) < 0)
+			{
+				ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "failed to redirect stdin.");
+				return -2;
```
markus2330: We have already forked here, this might not do what you expect. (It does not set an error in our process, same problem also 3x below.) See also #1590

petermax2:
@markus2330 I fixed the warnings from #1579 . It should be ready to merge.

src/plugins/fcrypt/fcrypt.c
```diff
@@ -113,15 +114,27 @@ static int shredTemporaryFile (int fd, Key * errorKey)
...
-		return -1;
+		goto error;
 	}
 
 	for (off_t i = 0; i < tmpStat.st_size; i += sizeof (buffer))
 	{
-		write (fd, buffer, sizeof (buffer));
+		if (write (fd, buffer, sizeof (buffer)) != sizeof (buffer))
+		{
+			// save the error state but keep on writing in the hope that further writes wont't fail
```
markus2330: Wouldn't that write a corrupted file?

src/plugins/crypto/gpg.c
petermax2: would it be sufficient to call `exit(-2)` like we do in `external.cpp:196`?

src/plugins/crypto/botan_operations.cpp
petermax2: it is caught directly in line 305

petermax2:
I fixed the error handling within the forked child process in `gpg.c` and the `write()`calls to possibly corrupted files in `fcrypt.c`.

src/plugins/crypto/botan_operations.cpp
markus2330: Ohh sorry, was cut off by github. Isn't this a bit of abuse for exceptions as control flow? If it is a case that never happens it would be okay, though.

src/plugins/crypto/gpg.c
```diff
@@ -784,25 +784,37 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
 		if (msgKey)
 		{
 			close (STDIN_FILENO);
-			dup (pipe_stdin[0]);
+			if (dup (pipe_stdin[0]) < 0)
+			{
+				ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "failed to redirect stdin.");
+				exit (42);
```
markus2330: Is 42 not used by gpg?

src/plugins/crypto/gpg.c
```diff
@@ -784,25 +784,37 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
...
+			exit (42);
+		}
 		close (pipe_stderr[1]);
 
 		// finally call the gpg executable
 		if (execv (argv[0], argv) < 0)
 		{
 			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "failed to start the gpg binary: %s", argv[0]);
-			return -1;
+			exit (42);
```
markus2330: If 42 is used everywhere, you cannot distinguish in the parent process which error happened.

src/plugins/crypto/gpg.c
```diff
@@ -841,6 +863,10 @@ int CRYPTO_PLUGIN_FUNCTION (gpgCall) (KeySet * conf, Key * errorKey, Key * msgKe
 		ELEKTRA_SET_ERROR (ELEKTRA_ERROR_CRYPTO_GPG, errorKey, "GPG reported a bad signature");
 		break;
 
+	case 42:
+		// error has been set to errorKey by the child process
```
markus2330: Yes, but the child process has already exited, who will set the error here (in the parent process)? You need to communicate which error to set.

petermax2: But isn't `errorKey` a pointer pointing to the exact same address in both the child and the parent process?

src/plugins/crypto/gpg.c
petermax2: Sadly GPG does not document it's return values/error codes very well. See [man-page](https://www.gnupg.org/documentation/manpage.html)

src/plugins/crypto/gpg.c
petermax2: Oop sorry, I see where I went wrong.... I'll fix it right away.

petermax2:
Sorry for the confusion. I was thinking in terms of threads not processes. Of course the last error communication did not work.

BTW GPG does not document much about its error codes.

markus2330:
Thank you for the fix & cleanup!

> BTW GPG does not document much about its error codes.

Yes, thus I thought it would be preferable to use the pipe to communicate the error (at least for execv failures). Then we could also pass the errno. But its a minor thing. I hope these large error codes are supported everywhere.

PR #1603, opened 2017-09-12
Zsh completion

reox:
# Purpose

use kdb complete instead of kdb ls for completion


reox:
on installation you need to strip the first three lines from the file. otherwise it will not work as a completion file! completion files need to start with `#compdef kdb`. I hope you do that somewhere :)

markus2330:
Works like a charm, thank you so much! (Even completes on mountpoints, thanks to @e1528532)

No, the 3 lines were added by accident (to suppress warnings by a bash-ism detector). I will remove them.

PR #1604, opened 2017-09-13
Fix Compiler Warnings & Add ASAN Build Jobs

sanssecours:
# Purpose

This pull request fixes various compiler warnings. It also adds two ASAN build jobs for Travis. Some of the changes might not be ideal:

- Instead of disabling Swig warning 317 it would have been better to just address the warning. However, since I do not know anything about Swig, I leave that to the actual authors of the bindings. It might also make sense to disable the warning at a central location. Specifying the code 

   ```c
   #pragma SWIG nowarn=317
  ```

  inside `common.i` did not work, when I tried it though.

- I am quite puzzled why checking for `sys/time.h` on macOS does not work anymore if we add the switch `-Werror`. If someone has a better solution, than to check for header files before setting the compilation switches, then please comment below.

- Maybe we sometimes forgot to add a final `break` in some of the fall-through code inside `switch` statements. In that case it might make sense to replace some of the `// FALLTHROUGH` comments by a `break;` statement in commit 17f3320.

# Checklist

- [x] The reference in commit 94d0bca9 should be correct. I also checked all other commit messages.
- [x] I added code comments for the controversial changes mentioned above. 


markus2330:
Thank you for fixing the warnings!

>  leave that to the actual authors of the bindings

Afaik its not a matter of the bindings but of SWIG itself.

> I am quite puzzled why checking for sys/time.h on macOS 

This seems to be a problem of macOS? A better solution would be to filter -Werror from the flags only for MacOSX at that place where the header file checks have been earlier (maybe CMAKE_REQUIRED_FLAGS helps, too?).

> Maybe we sometimes forgot to add a final break in some of the fall-through code inside switch statements.

Yes, in some cases it looks so. Can you at least mark them with TODO? Or even leave the warnings? Or separate this in another PR, so that we can merge the safe parts already?

sanssecours:
> This seems to be a problem of macOS? A better solution would be to filter -Werror from the flags only for MacOSX at that place where the header file checks have been earlier (maybe CMAKE_REQUIRED_FLAGS helps, too?).

I tried to use `CMAKE_REQUIRED_FLAGS` and `CMAKE_REQUIRED_DEFINITIONS` but unfortunately was not able to make it work that way. The newest code just removes the switch `-Werror` before it checks `sys/time.h` and then reenables the old switches afterwards.

> Yes, in some cases it looks so. Can you at least mark them with TODO? Or even leave the warnings? Or separate this in another PR, so that we can merge the safe parts already?

I inserted `break` statements wherever adding them did not break the tests. Hope this works for you.

markus2330:
Thank you! Can you please create a bug report if you have been unsure in a fallthrough (especially if there was no test case for it).

For future: Please avoid silent exits in shell scripts and for workarounds of a specific OS in CMake, please use `if (OS)`, however innocent the statement looks. Sometimes it breaks another OS nobody thought of.

But fixing it at all is way more important than such nuances, thus the accepted merge.

sanssecours:
> Please avoid silent exits in shell scripts…

If you refer to 

```sh
cd "@CMAKE_SOURCE_DIR@" || exit
```

, then I think exiting without a message in this case should be okay, since `cd` already prints an error message. For example, if I enter `cd bla || exit`, then bash prints the following lines

```
bash: cd: bla: No such file or directory
exit
```

. I did not think of that when I added the code though 😊.

PR #1608, opened 2017-09-21
Spelling Fixes &  DocSet Generation

sanssecours:
# Purpose

This pull request adds DocSet generation to Elektra. DocSets are basically folders containing HTML
documentation together with an index for fast searching. The DocSet format is the standard file format of popular documentation utilities such as [Dash][] and [Zeal][].

[Dash]: https://kapeli.com/dash
[Zeal]: https://zealdocs.org

The screenshot below shows Elektra’s DocSet inside Dash:

dash

. There are basically two minor issues with the DocSet generation. 

1. If we generate the DocSet, then the HTML documentation will **not** contain an index. The index is very helpful if we look at the HTML documentation directly, but is quite useless inside the DocSet:

   dash - index

   . I decided against creating additional HTML documentation just for the DocSet, since this would probably mean a lot more work and extended build time for a very minor enhancement.

2. Changing the documentation (e.g. array.c) and then starting the build process does not update the DocSet. However since this process does also not update the corresponding HTML file `array_8c.html` I think the current behavior is fine.

# Checklist

- [x] I checked all commit messages. None of the commits fixes an open issue.


markus2330:
Thank you, great PR and the great description what DocSet does. Unfortunately, the docu in PR descriptions are not easy to find for others (afaik we are the only one subscribed to all notifications), let us discuss in the next meeting how we better communicate that such enhancements were done (and how to use them).

Issue #1617, opened 2017-10-07
KDB Command to Get Binary Values


sanssecours:
# Proposal

A tool to retrieve binary data from the KDB could be quite handy. In the first step it would make sense that the tool writes out binary data encoded as 8 bit hexadecimal values. Let us assume we call the command `kdb get-binary`, and the binary data at the location `/test/binary` stores the ASCII string `"Hi"` (the decimal numbers 72, 105, and 0). The example below shows an invocation of the proposed command using Markdown Shell Recorder syntax. 

```sh
kdb get-binary /test/binary
#> \x48\x69\x00
```

We could also integrate the functionality to read binary data directly into `kdb get`:

```sh
kdb get -b /test/binary
#> \x48\x69\x00
```

, or even better `kdb get` could just automatically encode binary values:

```sh
kdb get /test/binary
#> \x48\x69\x00
```

. Either way the current situation is not ideal:

```
The command kdb get terminated unsuccessfully with the info:
Binary/String key mismatch, use proper getString()/getBinary() or use getValue() to get both.
Please report the issue at https://issues.libelektra.org/
```

. It might make sense to implement a `kdb` command to add binary data to the database first (see issue #1614).

markus2330:
Seems like you already implemented it but nevertheless here an idea for future reference: We could use plugins as filter to be used for "kdb get". A filter converting binary to a hexadecimal string (enabled by default so that nobody gets the "Binary/String key mismatch" error) would then exactly do what we want.

But your implementation also has advantages (no problems with missing plugins).

Issue #1466, opened 2017-04-11
JNI: Plugin Test Fails


sanssecours:
# Steps to Reproduce the Problem

1. Set the environment variable `JAVA_HOME`

    ```sh
    export JAVA_HOME=$(/usr/libexec/java_home | tr -d '\n')
    ```

1. Clone the repository

    ```sh
    git clone https://github.com/ElektraInitiative/libelektra.git
    ```

2. Create build directory

    ```sh
    mkdir -p build
    ```

3. Create the build system using the following command

    ```sh
    cd build
    cmake ..                                       \
      -GNinja                                      \
      -DENABLE_ASAN=OFF                            \
      -DENABLE_LOGGER=OFF                          \
      -DENABLE_DEBUG=ON                            \
      -DENABLE_TESTING=ON                          \
      -DCMAKE_EXPORT_COMPILE_COMMANDS=ON           \
      -DBUILD_PDF=OFF                              \
      -DPDFLATEX_COMPILER=`which latexmk`          \
      -DPDFLATEX_COMPILER_OPTIONS='-pdf;-f;-quiet' \
      -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5       \
      -DTOOLS=ALL                                  \
      -DBINDINGS="ALL;-swig_python2;-swig_ruby"    \
      -DPLUGINS="ALL"
    ```

4. Compile Elektra

    ```sh
    ninja
    ```

5. Run `testmod_jni`

    ```sh
    ctest -V -R testmod_jni
    ```

# Expected Result

Everything works fine and the test finishes successfully.

# Actual Result

The test fails displaying the output below.

```
test 61
    Start 61: testmod_jni

61: Test command: Elektra/build/bin/testmod_jni "Elektra/src/plugins/jni"
61: Test timeout computed to be: 1500
61: JNI          TESTS
61: ==================
61:
61: There are 1 warnings
61: buffer is: warnings/#00
61: number: 11
61: description: open of plugin returned unsuccessfully (reason contains plugin, see other warnings for details)
61: ingroup: kdb
61: module:
61: file: ../src/libs/elektra/plugin.c
61: line: 297
61: reason: jni
61: reason:
61: reason:
61: ../src/plugins/jni/testmod_jni.c:31: error in test_helloWorld: warnings in kdbOpen for plugin jni
61: number: 102
61: description: : jni/java error
61: ingroup: : plugin
61: module: : jni
[...]

Issue #1042, opened 2016-10-22
ELEKTRA_PLUGIN_NAME


markus2330:
Build system should provide:
- a string that contains the plugin/module name
- a non-string variant of the plugin/module name

Then ELEKTRA_PLUGIN_EXPORT, ELEKTRA_PLUGIN_FUNCTION, ELEKTRA_README would have one paramter less.

markus2330:
@petermax2 Do you maybe want to implement this?

petermax2:
Yes, you can assign me if the implementation is not time critical. At the moment I have not so much spare time.

markus2330:
Thank you for working on this!

The actual reason to do this, is that every plugin has something like CRYPTO_PLUGIN_FUNCTION available and that CRYPTO_PLUGIN_FUNCTION is not needed anymore.

Btw. ARG_COMPILE_DEFINITIONS cannot be modified (due to restore_variable), so simply add further definitions in cmake/Modules/LibAddPlugin.cmake line 155 (where COMPILE_DEFINITIONS are defined).

markus2330:
Great that you have progress!

petermax2:
@markus2330  the build server reported

    Caught the following exception: null

Is this a know bug?

petermax2:
### Progress so far

#2360 : The CMake module `add_plugin` provides the following compiler definitions for every plugin:

- **ELEKTRA_PLUGIN_NAME**: the plugin name (including the variant if any) as string
- **ELEKTRA_PLUGIN_NAME_C**: the plugin name without quotes, so it can be used to construct function names, that depend on the plugin name/variant.

The removal of unnecessary parameters of `ELEKTRA_PLUGIN_FUNCTION` and `CRYPTO_PLUGIN_FUNCTION` will be implemented in another PR.

markus2330:
So ELEKTRA_PLUGIN_EXPORT and ELEKTRA_README are the last left-overs?

petermax2:
> So ELEKTRA_PLUGIN_EXPORT and ELEKTRA_README are the last left-overs?

Yes, should be very easy and straight-forward now that `ELEKTRA_PLUGIN_FUNCTION` works.

markus2330:
We finally did it!

PR #1618, opened 2017-10-10
Binary Data Support for YAML CPP

sanssecours:
# Purpose

This pull request adds support for binary data to the YAML CPP plugin via a modified version of the Base64 plugin called Base 666. The pull request also adds support for binary data to `kdb get` (issue #1617) and fixes the `CLASSPATH` detection of the JNI plugin test (issue #1466).

# Checklist

- [x] I checked all commit messages. The reference in commit a6ff6f28 should be correct.
- [x] I added a Markdown Shell Recorder test for the new Base 666 plugin.
- [x] I added code comments and logging code.


sanssecours:
The problem with the current version of the Base 666 (and Base 64) plugin seems to be the expansion of

```
ELEKTRA_PLUGIN_FUNCTION (ELEKTRA_PLUGIN_NAME_C, base64Decode)
```

. Since `ELEKTRA_PLUGIN_NAME_C` is a macro definition and `ELEKTRA_PLUGIN_FUNCTION` is a macro function `ELEKTRA_PLUGIN_NAME_C` is not replaced in the code and the [build server](https://travis-ci.org/ElektraInitiative/libelektra/jobs/285966384#L1576) produces the following output:

```
src/plugins/base666/CMakeFiles/elektra-base666-objects.dir/__/base64/base64_functions.c.o: In function `libelektra_ELEKTRA_PLUGIN_NAME_C_LTX_elektraPluginbase64Encode':
/home/travis/build/ElektraInitiative/libelektra/src/plugins/base64/base64_functions.c:22: multiple definition of `libelektra_ELEKTRA_PLUGIN_NAME_C_LTX_elektraPluginbase64Encode'
src/plugins/base64/CMakeFiles/elektra-base64-objects.dir/base64_functions.c.o:/home/travis/build/ElektraInitiative/libelektra/src/plugins/base64/base64_functions.c:22: first defined here
src/plugins/base666/CMakeFiles/elektra-base666-objects.dir/__/base64/base64_functions.c.o: In function `libelektra_ELEKTRA_PLUGIN_NAME_C_LTX_elektraPluginbase64Decode':
/home/travis/build/ElektraInitiative/libelektra/src/plugins/base64/base64_functions.c:103: multiple definition of `libelektra_ELEKTRA_PLUGIN_NAME_C_LTX_elektraPluginbase64Decode'
src/plugins/base64/CMakeFiles/elektra-base64-objects.dir/base64_functions.c.o:/home/travis/build/ElektraInitiative/libelektra/src/plugins/base64/base64_functions.c:103: first defined here
collect2: error: ld returned 1 exit status
```

. We see that `ELEKTRA_PLUGIN_NAME_C` is not replaced with the name of the plugin (`base64`). The following example (`test.c`) should make the problem more obvious:

```c
#define ELEKTRA_PLUGIN_FUNCTION(module, function)                              \
  libelektra_##module##_LTX_elektraPlugin##function

#define ELEKTRA_PLUGIN_NAME_C base64

ELEKTRA_PLUGIN_FUNCTION(ELEKTRA_PLUGIN_NAME_C, funcy)
```

. If we run the preprocessor (`gcc -E test.c` or `clang -E test.c`), then it produces

```
libelektra_ELEKTRA_PLUGIN_NAME_C_LTX_elektraPluginfuncy
``` 

instead of the text we want:

```
libelektra_base64_LTX_elektraPluginfuncy
```

. I am not really sure if we can fix the problem above with the preprocessor alone. If someone has a simple solution for the issue (apart from using CMake’s `configure_file`, code generation or everyone’s favorite tool “copy & paste”), then please just comment below.

markus2330:
Great to see this functionality within an plugin (and not directly within yamlcpp). If possible I would prefer that we only have a single base64 plugin. Note that you can pass run-time configuration to plugins via the contract. See be878fb398ed62c8737e5fcfaf64b67c967dfeba for how to do that.

In your case you would pass something like `keyNew ("system/elektra/modules/yamlcpp/config/needs/base64/all", KEY_VALUE, "1", KEY_END),` to yamlcpp (to tell base64 to encode all binaries.

I would even prefer encoding all binaries to base64 as default and make the need of the crypto plugin as a special case, if we need a special case at all. @petermax2 what do you think is the best way to avoid the duplication of the plugin?

src/plugins/base64/README.md
```diff
@@ -41,4 +41,24 @@ sudo kdb umount /examples/base64/test
...
+```sh
+# Mount the INI and Base64 plugin
+kdb mount config.ini user/examples/base64 ini base64
+
+# Copy binary data
+kdb cp system/elektra/modules/dump/exports/unserialise user/examples/base64/binary
+
+# Print binary data
+kdb get user/examples/base64/binary
+# STDOUT-REGEX: (\\\\x[0-9a-f]+)+
```
markus2330: does this make sure that *only* `[0-9a-f]` are present? (is the regex implicitely surrounded with `^$`?)

.travis.yml
```diff
@@ -20,6 +20,10 @@ matrix:
       compiler: clang
       env: [ ASAN=ON ]
 
+    - os: linux
```
markus2330: Great to have one buildjob more. Is it possible to have a short description of the build jobs? (what means FULL)

src/plugins/base666/README.md
```diff
@@ -0,0 +1,26 @@
...
+- infos/placements = postgetstorage presetstorage
+- infos/status = discouraged experimental
+- infos/metadata =
+- infos/description = Base64 Encoding
+
+# Base 666
+
+> If You're 555 then I’m 666. What’s it like to be a heretic?
+
+Slipknot — The Heretic Anthem
```
markus2330: Spaces and — in the beginning to make clear that this is a citation

src/plugins/yamlcpp/README.md
```diff
@@ -3,7 +3,7 @@
 - infos/licence = BSD
 - infos/needs =
 - infos/provides = storage/yaml
-- infos/recommends =
+- infos/recommends = base666
```
markus2330: or require? What happens with binary data without base666? I think the user does not expect to receive base64 encoded strings? (Does the standard says something about it?)

markus2330:
> The problem with the current version of the Base 666 (and Base 64) plugin seems to be the expansion of

Isn't this already proposed in #1042? If not, can you please create a separate issue?

sanssecours:
> I would prefer a solution with only one base64 plugin, though.

I am currently working on it.

src/plugins/base64/README.md
sanssecours: Nope. I replaced the regex with the more rigorous `^(\\\\x[0-9a-f]{1,2})+$`.

.travis.yml
sanssecours: `FULL=ON` enables a “full build“. It is basically a shorthand for `BUILD_FULL=ON`.

src/plugins/yamlcpp/README.md
sanssecours: > What happens with binary data without base666?

Then the user receives a base64 encoded string.

> I think the user does not expect to receive base64 encoded strings?

That depends, if he read the documentation of the plugin, then he would probably know that YAML CPP does not support base64 encoded binary data fully without the Base 666 plugin.

> Does the standard says something about it?

As far as I can remember base64 support is not required. Otherwise [yaml-cpp](https://github.com/jbeder/yaml-cpp) would probably save the data in encoded form.

sanssecours:
> Isn't a solution to this problem already proposed in #1042?

Yeah, I guess the proposed solution would work. Anyway, I created a new issue [here](https://github.com/ElektraInitiative/libelektra/issues/1620).

markus2330:
Thank you, looks nice! Doesn't this PR also fix #1019 (by adding base64 as dep for tcl)?

PR #1623, opened 2017-10-16
haskell-bindings-improve-cmake

### LINKED ISSUE #1616: Build Fails If Cabal is Installed
# Steps to Reproduce

1. Install Cabal via Homebrew

   ```sh
   brew install cabal-install
   ```

2. Create the build system

   ```sh
   mkdir -p build
   cd build; cmake ..                                 \
         -GNinja                                      \
         -DENABLE_ASAN=OFF                            \
         -DENABLE_LOGGER=ON                           \
         -DENABLE_DEBUG=ON                            \
         -DCMAKE_EXPORT_COMPILE_COMMANDS=ON           \
         -DBUILD_DOCUMENTATION=OFF                    \
         -DBUILD_DOCSET=OFF                           \
         -DBUILD_PDF=OFF                              \
         -DPDFLATEX_COMPILER=`which latexmk`          \
         -DPDFLATEX_COMPILER_OPTIONS='-pdf;-f;-quiet' \
         -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5       \
         -DTOOLS="ALL"                                \
         -DBINDINGS="ALL;-swig_python2;-swig_ruby"    \
         -DPLUGINS="ALL;-jni;-python2;-lua"
   ```

# Expected Result

CMake configures the project without any problems.

# Actual Result

CMake prints the following error message:

```
CMake Error at cmake/Modules/LibAddPlugin.cmake:393 (add_dependencies):
  The dependency target "c2hs_haskell" of target "elektra-haskell" does not
  exist.
  cmake/Modules/LibAddHaskellPlugin.cmake:141 (add_plugin)
  src/plugins/haskell/CMakeLists.txt:3 (add_haskell_plugin)


CMake Error at cmake/Modules/LibAddPlugin.cmake:353 (add_dependencies):
  The dependency target "c2hs_haskell" of target "elektra-haskell-objects"
  does not exist.
  cmake/Modules/LibAddHaskellPlugin.cmake:141 (add_plugin)
  src/plugins/haskell/CMakeLists.txt:3 (add_haskell_plugin)
```

.

# System Information

- Elektra Version: master
- OS: macOS 10.13

e1528532:
# Purpose

Fixes #1616 build error, also makes the GHC library variables and executables advanced

# Checklist

@markus2330 Should be fine to merge i think, the issue was reproducible and seems to work now.

### DIFF
--- a/cmake/Modules/LibAddHaskellPlugin.cmake
+++ b/cmake/Modules/LibAddHaskellPlugin.cmake
@@ -150,7 +150,7 @@ macro (add_haskell_plugin target)
 			endif (GHC_FFI_LIB)
 
 		else (FINDEX GREATER -1)
-			remove_plugin (${target} "haskell bindings are not included in the make configuration")
+			remove_plugin (${target} "haskell bindings are not included in the cmake configuration")
 		endif (FINDEX GREATER -1)
 		else (GHC-PKG_EXECUTABLE)
 			remove_plugin (${target} "ghc-pkg not found")
--- a/cmake/Modules/LibAddHaskellPlugin.cmake
+++ b/cmake/Modules/LibAddHaskellPlugin.cmake
@@ -130,6 +130,7 @@ macro (add_haskell_plugin target)
 				COMMAND ${CABAL_EXECUTABLE} --ipid=${target} configure
 				COMMAND ${CABAL_EXECUTABLE} build
 				WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+				BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/dist/build/libHS${target}.a
 				DEPENDS ${target}-register
 			)
 
--- a/cmake/Modules/LibAddHaskellPlugin.cmake
+++ b/cmake/Modules/LibAddHaskellPlugin.cmake
@@ -17,8 +17,8 @@ macro (add_haskell_plugin target)
 		${ARGN}
 	)
 
-	SET (PLUGIN_NAME ${target})
-	SET (PLUGIN_NAME_CAPITALIZED ${target})
+	set (PLUGIN_NAME ${target})
+	set (PLUGIN_NAME_CAPITALIZED ${target})
 	string (SUBSTRING ${PLUGIN_NAME} 0 1 FIRST_LETTER)
 	string (TOUPPER ${FIRST_LETTER} FIRST_LETTER)
 	string (REGEX REPLACE "^.(.*)" "${FIRST_LETTER}\\1" PLUGIN_NAME_CAPITALIZED "${PLUGIN_NAME}")
@@ -33,6 +33,8 @@ macro (add_haskell_plugin target)
 		if (CABAL_EXECUTABLE)
 		if (GHC_EXECUTABLE)
 		if (GHC-PKG_EXECUTABLE)
+		list (FIND BINDINGS "haskell" FINDEX)
+		if (FINDEX GREATER -1)
 
 			# needed for HsFFI.h
 			execute_process (
@@ -71,6 +73,12 @@ macro (add_haskell_plugin target)
 				OUTPUT_VARIABLE GHC_PRIM_NAME OUTPUT_STRIP_TRAILING_WHITESPACE
 			)
 			find_library (GHC_PRIM_LIB "HS${GHC_PRIM_NAME}" ${GHC_LIB_DIR}/${GHC_PRIM_NAME})
+
+			if (GHC_FFI_LIB)
+			if (GHC_RTS_LIB)
+			if (GHC_BASE_LIB)
+			if (GHC_GMP_LIB)
+			if (GHC_PRIM_LIB)
 
 			set (GHC_LIB_DIRS
 				"${CMAKE_CURRENT_BINARY_DIR}/dist/build/libHS${target}.a"
@@ -125,6 +133,25 @@ macro (add_haskell_plugin target)
 				DEPENDS ${target}-register
 			)
 
+			else (GHC_PRIM_LIB)
+				remove_plugin (${target} "GHC_PRIM_LIB not found")
+			endif (GHC_PRIM_LIB)
+			else (GHC_GMP_LIB)
+				remove_plugin (${target} "GHC_GMP_LIB not found")
+			endif (GHC_GMP_LIB)
+			else (GHC_BASE_LIB)
+				remove_plugin (${target} "GHC_BASE_LIB not found")
+			endif (GHC_BASE_LIB)
+			else (GHC_RTS_LIB)
+				remove_plugin (${target} "GHC_RTS_LIB not found")
+			endif (GHC_RTS_LIB)
+			else (GHC_FFI_LIB)
+				remove_plugin (${target} "GHC_FFI_LIB not found")
+			endif (GHC_FFI_LIB)
+
+		else (FINDEX GREATER -1)
+			remove_plugin (${target} "haskell bindings are not included in the make configuration")
+		endif (FINDEX GREATER -1)
 		else (GHC-PKG_EXECUTABLE)
 			remove_plugin (${target} "ghc-pkg not found")
 		endif (GHC-PKG_EXECUTABLE)
@@ -149,5 +176,16 @@ macro (add_haskell_plugin target)
 		DEPENDS
 			${target} c2hs_haskell
 		ADD_TEST
+	)
+
+	mark_as_advanced (
+		GHC_EXECUTABLE
+		GHC-PKG_EXECUTABLE
+		CABAL_EXECUTABLE
+		GHC_FFI_LIB
+		GHC_RTS_LIB
+		GHC_BASE_LIB
+		GHC_GMP_LIB
+		GHC_PRIM_LIB
 	)
 endmacro (add_haskell_plugin)
--- a/src/bindings/haskell/CMakeLists.txt
+++ b/src/bindings/haskell/CMakeLists.txt
@@ -19,6 +19,7 @@ if (GHC_EXECUTABLE)
 		c2hs_haskell ALL
 		COMMAND ${CABAL_EXECUTABLE} --ipid=libelektra-haskell-${KDB_VERSION} configure
 		COMMAND ${CABAL_EXECUTABLE} build
+		BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/dist/build/libHSlibelektra-haskell-${KDB_VERSION}.a
 		WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
 	)
 
--- a/src/bindings/haskell/CMakeLists.txt
+++ b/src/bindings/haskell/CMakeLists.txt
@@ -6,8 +6,7 @@ if (CABAL_EXECUTABLE) # set by find_program
 if (C2HS_EXECUTABLE)
 if (GHC_EXECUTABLE)
 	
-	SET (CABAL_INCLUDE_DIRS "${CMAKE_SOURCE_DIR}/src/include, ${CMAKE_BINARY_DIR}/src/include")
-	SET (CABAL_EXTRA_LIB_DIRS "")
+	set (CABAL_INCLUDE_DIRS "${CMAKE_SOURCE_DIR}/src/include, ${CMAKE_BINARY_DIR}/src/include")
 
 	# configure include paths
 	configure_file (
@@ -54,3 +53,9 @@ endif (C2HS_EXECUTABLE)
 else (CABAL_EXECUTABLE)
 	remove_binding (c2hs_haskell "cabal not found")
 endif (CABAL_EXECUTABLE)
+
+mark_as_advanced (
+	GHC_EXECUTABLE
+	C2HS_EXECUTABLE
+	CABAL_EXECUTABLE
+)
--- a/src/bindings/haskell/src/Elektra/Key.chs
+++ b/src/bindings/haskell/src/Elektra/Key.chs
@@ -122,7 +122,7 @@ keyListMeta key = keyRewindMeta key >> listMeta []
         listMeta res = do
             cur <- keyNextMeta key
             isNull <- keyPtrNull cur
-            if isNull then return res else liftM (cur :) (listMeta res)
+            if isNull then return res else keyIncRef cur >> liftM (cur :) (listMeta res)
 
 -- ***
 -- KEY TESTING METHODS


cmake/Modules/LibAddHaskellPlugin.cmake
```diff
@@ -125,6 +133,25 @@ macro (add_haskell_plugin target)
...
+			endif (GHC_BASE_LIB)
+			else (GHC_RTS_LIB)
+				remove_plugin (${target} "GHC_RTS_LIB not found")
+			endif (GHC_RTS_LIB)
+			else (GHC_FFI_LIB)
+				remove_plugin (${target} "GHC_FFI_LIB not found")
+			endif (GHC_FFI_LIB)
+
+		else (FINDEX GREATER -1)
+			remove_plugin (${target} "haskell bindings are not included in the make configuration")
```
sanssecours: Typo: make → CMake

markus2330:
> then the Haskell plugin is excluded with the following message

Is there also an output line regarding an excluded haskell binding?

Did you try to add the haskell binding manually? If it is experimental, it won't be added with `-DBINDINGS="ALL"`.

e1528532:
Yes markus is correct, as they are experimental the haskell bindings are not included with ALL, therefore the plugins which depend on the bindings aren't included either in that case. This will work in the future once we add the haskell bindings to ALL, so its not really an issue but rather works as intended i'd say.

Fixed the typo along with a small reference counting issue in the bindings i encountered during my other developments.

sanssecours:
> Is there also an output line regarding an excluded haskell binding?

Nope.

>  If it is experimental, it won't be added with -DBINDINGS="ALL".

Thank you for the info. The behavior was not something I would have expected, since e.g. `DPLUGINS="ALL"` [works differently](https://github.com/ElektraInitiative/libelektra/blob/master/doc/COMPILE.md#plugins):

> By default nearly all plugins are added. Only experimental plugins
> will be omitted:
> 
>     -DPLUGINS="ALL;-EXPERIMENTAL"
> 
> To add also experimental plugins, you can use:
> 
>     -DPLUGINS=ALL

.

sanssecours:
I configured Elektra with `BINDINGS="ALL;haskell"`, installed `c2hs` and tried to compile the code afterwards. Ninja now prints the following error message:

```
ninja: error: 'src/plugins/haskell/dist/build/libHShaskell.a', needed by 'lib/libelektra-haskell.so', missing and no known rule to make it
```

.

markus2330:
> The behavior was not something I would have expected

Yes, the behavior is wrong (only PLUGINS was fixed so far, unfortunately it is not easy to fix TOOLS and PLUGINS), I updated the docu in 469e578649ec7907c1f79b4cfa579c156cf926f1

e1528532:
> I configured Elektra with BINDINGS="ALL;haskell", installed c2hs and tried to compile the code afterwards. Ninja now prints the following error message:

First of all thank your for trying this out, you often have all sorts of stuff configured on your development system which are not present on a "fresh" system so then while it works for you it won't for others.

Unfortunately i cannot reproduce your last issue, i tried it on a vm which had no haskell installed before and there it compiled successfully after only installing ghc, cabal and c2hs. Do you have any further information how to possibly reproduce?

libHShaskell.a is the output of compiling the haskell part of the haskell plugin, which then gets linked to the shared library libelektra-haskell.so that is the c proxy that initializes ghc and forwards the calls to libHShaskell.a . So unless there was some compilation error on that file, it should be present in theory. Do you see it in the build folder in build/src/plugins/haskell/dist/build? It should contain files like libHShaskell-ghc8.2.1.dylib, libHShaskell_p.a and libHShaskell.a

sanssecours:
> Do you have any further information how to possibly reproduce?

Unfortunately in the moment I can not install `c2hs` on my machine:

```
cabal install c2hs
Resolving dependencies...
Downloading c2hs-0.28.2...
Configuring c2hs-0.28.2...
Building c2hs-0.28.2...
Failed to install c2hs-0.28.2
Build log ( /Users/rene/.cabal/logs/ghc-8.2.1/c2hs-0.28.2-JC7ouNkfRzBIpk1MG6chXz.log ):
cabal: Entering directory '/var/folders/hx/flbncdhj4fs87095gzxvnj3h0000gn/T/cabal-tmp-51019/c2hs-0.28.2'
Configuring c2hs-0.28.2...
Preprocessing executable 'c2hs' for c2hs-0.28.2..
Building executable 'c2hs' for c2hs-0.28.2..

<no location info>: warning: [-Wmissing-home-modules]
    These modules are needed for compilation but not listed in your .cabal file's other-modules: Paths_c2hs
[ 1 of 26] Compiling C2HS.C.Info      ( src/C2HS/C/Info.hs, dist/build/c2hs/c2hs-tmp/C2HS/C/Info.o )
[ 2 of 26] Compiling C2HS.Config      ( src/C2HS/Config.hs, dist/build/c2hs/c2hs-tmp/C2HS/Config.o )
[ 3 of 26] Compiling C2HS.Switches    ( src/C2HS/Switches.hs, dist/build/c2hs/c2hs-tmp/C2HS/Switches.o )
[ 4 of 26] Compiling Control.StateTrans ( src/Control/StateTrans.hs, dist/build/c2hs/c2hs-tmp/Control/StateTrans.o )

src/Control/StateTrans.hs:80:1: warning: [-Wdodgy-imports]
    Module ‘Prelude’ does not export ‘catch’
   |
80 | import Prelude hiding (catch)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ 5 of 26] Compiling Data.Errors      ( src/Data/Errors.hs, dist/build/c2hs/c2hs-tmp/Data/Errors.o )

src/Data/Errors.hs:41:1: error:
    Could not find module ‘Language.C.Data.Error’
    There are files missing in the ‘language-c-0.6.1’ package,
    try running 'ghc-pkg check'.
    Use -v to see a list of the files searched for.
   |
41 | import Language.C.Data.Error hiding (Error)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

src/Data/Errors.hs:42:1: error:
    Could not find module ‘Language.C.Data.Position’
    There are files missing in the ‘language-c-0.6.1’ package,
    try running 'ghc-pkg check'.
    Use -v to see a list of the files searched for.
   |
42 | import Language.C.Data.Position
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cabal: Leaving directory '/var/folders/hx/flbncdhj4fs87095gzxvnj3h0000gn/T/cabal-tmp-51019/c2hs-0.28.2'
cabal: Error: some packages failed to install:
c2hs-0.28.2-JC7ouNkfRzBIpk1MG6chXz failed during the building phase. The
exception was:
ExitFailure 1
```

. I will try to install Haskell on the Travis build server and check if building Elektra works there. Since the macOS instances on Travis are quite slow, this will probably take some time though.

e1528532:
Unfortunately cabal often has its issues when its used in the standard global mode leading to exceptions like that, though it gets better. In my experience errors like this can often be resolved by reinstalling the mentioned package using `cabal install language-c --force-reinstall`. Refreshing the package cache using `ghc-pkg recache`might also help.

I am thinking about switching the build system to stack https://haskellstack.org because of that reason which behaves more like a sandbox and uses its own packages for each project, in fact you could even have your own independent ghc versions for your projects. Ultimately it seems to be more robust to use as it then doesn't depend on the users global haskell environments so much. But this would be quite some effort to do so I'd postpone it for now.

sanssecours:
> In my experience errors like this can often be resolved by reinstalling the mentioned package using `cabal install language-c --force-reinstall`. Refreshing the package cache using `ghc-pkg recache`might also help.

Thank you for the advice. Unfortunately neither command helps on my computer.

> Ultimately it seems to be more robust to use as it then doesn't depend on the users global haskell environments so much.

Since my initial Haskell environment does not include any package from Cabal (`~/.cabal` is empty), I do not think that using another package manager on top of Cabal would help much.

> But this would be quite some effort to do so I'd postpone it for now.

That makes sense. Personally I would advice against using another build system. Adding additional complexity is usually not a good idea, especially if it means using a tool based on another unreliable tool.

e1528532:
>Since my initial Haskell environment does not include any package from Cabal (~/.cabal is empty), I do not think that using another package manager on top of Cabal would help much.

Does that mean you basically had a more or less fresh installation of haskell too?

Maybe it depends on the versions used. You can check the versions of the relevant tools with `ghc --version` and `cabal --version`, mine are 8.2.1 for ghc and 2.0.0.0 of cabal-install using 2.0.0.2 of the cabal library. What are yours?

Have you had older versions of ghc/cabal installed before, that may also lead to issues in my experience. The output of the command `ghc-pkg list` could maybe also give some further insight if you'd provide it.

>Personally I would advice against using another build system.

Stack is pretty widespread in the haskell community but you are right, my initial decision for cabal was also because i wanted to avoid using yet another tool and stick to cabal which is usually bundled with haskell installations.

sanssecours:
> What are yours?

I use exactly the same versions of `cabal` and `ghc`.

> The output of the command `ghc-pkg list` could maybe also give some further insight if you'd provide it.

That advice was very helpful, since it showed me that there was still an `.ghc` directory (from yesterday) in my home folder. I removed the directory and then installed `happy`, `alex` and finally `c2hs`. After I did that, I configured the build system with `BINDINGS="ALL;haskell"` and invoked `ninja`. Ninja still shows the same error message as yesterday. 

In theory the two active (non-canceled) macOS build jobs [here](https://travis-ci.org/sanssecours/libelektra/builds/289142841) should show us if the problem only exists on my computer.

sanssecours:
Travis now finished the two build jobs I mentioned before. They both [[1](https://travis-ci.org/sanssecours/libelektra/jobs/289142846#L5563), [2](https://travis-ci.org/sanssecours/libelektra/jobs/289142847#L2751)] show the same problem I had locally:

```
ninja: error: 'src/plugins/haskell/dist/build/libHShaskell.a', needed by 'lib/libelektra-haskell.so', missing and no known rule to make it
```

. The configuration of the build jobs is available [here](https://github.com/sanssecours/libelektra/blob/haskell/.travis.yml). I think we should either

1. fix the problem in the current pull request, or
2. merge the pull request as is and open a new issue describing the bug

. What do you think?

e1528532:
I noticed this in the log files:

> Configuring libelektra-haskell-0.8.19...
> cabal: Encountered missing dependencies:
> hspec -any

This is required for the test execution of the haskell tests, you can install these files with cabal. QuickCheck is also required by hspec as i included one QuickCheck test. The commands should be first `cabal install QuickCheck` followed by `cabal install hspec`.

I have also figured out the other issue. It only happens with ninja. You have to specify the byproducts of my custom cabal commands which build the haskell libraries if they are dependencies of other targets, otherwise it doesn't figure how those required libraries get built. I have fixed this in my last commit, it compiled fine now locally with ninja. Please try again if it also works on travis now, and thanks for your valuable inputs!

src/bindings/haskell/CMakeLists.txt
```diff
@@ -6,8 +6,7 @@ if (CABAL_EXECUTABLE) # set by find_program
 if (C2HS_EXECUTABLE)
 if (GHC_EXECUTABLE)
 	
-	SET (CABAL_INCLUDE_DIRS "${CMAKE_SOURCE_DIR}/src/include, ${CMAKE_BINARY_DIR}/src/include")
-	SET (CABAL_EXTRA_LIB_DIRS "")
+	set (CABAL_INCLUDE_DIRS "${CMAKE_SOURCE_DIR}/src/include, ${CMAKE_BINARY_DIR}/src/include")
```
sanssecours: I think 

```cmake
set (CABAL_INCLUDE_DIRS "\"${CMAKE_SOURCE_DIR}/src/include\", \"${CMAKE_BINARY_DIR}/src/include\"")
```

would be better, since it makes people that have space characters in their build path happier 😊. As far as I can tell `@CMAKE_CURRENT_SOURCE_DIR@` in `libelektra-haskell.cabal.in` should also be quoted:

```
"@CMAKE_CURRENT_SOURCE_DIR@"/src
```

.

sanssecours:
Looks like building on Travis works. On my machine it works partially after I [quote the library paths](https://github.com/sanssecours/libelektra/commit/cd6edc668856a7d510c509331c533787c2a90b9a). For example, if I invoke

```c
cmake -GNinja -DBINDINGS="ALL;haskell" ..
```

and then call Ninja everything is fine. However if I invoke

```
cmake ..                                       \
  -GNinja                                      \
  -DBUILD_STATIC=OFF                           \
  -DBUILD_FULL=OFF                             \
  -DENABLE_ASAN=OFF                            \
  -DENABLE_LOGGER=ON                           \
  -DENABLE_DEBUG=ON                            \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON           \
  -DBUILD_DOCUMENTATION=ON                     \
  -DBUILD_DOCSET=OFF                           \
  -DBUILD_PDF=OFF                              \
  -DPDFLATEX_COMPILER=(which latexmk)          \
  -DPDFLATEX_COMPILER_OPTIONS='-pdf;-f;-quiet' \
  -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5       \
  -DTOOLS="ALL"                                \
  -DBINDINGS="ALL;haskell"                     \
  -DPLUGINS="ALL;-jni;-python2;-lua"
```

then the build fails with the error:

```
FAILED: src/bindings/haskell/dist/build/libHSlibelektra-haskell-0.8.19.a
cd "/Users/rene/Documents/Studium/Master Thesis/Configuration Parsing/Source/Elektra/build/src/bindings/haskell" && /usr/local/bin/cabal --ipid=libelektra-haskell-0.8.19 configure && /usr/local/bin/cabal build
Resolving dependencies...
Configuring libelektra-haskell-0.8.19...
cabal: Missing dependency on a foreign library:
* Missing (or bad) header file: kdb.h
This problem can usually be solved by installing the system package that
provides this library (you may need the "-dev" version). If the library is
already installed but in a non-standard location then you can use the flags
--extra-include-dirs= and --extra-lib-dirs= to specify where it is.
If the header file does exist, it may contain errors that are caught by the C
compiler at the preprocessing stage. In this case you can re-run configure
with the verbosity flag -v3 to see the error messages.
[99/887] Building CXX object gtest/CMakeFiles/gtest.dir/src/gtest-all.cc.o
ninja: build stopped: subcommand failed.
```

.

sanssecours:
After I reinstalled the command line tools – I broke part of my system [yesterday](https://github.com/ElektraInitiative/libelektra/pull/1627#issuecomment-337527040) – everything seems to work fine now 🙌.

markus2330:
@e1528532 Can you set ready to merge please?

e1528532:
Ok so everything works fine now for @sanssecours and there is no quoting issue left?

In that case it should be fine to merge i think, otherwise i can still try to resolve quoting issues if there are any.

sanssecours:
> Ok so everything works fine now for @sanssecours and there is no quoting issue left?

I already fixed the quoting issues. I will open a pull request containing a fix later.

PR #1624, opened 2017-10-16
Documentation: Integrate Man Pages Into Repository

sanssecours:
# Purpose

This pull request adds the man pages (usually produced by [ronn](https://github.com/rtomayko/ronn)) to the repository. This change allows us to drop ronn from the list of essential build tools.

# Checklist

- [x] I checked all commit messages. The issue reference in commit 23087b0 should be correct.


doc/help/CMakeLists.txt
```diff
@@ -16,5 +16,11 @@ if (RONN_LOC)
 		generate_manpage (${NAME} SECTION 7)
 	endforeach ()
 else ()
-	message (WARNING "Sorry, I cannot find ronn, thus I cannot build the man pages. Please install https://github.com/rtomayko/ronn.")
+	message ("Sorry, I cannot find ronn, thus I will install man pages directly from the repository.")
```
markus2330: I would prefer if we do not remove the link, maybe change it to:
`Sorry, I cannot find ronn (https://github.com/rtomayko/ronn), thus the script update-manpages won't work and I will install....`

src/bindings/swig/ruby/CMakeLists.txt
```diff
@@ -63,8 +63,11 @@ else()
 	set (SWIG_COMPILE_FLAGS "${SWIG_COMPILE_FLAGS} -DSWIG_TYPE_TABLE=kdb")
 
 	# disable certain compiler warnings for SWIG generated files
-	set (SWIG_COMPILE_FLAGS "${SWIG_COMPILE_FLAGS} -Wno-unused-parameter -Wno-unused-but-set-variable")
+	set (SWIG_COMPILE_FLAGS "${SWIG_COMPILE_FLAGS} -Wno-unused-parameter")
 	set (SWIG_COMPILE_FLAGS "${SWIG_COMPILE_FLAGS} -Wno-sign-compare")
+	if (NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
markus2330: Is this Clang or Apple specific?

scripts/update-manpages
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
```
markus2330: Please add some metadata so that it will work with `find-tools`. `scripts/README.md` describes how to do that.

scripts/update-manpages
```diff
@@ -0,0 +1,48 @@
...
+cleanup() {
+        printf '👋  Change back to original directory “%s”\n' "$CURRENT_DIR"
+        cd "$CURRENT_DIR" || exit
+}
+
+# ========
+# = Init =
+# ========
+
+CURRENT_DIR="$PWD"
```
markus2330: Executing `pwd` sounds more reliable to me. (PWD can simply be modified.) Alternatively, we should fix the other scripts.

scripts/update-manpages
```diff
@@ -0,0 +1,48 @@
...
+SCRIPTS_DIR=$(dirname "$0")
+. "$SCRIPTS_DIR/realpath"
```
markus2330: why not directly execute include-common, it already sets some of the needed vars.

scripts/update-manpages
```diff
@@ -0,0 +1,48 @@
...
+command -v cmake >/dev/null 2>&1 || { printf 'Unable to locate CMake' >&2; exit 1; }
+
+mkdir -p "$BUILD_DIR" || { printf 'Unable to create “%s”' "$BUILD_DIR" >&2; exit 1; }
+
+# ========
+# = Main =
+# ========
+
+printf '👋  Change into build directory “%s”\n' "$BUILD_DIR"
+cd "$BUILD_DIR" || exit
```
markus2330: Is it necessary to start a fresh build here? What about splitting the configure phase and the copying phase into two scripts?

doc/help/CMakeLists.txt
```diff
@@ -16,5 +16,11 @@ if (RONN_LOC)
...
+	set (MAN_PAGES_DIR "${CMAKE_SOURCE_DIR}/doc/man")
+
+	if (INSTALL_DOCUMENTATION)
```
markus2330: What about providing a target that rebuilds all man pages using update-manpages? (Maybe EXCLUDE_FROM_ALL).

src/bindings/swig/ruby/CMakeLists.txt
sanssecours: As far as I can tell the compiler warning `unused-but-set-variable` is only supported by GCC. I changed the check to

```cmake
if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
```

.

doc/help/CMakeLists.txt
sanssecours: I do not think that makes that makes much sense, since calling `ninja update-manpages` or `update-manpages` does not make that much of a difference in my opinion.

scripts/update-manpages
sanssecours: > Is it necessary to start a fresh build here?

Not really. The script now checks the CMake cache and does not regenerate the build system if the option `BUILD_DOCUMENTATION` is enabled.

sanssecours:
This pull request should be ready. If I should change anything, then please just add a comment.

markus2330:
Thank you!

PR #1627, opened 2017-10-17
Uninstall: Remove Additional Files

sanssecours:
# Purpose

After this pull request `build-system uninstall` – where build-system is something like `make` or `ninja` – should remove (almost) all files produced by Elektra.

# Checklist

- [x] The issue reference in commit 5a093e4 should be correct.
- [x] I installed and uninstalled Elektra multiple time to check if there were any remaining files.


sanssecours:
I also tested the changes with the standard Elektra installation path, which resulted in an empty `/usr/local/bin` directory 😊. Anyway, I then reinstalled PGP, Homebrew and fixed the problematic code. After that I installed and uninstalled Elektra about three times, which seemed to work. At least `brew`, `gpg` and other tools still work fine on my machine. It might makes sense for someone else to review and test the changes before we merge this pull request though.

cmake/ElektraUninstall.cmake
```diff
@@ -28,3 +33,78 @@ foreach (file ${files})
...
+
+			if (NOT "${rm_retval}" STREQUAL 0)
+				message (FATAL_ERROR "Problem when removing ${dir}")
+			endif (NOT "${rm_retval}" STREQUAL 0)
+		endif (EXISTS "${dir}")
+	endforeach (directory ${directories})
+endfunction(remove_directories)
+
+set (DIRECTORIES
+	"/usr/local/include/elektra"
```
markus2330: Seems to ignore CMAKE_INSTALL_PREFIX?

cmake/ElektraUninstall.cmake
```diff
@@ -11,7 +11,12 @@ endif (NOT EXISTS "${MANIFEST}")
 
 #message (MANIFEST IS ${MANIFEST})
 
+# =========
+# = Files =
+# =========
+
 file (READ "${MANIFEST}" files)
+string (APPEND files "\n/usr/local/lib/python2.7/site-packages/elektra_gen-${KDB_VERSION}-py2.7.egg-info")
```
markus2330: Seems to ignore CMAKE_INSTALL_PREFIX? And also python version? (Please look into the swig binding how the path is concatinated)

cmake/ElektraUninstall.cmake
```diff
@@ -28,3 +33,78 @@ foreach (file ${files})
...
+	"/usr/local/share/doc/elektra"
+	"/usr/local/share/doc/elektra-api"
+	"/usr/local/share/elektra"
+	"/usr/local/share/share/elektra"
+	"/usr/local/lib/python2.7/site-packages/support"
+)
+
+remove_directories ("${DIRECTORIES}")
+
+# The following directories might be empty. The order of the directories is important, since we remove them in the given order.
```
markus2330: Seems to ignore CMAKE_INSTALL_PREFIX?

And it would be bad to remove directories we did not create. Isn't easy to fix though (we do not want to reimplement package managers).

sanssecours: > And it would be bad to remove directories we did not create.

Removing empty directories should not be a problem in my opinion. Do you know a scenario where this behavior would cause issues?

cmake/ElektraUninstall.cmake
sanssecours: I fixed it in my latest commit. Thank you for the advice.

cmake/ElektraUninstall.cmake
sanssecours: > And also python version? (Please look into the swig binding how the path is concatenated)

The problem should be fixed in my latest update. Thank you for the helpful input.

PR #1634, opened 2017-10-23
haskell-bindings-fix-tests

e1528532:
# Purpose

(Hopefully) allows the haskell tests to work even if elektra has only been built but not installed yet. Also recognized changes on the haskell binding source files so ninja recompiles it correctly.

# Checklist

- [X] I added unit tests (restructured ;) )

@sanssecours lets see if this also works on travis now, seems fine locally but maybe i had some leftovers that make it work although it shouldn't (did clean the build directory though)

### DIFF
--- a/src/bindings/haskell/CMakeLists.txt
+++ b/src/bindings/haskell/CMakeLists.txt
@@ -5,6 +5,7 @@ find_program (C2HS_EXECUTABLE c2hs)
 if (CABAL_EXECUTABLE) # set by find_program
 if (C2HS_EXECUTABLE)
 if (GHC_EXECUTABLE)
+if (NOT BUILD_STATIC)
 
 	set (CABAL_INCLUDE_DIRS "\"${CMAKE_SOURCE_DIR}/src/include\", \"${CMAKE_BINARY_DIR}/src/include\"")
 
@@ -128,14 +129,17 @@ if (GHC_EXECUTABLE)
 			set_property (TEST testhaskell_realworld_optimized PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib")
 		endif (ENABLE_KDB_TESTING)
 	endif (BUILD_TESTING)
+else (NOT BUILD_STATIC)
+	remove_binding (haskell "BUILD_STATIC is currently not compatible with the haskell bindings")
+endif (NOT BUILD_STATIC)
 else (GHC_EXECUTABLE)
-	remove_binding (c2hs_haskell "GHC not found")
+	remove_binding (haskell "GHC not found")
 endif (GHC_EXECUTABLE)
 else (C2HS_EXECUTABLE)
-	remove_binding (c2hs_haskell "c2hs not found")
+	remove_binding (haskell "c2hs not found")
 endif (C2HS_EXECUTABLE)
 else (CABAL_EXECUTABLE)
-	remove_binding (c2hs_haskell "cabal not found")
+	remove_binding (haskell "cabal not found")
 endif (CABAL_EXECUTABLE)
 
 mark_as_advanced (
--- a/src/bindings/haskell/CMakeLists.txt
+++ b/src/bindings/haskell/CMakeLists.txt
@@ -8,6 +8,32 @@ if (GHC_EXECUTABLE)
 
 	set (CABAL_INCLUDE_DIRS "\"${CMAKE_SOURCE_DIR}/src/include\", \"${CMAKE_BINARY_DIR}/src/include\"")
 
+	execute_process (
+		COMMAND ${GHC_EXECUTABLE} --numeric-version
+		OUTPUT_VARIABLE GHC_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE
+	)
+
+	set (BINDING_HASKELL_NAME "${CMAKE_CURRENT_BINARY_DIR}/dist/build/libHSlibelektra-haskell-${KDB_VERSION}")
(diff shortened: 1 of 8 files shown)


e1528532:
the ffi lib should be independent of wheter elektra is installed or not, it is a ghc component. But strange that you get this issue, my rts folder looks basically the same:

```
ls /usr/local/Cellar/ghc/8.2.1/lib/ghc-8.2.1/rts
libCffi.a				libHSrts_l-ghc8.2.1.dylib
libCffi_debug.a				libHSrts_l.a
libCffi_l.a				libHSrts_p.a
libCffi_p.a				libHSrts_thr-ghc8.2.1.dylib
libCffi_thr.a				libHSrts_thr.a
libCffi_thr_debug.a			libHSrts_thr_debug-ghc8.2.1.dylib
libCffi_thr_l.a				libHSrts_thr_debug.a
libCffi_thr_p.a				libHSrts_thr_l-ghc8.2.1.dylib
libHSrts-ghc8.2.1.dylib			libHSrts_thr_l.a
libHSrts.a				libHSrts_thr_p.a
libHSrts_debug-ghc8.2.1.dylib		libffi.dylib
libHSrts_debug.a
```

What is your GHC_LIB_DIR path? Maybe it contains spaces that need quoting? Basically if the path is right cmake's find_library should recognize that libCffi.a corresponds to Cffi (and the same for HSrts).

sanssecours:
> What is your GHC_LIB_DIR path?

The command `ghc --print-libdir` prints the path `/usr/local/Cellar/ghc/8.2.1/lib/ghc-8.2.1`.

Seems like the Travis build job has the [same problem](https://travis-ci.org/sanssecours/libelektra/jobs/291563479#L2043).

e1528532:
I think that was caused by a leftover check for FFI_LIB which happened before it even searched for it. Should be fixed now.
I am still working on linux support, things are slightly different there i noticed.

markus2330:
@e1528532 Thank you for working on this! Would be good to get more bugs fixed to have a better release.

e1528532:
Ok after trying to mix static and shared libraries which works fine in osx and doesn't work at all in linux, i decided to go the full shared route as it seems "cleaner" anyway (e.g. build_shared -> make everything shared and build_static -> build everything static).

While with the default ENABLE_SHARED it worked both in osx and linux for me now. Now i have to find a way to make build_shared and build_full happy too.

src/bindings/haskell/CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ if (GHC_EXECUTABLE)
 	add_custom_target (c2hs_haskell ALL DEPENDS "${BINDING_HASKELL_NAME}")
 
 	# build and install it to the cabal db
-	install (CODE "execute_process (COMMAND ${CABAL_EXECUTABLE} install WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})")
+	install (CODE "execute_process (COMMAND ${CABAL_EXECUTABLE} install WORKING_DIRECTORY \"${CMAKE_CURRENT_BINARY_DIR}\")")
```
sanssecours: Thank you!

e1528532:
Ok so i've also made BUILD_FULL happy. BUILD_STATIC does still not work though, it tells me

```
Linking dist/build/testhaskell_realworld/testhaskell_realworld ...
  "_hs_elektraHaskellCheckConfig", referenced from:
      _elektraHaskellCheckConfig in libelektra-static.a(haskell.c.o)
  "_hs_elektraHaskellSet", referenced from:
      _elektraHaskellSet in libelektra-static.a(haskell.c.o)
  "_hs_elektraHaskellGet", referenced from:
      _elektraHaskellGet in libelektra-static.a(haskell.c.o)
  "typeinfo for std::__1::basic_streambuf<char, std::__1::char_traits<char> >", referenced from:
      typeinfo for std::__1::basic_stringbuf<char, std::__1::char_traits<char>, std::__1::allocator<char> > in libelektra-static.a(dump.cpp.o)
```

when building the haskell tests for the bindings. However when excluding the haskell binding tests the remaining tests do work (including the haskell plugin test). So i assume this is some issue with how the tests get linked. Tried a bunch of things already but nothing worked so far.

I am not sure if i can fix this in time and will do some of the java things now. We should at least include this PR before the release, as the haskell bindings don't really work without it.
I think shared and full are usually used and as these plugin/bindings are experimental anyway it shouldn't be the biggest deal that it doesn't yet support static builds in the meanwhile.

By the way, is it correct that build_shared, build_full and build_static are not meant to be mixed with each other? Those 3 compile flags aren't really explained in the documentation https://www.libelektra.org/docgettingstarted/compiling .

markus2330:
> I am not sure if i can fix this in time
> it shouldn't be the biggest deal that it doesn't yet support static builds in the meanwhile.

Yes, it is no big deal because by default both `BUILD_STATIC` and the `haskell` plugin are excluded. Please warn in the README and in the build process that these two options currently conflict.

> By the way, is it correct that build_shared, build_full and build_static are not meant to be mixed with each other? 

Yes, they can be intermixed without limitations.

> Those 3 compile flags aren't really explained in the documentation 

I added documentation in 9506cec701e3cbb7b839fa03f9afe2758751edee

e1528532:
ok, i think we can merge this then to have the shared and full builds for haskell bindings

e1528532:
Could it be that jenkins had an issue therefore the three builds there failed, which are unrelated to this PR?

markus2330:
Sorry, yes, these errors are unrelated. I tried to revert your change with the new constructor of PluginSpec because it seems to break ABI (there are some segfaults in the ABI tests), but seems like other parts are not compiling anymore because of other changes in the other constructors (maybe the explicit?). I am working on it.

Issue #1633, opened 2017-10-21
spec-mount rest-backend-config does not work any more


BernhardDenner:
# Describe what you wanted to do

prepare the rest-backend configuration with `kdb mount-rest-backend-config`, which involves

```shell
kdb mount rest-backend-spec.ini spec/sw/elektra/restbackend/#0/ ni
kdb spec-mount /sw/elektra/restbackend/#0/ ni
```

This leads to

```
Got mountpoint from spec/sw/elektra/restbackend/#0/current with 33 nodes, configfile:  and mountpoint: /sw/elektra/restbackend/#0/current/backend/api/description/html
The command kdb spec-mount terminated unsuccessfully with the info:
The path you entered is invalid.
Try to add another path instead.

Filenames are typically (depending on
resolver) not allowed to contain '..'.

For more information see:
 kdb info <your resolver>
Please report the issue at https://issues.libelektra.org/
```

# Describe what you expected

rest-backend-config.ini mounted according to defined specification.

# Describe what actually happened

The command `kdb spec-mount /sw/elektra/restbackend/#0/ ni` gives the above error message. 

Steps to reproduce:

```shell
# in case you do not have rest-backend compiled
cp <libelektra-src>/src/tools/rest-backend/config-specifiction.ini.in /usr/local/share/elektra/specification/rest-backend-config.ini
# manually substitute cmake vars
vim /usr/local/share/elektra/specification/rest-backend-config.ini
kdb mount rest-backend-config.ini spec/sw/elektra/restbackend/#0 ni
kdb spec-mount /sw/elektra/restbackend/#0
```

The elektra version GIT sha1: e5624dd2ddb1a9ee09e467a9c65f40f1cea76925 (master from end of May) works as expected.

# System Information

- Elektra Version: master
- verified on Debian 9 and Ubuntu 16.04

BernhardDenner:
Some debugging I did:

The comand `kdb ls spec/sw/elektra/restbackend` gives

with the current master build (spec-mount not working)

```
/sw/elektra/restbackend/#0/current/backend/api/description/html
...
spec/sw/elektra/restbackend/#0/current
...
```

while the old build (from mentioned GIT hash - spec-mount working)

```
spec/sw/elektra/restbackend/#0/current
...
```

So, now the `kdb ls` contains cascading keys too.
I've done a short verification if this triggers the but with the help of the Ruby bindings

```ruby
require 'kdbtools'

db = Kdb.open
[...]

PR #1638, opened 2017-10-23
specreader: ignore non-spec keys

BernhardDenner:
# Purpose

Fixes issue #1633
Don't know if this patch may be considered to be an ugly patch, also I didn't find a nicer way to filter spec keys. However, this filters out all non-spec keys (in that case all cascading) before we try to parse the specification and build the backend.

# Checklist

- [ ] I added unit tests


markus2330:
Thank you for the bug triage and the fix!

With std::remove_if and lambdas you could do the same with a one-liner, but your fix might be more clear for people with less C++ background.

PR #1656, opened 2017-10-28
Build: Simplify Man Page Generation

sanssecours:
# Purpose

This pull request simplifies man page generation. Now the target `man` will directly put the generated man pages into `src/build/man`. Before commit 5c96061 we would put them into `build/doc/help` instead and then copy them to `src/build/man` afterwards.

# Checklist

- [x] I checked all commit messages. The issue reference in commit 5c96061 should be correct.
- [x] I tried building Elektra with and without `ronn`.
- [x] I modified one of the input sources for the man pages (`src/libs/README.md`), which resulted in an updated man page after I invoked `ninja man`.
- [x] After I installed Elektra `man kdb` printed the correct man page.


markus2330:
Thank you!

PR #1657, opened 2017-10-28
restbackend: spec use 0/1 for booleans

BernhardDenner:
The boolean plugin does not canonicalize boolean values here, since it
uses 'type' instead of 'check/type' meta key.
The type plugin requires integer values for boolean types.
So for a quick fix, provide valid defaults.

I wonder if this ever worked ...

- [ ] I added unit tests


markus2330:
Sounds like a practical solution, thank you!

PR #1658, opened 2017-10-28
build-homepage: late fixes

BernhardDenner:
Do we really want to run the backend service as an ASAN build? It already crashed two times because of that.
    
Remove installed libraries before installing the new one. This way we ensure there are no old unwanted plugins that might cause problems.

Ensure rest-backend is started in daemon mode, otherwise the script will be blocked.

### DIFF
--- a/scripts/build-homepage
+++ b/scripts/build-homepage
@@ -19,7 +19,7 @@ fi
 C_FLAGS='-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-protector-strong -Wstack-protector -fPIE -pie'
 LD_FLAGS='-Wl,-z,now -Wl,-z,relro'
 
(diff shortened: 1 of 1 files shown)


markus2330:
> Do we really want to run the backend service as an ASAN build?

Yes, we want to avoid that people can exploit undefined code. But if it does not work otherwise, we can disable ASAN for the release.

> This way we ensure there are no old unwanted plugins that might cause problems.

Why not remove it completely? Or move it once so that we have a backup of the last-working situation.

> Ensure rest-backend is started in daemon mode, otherwise the script will be blocked.

Shouldn't that be set by puppet?

markus2330:
Please revert if you find a better solution.

markus2330:
Thanks for your quick fix!

Issue #1670, opened 2017-10-29
haskell: building bindings failed


markus2330:
```
Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
update' to download it.
Resolving dependencies...
Warning: solver failed to find a solution:
Could not resolve dependencies:
rejecting: libelektra-haskell-0.8.20:!test (constraint from config file,
command line flag, or user target requires opposite flag selection)
trying: libelektra-haskell-0.8.20:*test
unknown package: QuickCheck (dependency of libelektra-haskell-0.8.20:*test)
Dependency tree exhaustively searched.
Trying configure anyway.
Configuring libelektra-haskell-0.8.20...
cabal: Encountered missing dependencies:
QuickCheck -any, hspec -any
```

Seems like a `cabal update && cabal install QuickCheck hspec` is missing to make the binding happy. Docu is quite sparse and does not mention this at all.

e1528532:
The build seems to be fixed with the PR. I think @sanssecours did some changes to have it treated differently by the build server / have an own build for that. Is that still necessary? If we can just build this in the normal macOS build we can free up valuable macOS instances so the travis builds are faster.

sanssecours:
>  I think @sanssecours did some changes to have the haskell bindings treated differently by the build server / have an own build for that. Is that still necessary?

If you want to continue testing the Haskell bindings, then the answer is probably yes. We could add Haskell to the general build instances, but I fear that would mean that Travis would report a lot of timeouts (Mac build jobs can only run for [50 minutes](https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts) until they are canceled).

sanssecours:
I added Haskell to the main Mac build jobs in a [branch of my repo](https://travis-ci.org/sanssecours/libelektra/builds/295198476). Unfortunately we have to wait until [Travis works correctly again](https://www.traviscistatus.com) to see the results.

markus2330:
As haskell is now more important (@e1528532 is actively working on it) we could remove some other less important binding/plugin to reduce build time. What above removing java binding+plugin from the build, would that change the build time? (It might be needed by @vanessaNWK soon but not now.) Or we remove the qt-gui from the build. (Also no changes planned in the near future.)

Is it possible to talk to travis like to jenkins (from github)? Then we could start excluded parts explicitly. (This feature is low priority).

sanssecours:
> What above removing java binding+plugin from the build, would that change the build time?

[...]

PR #1671, opened 2017-10-29
Haskell bindings check dependencies

e1528532:
# Purpose

Improves the dependency management of the haskell bindings mentioned in #1670 . The readme is also more precise about the dependencies now.

# Checklist

@markus2330 There is one little issue left. If one dependency is missing, and it excludes the bindings, then the haskell plugin will fail later on. This is because for what i've seen the plugins get configured first, so the bindings are still included and thus the haskell plugin is happy. But then it configures the bindings, notices some lib is missing and excludes the bindings. Thus then in the later phases the haskell plugin fails because it depends on a target added by the bindings, which obviously is missing now. 

We currently don't have a simple mechanism to handle such cases, do we? Anyway its probably not the biggest deal as the user will see that the bindings got excluded, it just would be nicer to hide the cmake error which follows then.

### DIFF
--- a/src/bindings/haskell/CMakeLists.txt
+++ b/src/bindings/haskell/CMakeLists.txt
@@ -1,29 +1,7 @@
-find_program (GHC_EXECUTABLE ghc)
-find_program (CABAL_EXECUTABLE cabal)
-find_program (C2HS_EXECUTABLE c2hs)
-find_program (GHC-PKG_EXECUTABLE ghc-pkg)
+find_package (Haskell)
 
-if (CABAL_EXECUTABLE) # set by find_program
-if (C2HS_EXECUTABLE)
-if (GHC_EXECUTABLE)
-if (GHC-PKG_EXECUTABLE)
+if (HASKELL_FOUND) 
 if (NOT BUILD_STATIC)
-
-# check for hspec and QuickCheck
-# ghc-pkg return code is 0 on success, 1 otherwise
-execute_process (
-	COMMAND ${GHC-PKG_EXECUTABLE} latest hspec
-	RESULT_VARIABLE GHC_HSPEC_FOUND
-	OUTPUT_QUIET ERROR_QUIET
-)
-if (GHC_HSPEC_FOUND EQUAL 0)
-
-execute_process (
-	COMMAND ${GHC-PKG_EXECUTABLE} latest QuickCheck
-	RESULT_VARIABLE GHC_QUICKCHECK_FOUND
-	OUTPUT_QUIET ERROR_QUIET
-)
-if (GHC_QUICKCHECK_FOUND EQUAL 0)
 
 	set (CABAL_INCLUDE_DIRS "\"${CMAKE_SOURCE_DIR}/src/include\", \"${CMAKE_BINARY_DIR}/src/include\"")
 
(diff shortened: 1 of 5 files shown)


src/bindings/haskell/CMakeLists.txt
```diff
@@ -129,9 +147,18 @@ if (NOT BUILD_STATIC)
 			set_property (TEST testhaskell_realworld_optimized PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib")
 		endif (ENABLE_KDB_TESTING)
 	endif (BUILD_TESTING)
+else (GHC_QUICKCHECK_FOUND EQUAL 0)
+	remove_binding (haskell "QuickCheck library not found")
```
markus2330: give some reference to README or say what to do here. (also for others)

src/bindings/haskell/README.md
```diff
@@ -4,12 +4,33 @@ A full haskell binding using c2hs.
...
 GHC, cabal and C2HS, which generates the actual bindings out of a metacode and 
-will be called during the build automatically (Call `cabal install c2hs` to 
-install it globally).
+will be called during the build automatically.
+
+There are some examples how the bindings can be used in
+[examples](src/bindings/haskell/examples). Furthermore there exists the possibility 
+to write [plugins](src/plugins/haskell/) using the haskell bindings.
+
+To build and test it manually call `cabal test`. Otherwise the bindings will be
```
markus2330: It is unclear in which folder to execute `cabal test`.

src/bindings/haskell/README.md
```diff
@@ -4,12 +4,33 @@ A full haskell binding using c2hs.
...
+GHC and cabal can be either obtained using a package manager or downloaded directly
+from [the Haskell homepage](https://www.haskell.org/platform/).
+
+Before installing the remaining dependencies, make sure to update your cabal database
+which contains the current list of haskell packages by invoking `cabal update`, this
+is not done automatically.
+
+- c2hs `cabal install c2hs` > 0.28.2 (older versions untested)
+- hspec `cabal install hspec` (only if the tests are being built, v2.4.4 tested)
+- QuickCheck `cabal install QuickCheck` (only if the tests are being built, v2.10.1 tested)
```
markus2330: For convenience you could give a single command which does everything.

Isn't it better to install c2hs from the package manager? (At least it is included in Debian.)

e1528532: if it is available in the package manager it most likely is. In homebrew its not available so therefore i am used to install it via cabal, but i'll add this recommendation

e1528532:
> Currently you need to duplicate the checks

So you mean make the haskell plugin, which doesn't basically need hspec depend on it as well, because if there is no hspec, the bindings will be missing?

> Maybe write a "FindHaskell" that does everything?

Ah so this plugin would just set a single HASKELL_FOUND variable and maybe a second one with details of what went wrong, instead of repeating the checks everywhere? Thats a good idea i think. Is it fine for the release if i do it tomorrow, its slowly getting late

markus2330:
> So you mean make the haskell plugin, which doesn't basically need hspec depend on it as well, because if there is no hspec, the bindings will be missing?

Yes.

> Ah so this plugin would just set a single HASKELL_FOUND variable and maybe a second one with details of what went wrong, instead of repeating the checks everywhere?

Exactly

> Is it fine for the release if i do it tomorrow, its slowly getting late

Perfectly fine.

e1528532:
Added the improvements. Somehow travis seems to have an issue with mac builds, even the one started yesterday is still queued https://travis-ci.org/ElektraInitiative/libelektra/builds/294567753?utm_source=github_status&utm_medium=notification . It would be interesting though to see the result of the HASKELL=ON build to judge whether it only works locally or at least on the build server too.

markus2330:
Maybe @sanssecours has an idea. Seems like everything rebuilds now anyway.

sanssecours:
The Mac build are just fine. You need to wait a little bit though, since Travis only allows [two concurrent Mac builds](https://blog.travis-ci.com/2017-09-22-macos-update) for a single repository. If you want jobs to run faster you can either cancel old jobs, or just register with your GitHub account at https://travis-ci.org and let Travis run on your own version of the repository. It is really easy. Even I was able to do it 😊.

e1528532:
Hmm, i added my fork of elektra to travis but it doesn't recognize my branches so i can't trigger a build for my version of this branch. It simply says no other branches for this repository. Any idea what i've missed? Do i really have to commit something to trigger a build?

sanssecours:
As far as I can tell the only thing I did apart from registering was flipping the switch for the repository:

switch

.

> Do i really have to commit something to trigger a build?

You could publish a copy of your current branch under a different name. By default Travis should trigger builds for new branches.

e1528532:
Right, it only seems to work for new/updated branches. Triggered my own build at https://travis-ci.org/e1528532/libelektra/builds/294981392 now, lets see

markus2330:
Thank you, your build was successful!

PR #1673, opened 2017-10-29
build-homepage: build frontend with --force

BernhardDenner:
The Backend throws an excpetion if 'database' endpoint is called without
a 'sortBy' field. Therefore the backend returns with HTTP 500 which
aborts the frontend build process.

For now, we simply force the frontend build to be not blocked by this
minor issue.


markus2330:
Doesn't this disable the site map?

Do you want to use this PR to directly build the website or are there issues doing so?

PR #1677, opened 2017-10-31
haskell-bindings-fix-regression: add dependency as the binding librar…

e1528532:
# Purpose

I fear i introduced a little regression by the latest change which makes the build non-deterministic, depending on the order in which cmake configures/compiles things. This should fix it again, initially i removed the dependency because of the static builds having a cyclic dependency then (elektra depending on haskell plugins, haskell plugins depending on haskell bindings, haskell bindings depending on elektra) but since static builds are not supported anyway for the haskell bindings it should be fine for the meanwhile.

# Checklist

- [X] I ran all tests and everything went fine (at least locally)

@sanssecours i hope this resolves the issue you reported in #1670. My own build is triggered here in case travis is slow again/still https://travis-ci.org/e1528532/libelektra/builds/295256480


sanssecours:
jenkins build elektra-gcc-configure-debian-withspace please

sanssecours:
jenkins build elektra-mergerequests-stable please

sanssecours:
Thank you!

PR #1724, opened 2017-12-10
kdb: proposal for nokdb profile

markus2330:
# Purpose

When KDB is severely broken (e.g. KDB.get does not work) currently the kdb tool always emits errors when trying to read its own configuration.

This is a proposal to add the profile "nokdb" which allows to not use KDB for the configuration of the kdb-tool. (Of course the main actions of `kdb` still use KDB.)

- This idea might be a misfeature (to give profiles a semantics), so I thought it would be good to discuss it first.
- The formatting is not really nice.

# Checklist

needs to be checked.

TODO:

- [ ] I added unit tests

### DIFF
--- a/doc/help/kdb.md
+++ b/doc/help/kdb.md
@@ -76,7 +76,7 @@ For example, to permanently change verbosity one can use:
 
 Profiles allow users to change many/all configuration options of a tool
 at once. It influences from where the KDB entries are read.
-For example if you use:  
+For example if you use:<br>
 	`kdb export -p admin system`
 
 It will read its format configuration from `/sw/elektra/kdb/#0/admin/format`.
@@ -88,13 +88,15 @@ be chosen automatically according to the current user or current working directo
 Sometimes it is useful to start with default options, for example it is not
 possible to invert the `-q` option.
 In such situations one can simply select a non-existing profile, then `-q`
-works as usual:  
+works as usual:<br>
 	`kdb mount -p nonexist -q /abc dir/abc`
 
 There are two special profiles:
 
-- `%` is a fallback profile. It does not need to be selected and will be used if a key cannot be found in the main profile.
-- `nokdb` disables reading from `KDB`. Then only command-line arguments are used.
+- `%`:
+  Is a fallback profile. It does not need to be selected and will be used if a key cannot be found in the main profile.
+- `nokdb`:
+  Disables reading from `KDB`. Then only command-line arguments are used.
 
 ## BOOKMARKS
 
--- a/doc/help/kdb.md
+++ b/doc/help/kdb.md
@@ -59,10 +59,13 @@ The `kdb` utility reads its own configuration from three sources
 within the KDB (key database):
 
 1. /sw/kdb/**profile**/ (for legacy configuration)
-2. /sw/elektra/kdb/#0/%/ (for empty profile)
-3. /sw/elektra/kdb/#0/**profile**/ (for current profile)
+2. /sw/elektra/kdb/#0/%/ (for empty profile, `%` in Elektra
+   is used to represent a emptiness)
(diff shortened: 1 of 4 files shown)


markus2330:
> If we can not read the database, does it still make sense to use the kdb tool at all?

Yes, I even had such a case. It always makes sense when the mountpoint where kdb reads its configuration from is a different mountpoint than the one you work with.

I had for example that .config/default.ecf was a broken dump file (mounted at user) but all other namespaces where working fine.

With this PR the only difference is that no error is shown (if `-p nokdb` is passed), the functionality is not effected. At the moment you can also use the `kdb` tool to work with non-broken mountpoints even if the mountpoint with kdb configuration is broken.

markus2330:
I added a bit of docu and also fixed the other case.

The only remaining issue for me if we should keep the name "nokdb". An alternative would be to use "%", because it does not make sense to use it as profile otherwise anyway.

doc/help/kdb.md
```diff
@@ -76,7 +76,7 @@ For example, to permanently change verbosity one can use:
 
 Profiles allow users to change many/all configuration options of a tool
 at once. It influences from where the KDB entries are read.
-For example if you use:
+For example if you use:
```
sanssecours: Did you add the whitespace here and in line 91 intentionally?

sanssecours:
> The only remaining issue for me is if we should keep the name "nokdb". An alternative would be to use "%", because it does not make sense to use it as profile otherwise anyway.

Is `%` not already used for the fallback profile? As far as I can tell, we could add fallback values for the `kdb` configuration to `%`. Would `kdb get -p % …` then read the fallback values or ignore the configuration? 

I am not sure if the text above describes a valid problem at all though, since I usually do not use profiles at all.

markus2330:
> Is % not already used for the fallback profile?

Yes, exactly.

> As far as I can tell, we could add fallback values for the kdb configuration to %. Would kdb get -p % … then read the fallback values or ignore the configuration?

Yes, it would read the fallback twice and not read the "current" configuration.

> I am not sure if the text above describes a valid problem at all though,

You are right, it is a valid problem. But to solve the problem *any* not-used profile works. "%" leads to a double-lookup and is not better suited than any other profile.

> since I usually do not use profiles at all.

I'll add profiles as topic for the next meeting. It is an important underused functionality.

doc/help/kdb.md
markus2330: Yes, unfortunately Markdown requires such whitespaces to indicate a line break (but not a paragraph break). Without these two spaces the resulting man page is ugly. (Example would not start in a new line.)

I also dislike it but it seems to be the way how it is done in Markdown.

sanssecours: Can we please (please please 🙏) replace these two trailing spaces with the HTML tag `<br />`. The tag should have exactly the same effect as the two trailing spaces, but makes your intention much clearer to someone who looks at the plain text file.

markus2330: If all of our 4 markdown renderer (website, ronn, GitHub, discount in qt-gui) support the `<br>` I also prefer to use that. Then we can systematically replace all occurrences of two spaces.

sanssecours: > If all of our 4 markdown renderer (website, ronn, GitHub, discount in qt-gui) support the `<br>` I also prefer to use that.

I tried GitHub’s Markdown renderer, `discount` (`qt-gui`) and `ronn`. The tag `<br>` works in all of these applications. I assume the Website supports the `<br>` tag too.

I already replaced every occurrence of two or more spaces at the end of a line with the `<br>` tag in my local copy of the repo. The output of `ronn` looks a little bit different in some places, while in other places `ronn` produces exactly the same output as before. I will open a pull request containing the changes later.

Issue #1764, opened 2017-12-31
Travis jobs do not report failures


markus2330:
As discussed in #1698 there seems to be a severe problem in the build logic of travis. Even though test cases fail, the build job succeeds:

```
        Start  67: testshell_hosts
 67/110 Test  #67: testshell_hosts .........................***Failed    1.19 sec
kdb mount /tmp/hosts system/tmount/hosts hosts --with-recommends
Return value “7” does not match “0”
kdb set system/tmount/hosts/ipv4/localhost 127.0.0.1
kdb ls system/tmount/hosts
ERROR - STDOUT:
“system/tmount/hosts/ipv4/localhost”
does not match
“system/tmount/hosts(⏎system/tmount/hosts/ipv[4,6]/localhost){1,2}”
kdb get system/tmount/hosts/ipv4/localhost
kdb set system/tmount/hosts/ipv4/localhost ::1
Return value “0” does not match “5”
kdb umount system/tmount/hosts
shell_recorder /Users/travis/build/ElektraInitiative/libelektra/tests/shell/shell_recorder/hosts.esr RESULTS: 22 test(s) done 3 error(s).
📕
Protocol File: /var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/elektraenv.XXXXXXXXX.y92sbEwY

...

The following tests FAILED:
	 67 - testshell_hosts (Failed)
Errors while running CTest
FAILED: tests/CMakeFiles/run_all 
cd /Users/travis/build/ElektraInitiative/build && /Users/travis/build/ElektraInitiative/libelektra/scripts/run_all RelWithDebInfo
ninja: build stopped: subcommand failed.
```
This error is ignored and afterwards the local tests start.

https://travis-ci.org/ElektraInitiative/libelektra/jobs/323217679

PR #1765, opened 2017-12-31
Travis: Always Consider Exit Status of `run_all`

sanssecours:
# Purpose

This commit should fix the problem reported in issue #1764.

# Checklist

- [x] I checked the commit message. The issue reference in commit dfcd490 should be correct.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -163,7 +163,6 @@ script:
       # Unfortunately the Haskell tests do not work currently:
       # https://github.com/ElektraInitiative/libelektra/issues/1631
       if [[ "$HASKELL" != "ON" ]]; then
-        ninja run_all
-        kdb run_all
+        ninja run_all && kdb run_all
       fi
     fi


markus2330:
Happy new year!

Are you sure this will fix the problem? If the language used in travis is similar to sh, this will only avoid `kdb run_all` to be  executed but will not fix the overall exit status.

In my /bin/sh shell, the "fix" triggers the problem we want to solve:

```
#!/bin/sh

set -ex

true

if true
then
	false
	true
fi

true
```

executes as wanted:
```
+ true
+ true
+ false
Exit-status: Failure
```

but:

```
#!/bin/sh

set -ex

true

if true
then
	false && true
fi

true
```

executes like:

```
+ true
+ true
+ false
+ true
Exit-status: Success
```

which would lead to the problem as described in #1764. But travis seems to have different semantics than my shell.

sanssecours:
> Happy new year!

🍾 Happy new year to you too 🥂.

> Are you sure this will fix the problem?

I would assume so. I pushed a debug commit, to check if everything works as expected.

sanssecours:
Seems the updated code works as [expected](https://travis-ci.org/ElektraInitiative/libelektra/jobs/323746774#L2324).

markus2330:
> 🥂

🥂

Thank you for checking!

Please add ready to merge when you reverted the debug commit.

markus2330:
Thank you for the fix!

Issue #1768, opened 2018-01-03
Global Initialization Logic for Plugins


e1528532:
Unfortunately the GHC runtime has a deficiency that renders it unable to reinitialize the runtime system after it has been closed down (see https://ghc.haskell.org/trac/ghc/ticket/13693). While it most likely does not matter for normal use cases it is a real show stopper for using it with libelektra, as libelektra dynamically loads and unloads plugins all the time in one process. It looks like the runtime can only be started once per process, even if the associated shared libraries are unloaded inbetween, otherwise it will lead to random behavior like deadloops or segfaults or similar things.

I've already tried all sorts of "easy solutions" in the form of different ways of starting/shutting down the runtime, not doing a real shutdown but manually hack some cleanup actions together picked from the hs_exit routine, and similar, but none of them worked and i slowly run out of ideas.

So basically i see two options here:

- Grab the ghc sources and try to fix this issue there. The downside is that this is quite unpredictable, it could be possibly a quick fix maybe, or its going to be a matter of weeks, or i won't ever find a solution in the worst case. In any way if I managed to fix it it would still take a long time until it reaches upstream ghc versions, especially on more conservative distributions like debian, so i'd avoid this approach.

- Somehow build global plugin initialization logic into libelektra itself which ensures the runtime gets started up before any other work is done (or lazily upon the first hs plugin call) and then gets unloaded after everything is done and the process is about to exit, so that it remains open in a memory location that doesn't get unloaded during the process lifetime.

i thought about using a global plugin but it seems to suffer from the same issue - loading and unloading everything dynamically which makes it unsuitable to use either. Maybe there is already something like i need available and i'm just not aware about it? If not we should probably discuss how such system would look like.

So summing up the requirements:
- load the haskell runtime before any haskell plugin gets called
- keep the haskell runtime and the associated rts libraries open throughout the process lifetime
- exit once there will be no more plugin calls, if it hasn't been started from a haskell context (e.g. standalone haskell application using the bindings)

markus2330:
Thank you for bringing this up before crashes are reported! We should really write a design decision to document this once and for all.

> Grab the ghc sources and try to fix this issue there.

I am afraid that the effort for this would be too high. Most likely they have much global state.

It is discussed in many issues (e.g. #767), and and a proposal was written in #418.

[...]

PR #1774, opened 2018-01-11
pluginprocess

e1528532:
# Purpose

This adds a library that allows plugins to be executed in a child process to workaround the global initialization issue described in #1768. 

It uses fork with a simple communication protocol based on elektra itself, invoking the dump plugin to let a plugin communicate with its child process via named pipes (as the dump plugin expects a file as the input argument to serialize from/to) using a few extra keys which indicate which plugin function to execute and what was the original key for plugin function calls.

# Checklist

- [ ] I added unit tests

@markus2330 One more puzzle piece towards the typechecker. Actually, I have already used this library for it and indeed my reinitialization issues are now gone with it. I will make a separate PR which uses this for haskell plugins though to keep the PRs smaller (then you also see how its used exactly). You can preview it [here ](https://github.com/e1528532/libelektra/blob/haskell-pluginprocess/src/plugins/haskell/haskell.c.in).

Is some more in-depth documentation for this required? Basically i have added small examples to the functions and it's usage in the base haskell plugin should serve as a perfect example for it. We should maybe add a hint to the developer docu so other people encountering such reinitialization issues will be guided towards this library. As you suggested I've designed it for general usage, not only with haskell plugins.

Some branches that build on this (and the cabal sandboxing) thus i have no PRs for them yet:

[Using this library for haskell plugins](https://github.com/e1528532/libelektra/tree/haskell-pluginprocess)

and ultimately....
[Typechecker Prototype](https://github.com/e1528532/libelektra/tree/typechecker-prototype)
(though i still need to adjust one thing or two with it tomorrow before it can be finally tried out and it needs some README how to use it, currently i still have a path hardcoded)

### DIFF
--- a/src/error/specification
+++ b/src/error/specification
@@ -1190,8 +1190,8 @@ module:pluginprocess
 macro:PLUGIN_PROCESS_INITIALIZATION
 
 number:191
-description:An issue happened while executing a pluginprocess command for a plugin
-severity:warning
+description:Failed to communicate with the child process
+severity:error
 ingroup:plugin
 module:pluginprocess
 macro:PLUGIN_PROCESS_COMMAND
--- a/src/error/specification
+++ b/src/error/specification
@@ -1181,3 +1181,17 @@ severity:error
 ingroup:plugin
 module:directoryvalue
 macro:DIRECTORY_VALUE_ARRAY
+
+number:190
+description:Failed to initialize the pluginprocess library for a plugin
+severity:error
+ingroup:plugin
+module:pluginprocess
+macro:PLUGIN_PROCESS_INITIALIZATION
+
+number:191
+description:An issue happened while executing a pluginprocess command for a plugin
+severity:warning
+ingroup:plugin
+module:pluginprocess
+macro:PLUGIN_PROCESS_COMMAND
--- a/src/libs/pluginprocess/pluginprocess.c
+++ b/src/libs/pluginprocess/pluginprocess.c
@@ -260,13 +260,23 @@ static int elektraPluginProcessFork (ElektraPluginProcess * pp, Key * errorKey)
 	return 0;
 }
 
-static char * getPipename (Plugin * handle, char * suffix)
+static char * getPipename (Key * errorKey, char * suffix)
 {
 	int pipenameSize;
 	char * pipename;
-	pipenameSize = snprintf (NULL, 0, "/tmp/libelektra-process-%p-%s", (void *)handle, suffix);
+	pipenameSize = snprintf (NULL, 0, "/tmp/libelektra-process-%s-XXXXXX", suffix);
 	pipename = elektraMalloc (pipenameSize + 1);
-	pipenameSize = snprintf (pipename, pipenameSize + 1, "/tmp/libelektra-process-%p-%s", (void *)handle, suffix);
+	pipenameSize = snprintf (pipename, pipenameSize + 1, "/tmp/libelektra-process-%s-XXXXXX", suffix);
+
+	int prevErrno = errno;
+	if (mktemp (pipename) == NULL || errno == ENOTDIR || errno == EINVAL)
+	{
+		ELEKTRA_SET_ERRORF (190, errorKey, "Failed to generate a temporary pipename, mktemp returned errno %d", errno);
+		elektraFree (pipename);
+		errno = prevErrno;
+		return NULL;
+	}
+	errno = prevErrno;
 	return pipename;
 }
 
@@ -303,13 +313,24 @@ int elektraPluginOpen (Plugin * handle, Key * errorKey)
  * @retval a pointer to the information
  * @ingroup processplugin
  **/
-ElektraPluginProcess * elektraPluginProcessInit (Plugin * handle, Key * errorKey)
+ElektraPluginProcess * elektraPluginProcessInit (Key * errorKey)
 {
 	// First time initialization
 	ElektraPluginProcess * pp;
 	pp = elektraMalloc (sizeof (ElektraPluginProcess));
-	pp->commandPipe = getPipename (handle, "command");
-	pp->resultPipe = getPipename (handle, "result");
+	pp->commandPipe = getPipename (errorKey, "command");
+	if (pp->commandPipe == NULL)
+	{
+		elektraFree (pp);
+		return NULL;
+	}
+	pp->resultPipe = getPipename (errorKey, "result");
+	if (pp->resultPipe == NULL)
+	{
+		elektraFree (pp->commandPipe);
+		elektraFree (pp);
+		return NULL;
+	}
 	pp->dump = elektraInvokeOpen ("dump", 0);
 	pp->counter = 0;
 	if (pp->commandPipe && pp->resultPipe && pp->dump)
--- a/src/libs/pluginprocess/pluginprocess.c
+++ b/src/libs/pluginprocess/pluginprocess.c
@@ -10,17 +10,15 @@
  */
 
 #include "kdbpluginprocess.h"
-#include <kdbinvoke.h>
-// To access the plugin function pointers
 #include <kdberrors.h>
+#include <kdbinvoke.h>
 #include <kdblogger.h>
-#include <kdbprivate.h>
+#include <kdbprivate.h> // To access the plugin function pointers
 
 #include <errno.h>
 #include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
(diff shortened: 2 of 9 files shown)


src/error/specification
```diff
@@ -1181,3 +1181,17 @@ severity:error
...
+
+number:190
+description:Failed to initialize the pluginprocess library for a plugin
+severity:error
+ingroup:plugin
+module:pluginprocess
+macro:PLUGIN_PROCESS_INITIALIZATION
+
+number:191
+description:An issue happened while executing a pluginprocess command for a plugin
```
markus2330: This statement is really generic, except of "pluginprocess" this is true for every warning/error/log

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+ *
+ * Executes plugins in a separate process via fork and uses a simple
+ * communication protocol based on the dump plugin via named pipes.
+ *
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ */
+
+#include "kdbpluginprocess.h"
+#include <kdbinvoke.h>
+// To access the plugin function pointers
```
markus2330: Which header? Write it next to the include.

And why is this needed? Is kdbinvoke.h missing something?

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+ * is typically called in a plugin's open function.
+ *
+ * @param handle the plugin's handle
+ * @param pp the data structure containing the plugin's process information
+ * @see elektraPluginProcessInit how to use this function in a plugin
+ * @ingroup processplugin
+ **/
+void elektraPluginProcessStart (Plugin * handle, ElektraPluginProcess * pp)
+{
+	Key * commandPipeKey = keyNew ("/pluginprocess/commandPipe", KEY_VALUE, pp->commandPipe, KEY_END);
```
markus2330: please avoid camelCase or any other separator other than / in key names. Use `/pluginprocess/pipe/command` and `/pluginprocess/pipe/result`

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+
+		int result = ELEKTRA_PLUGIN_STATUS_ERROR;
+		if (*endPtr == '\0' && errno != ERANGE)
+		{
+			ELEKTRA_LOG ("C: We want to execute the command with the value %d now", command);
+			switch (command)
+			{
+			case ELEKTRA_PLUGIN_OPEN:
+				counter++;
+				result = handle->kdbOpen (handle, originalKey);
```
markus2330: Why calling functions of the plugin itself, isn't this a proxy plugin to call other plugins?

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+{
+	Key * commandPipeKey = keyNew ("/pluginprocess/commandPipe", KEY_VALUE, pp->commandPipe, KEY_END);
+	Key * resultPipeKey = keyNew ("/pluginprocess/resultPipe", KEY_VALUE, pp->resultPipe, KEY_END);
+	KeySet * keySet = ksNew (0, KS_END);
+	int counter = 0;
+	plugin_t command = ELEKTRA_PLUGIN_END;
+
+	do
+	{
+		ELEKTRA_LOG_DEBUG ("C: Wait for a KeySet");
```
markus2330: What does this "C:" mean? Why not writing "Child:" or even the pid?

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+	// Some plugin functions don't use keysets, but we need one for our
+	// simple communication protocol, so create a temporary one
+	KeySet * keySet = originalKeySet == NULL ? ksNew (3, KS_END) : ksDup (originalKeySet);
+	ksAppendKey (keySet, keyDup (key));
+	char * commandStr = intToStr (command);
+	ksAppendKey (keySet, keyNew ("/pluginprocess/command", KEY_VALUE, commandStr, KEY_END));
+	Key * originalKey = keyNew ("/pluginprocess/original", KEY_VALUE, keyName (key), KEY_END);
+	ksAppendKey (keySet, originalKey);
+
+	// Serialize, currently statically use dini our default format, this already writes everything out to the pipe
```
markus2330: Comment wrong: dump is used.

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+}
+
+/** Cleanup a plugin process
+ *
+ * This will decrease the internal counter how often open/close has been called, 
+ * closing opened pipes when the counter reaches 0. This will not delete the
+ * plugin data associated with the handle as it may contain other data out of
+ * the scope of this library, so this has to be done manually like
+ * @code
+int elektraPluginClose (Plugin * handle, Key * errorKey)
```
markus2330: leftover?

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+	pp->pid = pid;
+	ELEKTRA_LOG_DEBUG ("The pluginprocess is set using the pipes %s and %s with the pid %d", pp->commandPipe, pp->resultPipe, pid);
+	return 0;
+}
+
+static char * getPipename (Plugin * handle, char * suffix)
+{
+	int pipenameSize;
+	char * pipename;
+	pipenameSize = snprintf (NULL, 0, "/tmp/libelektra-process-%p-%s", (void *)handle, suffix);
```
markus2330: To use the handle as only distinction is maybe a good idea on systems with Address space layout randomization but we have no guarantees for that. So we should encode some additional randomness in this path. Did you consider to use mkdtemp? (see tests/cframework/tests.c 95). Alternatively you can take a look at elektraGenTempFilename (which could be called by libinvoke)

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -0,0 +1,379 @@
...
+	elektraInvoke2Args (pp->dump, "get", keySet, resultPipeKey);
+	ELEKTRA_LOG ("P: We received %zd keys in return", ksGetSize (keySet));
+
+	// Bring everything back in order by removing our process-related keys
+	Key * originalDeserializedKey = ksLookupByName (keySet, keyName (key), KDB_O_POP);
+	Key * resultKey = ksLookupByName (keySet, "/pluginprocess/result", KDB_O_POP);
+	char * endPtr;
+	int prevErrno = errno;
+	errno = 0;
+	int result = (int)strtol (keyString (resultKey), &endPtr, 10);
```
markus2330: This cast is ugly (why not store it in a long int?) and error checks are missing.

src/libs/pluginprocess/pluginprocess.c
e1528532: No i think we have a misunderstanding here. It is a library you can use to add the forking and communication functionality to your plugins, not a plugin on its own. [I'll link the usage example again and i hope this clears it up.](https://github.com/e1528532/libelektra/blob/haskell-pluginprocess/src/plugins/haskell/haskell.c.in).

I was not really sure how to design this as a plugin on its own, upon mounting people most like won't like to have to deal with that and simply mount a specification e.g. via `kdb mount /testSpecification.ini /spec/test ini typechecker` (assuming typechecker is a haskell plugin using this library) and not something like `kdb mount .... ini processplugin proxy=typechecker` (using a configuration to specify which plugin shall be proxied).

Maybe there is a simple way to have it as a plugin though, in that case I'd be fine with refactoring it to a plugin as the general logic most likely will stay pretty much the same.

src/libs/pluginprocess/pluginprocess.c
e1528532: the error checks are further down below, i guess this is not a good idea since you've overlooked them :) I'll refactor it.

src/libs/pluginprocess/pluginprocess.c
e1528532: No i think not, as its a library and people probably have other data stored in their plugin data handle besides the processplugin information i've left it to the plugin implementor to clear the handle afterwards by setting it to null (or do whatever he wants to do), and this function tells you when the plugin process has ended.

src/libs/pluginprocess/pluginprocess.c
markus2330: Ahh, okay. I see. Yes, it is a good idea to have it as library. Then plugins which do not work at all without being in a separate process can link against it.

For those who both modes make somehow sense (e.g. crypto) we should implement the plugin nevertheless. With your library writing such a plugin would be trivial anyway.

The usage would be `kdb mount file.ini ini process plugin=crypto` to have the crypto plugin running in a separate process. Let us discuss this today.

e1528532: Ok so this should be fine in the context that it's a library, a plugin that uses this library to implement the functionality you described will follow.

markus2330: Thank you! It is easier to comment if you create PRs, even if they are far from finished.

When looking at your code, I was missing a proxy for the contract: For the clause "infos/placements" you need to return what the plugin we are a proxy for returned. (You will need this logic also for the haskell plugin.). You can achieve this by renaming the respective clauses of the original plugin.

If you have any troubles implementing, please create the PR then we can easier discuss the problems.

e1528532:
i don't think this one error is related to my changes, as i only add a library in this PR without using it yet. Therefore i'd say it can be merged

sanssecours:
jenkins build all please

markus2330:
Which error do you mean? "jenkins build all please" (as called by @sanssecours just now) should always pass without any troubles.

sanssecours:
jenkins build mingw64 please

e1528532:
as far as i know mingw doesn't support fork or mkfifo, is it acceptable that this library will not work on these platform?

markus2330:
Yes, of course it is acceptable but you need to make sure that the library will not be built in mingw (or on any other non-POSIX platform). You can also check for fork/mkfifo, these checks might already exist for crypto.

e1528532:
Added a possibility to check the prerequisites for this library (Obviously using it for the library itself). 

I've designed it in a way so the check is also useable for plugins depending on this library. Ideally we'd automatically exclude all plugins depending on pluginprocess if it can't be built on the target platform, but i've found that to be way more effort to implement it than the current find approach (which plugins need to check as well), so i hope this is fine for now.

e1528532:
jenkins build elektra-multiconfig-gcc-stable please

e1528532:
jenkins build elektra-multiconfig-gcc-stable please

e1528532:
These tests are unfortunately slowing everything down i have to say, especially the travis ones. Everytime i fix something i have to wait almost a whole day until they all went through where i can't do much than hope they'll pass and if something fails the whole procedure starts again after trying a fix or trying to figure out what went wrong. I do run them locally beforehand but often issues only apply to a certain platform or configuration. 

Unfortunately there is yet another (new) issue which i can't reproduce locally and suspect to be some kind of race condition https://build.libelektra.org/jenkins/job/elektra-multiconfig-gcc-stable/179/build_type=Debug,plugins=ALL/console . @markus2330 Shouldn't i be able to restart the test by saying jenkins build elektra-multiconfig-gcc-stable please? It doesn't seem to listen to me.

sanssecours:
> Shouldn't i be able to restart the test by saying jenkins build elektra-multiconfig-gcc-stable please?

You need to remove the prefix`elektra-` from the build phrase. The correct build phrase is

```
```

. For more information, please take a look at [`doc/GIT.md`](https://github.com/ElektraInitiative/libelektra/blob/master/doc/GIT.md#run-jobs).

sanssecours:

e1528532:
this also looks good now. Tests are basically added with my next PR which includes a plugin based on this, so that everything is covered then.

markus2330:
Thank you, great job! Looking forward to a plugin based on this!

PR #1783, opened 2018-01-21
Process Plugin

e1528532:
# Purpose

Implements a plugin which uses libpluginprocess to encapsulate arbitrary plugins in an own process

# Checklist

- [X] I added unit tests
- [X] release notes are updated (doc/news/_preparation_next_release.md)

@markus2330 as promised

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -279,6 +279,5 @@ before_script:
     $TRAVIS_BUILD_DIR
 
 script:
-  - find /usr/ -name '*libclang_rt.asan-x86_64*'
   - ninja
   - ${TEST_COMMAND:-eval 'ninja install && ninja run_all && kdb run_all'}
--- a/.travis.yml
+++ b/.travis.yml
@@ -279,5 +279,6 @@ before_script:
     $TRAVIS_BUILD_DIR
 
 script:
+  - find /usr/ -name '*libclang_rt.asan-x86_64*'
   - ninja
   - ${TEST_COMMAND:-eval 'ninja install && ninja run_all && kdb run_all'}
--- a/cmake/Modules/LibAddPlugin.cmake
+++ b/cmake/Modules/LibAddPlugin.cmake
@@ -179,6 +179,7 @@ function (add_plugintest testname)
 endfunction (add_plugintest)
 
 function (plugin_check_if_included PLUGIN_SHORT_NAME)
+	set (NOT_INCLUDED "" PARENT_SCOPE)
 	list (FIND PLUGINS
 		   "-${PLUGIN_SHORT_NAME}"
 		   FOUND_EXCLUDE_NAME)
--- a/src/error/specification
+++ b/src/error/specification
@@ -1230,3 +1230,17 @@ severity:warning
 ingroup:plugin
 module:zeromqsend
 macro:ZEROMQSEND_ERROR
+
+number: 197
+description:Failed to initialize the execution of a plugin in a child process
+severity:error
+ingroup:plugin
+module:process
+macro:PROCESS_INITIALIZATION
+
+number: 198
+description:Plugin name to proxy in a process is invalid, it will not be loaded
+severity:warning
+ingroup:plugin
+module:process
+macro:PROCESS_INITIALIZATION
--- a/src/error/specification
+++ b/src/error/specification
@@ -1232,13 +1232,6 @@ module:zeromqsend
 macro:ZEROMQSEND_ERROR
 
 number: 197
-description:Failed to initialize the execution of a plugin in a child process
-severity:error
-ingroup:plugin
-module:process
-macro:PROCESS_INITIALIZATION
-
-number: 198
 description:Plugin name to proxy in a process is invalid, it will not be loaded
 severity:warning
 ingroup:plugin
--- a/src/libs/invoke/invoke.c
+++ b/src/libs/invoke/invoke.c
@@ -20,28 +20,6 @@ struct _ElektraInvokeHandle
 };
 
 /**
- * Structure for deferred calls
- * @internal
- */
-typedef struct _ElektraDeferredCall
-{
-	char * name;
-	KeySet * parameters;
-	struct _ElektraDeferredCall * next;
-} _ElektraDeferredCall;
-
-/**
- * Structure for internal plugin state
- * @internal
- */
-struct _ElektraDeferredCallList
-{
-	_ElektraDeferredCall * head;
-	_ElektraDeferredCall * last;
-};
-
-
-/**
  * @defgroup invoke
  * @brief Functionality to use plugins and invoke functions
  *
@@ -219,7 +197,7 @@ KeySet * elektraInvokeGetPluginConfig (ElektraInvokeHandle * handle)
  *
  * @pre handle must be as returned from elektraInvokeOpen()
  *
- * @return the name of the plugin
+ * @return the name of the plugin 
  */
 const char * elektraInvokeGetPluginName (ElektraInvokeHandle * handle)
 {
@@ -294,14 +272,16 @@ int elektraInvoke2Args (ElektraInvokeHandle * handle, const char * elektraPlugin
 {
 	if (!handle || !elektraPluginFunctionName) return -2;
 
-	typedef int (*elektra2Args) (Plugin *, KeySet *, Key *);
-	elektra2Args func = *(elektra2Args *) elektraInvokeGetFunction (handle, elektraPluginFunctionName);
+	// If we cast this right away it will hang if the function is not
+	const void * uncastedFunc = elektraInvokeGetFunction (handle, elektraPluginFunctionName);
 
-	if (!func)
+	if (!uncastedFunc)
 	{
 		return -2;
 	}
 
+	typedef int (*elektra2Args) (Plugin *, KeySet *, Key *);
+	elektra2Args func = *(elektra2Args *) uncastedFunc;
 	return func (handle->plugin, ks, k);
 }
 
@@ -338,6 +318,7 @@ void elektraInvokeClose (ElektraInvokeHandle * handle, Key * errorKey)
 }
 
 /**
+<<<<<<< HEAD
  * Invokes a deferable function on an invoke handle.
  * If the function is exported by the plugin it is directly invoked,
  * if the plugin supports deferring calls, the call is deferred.
--- a/src/libs/invoke/invoke.c
+++ b/src/libs/invoke/invoke.c
@@ -294,16 +294,14 @@ int elektraInvoke2Args (ElektraInvokeHandle * handle, const char * elektraPlugin
 {
 	if (!handle || !elektraPluginFunctionName) return -2;
 
-	// If we cast this right away it will hang if the function is not exported by the plugin
-	const void * uncastedFunc = elektraInvokeGetFunction (handle, elektraPluginFunctionName);
+	// If we cast this right away although the function wasn't found it will cause a deadlock
+	const void * rawFunc = elektraInvokeGetFunction (handle, elektraPluginFunctionName);
 
-	if (!uncastedFunc)
-	{
-		return -2;
-	}
+	if (!rawFunc) return -2;
 
 	typedef int (*elektra2Args) (Plugin *, KeySet *, Key *);
-	elektra2Args func = *(elektra2Args *)uncastedFunc;
+	elektra2Args func = *(elektra2Args *) rawFunc;
+
 	return func (handle->plugin, ks, k);
 }
 
--- a/src/libs/invoke/invoke.c
+++ b/src/libs/invoke/invoke.c
@@ -303,7 +303,7 @@ int elektraInvoke2Args (ElektraInvokeHandle * handle, const char * elektraPlugin
 	}
 
 	typedef int (*elektra2Args) (Plugin *, KeySet *, Key *);
-	elektra2Args func = *(elektra2Args *) uncastedFunc;
+	elektra2Args func = *(elektra2Args *)uncastedFunc;
 	return func (handle->plugin, ks, k);
 }
 
--- a/src/libs/invoke/invoke.c
+++ b/src/libs/invoke/invoke.c
@@ -20,6 +20,28 @@ struct _ElektraInvokeHandle
 };
 
 /**
+ * Structure for deferred calls
+ * @internal
+ */
+typedef struct _ElektraDeferredCall
+{
+	char * name;
+	KeySet * parameters;
+	struct _ElektraDeferredCall * next;
+} _ElektraDeferredCall;
+
+/**
+ * Structure for internal plugin state
+ * @internal
+ */
+struct _ElektraDeferredCallList
+{
+	_ElektraDeferredCall * head;
+	_ElektraDeferredCall * last;
+};
+
+
+/**
  * @defgroup invoke
  * @brief Functionality to use plugins and invoke functions
  *
@@ -197,7 +219,7 @@ KeySet * elektraInvokeGetPluginConfig (ElektraInvokeHandle * handle)
  *
  * @pre handle must be as returned from elektraInvokeOpen()
  *
- * @return the name of the plugin 
+ * @return the name of the plugin
  */
 const char * elektraInvokeGetPluginName (ElektraInvokeHandle * handle)
 {
@@ -318,7 +340,6 @@ void elektraInvokeClose (ElektraInvokeHandle * handle, Key * errorKey)
 }
 
 /**
-<<<<<<< HEAD
  * Invokes a deferable function on an invoke handle.
  * If the function is exported by the plugin it is directly invoked,
  * if the plugin supports deferring calls, the call is deferred.
--- a/src/libs/invoke/invoke.c
+++ b/src/libs/invoke/invoke.c
@@ -294,7 +294,7 @@ int elektraInvoke2Args (ElektraInvokeHandle * handle, const char * elektraPlugin
 {
 	if (!handle || !elektraPluginFunctionName) return -2;
 
-	// If we cast this right away it will hang if the function is not
+	// If we cast this right away it will hang if the function is not exported by the plugin
 	const void * uncastedFunc = elektraInvokeGetFunction (handle, elektraPluginFunctionName);
 
 	if (!uncastedFunc)
--- a/src/plugins/process/CMakeLists.txt
+++ b/src/plugins/process/CMakeLists.txt
@@ -17,7 +17,6 @@ if (PLUGINPROCESS_FOUND)
 		    TEST_REQUIRED_PLUGINS "dump"
 		    LINK_ELEKTRA elektra-pluginprocess
 				 elektra-invoke)
-
 else (PLUGINPROCESS_FOUND)
-	message ("${PLUGINPROCESS_NOTFOUND_INFO}, excluding pluginprocess library, thus excluding the process plugin")
+	exclude_plugin (process "${PLUGINPROCESS_NOTFOUND_INFO}, pluginprocess library excluded, thus also excluding the process plugin")
 endif (PLUGINPROCESS_FOUND)
--- a/src/plugins/process/CMakeLists.txt
+++ b/src/plugins/process/CMakeLists.txt
@@ -5,6 +5,9 @@ add_plugin (process
 		process.h
 		process.c
 	ADD_TEST
+	TEST_README
+	TEST_REQUIRED_PLUGINS
+		"dump"
 	LINK_ELEKTRA
 		elektra-pluginprocess
 		elektra-invoke
--- a/src/plugins/process/CMakeLists.txt
+++ b/src/plugins/process/CMakeLists.txt
@@ -9,18 +9,15 @@ if (NOT_INCLUDED)
 endif (NOT_INCLUDED)
 
 if (PLUGINPROCESS_FOUND)
-add_plugin (process
-	SOURCES
-		process.h
-		process.c
-	ADD_TEST
-	TEST_README
-	TEST_REQUIRED_PLUGINS
-		"dump"
-	LINK_ELEKTRA
-		elektra-pluginprocess
-		elektra-invoke
-	)
+	add_plugin (process
+		    SOURCES process.h
+			    process.c
+		    ADD_TEST
+		    TEST_README
+		    TEST_REQUIRED_PLUGINS "dump"
+		    LINK_ELEKTRA elektra-pluginprocess
+				 elektra-invoke)
+
 else (PLUGINPROCESS_FOUND)
 	message ("${PLUGINPROCESS_NOTFOUND_INFO}, excluding pluginprocess library, thus excluding the process plugin")
 endif (PLUGINPROCESS_FOUND)
--- a/src/plugins/process/CMakeLists.txt
+++ b/src/plugins/process/CMakeLists.txt
@@ -1,5 +1,8 @@
 include (LibAddMacros)
 
+find_package (Pluginprocess)
+
+if (PLUGINPROCESS_FOUND)
 add_plugin (process
 	SOURCES
 		process.h
@@ -12,3 +15,6 @@ add_plugin (process
 		elektra-pluginprocess
 		elektra-invoke
 	)
+else (PLUGINPROCESS_FOUND)
+	message ("${PLUGINPROCESS_NOTFOUND_INFO}, excluding pluginprocess library, thus excluding the process plugin")
+endif(PLUGINPROCESS_FOUND)
--- a/src/plugins/process/CMakeLists.txt
+++ b/src/plugins/process/CMakeLists.txt
@@ -1,7 +1,9 @@
 include (LibAddMacros)
 
 find_package (Pluginprocess)
+plugin_check_if_included ("dump")
 
+if (NOT NOT_INCLUDED)
 if (PLUGINPROCESS_FOUND)
 add_plugin (process
 	SOURCES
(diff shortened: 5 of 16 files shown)


e1528532:
jenkins build all please

e1528532:
jenkins build all please

e1528532:
interesting, output_warnings crashed in case there were multiple warnings returned in a test. However this fix was so tiny and this was apparently the first plugin that made use of that in a test, i just did it here as i think after that this PR is fine too.

e1528532:
Now that invoke has error messages i see the reason why this one test fails, apparently it fails to resolve some asan symbol correctly. Maybe travis doesn't set up the symbolizer paths appropriately as described in the documentation. Lets see if this works on our own build server.

e1528532:
jenkins build clang-asan please

e1528532:
Unfortunately, the issue [still seems to persist](https://travis-ci.org/ElektraInitiative/libelektra/jobs/338590555):

```
reason: of module: libelektra-dump.so, because: /home/travis/build/ElektraInitiative/build/lib/libelektra-dump.so: undefined symbol: __ubsan_vptr_type_cache
```

I cannot reproduce this with my local debian though. The only idea that i'd still have is that this LD_PRELOAD thing is required as well for that one build, though it'd be strange as all the other tests using asan work fine as well. But the error message looks like the asan library is missing to me..

>LD_PRELOAD=/usr/lib/clang/3.8.0/lib/linux/libclang_rt.asan-x86_64.so run_asan
>or on Debian:
>LD_PRELOAD=/usr/lib/llvm-3.8/lib/clang/3.8.1/lib/linux/libclang_rt.asan-x86_64.so run_asan

However the travis job seems to use version 5 of clang. Is there any way i can find out the path of this library on that one travis machine somehow, without blindly trying around until it fits @sanssecours (you seem to know a lot about travis, so maybe you know this too) @markus2330 ?

markus2330:
You could run a `find / -name '*libclang_rt.asan-x86_64*'` or similar on the travis server.

Wouldn't it be the better approach to search for this library using CMake? CMake searches in various standard folders by default, so we have a much better chance that it works on new OSs.

e1528532:
> Wouldn't it be the better approach to search for this library using CMake

On the long run it most likely is, but that'd be worth an own PR i guess and to be honest i want to focus more on my type checker to move forward there, so i'd be happy with an LD_PRELOAD approach for now ;)

.travis.yml
```diff
@@ -173,12 +173,13 @@ before_script:
...
 
 script:
+  - find /local/mnt/workspace/ -name '*libclang_rt.asan-x86_64.so'
   - ninja
   - |
     if [[ "$FAST" == "ON" ]]; then
       ninja run_all
     elif [[ "$ASAN" = "ON" && "$TRAVIS_OS_NAME" == "linux" ]]; then
-      LD_PRELOAD=$(find /usr/ -name '*libclang_rt.asan-x86_64.so') ASAN_OPTIONS=symbolize=1 ASAN_SYMBOLIZER_PATH=$(which llvm-symbolizer) ninja run_all
+      LD_PRELOAD=$(find /local/mnt/workspace/tmp/ -name '*libclang_rt.asan-x86_64.so') ASAN_OPTIONS=symbolize=1 ASAN_SYMBOLIZER_PATH=$(which llvm-symbolizer) ninja run_all
```
markus2330: Simply use `find /` or does it take so long? @sanssecours what do you think?

e1528532: the thing is the asan library in there doesn't work it says linked to incompatible asan runtime there and points me to some strange path /local/mnt/workspace/tmp which i am now hackily investigating what it contains for us ;)

e1528532: ok apparently that path doesn't even exist and asan doesn't like the runtime in /usr/ although there are only 5.0.0 runtimes and i have no idea where else it magically picks it up from. lets just search / then.

e1528532: The search yielded nothing new, there seems to be only /usr/local/clang-5.0.0/lib/clang/5.0.0/lib/linux/libclang_rt.asan-x86_64.so but when i LD_PRELOAD on that it complains its  not the asan runtime linked against...

```
clang version 5.0.0 (tags/RELEASE_500/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/clang-5.0.0/bin
```

this would also indicate that the runtime in local/clang/... should be correct so i'm not sure how to continue. I'm tempted to just exclude the process plugin in that constellation as we have our own asan build anyway. Its not executed yet though so lets see if that works as a replacement.

markus2330: Yes, it is difficult to find the correct one afterwards. I still suggest to do it via CMake, e.g., did you look at https://github.com/arsenm/sanitizers-cmake

e1528532: while this seems like a fine idea, in my experience changing such build-related things usually turn out to be a lot of work even if it only looks like changing a few simple line. So I'd prefer to follow another approach for this PR for now (like blacklisting it for this one travis asan build in case it works on our own asan build). We can open an issue about using such a cmake approach for the sanitizer for the future.

e1528532:
@markus2330 https://build.libelektra.org/jenkins/job/elektra-clang-asan/393/ looks like our own asan build works fine and only this one linux config on travis is affected. If its ok, I'd disable the travis asan test for the process plugin for now until we've fixed the asan configuration via cmake or something like that.

sanssecours:
Looks like we need to preload [`libclang_rt.ubsan_standalone_cxx-x86_64.so`](https://travis-ci.org/sanssecours/elektra/builds/338933829) instead of `libclang_rt.asan-x86_64.so`. I updated the PR accordingly in my copy of the repository.

@e1528532 🙏 Please wait until I update this PR.

sanssecours:
I just pushed an update containing my changes. I also dropped and merged some of the latest commits using `git rebase --interactive`. If you think the updates are too intrusive Armin, then please feel free to just overwrite them.

> @sanssecours what do you think?

Preloading a library via an environment variable does not sound like a good idea. Hopefully we find a better solution for this problem in the future.

e1528532:
> I just pushed an update containing my changes. I also dropped and merged some of the latest commits using git rebase --interactive

very nice! i wasn't even aware that you can get rid of such "try&error" style commits after you've pushed them easily. its much cleaner now without the unnecessary commits, so thanks for that.

> Hopefully we find a better solution for this problem in the future.

yes we should evaluate the cmake asan integration - maybe that would take care about such issues.

However, lets see now if the tests will pass this time. It'd be great to see this merged!

e1528532:
jenkins build all please

e1528532:
thanks to @sanssecours it finally looks good now. @markus2330 i think we can merge this first version now., its still experimental anyway i haven't tried too much with it yet apart from the simple test.

src/libs/invoke/invoke.c
```diff
@@ -264,14 +264,16 @@ int elektraInvoke2Args (ElektraInvokeHandle * handle, const char * elektraPlugin
 {
 	if (!handle || !elektraPluginFunctionName) return -2;
 
-	typedef int (*elektra2Args) (Plugin *, KeySet *, Key *);
-	elektra2Args func = *(elektra2Args *)elektraInvokeGetFunction (handle, elektraPluginFunctionName);
+	// If we cast this right away it will hang if the function is not
```
markus2330: sentence?

src/plugins/process/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+include (LibAddMacros)
+
+find_package (Pluginprocess)
+
+if (PLUGINPROCESS_FOUND)
```
markus2330: Isn't dump needed, too?

src/plugins/process/README.md
```diff
@@ -0,0 +1,53 @@
+- infos = Information about the process plugin is in keys below
+- infos/author = e1528532 <e1528532@libelektra.org>
```
markus2330: Your name please.

src/plugins/process/README.md
```diff
@@ -0,0 +1,53 @@
...
+- infos/licence = BSD
+- infos/needs = dump
```
markus2330: needs tells that dump should be loaded as separate plugin, this is not what we want

src/plugins/process/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Dependencies
+
+The `pluginprocess` library, which is only available on POSIX environments currently.
+
+## Examples
+
+```sh
+# Mount the dump plugin a separate process via the process plugin to the cascading namespace `/examples/process`
+sudo kdb mount config.file /examples/process process plugin=dump
```
markus2330: Does not work for me, I get the error:
```
The command bin/kdb mount terminated unsuccessfully with the info:
Was not able to load such a plugin!

Maybe you misspelled it, there is no such plugin or the loader has problems.
You might want to try to set LD_LIBRARY_PATH, use kdb-full or kdb-static.
Errors/Warnings during loading were:

```

Even "kdb check process" fails.


Please also add this shell recorder test in tests/shell/shell_recorder/tutorial_wrapper/CMakeLists.txt then hopefully the build server will tell you the same problem I have locally.

src/plugins/process/process.c
```diff
@@ -0,0 +1,243 @@
...
+}
+
+int elektraInvoke1Arg (ElektraInvokeHandle * handle, const char * elektraPluginFunctionName, Key * k)
+{
+	if (!handle || !elektraPluginFunctionName) return -2;
+
+	// If we cast this right away although the function wasn't found it will cause a deadlock
+	const void * rawFunc = elektraInvokeGetFunction (handle, elektraPluginFunctionName);
+
+	if (!rawFunc) return -2;
```
markus2330: No error set?

src/plugins/process/process.c
```diff
@@ -0,0 +1,243 @@
...
+	return ELEKTRA_PLUGIN_STATUS_NO_UPDATE;
+}
+
+Plugin * ELEKTRA_PLUGIN_EXPORT (process)
+{
+	// clang-format off
+	return elektraPluginExport ("process",
+		ELEKTRA_PLUGIN_OPEN,	&elektraProcessOpen,
+		ELEKTRA_PLUGIN_CLOSE,	&elektraProcessClose,
+		ELEKTRA_PLUGIN_GET,		&elektraProcessGet,
```
markus2330: formatting?

src/plugins/process/process.c
```diff
@@ -0,0 +1,243 @@
...
+	if (pluginNameKey == NULL)
+	{
+		ELEKTRA_SET_ERROR (192, errorKey, "Missing plugin configuration parameter plugin=<name of plugin to be proxied>");
+		return ELEKTRA_PLUGIN_STATUS_ERROR;
+	}
+
+	// and this key should obviously contain the plugin's name, so check for any name
+	// furthermore by invoking process in a process we'd create a deadloop, check that too
+	const char * pluginName = keyString (pluginNameKey);
+	if (elektraStrCmp (pluginName, "(null)") == 0 || elektraStrCmp (pluginName, "(binary)") == 0)
```
markus2330: Isn't keyIsBinary the better check?

src/plugins/process/process.c
e1528532: no it just crashed right away and when i separated the cast from getting the pointer it worked - so that was more fixed by trial and error ;)

src/plugins/process/CMakeLists.txt
e1528532: yes thats what i thought infos/needs takes care about, will correct it

src/plugins/process/README.md
e1528532: it is expected that kdb check process fails, as without a plugin configuration it does not work because it doesn't know what plugin shall be proxied.

e1528532: @markus2330 regarding the mounting issue, thats strange, i am pretty sure this worked at some point but i can reproduce it also. Again, logging reveals the actual cause (I've added a simple logging statement for debugging just after it gets the plugin configuration like
```
KeySet * processConfig = elektraPluginGetConfig (handle);
Key * pluginName = ksLookupByName (processConfig, "/plugin", KDB_O_NONE);
ELEKTRA_LOG ("we are trying to open the plugin %s", keyString (pluginName));
```
)

but it looks like the configuration parameter is not present? Other plugins using config seem to handle it the same way though, so i'm not exactly sure yet where the error hides.

```
sudo kdb mount test.dump user/examples/process process plugin=dump
...
src/plugins/process/process.c:62:elektraProcessOpen: we are trying to open the plugin (null)
```

markus2330: You should definitely always add warnings in such situations. It should never be required to enable logging to debug wrong plugin parameters.

> which would have complained about the missing name?

Where do you implement this check? Do you also pass the checkConfig calls to the proxied plugins?

markus2330:
#1817 was merged, please rebase and make this ready to merge.

src/plugins/process/README.md
e1528532: > You should definitely always add warnings in such situations.

done

> Where do you implement this check?

Currently it was only in the checkConfig function but now i've also included the same check in the startup function so users get the warning in any case. the little overhead should be acceptable for that.

I don't pass checkConfig calls to proxied plugins though, as there is no Plugin handle in the checkConfig function signature so i can't really access the child process in there.

e1528532:
I've fixed the warning and noticed that if the key you use for the proxied plugins has an invalid name, it won't be appended to the keyset to be transferred to the child process, hence resulting in strange results and missing error messages. Fixed this issue now in general, so we have warning messages always.

However, the issue with the plugin configuration remains. Somehow it doesn't pass the configuration parameter to the plugin. I tried to print the whole configuration upon elektraProcessOpen by adding these few debug statements:

```
		KeySet * processConfig = elektraPluginGetConfig (handle);
		ksRewind (processConfig);
		Key * k;
		while ((k = ksNext (processConfig)) != NULL)
			ELEKTRA_LOG ("config got key %s %s", keyName (k), keyString (k));
```

However, when i mount a random file then using a proxied dump plugin like `sudo kdb mount test.dump user/examples/process process process=dump` the iteration doesn't print the /plugin key i'd expect to see what plugin should be proxied. The only message i get is about some `system/module` key in the plugin configuration: `src/plugins/process/process.c:64:elektraProcessOpen: config got key system/module this plugin was loaded for the status`
@markus2330 any idea why the configuration is not passed upon mounting?

markus2330:
> any idea why the configuration is not passed upon mounting?

No, I am afraid you have to debug where the info is lost. Maybe it is not correctly passed to the other process and never reaches the proxied plugin? Please also add tests for this.

markus2330:
Any progress here? I have yet another use case where this plugin would be handy. (For python)

e1528532:
i see that you have already started to use the library directly for the python plugin. IMO for interpreters that always have to be executed inside an own process this is the better way than using this plugin as a proxy, because if it doesn't work without being executed in its own process it is of little use on its own.
I don't plan to continue on that before the typechecker and thesis is done due to the deadline

markus2330:
Yes, I agree. Patching ruby and java to directly use the library also makes sense.

For other plugins, like xerces, however, it would be good to be able to choose if fork is used. So the plugin is still relevant. But yes, you need to concentrate on the type checker.

e1528532:
rebased everything on the newest pluginprocess improvements, lets see if it helps

markus2330:
Looks good. There only seems to be some issue if logging+debugging is turned on.

@ingwinlu Wouldn't it make sense to rename the -optimization to -log-debug? It is much more likely to trigger a bug in log+debug than to trigger a bug in the non-optimization case.

e1528532:
Are they related to this plugin? Judging from the failed tests i don't see a direct connection to it, apart from check_kdb_internal_check:

```
	112 - testshell_markdown_ini (Failed)
	174 - testscr_check_kdb_internal_check (Failed)
	193 - testshell_replay_ls (Failed)
	199 - testshell_markdown_tutorial_cascading (Failed)
	200 - testshell_markdown_tutorial_validation (Failed)
	201 - testshell_markdown_msr_syntax (Failed)
	202 - testshell_markdown_readme_msr (Failed)
	203 - testshell_markdown_issue_template (Failed)
```

e1528532:
Wait i see i accidentially committed log.c, maybe that causes it

e1528532:
Finally found out that this plugin seems to be impossible with the current backend behavior.
Backend fetches the plugin's contract in `backendOpenModules` in `backend.c`, however it only provides a default config there for the plugin.

But this plugin needs to have its config available already, as it needs to grab the `/plugin` config in order to deliver the correct contract. The process plugin has to mimic the contract of the plugin it proxies, and for that i wanted to use `invoke` to call the get method of the proxied plugin with the contract key.

I've bypassed this for now by assuming there won't be a config available when grabbing the contract and returning a generic default, however this is most likely not what shall be done as this makes the information in `system/elektra/modules/process` wrong. The issue is that this information can't really be determined statically, as it totally depends on the plugin that is being proxied. Any idea how to work around this without tons of work? @markus2330

e1528532:
In case there is no easy workaround i'd suggest merging this anyway after it passes the build in its current state, i've added this fallacy to the README. As the plugin is experimental i think its good to have it in master, maybe someone will continue on it one day and for simple things like proxying dump it works (to show the experiment)

src/plugins/process/process.c
markus2330: still not resolved

src/plugins/process/process.c
```diff
@@ -0,0 +1,286 @@
...
+			       keyNew ("system/elektra/modules/process/exports/get", KEY_FUNC, elektraProcessGet, KEY_END),
+			       keyNew ("system/elektra/modules/process/exports/set", KEY_FUNC, elektraProcessSet, KEY_END),
+			       keyNew ("system/elektra/modules/process/exports/error", KEY_FUNC, elektraProcessError, KEY_END),
+			       keyNew ("system/elektra/modules/process/exports/checkconf", KEY_FUNC, elektraProcessCheckConfig, KEY_END),
+#include ELEKTRA_README (process)
+			       keyNew ("system/elektra/modules/process/infos/version", KEY_VALUE, PLUGINVERSION, KEY_END), KS_END);
+		ksAppend (returned, contract);
+		ksDel (contract);
+
+		if (!validPluginName (pluginName, parentKey)) return ELEKTRA_PLUGIN_STATUS_SUCCESS;
```
markus2330: Please comment what is done here?

Shouldn't the contract be renamed to pluginName?

src/plugins/process/process.c
```diff
@@ -0,0 +1,286 @@
...
+	Key * cur;
+	while ((cur = ksNext (pluginContract)) != NULL)
+	{
+		Key * cpy = keyDup (cur);
+		keySetBaseName (cpy, NULL);
+		if (!elektraStrCmp ("infos", keyBaseName (cpy)))
+		{
+			keySetBaseName (cpy, NULL);
+			keySetBaseName (cpy, NULL);
+			keyAddBaseName (cpy, "process");
```
markus2330: Why to process and not pluginName?

src/plugins/process/process.c
markus2330: still not resolved

src/plugins/process/README.md
```diff
@@ -0,0 +1,52 @@
...
+#> value
+
+# Undo modifications
+kdb rm -r /tests/process
+sudo kdb umount /tests/process
+```
+
+## Limitations
+
+This plugin cannot act as a proxy for itself to prevent loops. Currently only plugins that
```
markus2330: Make a list and turn around. The second limitation is much more important.

src/plugins/process/README.md
```diff
@@ -0,0 +1,52 @@
...
+- infos/status = maintained unittest shelltest tested configurable experimental
+- infos/metadata = 
+- infos/description = executes other plugins inside an own process
+
+## Introduction
+
+The process plugin uses the `pluginprocess` library in order to execute other plugins
+inside an own process.
+This is useful for plugins which cause memory leaks to be isolated in an own process.
+Furthermore this is useful for runtimes or libraries that cannot be reinitialized in
```
markus2330: Second reason more important (swap)

src/plugins/process/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+		    SOURCES process.h
+			    process.c
+		    ADD_TEST
+		    TEST_README
+		    TEST_REQUIRED_PLUGINS "dump"
+		    LINK_ELEKTRA elektra-pluginprocess
+				 elektra-invoke)
+
+else (PLUGINPROCESS_FOUND)
+	message ("${PLUGINPROCESS_NOTFOUND_INFO}, excluding pluginprocess library, thus excluding the process plugin")
```
markus2330: pluginprocess library excluded, thus...

src/plugins/process/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+find_package (Pluginprocess)
+plugin_check_if_included ("dump")
+
+if (NOT_INCLUDED)
+	message ("dump plugin not found (${NOT_INCLUDED}), excluding the process plugin")
+	return ()
+endif (NOT_INCLUDED)
+
+if (PLUGINPROCESS_FOUND)
```
markus2330: wrongly used, please properly exclude plugin

src/error/specification
```diff
@@ -1230,3 +1230,17 @@ severity:warning
 ingroup:plugin
 module:zeromqsend
 macro:ZEROMQSEND_ERROR
+
+number: 197
+description:Failed to initialize the execution of a plugin in a child process
```
markus2330: reuse errors from pluginprocess?

cmake/Modules/LibAddPlugin.cmake
```diff
@@ -179,6 +179,7 @@ function (add_plugintest testname)
 endfunction (add_plugintest)
 
 function (plugin_check_if_included PLUGIN_SHORT_NAME)
+	set (NOT_INCLUDED "" PARENT_SCOPE)
```
markus2330: What is done here?

src/plugins/process/README.md
```diff
@@ -0,0 +1,52 @@
+- infos = Information about the process plugin is in keys below
+- infos/author = Armin Wurzinger <e1528532@libelektra.org>
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/recommends =
+- infos/placements = pregetstorage getstorage postgetstorage presetstorage setstorage
+- infos/status = maintained unittest shelltest tested configurable experimental
```
markus2330: status unfinished missing. Is it really tested?

src/plugins/process/process.c
e1528532: the contract returned from this plugin seems to be expected to be called process, i don't recall the error but i think it was something like contract missing.

src/plugins/process/process.c
e1528532: as said below, it seems to be expected to be called process

src/plugins/process/process.c
markus2330: Ok. Did you try to use `process#1 plugin=dump process#2 plugin=other`? Iirc this should allow different contracts for plugins with the same name.

e1528532: unfortunately this still doesn't resolve the issue of not knowing which plugin shall be proxied upon getting the contract, so it is only of little use currently.

markus2330: Ok, please resolve the other small things so that we can get this into master.

markus2330:
@e1528532  would be great to have this for the upcoming release which hopefully happens today.

Please also improve the release notes about the type checker plugin

markus2330:
Thank you, the sooner the better :+1:

Actually only the update of the release notes is necessary. If the processplugin is within this release or the next does not really matter.

src/plugins/process/process.c
e1528532: funny, on my system it looks normal this way, but ok deleted one tab ;)

src/error/specification
e1528532: Yes i think that error can be reused to avoid introducing more as this plugin and pluginprocess are quite closely related

src/plugins/process/process.c
markus2330: Hopefully GitHub did not change their tab setting (or is different than ours)... I trusted the GitHub view and did not check locally. Our settings are in .clang-format: TabWidth is 8

src/plugins/process/README.md
markus2330: Yes, thank you. Better to understate.

src/plugins/process/CMakeLists.txt
e1528532: done

src/plugins/process/CMakeLists.txt
e1528532: done

src/plugins/process/README.md
e1528532: done

src/plugins/process/README.md
e1528532: done

markus2330:
The build fails now, let us merge this after the release.

e1528532:
accidentially wrote exclude_plugin instead of remove_plugin so that one test fails. debian-unstable-clang failed due to zeromq,so i guess thats not my fault. I've corrected the exclude_plugin but since all the other tests ran successfully i don't think its necessary to wait for them to complete.

e1528532:
adding ready to merge, judge for yourself if you trust my explanation ;)

markus2330:
Thank you! Sorry it was too late, I already started the release when I saw your "ready to merge".

Please rebase for the release notes.

e1528532:

markus2330:
Thank you!

Issue #1782, opened 2018-01-21
setup.hs compilation error


markus2330:
Seems like I merged the PRs too early.

On Debian Stretch the compilation failed:
```
dist/setup/setup.hs:141:11: error: parse error on input ‘->’
src/bindings/haskell/CMakeFiles/c2hs_haskell.dir/build.make:71: for „src/bindings/haskell/libHSlibelektra-haskell-ghc8.0.1.so“ failed
make[2]: *** [src/bindings/haskell/libHSlibelektra-haskell-ghc8.0.1.so] Error 1
```

And there is also a new warning:
```
CMakeFiles/elektra-pluginprocess.dir/pluginprocess.c.o: In function `getPipename':
/home/markus/Projekte/Elektra/current/src/libs/pluginprocess/pluginprocess.c:272: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
[ 82%] Built target elektra-pluginprocess
```

e1528532:
could you please state your ghc and cabal versions? Most likely that is some version difference. We should probably also introduce a haskell test to be executed on the debian platforms, currently its only really tested on mac like i use for developing. I think i should set up a debian vm additionally.

The latter issue can be avoided by creating a temporary directory instead and placing the pipe files into there. i'll take care about that.

markus2330:
Thank you for the quick response. I use the versions from Debian Stretch.

```
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.0.1

$ cabal --version
cabal-install version 1.24.0.1
compiled using version 1.24.1.0 of the Cabal library
```

> I think i should set up a debian vm additionally.

This should be easily done with the docker builds. But I am also okay to install ghc and cabal on every agent (or at least on these where we plan to support it. Maybe Jessie is already too old. We should support stretch, though.).

e1528532:
that should be covered by #1787 and checked via the new haskell build job on v2. closing.

markus2330:
very good that so many bugs are being closed right now :-)

PR #1785, opened 2018-01-21
pluginprocess-mktemp-warning-fix

e1528532:
# Purpose

Uses mkdtemp and puts the pipe inside this directory to avoid the mktemp warning (possible race condition between file name generation and pipe creation, maybe something else picks the name in the meanwhile, highly unlikely but theoretically possible). Fixes the issue mentioned in #1782 .

# Checklist

@markus2330 tested by using this in conjunction with the process PR, i doubt any separate testing makes sense.

### DIFF
--- a/src/libs/pluginprocess/pluginprocess.c
+++ b/src/libs/pluginprocess/pluginprocess.c
@@ -25,6 +25,7 @@
 
 struct _ElektraPluginProcess
 {
+	char * pipeDirectory;
 	char * commandPipe;
 	char * resultPipe;
 	int pid;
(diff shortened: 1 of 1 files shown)


e1528532:
jenkins build all please

e1528532:
jenkins build incremental please

e1528532:

e1528532:
looks good. the jenkins docker build is actually done if you click the link and check the status, git doesn't seem to update the status. and the incremental test failure is strange and i guess unrelated to this PR. so imo its ready to merge unless you have any further objections.

markus2330:
Thank you for the fix!

How does cleaning up the directory work? (unlinking it and the pipe)

e1528532:
`elektraPluginProcessClose` which gets called in a plugin's close method calls the `cleanup` function (both in `pluginprocess.c`), which unlinks the pipes and then removes the directory. Additionally, if there is an initialization failure, it also cleans up any created pipes/directories. Ultimately if something goes very wrong then it stays, but as it gets created in `/tmp` distributions will usually clean it up themselves.

markus2330:
Thank you, great!

If you can easily fix some cases where the cleanup does not take place, please go ahead.

Issue #1786, opened 2018-01-22
security: vulnerable dependency


markus2330:
Hello Marvin,

Can you easily update the version number for jquery?

src/tools/rest-frontend/package.json

Known  moderate severity security vulnerability detected in jquery < 3.0.0 defined in package.json. 

best regards,

markus2330:
Thank you for the fast reply.

The warning was generated from GitHub, I gave you also the access to these security messages (should be in "Inights"). I also forwarded you the email I got from GitHub.

Namoshek:
Nope, not seeing the vulnerabilities, only the dependency graph. However, I have a concrete clue about updating jQuery now. I will have a look at it but I cannot promise anything as I cannot check the stuff easily as I still couldn't get back my dev VM.

Namoshek:
Couldn't get anything working yesterday, so I figured it might be worth trying to just increment the version on it's own. If it doesn't work, we can just go ahead and undo the changes.

PR #1788, opened 2018-01-23
increment jquery version in package.json

Namoshek:
Due to a wont-fix vulnerability in jquery v2.x.x, it is necessary to update to jquery v3.x.x. For more details see #1786.

### DIFF
--- a/src/tools/rest-frontend/package.json
+++ b/src/tools/rest-frontend/package.json
@@ -32,7 +32,7 @@
         "angular-typewriter": "0.0.15",
         "angular-ui-bootstrap": "^2.2.0",
         "angular-ui-notification": "^0.3.6",
-        "angular-ui-router": "^0.3.1",
+        "angular-ui-router": "^0.4.2",
         "bootstrap": "^3.3.7",
         "connect-modrewrite": "^0.9.0",
         "fs-extra": "^0.30.0",
--- a/src/tools/rest-frontend/package.json
+++ b/src/tools/rest-frontend/package.json
@@ -50,7 +50,7 @@
         "grunt-contrib-watch": "latest",
         "grunt-preprocess": "latest",
         "highlight.js": "^9.7.0",
-        "jquery": "^2.2.4",
+        "jquery": "^3.3.1",
         "jshint-stylish": "latest",
         "ng-tags-input": "^3.1.1",
         "pace-progress": "^1.0.2",


Namoshek:
jenkins build homepage please

markus2330:
New website is online and seems to work. So any objections to merge it?

Namoshek:
There is a minor issue with the `ui-router` package, which wasn't there before the version upgrade afaik. It seems to still work though and as if the error was purely cosmetic (at least for the moment).

I noticed that the backend has a broken database, though: https://restapi.libelektra.org/database/
The snippet converter does still work, so I think it is no general issue with the backend but a maybe a corrupted database?

markus2330:
Thank you for looking into it!

> There is a minor issue with the ui-router package, which wasn't there before the version upgrade afaik.

Is it possible to update ui-router, too?

> as if the error was purely cosmetic (at least for the moment).

Where does the error occur?

> I noticed that the backend has a broken database, though: https://restapi.libelektra.org/database/

Does map::at mean that an exception was thrown when calling map::at? Is there any easy way to get a stack trace? (Might not be ideal for security, though)

> The snippet converter does still work, so I think it is no general issue with the backend but a maybe a corrupted database?

The last commit in the database was Tue Apr 4  2017 but maybe changes in the xerces (or some other) plugin cause the problem? (import of previously working plugins might fail?)

Namoshek:
jenkins build homepage please

Namoshek:
> Is it possible to update ui-router, too?

Sure, let's see what happens. I had a look at it's new dependencies and didn't find any issue. I only went to v0.4.x though and not to v1.x.x as this really introduced breaking changes.

> Where does the error occur?

Browser console... that's why I said purely cosmetic. Not even visible for most people.

> Does map::at mean that an exception was thrown when calling map::at? Is there any easy way to get a stack trace? (Might not be ideal for security, though)

I have actually no idea at all. It's also over one year ago I last used C++, so I'm kind of out for this question...

> The last commit in the database was Tue Apr 4 2017 but maybe changes in the xerces (or some other) plugin cause the problem? (import of previously working plugins might fail?)

Was this before or after the server got compromised and had to be renewed?

markus2330:
> Sure, let's see what happens.

The website looks good. Any objections to merge it now?
     

> Was this before or after the server got compromised and had to be renewed?

Shortly before, our server was unblocked at 30.5.2017.

Namoshek:
> The website looks good. Any objections to merge it now?

No, it looks fine, go ahead!

> Shortly before, our server was unblocked at 30.5.2017.

Is it possible that something got corrupted during this process?

markus2330:
Thank you for your work!

>   Is it possible that something got corrupted during this process?

Yes, certainly it is possible. Or the setup is simply done wrongly and the database is not even found by the restbackend?

I created #1791 for the database error. Please continue discussions there.

PR #1796, opened 2018-01-28
Shell Recorder: Fix Syntax Error

sanssecours:
# Purpose

Fix a syntax error in the Shell Recorder script

# Checklist

- [x] I checked the commit message. 
- [x] I ran all tests and everything went fine.

PR #1730, opened 2017-12-12
Opmphm Integration

KurtMi:
# Purpose

Integrate the Order Preserving Minimal Perfect Hash Map

# Checklist

- [x] commit messages are fine (with references to issues)
- [x] I added unit tests
- [x] I ran all tests and everything went fine
- [x] affected documentation is fixed
- [x] I added code comments, logging, and assertions (see doc/CODING.md)
- [x] release notes are updated (doc/news/_preparation_next_release.md)

Issue #1260, opened 2017-01-08
Semlock: Unit Tests Fail on Debian with Build Type RelWithDebInfo


markus2330:
To be fixed together after working global plugins (i.e. when semlock can be tested)

https://build.libelektra.org/job/elektra-gcc-configure-debian-jessie-minimal/lastFailedBuild/console

```
        Start  66: testmod_semlock
 66/138 Test  #66: testmod_semlock ....................***Failed    0.12 sec
SEMLOCK     TESTS
==================

test Open & Close
There are 2 warnings
buffer is: warnings/#00
number: 11
description: open of plugin returned unsuccessfully (reason contains plugin, see other warnings for details)
ingroup: kdb
module: 
file: /home/jenkins/workspace/workspace/elektra-gcc-configure-debian-jessie-minimal/src/libs/elektra/plugin.c
line: 297
reason: semlock
reason: 
reason: 
buffer is: warnings/#01/configfile
number: (null)
description: kdbClose() failed
ingroup: kdb
module: 
file: /home/jenkins/workspace/workspace/elektra-gcc-configure-debian-jessie-minimal/src/libs/elektra/plugin.c
line: 328
reason: kdbClose() failed
reason: 
reason: 
/home/jenkins/workspace/workspace/elektra-gcc-configure-debian-jessie-minimal/src/plugins/semlock/testmod_semlock.c:23: error in test_OpenClose: warnings in kdbOpen for plugin semlock
number: 145
description: : semlock plugin error
ingroup: : plugin
module: : semlock
at: /home/jenkins/workspace/workspace/elektra-gcc-configure-debian-jessie-minimal/src/plugins/semlock/semlock.c:58
reason: : Open semaphore: ENOSYS. /dev/shm should be mounted as tempfs. Please look in the README using "kdb info semlock"!

mountpoint: : 
configfile: : 
/home/jenkins/workspace/workspace/elektra-gcc-configure-debian-jessie-minimal/src/plugins/semlock/testmod_semlock.c:23: error in test_OpenClose: error in kdbOpen for plugin semlock
/home/jenkins/workspace/workspace/elektra-gcc-configure-debian-jessie-minimal/src/plugins/semlock/testmod_semlock.c:23: fatal in test_OpenClose: could not open semlock plugin

        Start  67: testmod_shell
...
        Start  80: testscr_check_kdb_internal_check
 80/138 Test  #80: testscr_check_kdb_internal_check ...***Failed    0.55 sec

[...]

PR #1798, opened 2018-01-28
Tests: Shell: kdb_internal_check exclude semlock

KurtMi:
# Purpose

`check_kdb_internal_check.sh` checks all added plugins and does not care about excluded testcases,
therefore `semlock` could fail and should be removed till fixed properly.

# Checklist

- [ ] I added unit tests

### DIFF
--- a/tests/shell/check_kdb_internal_check.sh
+++ b/tests/shell/check_kdb_internal_check.sh
@@ -28,7 +28,7 @@ do
 		continue
 		;;
 	"semlock")
-		# exclude due to issue #1781
+		# exclude due to issue 1781
 		continue
 		;;
 	esac
--- a/tests/shell/check_kdb_internal_check.sh
+++ b/tests/shell/check_kdb_internal_check.sh
@@ -27,6 +27,10 @@ do
 		# output on open/close
 		continue
 		;;
+	"semlock")
+		# output on open/close
+		continue
+		;;
 	esac
 
 	> $FILE
--- a/tests/shell/check_kdb_internal_check.sh
+++ b/tests/shell/check_kdb_internal_check.sh
@@ -28,7 +28,7 @@ do
 		continue
 		;;
 	"semlock")
-		# output on open/close
+		# exclude due to issue #1781
 		continue
 		;;
 	esac


KurtMi:

markus2330:
Thank you, good solution! But is the comment correct? Is the problem an output or something else?

KurtMi:
> Thank you, good solution! But is the comment correct? Is the problem an output or something else?

The comment is a copy paste comment and not correct, I will update it!

KurtMi:
@markus2330 why is the `RelWithDebInfo` in https://build.libelektra.org/jenkins/job/elektra-multiconfig-gcc-stable/203/ skipped?

markus2330:
Thank you for the fix!

> why is the RelWithDebInfo in https://build.libelektra.org/jenkins/job/elektra-multiconfig-gcc-stable/203/ skipped?

I don't know. I added the problem with RelWithDebInfo in #160.

sanssecours:
> why is the `RelWithDebInfo` in https://build.libelektra.org/jenkins/job/elektra-multiconfig-gcc-stable/203/ skipped?

The Semlock plugin always [caused an error](https://github.com/ElektraInitiative/libelektra/issues/1260) when we used this configuration.

sanssecours:
Anyway, how can this be `ready to merge` if nobody called `jenkins build all` beforehand?

KurtMi:
> Anyway, how can this be ready to merge if nobody called jenkins build all beforehand?

This policy for a such easy fix is new to me, but I will rebase #1730 to master and run there build all, I have to do it anyway.

markus2330:
>  The Semlock plugin always caused an error when we used this configuration.

So is this case disabled? Can you enable it now again?

> Anyway, how can this be ready to merge if nobody called jenkins build all beforehand?

Good point I thought it won't be necessary for excluding a unit tests. From costs benefits: there is a heavy cost running all tests and the benefit is small: It was unlikely that we both overlook a syntax error.

If we always enforce "jenkins build all please" also for trivial patches we will encourage people to build larger PRs including unrelated stuff, which is problematic for other reasons (like endless review rounds).

sanssecours:
> So is this case disabled? Can you enable it now again?

I just did that.

> Good point I thought it won't be necessary for excluding a unit tests. From costs benefits: there is a heavy cost running all tests and the benefit is small: It was unlikely that we both overlook a syntax error.

In my opinion we should run `jenkins build all` for all PRs, that could break anything. Having a stable `master` branch is much more important than being able to merge something a few hours earlier.

> If we always enforce "jenkins build all please" also for trivial patches we will encourage people to build larger PRs including unrelated stuff, which is problematic for other reasons (like endless review rounds).

I do not think that is the case. Case in point: PR #1796.

PR #1802, opened 2018-01-30
debian: updated and added licenses in "copyright"

waht:
# Purpose

Moved updated licenses from #1780 to this PR.

# Checklist

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -201,7 +201,20 @@ License: OFL-1.1
 
 Files: src/tools/rest-frontend/public/vendor/asciinema/asciinema-player.*
 Copyright: 2011-2016 Marcin Kulik
-License: Apache-2.0
+License: GPL-3-asciinema
+ This program is free software: you can redistribute it and/or modify it under
+ the terms of the GNU General Public License as published by the Free Software
+ Foundation, version 3.
+ .
+ This program is distributed in the hope that it will be useful, but WITHOUT ANY
+ WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ PARTICULAR PURPOSE. See the GNU General Public License for more details.
+ .
+ You should have received a copy of the GNU General Public License along with
+ this program. If not, see http://www.gnu.org/licenses/
+ .
+ On Debian systems, the complete texts of the GNU General Public Licenses
+ version 3 can be found in `/usr/share/common-licenses/GPL-3'.
 
 Files: src/tools/rest-frontend/resources/assets/skin/base/fonts/font-awesome/*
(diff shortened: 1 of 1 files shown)


debian/copyright
```diff
@@ -154,7 +154,7 @@ License: BSD-3-clause-inih
 
 Files: src/plugins/ni/nickel-1.1.0/*
 Copyright: 2008 Charles Lindsay <charles@chaoslizard.org>
-License: BSD-3-clause-nickel
+License: Zlib
```
markus2330: Ohh, thanks for fixing this! Is it a completely unmodified Zlib license? Otherwise Zlib-nickel might be better.

waht: You're welcome. Stumbled upon it and at first thought it was BSD-3-clause as well :)
It's unmodified except for the copyright header...i'll change it to Zlib-nickel and include the content of the license file here.

waht: Wait...the license below does not have the header and copyright is included in the line above. I guess it's to leave it as "Zlib" fine then?

markus2330: > I guess it's fine to leave it as "Zlib"?

Sorry, I do not know, this file originally is from @pinotree

waht: Ok I did some reading on the topci and as long as the license text is unmodified an identifier (specified with `License: `) can be reused.

debian/copyright
```diff
@@ -201,15 +201,55 @@ License: OFL-1.1
 
 Files: src/tools/rest-frontend/public/vendor/asciinema/asciinema-player.*
 Copyright: 2011-2016 Marcin Kulik
-License: Apache-2.0
+License: GPL-3-asciinema
```
waht: The license was changed from GPL-3 to Apache-2.0 in [this commit from June 2017](https://github.com/asciinema/asciinema-player/commit/8b24d8dcefe1f7419d06bc5185c4896ca51aff12) but according to our history the files were added before. So technically they were GPL-3 then.

I could not determine whether it is/was [GPL-3](https://spdx.org/licenses/GPL-3.0-only.html) or [GPL-3+](https://spdx.org/licenses/GPL-3.0-or-later.html). The only difference between those licenses I could find seems to be in the license header within files. Since the player's source files do not contain such a header I assumed it was GPL-3 only.

markus2330: This would be a problem, we could not distribute it within our repo then. Can you please replace the file to the one that uses Apache-2.0?

waht: Oh I see. It's, done - PR #1804 has the necessary changes.

markus2330:
Thank you so much for looking into these license issues! This is a great contribution to make the package maintainer's life easier.

PR #1803, opened 2018-01-31
YAML CPP: Update CMake Code for yaml-cpp 0.6

sanssecours:
# Purpose

Do not require [Boost](http://www.boost.org) for the YAML CPP plugin anymore, if CMake locates yaml-cpp 0.6 or newer. 

# Checklist

- [x] The release notes are up do date.

### DIFF
--- a/src/plugins/yamlcpp/CMakeLists.txt
+++ b/src/plugins/yamlcpp/CMakeLists.txt
@@ -1,20 +1,22 @@
 if (DEPENDENCY_PHASE)
-	find_package (Boost QUIET)
-	if (NOT Boost_FOUND)
-		remove_plugin (yamlcpp "boost was not found")
-	endif (NOT Boost_FOUND)
-
 	find_package (yaml-cpp QUIET 0.5)
 	if (NOT YAML-CPP_FOUND)
 		remove_plugin (yamlcpp "yaml-cpp (libyaml-cpp-dev >= 0.5) not found")
 	else (NOT YAML-CPP_FOUND)
-		# Disable warnings about shadow declarations in yaml-cpp source code
-		if (APPLE AND CMAKE_COMPILER_IS_GNUCXX)
-			string (REPLACE "-Wshadow" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-		endif (APPLE AND CMAKE_COMPILER_IS_GNUCXX)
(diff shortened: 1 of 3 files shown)


sanssecours:
jenkins build all please

markus2330:
Thank you for your proactive maintenance!

src/plugins/yamlcpp/CMakeLists.txt
```diff
@@ -1,20 +1,22 @@
...
 		remove_plugin (yamlcpp "yaml-cpp (libyaml-cpp-dev >= 0.5) not found")
 	else (NOT YAML-CPP_FOUND)
-		# Disable warnings about shadow declarations in yaml-cpp source code
-		if (APPLE AND CMAKE_COMPILER_IS_GNUCXX)
-			string (REPLACE "-Wshadow" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-		endif (APPLE AND CMAKE_COMPILER_IS_GNUCXX)
+		if (YAML-CPP_VERSION VERSION_LESS "0.6.0")
+			find_package (Boost QUIET)
+			if (NOT Boost_FOUND)
+				remove_plugin (yamlcpp "yaml-cpp ${YAML-CPP_VERSION} requires Boost")
```
markus2330: I wanted to merge it as is but because there is a merge conflict in the release notes anyway, here a small note:

Can you add here the information that yaml-cpp <= 0.6.0 requires boost? If this is clearly stated in the CMake message of remove_plugin some people might decide to upgrade yamlcpp instead of installing boost.

sanssecours: > Can you add here the information that yaml-cpp <= 0.6.0 requires boost?

In my opinion the current error message:

```
Exclude Plugin yamlcpp because yaml-cpp 0.5.3 requires Boost
```

already states that.

markus2330: Yes, it describes the situation properly but it still might be misleading. It somehow suggests to install boost but maybe upgrading yamlcpp would be the better way. But that a upgrade of yamlcpp might help is not part of the message.

But it is minor, most likely nobody will ever run into this situation.

markus2330:
Again thank you!

If you decide to change the message, you can directly commit into master.

PR #1806, opened 2018-02-01
fix-check-distribution-concurrency

### LINKED ISSUE #1805: Multiple builds fail due to race condition in one test

On the build server we have a new node which can basically execute several builds in parallel. Basically just launch a multiconfig-gcc47-cmake-options build.

## Expected Result

I expect that all builds pass.

## Actual Result

The subbuilds of the multiconfig build sometimes fail due to a race condition in `testscr_check_distribution` as it creates a file in /tmp, so multiple builds can't run the tests at the same time due to this.

```
Check if script tests the correct version
Testing sibling
  Check distribution of system/tests/script/distribution/a1 and system/tests/script/distribution/b2
fatal (no cleanup necessary): /tmp/file1 already exists
```

Maybe it is possible do create a temporary directory with mktemp and do the tests inside there to avoid this race condition?

In any way I've limited the number of builds back to 1 for the new node in the meanwhile, though its capable of more which we should utilize if possible.

e1528532:
# Purpose

fixes the concurrency issue when running multiple tests simultaneously by using mktemp instead of a fixed filename.

# Checklist

@markus2330 hopefully it was a simple fix, lets see what build all thinks

### DIFF
--- a/tests/shell/check_distribution.sh
+++ b/tests/shell/check_distribution.sh
@@ -15,16 +15,10 @@ check_distribution()
 	echo "  Check distribution of $1 and $2"
 	MOUNTPOINT1="$1"
 	MOUNTPOINT2="$2"
-	FILE1="/tmp/file1"
-	FILE2="/tmp/file2"
+	FILE1=`mktemp /tmp/file1XXXXXX`
+	FILE2=`mktemp /tmp/file2XXXXXX`
 	VALUE1="value111111"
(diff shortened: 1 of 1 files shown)


e1528532:
jenkins build all please

markus2330:
If the build jobs passed this is safe to merge. You can also click the button.

e1528532:
jenkins build fast please

Issue #1824, opened 2018-02-23
Test for Notification Library Causes Memory Leaks


sanssecours:
# Problem Description

It looks like `testlib_notification` [leaks some memory](https://travis-ci.org/ElektraInitiative/libelektra/jobs/344995137#L1526) since we [changed the default storage](https://github.com/ElektraInitiative/libelektra/commit/91c3247edcd1dd5b48579911b323eca62bad9ad6) back to `dump`. This causes problems on Travis where the ASAN enabled Linux build job currently fails.

markus2330:
Thank you for reporting!

@waht Do you have an idea?

waht:
Thanks! I had a quick look at the traces but I'm not sure what causes the leak.

I will look into it after finishing changes on the notification library in PR #1825 which should done at the beginning of March.

markus2330:
Thank you for looking into it so quickly!

The issue might be also related to the mem leak discussed in #1826. (which might be caused by global plugins, like this issue here).

We should add extra tests for mountGlobalPlugin and better test the global plugins in general. Maybe @mpranj will find the problem while testing the placements for the mmap global plugin.

mpranj:
Interestingly, the leaks in the log all have a root in `test_openclose` from `testlib_notification.c` from kdbOpen. The other tests do kdbOpen too, but not with a cascading key. Maybe that is just a random observation though.

markus2330:
Thanks for looking into it.

If I make a wild guess there is some assumption in the global plugin processing about present keys, which is only fulfilled with INI (because it always adds section keys).

I have some doubts, however, that this is the only problem.

waht:
Good catch! I gave it [a shot](https://travis-ci.org/ElektraInitiative/libelektra/jobs/345999946#L1547) but the trace stayed the same.

markus2330:
So the issue has nothing to do with INI/dump?

waht:
I was able to locate the source of the "leak". The problem was that the test aborted because `kdbOpen()` failed due to a missing symbol in the storage plugin:
```
root@92a127a364ed:/mnt/share/build_docker# ASAN_OPTIONS=symbolize=1 ASAN_SYMBOLIZER_PATH=/usr/lib/llvm-5.0/bin/llvm-symbolizer ./bin/testlib_notification 
test open & close
src/libs/elektra/kdb.c:260:kdbOpen: called with system/sw/tests/testlib_notification
[...]

Issue #1134, opened 2016-11-23
dbus synopsis


beku:
The [dbus readme](../blob/master/src/plugins/dbus/README.md) gives an example.
However, it is unclear for me on how to adapt to our needs.

What I want:
* sync client views of settings across a user desktop session
* obtain messages for key changes in our name space (org/freedesktop/openicc/...)
* use Qt5DBus for Qt based GUI: [Synnefo](https://github.com/oyranos-cms/Synnefo)
* observe in [FLTK](http://www.fltk.org), (even not sure what is the best way there.)

markus2330:
Hi Kai-Uwe, thanks for reporting this issue.

@waht is working on this topic and hopefully will write some nice tutorials to help you with these questions when he is ready.

You requirements are quite straight-forward and I do not think we will have any issues in supporting them. Just out of interest: Is it also a requirement for you that dbus is used or are other bus systems okay, too?

In https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/dbus/receivemessage.c you also see how to receive a message in C.
@waht also looks into the issues of main/event loop integration.

But it is an ongoing topic and there are some issues: #955, #404 but the basic things should work.

beku:
Oh. Thanks.

What I am looking for is more specifically the mount dbus key_path synopsis.

dbus-monitor type='signal',interface='org.libelektra',path='/org/libelektra/configuration' # gives no messages atm

markus2330:
You got the synopsis right and it works for me.

```sh
kdb mount file.dump /dbus dump dbus
kdb set user/dbus/y a  # should fire in session bus
kdb set system/dbus/y   # should fire in system bus
```

For me both dbus-monitor and the C code works:
```sh
dbus-monitor --system type='signal',interface='org.libelektra',path='/org/libelektra/configuration'
#> signal sender=:1.100875 -> dest=(null destination) serial=2 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
   string "user/dbus/x"
kdb testmod_dbus receive_system 
#> Notify received
```

Maybe you only tested with system and forgot to add `--system`? I clarified the docu.

beku:
Following appears not to mount successfully (appreciate a correct mount line ;)

    kdb mount file.yajl org/freedesktop/openicc yajl dbus

mount-openicc says:

[...]

Issue #868, opened 2016-07-29
proposal: kdbEnsure


markus2330:
As discussed we need for some applications a way to ensure that some global plugin are present. For example, e.g. applications that need the cachefilter (written by @Namoshek) or need the internalnotification ( #828 written by @waht).

More generally put, some Applications need behaviour from Elektra that was not there some versions ago and cannot be the new default.

In the long term, we should extend `kdbOpen` so that users can request any functionality (either by plugins or the core itself).

In the shorter term, we could add a `kdbEnsure` that will add a global plugin if it was not there before, exactly fulfilling the current needs.

A possible API would be:

```
int kdbEnsure (KDB *kdb, KeySet *contractToEnsure, Key *errorKey);
```

Namoshek:
`kdbEnsure` sounds like a good way to go about it for now. To me changing `kdbOpen` looks like a deep change that would probably fit more during a new major version of the library where breaking changes are fine. I know it is possible to overload functions with some tricks since C11, but I'd rather not deploy such a big change with a minor version.

markus2330:
The current plan is that we create new APIs (similar to `keyRel2` #826) in libproposal, later we can use symbol versioning like libc does, and with `1.0.0` we get rid of the legacy stuff (maybe move it to `libelektra-legacy`).

markus2330:
@kodebach Is this maybe a byproduct of the checks done within the high-level API?

markus2330:
In #2383 we found that sometimes the absence of plugins might be required (e.g. for cmd-line options).

So for now, we either want to say that a global plugin needs to be present (internalnotification or also cmd-line options) or should be absent (cmd-line options for maintenance tasks).

kodebach:
Instead of creating `kdbEnsure`, we could also pass the information to `kdbOpen` or `kdbGet` directly. 

The advantage of passing the information to `kdbOpen` directly is that we might be able to avoid doing unnecessary work, which would only be undone by `kdbEnsure`. However, the only way of passing information to `kdbOpen` would be using the metadata of the `errorKey` parameter. This seems kind of inconvenient compared to passing an actual `KeySet`. One possibility would be to add `kdbOpen2` to libproposal as mentioned above.

`kdbGet` meanwhile already receives a `KeySet`, so we could simply use keys below `proc/elektra` (so they won't interfere with actual keys retrieved by `kdbGet`). The only real advantage here is that no extra function call and therefore no extra error handling is needed.

[...]

PR #1825, opened 2018-02-23
notification: dbus transport plugins

waht:
# Purpose

This PR contains dbus transport plugins for the notification feature.

# Checklist

- [x] dbus[send] plugin (adds asynchronous sending of notifications to existing dbus plugin)
- [x] dbusrecv plugin (asynchronous receiving of notifications via dbus)
- [x] I added unit tests
- [x] find and fix memory leak (#1824)
- [x] dbusrecv: add support for "Commit" signals
- [x] notification: check registered keys on notifications
- [x] architecture: improve plugin callbacks


src/libs/io/io.c
```diff
@@ -9,13 +9,97 @@
...
+
+/**
+ * @internal
+ * Retrieves a function exported by a plugin.
+ *
+ * @param  plugin Plugin handle
+ * @param  name   Function name
+ * @return        Pointer to function
+ */
+static size_t getPluginFunction (Plugin * plugin, const char * name)
```
waht: @markus2330: Is there a way to share functions like these internally (e.g. with libraries and plugins) and not export them? I feel like I'm missing some obvious choice here :)

markus2330: To share a library function (in any way outside of the source file/library itself) requires the function to be exported (including the c file is a hack only suitable for tests).

To share a function in a plugin you can use the contract (the `exports`). Of course you could realize something similar for your library but that most likely is overkill.

Btw. do you know elektraInvokeGetFunction from the elektra-invoke library implemented in src/libs/invoke/invoke.c?
It looks very similar to your code.

waht: Ok, thank you very much!

Yes I have seen it, but it uses `ElektraInvokeHandle` which wraps `Plugin`. I'm using it at various points (like when [iterating global plugins](https://github.com/ElektraInitiative/libelektra/pull/1825/files#diff-c7a393ddf5f92626daa6ec9e73774309R94) or [passing the I/O binding to list's subplugins](https://github.com/ElektraInitiative/libelektra/pull/1825/files#diff-2233b84e0123b41a7f3eca5fa59dec4dR363) and in the notification library to access register*-functions) where I directly have a Plugin handle available.

How about adding this particular function to `libplugin` as `elektraPluginGetFunction`?

markus2330: Yes, this is a very good idea!

I am even thinking about making elektra-invoke dependent on elektra-plugin to use your function. Both getPluginFunction and elektraInvokeGetFunction are lacking the optimization to directly use plugin->kdb* when requesting one of the 5 standard functions (open, close, get, set, error).

If we add the dep to elektra-plugin from elektra-invoke we could get rid of any access to Plugin's struct within elektra-invoke, making elektra-invoke independent of internals of the Plugin's struct. elektra-plugin obviously needs by design access to the struct.

waht: Ok, great! I will add it to libplugin. 

I can make the changes to elektra-invoke later on within this PR.

src/libs/io/io.c
```diff
@@ -9,13 +9,97 @@
...
+	size_t func = *buffer;
+
+	elektraFree (buffer);
+	ksDel (exports);
+	keyDel (pk);
+
+	return func;
+}
+
+static void debugKeySet (char * name, KeySet * ks)
```
markus2330: This function yields a warning because static functions need to be used:
https://build.libelektra.org/jenkins/view/compiler%20warnings/job/elektra-gcc-configure-debian-log/1309/warningsResult/source.-2911739015347291038/#57

waht:
I would like to add automated tests for the dbus plugins. This requires either mocking of the dbus library functions or having a dbus daemon running on the build server. 
It is possible to start dbus manually with `dbus-daemon --system` (as root at least).
What do you think, @sanssecours @markus2330?

src/libs/notification/notification.c
```diff
@@ -660,6 +617,65 @@ int unmountGlobalPlugin (KDB * kdb, Plugin * plugin)
...
+			}
+			ElektraNotificationCloseNotification closeNotification = (ElektraNotificationCloseNotification)func;
+			closeNotification (plugin);
+		}
+	}
+}
+
+void elektraNotificationDoUpdate (Key * key, void * data)
+{
+	KDB * kdb = (KDB *)data;
```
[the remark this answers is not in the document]
markus2330: Why is `void*` used instead of `KDB*`?

Yes, it can make sense to have an extra struct, then you can pass more than a `KDB*` handle to the callback.

In any case: This is a question of API design. Why should it be related to security?

sanssecours:
> It is possible to start dbus manually with dbus-daemon --system (as root at least).

At least on macOS (Travis) you should be able to start dbus with `brew services start dbus`. Starting `dbus` with

```sh
sudo dbus-daemon --system
```

fails with the message

> dbus-daemon[94768]: Failed to start message bus: Could not get UID and GID for username "messagebus"

, at least on my machine. Maybe the same command works on the Linux Travis build jobs. Anyway, I would just recommend you try it. 

If you think the build times are too long, I would recommend you just remove most of the code in `travis.yml`, register with your GitHub account at http://travis-ci.org and try to change `.travis.yml` at your copy of the repo. This way you do not have to share the Travis resources with the other Elektra developers here.

src/libs/notification/notification.c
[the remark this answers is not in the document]
waht: I haven't decided about the struct but didn't want to use `KDB` so I used void for now.

I thought in our discussion on why `KBD` was not in `Plugin` security was the issue.

If having KDB passed in a struct to plugins is not an issue, I'll use a struct to keep the API easier to maintain.

markus2330:
Installing dbus on the Jenkins agents is no problem, I suppose most agents already have dbus.

In any case, it would be good if you check within tests if dbus is available and stop the test if it won't work.

waht:
Ok great! Thanks for looking into it and for the tip, @sanssecours! Build times should not be affected (unless I need dbus needs to be compiled of course), but my own Travis account is probably useful until I figure out the build setup. 

@markus2330 Sure, I will make the tests fail with a warning if dbus is not available.

markus2330:
> Sure, I will make the tests fail with a warning if dbus is not available.

Please succeed with warning. If test cases do not run through, people won't run them.

waht:
The dbus plugin tests now only fail when no D-Bus connection (neither system nor session bus) was possible. On MacOs the tests use the session bus, on Linux the system bus.

You were talking about having a warning (i.e. a printf) and skipping the further tests, right? My concern is that the warning would not be visible when running the tests with `run_all` & co.

@markus2330 please review my pull request - I'll go hunting for the memory leak from #1824 :)

cmake/Modules/LibAddPlugin.cmake
```diff
@@ -45,7 +45,7 @@ function (add_plugintest testname)
 		restore_variable (${PLUGIN_NAME} ARG_COMPILE_DEFINITIONS)
 		restore_variable (${PLUGIN_NAME} ARG_INCLUDE_DIRECTORIES)
 		restore_variable (${PLUGIN_NAME} ARG_INCLUDE_SYSTEM_DIRECTORIES)
-		restore_variable (${PLUGIN_NAME} ARG_LINK_ELEKTRA)
+		restore_variable (${PLUGIN_NAME} ARG_LINK_ELEKTRA ALLOW_MATCH)
```
waht: @markus2330 the dbusrecv plugin [required io_uv for testing](https://github.com/ElektraInitiative/libelektra/pull/1825/files#diff-67c9a608911fa20894d47412d4fba5f2R28): While the dbusrecv plugin relies on the ElektraIoInterface, to test the code a I/O binding was required.

I relaxed the consistency checks so that `add_plugintest` allows additional libraries. The plugin libraries still need to be included.

waht:
jenkins build all please

waht:
jenkins build gcc-configure-debian-musl please

waht:
jenkins build multiconfig-gcc47-cmake-options please

waht:
jenkins build gcc-configure-debian-musl please

waht:
jenkins build all please

waht:
jenkins build gcc-configure-debian-debug please

waht:

waht:

waht:

waht:
Okay, we've located the source for the memleak in #1824: The ubsan library was missing - the workaround (preloading it) is included in this PR or should be wait for the process plugin PR (#1783) which includes this fix as well?

markus2330:
@e1528532 Any plans to finish #1783?

In any case: if you git cherry-pick a fix or your fix is identical (char-by-char) there is no problem if it is applied twice.

waht:
Ok I realized that we have a similar issue (and curiously the same missing symbol `__ubsan_vptr_type_cache`) which is fixed by preloading different libraries.

In order to resolve my issue preloading the UbSan c++ library is required. The ASan library (as in #1783) does not work for me (the output says `==28==Your application is linked against incompatible ASan runtimes.`).

I changed my PR so that it is possible to preload different libraries without creating conflicts with git.

waht:
jenkins build haskell please

waht:
jenkins build haskell please

sanssecours:
jenkins build haskell please

sanssecours:
jenkins build docker please

sanssecours:
Seems like I somehow misconfigured the build job [`elektra-haskell`](https://build.libelektra.org/jenkins/job/elektra-haskell/). Sorry for that. The build phrase `jenkins build haskell please` should now work again.

waht:
Great, thanks!

src/plugins/list/list.h
```diff
@@ -19,6 +21,9 @@ int elektraListSet (Plugin * handle, KeySet * ks, Key * parentKey);
 int elektraListError (Plugin * handle, KeySet * ks, Key * parentKey);
 int elektraListAddPlugin (Plugin * handle, KeySet * pluginConfig);
 int elektraListEditPlugin (Plugin * handle, KeySet * pluginConfig);
+void elektraListSetIoBinding (Plugin * handle, ElektraIoInterface * binding);
```
markus2330: Is this the right place? It couples the ioplugin interface to everyone using the list plugin.

src/plugins/list/list.c
```diff
@@ -311,6 +317,33 @@ static int runPlugins (KeySet * pluginKS, KeySet * modules, KeySet * plugins, Ke
 				keyDel (slaveKey);
 			}
 		}
+		if (ioBinding)
```
markus2330: Is this coupling necessary? (What about callbacks or some other more generic mechanism) If it is necessary: The function is already very long, can you refactor this a bit?

src/plugins/internalnotification/README.md
```diff
@@ -40,6 +40,10 @@ Cascading keys as `key` names are also supported.
     `system/elektra/modules/internalnotification/exports/handle`.
 - *key* Key to watch for changes.
 
+*Returns*
+
+1 on success, 0 otherwise
```
markus2330: Isn't this information already in API docu? What about putting a link there?

src/plugins/dbusrecv/testmod_dbusrecv.c
```diff
@@ -0,0 +1,253 @@
...
+	test_prerequisites ();
+
+	uv_loop_t * loop = uv_default_loop ();
+	ElektraIoInterface * binding = elektraIoUvNew (loop);
+
+	// Test added, changed & deleted
+	test_keyAdded (loop, binding);
+	test_keyChanged (loop, binding);
+
+	printf ("\ntestmod_dbusrecv RESULTS: %d test(s) done. %d error(s).\n", nbTest, nbError);
```
markus2330: @sanssecours wrote a print_result which makes the output the same across all tests.

src/plugins/dbusrecv/README.md
```diff
@@ -0,0 +1,45 @@
+- infos = Information about the dbus plugin is in keys below
+- infos/author = Markus Raab <elektra@libelektra.org>
```
markus2330: You can put your name here.

Is this copied or moved from dbus plugin?

src/plugins/dbusrecv/README.md
```diff
@@ -0,0 +1,45 @@
...
+- infos/licence = BSD
+- infos/provides = notification
```
markus2330: maybe we should give a different provider name for your async io notification receiving and sending plugins?

src/plugins/dbusrecv/README.md
```diff
@@ -0,0 +1,45 @@
...
+- infos/needs =
+- infos/recommends =
+- infos/placements = postgetstorage postcommit
+- infos/status = maintained unittest libc global experimental
+- infos/description = Sends DBus signals when a method is called
```
markus2330: "a method is called"?

src/plugins/dbus/receivemessage.c
```diff
@@ -8,8 +8,91 @@
...
+	dbus_bus_remove_match (connection, RECEIVE_MATCH_RULE, &error);
+	if (dbus_error_is_set (&error)) goto error;
+
+	dbus_connection_remove_filter (connection, filter_func, data);
+
+	dbus_error_free (&error);
+	return 1;
+
+error:
+	printf ("Error occurred\n");
```
markus2330: Is this library or test code?

src/plugins/dbus/README.md
```diff
@@ -187,3 +187,33 @@ of the whole system knows that a chain of reactions will terminate.
...
+
+Mount this plugin globally with default settings to use it as *sending*
+transport plugin for Elektra's notification feature:
+
+> kdb global-mount dbus
+
+# Notification Format
+
+This plugin uses D-Bus signal messages as notifications.
+Notifications we path `/org/libelektra/configuration` and the D-Bus interface
```
markus2330: "we path"?

src/plugins/dbus/README.md
```diff
@@ -187,3 +187,33 @@ of the whole system knows that a chain of reactions will terminate.
...
+The first argument contains the name of the changed key.
+The system bus is used if the affected keys is below the `system` namespace.
+If the key is below the `user` namespace the session bus is used.
+
+Example output from `dbus-monitor`:
+
+```
+signal time=1519856379.538736 sender=:1.1 -> destination=(null destination) serial=18 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
+   string "system/tests/foo/bar"
+signal time=1519856379.538753 sender=:1.1 -> destination=(null destination) serial=19 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyChanged
```
markus2330: Is this information used within your internal notification?

Otherwise we already said we do not need a notification for every key individually. See #1134

src/plugins/dbus/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ add_plugin (dbus
 	INCLUDE_DIRECTORIES
 		${DBUS_INCLUDE_DIR}
 		${DBUS_ARCH_INCLUDE_DIR}
+	LINK_ELEKTRA
+		elektra-io-adapter-dbus
```
markus2330: Does sending dbus messages also work without any support from applications?

Would be good to have some documentation/tutorials/...

src/libs/notification/example/CMakeLists.txt
```diff
@@ -1,17 +1,45 @@
...
 # target_include_directories (${example} SYSTEM PUBLIC ${LIBUV_INCLUDE_DIRS})
 
 target_link_elektra (${example} elektra-kdb elektra-notification)
+
+# The asynchronous example requires io_uv
+check_binding_included ("io_uv" IS_INCLUDED SUBDIRECTORY "io/uv" SILENT)
+find_package (PkgConfig)
+pkg_check_modules (LIBUV QUIET libuv>=1)
+if (IS_INCLUDED AND LIBUV_FOUND)
+  ## Build example_notification_async
```
markus2330: spaces?

src/libs/notification/example/CMakeLists.txt
```diff
@@ -1,17 +1,45 @@
...
+  set (example example_notification_async)
+
+  SET (SRC_FILES
```
markus2330: SET (capitals?)

src/libs/io/adapter/dbus/dbus.c
```diff
@@ -0,0 +1,274 @@
...
+		flags |= ELEKTRA_IO_WRITABLE;
+	}
+	// from docs: "all watches implicitly include a watch for hangups, errors, and other exceptional conditions."
+	// TODO add DBUS_WATCH_ERROR and DBUS_WATCH_ERROR to flags
+	return flags;
+}
+
+static void dbusWrapperFree (void * memory)
+{
+	// fprintf(stderr, "dbusWrapperFree\n");
```
markus2330: Please use logger here.

cmake/ElektraCache.cmake
```diff
@@ -346,11 +346,16 @@ set (LIB_SUFFIX ""
 		"Optional suffix to use on lib folders (e.g. 64 for lib64)"
     )
 
-set(MEMORYCHECK_SUPPRESSIONS_FILE
+set (MEMORYCHECK_SUPPRESSIONS_FILE
 		"${CMAKE_SOURCE_DIR}/tests/valgrind.suppression"
 		CACHE FILEPATH
 		"Full path to suppression file for valgrind")
 
+set (ENABLE_IO_ADAPTER_DBUS "ON"
```
markus2330: I am not very happy about even more CMake vars. Why not explictly exclude debian-musl at the place where needed? Btw. What is the compilation error, why cannot it be fixed?

cmake/Modules/LibAddPlugin.cmake
markus2330: Would be better if we can avoid circular deps.

If we really would need such a hack, we should introduce LINK_LIBRARIES_CIRCULAR or similar which adds the ALLOW_MATCH (the name ALLOW_MATCH is not really good, too?)

scripts/configure-debian-musl
```diff
@@ -8,8 +8,9 @@ cd $BUILD
 CC=musl-gcc CXX=g++ cmake \
 	-DINSTALL_BUILD_TOOLS=ON \
 	-DBUILD_TESTING=OFF -DBUILD_FULL=OFF -DBUILD_STATIC=OFF \
-	-DPLUGINS="ALL;-yajl;-journald;-CRYPTO;-dbus;-xmltool;-curlget;-multifile" \
+	-DPLUGINS="ALL;-yajl;-journald;-CRYPTO;-dbus;-dbusrecv;-xmltool;-curlget;-multifile" \
 	-DBINDINGS="cpp" \
 	-KDB_DEFAULT_STORAGE=ini \
 	-DTARGET_PLUGIN_FOLDER="" \
+	-DENABLE_IO_ADAPTER_DBUS="OFF"\
```
markus2330: please avoid

src/bindings/io/uv/CMakeLists.txt
```diff
@@ -69,18 +60,20 @@ else ()
 	list (APPEND TEST_SOURCES ${IO_TEST_SRC_FILES})
 	list (APPEND TEST_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/testio_${BINDING_VARIANT}.c")
 
+	if (BUILD_FULL OR BUILD_STATIC)
+    # add sources for elektra-io-uv for static and full builds
```
markus2330: spaces?

src/bindings/io/uv/example/exampleio_uv.c
```diff
@@ -127,11 +127,7 @@ int main (void)
 	// Start the event loop
 	uv_run (loop, UV_RUN_DEFAULT);
 
-#ifdef HAVE_LIBUV1
 	uv_loop_close (loop);
-#elif HAVE_LIBUV0
-	uv_loop_delete (loop);
```
markus2330: Why remove support for older libs?

src/include/kdbnotification.h
```diff
@@ -34,7 +34,8 @@
  * (see
  * <a href="https://www.libelektra.org/plugins/internalnotification">its plugin docs</a>).
  *
- * - If no plugin is mounted at a desired position it is simply mounted.
+ * - If no plugin is mounted at a desired position the internalnotification
```
markus2330: Sentence still not clear.

src/include/kdbprivate.h
```diff
@@ -15,6 +15,7 @@
 #include <kdbhelper.h>
 #include <kdbio.h>
 #include <kdbmacros.h>
+#include <kdbnotificationinternal.h>
```
markus2330: Do we need the internal header file only for a callback?

src/libs/io/adapter/dbus/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+include (LibAddMacros)
+
+if (NOT ENABLE_IO_ADAPTER_DBUS)
+	# required for debian-musl build - not other purpose
```
markus2330: "not other purpose"?

And as written above: Please directly check for the reason why the compilation fails.

src/libs/io/adapter/dbus/dbus.c
```diff
@@ -0,0 +1,274 @@
...
+	if (dbus & DBUS_WATCH_READABLE)
+	{
+		flags |= ELEKTRA_IO_READABLE;
+	}
+	if (dbus & DBUS_WATCH_WRITABLE)
+	{
+		flags |= ELEKTRA_IO_WRITABLE;
+	}
+	// from docs: "all watches implicitly include a watch for hangups, errors, and other exceptional conditions."
+	// TODO add DBUS_WATCH_ERROR and DBUS_WATCH_ERROR to flags
```
markus2330: todo

src/libs/io/adapter/dbus/CMakeLists.txt
waht: Should have been "no other purpose" :)

src/plugins/dbusrecv/README.md
waht: I also call them "transport plugins". Maybe `notificationtransport`?

src/plugins/dbusrecv/README.md
waht: Sure :)

src/plugins/list/list.c
waht: It was the only solution I could find: Without a KDB handle in the Plugin structure the plugins cannot get the notification callback or I/O binding from stored in the KDB handle. So the notification library and `elektraIoSetBinding()` need to "push" callbacks to the plugins using exported functions (`open/closeNotification` and `setIoBinding`).

Since the list plugin loads its plugins internally I cannot get these functions from `system/elektra/modules/...`. Therefore the list plugin needs to act as a proxy for these exported functions. 
The plugin cannot set it directly because it lazy-loads the plugins later at the first `kdbGet()` or `kdbSet()` which is why I had to add it here.

But of course in the long-term it is not feasible to implement every possible exported function that needs to be passed to plugins. How about these two solutions for more generic exported functions?

A simple generic solution would be to have a `getNestedPlugins` function exported by the list plugin and similar plugins. The list plugin would have to load all plugins when this funciton is called for `setIoBinding` and `openNotification` to work since these are called before the first `kdbGet/Set()`.

Another solution that would make the exported functions also generic for the caller would be to have a `getInterfaces` function exported by compatible plugins. The returned "interfaces" could be a `KeySet` with well-known interface name like "notification" or "ioBinding" and the required functions (e.g. `open`, `close` for "notification", or `set` for "ioBinding").
Normal plugins would simply return their interfaces. Plugins like the list plugin could collect interfaces from its nested plugins (again after loading them) and return a KeySet that combines all interfaces from the nested plugins.

Here is an example interface-`KeySet` from a list plugin with two nested plugins A and B both implementing the "notification" interface. Plugin B also has the "ioBinding" interface:
- `/notification/#0/open` => address of `openNotification` of plugin A
- `/notification/#0/close` => address of `closeNotification` of plugin A
- `/notification/#1/open` => address of `openNotification` of plugin B
- `/notification/#1/close` => address of `closeNotification` of plugin B
- `/ioBinding/#0/set` => address of `setIoBinding` of plugin B

Both solutions would only circumvent the list plugin's lazy-loading when the notification features are used by an application. The second solution might be a bit over-engineered :)

src/plugins/list/list.h
waht: Yes it couples it, but it is optional to call the exported function. I added more details in the comment below.

src/plugins/internalnotification/README.md
waht: Yes, it kind of is in the API docs: The elektraNotificationRegister* functions are in the notification API docs, but not the functions exported by the plugin.
Since application developers should use the notification API instead of the exported functions I will add a link there.

src/plugins/dbus/README.md
waht: Ops - this didn't make sense.

src/plugins/dbus/CMakeLists.txt
waht: Yes it works without modifications. If an I/O binding is set by an application using `elektraIoSetBinding()` the plugin uses it to send dbus messages.

I will add this to the dbus plugin readme.

waht: Well, after reading the API docs for `dbus_connection_send()` I discovered that it already operates asynchonously. So the adapter is only needed in the dbusrecv plugin.

src/libs/io/adapter/dbus/dbus.c
waht: These were some leftovers from testing - I removed the fprintf statements.

cmake/ElektraCache.cmake
waht: The [build job](https://build.libelektra.org/jenkins/job/elektra-gcc-configure-debian-musl/695/consoleFull) failed with `/usr/bin/ld: cannot find -ldbus-1`.
When configuring CMake it is present (`-- Found D-Bus: /usr/lib/x86_64-linux-gnu/libdbus-1.so`) so it seems like musl cannot link against it.
The dbus plugin was already disabled for this build job.

I agree. I will remove the variable and check with `${CMAKE_C_COMPILER} MATCHES "musl"`.

src/include/kdbnotification.h
waht: I rephrased it to "If there is no plugin mounted at a required position the internalnotification plugin is mounted at this position."

src/include/kdbprivate.h
waht: Since the kdbnotification.h is installed and user-facing I didn't want to include these definitions there. They are only used internally by the internalnotification plugin, the notification library and transport plugins.

src/plugins/dbus/receivemessage.c
waht: It is test code. I added this info to the file header and excluded the file from the plugin sources.

src/libs/notification/example/CMakeLists.txt
waht: automatic formatting for cmake would be great :)

cmake/ElektraCache.cmake
markus2330: We should be very precise here with the exclusion, otherwise we will disable these parts for OpenWRT, too.

Why does it even try to link against dbus? Only the plugin should link against dbus and the plugin is not built on our debian-musl, isn't it?

If the exclusion is really necessary, I am afraid we need to write a little test program with `try_compile` which tries to compile and link with dbus. This is already done several times, e.g.  in "src/plugins/augeas/CMakeLists.txt:              try_compile (HAS_LIBAUGEAS16"

waht: Okay, sure we don't want that.

Additional to the I/O bindings there are reusable "adapters" for D-Bus und ZeroMq (different PR). Plugins can use these adapters. 
It tries to link against D-Bus since these adapters are built as long as the required libraries are available - I didn't want to introduce a new build variable and these adapters are not intended to be used outside elektra.

`try_compile` seems more reasonable (testing capability vs. testing presumed capability) - I'll use that.

markus2330: The initial idea was that everything with deps should be either a tool, plugin or a binding. Then users can always manually exclude tools, plugins or bindings as wished or to make Elektra compile on erroneous platforms (which claim something is available but it isn't).

It seems like, however, that now also libs start to have external deps. (Yours is not the first, with src/libs/pluginprocess we already had similar problems.) So a LIBS (like BINDINGS and PLUGINS) might be necessary in the long run?

src/plugins/dbus/README.md
waht: Ok that wasn't very clear when reading the conversation. Currently when a notification arrives `kdbGet()` is executed with this key. I plan to add a check if the application has registered a key below the changed key. The type of change is actually not used.

So I see no problem using/supporting the "Commit" announcement of the dbus plugin.

src/bindings/io/uv/example/exampleio_uv.c
waht: When fixing static & full builds with the "io_uv" binding I ran into problems. But I could solve them and restore support for libuv 0.x.

cmake/Modules/LibAddPlugin.cmake
waht: Why would the deps be circluar?
It allows to add additional libraries for plugin tests.

> If we really would need such a hack, we should introduce LINK_LIBRARIES_CIRCULAR or similar which adds the ALLOW_MATCH (the name ALLOW_MATCH is not really good, too?)

Okay. How about LINK_LIBRARIES_ADDITIONAL and "ALLOW_ADDITIONAL"?

src/plugins/dbus/README.md
markus2330: > keyNeedSync

Unfortunately this approach is buggy, see #404

> Currently when a notification arrives kdbGet() is executed with this key.

So it might lead to a massive amount of kdbGet()?

> So I see no problem using/supporting the "Commit" announcement of the dbus plugin.

Commit messages are necessary, even when sending only a single message per backend the notification system somehow needs to know when to trigger a reload (when all backends are executed). A proper design for this is badly needed.

waht:
jenkins build gcc-configure-debian-musl please

src/plugins/dbus/README.md
waht: Ok, so the change detection path is a minefield :)

> So it might lead to a massive amount of kdbGet()?

In the current state yes but I will add the necessary changes to this PR. I had some doubts about overwritten keys and cascading, but kdbGet() should handle these fine.

I thought for global plugins the `postcommit` position ensures that the backends are done and the data is final. A "Commit" message on kdbSet() with the parentKey would notify other applications about changes below this key.

cmake/ElektraCache.cmake
waht: It may be necessary, but for libs that are not only used internally I think we should prevent that. In my opinion users should not bother too much with internals.

For the adapter problem I could try to change it into an object library. The CMake documentation claims it does not link the files so the dbus issue would be moved to plugins using the adapters.

waht:
jenkins build multiconfig-gcc47-cmake-options please

waht:
jenkins build gcc-configure-debian-musl please

waht:
jenkins build multiconfig-gcc47-cmake-options please

src/plugins/list/list.c
markus2330: > need to "push" callbacks to the plugins using exported functions (open/closeNotification and setIoBinding)

There should be nothing wrong with that but the callback interface should be more generic (not specific to notification).

> The list plugin would have to load all plugins

The idea of the list plugin is to avoid exactly this. So not circumventing lazy-loading (which is the main reason for the list plugin in the first place) would be preferable.

waht: Ok, to sum up the requirements:

1. the callback interface on the list plugin should be generic (independent from notification & I/O binding)
2. the list plugin should not load all plugins before they are actually required within kdbGet()/kdbSet()
3. there is this notification initialization sequence: 
    - `kdbOpen(kdb, ...)`
    - `elektraIoSetBinding(kdb, binding)`: plugins are checked for `setIoBinding()` exports
    - `elektraNotificationOpen(kdb)`: plugins are checked for `openNotification()` exports
    - `kdbGet(kdb, ...)`: list plugin loads it's plugins

I can think of some "delayCallback" function exported by the list plugin which is used if either exported function is not present. It takes a function name and a list of arguments. When the list plugin initializes it's plugins it would check for the exported functions and call this function with the arguments. Since C is statically typed that will change the definition of both `setIoBinding()` and `openNotification()` to a KeySet (or any list) containing the arguments. 

Would that be better?

markus2330: Another requirement is that there should be an interface which is not entangled with details of the list plugin. The list plugin might be replaced and we want to avoid that your I/O code needs changes in such a situation.

> there is this notification initialization sequence: 

Do all these calls happen from the application?

Why not have an elektraNotificationOpen which also takes the binding as second parameter?

> Would that be better?

Yes, having a way to call functions within plugins
1.) without being entangled with list plugin details,
2.) for anyone (and not only I/O bindings)
3.) and still preserving lazyness
would be important.

This is needed for everything in the direction of #868

It would be great to have written everything together in a design decision. The discussion is already quite distributed.

waht: > Why not have an elektraNotificationOpen which also takes the binding as second parameter?

As long as I/O binding and notification are only used together this makes sense - in the long term I/O bindings might be used alone.

> It would be great to have written everything together in a design decision. The discussion is already quite distributed.

Okay I'll write up a design decision and add it to this PR.

waht: I added the decision and called it [deferred plugin calls](https://github.com/ElektraInitiative/libelektra/pull/1825/files#diff-4192311c812b52df60a9bd3eca54c457).

waht: I also added an initial implementation of the proposed solution. The generic functions for handling the deferred calls from the decision are now in the "list" plugin.

What would be the best place to put them? src/libs/elektra/plugin.c (with according prefix) or is an object library better which can be included on demand?

cmake/Modules/LibAddPlugin.cmake
markus2330: Ahh, I misunderstood what you are trying to achieve.

Yes, fresh names are a much better solution. What about UNITTEST_LINK_LIBRARIES? (ADDITIONAL is too generic, then it is unclear -- like it was for me -- that you only need it for the tests.)

cmake/Modules/LibAddPlugin.cmake
```diff
@@ -42,16 +42,18 @@ function (add_plugintest testname)
...
-		restore_variable (${PLUGIN_NAME} ARG_LINK_LIBRARIES)
+		restore_variable (${PLUGIN_NAME} ARG_LINK_LIBRARIES ALLOW_MATCH)
 		restore_variable (${PLUGIN_NAME} ARG_COMPILE_DEFINITIONS)
 		restore_variable (${PLUGIN_NAME} ARG_INCLUDE_DIRECTORIES)
 		restore_variable (${PLUGIN_NAME} ARG_INCLUDE_SYSTEM_DIRECTORIES)
-		restore_variable (${PLUGIN_NAME} ARG_LINK_ELEKTRA)
+		restore_variable (${PLUGIN_NAME} ARG_LINK_ELEKTRA ALLOW_MATCH)
 		restore_variable (${PLUGIN_NAME} ARG_ADD_TEST)
 		restore_variable (${PLUGIN_NAME} ARG_INSTALL_TEST_DATA)
+		restore_variable (${PLUGIN_NAME} ARG_OBJECT_SOURCES)
```
markus2330: also mention here that it is for tests.

cmake/Modules/LibAddPlugin.cmake
```diff
@@ -176,14 +178,27 @@ function (plugin_check_if_included PLUGIN_SHORT_NAME)
 endfunction ()
 
 function (restore_variable PLUGIN_NAME VARIABLE)
+	cmake_parse_arguments (ARG
+		"ALLOW_MATCH" # optional keywords
```
markus2330: should be removed then.

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
+# Deferred Plugin Calls
+
+## Issue
+
+Since some plugins may lazy-load encapsulated plugins at a later time a
```
markus2330: either "some" or "may". (some do it, in general plugins may do it).

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
...
+mechanism for deferring function calls to this encapsulated plugins is needed.
+
+For example when setting I/O bindings with `elektraIoSetBinding()` the exported
+function `setIoBinding` is called for all globally mounted plugins.
+The "list" does not export this functions and cannot implement every exported
+function for its encapsulated plugins.
```
markus2330: You should also give other examples, e.g. dini and multifile, and explain if your decision also applies to them.

At the moment the decision is too much centered to your use case.

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
...
+## Assumptions
+
+1. The called functions do not return a value (e.g. `set`, `open`, `close`, ...)
+
+## Considered Alternatives
+
+- Plugins like the "list" plugin export a function called
+  `getEncapsulatedPlugins`. This would break encapsulation and
+  break lazy-loading since to return the encapsulated plugin handles they have
+  to be loaded.
```
markus2330: You also proposed other (good) alternatives (which are worth mentioning).

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Decision
+
+Encapsulating plugins export a function called `deferFunctionCall` with the
+declaration
+`void deferFunctionCall (Plugin * plugin, char * name, KeySet * parameters)`.
```
markus2330: should start with elektra, e.g. `elektraDeferFunctionCall`.

Why not simply `elektraCallFunction`?

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
...
+the `elektraIoSetBinding()` or `elektraNotificationOpen()`.
+
+## Implications
+
+The called functions have the declaration
+`void function (Plugin * plugin, KeySet * parameters)`.
+This means that the functions also receive their parameters via a KeySet.
+
+## Related decisions
+
```
markus2330: You should describe why elektra-invoke does not apply for your situation.

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
...
+## Decision
+
+Encapsulating plugins export a function called `deferFunctionCall` with the
+declaration
+`void deferFunctionCall (Plugin * plugin, char * name, KeySet * parameters)`.
+Encapsulating plugins shall save multiple deferred calls and call the exported
+functions specified by `name` passing the `parameters` KeySet when a plugin is
+initialized in the same order as received.
+Encapsulated plugins that do not export a specified function are omitted.
+
```
markus2330: What is required by plugins supporting such deferred function calls?

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
...
+## Constraints
+
+1. Plugins encapsulating other plugins shall be able to lazy-load them.
+2. Callers shall not be entangled in encapsulating plugin details.
+3. The encapsulation of the plugins shall not be broken.
+4. The mechanism shall be generic (not only for I/O bindings).
+
+## Assumptions
+
+1. The called functions do not return a value (e.g. `set`, `open`, `close`, ...)
```
markus2330: Is it possible to communicate in both directions via the KeySet parameter?

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,61 @@
...
+
+The called functions have the declaration
+`void function (Plugin * plugin, KeySet * parameters)`.
+This means that the functions also receive their parameters via a KeySet.
+
+## Related decisions
+
+## Notes
+
+Utility functions that help with managing deferred calls would be nice:
```
markus2330: They absolutely would be nice. What about moving them to elektra-invoke to make elektra-plugin not too bloated?

cmake/Modules/LibAddPlugin.cmake
waht: This is used for the I/O adapters which are required for the plugin (not tests only).

src/plugins/list/list.c
markus2330: As written as comment, plugin.c should stay minimal. Only use it for functions which manipulate the plugin structure.

For the convenience functions elektra-invoke might be suitable?

cmake/Modules/LibAddPlugin.cmake
```diff
@@ -49,9 +49,13 @@ function (add_plugintest testname)
 		restore_variable (${PLUGIN_NAME} ARG_LINK_ELEKTRA)
 		restore_variable (${PLUGIN_NAME} ARG_ADD_TEST)
 		restore_variable (${PLUGIN_NAME} ARG_INSTALL_TEST_DATA)
+		restore_variable (${PLUGIN_NAME} ARG_OBJECT_SOURCES)
+		restore_variable (${PLUGIN_NAME} ARG_TEST_LINK_LIBRARIES)
+		restore_variable (${PLUGIN_NAME} ARG_TEST_LINK_ELEKTRA)
```
waht: @markus2330 I named the new options TEST_LINK_ELEKTRA/LIBRARIES.

doc/decisions/deferred_plugin_calls.md
waht: It is possible but I added that is is not recommended: There is no defined moment where the data should be available and defining such a moment would break lazy-loading.

waht:
Okay. I've improved the decision and moved the new functions to elektra-invoke.

waht:
jenkins build all please

waht:
Obviously the dbus tests fail without the daemon. Shall we wait until the build agents are updated or simply make the tests print a warning and exit normally?

markus2330:
> Shall we wait until the build agents are updated or simply make the tests print a warning and exit normally?

Yes, that is okay.  I even wrote that 22 days ago, see above ("Please succeed with warning. If test cases do not run through, people won't run them.").

waht:
Ah I see :) I will change the tests accordingly.

waht:
jenkins build all please

waht:
jenkins build all please

markus2330:
I'll check the PRs asap. Is there any order I need to merge them?

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,120 @@
...
+  plugins.
+  This approach breaks lazy-loading since for combining all interfaces the
+  plugins have to be loaded and their `getInterfaces` functions have to be
+  called.
+
+## Decision
+
+Encapsulating plugins export a function called `deferredCall` with the
+declaration
+`void ElektraDeferredCall (Plugin * plugin, char * name, KeySet * parameters)`.
```
markus2330: E should be lower case

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,120 @@
...
+
+The solution allows to change encapsulating plugin implementations without
+breaking callers.
+
+## Implications
+
+The called function receive their parameters via a KeySet.
+
+While called functions could return data using the `parameters` KeySet (or a
+seperate KeySet) there is no defined moment when the data can be collected.
```
markus2330: separate

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,120 @@
...
+Encapsulating plugins export a function called `deferredCall` with the
+declaration
+`void ElektraDeferredCall (Plugin * plugin, char * name, KeySet * parameters)`.
+Encapsulating plugins shall save multiple deferred calls and call the exported
+functions specified by `name` passing the `parameters` KeySet when a plugin is
+initialized in the same order as received.
+
+Plugins that support deferred calls shall have the following declaration for
+their functions
+`void ElektraDeferredCallable (Plugin * plugin, KeySet * parameters)`.
```
markus2330: Doesn't the absence of deferredCall already tell? What are the parameters for?

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,120 @@
...
+
+While called functions could return data using the `parameters` KeySet (or a
+seperate KeySet) there is no defined moment when the data can be collected.
+Defining such a moment would break the lazy-loading constraint.
+It is recommended to use callbacks passed as `parameters`.
+The callback function declaration is not affected by this decision.
+
+## Related decisions
+
+- Elektra's invoke functionality will be extended to also allow use of deferred
```
markus2330: allow us to use

doc/decisions/deferred_plugin_calls.md
```diff
@@ -0,0 +1,120 @@
...
+
+The called function receive their parameters via a KeySet.
+
+While called functions could return data using the `parameters` KeySet (or a
+seperate KeySet) there is no defined moment when the data can be collected.
+Defining such a moment would break the lazy-loading constraint.
+It is recommended to use callbacks passed as `parameters`.
+The callback function declaration is not affected by this decision.
+
+## Related decisions
```
markus2330: Is this also already implemented? Why related decision?

src/libs/io/adapter/dbus/dbus.c
```diff
@@ -0,0 +1,252 @@
...
+
+	return priv;
+}
+
+// TODO rename to elektraIoAdapterDbusDetach when complete reversal is possible
+int elektraIoAdapterDbusCleanup (ElektraIoAdapterDbusHandle * priv)
+{
+	DBusConnection * connection = priv->connection;
+
+	// TODO currently not possible because dbus uses multiple watches for the same fd
```
markus2330: todo

src/plugins/list/list.h
markus2330: I could not find the comment below anymore. Can you describe the deps somewhere explicitly?

markus2330:
Thank you, great work!

I added some small comments to be fixed in a later PR.

Issue #1016, opened 2016-10-10
Document dependencies via CONTRACT


markus2330:
@Namoshek wrote:

> What I personally would like to see is some collection of all packages you need to get a full installation running.

Currently, most plugins print a warning when they are excluded which dependency was missing. This solution seems to be not satisfactory (see #1015).

## Should replace

- [ ] info in doc/COMPILE.md
- [ ] lists in .travis.yml
- [ ] list in doc/docker/Dockerfile
- [ ] list in doc/vagrant/provision.sh
- [ ] list in puppet hiera
- [ ] the output some plugins provide when they are excluded because of missing deps (see #1889)

## Should support

- [ ] Debian-based systems
- [ ] MacOSX-based systems
- [ ] Others? (Fedora-based systems, Arch-linux-based systems)

## Open questions

Implement it as CMake or Shell script? (Advantage of CMake: already implements logic for interpreting PLUGINS variable, Disadvantage: adds additional step during installation if cmake not available)

I would propose that we describe deps similar to  #1012: in a way that they are written down decentral but we can collect them easily. The main problem is that packages are called differently across distros and versions. So how can we specify them?

What are your requirements and expectations for documentations of the dependencies?

Namoshek:
The most important distros should be documented (debian, ubuntu, redhat, ...). Ideally you could somehow generate package manager commands (`apt-get install ...`) from the enumeration.

If enumerating all possible variations gets too much (quite likely considering how many distros there are), we should maybe limit supported distros for this specific part of documentation? For technical implementation, I think a space separated list of deps for each supported distro(+version) should be enough.

markus2330:
Result of last discussion:

Let us add dependencies information via the contract. First we have to add the dependencies for some package systems (contributors?). Later we can also add functionality in `kdb info` to output all deps of all plugins.

@reox can you add deps for debian/cent os (rpm)?

reox:
Dependecies for CentOS 7 (got from rpm's spec file):
```
boost-devel cmake3 db4-devel docbook-style-xsl doxygen gawk GConf2-devel gettext-devel libxml2-devel libxslt yajl-devel
```

reox:
Found out, there are more dependencies. This is my current list (using CentOS 7.2 and EPEL Repos):

```
[...]

PR #1866, opened 2018-03-29
Prepare Jenkinsfile

ingwinlu:
# Purpose
Start porting build jobs into Jenkinsfile to improve traceability and maintainability of test runs.

This PR replaces the following build jobs:
* ~~elektra-clang (by debian-stable-full-clang)~~ delayed to next release
* elektra-doc (by doc)
* elektra-gcc (by debian-stable-full)
* ~~elektra-gcc-asan (by debian-stable-asan)~~ delayed to next release
* elektra-gcc-configure-debian (by debian-unstable-full)
* elektra-gcc-configure-debian-debug (by debian-unstable-full)
* elektra-gcc-configure-debian-notest (by debian-unstable-full)
* elektra-gcc-configure-debian-stretch-minimal (by debian-stable-minimal)
* elektra-haskell (by debian-stable-full)
* elektra-mergerequests-fast (by debian-unstable-fast)
* elektra-mergerequests-stable (by debian-oldstable-full)
* elektra-test-bindings (by debian-unstable-full)
* elektra-todo (by todo)
* test-docker (by debian-stable-full, debian-oldstable-full, ubuntu-xenial)

# Checklist

# Expanded Checklist
- [ ] disable old build job after merge to speed up build all
- [ ] toggle build job to run on scm changes


ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
@markus2330 any idea why the make options `KDB_DB_SYTEM` and `KDB_DB_SPEC` get mangled?

[Build log](https://build.libelektra.org/jenkins/job/elektra-jenkinsfile/15/execution/node/67/log/?consoleFull) shows that
```
/home/jenkins/workspace/elektra-jenkinsfile/scripts/configure-debian -DJAVA_AWT_INCLUDE_PATH=/usr/lib/jvm/java-8-openjdk-amd64/include/ -DBUILD_DOCUMENTATION=OFF -DKDB_DB_SYSTEM=/tmp/home/.config/kdb/system -DKDB_DB_SPEC=/tmp/home/.config/kdb/spec /home/jenkins/workspace/elektra-jenkinsfile
```
becomes
```
	Reason: Could not create configuration file "/.config/default.ecf" because No such file or directory
	Mountpoint: user
	Configfile: /.config/default.ecf.9333:1522349296.903782.tmp
```

markus2330:
Btw. I know its against intuition but please try to ask general questions in separate issues. Otherwise PRs sometimes get *really* confusing.

markus2330:
Btw2: Please create issues if you find bugs. It seems like testkdb_allplugins crashes in your job. Crashes caused by the library (and not the test cases themselves) are always bugs. (We would need a stack trace, though.)

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build docker please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
I will probably move from declerative jenkinsfiles to scripted ones.

while I wanted to avoid scripted ones as the declerative option is generally 'nicer to read', it is too restrictive for what I want to do.

ingwinlu:
jenkins build jenkinsfile-scripted please

markus2330:
Btw. I get one mail after the other that we exceed the traffic limits of 500MB/h. Maybe you find a way to avoid recloning the repo for every build? If this is not easily possible, I will change the notification to 1GB/h. But we might exceed the TB we have in a month, which would lead to reduction of our bandwidth for this month. Internal traffic does not count, so cloning from one shared repo on the community server would solve the problem.

ingwinlu:
@markus2330 500mb/h on what server (i guess build?)

Since we have to get the Jenkinsfile for the build every time I always will need to clone. I am already doing shallow clones hence I expected the traffic to be minimal.

After the Jenkinsfile is parsed all the nodes will have to checkout the repo as well (since they are on other hardware and will have to use the files in the repo, again only shallow with depth 1 is used).

Not sure if we can do more.

markus2330:
> 500mb/h on what server (i guess build?)

The community server (i.e. Jenkins and all -mr build agents).

> Since we have to get the Jenkinsfile for the build every time I always will need to clone.

We could clone via a local proxy. But setting up a proxy is obviously not part of your job description. And thus we will move the build jobs from -mr anyway, I simply changed the notification to 1GB.

> I am already doing shallow clones hence I expected the traffic to be minimal.

Great!

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build docker please

ingwinlu:
jenkins build docker please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted[blablubb] please

markus2330:
Build triggered. sha1 is merged.

markus2330:
Build triggered. sha1 is merged.

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile-scripted please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

doc/GIT.md
```diff
@@ -147,13 +144,8 @@ After being added to whitelist you can trigger buildjobs by saying:
...
 * jenkins build [multiconfig-gcc47-cmake-options](https://build.libelektra.org/job/elektra-multiconfig-gcc47-cmake-options/) please
 * jenkins build [source-package-test](https://build.libelektra.org/job/elektra-source-package-test/) please
-* jenkins build [stable](https://build.libelektra.org/job/elektra-mergerequests-stable/) please
-* jenkins build [unstable](https://build.libelektra.org/job/elektra-mergerequests-unstable/) please
 * jenkins build [homepage](https://build.libelektra.org/job/elektra-homepage/) please
-* jenkins build [gcc-configure-debian-stretch-minimal](https://build.libelektra.org/job/elektra-gcc-configure-debian-stretch-minimal/) please
-* jenkins build [gcc-configure-debian-jessie-minimal](https://build.libelektra.org/job/elektra-gcc-configure-debian-jessie-minimal/) please
-* jenkins build [docker](https://build.libelektra.org/job/test-docker/) please
-* jenkins build [haskell](https://build.libelektra.org/job/elektra-haskell/) please
+* jenkins build [jenkinsfile](https://build.libelektra.org/jenkins/job/elektra-jenkinsfile/) please
```
markus2330: Is there no way to trigger individual parts of this job?

doc/docker/debian/jessie/Dockerfile
```diff
@@ -0,0 +1,70 @@
+FROM debian:jessie
```
markus2330: Please move to scripts/docker.

doc/docker/debian/jessie/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+ENV LANG C.UTF-8
+ENV LANGUAGE C.UTF-8
+ENV LC_ALL C.UTF-8
+
+RUN apt-get update && apt-get -y install \
+    curl \
```
markus2330: Is there a way to move all deps to one place? Ideally fixing #1016 and #1886

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,515 @@
...
+// ## Setup log discarder ##
+properties([
+    buildDiscarder(logRotator(numToKeepStr: '60', artifactNumToKeepStr: '60'))
+])
+
+// ## Define Docker images ##
+DOCKER_IMAGES = [
+    sid: [
+        id: docker_image_prefix + 'debian-sid',
+        context: "./doc/docker/debian/sid",
```
markus2330: should be scripts..

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,515 @@
...
+]
+
+BUILD_FLAGS_ALL = [
+    'BINDINGS': 'ALL;EXPERIMENTAL;-DEPRECATED',
+    'PLUGINS': 'ALL;EXPERIMENTAL;-DEPRECATED',
+    'TOOLS': 'ALL'
+]
+
+BUILD_FLAGS_COVERAGE = [
+    'ENABLE_COVERAGE': 'ON'
```
markus2330: For every build job?

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,515 @@
...
+// ## Define Test Flags ##
+TEST_MEM = 'mem'
+TEST_NOKDB = 'nokdb'
+TEST_ALL = 'all'
+TEST_INSTALL = 'install'
+
+/* # Determine what needs to be build #
+target = "light"
+
+matcher = /jenkins build jenkinsfile(?:\[(.*)\])? please/
```
markus2330: seems really useful :100:

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,515 @@
...
+}
+
+def image_id(image) {
+    def t = checksum(image.file)
+    return "${image.id}:${t}"
+}
+
+def checksum(file) {
+    return sh(returnStdout: true,
+              script: "cat $file | sha256sum | dd bs=1 count=8 status=none")
```
markus2330: weird statement: useless cat and why using sha256sum if most of the 256 bits are discarded? crc32 seems to have the desired length.

doc/GIT.md
ingwinlu: No, I implemented it and reverted it as I did not see the point. Build all is fast enough. And if you need to debug an specific issue you can replicate the environment on your local workstation and can make use of incremental builds.

scripts/jenkins/Jenkinsfile
ingwinlu: Can you elaborate? It only enables the coverage when used in the test() functions later (and modifies the job accordingly so usage is recorded)

scripts/jenkins/Jenkinsfile
ingwinlu: I took this statement from the i3wm build pipeline. 

While you are technically correct, I would argue that sha256 bins are distributed on every distribution I know of, while for example that my personal system does not ship crc32 bin's anymore.
Hence this is much less taken because it is perfect and more because it is available everywhere,. The added downside of computation time is complete neglectable.

doc/docker/debian/jessie/Dockerfile
markus2330: Yes please, move all docker files as discussed in #1912

doc/docker/debian/jessie/Dockerfile
markus2330: Yes, but with #1016 we could add an entry for every distribution (+versions of distributions) we want to support.

scripts/jenkins/Jenkinsfile
markus2330: For me some docu was missing. It was unclear if these flags are applied to every build jobs.

So simply move what you said to the docu.

scripts/jenkins/Jenkinsfile
markus2330: Okay, but what is the point of cutting it at all? Is sounds like a dangerous decision to save some bits and have the risk of easily get collisions.

doc/GIT.md
ingwinlu: > Job was aborted (restarts etc.)

if the job was aborted you need to rerun it completely anyway as you will not know if it could have completed the stages completely

> here was spurious error (some network connection failed, v2 segfault, ...)

but then you would not have a run where all results are published together, only the part that you reran.

> The problem can be easy fixed and/or local replication is unwanted, ...

if it can be easily fixed you want to do another build all anyway to see if it did not affect any of the other environments

> Jenkinsfile itself gets updates

that one i don't understand. if the jenkinsfile gets an update you will need to rebuild anything anyway as this is the only way to see what the jenkinsfile is actually doing?

> Always re-triggering everything is an easily avoidable waste of energy. It is imho already done too often.

What I see is the following behaviour: tests get run, some fail. new commit is uploaded, failing tests are rerun, either repeat or, rerun all afterwards. With the simplified setup you are only cutting out the middle part. And actually saving energy because you do the build all right away

if you run tests in an isolated matter then you don't need a ci system. you can just run your tests on each developer machine and be done with it. 

Reactivating the behavior is easy, I tested it and it works exactly as the current one. It is there and just commented out. However:
What will never be possible with this approach is having the current 40 build results all displayed individually like it is right now. one job means one result in the github pr ui (though you can modify the text displayed).

If you want to avoid waste of energy you need to provide clear guidlines on what you want to have tested and how because the current state is that there are tests rerun 10+ times in the too similar environments in different jobs. Which is the way bigger issue.

scripts/jenkins/Jenkinsfile
markus2330: > It needs to be unique enough so that a change to the dockerfile immediately is reflected in the hash itself.

Hashes solve this problem. So crc32 actually is not a good choice.

> The length is long enough to guarantee that while not too long to be cumbersome in logs.

E.g. md5 would do? Hashes are not designed to only use parts of it. Systems that show only parts (like git) have the full shas stored, and only show small parts if the small part is actually unique.

> what exactly would be dangerous here? The only thing you can achive (theoretically) is that the dockerfile is modified and the build system does not initiate a rebuild...

I did not know that, some documentation would help. I was afraid that wrong Dockerfiles are used then (or similar).

I only wondered about the code smell. If it is not important anyway, please keep it as is.

ingwinlu: Any hash function sufficies by definition as small changes in the input need to generate a completely different hash.

Functionality wise it is a quick check to see if the contents of the Dockerfile changes (which is enough in this case as we don't use any other file to generate the images). If it does it triggers a rebuild of the images as the old one can not be found locally or in the registry
https://github.com/ingwinlu/libelektra/blob/92f1cad3919a1135b8215c5dd072e4a1e5aa75c2/scripts/jenkins/Jenkinsfile#L145

I can pad this section with comments as it might not be clear from the functions used

doc/GIT.md
markus2330: > I already had this discussion with myself a week ago so I am happy to actually bring it to paper (or at list some hard drive in the cloud) :D.

Good that I asked then :smile: (Btw. I use github-backup)

> let me state some things that we completely left out of the discussion till now:

I think we are now on the same page, so there is no need to endlessly discuss what might be.
Lets face the fact:

It is a huge change, we both do not know the impact it will have on the community's behavior. So the only scientific way forward is to run an experiment if phrases for partial retrigger are needed or not.

> Hence my proposal moving forward: Keep build selection out for now.

How will we know if people find the partial retriggering useful if it is not available? It is easily possible that someone thinks he/she wants retriggering but after trying it out he/she finds it is useless. (E.g. because artifacts were not published.)

So I propose that we have it enabled and we will discuss in a future meeting if people found it useful. If there is doubt on its usefulness we remove it for simplicity reasons.

scripts/jenkins/Jenkinsfile
markus2330: > Any hash function sufficies by definition as small changes in the input need to generate a completely different hash.

It might be the rationale for hash functions but I could not find something like that in the [RFC for SHA](https://tools.ietf.org/html/rfc4634):

> it is computationally infeasible to (1) find a message that
>  corresponds to a given message digest, or (2) find two different
>  messages that produce the same message digest.

If you say (and document) that it does not matter, it is okay. But the code smells. I could be wrong though, maybe he/she actually proved that this reduction is something sane to do. I heard that it was very difficult to prove that AES256 is more secure than AES128. md5 might be weak in deliberate attacks but at least its properties are known.

Btw. I was already involved in the bug triage where a hash sum started with a file name which caused http requests to be redirected wrongly.

ingwinlu:
jenkins build jenkinsfile[.*fast] please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
@markus2330 I can not figure out how I can modify the status pushed by the pull request builder manually. (the plugin is abandoned and does not play very nice with pipelines). It should be possible but I don't want to dive into it too deep right now.

For now the current state of the PR marks the build as unstable if it is not a 'full build' which translates to a build failure on github. Additionally jenkins pipelines are not supporting the concept of 'unstable stage' yet hence the full build has to be marked as unstable.

but you can now do jenkins build jenkinsfile[REGEX] please where regex will have to regex match the stages names you want to build.

So for a quick rundown:
* `.*` is a full build
* `.*-fast` only fast builds
* `doc` only build doc

Dockerfiles will always be build and never skipped (if there are changes to the dockerfile) to avoid general strangeness.

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

scripts/jenkins/Jenkinsfile
markus2330: Btw. why not simply *use* the full hash and *log* the first 8 bytes.

This might lead to a small confusion in the logs but at least it will not bring us in a situation where docker images are not rebuilt and nobody knows why.

ingwinlu: I will use full length hashes

ingwinlu:
jenkins build jenkinsfile please

doc/GIT.md
```diff
@@ -147,13 +144,10 @@ After being added to whitelist you can trigger buildjobs by saying:
...
-* jenkins build [stable](https://build.libelektra.org/job/elektra-mergerequests-stable/) please
-* jenkins build [unstable](https://build.libelektra.org/job/elektra-mergerequests-unstable/) please
 * jenkins build [homepage](https://build.libelektra.org/job/elektra-homepage/) please
-* jenkins build [gcc-configure-debian-stretch-minimal](https://build.libelektra.org/job/elektra-gcc-configure-debian-stretch-minimal/) please
-* jenkins build [gcc-configure-debian-jessie-minimal](https://build.libelektra.org/job/elektra-gcc-configure-debian-jessie-minimal/) please
-* jenkins build [docker](https://build.libelektra.org/job/test-docker/) please
-* jenkins build [haskell](https://build.libelektra.org/job/elektra-haskell/) please
+* jenkins build [jenkinsfile](https://build.libelektra.org/jenkins/job/elektra-jenkinsfile/)[REGEX] please
+  * Use the optional `REGEX` field to specify which stages you want to run
+    The REGEX has to match the stagename.
```
markus2330: Can you move the deleted jobs as sub-items here? These names can be used for the regex, can't they?

scripts/docker/centos/6/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM centos:6
+
+RUN yum install -y cmake3 git gcc-c++
```
markus2330: Great to have centos, too! :100:

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,538 @@
+// FIXME sloc publish errors
```
markus2330: better to start with some description what this is and not with a TODO list :smile:

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,538 @@
...
+// XXX add missing jobs
+// TODO have a per plugin/binding deps in Dockerfile for easier maintanence
+// TODO add warnings plugins to scan for compiler warnings
+
+// # Imports #
+import java.text.SimpleDateFormat
+
+// # Set Variables #
+CMPVERSION = '0.8.22'
+VERSION = '0.8.23'
```
markus2330: doc/todo/RELEASE should have information when this needs to be updated.

scripts/jenkins/Jenkinsfile
markus2330: still open: please add more docu

Ideally give a reader some guide how to read the Jenkinsfile in the very top and guide through some standard use cases (e.g. how to add a build job)

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,538 @@
...
+TEST_MEM = 'mem'
+TEST_NOKDB = 'nokdb'
+TEST_ALL = 'all'
+TEST_INSTALL = 'install'
+
+// ## DATE ##
+NOW = new Date()
+
+// # Determine what needs to be build #
+target = ".*"
```
markus2330: Does this mean by default everything is build? (The comment directly above is confusing)

scripts/docker/centos/6/Dockerfile
ingwinlu: actually I must have accidentally added the file in a commit. Centos6 has an too old kernel to be dockerized. While I did plan to add them they should not be part of the PR and I will delete it.

scripts/jenkins/Jenkinsfile
ingwinlu: It plans to build everything per default. But if some of the fast builds fail it will not queue the more intensive ones which occur later

scripts/jenkins/Jenkinsfile
ingwinlu: since it is currently not used and was only there as an example i can simply remove it

doc/GIT.md
ingwinlu: no they can not

ingwinlu: ` The REGEX has to match the stagename.`  should indicate that you are to use the stage names of the jenkins file which can be easily read from the webui

scripts/jenkins/Jenkinsfile
ingwinlu: Please add a suggestion on how to archive that. I don't see comments in the c files explaining what a global variable is.

If I add comments for each bit of groovy syntax it will get unreadable very quickly.

doc/GIT.md
markus2330: Can you please add a link to the webui then?

scripts/jenkins/Jenkinsfile
markus2330: For the icheck and the Debian jobs you will need them. But for this PR you can remove them if you prefer so.

ingwinlu:
jenkins build jenkinsfile please

doc/GIT.md
ingwinlu: it is literally there in line 148, even clickable...

scripts/jenkins/Jenkinsfile
ingwinlu: Hopefully we can parse the version numbers from somewhere in the repository?

ingwinlu:
jenkins build jenkinsfile please

scripts/jenkins/Jenkinsfile
markus2330: Why parse? The current version number is in the CMake variable KDB_VERSION.

But the numbers above are not the current version but the version to build (Debian Packages for) and the version to compare to. They differ from KDB_VERSION respective KDB_VERSION-1 during the release process.

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
@markus2330 Renamed some of the variables and functions as discussed yesterday

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+                                     script: 'which llvm-symbolizer').trim()
+            withEnv(["ASAN_OPTIONS='symbolize=1'",
+                     "ASAN_SYMBOLIZER_PATH=${llvm_symbolizer}"]){
+                ctest()
+            }
+        }
+    }
+}
+
+/* Helper to generate a typical elektra test environment
```
markus2330: This documentation is now useful, please also add similar documentation to all other vars and functions.

I think it would be useful if you use doxygen style (in particular @param) even if doxygen cannot parse it. But we are used in reading function docu in that way.

For the bigger picture (how jenkinsfiles and Dockerfiles interact and similar) a separate README.md would be useful. The current intro is not enough to understand how to read this file. The README.md can be done in a later PR.

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+import java.text.SimpleDateFormat
+
+// Buildjob properties
+properties([
+    buildDiscarder(logRotator(numToKeepStr: '60', artifactNumToKeepStr: '60'))
+])
+
+// Globals
+docker_node_label = 'docker'
+docker_image_prefix = 'build-elektra-'
```
markus2330: prefix (or function) for the file name, too?

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+        file: "./scripts/docker/debian/jessie/Dockerfile",
+    ],
+    xenial: [
+        id: docker_image_prefix + 'ubuntu-xenial',
+        context: "./scripts/docker/ubuntu/xenial",
+        file: "./scripts/docker/ubuntu/xenial/Dockerfile",
+    ]
+]
+
+// Define reusable cmake Flag globals
```
markus2330: why do we need globals here? what are they for? when would I need to declare them?

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+    'SITE': '${STAGE_NAME}',
+    'KDB_DB_SYSTEM': '${HOME}/.config/kdb/system',
+    'KDB_DB_SPEC': '${HOME}/.config/kdb/spec',
+    'CMAKE_INSTALL_PREFIX': '${WORKSPACE}/system',
+    'INSTALL_SYSTEM_FILES': 'OFF',
+    'BUILD_DOCUMENTATION': 'OFF'
+]
+
+CMAKE_FLAGS_BUILD_ALL = [
+    'BINDINGS': 'ALL;EXPERIMENTAL;-DEPRECATED',
```
markus2330: EXPERIMENTAL is automatically added if not removed. Why calling it ALL if DEPRECATED is removed?

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+    'PLUGINS': 'ALL;EXPERIMENTAL;-DEPRECATED',
+    'TOOLS': 'ALL'
+]
+
+CMAKE_FLAGS_COVERAGE = ['ENABLE_COVERAGE': 'ON']
+
+CMAKE_FLAGS_JAVA = [
+    'JAVA_AWT_INCLUDE_PATH': '/usr/lib/jvm/java-8-openjdk-amd64/include/'
```
markus2330: Isn't this dependent on the docker image?

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+CMAKE_FLAGS_DEBUG = ['ENABLE_DEBUG': 'ON']
+CMAKE_FLAGS_LOGGER = ['ENABLE_LOGGER': 'ON']
+
+// Define Test Flag globals
+TEST_MEM = 'mem'
+TEST_NOKDB = 'nokdb'
+TEST_ALL = 'all'
+TEST_INSTALL = 'install'
+
+// DATE
```
markus2330: this comment is imho not needed

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+lock('docker-images') {
+    stage("Build docker images") {
+        // makes sure docker images corresponding to the Dockerfiles are
+        //   available to run the tests in
+        parallel generate_docker_images()
+    }
+}
+
+stage("Fast builds") {
+    // quick builds to give fast initial feedback to programmers
```
markus2330: Please move comments up

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+/* Determine what needs to be build
+ *
+ * Execution of this script is triggered by SCM changes or manually by
+ * commenting 'jenkins build Jenkinsfile[REGEX] please. The [REGEX] is
+ * optional but when it exists it allows for specification of which
+ * stages should be build.
+ */
+target = determineBuildTarget()
+
+/* Main Stages
```
markus2330: Put some special markers for the new "sections" (e.g. a full line of *)

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+    parallel generate_elektra_builds_fast()
+}
+
+stage("Full builds") {
+    // tests to reach full coverage
+    parallel generate_elektra_builds_full()
+}
+
+
+// Stage Generators
```
markus2330: What is a stage generator?

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+def generate_docker_images() {
+    def tasks = [:]
+
+    DOCKER_IMAGES.each { id, image ->
+        tasks << build_image(image)
+    }
+    return tasks
+}
+
+def build_image(image) {
```
markus2330: which kind of image? why do we need to build it, if we already pass it as parameter?

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+                        )
+                        i.push()
+                    }
+                }
+            }
+        }
+    }]
+}
+
+def generate_elektra_builds_fast() {
```
markus2330: again, unclear what it does (also below...)

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+            }
+            if(test_mem || test_nokdb || test_all) {
+                xunitUpload()
+            }
+            deleteDir()
+        }
+    }
+}
+
+// Define helper functions
```
markus2330: again, this should be a clear section marker

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,567 @@
...
+def archive(paths) {
```
markus2330: please document...

scripts/jenkins/Jenkinsfile
ingwinlu: a function that generates stages

ingwinlu:
jenkins build jenkinsfile please

scripts/jenkins/Jenkinsfile
ingwinlu: kind of. It is a monkey patch kind of solution anyway as cmake should be able to find it without providing the path explicitly but it does not.

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

scripts/jenkins/Jenkinsfile
ingwinlu: > EXPERIMENTAL is automatically added if not removed.

Pretty sure it wasn't including something hence it is stated explicitly. Deprecated was removed because gi threw test errors.

markus2330: > Pretty sure it wasn't including something hence it is stated explicitly.

That would be a bug, please report it.

> Deprecated was removed because gi threw test errors.

Isn't not installing gobject-introspection the easier way? Or do you have a check if all plugins were actually added? (Would be a useful check...)

scripts/jenkins/Jenkinsfile
markus2330: Yeah, Debian loves shipping CMake files that are unable to find its own files, already happened multiple times (e.g. python or swig3).

Forking all these CMake files, however, puts quite a maintenance burden on us. For Java we actually nearly included it but we decided to avoid forking CMake Modules.

Now the scripts/configure- contains all our workarounds, maybe you want to use them, too?

ingwinlu: there is nothing java related in the configure scripts anymore (only commented out section about setting JAVA_HOME, which I already do in the docker image)

ingwinlu: cmakes unsets the var because it can not find `jawt.h`. The file is not available in stretch https://packages.debian.org/stretch/amd64/openjdk-8-jdk-headless/filelist which seems like a bug as it is available in the i386 package https://packages.debian.org/stretch/i386/openjdk-8-jdk/filelist ....

ingwinlu:
JAVA_AWT_INCLUDE_PATH

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

scripts/jenkins/Jenkinsfile
ingwinlu: I will not continue working  on this because I will have  to redo it in the next PR.

If i want to add something that is similar to the current matrix builds I will need another way to describe possible values of an configuration anyway.

scripts/jenkins/Jenkinsfile
ingwinlu: > EXPERIMENTAL is automatically added if not removed.

https://build.libelektra.org/jenkins/blue/organizations/jenkins/elektra-jenkinsfile/detail/elektra-jenkinsfile/187/pipeline/150

ingwinlu: ah now i remember, it was because of the haskell plugin and has been reported in #1890 and an open pr exists at #1894

scripts/jenkins/Jenkinsfile
ingwinlu: Just because there is an image definition does not mean it has been built yet hence the need to try and build it.

scripts/jenkins/Jenkinsfile
markus2330: Ahh, perfect. Then simply add a TODO that this needs to be removed after  #1894 is merged :smile:

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

scripts/jenkins/Jenkinsfile
markus2330: The questions should be answered in documentation, please :smile:

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
jenkins build jenkinsfile please

ingwinlu:
@markus2330 I renamed the generator functions to more clearly express what they are doing. I also renamed build_image to maybe_build_image to reflect the change that it no longer always tries to build.

hopefully documented enough now.

markus2330:
Thank you, it is much better now. We will see if it is enough when we actually try to extend the jobs. Some README.md for a big picture is certainly still needed (describing what Jenkins and the build agent do).

Please point me now to where the Debian packages fail.

ingwinlu:
I managed to build the packages today. I needed to create a new tag on checkout which was done in the build job configuration to get rid of the issues with the diff. Hopefully will get it done tomorrow.

Issue #1887, opened 2018-04-04
shellrecorder: fail if writing outside of /tests


markus2330:
Although TESTING.md clearly defines that only `/tests/<testname>` is allowed for writing, quite some tutorials, plugins, tests do differently. Following Elektra paths seem to be used:

- [ ] `/tests` see #1816
- [ ] tutorials, examples, among others... in tutorials and plugin examples

markus2330:
As discussed, only `/tests` is allowed.

markus2330:
The next step would be to let the shell recorder fail if data outside of /tests is accessed.

PR #1850, opened 2018-03-21
notification: improved API

waht:
# Purpose

This PR improves the notification API. It adds new types (both C and kdbtypes.h). It also adds macros for generating API functions for new types and also adds C99 definitions to kdbtypes.h.

# Checklist

- [x] commit messages are fine (with references to issues)
- [x] I added unit tests
- [x] I ran all tests and everything went fine
- [x] affected documentation is fixed
- [x] I added code comments, logging, and assertions (see doc/CODING.md)
- [x] release notes are updated (doc/news/_preparation_next_release.md)
- [x] rebase when dbus PR (#1825) is merged into `master`.
- [x] update notification tutorial: explain emergent misbehavior & add guidelines
- [x] update architecture diagram

@markus2330 please review my pull request

Issue #1398, opened 2017-03-02
units-of-measurements plugin


markus2330:
A plugin that gives numbers a more human-friendly syntax would be great.

I am thinking of following features, as specified in meta data:

- prefixing it with 0b and 0x to get binary or hexadecimal base
- postfixing it with a [Metric prefix](https://en.wikipedia.org/wiki/Metric_prefix) such as 20k to get 20000. Or for floats a [normalized number](https://en.wikipedia.org/wiki/Normalized_number), i.e. 2.0e4 (if we decide to support that for floats, see #1388)
- some e-notation or scientific notation, such as 20e3
- postfixing it with arbitrary strings (or SI-units) that are checked for presence. For example, "s" demands that it is 20ks or 20s, but 20 would be invalid then.

For writing out, it restores the same pre and postfix as it was when getting the value. For new values it uses the specified default.

The plugin either transforms to a value, or yields an error if the value cannot be transformed with the given specification.

For example:

```
[key]
type = long
unit/base/#0 = decimal
unit/base/#1 = hexadecimal
unit/notation/#0 = metric
unit/notation/#1 = e
unit/measurement = s
```

decimal and metric would be default, as they are first in the specifications. hexadecimal and e-notation would be allowed (and are also kept this way if it was so before). In any case, the key must be postfixed with s to be valid.

As first step, `0x` support for lcdproc would be enough. @tom-wa Do you think `0b` is needed for lcdproc?

What do you think?

tom-wa:
i haven't seen any binary literals in the configs, and if they'd be needed somewhere i'm not sure if and how they should be interpreted. so let's avoid them for now.

about the rest: seems fine to me.

markus2330:
Then lets start with hex-only and add later more if needed.

Do you also need escaping of strings for INI?

Do we need a transform dispatcher to have the hex and size plugin #1395 present without hassle?

tom-wa:
>Do you also need escaping of strings for INI?

unescaping

>Do we need a transform dispatcher to have the hex and size plugin #1395 present without hassle?

i'm not really sure if we need it for now, but i'd prefer to have it. it's easier to specify the order of operations, you don't need to remount everything it you want to add another plugin, ....

markus2330:
> unescaping

details? For `=`? Only for values or also for names? And metanames, too?

> but i'd prefer to have it. it's easier to specify the order of operations, you don't need to remount everything it you want to add another plugin, ....

[...]

PR #1903, opened 2018-04-12
Hexnumber

kodebach:
# Purpose

Implementation of hexnumber plugin

# Checklist

needs to be checked.

- [x] I added unit tests


src/plugins/hexnumber/README.md
```diff
@@ -0,0 +1,31 @@
+- infos = Information about the template plugin is in keys below
+- infos/author = Klemens Böswirth <k.boeswirth+git@gmail.com>
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/recommends =
+- infos/placements = postgetstorage presetstorage
+- infos/status = nodep experimental unfinished
+- infos/metadata =
+- infos/description = interprets strings with prefix 0x as hexadecimal numbers
```
markus2330: `infos/ordering = type` is needed: you want to be sure that your plugin runs before the type plugin, otherwise the type plugin would reject numbers with `0x..`.

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,275 @@
...
+ * @param key The Key that should be checked.
+ *
+ * @retval #true if the Key's type metadata is #ELEKTRA_HEXNUMBER_META_TYPE
+ * @retval #false otherwise
+ */
+static bool hasHexType (const Key * key)
+{
+	const Key * typeMeta = keyGetMeta (key, "type");
+	return typeMeta && strcmp (keyString (typeMeta), ELEKTRA_HEXNUMBER_META_TYPE) == 0;
+}
```
markus2330: Nice little functions! This makes the code very clear.

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,275 @@
...
+
+/**
+ * Establish the plugin contract and convert all hexadecimal values in the KeySet to decimal.
+ *
+ * @note The plugin will attempt to convert ALL values starting with 0x from hexadecimal into decimal.
```
markus2330: Notes here will not be read by the user of the plugin.

The feature seems to dangerous, see other discussions and #1398.

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,275 @@
...
+		return ELEKTRA_PLUGIN_STATUS_SUCCESS;
+	}
+
+	Key * cur;
+	ksRewind (returned);
+
+	int status = ELEKTRA_PLUGIN_STATUS_NO_UPDATE;
+	while ((cur = ksNext (returned)) != NULL)
+	{
+		if (!keyIsString (cur) || !isHexString (cur))
```
markus2330: please invert the expression, then you get rid of continue.

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,275 @@
...
+ */
+int elektraHexnumberSet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_UNUSED, Key * parentKey ELEKTRA_UNUSED)
+{
+	Key * cur;
+	ksRewind (returned);
+
+	int status = ELEKTRA_PLUGIN_STATUS_NO_UPDATE;
+	while ((cur = ksNext (returned)) != NULL)
+	{
+		if (!keyIsString (cur) || !hasHexType (cur))
```
markus2330: Again, please invert logic.

src/plugins/hexnumber/hexnumber.h
```diff
@@ -0,0 +1,23 @@
...
+ *
+ */
+
+
+#ifndef ELEKTRA_HEXNUMBER_H
+#define ELEKTRA_HEXNUMBER_H
+
+#include <kdbplugin.h>
+
+#define ELEKTRA_HEXNUMBER_META_TYPE "hexnumber"
```
markus2330: You have to separate meta data for internal purposes and for the user to be specified.

1. To remember what you have transformed to transform it back please use meta keys in `internal/<plugin>`
2. For what the user needs to specify, we should discuss this in #1398 first.

For non-internal meta-data please add the specification in doc/METADATA.ini and in the plugin's contract.

src/plugins/hexnumber/testmod_hexnumber.c
```diff
@@ -0,0 +1,81 @@
...
+
+int main (int argc, char ** argv)
+{
+	printf ("HEXNUMBER    TESTS\n");
+	printf ("==================\n\n");
+
+	init (argc, argv);
+
+	test_basics ();
+	test_default ();
```
markus2330: Please check more for errors and the validation parts of this plugin.

src/plugins/hexnumber/README.md
```diff
@@ -0,0 +1,31 @@
+- infos = Information about the template plugin is in keys below
```
sanssecours: Please replace ”template“ with the name of the plugin (“hexnumber“).

markus2330: Thank you!

The `../../scripts/copy-template` approach avoids such problems.

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,275 @@
...
+	// get hex string from key
+	const char * hexValue = keyString (key);
+
+	int errnoSaved = errno;
+
+	// convert hex string to long long int
+	errno = 0;
+	char * endPtr;
+	unsigned long long int value = strtoull (hexValue, &endPtr, 16);
+	if (errno == ERANGE && value == ULLONG_MAX)
```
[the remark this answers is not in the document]
kodebach: the supermacro is part of #1850 which isn't merged yet, so I will leave it for now

markus2330:
Is this ready for review? (remove "work in progress" if it is) Please rebase after the release.

kodebach:
@markus2330 everything you noted in your original review should be fixed now, please review again

I am not removing the "work in progress" tag, because the `testshell_markdown_hexnumber` test still fails, but I am not sure why, because manually running the code in the README.md file works fine.
What exactly does the test do other than executing the code?

markus2330:
The "work in progress" basically only means "I am still working on it, please do not review". If build jobs do not work is a separate problem (except you are still trying to fix the build jobs)

The build server might have less packages and environment variables. Furthermore, it is guaranteed to make a clean build with only the files as pushed in the repo.

The first step to reproduce is to rebuild everything from scratch locally.

sanssecours:
> What exactly does the test do other than executing the code?

Basically the Shell Recorder executes [each line via `sh -c`](https://github.com/ElektraInitiative/libelektra/blob/master/tests/shell/shell_recorder/shell_recorder.sh#L70). To learn more check out [`markdown_shell_recorder.sh`](https://master.libelektra.org/tests/shell/shell_recorder/tutorial_wrapper/markdown_shell_recorder.sh) and [`shell_recorder.sh`](https://github.com/ElektraInitiative/libelektra/blob/master/tests/shell/shell_recorder/shell_recorder.sh).

I executed some of the commands from the ReadMe in my shell. While the following works:

```sh
kdb mount test.ecf user/examples/hexnumber/test hexnumber
kdb set user/examples/hexnumber/test/hex 0x1F
kdb get user/examples/hexnumber/test/hex
#> 0x1F
```

, the commands:

```sh
kdb mount test.ecf user/examples/hexnumber/test hexnumber
kdb set user/examples/hexnumber/test/hex 0x1F

kdb setmeta user/examples/hexnumber/test/hex type int
kdb set user/examples/hexnumber/test/dec 26 # Without this command everything works fine
kdb get user/examples/hexnumber/test/hex # This command should print the value “31”
#> 1
```

produce unexpected output.

By the way:  We decided to only [use the namespace `user/tests`](https://github.com/ElektraInitiative/libelektra/issues/1887) in tests. Please just replace `user/examples` with `user/tests` accordingly.

markus2330:
Thank you for mentioning `/tests`.

Btw. why are you linking to the shell scripts and not the accompanied README.md? How can we improve the README.md?

sanssecours:
> Btw. why are you linking to the shell scripts and not the accompanied README.md?

Because the code tells you exactly how the tests work (including all the warts).

> How can we improve the README.md?

I assume the best way would be to let someone without any knowledge about the (Markdown) Shell Recorder write a Markdown Shell Recorder test. Afterwards this person could then update and extend the ReadMe, adding information that might not be obvious. Another person with extended knowledge of the MSR (either you or Thomas 😊) could then write a review of the extended ReadMe correcting wrong, unclear and outdated information.

kodebach:
I found the bug, everything should be working now.

> The "work in progress" basically only means "I am still working on it, please do not review"
I also do not have the access rights to remove the label.

> How can we improve the README.md?
@sanssecours explanation matches what I understood from the README, so I don't think we need to change anything.

markus2330:
> I also do not have the access rights to remove the label.

You should now have the rights! You were added to ElektraDevelopers

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,439 @@
...
+	int count = 1;
+	const char * values = keyString (key);
+	char * ptr = (char *) values;
+	while (*ptr)
+	{
+		if (*ptr == ';') ++count;
+		++ptr;
+	}
+	*array = elektraCalloc ((count + 1) * sizeof (char *));
+	char * localString = strdup (values);
```
markus2330: please use elektraStrDup (for everything that allocates memory, there is an elektra* method)

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,439 @@
...
+	const Key * typeMeta = keyGetMeta (key, "type");
+	if (!typeMeta)
+	{
+		return false;
+	}
+
+	const char * type = keyString (typeMeta);
+	for (int i = 0; types[i] != NULL; ++i)
+	{
+		if (strcmp (types[i], type) == 0)
```
markus2330: Please use elektraStrCmp

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,439 @@
...
+ * Checks whether a given Key's value is a hexadecimal string.
+ *
+ * @param key The Key that should be checked.
+ *
+ * @retval #true if the Key's value starts with 0x
+ * @retval #false otherwise
+ */
+static bool isHexString (const Key * key)
+{
+	return strncasecmp (keyString (key), "0x", 2) == 0;
```
markus2330: Please use elektraStrCaseCmp

ingwinlu:

src/plugins/hexnumber/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. To enable forced conversion mode set `/force` to any value. In forced conversion mode the plugin converts **ALL** strings starting with 
+   `0x` (or `0X`) into decimal before passing the value on to the rest of Elektra.
+
+    ```
+    sudo kdb mount test.ecf /examples/hexnumber/forced hexnumber /force=1
+    ```
+
+2. The types recognized as integers can be configured. For this purpose specify all types you want to be considered for possible hexadecimal
+   conversion as a semicolon(;)-delimited list. This overwrites the list of default types completely.
```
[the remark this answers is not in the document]
kodebach: I used the same code as the boolean plugin uses for its lists of true/false values.

I think an array would be very cumbersome in this case. If I want to use all the default types plus one extra type `customint`, I would need to mount like this:
`kdb mount file.ecf /path/to/mount hexnumber /accept/type/#0=byte /accept/type/#1=short /accept/type/#2=unsigned_short /accept/type/#3=long /accept/type/#4=unsigned_long /accept/type/#5=long_long /accept/type/#6=unsigned_long_long /accept/type/#7=customint`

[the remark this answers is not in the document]
markus2330: Yes, the boolean plugin should be changed, too :wink: 

As I thought of it, the default types are always present. The user might choose to add more. But that is not the main point.

Usually the mount command is not done imperatively but with an specification. And using specifications you can use file-format specific extensions to specify arrays, e.g. in INI you might write:

```
[path/to/mount]
accept/type=int
accept/type=customint
```

(Unfortunately there is currently no meta-int array in INI and ni does not support arrays but I hope you get the idea how it should be done.) Alternatively, we could also introduce a syntax to specify arrays in `kdb mount` configuration. But we should not introduce parsing mechanisms per plugin.

To make the implementation of plugins sane, you can use the array commands that already exist in libelektra-ease. Extensions in this library are very welcomed (e.g. better iteration over array entries?) :smile:

src/plugins/hexnumber/hexnumber.c
```diff
@@ -251,76 +255,31 @@ static bool hasHexType (const Key * key)
 	return typeMeta && elektraStrCmp (keyString (typeMeta), "1") == 0;
 }
 
-/**
- * copied from boolean/boolean.c
```
markus2330: Thank you so much for getting rid of this! :heart:

src/plugins/hexnumber/README.md
```diff
@@ -0,0 +1,95 @@
...
+- infos/placements = postgetstorage presetstorage
+- infos/status = nodep unittest maintained
+- infos/metadata = unit/base type
+- infos/ordering = type
+- infos/description = converts hexadecimal values into decimal and back
+
+## Introduction
+
+This plugin is used to read configuration files that use hexadecimal values. All "hex-values" (see below) will be 
+converted into decimal when the Elektra reads values from the mounted file. When Elektra writes back to the file the converted values
```
sanssecours: Typo: `the Elektra` → `Elektra`

src/plugins/hexnumber/hexnumber.c
```diff
@@ -0,0 +1,429 @@
...
+		if (keyIsString (cur) && hasHexType (cur))
+		{
+			status |= convertDecToHex (cur, parentKey);
+		}
+	}
+
+	return status;
+}
+
+int elektraHexnumberClose (Plugin * handle ELEKTRA_UNUSED, Key * parentKey ELEKTRA_UNUSED)
```
sanssecours: Please remove `ELEKTRA_UNUSED` after the argument `handle`.

kodebach:
everything should be fixed now

markus2330:
jenkins build all please

src/plugins/hexnumber/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+include (LibAddMacros)
```
sanssecours: Please format this file with the script [`reformat-cmake`](https://github.com/ElektraInitiative/libelektra/blob/master/doc/CODING.md#cmake-guidelines).

markus2330: Thanks for noting, we should enforce that with the buildserver. Isn't this just a matter of installing two tools?

Ideally the build server should directly tell that something like that is wrong.

@ingwinlu What do you think?

sanssecours: > Thanks for noting, we should enforce that with the buildserver.

We already do that in the [Travis Linux builds](https://travis-ci.org/ElektraInitiative/libelektra/jobs/380362048#L3163).

> Isn't this just a matter of installing two tools?

[Yes](https://github.com/ElektraInitiative/libelektra/blob/master/.travis.yml#L139-L140).

ingwinlu: knock yourself out :D

markus2330:
Please add elektraArrayGetStrings to tests/icheck.suppression (It is a newly added API).

I am not sure if some kind of iteration wouldn't be a better choice. Freeing the structure you return needs a loop (potential errors..). And it seems it cannot handle NULL-keys (binary Keys from which the value is a NULL pointer) because what you return needs to be NULL-terminated. And empty arrays might also be a problem?

What about something like elektraKsNextArrayEntry which simply forwards to the next array entry with the internal cursor? If it can be used within the high-level API, it would be ideal.

src/plugins/hexnumber/CMakeLists.txt
kodebach: @sanssecours It seems you need to manually install libyaml as well as [PyYAML](https://github.com/yaml/pyyaml) for the `reformat-cmake` script to work. This should probably be reflected in our documentation. Especially because otherwise the script throws an error for each CMake file and also deletes the contents of all these files.

sanssecours: > @sanssecours It seems you need to manually install libyaml as well as [PyYAML](https://github.com/yaml/pyyaml) for the `reformat-cmake` script to work. 

As far as I know `cmake-format` has no dependencies:

```sh
pip show cmake-format
#> Name: cmake-format
#> Version: 0.3.6
#> Summary: Can format your listfiles so they don't look like crap
#> Home-page: https://github.com/cheshirekow/cmake_format
#> Author: Josh Bialkowski
#> Author-email: josh.bialkowski@gmail.com
#> License: UNKNOWN
#> Location: /Users/rene/.pyenv/versions/3.6.5/lib/python3.6/site-packages
#> Requires:
#> Required-by:
```

. 

> This should probably be reflected in our documentation. Especially because otherwise the script throws an error for each CMake file and also deletes the contents of all these files.

While [`reformat-cmake`](https://github.com/ElektraInitiative/libelektra/blob/master/scripts/reformat-cmake) is certainly not perfect, it checks for the binaries `cmake-format` and `sponge` beforehand. If you did not install one of these tools it should just fail printing a (broken) error message. I tried this myself by uninstalling `cmake-format` and after that uninstalling `moreutils`. Are you sure that the script deleted you CMake files. If so, can you please describe steps to reproduce the problem?

kodebach: > Are you sure that the script deleted you CMake files?

It did not delete the files themselves, just their contents. I think this is because stdout of `cmake-format` is piped back into the CMake file, but stdout is empty because of the error.

> As far as I know cmake-format has no dependencies

I also looked at the `cmake-format` github page and could not find any mention of the dependency. But [here](https://github.com/cheshirekow/cmake_format/blob/2e2aff2abaa5040bb89e6697d15ecbd628d37b07/cmake_format/__main__.py#L101) the module `yaml` is imported, which is the cause of the error. 

The problem could probably be solved by using the `json` or `python` format for the `.cmake-format` file, because these two modules are part of the python standard library.


If you still want me to, I can try to find out how to reproduce the error.

sanssecours: Thank you for the detailed description. I opened pull request #2011, which should fix the problem by checking the exit code of `cmake-format CMakeLists.txt` before it updates any file.

> The problem could probably be solved by using the `json` or `python` format for the `.cmake-format` file, because these two modules are part of the python standard library.

That might even be a better solution, than the one I came up with. Anyway, since I prefer the human readability of YAML over JSON or the Python config style, I did not change the format of the config file.

kodebach:
> Please add elektraArrayGetStrings to tests/icheck.suppression (It is a newly added API).

I decided to remove the function again and use a `KeySet` together with `elektraKsFilter` instead. There should now be no more memory leaks.

markus2330:
jenkins build all please

markus2330:
Thank you, great job: You could store the old string of the hex number into the internal metadata, then you would not lose the capitalization of hex numbers in the configuration files.

Issue #1908, opened 2018-04-14
build: introduce Jenkinsfile.daily


ingwinlu:
[Jenkinsfile.daily](https://build.libelektra.org/jenkins/job/elektra-jenkinsfile-daily/) should be used to do maintenance work on build nodes:

- [x] create/update reference repos
- [x] cleanup docker containers

optional:
- [ ] prepare daily builds

ingwinlu:
I am closing this as daily builds for incrementals are currently not a priority anymore

Issue #1907, opened 2018-04-14
build: Set up reference repos


ingwinlu:
To safe io (network + disc) we should use reference git repos on all jobs.
https://support.cloudbees.com/hc/en-us/articles/115001728812-Using-a-Git-reference-repository

Additionally a 'daily' job to update it could be introduced (`Jenkinsfile.daily` #1908 ).

ingwinlu:
Created https://build.libelektra.org/jenkins/job/elektra-jenkinsfile-daily/, can add to existing jobs via reference repo: `${HOME}/git_mirrors/libelektra`

markus2330:
Thank you!

PR #1910, opened 2018-04-15
Add Jenkinsfile.daily

ingwinlu:
# Purpose

Add the [existing pipeline](https://build.libelektra.org/jenkins/job/elektra-jenkinsfile-daily/) to VCS for better maintanence.

# Checklist

- [x] Add reference repository to all jobs

# post merge checklist

- [ ] Change https://build.libelektra.org/jenkins/job/elektra-jenkinsfile-daily/ to read from repository

### DIFF
--- a/Jenkinsfile.daily
+++ b/Jenkinsfile.daily
@@ -1,0 +1,65 @@
+properties([
+    pipelineTriggers([cron('H H(1-3) * * *')]),
+    buildDiscarder(logRotator(numToKeepStr: '30', artifactNumToKeepStr: '30')),
+    disableConcurrentBuilds()
+])
+
+env.gitmirror_node_label = "gitmirror"
+
+stage ("Maintain Reference Repositories") {
+    def nodes = nodesByLabel label:env.gitmirror_node_label
+    nodes.add("master")
+    println "Update Git Mirrors on all nodes with tag ${env.gitmirror_node_label} + master\n${nodes}"
+    def repos = [
+        [
+            url:"https://github.com/ElektraInitiative/libelektra.git",
+            dest: "libelektra"
+        ]
+    ]
+    parallel maintain_repositories("git_mirrors", nodes, repos)
+}
+
+
+def maintain_repositories(basedir, nodes, repos){
+    def tasks = [:]
+    for(int i=0; i< nodes.size(); i++) {
+        def axisNodeValue = nodes[i]
+        for(int j=0; j<repos.size(); j++){
+            def repo = repos[j]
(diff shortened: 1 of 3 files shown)


ingwinlu:
ref #1907
ref #1908 
ref #160

markus2330:
Okay, if there is more than one Jenkinsfile it is crystal-clear for me that I want these files in a different folder.

I think "scripts/jenkins" is best suited for them, see #1912

Jenkinsfile.daily
```diff
@@ -0,0 +1,65 @@
+properties([
```
markus2330: Can you add some comments for us newbies having no clue about Jenkinsfiles?

Jenkinsfile.daily
```diff
@@ -0,0 +1,65 @@
...
+    pipelineTriggers([cron('H H(1-3) * * *')]),
```
markus2330: 1h sometimes people are working. What about 2-6? or even 3-6?

Jenkinsfile.daily
```diff
@@ -0,0 +1,65 @@
...
+    buildDiscarder(logRotator(numToKeepStr: '30', artifactNumToKeepStr: '30')),
```
markus2330: Is it possible to fetch this from a single place? I assume that will also be needed by other Jenkinsfiles?

Jenkinsfile.daily
```diff
@@ -0,0 +1,65 @@
...
+        [
+            url:"https://github.com/ElektraInitiative/libelektra.git",
+            dest: "libelektra"
+        ]
+    ]
+    parallel maintain_repositories("git_mirrors", nodes, repos)
+}
+
+
+def maintain_repositories(basedir, nodes, repos){
```
markus2330: space before `{` or maybe even `(` (like we use it for C). Please have a quick look at doc/CODING.md and ideally define something for Groovy.

Jenkinsfile.daily
ingwinlu: While there are shared libraries I don't think it makes sense here as those settings are specific to that build job.

Jenkinsfile.daily
ingwinlu: The impact here is minimal for now with under a minute of execution time. It can be moved around when needed later.

markus2330:
Please set "ready to merge" again if everything is finished and all tests passed.

ingwinlu:
tests should not matter as we don't impact any with this change.

markus2330:
Thank you! Would be good to get all tests (and build jobs) fixed for 0.8.23 nevertheless.

PR #1927, opened 2018-04-17
build: Rewrite Jenkinsfile.daily

### LINKED ISSUE #1916: Jenkinsfile.daily - properties
The current implementation has several problems:
* properties: a initial run is needed for picking them up
* not testable in PR's: you can not specificly run it as part of an pr

So I will create a pr and remove the properties settings from Jenkinsfiles and let them stay in the build server configuration.

ingwinlu:
* Remove pipelineTriggers
* Keep stages better seperated
* Split up into more functions
* Add 'Stage2' of #1908

Resolves #1916 

### post merge checklist
- [ ] Setup triggers in job

### DIFF
--- a/scripts/jenkins/Jenkinsfile.daily
+++ b/scripts/jenkins/Jenkinsfile.daily
@@ -2,9 +2,6 @@
 
 // Set properties for buildjob
 properties([
-    // Trigger only at specified cron time
-    pipelineTriggers([cron('H H(1-3) * * *')]),
-
     // Discard all but the latest 30 logs
     buildDiscarder(logRotator(numToKeepStr: '30', artifactNumToKeepStr: '30')),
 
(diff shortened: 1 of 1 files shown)


ingwinlu:
[Full Testrun](https://build.libelektra.org/jenkins/blue/organizations/jenkins/pipeline-playground/detail/pipeline-playground/83/pipeline/237)

markus2330:
Thank you for the improvements!

Is Jenkinsfile.daily still the correct name? It seems like it does not execute daily anymore?

ingwinlu:
@markus2330 what makes you think that?
I just moved the trigger configuration out of the file into the jenkins build job due to #1916. It is a chicken - egg problem unfortunately.

markus2330:
> what makes you think that? 

I thought so because of the removal of the cron line. For example, you could always update the reference repos (on every build).

> I just moved the trigger configuration out of the file into the jenkins build job due to #1916.

Thanks for the clarification!

PR #1947, opened 2018-04-29
kdb-find: initial commit

### LINKED ISSUE #1943: kdb find
Implement find (search) functionality for Elektra keys.

Task: iterate over all keys and print the name of those who match. `-0` feature is required.

Synopsis:

`kdb find <string> [<root>]`

@omnidan any other requirements?

markus2330:
# Purpose

Closes #1943

# Checklist

needs to be checked.

- [X] commit messages are fine ("module: short statement" syntax and refer to issues)
- [X] I added unit tests

@omnidan  please review my pull request

### DIFF
--- a/doc/help/kdb-find.md
+++ b/doc/help/kdb-find.md
@@ -1,0 +1,59 @@
+kdb-find(1) -- Find keys in the key database
+================================
+
+## SYNOPSIS
+
+`kdb find <regex>`
+
+Where `regex` is a regular expression which contains the key to find.
+
+## DESCRIPTION
+
+This command will list the name of all keys that contain `regex`.
+
+## OPTIONS
+
+- `-H`, `--help`:
+  Show the man page.
+- `-V`, `--version`:
+  Print version info.
+- `-p`, `--profile <profile>`:
+  Use a different kdb profile.
+- `-C`, `--color <when>`:
+  Print never/auto(default)/always colored output.
+- `-v`, `--verbose`:
+  Explain what is happening.
+- `-0`, `--null`:
+  Use binary 0 termination.
+
+## EXAMPLES
+
+```sh
+# Backup-and-Restore: /test
+
+# Create the keys we use for the examples
+kdb set /test/test val1
+kdb set /test/test/foo/bar val2
+kdb set /test/test/fizz/buzz fizzbuzz
+kdb set /test/tostfizz val3
+kdb set /test/tust/level lvl
+
+# list all keys containing /test/t[eo]
+kdb find '/test/t[eo]'
+#>user/test/test
+#>user/test/test/fizz/buzz
+#>user/test/test/foo/fizzbar
+#>user/test/tostfizz
+
+# list all keys containing fizz
+kdb find 'fizz'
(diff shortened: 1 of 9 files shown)


sanssecours:
Seems rather strange that some of the Markdown Shell Recorder tests fail on Linux. I modified the code in commit f271a17b. Now `kdb find` also prints a descriptive error message for regex errors. Strangely enough the updated code also [seems to fix the failing Linux builds on Travis](https://travis-ci.org/sanssecours/elektra/builds/373363517).

@markus2330: Please feel free to overwrite or delete commit f271a17b if you do not like it.

markus2330:
Thank you for the fix!

markus2330:
jenkins build all please

markus2330:
@ingwinlu It would be great if clang jobs could be fixed. Is it fixed by #1866? Using clang 5.0 should already fix the problem.

ingwinlu:
i did try switching to clang 5.0 in #1866 but it reported the same errors as with 3.8 so I reverted it.

sanssecours:
jenkins build gcc-configure-debian-debug please

sanssecours:
jenkins build gcc-configure-debian-log please

markus2330:
@sanssecours Thank you for the update!

@omnidan Does this PR do what you want?

sanssecours:
Seems like the step [“Deploy Homepage” failed](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-1947/7/pipeline/):

> docker: Error response from daemon: Conflict. The container name "/elektra-backend" is already in use by container "e7dc071c4d2d0c64e1759cfdd45636dbdee83c691bb960716452b2a3bb82ec01". You have to remove (or rename) that container to be able to reuse that name.

.

ingwinlu:
might have been a race condition with another build as it should not run
for PR's. It is fixed in the Jenkinsfile misc pr.

ingwinlu:
actually the backend was repeatedly restarting and hence could not be removed by the build server. sadly no indication what triggered it.

sanssecours:

omnidan:
@markus2330 could you add a flag to also search for key values containing the string?

markus2330:
The problem is we are running out of flags in kdb suite (flags are shared between all kdb commands). We could always search for values, too. Or do you plan to give the user a choice which parts are considered?

omnidan:
@markus2330 in that case I would always search by values too

markus2330:
Is it okay that we merge it as-is and implement the search-by-value later? Can you extend Web-UI 1.5 then to use it?

I am sure we'll find a good way to encode the search text for keys and values within a single string. One way would be match the regex against a string which has the key and the value concatenated with a null-byte in between. Or we make a query language. The web UI does not need to be concerned about that, you just call `kdb find <string>`.

omnidan:
@markus2330 yes we can do that - for now we could also just apply the same search string to names and values and return a result if either matches (that is how I would implement it in the Web UI)

markus2330:
Thank you! I merged it!

As discussed, we will extend it so that values are considered to later. At the moment it only searches for keys.

PR #1997, opened 2018-05-15
mINI: Require `ccode` Plugin Instead of `code`

sanssecours:
# Purpose

After this update the test `testshell_markdown_mini` should also complete successfully if we

- build Elektra with enabled address sanitizer and
- enable the leak sanitizer (via `ASAN_OPTIONS=detect_leaks=1`)

.

# Checklist


markus2330:
Thank you for fixing it!

We should avoid that a provider name is also a plugin name.

sanssecours:
jenkins build all please

Issue #1902, opened 2018-04-12
haskell: cmake issues


ingwinlu:
## Steps to Reproduce the Problem

Build without internet access or right now (since hackage is down) + haskell binding

## Expected Result

cmake should fail if required packages are not available

## Actual Result

cmake passes and make tries to call cabal during build

## System Information

- Elektra Version: master

## Further Log Files and Output
```
[ 92%] Linking C executable ../../bin/testmod_augeas
Scanning dependencies of target testmod_hosts
[ 92%] Building C object src/plugins_tests/CMakeFiles/testmod_hosts.dir/hosts/testmod_hosts.c.o
[ 92%] Built target testmod_dbusrecv
cabal: Failed to download http://hackage.haskell.org/00-index.tar.gz : HTTP
code 503
src/bindings/haskell/CMakeFiles/c2hs_haskell.dir/build.make:71: recipe for target 'src/bindings/haskell/libHSlibelektra-haskell.so' failed
make[2]: *** [src/bindings/haskell/libHSlibelektra-haskell.so] Error 1
CMakeFiles/Makefile2:17443: recipe for target 'src/bindings/haskell/CMakeFiles/c2hs_haskell.dir/all' failed
make[1]: *** [src/bindings/haskell/CMakeFiles/c2hs_haskell.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
Scanning dependencies of target testmod_ini
[ 92%] Building C object src/plugins_tests/CMakeFiles/testmod_ini.dir/ini/testmod_ini.c.o
```
[Full Log](https://build.libelektra.org/jenkins/job/elektra-jenkinsfile/85/execution/node/114/log/)

markus2330:
Thank you for reporting!

ingwinlu:
@e1528532 any idea why we call `cabal update` [here](https://github.com/ElektraInitiative/libelektra/blob/master/src/bindings/haskell/CMakeLists.txt#L56)?

Any chance to remove that from the build process and instead do it earlier?

ingwinlu:
```

[100%] Building CXX object src/bindings/intercept/env/tests/CMakeFiles/test_context.dir/test_context.cpp.o
[100%] Linking CXX executable ../../../../../bin/test_fork
[100%] Linking CXX executable ../../../../../bin/test_context
[100%] Built target test_fork
[100%] Built target test_context
[100%] Linking CXX executable ../../../../../bin/test_getenv
[...]

PR #2004, opened 2018-05-17
Haskell cmake improvements

e1528532:
# Purpose

This tries to address the issues described in #1902 . The idea is that i have a script which generates me a full cabal install command by scanning the existing cabal files which i can put into the haskell readme and the docker file that builds up a sandbox at some custom location, e.g. the output will look like `cabal install 'containers >=0.4 && <0.6' 'base >=4.7 && <5' 'directory >=1.1 && <1.4' 'process >=1.2 && <1.7' 'filepath >=1.3 && <1.5'`.

The build will then use this sandbox and won't install any dependencies by itself due to cabals '-offline' switch. Therefore, the dependencies have to be set up beforehand like its handled for all the other plugins, e.g. in the dockerfile and only has to be rebuilt if the dependencies change.

Its probably not the best approach still, but i guess it should solve the cmake issue while still being easy enough to use.

# Checklist

needs to be checked.

- [ ] I added unit tests

Still very WIP, but i wanted to share the idea with this PR.


scripts/docker/debian/stretch/Dockerfile
```diff
@@ -98,3 +98,11 @@ USER ${JENKINS_USERID}
...
 RUN git config --global user.email 'Jenkins <autobuilder@libelektra.org>' \
     && git config --global user.name 'Jenkins'
+
+# Handle Haskell dependencies
+WORKDIR /home/jenkins/elektra-cabal-sandbox
+RUN cabal update && \
+cabal sandbox init && \
+cabal install 'containers >=0.4 && <0.6' 'base >=4.7 && <5' 'directory >=1.1 && <1.4' \
+'process >=1.2 && <1.7' 'filepath >=1.3 && <1.5' 'base -any' 'hspec -any' 'QuickCheck -any' && \
+echo 'export HASKELL_SHARED_SANDBOX=/home/jenkins/elektra-cabal-sandbox' >> /etc/bash.bashrc
```
ingwinlu: That is not how docker works.
Use 
```
ENV HASKELL_SHARED_SANDBOX=/home/jenkins/elektra-cabal-sandbox
RUN mkdir -p $HASKELL_SHARED_SANDBOX && cabal init....
```

And do not forget to add it in the other containers that do haskell testing as well.

scripts/docker/debian/stretch/Dockerfile
```diff
@@ -105,5 +105,5 @@ WORKDIR /home/jenkins/elektra-cabal-sandbox
 RUN cabal update && \
 cabal sandbox init && \
 cabal install 'containers >=0.4 && <0.6' 'base >=4.7 && <5' 'directory >=1.1 && <1.4' \
-'process >=1.2 && <1.7' 'filepath >=1.3 && <1.5' 'base -any' 'hspec -any' 'QuickCheck -any' && \
-echo 'export HASKELL_SHARED_SANDBOX=/home/jenkins/elektra-cabal-sandbox' >> /etc/bash.bashrc
+'process >=1.2 && <1.7' 'filepath >=1.3 && <1.5' 'base -any' 'hspec -any' 'QuickCheck -any'
+ENV HASKELL_SHARED_SANDBOX /home/jenkins/elektra-cabal-sandbox
```
ingwinlu: You will need to set ENV before you try to use it.

e1528532: but i don't try to use it here it will be used later on by cmake then

ingwinlu: Ah so it is not a envvar for cabal?

Anyway your workdir instruction is lingering into the finished image so I would suggest you either remove it and set it via a RUN cd ... (faster) or add another WORKDIR instruction at the end (either to / as it was before or /home/jenkins)

ingwinlu: Also, you should probably use the ENV variable to navigate to that directory

ingwinlu:
```
# Handle Haskell dependencies
ENV HASKELL_SHARED_SANDBOX /home/jenkins/elektra-cabal-sandbox
RUN mkdir -p ${HASKELL_SHARED_SANDBOX} \
    && cd ${HASKELL_SHARED_SANDBOX} \
    && cabal update && cabal sandbox init && cabal install ' \
           containers.....'
```

e1528532: no its a envvar for cmake so it uses this sandbox for the build process. Ok will update, thanks for the hints!

e1528532:
Looks good on the build server so far. Can i teach the bashism checker that a .hs file is irrelevant for it?

e1528532:
that last issue is very strange (and of course not reproducible locally in any way). When that test happens in debian-stable-fast it works but then in debian-stable-full it suddenly breaks.

@ingwinlu Does the debian-stable-fast and debian-stable-full builds happen in the same docker image, so its basically like you run it locally 2 times with a few different options?

ingwinlu:
yes but it does not run `testshell_markdown_typechecker` (as it is tagged as `kdbtests`)

```
Reason: GhcException "can't load .so/.DLL for: /home/jenkins/elektra-cabal-sandbox/.cabal-sandbox/lib/x86_64-linux-ghc-8.0.1/libHSlibfa-1.0.8.23-9qpPxaWMWTxHH2zGjl4A3M-ghc8.0.1.so (/home/jenkins/elektra-cabal-sandbox/.cabal-sandbox/lib/x86_64-linux-ghc-8.0.1/libHSlibfa-1.0.8.23-9qpPxaWMWTxHH2zGjl4A3M-ghc8.0.1.so: undefined symbol: stg_newMutVarzh)"
```
Maybe a linking problem?

e1528532:

e1528532:
finally looks good so far! i think the whole haskell build process is now much more mature plus we got around the dependency reinstallation issue. lets wait for travis to finish as well.

e1528532:
https://travis-ci.org/e1528532/libelektra on my travis its already finished successfully

ingwinlu:
The new dependencies are still not reflected in all Dockerfiles.

scripts/docker/debian/sid/Dockerfile
```diff
@@ -78,3 +79,16 @@ RUN useradd \
     jenkins
 
 USER ${JENKINS_USERID}
+
+# Handle Haskell dependencies
+ENV HASKELL_SHARED_SANDBOX /home/jenkins/elektra-cabal-sandbox
+RUN cd /home/jenkins \
```
ingwinlu: Are you sure you don't want to reuse the ENV variable from above here?
```
RUN mkdir -p $HASKELL_SHARED_SANDBOX && cd $HASKELL_SHARED_SANDBOX && cabal bla
```
This would remove the double definition that currently is occuring.

scripts/docker/debian/sid/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN apt-get update && apt-get -y install \
     autotools-dev \
     automake \
     cmake \
+    libcurl3 \
```
ingwinlu: I don't think all mirrors are updated yet and thus some builds for sid are currently failing (go team debian).

However I don't think explicitly requiring it here is the way to go.

scripts/docker/debian/sid/Dockerfile
e1528532: good point, will adjust

scripts/docker/debian/sid/Dockerfile
e1528532: to me it just looked like the quickest solution without much hassling around. But if you think it should work without it, i can remove it again.

ingwinlu: I had the same error when testing the gtest changes an hour ago and a rerun of the buildjob made it pass (hence I assumed faulty mirrors).

e1528532:
sid still fails due to that cmake dependency... lets see if it works tomorrow then

ingwinlu:
curl requries libcurl4, cmake requires libcurl3. But libcurl3 conflicts with libcurl4. Lets see if they sort it out themselves.

e1528532:

e1528532:
@markus2330 next try to get it merged ;)

markus2330:
jenkins build all please

markus2330:
Thank you, great job!

PR #1820, opened 2018-02-18
cmake tests

markus2330:
# Purpose

Playing around with:

-  [WILL_FAIL](https://cmake.org/cmake/help/v3.0/prop_test/WILL_FAIL.html) (not yet done)
- [PASS_REGULAR_EXPRESSION](https://cmake.org/cmake/help/v3.0/prop_test/PASS_REGULAR_EXPRESSION.html) (not yet done)
- prefix tests with /tests
- [CMake has a skipped/not run feature](https://cmake.org/cmake/help/v3.0/prop_test/SKIP_RETURN_CODE.html), which [might be useless, though](https://gitlab.kitware.com/cmake/cmake/issues/13825)

# Checklist

Please only check relevant points.
For docu fixes, spell checking and similar nothing
needs to be checked.

- [ ] commit messages are fine (with references to issues)
- [ ] I added unit tests
- [ ] I ran all tests and everything went fine
- [ ] affected documentation is fixed
- [ ] I added code comments, logging, and assertions (see doc/CODING.md)
- [ ] meta data is updated (e.g. README.md of plugins)
- [ ] release notes are updated (doc/news/_preparation_next_release.md)

PR #2016, opened 2018-05-20
Save MSR Test Data Below `/tests` & Other Minor Fixes

sanssecours:
# Purpose

This PR should address parts of issue #1887.

# Checklist

- [x] I checked all commit messages. The issue reference in commit 891528b should be correct.
- [x] This pull request adds additional Markdown Shell Recorder tests.
- [x] The PR contains an updated version of the release notes.


markus2330:
Wow, impressive! Does this make the /tests changes in #1820 obsolete?

> This PR should address parts of issue #1887.

Which parts are missing?

Btw. did you also consider to simply prefix every path with /tests/testname within shellrecorder? (Yes, it has some disadvantages, e.g., if you want to check what the shellrecorder did within an interactive shell users won't expect that a prefix exists.)

markus2330:
What we could do (without surprising anyone) is to track if any key is set, which is not below /tests/testname (and /elektra) and fail the tests in such cases. Then we could be sure that all tests are fixed and it would also avoid regressions.

sanssecours:
> What we could do (without surprising anyone) is to track if any key is set, which is not below /tests/testname (and /elektra) and fail the tests in such cases.

I added a simple check for write operations outside of `/tests` in commit 30769620. The check seems to work as it showed that the MSR test of the `range` plugin did not save data below `/tests`.

sanssecours:
jenkins build all please

markus2330:
Any idea why the test case fails?

sanssecours:
Oh , I think I know what the problem is. The `file` util is missing…

ingwinlu:
I can confirm that there is no file utility installed in the minimal image.

scripts/docker/debian/stretch/Dockerfile.minimal
```diff
@@ -5,7 +5,7 @@ ENV LANGUAGE C.UTF-8
 ENV LC_ALL C.UTF-8
 
 RUN apt-get update && apt-get -y install \
-    cmake git build-essential curl
+    cmake git build-essential curl file
```
ingwinlu: This kind of defeats the purpose of having a minimal test environment.
The packages listed here are taken straight from COMPILE.md and should be enough to build + run tests on the 'core'.

To make this change make sense we either need to
  * add `file` as a dependency to COMPILE.md or
  * disable the test if `file` is not available
.

sanssecours: I really think `file` should be part of every Unix system. The `file` utility is also [part of the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/file.html). If we really are that picky about this stuff, then we would also have to mention

- `awk`,
- `bash`
- `grep`,
- `find` and
- `sed`

as dependencies of Elektra.

> add `file` as a dependency to COMPILE.md

We do not need the `file` utility to **compile** Elektra.

ingwinlu: then we should probably list those deps for the test helpers in compile.md and explicitly install them in the images to reflect that.

sanssecours: > then we should probably list those deps for the test helpers in compile.md

I added the above tools to `Testing.md`. Hope that works for you.

src/plugins/iconv/README.md
```diff
@@ -41,13 +41,25 @@ are still properly encoded for anyone.
...
-sudo kdb mount iconv.ini /example/iconv ini iconv from=UTF-8,to=ISO-8859-1
-file ~/.config/iconv.ini             # iconv.ini: ISO-8859 text
-kdb get user/example/iconv/a         # converts ISO-8859 to UTF-8
-kdb set user/example/iconv/a öäß     # converts UTF-8 to ISO-8859
-```
+# Mount the file `iconv/iconv.ini` using the `ini` plugin together with `iconv`
+sudo kdb mount "$PWD/src/plugins/iconv/iconv/iconv.ini" system/tests/iconv ini iconv from=UTF-8,to=ISO-8859-1
+
+# Check the file type
+file -b "$PWD/src/plugins/iconv/iconv/iconv.ini"
```
markus2330: Is this the only occurrence where file is used?

Because this is useless, it only checks the content of a file from the repo.

We should also use the file utility to check for the correct encodings after conversions happened.

scripts/docker/debian/stretch/Dockerfile.minimal
markus2330: Thank you for the hints and the discussion! You already found a good solution.

doc/TESTING.md
```diff
@@ -91,10 +91,17 @@ For `make run_all` following development tools enable even more tests:
 
 - The script `checkbashisms` is needed to check for bashism (tests/shell/check_bashisms.sh),
   it is part of `devscripts`.
-- `git` and `clang-reformat-5` (to 7) to check formatting.
+- `git`, `clang-format` (version 5 up to version 7), and [cmake-format](https://github.com/cheshirekow/cmake_format) to check formatting.
 - `pkg-config` must be available (check_external.sh and check_gen.sh).
 - A build environment including gcc (check_gen.sh).
-
+- You also need the following Unix utilities:
```
markus2330: We should also write here, that all but bash are POSIX.

tests/shell/shell_recorder/tutorial_wrapper/markdown_shell_recorder.sh
```diff
@@ -29,6 +29,15 @@ writeBlock()
 	fi
 	COMMAND=$(sed s/sudo\ //g <<< "$COMMAND")
 	while read -r cmd; do
+		match_command='(kdb[ \t]+(set(meta)?|rm)|kdbSet|keySetName)'
```
markus2330: Other variables in this file are written in uppercase. We should introduce shell coding guidelines in doc/CODING.md though.

sanssecours: Yeah, basically we use all kinds of variable styles all over the place inside shell scripts. Since this is already the case, I used one of the most common styles for scripting languages.

src/plugins/iconv/README.md
sanssecours: > Because this is useless,…

I do not think so. I think the command works both as a nice piece of user documentation and it also makes sure, that the test fails

- if someone changes the file encoding of `iconv.ini`, and
- the `iconv` plugin does not change the encoding at all

.

> We should also use the file utility to check for the correct encodings after conversions happened.

I changed the command to check the file type of the mounted file in the latest commit.

tests/shell/shell_recorder/tutorial_wrapper/markdown_shell_recorder.sh
markus2330: Yes, we definitely need some style guide. Breaking the style within a file can be very confusing. (People think these variables are used differently than the other ones.)

sanssecours:
jenkins build all please

sanssecours:

sanssecours:
jenkins build all please

sanssecours:

sanssecours:

markus2330:
Thank you, great job!

PR #2019, opened 2018-05-24
Travis: Use Latest Version of GCC and Clang

sanssecours:
# Purpose

Turns out we only installed recent versions of compilers on most of the Linux build jobs, but did not use them to compile Elektra 😳. This update should fix this problem. The PR also disables `testlib_notification` on ASAN enabled build jobs, since the test [seems to leak memory](https://travis-ci.org/sanssecours/elektra/jobs/383014685#L2197).

# Checklist

- [x] This PR adds a code comment about the disabled test `testlib_notification` to the relevant CMake code.
- [x] The pull request contains an updated version of the release notes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,11 +28,11 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
           packages:
-            - g++-7
+            - g++-8
       env:
         - ASAN=ON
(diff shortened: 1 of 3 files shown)


sanssecours:
jenkins build all please

markus2330:
Thank you for the update/fix!

PR #2028, opened 2018-05-26
Build: Add Google Test as External Project

sanssecours:
# Purpose

This pull request removes the outdated copy of [GTest](https://github.com/google/googletest) from the repository. Instead of the integrated GTest we know use [DownloadProject](https://github.com/Crascit/DownloadProject/blob/master/CMakeLists.txt) to fetch the latest version of Google Test at configuration time.

# Checklist

- [x] I removed documentation concerning the old version of Google Test we included before. If I missed something please comment below. 


ingwinlu:
I am very much against cmake fetching deps.

All requirements should be installed first, cmake should only do checks + config.

ingwinlu:
can we use https://packages.debian.org/stretch/googletest as an alternative?

sanssecours:
> I am very much against cmake fetching deps.

> All requirements should be installed first, cmake should only do checks + config.

While I agree in general, it is not **recommended to install Google Tests** globally:

```sh
brew install googletest
#> Error: No available formula with the name "googletest"
#> Installing gtest system-wide is not recommended; it should be vendored
#> in your projects that use it.
```

. Sometimes it is also not that easy to install a recent version of a package. For example even Debian Unstable does not provide a package for a [recent version of ANTLR 4](https://packages.debian.org/sid/antlr4).

> can we use https://packages.debian.org/stretch/googletest as an alternative?

I guess we could check for an installed version of Google Test beforehand. However, I am not sure if doing that is worth the effort, since the procedure proposed in this PR should work on every OS we support:

- if you installed Git and CMake, and you 
- have a working internet connection the first time you configure the project

.

ingwinlu:
* The download could fail
* we could get blacklisted for wasting io (each cmake call will download a fresh copy, so that makes 10+ downloads for each build)
* Cmake will take longer to finish.
* You open up vulnerabilities if you do not verify what you download
* Depending on what is downloaded it might indicate failures unrelated to the test runs.
* ...

The list goes on, builds are just not reproducible anymore. And are you really sure you want to use master as a download target? That just screams trouble at least use a tag.

We should at least add it to the Dockerfiles so it at least only gets downloaded on changes instead of each testrun.

ingwinlu:
We also currently try to remove 'cmake downloads haskell stuff' as it makes
a ton of problems for haskell builds. Why even bother with creating a new
angle for build problems for such dubious advantages.

If you want to make it easier to maintain write a script that updates gtest
in the repository (and pins it to the new version).

sanssecours:
> - The download could fail.

That is correct. But from the standpoint of the build server that should not be that different from failing to download CMake via `apt-get install cmake`.

> we could get blacklisted for wasting io (each cmake call will download a fresh copy, so that makes 10+ downloads for each build)

The download size should be about 36.7 MiB then (10 · 3.67 MiB). I think GitHub is able to handle that ☺️.

> - Cmake will take longer to finish.

That is not that big of a problem in my opinion, since it takes about 5 seconds more to configure the project with the proposed changes on my computer (≅15s Vs. ≅20s).

> - You open up vulnerabilities if you do not verify what you download

Since we retrieve the data from the official Google Test repository (and a lot other people do that too) I do not think there is that big of a security concern. I could be wrong though. I am certainly not a security expert.

> - Depending on what is downloaded it might indicate failures unrelated to the test runs.

If you talk about a broken download of Google Test causing build failures, I think that is certainly possible, but unlikely.

> And are you really sure you want to use master as a download target? That just screams trouble at least use a tag.

I do not think that is that big of a problem. At least switching from a five year old version of GTest to the current version of Google Test did not cause any problems as far as I can tell.

> We should at least add it to the Dockerfiles so it at least only gets downloaded on changes instead of each testrun.

That sounds like a good idea. How can I do that?

> We also currently try to remove 'cmake downloads haskell stuff' as it makes a ton of problems for haskell builds. 

I do not think downloading some source files into the build directory is comparable to the complex procedure we use to retrieve the various Haskell dependencies.

> Why even bother with creating a new angle for build problems for such dubious advantages.

I do not like keeping old dependencies around in the source tree. I can understand your point of view too. I guess we just have to disagree on this one 😊.

> If you want to make it easier to maintain write a script that updates gtest in the repository (and pins it to the new version).

Thank you for the advice and your other helpful input.

sanssecours:
jenkins build all please

ingwinlu:
> apt-get install cmake

not comparable as it is cached and only reinstalled if invalidated (currently set to each month or when new debian base images are released). That is also as far as I am willing to go with having the sources in the Docker image.

> That sounds like a good idea. How can I do that?

Modify Dockerfiles so google_test is fetched (see the homepage builds for how it is done for cppcms and co). Add an ARG for the version so it is easily modifiable. Add cmake checks to detect the installed google_test and use it.

> The download size should be about 36.7 MiB then (10 · 3.67 MiB).

for each job. maybe look at the % increase in io needed vs now. (if you ignore the haskell part)

> master

http://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/
Nobody will ever tell you to not pin dependencies to a version.

> If you talk about a broken download of Google Test causing build failures, I think that is certainly possible, but unlikely.

that does not matter, it introduces the possibility. that is enough of a reason. Every build should be as reproducible as possible.

sanssecours:
> Modify Dockerfiles so google_test is fetched (see the homepage builds for how it is done for cppcms and co). Add an ARG for the version so it is easily modifiable. Add cmake checks to detect the installed google_test and use it.

Thank you for the informative reply. Downloading of Google Test directly via Docker is not something I am interested in. Sorry.

By the way, since the tests currently fail if we use GCC 4.7 and the current version of Google Test: When will we drop support for GCC 4.7 and GCC 4.8? Those two compilers have some [serious problems when it comes to regex support](https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions).

markus2330:
Thank you for this work and the interesting discussion.

We already had a very similar discussion, in which @pinotree argued that it is not a good idea that we distribute gtest in our source. He had many troubles caused by this decision when packaging Elektra. I assume, he also does not like cmake to download gtest during creating Debian packages. But from development point (which was mainly my argument), having the tests just work  make things easier (for devs).

A lot changed since then: we now have much better testing coverage on the build server, which makes it less important that the unit tests are executed locally. Nevertheless, maybe following setup can make everyone happy:

1. We continue to support GTEST_ROOT and use GTEST_ROOT within docker (like done while building Debian packages). This should address @ingwinlu's concerns.
2. We stop distributing gtest and download it instead (as done by this PR) **only if** GTEST_ROOT is not set. (If GTEST_ROOT is set wrongly, we fail.) This should address @pinotree's additional concerns that it is not a good idea to distribute the source. And it should fix @sanssecours problem that outdated gtest is used. (But as discussed pinned to some reasonable version.) And it still makes it easy for devs to have gtest (without installing a dep and without fiddling around with GTEST_ROOT).

What do you think?

> When will we drop support for GCC 4.7 and GCC 4.8?

As written in the release notes: "Directly after the release, Jessie (oldstable) with gcc 4.8.4 will be the oldest supported platform." There is no plan to remove the support of 4.8.4. So we need to pin to some version of gtest where this is still supported.

sanssecours:
> What do you think?

Sounds like a good compromise to me.

> As written in the release notes: "Directly after the release, Jessie (oldstable) with gcc 4.8.4 will be the oldest supported platform." There is no plan to remove the support of 4.8.4. So we need to pin to some version of gtest where this is still supported.

How  about we just support the current version of GCC on Debian [GCC `4.9.2-2`](https://packages.debian.org/jessie/gcc) instead? This version of GCC seems to support both regular expressions and the current version of Google Test properly.

markus2330:
>  How about we just support the current version of GCC on Debian GCC 4.9.2-2 instead?

This would be certainly possible but is there really a feature in Google Test (from versions not supporting GCC 4.8) we need? I'll reply about regular expressions in #2031.

sanssecours:
> This would be certainly possible but is there really a feature in Google Test (from versions not supporting GCC 4.8) we need?

Not that I know of. Maybe GCC `4.8` will work well in conjunction with the current version of Google Test anyway. Do we have a build job that uses GCC `4.8`? As far as I can tell `debian-oldstable-full` [uses GCC `4.9.2`](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2028/7/pipeline/350#step-475-log-3).

cmake/Modules/DownloadProject.cmake
```diff
@@ -0,0 +1,174 @@
+# ~~~
+# Distributed under the OSI-approved MIT License.  See accompanying
```
markus2330: For all non-BSD licenses we have to update debian/copyright in the Debian branch.

Please do not create a cmake/Modules/LICENSE, it suggests that all cmake modules are MIT licensed, which they are not.

sanssecours:
jenkins build all please

sanssecours:
jenkins build ini please

sanssecours:
jenkins build all please

sanssecours:

sanssecours:

sanssecours:
jenkins build all please

sanssecours:

sanssecours:

markus2330:
@ingwinlu Is this okay for you?

ingwinlu:
yes

markus2330:
Thanks for the great work!

Issue #1951, opened 2018-05-01
web-ui 1.5 leftovers


markus2330:
Small usability things:

- [x] When clicking left next to arrow, a subtree collapses, but does not open. When clicking right you get exactly the opposite.
- [x] ~~edit value of non-leaves does not have a way to abort~~
- [x] ~~also metadata has no way to abort~~

Other small things:

- [x] ~~Why does "run.sh" start webd and not client?~~

- [x] ~~When starting there is some unnecessary output like:~~

```
The project was built assuming it is hosted at the server root.
To override this, specify the homepage in your package.json.
For example, add this to build it for GitHub Pages:
```

Other problems:

- [x] metadata does not go away afterwards (e.g. click on restrict/remove, uncheck restrict/remove, still leaves restrict/remove = 0 as metadata)

omnidan:
> edit value of non-leaves does not have a way to abort
> also metadata has no way to abort

Elektra Web always saves values when the field is changed (this is consistent across the whole application). Instead of "aborting", there is the undo/redo feature which serves the same purpose.

> Why does "run.sh" start webd and not client?

Because webd still serves the client. Even if you run `npm start` in the client directory, it will run webd which serves the client.

omnidan:
should be fixed in #2029

markus2330:
most of the parts were not fixed (or at least not with Firefox ESR), I added them to #2032

Issue #1901, opened 2018-04-12
create live demo of web-ui


markus2330:
The web-ui is now in a quite stable state (only little bugs are open, see #1892) so we could now start to create a live demo.

The deployment script should be very similar to what we already have for our homepage, it is basically:

- compile with -DTOOLS=..;web 
- make install
- kdb run-elektrad &
- kdb run-web &

We obviously need a completely separated container for that, the web-ui basically allows remote execution of kdb commands with the user elektrad is running.

Such a live demo might be one of the best advertisement *iff* it works well. What do you think?

markus2330:
The only realistic plan for this release is to build and host a docker image with a manual build of web-ui.

@omnidan Can you do that?

markus2330:
apiary of the elektrad+web would be also part of this.

markus2330:
@omnidan did you have time to work on this or the video?

omnidan:
I already made a video: https://drive.google.com/open?id=13dSC0ereCQGusrwlIodaDostW6eEes72

I will try building a docker image with the web-ui installed. Is there already a docker image that has libelektra installed? Where would we host the docker image?

markus2330:
Thank you, the video is great! If you happen to do another video sometimes, having a video without students as example would be excellent :smile:  (e.g. editing /etc/hosts)

What is our best option to host the video, so that playing the video works smoothly across browsers? (In the link above two browsers were not able to play it directly.)

I have a nextcloud installation, we could try it there.

Or is it only a matter of the video format it is in?

> Is there already a docker image that has libelektra installed

See scripts/docker for building images.

>  Where would we host the docker image?

We have our own docker registry at https://hub.libelektra.org but as far as I know it is not yet publicly available.  @ingwinlu knows more about that.

omnidan:
@markus2330 I uploaded it to youtube (unlisted), that should work in all browsers: https://youtu.be/lLg9sk6Hx-E

markus2330:
Why unlisted?

omnidan:
@markus2330 I can make it public, I just thought you only wanted to use it for embedding on the website.

markus2330:
Public is fine for me. We can embed it into the next release notes or even post it as comment to the current reactions of the release notes?

markus2330:
What is the standard youtube license?

markus2330:
I think a docker image with a full (and latest) Elektra preinstalled could be very useful for many demo other use cases, too (not only the Web UI).

[...]

PR #2029, opened 2018-05-26
Elektra Web 1.5

omnidan:
- [x] leftovers from #1951 
- [x] array key improvements from usability test
- [x] improved search #1947 
- [x] test with Firefox ESR

@markus2330 please review

---

## Release notes

Many UX improvements from the usability test!

- search completely reworked - it does not act as a filter on already opened keys anymore, and instead searches the whole key database - feedback from the search was also greatly improved (pulsating while searching, glowing blue when done)
- added "abort" buttons to dialogs to revert actions
- added "create array" button to easily create arrays
- removed confirmation dialog before deletion (undo can be used instead)
- created a docker image: `elektra/web`
- small fixes:
  - updated visibility levels
  - removed "done" button in main view
  - fixed issues with the opener click area
  - remove metakeys when they are set to the default value or empty/0


src/tools/web/client/package-lock.json
```diff
@@ -11324,14 +11324,6 @@
         }
       }
     },
-    "react-tap-event-plugin": {
```
markus2330: Could you please write docu how to udpate this file?

src/tools/web/client/src/components/TreeItem/dialogs/SettingsDialog.jsx
```diff
@@ -99,7 +99,12 @@ export default class SettingsDialog extends Component {
       }
       // persist value to kdb and show notification
       const { timeout } = this.state[key] || {}
-      setMeta(key, value)
+
+      // remove metakey when setting to 0
```
markus2330: and empty value?

src/tools/web/README.md
```diff
@@ -153,3 +153,52 @@ authenticate users, e.g. by [username/password auth](https://www.digitalocean.co
...
+  in `client/src/components/TreeItem/dialogs/utils.js`:
+
+```diff
+export const HANDLED_METADATA = [
+  ...,
+  'visibility',
+  'binary',
++ 'check/something',
+]
+```
```
markus2330: And then, how to do validation or similar?

omnidan:
@markus2330 all tasks from #2032 should be done now, I also tested with Firefox ESR.

you can review already, because only the docker container for the live demo is missing now.

markus2330:
The docker container would make it easier to test (and reproduce) :wink:

omnidan:
@markus2330 I created a Dockerfile (see https://github.com/ElektraInitiative/libelektra/issues/1901#issuecomment-393569029). I also released an image, but it is still 1.4.0 (as the Dockerfile builds from the official git repo).

markus2330:
I am okay with merging this PR so that creating the Docker image is easier. If everything is finished, please set "ready to merge".

src/tools/web/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:16.04
```
markus2330: Any reason why this file is not in scripts/docker/?

omnidan:
@markus2330 I already created a `1.5.0-beta` image, no worries. you can install it via:

```
docker run -d -it -p 33333:33333 -p 33334:33334 elektra/web:1.5.0-beta
```

ingwinlu:
why the two ports? you are missing `EXPOSE` entries in the Dockerfile. You might also be interested in removing some layers to reduce image size and speed up builds.

Additionally you don't have a PID1 that controls signals sent to the container.

markus2330:
@omnidan Thank you, works for me. As said, mountpoints would make the image more interesting. (Maybe mount /tmp/hosts to not destroy anything.) What about providing a start-web script instead of creating it only for the docker image? And release notes are missing! You are welcome to highlight your changes.

@ingwinlu Thank you for reviewing the Dockerfile!

markus2330:
@omnidan small things: "edit value" does not automatically focus to the edit field. Please also check if tabs work as wanted (they also cycle in the background of the dialog). And numbering of arrays sometimes does not work (if I clicked somewhere else before.) Should I create an issue for this?

omnidan:
@ingwinlu thanks for the hints! I added the two ports via EXPOSE. One port is for elektrad (33333), the other for webd/client (33334). I also tried reducing the image size by following instructions from: https://blog.florianlopes.io/5-tips-to-reduce-docker-image-size/

I'm not sure what you mean by "you don't have a PID 1" - I found out that nodejs is not supposed to run as PID 1 in docker: https://www.elastic.io/nodejs-as-pid-1-under-docker-images/

omnidan:
@markus2330 

> "edit value" does not automatically focus to the edit field. 

thanks for the hint - I fixed this!

> Please also check if tabs work as wanted (they also cycle in the background of the dialog).

unfortunately trapping the focus in the dialog does not work with the current version of the UI library I use (MUI). I tried using an external library to achieve this, but with the external library I can only trap focus within the dialog contents, which excludes the dialog buttons. so when I enable the focus trap, it is not possible to navigate to the dialog buttons anymore.

focus trapping dialogs are implemented in the latest beta version of MUI, however: https://github.com/mui-org/material-ui/issues/4384 - should I make an issue about things that can be fixed by upgrading to MUI 1.0 when it comes out (selecting text when pressing the copy button, trapping focus in dialogs)?

> And numbering of arrays sometimes does not work (if I clicked somewhere else before.)

Where did you click? I'm not sure how to reproduce this.

omnidan:
@markus2330 

> Thank you, works for me. As said, mountpoints would make the image more interesting. (Maybe mount /tmp/hosts to not destroy anything.)

We could push an example kdb export to a git repo and then pull that to build the docker image for the live demo (I would create a separate image for that though).

> And release notes are missing! You are welcome to highlight your changes.

I added them to the initial post of this PR!

markus2330:
Thank you look great.

But please describe changes in changelog (news). (your text in this PR is quite good, but it should be in doc/news/_preparation_next_release.md)

You can also finish (some) tasks of #2037 here.

markus2330:
jenkins build all please

markus2330:
Please add some info to the release notes.

markus2330:
Thank you. Seems like you need to rebase.

omnidan:
@markus2330 done!

markus2330:
Thank you! Could you please rebase to get rid of the old commits.

@ingwinlu Can you take a look at the docker file?

omnidan:
@markus2330 I already rebased, what do you mean "get rid of the old commits"?

markus2330:
The github view shows 250+ commits. Something seems to be wrong here.

markus2330:
And maybe it is a good idea to split the docker files and the WebUI changes.

omnidan:
> The github view shows 250+ commits. Something seems to be wrong here.

@markus2330 that is because I rebased and it added all commits from upstream master to the branch, they should not be duplicated after merging.

omnidan:
> And maybe it is a good idea to split the docker files and the WebUI changes.

I will remove them from this PR for now and create a new PR with the updated docker files tomorrow.

markus2330:
There is still something very wrong in this PR. There are 250+ commits.

omnidan:
@markus2330 as I said, that is a result of rebasing, it happened in the previous PR too. Look at the "Files changed" tab, it does not actually contain the changes of the commits.

omnidan:
see #2099

ingwinlu:
@omnidan out of curiosity: how do you rebase?

markus2330:
> it happened in the previous PR too.

This does not say it was correct. It also messed up the history back then.

The problem is that the common commit is old (26.05) and that you rebased something that already contained a merge commit.

A rebase should be very simple in your case, nobody else modified the directories you are working in.

You can also create a new PR if it is easier for you.

omnidan:
@ingwinlu https://help.github.com/articles/syncing-a-fork/ and then `git rebase master`

markus2330:
I do not think you should mix these two.

You can either merge with `-ff` or even `-ff-only` or do a pull with `--rebase`.

ingwinlu:
Make sure upstream is really pointing at the correct repository via `git remote -v`.
Your process seems fine to me otherwise.

omnidan:
@ingwinlu I read that it might be the way I created the branch, not the way I rebased: https://stackoverflow.com/a/40413455

PR #2031, opened 2018-05-27
KDB Find: Exclude Command on GCC 4.8 and Before

sanssecours:
# Purpose

Since GCC 4.7 and GCC 4.8 ship with [broken support for regular expressions](https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions) 😭, we exclude the tool `kdb find` if we detect these compiler versions.

This pull request should fix the broken build jobs

- [`elektra-gcc-configure-debian-wheezy`](https://build.libelektra.org/jenkins/blue/organizations/jenkins/elektra-gcc-configure-debian-wheezy/detail/elektra-gcc-configure-debian-wheezy/1104/pipeline)  
- [`elektra-gcc47-all`](https://build.libelektra.org/jenkins/blue/organizations/jenkins/elektra-gcc47-all/detail/elektra-gcc47-all/1357/pipeline)

.

# Checklist

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -193,6 +193,7 @@ Many problems were resolved with the following fixes:
   when checking if the ABI changed. *(Lukas Winkler)*
 - You can now build the [Qt-GUI](https://www.libelektra.org/tools/qt-gui) using Qt `5.11`. *(René Schwaiger)*
 - We fixed various minor spelling mistakes in the documentation. *(René Schwaiger)*
+- We exclude the command `kdb find`, if the we know that the current compiler ships with broken support for regular expressions. *(René Schwaiger)*
 
 
 ## Outlook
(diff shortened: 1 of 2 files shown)


sanssecours:

markus2330:
We do not support Wheezy anymore, I disabled both jobs.

For Jessie (afaik) the unit tests pass. Even if regex support might be broken, finding keys by text is also useful without regular expressions. So completely disabling the feature might be too much.

We could document that regex support on platforms which have broken regex implementations is broken. But listing the platforms is difficult and the statement itself is obvious, isn't it?

sanssecours:
> We do not support Wheezy anymore, I disabled both jobs.

Thank you!

> But listing the platforms is difficult and the statement itself is obvious, isn't it?

Yep.

PR #2008, opened 2018-05-17
crypto: stop gpg-agents after unit test

petermax2:
# Purpose

This PR affects unit tests of `crypto` and `fcrypt`.

The general idea is to:

1. set `GNUPGHOME` to a temporary directory,
2. delete the temporary directory and all of its contents as last step of the unit tests, and
3. shut down `gpg-agent` when the unit tests are done.

Step 2 instructs `gpg-agent` to shut down. A big thank you to @ingwinlu for finding out how to properly shut down the `gpg-agent`.

Closes #1928 .

# Checklist

Check relevant points but please do not remove entries.
For docu fixes, spell checking, and similar nothing
needs to be checked.

- [x] commit messages are fine ("module: short statement" syntax and refer to issues)
- [x] I added unit tests
- [x] I ran all tests locally and everything went fine
- [ ] affected documentation is fixed
- [x] I added code comments, logging, and assertions (see doc/CODING.md)
- [ ] meta data is updated (e.g. README.md of plugins)
- [ ] release notes are updated (doc/news/_preparation_next_release.md)

# TODOs

- [x] improve code quality
- [ ] adapt `fcrypt` unit test

Issue #1973, opened 2018-05-11
Tests behave oddly in certain environments (debian packages)


ingwinlu:
Before bc804c207df391c339d880d3487ea3c16eb4ee4e the following tests failed in some environments when building debian packages:
* testmod_blockresolver
* testmod_crypto_openssl
* testmod_crypto_gcrypt
* testmod_crypto_botan
* testmod_fcrypt

EDIT markus: You can also reproduce the blockresolver and fcrypt problems by manipulating HOME and TMPDIR. No building of a package is needed.

### blockresolver
```
[package/stretch]  42/118 Test  #48: testmod_blockresolver ............***Failed    0.00 sec
[package/stretch] BLOCKRESOLVER     TESTS
[package/stretch] ==================
[package/stretch] 
[package/stretch] /home/jenkins/workspace/elektra-jenkinsfile/libelektra/src/plugins/blockresolver/testmod_blockresolver.c:70: error in test_BlockresolverWrite: blockresolver->kdbSet failed
[package/stretch] Compare <key = only the inside has changed
[package/stretch] >, with <key = inside block
[package/stretch] >
[package/stretch] in file /home/jenkins/workspace/elektra-jenkinsfile/libelektra/src/plugins/blockresolver/blockresolver/compare.block, line 5.
[package/stretch] /home/jenkins/workspace/elektra-jenkinsfile/libelektra/tests/cframework/tests.c:157: error in compare_line_files: comparing lines failed
[package/stretch] /home/jenkins/workspace/elektra-jenkinsfile/libelektra/src/plugins/blockresolver/testmod_blockresolver.c:72: error in test_BlockresolverWrite: files do not match as expected
[package/stretch] 
[package/stretch] testmod_blockresolver Results: 15 Tests done — 3 errors.
```

### crypto_openssl
```
[package/stretch]  48/118 Test  #60: testmod_crypto_openssl ...........***Failed    5.32 sec
[package/stretch] CYPTO        TESTS
[package/stretch] ==================
[package/stretch] 
[package/stretch] /home/jenkins/workspace/elektra-jenkinsfile/libelektra/src/plugins/crypto/test_internals.h:263: error in test_gpg: failed to install the GPG test key
[package/stretch] 
[package/stretch] crypto_openssl Results: 65 Tests done — 1 error.
```

### crypto_gcrypt
```
[package/stretch]  49/118 Test  #61: testmod_crypto_gcrypt ............***Failed    5.18 sec
[...]

Issue #896, opened 2016-08-30
crypto plugin: gpgme plugin variant


petermax2:
# Feature

The `gpgme` plugin is created to solely uses the `gpgme` library for crypto operations.

petermax2:
This is a follow-up task to #892 .

petermax2:
We have two options here:

1. We use the master password and call the gpg binary to perform symmetric encryption with the regular master password.
2. We do not generate and use a master password for crypto_gpg but instead we call the "default" --encrypt / --decrypt option of gpg.

I would prefer number 1 because of better comparability to the other compilation variants. However, number 2 might be cleaner and more secure (no passwords are written over the pipes).

@markus2330 What is your opinion on this matter?

markus2330:
You are right that a plugin variant `crypto_gpg` doing something completely different than the other variants would be strange. Do you really need yet another comparison? 3 variants should be enough? I do not think that we need a benchmark to show us that executing processes takes longer than library calls. Describing a full setup (case study) with getenv would be much more interesting!

What about having a separate plugin `gpg` that uses `--encrypt` and `--decrypt` or even `--sign`?

petermax2:
> I do not think that we need a benchmark to show us that executing processes takes longer than library calls.

Agreed.

> What about having a separate plugin gpg that uses --encrypt and --decrypt or even --sign?

I was thinking about separate plugins for encryption/decryption and signing. Like having sign_gcrypt, fsign (or fverify, not sure about the name yet). I would prefer to implement the signature operations outside the scope of my Bachelor's Thesis.

PR #2003, opened 2018-05-16
Jenkinsfile misc

ingwinlu:
# Purpose

* Single style for functions in Jenkinsfile d7e5fac
* abort previous run if it is still in the build system a23e7a5
* Rename Fast builds -> Main builds 4e6f064
* Only deploy homepage when building for master a73f783
* Rework Docker Image handling 776d645
* Re Enable clang builds on clang-5.0, and clang/asan builds + disable any existing tests that fail
* add doc/BUILDSERVER.md

closes #1962 
closes #2001 
closes #2034 

# Checklist

- [x] commit messages are fine ("module: short statement" syntax and refer to issues)
- [x] affected documentation is fixed
- [x] I added code comments, logging, and assertions (see doc/CODING.md)
- [x] release notes are updated (doc/news/_preparation_next_release.md)

PR #2035, opened 2018-05-31
gpgme: working gpgme unit test prototype

petermax2:
Affects #896 , #1973 , #2008

# Purpose

This is just a unit test that is supposed to run on the build server.

**PLEASE DO NOT MERGE!!**

# Checklist

needs to be checked.

- [x] I added unit tests


markus2330:
That was really quick! Thank you for testing this!

petermax2:
I have a running unit test that creates a 1024 bit RSA key by utilizing `gpgme`. Let's see how the build server responds.

petermax2:
jenkins build Dockerfile please

petermax2:

petermax2:
jenkins build clang please

petermax2:
@ingwinlu 

    Trying to acquire lock on [docker-images]
    Found 0 available resource(s). Waiting for correct amount: 1.
    [docker-images] is locked, waiting...

I may have broken something. Sorry!

ingwinlu:
That is on purpose. Only a single build is allowed to create docker images at a time (to avoid clogging the build system). This is further refined in #2003. I guess you had old builds running and hence the new ones were blocked. This is also improved in #2003 were new builds auto cancel the old ones.

There should be debian packages on sid and stretch for gpgme-dev: https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=libgpgme-dev

What other questions did you have?

petermax2:
> What other questions did you have?

Everything works :smiley: Thank you four your quick response!

markus2330:
#2003 is merged now.

@ingwinlu Does @petermax2 now need to rebase or will the changes take effect immediately?

petermax2:

petermax2:
So we have a working key generation. GPGME created three 512 bit RSA private keys with sub-keys. :tada: 

    139/223 Test  #99: testmod_gpgme ............................   Passed   96.54 sec

I wanted to repeat the test more often, but ran into a time-out.

So far, `gpgme` looks to be promising. Maybe it would be worth the effort to replace my `gpg.c` module with `gpgme`?

ingwinlu:
key generation is hard on the build server because there is only very
little entropy

markus2330:
An interesting case would be where the current test cases fail and only gpgme succeeds. Maybe you can run all crypto test cases in a loop until they fail? (`while ctest -R crypt; do; done`)

And afaik key generation isn't something our plugins do, so I would not test that.

petermax2:
I added the test case, which fails on the build server, which is: importing the Elektra test key. I repeat the test case 20 times. Let's see how the build server responds.

markus2330:
Unfortunately the build server is down for the weekend.

I think we need to repeat this more than 20 times. Afaik in all build jobs of the last weeks the failure happened only once. And I think you need to repeat the whole test case (with ctest) so that actually a new TMPDIR/HOME is used.

src/plugins/gpgme/gpgme.c
```diff
@@ -0,0 +1,194 @@
...
+	return 1;
+}
+
+int elektraGpgmeOpen (ELEKTRA_UNUSED Plugin * handle, ELEKTRA_UNUSED Key * errorKey)
+{
+	gpgme_error_t err;
+
+	gpgme_check_version (NULL);
+	setlocale (LC_ALL, "");
+	gpgme_set_locale (NULL, LC_CTYPE, setlocale (LC_CTYPE, NULL));
```
markus2330: We should not globally change anything (and especially not locales, this might make applications dysfunctional). If we set it, it should only be done for the gpgme context but why set it at all?

It looks like gpgme is engineered properly, as far as I saw it everything seems to be operating only on its context (and never globally, except in gpgme_set_locale if a NULL pointer is passed).

It might be useful to allocate the gpgme context only once (in open), it might be faster then.

ingwinlu:
The build server is currently configured to kill build jobs after a certain time of inactivity on stdout. This is to avoid zombie jobs.

The current key generation takes long enough without producing output that this policy is enforced.

src/plugins/gpgme/gpgme.c
petermax2: The `gpgme` manual recommends to set the locale for initializing `gpgme`. I am not sure why they recommend it.

ingwinlu: It's fine for prototyping but probably should not be used like this in any released software.
What stands against creating a gpgme context to solve the issue (and not have to rely on maybe wonky locale)?

markus2330: > What stands against creating a gpgme context to solve the issue (and not have to rely on maybe wonky locale)?

It would lead to yet another configuration setting of the plugin. If the locales are broken, pinentries shown by other applications will be broken, too, won't they?

But if we set the locales for the context is a minor thing. I only wanted to note that we must not set it globally.

petermax2:
I close this PR for now and continue the work later on.

PR #2041, opened 2018-06-04
build: Add -Werror to all builds

ingwinlu:
# Purpose

This PR enables -werror on all builds. It does not aim to actually implement all changes needed for builds to pass.

# Checklist

needs to be checked.

- [ ] I added unit tests


markus2330:
Thank you for working on this! It is already a big progress to add `-Werror` to the jobs that currently do not have warnings.

src/plugins/resolver/resolver.c
```diff
@@ -913,8 +913,6 @@ static void elektraUpdateFileTime (resolverHandle * pk, int fd, Key * parentKey)
 		ELEKTRA_ADD_WARNINGF (99, parentKey, "Could not update time stamp of \"%s\", because %s",
 				      fd == pk->fd ? pk->filename : pk->tempfile, strerror (errno));
 	}
-#else
-#warning futimens/futimes not defined
```
markus2330: Maybe this should even be an error.  The resolver plugin does not do what it promises if it cannot update file time.

If it is an error, the error that `#warning` is a GCC extension should be gone, too.

ingwinlu:
You are welcome to push fixes to this branch that tackle any problems with
the build.

markus2330:
As said, I think the first step is to only activate `-Werror` where currently are no warnings+document how to add it for other jobs. debian-stable-full has quite many warnings, it is clearly out-of-scope for this PR to fix them. One need to reproduce it locally to fix all of them.

I commented on the particular change because it is a potentially dangerous change. On systems not having futimens/futimes people might think resolver will work while in fact it does not.

`#error` might fix the problem you tried to solve and is not dangerous. Obviously, the correct fix actually is checking for futimens/futimes within CMake but this again is out of scope.

src/plugins/resolver/resolver.c
ingwinlu: even with the current implemntation i get warnings because apparently the function is undefined for the build.
No idea if that is a c preprocessor thing or if we really do not have the required libraries/settings.

cmake/ElektraCompiling.cmake
```diff
@@ -142,6 +142,7 @@ set (COMMON_FLAGS "${COMMON_FLAGS} -Wformat -Wformat-security")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wshadow")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wcomments -Wtrigraphs -Wundef")
 set (COMMON_FLAGS "${COMMON_FLAGS} -Wuninitialized -Winit-self")
+set (COMMON_FLAGS "${COMMON_FLAGS} -Werror")
```
markus2330: We cannot add `-Werror` here. For production builds it is always useful to *not* have `-Werror`. We can never know if a later version of a compiler will throw additional warnings. We do not want to fail in such situations.

src/plugins/resolver/resolver.c
markus2330: The resolver is something which almost certainly will never compile with c99. Use wresolver instead (see my cmake command).

ingwinlu:
closed in favor of whatever comes from #2092

PR #2046, opened 2018-06-05
Add static and full builds

ingwinlu:
# Purpose

Add statically and fully linked builds.

# Checklist

### DIFF
--- a/doc/BUILDSERVER.md
+++ b/doc/BUILDSERVER.md
@@ -92,7 +92,6 @@ build:
 * jenkins build [gcc-configure-debian-shared](https://build.libelektra.org/job/elektra-gcc-configure-debian-shared/) please
 * jenkins build [gcc-configure-debian-withspace](https://build.libelektra.org/job/elektra-gcc-configure-debian-withspace/) please
 * jenkins build [git-buildpackage-jessie](https://build.libelektra.org/job/elektra-git-buildpackage-jessie/) please
-* jenkins build [git-buildpackage-wheezy](https://build.libelektra.org/job/elektra-git-buildpackage-wheezy/) please
 * jenkins build [icc](https://build.libelektra.org/job/elektra-icc/) please
 * jenkins build [libelektra](https://build.libelektra.org/jenkins/job/libelektra/) please
 * jenkins build [local-installation](https://build.libelektra.org/job/elektra-local-installation/) please
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -249,6 +249,7 @@ Thanks to Michael Zronek and Vanessa Kos.
   - Abort of previously queued but unfinished runs on new commits.
   - Document how to locally replicate the Docker environment used for tests.
 - The Jenkins build server now also compiles and tests Elektra with enabled address sanitizer. *(Lukas Winkler)*
+- Add `STATIC` and `FULL` linked builds. *(Lukas Winkler)*
 - Ported GCC ASAN build job to new build system *(René Schwaiger + Lukas Winkler)*
 - Docker artifacts are now cleaned up in our daily build job. *(Lukas Winkler)*
 - `clang` tests have been ported to the new build system *(Lukas Winkler et al)*
--- a/scripts/jenkins/Jenkinsfile
+++ b/scripts/jenkins/Jenkinsfile
@@ -89,6 +89,10 @@ CMAKE_FLAGS_DEBUG = ['ENABLE_DEBUG': 'ON']
 CMAKE_FLAGS_LOGGER = ['ENABLE_LOGGER': 'ON']
 CMAKE_FLAGS_OPTIMIZATIONS_OFF = ['ENABLE_OPTIMIZATIONS': 'OFF']
 
+// CMAKE_FLAGS_BUILD_SHARED = ['BUILD_SHARED': 'ON']  is ON per default
+CMAKE_FLAGS_BUILD_FULL = ['BUILD_FULL': 'ON']
+CMAKE_FLAGS_BUILD_STATIC = ['BUILD_STATIC': 'ON']
+
 // Define TEST enum used in buildAndTest helper
 enum TEST {
     MEM, NOKDB, ALL, INSTALL
@@ -299,7 +303,10 @@ def generateMainBuildStages() {
         "debian-stable-full",
         DOCKER_IMAGES.stretch,
         CMAKE_FLAGS_BUILD_ALL +
-            CMAKE_FLAGS_COVERAGE,
+            CMAKE_FLAGS_BUILD_FULL +
+            CMAKE_FLAGS_BUILD_STATIC +
+            CMAKE_FLAGS_COVERAGE
+        ,
         [TEST.ALL, TEST.MEM, TEST.INSTALL]
     )
     tasks << buildIcheck()
@@ -348,7 +355,8 @@ def generateFullBuildStages() {
         "debian-unstable-full-clang",
         DOCKER_IMAGES.sid,
         CMAKE_FLAGS_BUILD_ALL +
-            CMAKE_FLAGS_CLANG,
+            CMAKE_FLAGS_CLANG
+        ,
         [TEST.ALL, TEST.MEM, TEST.INSTALL]
     )
     tasks << buildAndTestMingwW64()
--- a/tests/shell/check_kdb_internal_check.sh
+++ b/tests/shell/check_kdb_internal_check.sh
@@ -16,8 +16,7 @@ cleanup()
 ACTUAL_PLUGINS=$PLUGINS
 # Otherwise the test would fail as SHARED_ONLY plugins are not
 # available in full and static builds
-echo "$KDB" | grep -qE "(full|static)\$"
-if [ $? == "0" ]; then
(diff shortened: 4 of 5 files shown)


ingwinlu:
@e1528532 seems like some of the haskell builds need check for static + full linked builds.
```
ls: cannot access '/home/jenkins/.cabal/lib': No such file or directory
-- Found PythonInterp: /usr/bin/python3 (found version "3.5.3") 
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Configuring done
CMake Error at cmake/Modules/LibAddPlugin.cmake:447 (add_dependencies):
  The dependency target "c2hs_haskell" of target "elektra-haskell" does not
  exist.
  cmake/Modules/LibAddHaskellPlugin.cmake:333 (add_plugin)
  src/plugins/haskell/CMakeLists.txt:3 (add_haskell_plugin)

CMake Error at cmake/Modules/LibAddPlugin.cmake:407 (add_dependencies):
  The dependency target "c2hs_haskell" of target "elektra-haskell-objects"
  does not exist.
  cmake/Modules/LibAddHaskellPlugin.cmake:333 (add_plugin)
  src/plugins/haskell/CMakeLists.txt:3 (add_haskell_plugin)

CMake Error at cmake/Modules/LibAddPlugin.cmake:407 (add_dependencies):
  The dependency target "c2hs_haskell" of target
  "elektra-typechecker-objects" does not exist.
  cmake/Modules/LibAddHaskellPlugin.cmake:333 (add_plugin)
  src/plugins/typechecker/CMakeLists.txt:14 (add_haskell_plugin)

CMake Error at cmake/Modules/LibAddPlugin.cmake:447 (add_dependencies):
  The dependency target "c2hs_haskell" of target "elektra-typechecker" does
  not exist.
  cmake/Modules/LibAddHaskellPlugin.cmake:333 (add_plugin)
  src/plugins/typechecker/CMakeLists.txt:14 (add_haskell_plugin)
```

e1528532:
I see. static builds are not supported currently but i have never tried it so i guess i have overseen something over time or introduced a regression at some point. i'll take care about it

ingwinlu:
Don't worry. The existing build did not build all bindings, plugins, libs so that is why it has never shown up.

markus2330:
Doesn't https://build.libelektra.org/jenkins/job/elektra-multiconfig-gcc47-cmake-options/ include all these options in all variants already? Or is this being replicated in this PR?

ingwinlu:
multiconfig-gcc47 will never be replaced. It has been disabled for quite some time:
```
Disabled because gcc47 was available on wheezy and the slave is not powerful enough to run the full matrix in a reasonable timeframe.
```
I was aiming to replace `elektra-mergerequests-unstable`. While using unstable, (which we already test), the static and fully linked parts of the PR are not replicated anywhere yet.

markus2330:
multiconfig-gcc47 was one of the jobs that found most issues in our cmake setup.

Maybe it would be enough to only run `cmake`, just to see if all cmake options are selectable without any cmake error. And we can add more when we have more hardware.

ingwinlu:
> multiconfig-gcc47 was one of the jobs that found most issues in our cmake setup.

We cover most of what was configured in the matrix build already. The only thing that could be more enforced are the ENABLE_LOGGER + ENABLE_DEBUG flags but even those we cover in some builds.

markus2330:
> ENABLE_LOGGER + ENABLE_DEBUG

Yes, these are important to always work because devs should be able to rely on it.

> gcc47

gcc47 is irrelevant (historic name of the time when it was created), the build job was only about that all combinations of cmake variables work.

ingwinlu:

ingwinlu:
@e1528532 still seems to fail

e1528532:
The PR #2069 has been merged into master recently, is your branch already rebased on that because this is supposed to fix these issues.
I happens because the check_kdb_internal_check script uses kdb-full for the tests if it is build with FULL but only the shared version kdb includes the haskell plugins.

ingwinlu:
```
190/224 Test #162: testscr_check_kdb_internal_check .........***Failed    3.25 sec

ELEKTRA KDB INTERNAL TEST SUITE

Check if script tests the correct version
/home/jenkins/workspace/libelektra_PR-2046-QQUB5CQLSKE3BG46Q3LBTWAUHWVIHA4J6VEH4F63G64LS7BIEELA/build/tests/shell/testscr_check_kdb_internal_check.sh: 1021: [: 0: unexpected operator
error: check of plugin haskell failed
error: check of plugin haskell produced: "Was not able to load such a plugin! Maybe you misspelled it, there is no such plugin or the loader has problems. You might want to try to set LD_LIBRARY_PATH, use kdb-full or kdb-static. Errors/Warnings during loading were: Sorry, 1 warning was issued ;( Warning (#70): Description: could not load module, static load failed Ingroup: modules Module: static At: /home/jenkins/workspace/libelektra_PR-2046-QQUB5CQLSKE3BG46Q3LBTWAUHWVIHA4J6VEH4F63G64LS7BIEELA/src/libs/loader/static.c:81 Reason: Did not find module: haskell Mountpoint: Configfile: "
error: check of plugin regexdispatcher failed
error: check of plugin regexdispatcher produced: "Was not able to load such a plugin! Maybe you misspelled it, there is no such plugin or the loader has problems. You might want to try to set LD_LIBRARY_PATH, use kdb-full or kdb-static. Errors/Warnings during loading were: Sorry, 1 warning was issued ;( Warning (#70): Description: could not load module, static load failed Ingroup: modules Module: static At: /home/jenkins/workspace/libelektra_PR-2046-QQUB5CQLSKE3BG46Q3LBTWAUHWVIHA4J6VEH4F63G64LS7BIEELA/src/libs/loader/static.c:81 Reason: Did not find module: regexdispatcher Mountpoint: Configfile: "
error: check of plugin typechecker failed
error: check of plugin typechecker produced: "Was not able to load such a plugin! Maybe you misspelled it, there is no such plugin or the loader has problems. You might want to try to set LD_LIBRARY_PATH, use kdb-full or kdb-static. Errors/Warnings during loading were: Sorry, 1 warning was issued ;( Warning (#70): Description: could not load module, static load failed Ingroup: modules Module: static At: /home/jenkins/workspace/libelektra_PR-2046-QQUB5CQLSKE3BG46Q3LBTWAUHWVIHA4J6VEH4F63G64LS7BIEELA/src/libs/loader/static.c:81 Reason: Did not find module: typechecker Mountpoint: Configfile: "
testscr_check_kdb_internal_check.sh RESULTS: 196 test(s) done 6 error(s).
```

e1528532:
Hmm. Maybe it behaves differently on debian, only tried with osx, let me check

e1528532:
of course it behaves differently on debian...
I don't want to open a new PR for this little thingy.

Please change in check_kdb_internal_check.sh the contents to the following, issue was that i used == inside the if instead of only a single = that apparently works on mac and fails on debian.

```
echo "$KDB" | grep -qE "(full|static)\$"
if [ $? = "0" ]; then
	ACTUAL_PLUGINS=$ADDED_PLUGINS_WITHOUT_ONLY_SHARED
fi
```

ingwinlu:
> debian

Not debians fault, we use `sh` which is not consistent across platforms.

The whole grep construct seems strange, why not test the string directly?

ingwinlu:
there we go 97ace7bb01f6797bd54d062d05f706cd60514edc seems to fix the issue and contains a wonderfully portable + unreadable posix test ^^

sanssecours:
> Please change in check_kdb_internal_check.sh the contents to the following, issue was that i used == inside the if instead of only a single = that apparently works on mac and fails on debian.

> Not debians fault, we use `sh` which is not consistent across platforms.

From my limited personal experience the Shell `sh` works fine across platforms. The syntax `==` for comparison is a just a [Bash specific extension](https://unix.stackexchange.com/questions/16109/bash-double-equals-vs-eq).

We actually do not need the last `if` in the code. Something like 

```sh
echo "$KDB" | egrep -q '(full|static)$' && ACTUAL_PLUGINS=$ADDED_PLUGINS_WITHOUT_ONLY_SHARED
```

should work on a POSIX compatible version of `sh`.

ingwinlu:
> From my limited personal experience the Shell sh works fine across platforms.

sh is usually just a symlink to whatever default shell implementation a distribution uses. So for debian that might be dash. For alpine it is ash. On my arch system it is even bash.

So sh just means `give me a shell`. If your script has shell specific functions and you use the sh shebang you will run into inconsistencies across platforms.

sanssecours:
> sh is usually just a symlink to whatever default shell implementation a distribution uses. So for debian that might be dash. For alpine it is ash. On my arch system it is even bash.

Correct.

> So sh just means `give me a shell`.

Specifying `sh` in the shebang usually states [“give me a Shell that is **compatible with** `sh`”](https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash#5725402). Even the worst shells (I am looking at you [`dash`](https://bugs.launchpad.net/ubuntu/+source/dash/+bug/1499473)) provide **at least** the functions specified in the [POSIX spec](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html).

> If your script has shell specific functions and you use the sh shebang you will run into inconsistencies across platforms.

Also correct.

markus2330:
Thank you, great job!

PR #2047, opened 2018-06-05
build: port i386 job to new build system

ingwinlu:
# Purpose

port i386 job to new build system

# Checklist

### DIFF
--- a/doc/BUILDSERVER.md
+++ b/doc/BUILDSERVER.md
@@ -91,7 +91,6 @@ build:
 * jenkins build [gcc-configure-debian-musl](https://build.libelektra.org/job/elektra-gcc-configure-debian-musl/) please
 * jenkins build [gcc-configure-debian-shared](https://build.libelektra.org/job/elektra-gcc-configure-debian-shared/) please
 * jenkins build [gcc-configure-debian-withspace](https://build.libelektra.org/job/elektra-gcc-configure-debian-withspace/) please
-* jenkins build [gcc-i386](https://build.libelektra.org/job/elektra-gcc-i386/) please
 * jenkins build [git-buildpackage-jessie](https://build.libelektra.org/job/elektra-git-buildpackage-jessie/) please
 * jenkins build [git-buildpackage-wheezy](https://build.libelektra.org/job/elektra-git-buildpackage-wheezy/) please
 * jenkins build [icc](https://build.libelektra.org/job/elektra-icc/) please
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -212,7 +212,6 @@ Thanks to Armin Wurzinger.
   - Better Docker image handling.
   - Abort of previously queued but unfinished runs on new commits.
   - Document how to locally replicate the Docker environment used for tests.
-  *(Lukas Winkler)*
 - The Jenkins build server now also compiles and tests Elektra with enabled address sanitizer. *(Lukas Winkler)*
 - Ported GCC ASAN build job to new build system *(René Schwaiger + Lukas Winkler)*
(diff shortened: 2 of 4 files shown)


ingwinlu:
@markus2330 how did we handle installing i386 packages on the existing build slaves? manually?

markus2330:
libc6-dev-i386 and similar packages are installed, which uses Debian [Multiarch](https://wiki.debian.org/Multiarch/HOWTO).

But this was rather a workaround because we did not have real i386 build slaves. Please do not add these packages in our main Dockerfile, it is already large enough :smile: 

If you can create an i386 Docker container it would be much more what we wanted. Also `-m32` is only a workaround because we did not have a i386 system. Afaik, we never found any bug in this cross-compiling build job.

It might not harm to have a real i386 **and** a cross-compiling build job, though.

Btw. why are most stages failing here (and not only the i386 one)? https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2047/1/pipeline/330

ingwinlu:
> Btw. why are most stages failing here (and not only the i386 one)? 

I manually canceled it to not waste resources after I saw the problem with i386. Sadly the fancy new webui does not contain that information. But it is displayed at https://build.libelektra.org/jenkins/job/libelektra/job/PR-2047/1/ .

> Please do not add these packages in our main Dockerfile, it is already large enough

My thoughts exactly.

> If you can create an i386 Docker container it would be much more what we wanted. Also -m32 is only a workaround because we did not have a i386 system. Afaik, we never found any bug in this cross-compiling build job.

We can try to add a new image for cross compilation OR create a native i386 image that will only run on i386.

markus2330:
I am for the second option then (create a native i386 image that will only run on i386). Ideally also i386 Debian packages drop out of the job (into the same reprepro). I have still some i386 systems running :wink:

markus2330:
What do you mean? Isn't an amd64 docker able to start an i386 Docker image?

ingwinlu:
> What do you mean? Isn't an amd64 docker able to start an i386 Docker image?

Start maybe, not sure if it is supported. For example even if you execute the entrypoint you will still share the kernel with your host system (which would be x86).

sanssecours:
I think we should not waste any time on support for i386, unless somebody needs it. In my opinion support for other non-outdated CPU architectures, such as ARM (64 bit) or Power would be much nicer.

markus2330:
>   Start maybe, not sure if it is supported. For example even if you execute the entrypoint you will still share the kernel with your host system (which would be x86).

According to https://github.com/moby/moby/issues/611 it works but some 64 bit libs are needed.

> I think we should not waste any time on support for i386, unless somebody needs it.

According to https://popcon.debian.org/ it is still (by far) the second most popular architecture. And as said I have i386 computers which I currently manually need to compile Elektra for. So somebody needs it :smile:

> In my opinion support for other non-outdated CPU architectures, such as ARM (64 bit) or Power would be much nicer.

I fully agree. But this is more effort and much slower (if emulated, e.g., with qemu). If @ingwinlu is more interested in doing something like this I can live without i386 for some time. But if base our decision just on the numbers we should create an i386 build job.

ingwinlu:
From my personal experience cross compiling with qemu can result in some 'unique' problems. As a rule of thumb you should always strive to do the linking on the target platform.

However:
* The current build system is overloaded as is. Adding more build jobs should be done only when it is absolutely needed.
* We have no native nodes for i386, arm or anything else active at the moment

I therefore ported the i386 job now to replicate the old behaviour although in a Docker container.

markus2330:
Thank you for creating this job! Please rebase.

markus2330:
Thank you, great job!

Please describe the jobs/Dockerfiles as comment and in the release notes. Noteworthy is that it seems to change locales, is only a minimal build, ...

Issue #287, opened 2015-09-15
Clean up tmp files/directories during/after building


manuelm:
The /tmp directory of my build containers have a LOT of tmp files and directories which get created during the build process. We should either clean all of them at the end or during the build.

tmp.XXXXXXXXX files in /tmp: 1541
elektraenv.XXXXXXXXX directories in /tmp: 12
elektra-test.XXXXXX directories in /tmp: 7
getenv-XXXXXX.ii files in /tmp: 6
cpp11_benchmark_thread-XXXXXX.ii files in /tmp: 1

markus2330:
You are right. They also should all start with elektra-.

markus2330:
#284 is already a step in this direction.

markus2330:
You mention some other folders and /tmp/jna--* seems also to be from Elektra

markus2330:
shell recorder should do cleanup, too

markus2330:
A trap for cleanup after signals would be great, too.

markus2330:
What about changing TMPDIR and clean up the whole directory? Seems easier than to track down every single tmpfile in every test case.

markus2330:
There are still plenty of such cases, including /tmp/elektra-test and elektraenv.XXXXXXXXX.

shellrecorder could help for cleanup, and finally fixing TMPDIR, see #1973

@ingwinlu Can the build server check which leftovers the testcases have? (So that we at least not increase the number of leaking test cases.)

markus2330:
@sanssecours Is it a good idea to keep the Protocol File of the shell recorder? We already print it, so they could be removed, can't they?

sanssecours:
> We already print it, so they can be removed, can't they?

Makes sense, I opened pull request #1979 which updates the behavior of the Shell Recorder accordingly.

markus2330:
Is this now fixed with #2056? Or do we need cleanup for other test suites (like gtest), too?

Can we add a check in our build server which checks if we leave /tmp in the state as before?

PR #2056, opened 2018-06-07
tests: Properly remove temporary test home dir

ingwinlu:
# Purpose

* check if required functions are available
* set test feature macros explicitly if needed
* replace manual file deletion with nftw call if nftw is available.

This should resolve #2008 and some parts of #1973. But fcrypt still fails as soon as a `TMPDIR` is specified.

# Checklist

- [ ] I added unit tests

### DIFF
--- a/tests/cframework/tests.c
+++ b/tests/cframework/tests.c
@@ -21,10 +21,15 @@
 #include <locale.h>
 #endif
 
+#ifdef HAVE_NFTW
+#include <ftw.h>
+#include <stdlib.h>
+#define NOPENFD 20
+#endif
+
 #include <regex.h>
 
 #include <kdbinternal.h>
-
 
 int nbError;
 int nbTest;
@@ -481,6 +486,27 @@ int output_error (Key * errorKey)
 	return 0;
 }
 
+#ifdef HAVE_NFTW
+static int rm_all (const char * fpath, const struct stat * sb ELEKTRA_UNUSED, int tflag, struct FTW * ftwbuf ELEKTRA_UNUSED)
+{
+	if (tflag == FTW_F)
+	{
+		unlink (fpath);
+	}
+	else if (tflag == FTW_D || tflag == FTW_DP)
+	{
+		rmdir (fpath);
+	}
+	else
+	{
+		// not a file or dir we can delete
+		printf ("unexpected flag: %d\n", tflag);
+		return 1;
+	}
+	return 0;
+}
+#endif
+
 static void clean_temp_home (void)
 {
 	if (tmpfilename)
@@ -499,6 +525,10 @@ static void clean_temp_home (void)
 
 	if (tempHome)
 	{
+#ifdef HAVE_NFTW
+		int nftw_flags = FTW_DEPTH | FTW_PHYS;
+		succeed_if (nftw (tempHome, rm_all, NOPENFD, nftw_flags) == 0, "Could not delete TMPHOME via nftw");
+#else
 		size_t fileToCleanLen = tempHomeLen + 30;
 		char * fileToClean = elektraMalloc (fileToCleanLen);
 		snprintf (fileToClean, fileToCleanLen, "%s/.gnupg/random_seed", tempHome);
@@ -513,7 +543,8 @@ static void clean_temp_home (void)
 		rmdir (fileToClean);
 		elektraFree (fileToClean);
 
-		rmdir (tempHome);
+		succeed_if (rmdir (tempHome) == 0, "Could not delete TMPHOME manually");
+#endif
 		elektraFree (tempHome);
 		tempHome = NULL;
 		tempHomeLen = 0;
--- a/tests/cframework/tests.c
+++ b/tests/cframework/tests.c
@@ -21,7 +21,7 @@
 #include <locale.h>
 #endif
 
-#ifdef HAVE_NFTW
+#ifdef USE_NFTW
 #include <ftw.h>
 #include <stdlib.h>
 #define NOPENFD 20
@@ -36,10 +36,6 @@ int nbTest;
 
 char file[KDB_MAX_PATH_LENGTH];
 char srcdir[KDB_MAX_PATH_LENGTH];
-
-#ifdef HAVE_CLEARENV
-int clearenv (void);
-#endif
 
 char * tmpfilename;
 char * tempHome;
@@ -486,7 +482,7 @@ int output_error (Key * errorKey)
 	return 0;
 }
 
-#ifdef HAVE_NFTW
+#ifdef USE_NFTW
 static int rm_all (const char * fpath, const struct stat * sb ELEKTRA_UNUSED, int tflag, struct FTW * ftwbuf ELEKTRA_UNUSED)
 {
 	if (tflag == FTW_F)
@@ -525,7 +521,7 @@ static void clean_temp_home (void)
 
 	if (tempHome)
 	{
-#ifdef HAVE_NFTW
+#ifdef USE_NFTW
 		int nftw_flags = FTW_DEPTH | FTW_PHYS;
 		succeed_if (nftw (tempHome, rm_all, NOPENFD, nftw_flags) == 0, "Could not delete TMPHOME via nftw");
 #else
--- a/cmake/ElektraCompiling.cmake
+++ b/cmake/ElektraCompiling.cmake
@@ -117,6 +117,9 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
 	set (CMAKE_INCLUDE_SYSTEM_FLAG_CXX "-isystem ")
 endif ()
 
+#
+# ASAN
+#
 if (ENABLE_ASAN)
 	set (EXTRA_FLAGS "${EXTRA_FLAGS} -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer")
 	set (ASAN_LIBRARY "-lasan") # this is needed for GIR to put asan in front
--- a/cmake/ElektraCompiling.cmake
+++ b/cmake/ElektraCompiling.cmake
@@ -90,6 +90,17 @@ if (CMAKE_COMPILER_IS_GNUCXX)
 	endif (WIN32)
 endif (CMAKE_COMPILER_IS_GNUCXX)
 
+#
+# Platform specific settings
+#
+if (APPLE)
+
+	# Workaround to not loose functions as soon as feature test macros are used
+	# https://boringssl.googlesource.com/boringssl/+/63a0797ff247f13870b649c3f6239d80be202752
+	# https://charm.cs.illinois.edu/redmine/issues/1743
+	add_definitions (-D_DARWIN_C_SOURCE)
+endif(APPLE)
+
 if (WIN32)
 	set (HAVE_WIN32 "1")
 	message (STATUS "Win32 detected")
--- a/cmake/ElektraCompiling.cmake
+++ b/cmake/ElektraCompiling.cmake
@@ -93,14 +93,6 @@ endif (CMAKE_COMPILER_IS_GNUCXX)
 #
 # Platform specific settings
 #
-if (APPLE)
-
-	# Workaround to not loose functions as soon as feature test macros are used
-	# https://boringssl.googlesource.com/boringssl/+/63a0797ff247f13870b649c3f6239d80be202752
-	# https://charm.cs.illinois.edu/redmine/issues/1743
-	add_definitions (-D_DARWIN_C_SOURCE)
-endif (APPLE)
-
 if (WIN32)
 	set (HAVE_WIN32 "1")
 	message (STATUS "Win32 detected")
--- a/cmake/ElektraCompiling.cmake
+++ b/cmake/ElektraCompiling.cmake
@@ -99,7 +99,7 @@ if (APPLE)
 	# https://boringssl.googlesource.com/boringssl/+/63a0797ff247f13870b649c3f6239d80be202752
 	# https://charm.cs.illinois.edu/redmine/issues/1743
 	add_definitions (-D_DARWIN_C_SOURCE)
-endif(APPLE)
+endif (APPLE)
 
 if (WIN32)
 	set (HAVE_WIN32 "1")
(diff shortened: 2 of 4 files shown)


ingwinlu:
jenkins build all please

sanssecours:
Seems the updated code [does not work on macOS](https://travis-ci.org/ElektraInitiative/libelektra/jobs/389216279#L5579-L5587):

```
/Users/travis/build/ElektraInitiative/libelektra/tests/cframework/tests.c:85:14: error: implicit declaration of function 'mkdtemp' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        succeed_if (mkdtemp (tempHome) != 0, "mkdtemp failed");
                    ^
/Users/travis/build/ElektraInitiative/libelektra/tests/cframework/tests.c:85:14: error: this function declaration is not a prototype [-Werror,-Wstrict-prototypes]
/Users/travis/build/ElektraInitiative/libelektra/tests/cframework/tests.c:442:60: error: unused parameter 'sb' [-Werror,-Wunused-parameter]
static int rm_all (const char * fpath, const struct stat * sb, int tflag, struct FTW * ftwbuf)
                                                           ^
/Users/travis/build/ElektraInitiative/libelektra/tests/cframework/tests.c:442:88: error: unused parameter 'ftwbuf' [-Werror,-Wunused-parameter]
static int rm_all (const char * fpath, const struct stat * sb, int tflag, struct FTW * ftwbuf)
                                                                                       ^
4 errors generated.
```

.

sanssecours:
I fixed the compilation issues on macOS in the latest two commits. I hope the updated code also works on Linux.

ingwinlu:
http://man7.org/linux/man-pages/man3/mkdtemp.3.html
http://man7.org/linux/man-pages/man3/ftw.3.html
http://man7.org/linux/man-pages/man7/feature_test_macros.7.html

We probably will need a conditional here to make it work on macos, but I am not familiar which libc they use.

ingwinlu:
@sanssecours which libc is used on macos builds? I can not find any documentation on that.
I don't really want to guess which options we need to set.

We could also look into specifying the feature test macros via cmake.

markus2330:
@petermax2 already has written code in #2008 doing something similar which seems to work with Mac OS X. (ftw is deprecated in POSIX but at least seems to be available in Mac OS X + Linux). Alternative: we could check for availability of nftw and conditionally include the code. Then we simply not remove left-overs in Mac OS X and Windows (once someone ports the unit test framework to Windows).

I think it is a good idea to remove all left-overs (otherwise we can never fix #287). But maybe we should print the file names so that we see if there are buggy test cases? (Or even bugs like kdb rm -r not removing the files.) For releases we could have regression tests if the list of left-over files does not grow.

ingwinlu:
I would love to know what provides libc on macOS (and why they disable mkdtemp as soon as _POSIX_C_SOURCE is set)

sanssecours:
> which libc is used on macos builds?

You can download a fairly recent version of Apple’s Libc [here](https://opensource.apple.com/release/macos-10133.html).

> and why they disable mkdtemp as soon as _POSIX_C_SOURCE is set

Maybe they do that because `mkdtemp` is not part of the POSIX base standard but part of the extended API? At least that is what I assume after I read this [StackOverflow question](https://stackoverflow.com/questions/11440787/mkdtemp-requires-darwin-c-source-for-unistd-h#11443288).

Anyway, I defined `_DARWIN_C_SOURCE` in `test.c` in commit 3a5f8a3d. After this update the code [should work both on macOS and Linux](https://travis-ci.org/sanssecours/elektra/builds/390020876). If you think that is an acceptable fix, please feel free to incorporate my changes into this PR.

ingwinlu:
@sanssecours thanks I will keep it in mind.

However: I would love to not define those values somewhere buried in the c code (for example date.c sets another value), but instead have it in one place in the cmake file.
That way we have the same rules for compilation for all c/cxx files.

I only just noticed we are not using c99 as c standard but gnu99 which seems to make some problems as it is allowing a different set of functionality.

Currently I am shifting through the standards to find a combination that will work for us.

markus2330:
The goal is to have as much pure c99 code as possible. It would be better if the files are also compiled as such (so that we notice if non-c99 constructs are introduced). Unfortunately it did not work. That gnu99 is used everywhere is because all parts of Elektra are compiled together in the FULL/STATIC variants (even though initially only the resolver actually needed it, now there are some more plugins using it).

In this case (nftw for the test framework), however, we do not have to introduce "XOPEN_SOURCE" everywhere, because it is not needed to compile libelektra. What about adding this selectively to the cframework target?

It has the advantage that no single combination of flags needs to be found. It has the disadvantage that nftw cannot be used within modules (and is most likely only a matter of time until it is needed somewhere in plugins).

And you should use
https://cmake.org/cmake/help/v3.0/command/add_definitions.html
which does exactly what you want and is portable to other compilers.

ingwinlu:
here is a summary as far as i understand the issue:

* gnu99 defaults to set `_DEFAULT_SOURCE` which allows all the GNU stuff (which might not be posix or any other standard)
* setting `XOPEN_SOURCE` does not automatically set `_DEFAULT_SOURCE` anymore (which in turn enabled _POSIX_C_SOURCE etc...)

For linux based systems this behaviour is defined in `/usr/include/features.h`. I have no idea where it is for macOS.

I am not a fan of having those defined only for a subset of files.

Also if you try to enforce c99 you will see that there are a lot of parts of elektra not conforming to the standard.

ingwinlu:
@markus2330 if I have learned anything is that the current setup allows pretty much everything to be used. Especially clang on macos seems to enable all kind of compatibility flags if nothing is set to reduce the set that is available.

markus2330:
> I am not a fan of having those defined only for a subset of files.

This is already heavily used and there is no way around this in such a big code base.

We should not weaken our portability just to have the same flags for tests and the core.

> Also if you try to enforce c99 you will see that there are a lot of parts of elektra not conforming to the standard.

Yes, that is the problem. If the compiler (build server) does not check it, weird things always creep in. This causes huge trouble if someone actually needs to port it somewhere else.

What we could do is to have a build job which sets C_STD:

`-DC_STD="--std=c99" -DPLUGINS="wresolver;dump;list;spec" -DKDB_DEFAULT_RESOLVER=wresolver -DTOOLS=kdb -DBINDINGS=cpp -DBUILD_TESTING=OFF` 

It seems to compile, so we should add this to the buildserver (before someone breaks it).

There are, however, three warnings (gcc6):

- src/libs/elektra/internal.c:162:9: warning: implicit declaration of function ‘strcasecmp’ / ‘strncasecmp’. 

This is obviously something we need to reimplement if we hit a system that does not have these functions.

But:

- src/libs/pluginprocess/pluginprocess.c:275:6: warning: implicit declaration of function ‘mkdtemp’ [-Wimplicit-function-declaration
- src/libs/pluginprocess/pluginprocess.c:275:30: warning: comparison between pointer and integer

is because the build server was not strict enough with C_STD. It should not have been allowed as part of our core libraries. I created a separate issue for it in #2061.

ingwinlu:
> It seems to compile, so we should add this to the buildserver (before someone breaks it).

Just because it compiles doesn't mean it works. It should segfault as soon as the function is called.

markus2330:
The suggested test is not about if it "works" but if non-POSIX constructs are introduced in the core.

If you can fix the cframework to still compile even if the core is compiled with C99, we can also add functional tests. But you will see that such undertakings are not so easy. Thus I suggested to only catch the problem and not redo the same functional tests which are already done in nearly all other jobs.

ingwinlu:
Wouldn't that be fairly pointless as long as we do not have Warnings as Errors passing? Will it not just pass all the time?

markus2330:
Yes, you are right. As long we do not have a way to fail builds on new warnings, we have to fix #2061 and create these two trivial strcasecmp implementations (or simply suppress these warnings if we do not manage to build the cframework and the core in C99 at the same time).

ingwinlu:
> We should not weaken our portability just to have the same flags for tests and the core.

I am not sure if you got it but this PR is currently only expanding regarding the `D_XOPEN_SOURCE` setting. The other options are set implicitly currently. That is also why _XOPEN_SOURCE was not even needed for `ntfw` on clang + macOS.

So I would say the impact on portability is minimal at this time.

> file vs initiative wide setting of compatibility options

I would like to add that tracking compatibility an resulting issues might be harder if you do it on a file basis.

> using ftw as in #2008 

I am super surprised that this works. `ftw` should list directories before the subdirs/files in the directory. Hence the directories will not be empty when calling rmdir. Which makes rmdir fail.

markus2330:
If we get the c99 build job I am happy. For plugins the _XOPEN_SOURCE seems to be a sensible setting (500 is the lowest setting anyway). But for the core neither gnu99 nor the settings added here are okay.

> So I would say the impact on portability is minimal at this time.

Porting the cframework to non-POSIX systems definitely got harder. But if this fixes #287 it is worth it.

ingwinlu:
Maybe we need a POSIX/c99 issue to track all violations.

ELEKTRA_UNUSED for example is also not c99/posix.

markus2330:
> Maybe we need a POSIX/c99 issue to track all violations.

If there is anything else next to #2061 and strncasecmp in the core, please open an issue.

> ELEKTRA_UNUSED for example is also not c99/posix.

The purpose of the macro is to reduce warnings, it does not have functionality. If a compiler does not have an attribute for unused variables, we can simply not use it there. Which compiler has problems with it?

ingwinlu:
til apples libc headers are currently broken and do not respect _POSIX_C_SOURCE if _DARWIN_C_SOURCE is not also specified.

I also could not find any documentation (apples sites give 404) on the issue.

Should we just define _DARWIN_C_SOURCE explicitly globally for apple builds?
@sanssecours @markus2330

sanssecours:
> til apples libc headers are currently broken

I doubt that. Do you have any sources to back that claim? I would assume there is a reason why they remove some functionality if you set `_POSIX_C_SOURCE `. According to an old (but up to date version) of the [man pages](https://www.unix.com/man-page/osx/5/compat/) setting only `_POSIX_C_SOURCE ` removes functions that are not compatible with the [Single UNIX Specification version 3 (aka. POSIX:2001)](https://en.wikipedia.org/wiki/Single_UNIX_Specification#2001:_Single_UNIX_Specification_version_3,_POSIX:2001):

> Defining _POSIX_C_SOURCE or _DARWIN_C_SOURCE causes library and kernel calls to conform to
     the SUSv3 standards even if doing so would alter the behavior of functions used in 10.3.
     Defining _POSIX_C_SOURCE also removes functions, types, and other interfaces that are not
     part of SUSv3 from the normal C namespace, unless _DARWIN_C_SOURCE is also defined (i.e.,
     _DARWIN_C_SOURCE is _POSIX_C_SOURCE with non-POSIX extensions).

.

> Should we just define _DARWIN_C_SOURCE explicitly globally for apple builds?

As long as this change does not break anything, that works for me.

ingwinlu:
sure. I added them as link in the code
https://boringssl.googlesource.com/boringssl/+/63a0797ff247f13870b649c3f6239d80be202752
https://charm.cs.illinois.edu/redmine/issues/1743

ingwinlu:
the joke is setting _POSIX_C_SOURCE removes the functionality why i set _POSIX_C_SOURCE in the first place.

sanssecours:
> the joke is setting _POSIX_C_SOURCE removes the functionality why i set _POSIX_C_SOURCE in the first place.

I think that behavior is quite common. For example, the command `set -o posix` in `bash` also removes functionality that is not defined in the POSIX specification.

sanssecours:
>* For apple the behaviour seems (again i can find no coherent piece of documentation on this) to be similar to glibc. If nothing is defined everything is exported. But as seen in this PR as soon as I set _POSIX_C_SOURCE I suddenly do not have mkdtemp available anymore. Except if _DARWIN_C_SOURCE is also set.  **wat**. 
>
> so it is entirely not about removing behaviour that is not POSIX. it is about removing POSIX implementation as soon as setting a flag that should **ENABLE** POSIX implementations.

As far as I can tell macOS [only supports POSIX:2001 officially](https://www.opengroup.org/openbrand/register/brand3632.htm). If you only set `_POSIX_C_SOURCE` (and not `_DARWIN_C_SOURCE `) then macOS will behave according to this version of the standard, which [does not define `mkdtemp` as far as I know](http://pubs.opengroup.org/onlinepubs/007904975/idx/im.html).

ingwinlu:
@sanssecours thanks for digging that out.

While it is certainly a valid approach to the subject I feel it makes test feature macros hard to use consistently across platforms.
I am not sure if we need to replace the check_function_exists calls with check_symbol_exists now that we know the behaviour for apples systems.

tests/cframework/tests.c
```diff
@@ -513,7 +539,8 @@ static void clean_temp_home (void)
 		rmdir (fileToClean);
 		elektraFree (fileToClean);
 
-		rmdir (tempHome);
+		succeed_if (rmdir (tempHome) == 0, "Could not delete TMPHOME manually");
```
markus2330: won't this always fail? (Our tests are not clean at the moment)

markus2330:
So does this PR also work for Mac OS X?

tests/cframework/tests.c
ingwinlu: If you run gpg tests on a system with gpg2 instead of gpg that has no nftw exported, probably.

markus2330: I am in favor of removing this `succeed_if` and instead add a build-server check if all tmp files are cleaned up in the build jobs. What do you think? (see #287)

markus2330:
It also passed before. Let me rephrase: is #287 fixed?

ingwinlu:
@markus2330 no i still see at least 1 test that creates a a /tmp/elektra-tmp.XXXXXXX file (contains some chars). Seems to create a new file with same content on each run.

Seems like that is created in tests.c as well: https://github.com/ElektraInitiative/libelektra/blob/master/tests/cframework/tests.c#L91

No idea why though?

markus2330:
There are many tests that use elektraFilename to get a pre-defined temporary filename. (So that they do not have to care about creation and cleanup.)

The question is why it is not cleaned up. I do not see any call to `_exit` which could have caused this.

Maybe moving this filename into "home" will solve the problem?

Issue #734, opened 2016-05-18
XDG variables


markus2330:
Currently, XDG basically only has support for what is the `user` namespace in Elektra. (The specification works on files, the current implementation with per-key override of `system` already took quite some interpretation of [the standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html), e.g. because it does not specify what "importance" means.)

For the other namespaces, we are currently lacking environment variables:

- system
- system/elektra/mountpoints (if we want to handle it separately)
- spec
- dir

Should we add them, should we concentrate our XDG efforts on `user`, or should we discuss it on the XDG list?

beku:
XDG list discussion appears the best thing to clarify XDG spec intention and side effects beyond a few use cases. Whether it helps getting admin comments is a different question.

markus2330:
Btw. on the mailing list there was no interest to extend the standard: https://www.mail-archive.com/xdg@lists.freedesktop.org/msg07921.html

PR #2060, opened 2018-06-09
jenkins: port xdg build to new system

ingwinlu:
# Purpose

Long descriptions should be in documentation and not here.

# Checklist

needs to be checked.

- [ ] I added unit tests

### DIFF
--- a/scripts/jenkins/Jenkinsfile
+++ b/scripts/jenkins/Jenkinsfile
@@ -860,8 +860,8 @@ def withDockerEnv(image, opts=[], cl) {
                 def cpu_count = cpuCount()
                 withEnv(["MAKEFLAGS='-j${cpu_count+2} -l${cpu_count*2}'",
                          "CTEST_PARALLEL_LEVEL='${cpu_count+2}'",
-                         "XDG_CONFIG_HOME='${WORKSPACE}/xdg/user'",
-                         "XDG_CONFIG_DIRS='${WORKSPACE}/xdg/system'"]) {
+                         "XDG_CONFIG_HOME=${WORKSPACE}/xdg/user",
+                         "XDG_CONFIG_DIRS=${WORKSPACE}/xdg/system"]) {
                     echo "Starting ${env.STAGE_NAME} on ${env.NODE_NAME}"
                     checkout scm
                     docker.image(image.id)
--- a/scripts/jenkins/Jenkinsfile
+++ b/scripts/jenkins/Jenkinsfile
@@ -47,7 +47,6 @@ CMAKE_FLAGS_BASE = [
     'SITE': '${STAGE_NAME}',
     'KDB_DB_SYSTEM': '${HOME}/.config/kdb/system',
     'KDB_DB_SPEC': '${HOME}/.config/kdb/spec',
-    'XDG_CONFIG_DIRS': '${HOME}/.config/xdg/system',
     'CMAKE_INSTALL_PREFIX': '${WORKSPACE}/system',
     'INSTALL_SYSTEM_FILES': 'OFF',
     'BUILD_DOCUMENTATION': 'OFF'
--- a/scripts/jenkins/Jenkinsfile
+++ b/scripts/jenkins/Jenkinsfile
@@ -348,6 +348,16 @@ def generateFullBuildStages() {
         ,
         [TEST.ALL, TEST.MEM]
     )
+    def xdgResolver = 'resolver_mf_xp_x'
+    tasks << buildAndTest(
+        "debian-stable-full-xdg",
+        DOCKER_IMAGES.stretch,
+        [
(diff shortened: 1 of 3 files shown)


scripts/jenkins/Jenkinsfile
```diff
@@ -47,6 +47,7 @@ CMAKE_FLAGS_BASE = [
     'SITE': '${STAGE_NAME}',
     'KDB_DB_SYSTEM': '${HOME}/.config/kdb/system',
     'KDB_DB_SPEC': '${HOME}/.config/kdb/spec',
+    'XDG_CONFIG_DIRS': '${HOME}/.config/xdg/system',
```
markus2330: The idea of this build job is to replicate https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

which means to directly write to "${HOME}/.config", "/etc/xdg" and so on. The XDG env vars, like XDG_CONFIG_DIRS, are set within the resolver tests in tests/shell/check_resolver.sh, so they need not to be set outside.

ingwinlu: should we just make /etc/xdg globally writable? same for /etc/kdb instead of setting it to the home directory of the user?

markus2330: Sorry, seems like the old test has set the environment variables "XDG_CONFIG_HOME and
 XDG_CONFIG_DIRS". It is fine if we replicate that. Only for the spec we need some workaround (like changing CMAKE_INSTALL_PREFIX) because of #734.

ingwinlu: I am afraid I do not follow:
The old job did not set any of those variables. It just calls https://github.com/ElektraInitiative/libelektra/blob/master/scripts/configure-xdg.

markus2330: https://build.libelektra.org/jenkins/job/elektra-gcc-configure-xdg/configure has:

```
export XDG_CONFIG_HOME="$WORKSPACE/user"
export XDG_CONFIG_DIRS="$WORKSPACE/system"
mkdir build
cd build
../scripts/configure-xdg -DCMAKE_INSTALL_PREFIX=$WORKSPACE/local -DBUILD_DOCUMENTATION=OFF ..
run-make

run-make run_all
run-make run_memcheck
```

And yes, it would be good if we can continue to use the script.

scripts/jenkins/Jenkinsfile
```diff
@@ -849,7 +861,9 @@ def withDockerEnv(image, opts=[], cl) {
             timeout(activity: true, time: 5, unit: 'MINUTES') {
                 def cpu_count = cpuCount()
                 withEnv(["MAKEFLAGS='-j${cpu_count+2} -l${cpu_count*2}'",
-                         "CTEST_PARALLEL_LEVEL='${cpu_count+2}'"]) {
+                         "CTEST_PARALLEL_LEVEL='${cpu_count+2}'",
+                         "XDG_CONFIG_HOME=${WORKSPACE}/xdg/user",
+                         "XDG_CONFIG_DIRS=${WORKSPACE}/xdg/system"]) {
```
markus2330: It is confusing if the same environment is set for every build job but only used for some (or even one here). Isn't it possible to pass the env from the definition above (like PLUGINS and KDB_DEFAULT_RESOLVER)?

ingwinlu: those are fundamentally different settings.
PLUGINS, KDB_DEFAULT_RESOLVER = cmake flags
XDG_CONFIG_HOME, XDG_CONFIG_DIRS = environment variables

We could add a parameter to the buildAndTest function that sets those though

markus2330: Yes, that would be great.

markus2330:
I removed "ready to merge" so that there are no confusions when you have pushed your last commit.

Btw. it would be delightful if you add some text in the release notes what the jobs do. People do not know that Elektra supports XDG and even less that we have tests for it.

So a short description what the test does and a link to XDG might be helpful to some readers.

scripts/jenkins/Jenkinsfile
ingwinlu: I tried to implement that now for 5hours. Saldy it is not that easy:
* WORKSPACE is not available at the point where we want to set the variables that reference it
* Lazy evaluation does not work because there is not even a WORKSPACE reference at that point
* templating the strings seemed like the way to go but apparently WORKSPACE is not really set for the environment (maybe only for shell variables?). I dug deep in the jenkins code but did not find a reason for why it is that way. Anyway no dice either.

As there is no problem with having the variables defined for all Docker Environments (they are simply not used in other cases) I will stop trying to implement something that Jenkins definitely does not want you to do.

While we could simply not use the default buildAndTest function I would argue that the gains are not worth having the additional function.

ingwinlu: I removed the commits moving the XDG configuration due to the reasons above. I kept the commits in https://github.com/ingwinlu/libelektra/tree/impl_xdg_debug.

markus2330:
Thank you!

PR #2066, opened 2018-06-11
CCode Updates & Other Minor Enhancements

sanssecours:
# Purpose

This pull request adds support for key name escaping to the [CCode plugin](https://www.libelektra.org/plugins/ccode). The PR also contains other minor enhancements, including

- relative storage support for the [Tcl plugin](https://www.libelektra.org/plugins/tcl), and
- updates to `info/status` of various plugins

.

# Checklist

- [x] I added a Markdown Shell Recorder test for the CCode plugin.
- [x] The PR contains some logging code for the CCode plugin.
- [x] The release notes should be up to date.


sanssecours:
jenkins build all please

doc/help/kdb-find.md
```diff
@@ -55,7 +55,7 @@ kdb find 'fizz'
 #> user/tests/find/tests/fizz/buzz
 #> user/tests/find/tostfizz
 
-sudo kdb umount /tests/find
```
markus2330: We should umount nevertheless. The umount feature of the shell recorder should only be there to catch incorrectly written shell recorder scripts.

doc/help/kdb-ls.md
```diff
@@ -91,7 +91,7 @@ kdb ls /tests/examples/kdb-ls/ -v
 #> user/tests/examples/kdb-ls/tost
 #> user/tests/examples/kdb-ls/tost/level
 
-kdb umount /tests/examples
```
markus2330: Same here

doc/man/man1/kdb-find.1
```diff
@@ -71,7 +71,8 @@ kdb find \'fizz\'
 #> user/tests/find/tests/fizz/buzz
 #> user/tests/find/tostfizz
 
-sudo kdb umount /tests/find
+kdb rm \-r /tests/find
+kdb umount /tests/find
```
markus2330: Why is umount here but not above in the source? This seems to be correct.

doc/news/_preparation_next_release.md
```diff
@@ -71,10 +71,20 @@ Thanks to Armin Wurzinger.
...
+- We fixed various warnings in the source code reported by [OCLint](http://oclint.org). *(René Schwaiger)*
+- The plugin now also encodes and decodes key names in addition to key values. *(René Schwaiger)*
+
 ### Crypto
 
 - The `crypto` plugin now uses Elektra's `libinvoke` and the `base64` plugin in order to encode and decode Base64 strings. This improvement reduces code duplication between the two plugins. *(Peter Nirschl)*
 
+### Haskell
+
+- An issue when building Haskell plugins with a cached sandbox is fixed in case
```
markus2330: Thank you for reordering this. We should discuss it in the next meeting how to use your new organization.

doc/news/_preparation_next_release.md
```diff
@@ -93,13 +103,22 @@ Thanks to Armin Wurzinger.
...
 
 ### Regex Dispatcher
 
 - The plugin [regexdispatcher](https://www.libelektra.org/plugins/regexdispatcher) has been added. It calculates regex representations for
   commonly used specification keywords to be used with the [typechecker](https://www.libelektra.org/plugins/typechecker). Currently the
   keywords `check/range`, `check/enum` and `check/validation` are supported. *(Armin Wurzinger)*
 
+### Tcl
+
+- The [`tcl`](http://libelektra.org/plugins/tcl) plugin does not fail anymore, if its configuration file does not exist and you trie to
```
markus2330: try

doc/news/_preparation_next_release.md
```diff
@@ -146,13 +176,16 @@ Thanks to Armin Wurzinger.
 - The man pages for [`kdb change-resolver-symlink`](https://www.libelektra.org/manpages/kdb-change-resolver-symlink) and
    [`kdb change-storage-symlink`](https://www.libelektra.org/manpages/kdb-change-storage-symlink) referenced the wrong command.
    *(Lukas Winkler, René Schwaiger)*
+- The documentation for `kdb` and `kdb set` now mention the `--` option to stop option processing. This is useful for setting negative values among other things. *(Klemens Böswirth)*
```
markus2330: mentions

I would not call `--` an option/setting. (It is a special command-line argument which indicates where options/setting end.) (And we should use settings instead of options.)

doc/news/_preparation_next_release.md
```diff
@@ -161,9 +194,10 @@ Thanks to Armin Wurzinger.
 - The Markdown Shell Recorder now also tests if a command prints nothing to `stdout` if you add the check `#>`. *(René Schwaiger)*
 - We fixed some problems in the [Markdown Shell Recorder](https://master.libelektra.org/tests/shell/shell_recorder/tutorial_wrapper) test
   of [`kdb ls`](https://master.libelektra.org/doc/help/kdb-ls.md). *(René Schwaiger)*
-- The documentation for `kdb` and `kdb set` now mention the `--` option to stop option processing. This is useful for setting negative values among other things. *(Klemens Böswirth)*
-
-## Compatibility
+- The Markdown Shell Recorder test for [`kdb find`](https://master.libelektra.org/doc/help/kdb-find.md) now removes the configuration file
+  at the end of the test. *(René Schwaiger)*
+- The [Shell Recorder](https://master.libelektra.org/tests/shell/shell_recorder) now properly unmounts any additional mountpoints created
```
markus2330: I love the feature but we should still unmount everything also explicitely. Otherwise copy&paste of our scripts do not clean up.

src/plugins/ccode/README.md
```diff
@@ -8,13 +8,76 @@
...
+# Write and read a key value containing a tab character
+kdb set user/tests/ccode/tab 'Tab	Fabulous'
+kdb get user/tests/ccode/tab
+#> Tab	Fabulous
+
+# The plugin also escapes special characters inside key names
+kdb set 'user/tests/ccode/tab/t	a	b' 'Escaped Tabs'
+grep 'tab/' `kdb file user/tests/ccode` | sed 's/[[:space:]]*//'
+#> tab/t\ta\tb = Escaped Tabs
+
```
markus2330: What if the escape character itself is present?

src/plugins/ccode/README.md
```diff
@@ -8,13 +8,76 @@
...
+# Undo modifications to database
+kdb rm -r user/tests/ccode
+sudo kdb umount user/tests/ccode
+```
+
+### Custom Configuration
+
+```sh
+# We use `%` as escape character and map the space character (hex: `20`) to the character `_`.
+sudo kdb mount config.tcl user/tests/ccode tcl base64 ccode escape=`bash -c 'printf %x \"%'` chars= chars/20=`bash -c 'printf %x \"_'`
```
markus2330: What about not requiring `chars=`?

src/plugins/ccode/ccode.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+	CppKeySet keys{ returned };
+	CppKeySet unescaped{};
+	for (auto key : keys)
+	{
+		size_t const size = max (key.getBinarySize (), key.getNameSize ());
+		if (size > mapping->bufferSize)
+		{
+			mapping->bufferSize = size;
+			mapping->buffer = new unsigned char[mapping->bufferSize];
```
markus2330: In modern C++ "new" is avoided. Here there is a big memory leak. You should use an array/vector instead. It avoids the memory leak, size is then implicitely available, and vectors have resize().

src/plugins/ccode/ccode.hpp
```diff
@@ -11,27 +11,35 @@
 
 #include <kdbplugin.h>
 
+#ifdef __cplusplus
```
markus2330: Is there any reason to include this in C code?

src/plugins/ccode/testmod_ccode.c
```diff
@@ -51,48 +39,53 @@ void test_encode (void)
...
+#ifdef __llvm__
+	__attribute__ ((annotate ("oclint:suppress[deep nested block]"), annotate ("oclint:suppress[high cyclomatic complexity]"),
+			annotate ("oclint:suppress[high ncss method]")))
+#endif
+
 {
-	CCodeData * d = get_data ();
-	char buf[1000];
-	d->buf = buf;
+	CCodeData * mapping = get_data ();
```
markus2330: Does this function test every possible character?

For maintenance it would be nice if the check_reversibility can be shared between the different code plugins.

tests/shell/shell_recorder/shell_recorder.sh
```diff
@@ -304,9 +304,30 @@ BACKUP=1
...
+	MOUNTPOINTS_BACKUP_FILE=$(mktempfile_elektra)
+	printf '%s' "$MOUNTPOINTS_BACKUP" > "$MOUNTPOINTS_BACKUP_FILE"
+	MOUNTPOINTS_FILE=$(mktempfile_elektra)
+	printf '%s' "$MOUNTPOINTS" > "$MOUNTPOINTS_FILE"
+
+	TOUMOUNT=$(diff "$MOUNTPOINTS_BACKUP_FILE" "$MOUNTPOINTS_FILE" | grep -Eo '^>.*')
+	for line in $TOUMOUNT;
+	do
+		mp=$(printf '%s' "$line" | sed -n 's/.*with name \(.*\)/\1/p')
+		"$KDB" umount "$mp"
```
markus2330: We should also warn the user that umount was forgotten here.

doc/help/kdb-find.md
sanssecours: > The umount feature of the shell recorder should only be there to catch incorrectly written shell recorder scripts.

I agree. I removed the auto unmount feature in the latest updates. This way an incorrectly written test will always fail.

doc/man/man1/kdb-find.1
sanssecours: > Why is umount here but not above in the source?

I guess my script to remove those pesky man pages where only the date changed also reverted the (single-line) update to the man page.

The man page and the ReadMe for `kdb-find` should be in sync after the last update again.

doc/news/_preparation_next_release.md
sanssecours: > (And we should use settings instead of options.)

I think settings is not a common word for something like `-V` or `--version`. I used the word “command line **switch**” in the latest commit. Anyway, in my opinion `option` would have worked as well too.

src/plugins/ccode/README.md
sanssecours: In that case the plugin escapes the escape character. I changed the MSR test slightly to show this behavior in the latest update.

src/plugins/ccode/ccode.cpp
sanssecours: > In modern C++ "new" is avoided.

I know. I just replaced `malloc` with `new` and `free` with `delete`.

> Here there is a big memory leak.

I do not think so, since I **did not change the way the plugin handles memory**. If there is a memory leak, then the Linux Travis ASAN build should also fail, which [is not the case](https://travis-ci.org/ElektraInitiative/libelektra/builds/390917873).

> You should use an array/vector instead. It avoids the memory leak, size is then implicitely available, and vectors have resize().

This certainly makes sense (except for the part about the memory leak of course 😊).

src/plugins/ccode/ccode.hpp
sanssecours: Since C does not support `using` or `extern "C"` we have to remove the code below, if we include the header file in `testmod_ccode.c`.

doc/help/kdb-find.md
markus2330: Yes, in general tests should fail if they do not properly clean up everything.

I am currently thinking about writing a tutorial of how to setup a computer with Elektra. In this tutorial it would be wanted that everything stays mounted at the end. It is questionable, however, if this tutorial can be checked via the shell recorder because it has many local dependencies, so maybe it won't be run via the shell recorder anyway.

src/plugins/ccode/ccode.cpp
markus2330: The memory leak only occurs if the size is > 1000 (especially in gradually increasing sizes there might be a memleak in every iteration), seems like this is not tested. I did not say you introduced the problem but in C++ it is now easier to fix :smile: 

And in general it is not safe to translate every malloc -> new because C++ additionally has exceptions which can cause memleaks not possible in C.

src/plugins/ccode/ccode.cpp
```diff
@@ -0,0 +1,438 @@
...
+}
+
+// ====================
+// = Plugin Interface =
+// ====================
+
+/** @see elektraDocOpen */
+int elektraCcodeOpen (Plugin * handle, Key * key ELEKTRA_UNUSED)
+{
+	CCodeData * const mapping = new CCodeData ();
```
markus2330: Should be unique_ptr, see vector discussion.

src/plugins/ccode/README.md
sanssecours: That sounds like a good idea, especially since I did not add `chars=` at first and had to debug why only adding `chars/20` did not work. However, for which key should we look in `ksLookupByName` then? As far as I know `ksLookupByName` does not support something like `/chars*` (glob expression) or `/chars.*` (regex) as argument.

src/plugins/ccode/testmod_ccode.c
sanssecours: > Does this function test every possible character?

Nope, not as far as I can tell. I [did not write this function](https://github.com/ElektraInitiative/libelektra/commit/0f8f033c) though 😊.

src/plugins/ccode/README.md
markus2330: I think the configuration settings passed to the plugin won't be too big. So simply iterate over everything and check if it contains /chars.

src/plugins/ccode/testmod_ccode.c
markus2330: Yeah, seems like I was in a hurry finishing my master thesis :smile:

sanssecours:
jenkins build all please

doc/news/_preparation_next_release.md
```diff
@@ -142,6 +153,14 @@ Thanks to Michael Zronek and Vanessa Kos.
   commonly used specification keywords to be used with the [typechecker](https://www.libelektra.org/plugins/typechecker). Currently the
   keywords `check/range`, `check/enum` and `check/validation` are supported. *(Armin Wurzinger)*
 
+### Tcl
+
+- The [`tcl`](http://libelektra.org/plugins/tcl) plugin does not fail anymore, if its configuration file does not exist and you try to
+  retrieve the plugin contract. *(René Schwaiger)*
+- The plugin now uses relative key names. This update addresses issue [#51][]. *(René Schwaiger)*
+
+[#51]: https://github.com/ElektraInitiative/libelektra/issues/51
```
markus2330: issues.libelektra.org/51 is preferred

doc/news/_preparation_next_release.md
```diff
@@ -211,6 +244,9 @@ Thanks to Michael Zronek and Vanessa Kos.
 - The Markdown Shell Recorder now also tests if a command prints nothing to `stdout` if you add the check `#>`. *(René Schwaiger)*
 - We fixed some problems in the [Markdown Shell Recorder](https://master.libelektra.org/tests/shell/shell_recorder/tutorial_wrapper) test
   of [`kdb ls`](https://master.libelektra.org/doc/help/kdb-ls.md). *(René Schwaiger)*
+- The Markdown Shell Recorder test for [`kdb find`](https://master.libelektra.org/doc/help/kdb-find.md) now removes the configuration file
+  at the end of the test. *(René Schwaiger)*
+- We removed the broken auto unmounting feature from the [Markdown Shell Recorder][]. *(René Schwaiger)*
```
markus2330: What is the behavior now?

src/bindings/cpp/tests/tests.hpp
```diff
@@ -24,15 +24,8 @@
...
-	do                                                                                                                                 \
-	{                                                                                                                                  \
-		ASSERT_TRUE (x) << y;                                                                                                      \
-	} while (0)
-#define exit_if_fail(x, y)                                                                                                                 \
-	do                                                                                                                                 \
-	{                                                                                                                                  \
-		ASSERT_TRUE (x) << y;                                                                                                      \
-	} while (0)
+#define succeed_if(x, y) ASSERT_TRUE (x) << y;
```
markus2330: This is not a good change, see https://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block

src/plugins/ccode/CMakeLists.txt
```diff
@@ -1,3 +1,12 @@
-include (LibAddMacros)
+if (DEPENDENCY_PHASE)
+	include (LibAddTest)
+	add_gtest (testmod_ccode
```
markus2330: why not add_plugin_test? Why in the dependency phase?

src/plugins/ccode/coder.hpp
```diff
@@ -0,0 +1,55 @@
...
+#ifndef ELEKTRA_CODER_HPP
+#define ELEKTRA_CODER_HPP
+
+#include <vector>
+
+#include <kdblogger.h>
+
+#include <keyset.hpp>
+
+using CppKeySet = kdb::KeySet;
```
markus2330: please limit scope (could be moved to class Coder)

src/plugins/ccode/coder.hpp
```diff
@@ -0,0 +1,55 @@
...
+using CppKey = kdb::Key;
+
+using std::string;
```
markus2330: This should not be done in header files.

src/plugins/ccode/testmod_ccode.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <kdbprivate.h>
+
+#include <tests.hpp>
+
+using std::string;
+
+using CppKeySet = kdb::KeySet;
+using CppKey = kdb::Key;
+
+using ckdb::elektraModulesClose;
```
markus2330: It is okay as is but you could also use "using ckdb;".

src/plugins/ccode/testmod_ccode.cpp
```diff
@@ -0,0 +1,109 @@
...
+			  keyNew ("user/chars/3D", KEY_VALUE, "65", KEY_END), // = -> e
+			  keyNew ("user/chars/3B", KEY_VALUE, "73", KEY_END), // ; -> s
+			  keyNew ("user/escape", KEY_VALUE, "25", KEY_END),   // use % as escape character
+			  KS_END };
+	return config;
+}
+
+void testRoundTrip (string const decodedString, string const encodedString = "", CppKeySet config = defaultConfig ())
+#ifdef __llvm__
+	__attribute__ ((annotate ("oclint:suppress[empty if statement]"), annotate ("oclint:suppress[high cyclomatic complexity]"),
```
markus2330: Does the code below really emits all these problems? Maybe we should create Macros with these attributes?

src/plugins/ccode/testmod_ccode.cpp
```diff
@@ -0,0 +1,109 @@
...
+			annotate ("oclint:suppress[high ncss method]"), annotate ("oclint:suppress[too few branches in switch statement]")))
+#endif
+{
+	CppKeySet modules{ 0, KS_END };
```
markus2330: modern C++ style :+1: 

I wonder about the spacing of the formatter, though.

src/plugins/tcl/action.cpp
```diff
@@ -19,22 +19,25 @@
 
 #include "action.hpp"
 
+#include <kdbease.h>
```
markus2330: please add a comment what it is used for (elektraKeyGetRelativeName)

src/plugins/tcl/tcl.cpp
```diff
@@ -42,15 +42,17 @@ int elektraTclGet (Plugin *, KeySet * returned, Key * parentKey)
...
 
 	int errnosave = errno;
-	std::ifstream in (keyString (parentKey), std::ios::binary); // we get our input from this file
+	std::ifstream in (parent.getString (), std::ios::binary); // we get our input from this file
 	if (!in.is_open ())
 	{
-		ELEKTRA_SET_ERROR_GET (parentKey);
+		ELEKTRA_SET_ERROR_GET (*parent);
 		errno = errnosave;
+		parent.release ();
```
markus2330: If you happen to start a new C++ plugin: the type plugin solves this much more nicely with a delegate. Then you only have single place where you have to release the ownership.

doc/news/_preparation_next_release.md
sanssecours: The Markdown Shell Recorder does not do any auto unmounting. That behavior did not change. I just removed dead code.

src/plugins/ccode/CMakeLists.txt
sanssecours: > why not add_plugin_test?

I just copied this code from the [`type` plugin](https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/type/CMakeLists.txt). The [documentation](https://github.com/ElektraInitiative/libelektra/blob/master/cmake/Modules/LibAddPlugin.cmake#L250-L251) also states that you should use `add_gtest`. My favorite approach would be to just use `add_plugin`, but as far as I can tell this function and `add_plugin_test` [do not support C++ tests for a C++ plugin](https://github.com/ElektraInitiative/libelektra/blob/master/cmake/Modules/LibAddPlugin.cmake#L86).

> Why in the dependency phase?

Adding the plugin without `if (DEPENDENCY_PHASE)…` does not work:

```
CMake Error at cmake/Modules/LibAddTest.cmake:28 (add_executable):
  add_executable cannot create target "testmod_ccode" because another target
  with the same name already exists.  The existing target is an executable
  created in source directory "/Users/rene/Documents/Studium/Master
  Thesis/Configuration Parsing/Source/Elektra/src/plugins".  See
  documentation for policy CMP0002 for more details.
  src/plugins/ccode/CMakeLists.txt:1 (add_gtest)


CMake Error at cmake/Modules/LibAddMacros.cmake:80 (target_link_libraries):
  Attempt to add link library "elektra-core" to target "testmod_ccode" which
  is not built in this directory.
  cmake/Modules/LibAddTest.cmake:34 (target_link_elektra)
  src/plugins/ccode/CMakeLists.txt:1 (add_gtest)


CMake Error at cmake/Modules/LibAddTest.cmake:40 (target_link_libraries):
  Attempt to add link library "gtest_main" to target "testmod_ccode" which is
  not built in this directory.
  src/plugins/ccode/CMakeLists.txt:1 (add_gtest)


CMake Error at cmake/Modules/LibAddTest.cmake:43 (target_link_libraries):
  Attempt to add link library "gtest" to target "testmod_ccode" which is not
  built in this directory.
  src/plugins/ccode/CMakeLists.txt:1 (add_gtest)


CMake Error at cmake/Modules/LibAddTest.cmake:51 (install):
  install TARGETS given target "testmod_ccode" which does not exist in this
  directory.
  src/plugins/ccode/CMakeLists.txt:1 (add_gtest)
```

.

src/plugins/ccode/testmod_ccode.cpp
sanssecours: I know, I prefer to only explicitly “import” the functionality I need though.

src/plugins/ccode/testmod_ccode.cpp
sanssecours: Does the code below really emits all these problems?

Yes. The main problems seem to be the macros `succeed_if_same` and `exit_if_fail`. Every time we use them OCLint prints the following two warnings:

- “too few branches in switch statement”, and
- “empty if statement”

.

> Maybe we should create Macros with these attributes?

Nah, I think it is okay as it is. These annotations should not look pretty. Ideally we should fix the underlying problems, instead of using annotations to disable the warnings.

src/bindings/cpp/tests/tests.hpp
sanssecours: Oh, I forgot to remove the trailing semicolons. This should be fixed in the latest version of the PR.

src/plugins/ccode/coder.hpp
sanssecours: I move the code into the namespace `elektra` in one fo the latest commits. Hopefully that works for you.

src/plugins/tcl/tcl.cpp
sanssecours: > If you happen to start a new C++ plugin: the type plugin solves this much more nicely with a delegate.

I know, I used the delegate approach for the [CCode plugin updates](https://github.com/ElektraInitiative/libelektra/pull/2066/files#diff-8a51e9df9c5ae96399489342adaa7158R21) (that are part of this PR).

src/plugins/ccode/CMakeLists.txt
markus2330: > add_gtest

Yes, seems like we miss a feature in add_plugin.

>  Adding the plugin without if (DEPENDENCY_PHASE)… does not work:

I mean that ADDTESTING_PHASE should be the correct phase.

sanssecours: > I mean that ADDTESTING_PHASE should be the correct phase.

After I changed the phase `add_gtest` [did not find the source files for the test at configuration time](https://travis-ci.org/ElektraInitiative/libelektra/jobs/393646495#L5642) any more. I fixed this problem by adding the code

```cmake
foreach (file ${ARG_SOURCES} ${source}.cpp)
  if (EXISTS "${file}")
    list (APPEND SOURCES "${file}")
  else ()
    list (APPEND SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/${file}")
  endif ()
endforeach ()
```

to `add_gtest`. However now the build fails printing an error message about `kdb.hpp`

```
In file included from ../src/plugins/ccode/coder.cpp:12:
../src/plugins/ccode/coder.hpp:17:10: fatal error: 'keyset.hpp' file not found
#include <keyset.hpp>
         ^~~~~~~~~~~~
```

. For now I undid the changes to fix the problem.

sanssecours:
jenkins build all please

markus2330:
Thank you, great job!

Issue #2075, opened 2018-06-14
Replace `try_compile`


ingwinlu:
https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/passwd/CMakeLists.txt#L4

This is not reliable as it will not fail when the function is not available (no -Werror that would make try_compile fail when it is not linkable. so it just warns about an implicit declaration).

markus2330:
Thank you for raising the question!

There is not really an easy answer for this question. Systems that do not have the symbol should fail during linking in try_compile. (The name try_compile is confusing, it actually also links.) So for these systems everything should work as expected.

I assume you talk about systems that have the symbol but some header file magic might (not) include the prototype. In such cases we have undefined behavior (if the prototype does nothing it works; if the prototype has some magic to select the correct symbol it will segfault at run-time.)

`-Werror` might not always be the correct choice because compilers might generate unrelated warnings (e.g. return value not used). (The `-Werror` build job will hopefully catch such problems for our compilers but not for every compiler out there.)

So I think we should only fix the build system if it is broken, meaning we actually encounter (or know that we would encounter) problems on a specific system.

markus2330:
Yes, of course it is. This is a very good suggestion! And we already use it at some places.

ingwinlu:
Updated OP

ingwinlu:
The current implementation is definitely not working: 
https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2063/52/pipeline/444/
Cmake believes the function is available. segfaults during tests.

markus2330:
Yes, checkFunctionExists is the way to go. Then such problems are (hopefully) fixed by the CMake community.

ingwinlu:
checkFunctionExists will not magically take care of setting definitions. or if you want to have another look at it: checkFuncitonExists also returns true for functions that are not exported due to missing feature macros.

#2078 has a commit in it that uses checkFunctionExists and sets the definitions accordingly.

@markus2330 how do you like this approach? can we use that in other cases as well?

ingwinlu:
I am not convinced `checkFunctionExists` is the way to go anymore as it detects functions that might not be available (depending on various things).

I will test some more in #2056 and #2078.

markus2330:
Seems like we start using CMake in non-standard ways.

ingwinlu:
@markus2330 i don't think there is a 'standard' way to use CMake except for trivial projects :p,

[...]

PR #2078, opened 2018-06-15
passwd: properly set feature flags

ingwinlu:
# Purpose

* no longer assume we build for glibc
* properly set definitions when functions are detected
* dont try_compile, instead try to find function definitions directly

# Checklist

- [x] I added unit tests
   - debian stretch
   - alpine 3.7

### DIFF
--- a/src/plugins/passwd/CMakeLists.txt
+++ b/src/plugins/passwd/CMakeLists.txt
@@ -12,12 +12,12 @@ if (DEPENDENCY_PHASE)
 	if (TEST_FGETPWENT)
 		add_definitions (-DUSE_FGETPWENT)
 	elseif (TEST_GETLINE)
-		add_definitions (-DUSE_GETLINE)
+		add_definitions (-DUSE_FGETPWENT_LOCAL)
 	else ()
 		remove_plugin (${plugin} "could not find required functions: fgetpwent || getline")
 	endif ()
 
-	# no need to remove the plugin as our fallback has no deps
+	# no need to remove the plugin as our fallbacks have no deps
 	if (TEST_PUTPWENT)
 		add_definitions (-DUSE_PUTPWENT)
 	endif (TEST_PUTPWENT)
--- a/src/plugins/passwd/CMakeLists.txt
+++ b/src/plugins/passwd/CMakeLists.txt
@@ -1,12 +1,13 @@
-include (CheckSymbolExists)
+include (SafeCheckSymbolExists)
 include (LibAddMacros)
 
 set (plugin passwd)
 
 if (DEPENDENCY_PHASE)
-	check_symbol_exists (fgetpwent "stdio.h;sys/types.h;pwd.h" TEST_FGETPWENT)
-	check_symbol_exists (getline "stdio.h" TEST_GETLINE)
-	check_symbol_exists (putpwent "stdio.h;sys/types.h;pwd.h" TEST_PUTPWENT)
+	add_definitions (-D_GNU_SOURCE -D_DARWIN_C_SOURCE)
+	safe_check_symbol_exists (fgetpwent "stdio.h;sys/types.h;pwd.h" TEST_FGETPWENT)
+	safe_check_symbol_exists (getline "stdio.h" TEST_GETLINE)
+	safe_check_symbol_exists (putpwent "stdio.h;sys/types.h;pwd.h" TEST_PUTPWENT)
 
 	if (TEST_FGETPWENT)
 		add_definitions (-DUSE_FGETPWENT)
--- a/src/plugins/passwd/CMakeLists.txt
+++ b/src/plugins/passwd/CMakeLists.txt
@@ -1,20 +1,25 @@
+include (CheckFunctionExists)
 include (LibAddMacros)
 
-if (DEPENDENCY_PHASE)
-	try_compile (TEST_FGETPWENT "${CMAKE_BINARY_DIR}" "${PROJECT_SOURCE_DIR}/src/plugins/passwd/hasfgetpwent.c")
-	try_compile (TEST_PUTPWENT "${CMAKE_BINARY_DIR}" "${PROJECT_SOURCE_DIR}/src/plugins/passwd/hasputpwent.c")
+set (plugin passwd)
 
-	if (TEST_PUTPWENT)
-		set (PUTPWENT 1)
+if (DEPENDENCY_PHASE)
+	check_function_exists (fgetpwent TEST_FGETPWENT)
+	check_function_exists (getline TEST_GETLINE)
+	check_function_exists (putpwent TEST_PUTPWENT)
+
+	if (TEST_FGETPWENT AND ALLOW_GNU_SOURCE)
+		add_definitions (-D_GNU_SOURCE)
+	elseif (TEST_GETLINE)
+		add_definitions (-D_POSIX_C_SOURCE=200809L)
 	else ()
-		set (PUTPWENT 0)
+		remove_plugin (${plugin} "could not find required functions: fgetpwent || getline")
 	endif ()
-	if (TEST_FGETPWENT)
-		set (FGETPWENT 1)
-	else ()
-		set (FGETPWENT 0)
-	endif ()
-	set (PASSWD_COMPILE_DEFS "HAS_PUTPWENT=${PUTPWENT};HAS_FGETPWENT=${FGETPWENT}")
+
+	# no need to remove the plugin as our fallback has no deps
+	if (TEST_PUTPWENT AND ALLOW_GNU_SOURCE)
+		add_definitions (-D_GNU_SOURCE)
+	endif (TEST_PUTPWENT AND ALLOW_GNU_SOURCE)
 endif ()
 
 add_plugin (passwd
--- a/src/plugins/passwd/CMakeLists.txt
+++ b/src/plugins/passwd/CMakeLists.txt
@@ -9,16 +9,16 @@ if (DEPENDENCY_PHASE)
 	check_function_exists (putpwent TEST_PUTPWENT)
 
 	if (TEST_FGETPWENT AND ALLOW_GNU_SOURCE)
-		add_definitions (-D_GNU_SOURCE)
+		add_definitions (-D_GNU_SOURCE -DUSE_FGETPWENT)
 	elseif (TEST_GETLINE)
-		add_definitions (-D_POSIX_C_SOURCE=200809L)
+		add_definitions (-D_POSIX_C_SOURCE=200809L -DUSE_GETLINE)
 	else ()
 		remove_plugin (${plugin} "could not find required functions: fgetpwent || getline")
 	endif ()
 
 	# no need to remove the plugin as our fallback has no deps
 	if (TEST_PUTPWENT AND ALLOW_GNU_SOURCE)
-		add_definitions (-D_GNU_SOURCE)
+		add_definitions (-D_GNU_SOURCE -DUSE_PUTPWENT)
 	endif (TEST_PUTPWENT AND ALLOW_GNU_SOURCE)
 endif ()
 
--- a/src/plugins/passwd/CMakeLists.txt
+++ b/src/plugins/passwd/CMakeLists.txt
@@ -1,25 +1,25 @@
-include (CheckFunctionExists)
+include (CheckSymbolExists)
 include (LibAddMacros)
 
 set (plugin passwd)
 
 if (DEPENDENCY_PHASE)
-	check_function_exists (fgetpwent TEST_FGETPWENT)
-	check_function_exists (getline TEST_GETLINE)
-	check_function_exists (putpwent TEST_PUTPWENT)
+	check_symbol_exists (fgetpwent "stdio.h;sys/types.h;pwd.h" TEST_FGETPWENT)
+	check_symbol_exists (getline "stdio.h" TEST_GETLINE)
+	check_symbol_exists (putpwent "stdio.h;sys/types.h;pwd.h" TEST_PUTPWENT)
 
-	if (TEST_FGETPWENT AND ALLOW_GNU_SOURCE)
-		add_definitions (-D_GNU_SOURCE -DUSE_FGETPWENT)
+	if (TEST_FGETPWENT)
+		add_definitions (-DUSE_FGETPWENT)
 	elseif (TEST_GETLINE)
-		add_definitions (-D_POSIX_C_SOURCE=200809L -DUSE_GETLINE)
+		add_definitions (-DUSE_GETLINE)
 	else ()
 		remove_plugin (${plugin} "could not find required functions: fgetpwent || getline")
 	endif ()
 
 	# no need to remove the plugin as our fallback has no deps
-	if (TEST_PUTPWENT AND ALLOW_GNU_SOURCE)
-		add_definitions (-D_GNU_SOURCE -DUSE_PUTPWENT)
-	endif (TEST_PUTPWENT AND ALLOW_GNU_SOURCE)
+	if (TEST_PUTPWENT)
+		add_definitions (-DUSE_PUTPWENT)
+	endif (TEST_PUTPWENT)
 endif ()
 
 add_plugin (passwd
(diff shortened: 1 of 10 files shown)


ingwinlu:
https://github.com/ElektraInitiative/libelektra/issues/2075#issuecomment-397550628
> the #error might be triggered. It would be better if we can detect this during the CMake stage and simply remove the plugin.

src/plugins/passwd/CMakeLists.txt
```diff
@@ -1,20 +1,20 @@
...
-		set (PUTPWENT 0)
-	endif ()
 	if (TEST_FGETPWENT)
-		set (FGETPWENT 1)
+		add_definitions (-D_GNU_SOURCE)
+	elseif (TEST_PUTPWENT AND TEST_GETPWENT)
+		add_definitions (-D_X_OPEN_SOURCE=500)
 	else ()
-		set (FGETPWENT 0)
+		remove_plugin (${plugin} "Could not find required functions")
```
markus2330: error message unclear: Which functions are this? Where they are provided? (I assume glibc?)

src/plugins/passwd/CMakeLists.txt
```diff
@@ -1,20 +1,20 @@
...
+	check_function_exists (getpwent TEST_GETPWENT)
 
-	if (TEST_PUTPWENT)
-		set (PUTPWENT 1)
-	else ()
-		set (PUTPWENT 0)
-	endif ()
 	if (TEST_FGETPWENT)
-		set (FGETPWENT 1)
+		add_definitions (-D_GNU_SOURCE)
```
markus2330: Could you please also shortly describe in the README.md when which function is used. (What the deps are.)

src/plugins/passwd/passwd.c
```diff
@@ -129,24 +129,26 @@ int elektraPasswdGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_
...
 	{
 		ELEKTRA_SET_ERRORF (110, parentKey, "Failed to open configuration file %s\n", keyString (parentKey));
 		return -1;
 	}
 	while ((pwd = fgetpwent (pwfile)) != NULL)
-#else
+#elif defined(_X_OPEN_SOURCE)
 	while ((pwd = getpwent ()) != NULL)
+#else
+#error Need either _GNU_SOURCE or _X_OPEN_SOURCE
```
markus2330: please improve error message: doesn't this case also mean that there is a bug in CMake code?

src/plugins/passwd/CMakeLists.txt
ingwinlu: `check_function_exists` already logs output hence I did not specify the functions here again:
```
-- Looking for fgetpwent
-- Looking for fgetpwent - found
-- Looking for putpwent
-- Looking for putpwent - found
-- Looking for getpwent
-- Looking for getpwent - found
-- Include Plugin passwd
```
And no not glibc: it just needs to be available in the include paths (typically it is provided by your libc implementation, if you are use something at least somewhat POSIX).

src/plugins/passwd/passwd.c
ingwinlu: yes this should not be reached. Should I replace it with something else?

src/plugins/passwd/CMakeLists.txt
markus2330: You only see this output in the first cmake run. In the later runs you only get the error. So the error message needs to describe what the reason is.

src/plugins/passwd/passwd.c
markus2330: Simply say that CMake should make sure that this is not reached and that this is a bug in Elektra. (ideally with URL to issues.libelektra.org)

src/plugins/passwd/CMakeLists.txt
ingwinlu: Thanks for that, I just noticed putpwent is a terrible choice of function and should probably be replaced with setpwent.

src/plugins/passwd/passwd.c
ingwinlu: aa27585

src/plugins/passwd/CMakeLists.txt
ingwinlu: * ed28db3 
* 46c545d

src/plugins/passwd/README.md
```diff
@@ -15,7 +15,16 @@ This plugin parses `passwd` files, e.g. `/etc/passwd`.
 
 ## Implementation Details
 
-If present, the not-posix compliant `fgetpwent` function will be used to read the file supplied by the resolver, otherwise `getpwent` will be used. For writing, if present, `putpwent` will be used, if not a simple implementation writing straight to the config file.
+If present, the not-posix function `fgetpwent` (GNU_SOURCE) will be used to
+read the file supplied by the resolver.
+Alternatively `getpwent` (X_OPEN_SOURCE >= 500) will be used.
+
+For writing, if present, `putpwent` (DEFAULT_SOURCE) will be used, if not a
```
sanssecours: This sentence seems to be a little complicated. Maybe you could use two sentences instead of one:

> For writing, if present, `putpwent` (DEFAULT_SOURCE) will be used. If `putpwent` is not available, the plugin will write straight to the config file.

.

src/plugins/passwd/passwd.c
```diff
@@ -129,24 +129,26 @@ int elektraPasswdGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_
...
 	{
 		ELEKTRA_SET_ERRORF (110, parentKey, "Failed to open configuration file %s\n", keyString (parentKey));
 		return -1;
 	}
 	while ((pwd = fgetpwent (pwfile)) != NULL)
-#else
+#elif defined(_X_OPEN_SOURCE)
 	while ((pwd = getpwent ()) != NULL)
+#else
+#error Configuration error in CMakeLists.txt. Neither _GNU_SOURCE nor _X_OPEN_SOURCE were provided. Open an issue at https://issue.libelektra.org.
```
sanssecours: I would add the word “please” here:

> Please open an issue at https://issue.libelektra.org.

.

ingwinlu:
@sanssecours seems like the plugin can not work if there is no fgetpwent (or it will only work if you want to read the systems passwd)......

ingwinlu:
@sanssecours can you play around with the current `master` version a bit? I have a hard time believing that it is working as intended.

I wrote a fallback function that parses passwd files in b079ca8 but my C is not the best so not sure if we want to solve it that way.

sanssecours:
> can you play around with the current `master` version a bit?

I tried a slightly modified version of the example code from the ReadMe:

```sh
kdb mount /etc/passwd system/passwd passwd index=name
kdb export system/passwd/root yamlcpp
#> gecos: System Administrator
#> gid: 0
#> home: /var/root
#> passwd: "*"
#> shell: /bin/sh
#> uid: 0
```

, which seems to work as intended. If you want other debug output, please just write down the commands I should test and I will try to run them on my machine.

Another good option to debug the plugin might be a Markdown Shell Recorder test. Even if the test fails, the Shell Recorder should inform you about the output and return values of all commands you included in the test.

sanssecours:
> but if you would mount a different password file (at another location) it would probably still return the values of the original /etc/passwd as the used getpwent does not accept a path to the new file.

As far as I can tell, yes, that is the case:

```sh
printf 'bla' > test.passwd
kdb mount "$PWD/test.passwd" system/passwd passwd index=name
kdb export system/passwd/root yamlcpp
#> gecos: System Administrator
#> gid: 0
#> home: /var/root
#> passwd: "*"
#> shell: /bin/sh
#> uid: 0
```

.

tom-wa:
this implementation fails for 
`messagebus:x:104:111::/var/run/dbus:/bin/false`
(strtok skips consecutive delimiters)
i currently use musl's fgetpwent_r.c code in an other version of this plugin for some testings, works fine.

ingwinlu:
i can not reproduce the error that we have on apple machines locally. can somebody enlighten me on where the signed int comes from?
```
/Users/travis/build/ElektraInitiative/libelektra/src/plugins/passwd/passwd.c:64:19: error: comparison of integers of different signs: 'int' and 'unsigned int' [-Werror,-Wsign-compare]
        for (x=0; **s-'0'<10U; ++*s) x=10*x+(**s-'0');
                  ~~~~~~~^~~~
```

ingwinlu:
someone send me the following link:
https://stackoverflow.com/questions/46463064/what-causes-a-char-to-be-signed-or-unsigned-when-using-gcc

src/plugins/passwd/CMakeLists.txt
```diff
@@ -1,20 +1,26 @@
...
-		set (PUTPWENT 0)
-	endif ()
 	if (TEST_FGETPWENT)
-		set (FGETPWENT 1)
+		add_definitions (-DUSE_FGETPWENT)
+	elseif (TEST_GETLINE)
+		add_definitions (-DUSE_FGETPWENT_LOCAL)
 	else ()
-		set (FGETPWENT 0)
+		remove_plugin (${plugin} "could not find required functions: fgetpwent || getline")
```
sanssecours: I think it would make sense to use “or” instead of “||”. Probably everyone knows what the first term means, but only a few (programmers) know the second term. I would write something like:

> could not find one or both of the required functions: `fgetpwent`, `getline`

. Maybe it even makes sense to add one `remove_plugin` for each function to improve the output in case of a failure.

src/plugins/passwd/README.md
```diff
@@ -15,7 +15,17 @@ This plugin parses `passwd` files, e.g. `/etc/passwd`.
 
 ## Implementation Details
 
-If present, the not-posix compliant `fgetpwent` function will be used to read the file supplied by the resolver, otherwise `getpwent` will be used. For writing, if present, `putpwent` will be used, if not a simple implementation writing straight to the config file.
+The not-POSIX function `fgetpwent` (GNU_SOURCE) will be used to
```
sanssecours: Typo: `not-POSIX` → `non-POSIX`

src/plugins/passwd/CMakeLists.txt
ingwinlu: > I think it would make sense to use “or” instead of “||”. 

agreed

> Maybe it even makes sense to add one remove_plugin for each function to improve the output in case of a failure.

That would imply that both are required right? But the implementation with getline is a replacement if fgetpwent can not be found hence the single message as you only need one of the functions.

ingwinlu:
@sanssecours thx. I kind of want to test the fallback implementation more so I will probably add a testcase comparing the fgetpwent to fgetpwent_local that runs when both are available.

src/plugins/passwd/CMakeLists.txt
sanssecours: > hat would imply that both are required right?

Yep.

> But the implementation with getline is a replacement if fgetpwent can not be found hence the single message as you only need one of the functions.

In that case I would write something like:

> could not find `fgetpwent` nor `getline`

.

src/plugins/passwd/README.md
```diff
@@ -15,7 +15,17 @@ This plugin parses `passwd` files, e.g. `/etc/passwd`.
 
 ## Implementation Details
 
-If present, the not-posix compliant `fgetpwent` function will be used to read the file supplied by the resolver, otherwise `getpwent` will be used. For writing, if present, `putpwent` will be used, if not a simple implementation writing straight to the config file.
+The non-POSIX function `fgetpwent` (GNU_SOURCE) will be used to
+read the file supplied by the resolver.
+As a fallback we implemented our own version which might not behave correctly.
```
markus2330: What does this mean? "might not behave correctly"

src/plugins/passwd/passwd.c
```diff
@@ -49,6 +49,79 @@ static int validatepwent (struct passwd * pwd)
 	return 0;
 }
 
+#if defined(USE_FGETPWENT_LOCAL)
+// clang-format off
+/* Taken from musl libc
```
markus2330: This is MIT licensed, we need to document it in debian/copyright of the debian branch.

src/plugins/passwd/passwd.c
```diff
@@ -130,26 +203,25 @@ int elektraPasswdGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_
...
 		ELEKTRA_SET_ERRORF (110, parentKey, "Failed to open configuration file %s\n", keyString (parentKey));
 		return -1;
 	}
+#if defined(USE_FGETPWENT)
 	while ((pwd = fgetpwent (pwfile)) != NULL)
+#elif defined(USE_FGETPWENT_LOCAL)
+	while ((pwd = fgetpwent_l (pwfile)) != NULL)
 #else
-	while ((pwd = getpwent ()) != NULL)
+#error Configuration error in CMakeLists.txt. Neither fgetpwent nor getline were provided. Please open an issue at https://issue.libelektra.org.
```
markus2330: How can this be if we have our own build-in version __getpwent_a?

src/plugins/passwd/README.md
```diff
@@ -15,7 +15,17 @@ This plugin parses `passwd` files, e.g. `/etc/passwd`.
...
+
+For writing `putpwent` (GNU_SOURCE) will be used.
+If it is not available the plugin will write straight to the config file.
+
+## Requirements
+
+For the plugin to be build at least `POSIX_C_SOURCE >= 200809L` compatibility
```
markus2330: Also if built-in versions are used?

src/plugins/passwd/CMakeLists.txt
```diff
@@ -1,20 +1,26 @@
...
+		add_definitions (-DUSE_FGETPWENT_LOCAL)
 	else ()
-		set (FGETPWENT 0)
+		remove_plugin (${plugin} "could not find fgetpwent nor getline.")
 	endif ()
-	set (PASSWD_COMPILE_DEFS "HAS_PUTPWENT=${PUTPWENT};HAS_FGETPWENT=${FGETPWENT}")
+
+	# no need to remove the plugin as our fallbacks have no deps
+	if (TEST_PUTPWENT)
+		add_definitions (-DUSE_PUTPWENT)
```
markus2330: It is better if you add this as COMPILE_DEFINITIONS in add_plugin.

src/plugins/passwd/passwd.c
```diff
@@ -49,6 +49,79 @@ static int validatepwent (struct passwd * pwd)
...
+	return x;
+}
+
+int __getpwent_a(FILE *f, struct passwd *pw, char **line, size_t *size, struct passwd **res)
+{
+	ssize_t l;
+	char *s;
+	int rv = 0;
+	int cs;
+	pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &cs);
```
markus2330: find_package (Threads QUIET)
LINK_LIBRARIES ${CMAKE_THREAD_LIBS_INIT}
needed (see resolver)

src/plugins/passwd/README.md
ingwinlu: Because of the dep to getline()

src/plugins/passwd/README.md
ingwinlu: we do not test for full correctness with our backup implementations. hence there might be hidden issues when those fallbacks are used.

src/plugins/passwd/passwd.c
ingwinlu: because __getpwent_a needs getline. it still should not trigger because the plugin should be removed by cmake. It is just a safety net.

src/plugins/passwd/README.md
markus2330: Maybe telling that we use the musl implementation is more helpful?

src/plugins/passwd/passwd.c
ingwinlu: can we remove the pthread references all together? Does not seem worth it for the additional dependency.

src/plugins/passwd/CMakeLists.txt
ingwinlu: then the safe_check_symbol_exists will not work

src/plugins/passwd/passwd.c
markus2330: I did not look into what they do. But you have a guarantee that the same instance of a plugin is not called from multiple threads at the same time.

src/plugins/passwd/CMakeLists.txt
markus2330: Ok

markus2330:
Thank you for making the test logic easier.

PR #2094, opened 2018-06-18
Fix archivation

ingwinlu:
The space in buildDir can not be escaped during `cp` as it does not allow the shell to expand `*`.


markus2330:
Thank you for the quick fix!

PR #2107, opened 2018-06-21
Elektra Web 1.6

### LINKED ISSUE #2103: Web UI 1.6
:+1: many bugs are resolved now but there are still some bugs:

- [x] creating array bug:
 - click new key
 - enter "array" as name
 - press "CREATE ARRAY"
 - *do not open array* and press "+" for a subkey
 - BUG: `#0` not suggested as name
- [x] creating two arrays bug:
 - create a new array next to the old array (call it "array2")
 - add a subkey to it (`#1`)
 - BUG: when refreshing a session in another browser (F5) OR when trying to add a key in the other array, an already existing number is taken
 - (a simple way to reproduce: add keys in two arrays alternating)
- [x] set values bug
  - change values of array entry
  - add a new array entry
  - BUG: wrong name
- [x] arrays with different visibility as array members does not work:
  - create "array"
  - add `#1`
  - add `#2``and give it visibility "important"
  - click "configure metadata"
  - BUG: visibility is now "user"
- [x] endless "Loading instance... please wait" when http://localhost:33334/instances/X does not exist
- [ ] ~~undo on metadata still does not work~~
 - e.g. create a metakey and click "restrict/binary"
 - after undo both are still there, then abort
 - next undo removes the metakey, even though it was done in the different order
- [x] copying still does not copy (pressing ctrl is the accepted default afaik)
- [x] no error message if no "kdb find" present
- [ ] ~~not possible to click somewhere in the text (of values)~~ -> #2105

## Important functionalities missing:

- [x] there should be detailed information in showing the full error (for bug reports and similar)
- [x] removing all keys directly in namespaces missing

## Demo:

- [x] very little mountpoints and data present in Live Demo

## Performance/Concurrency:

Something seems to be very wrong how elektrad works. It seems to issue way to many `kdb` commands. For example:

- [x]  `kdb find` returns in about 1.5 seconds, elektrad nearly brought my system down (16GB ram swapped), many `kdb getmeta` were running concurrently. Only killing elektrad helps.
- [x]  You should at least find out what is wrong here.

omnidan:
Includes improvements from #2103 

I tested everything with Chrome and Firefox ESR. 

The demo now includes the real world config (at `user/realworld`).

I tested search with the real world config and it did freeze my browser tab when entering a query with a lot of matches (such as "a"). This was a result of a feature that automatically unfolded all search results (which isn't such a good idea for many results).

I also bumped the version and updated the release notes.

I will rebuild the docker images (from `master`) after the PR gets merged.

closes #2103


markus2330:
Thank you! Let us test this today!

Should I merge it in advance so that we can have it on webui.libelektra.org?

@ingwinlu Is there a manual action necessary to update webui.libelektra.org or will it automatically fetch the latest docker image once @omnidan pushed the image?

> I tested search with the real world config and it did freeze my browser tab when entering a query with a lot of matches (such as "a").

Can you maybe delay the unfolding for some time? Or simply not do it as long as there are too many matches? On the one hand, it is likely that the user will continue typing anyway. On the other hand, a "unfold all" feature might be useful in some cases.

omnidan:
> Can you maybe delay the unfolding for some time? Or simply not do it as long as there are too many matches? On the one hand, it is likely that the user will continue typing anyway. On the other hand, a "unfold all" feature might be useful in some cases.

I can make it unfold if there are less than 10-20 matches (I will test when it starts getting too slow). I agree, it is a useful feature if there are only a handful of results.

I will still implement this and then you can merge and I will rebuild the images.

markus2330:
Unfortunately it is not easily possible to determine a number that is always valid. E.g. Firefox ESR is much slower than a recent Firefox/Chrome.

An "unfold more" or even unfolding while scrolling would be the best solution. But I have to admit that this would be a new feature.

omnidan:
> Unfortunately it is not easily possible to determine a number that is always valid. 

You are right, I will set it to a very low number then (like 10). That should not cause any problems.

ingwinlu:
@markus2330 as I pointed out previously there is a mechanism in place that keeps the deployed image in sync with whatever is `elektra/web:latest` it should start deployment soon after the tag is updated.

omnidan:
- [ ] show values next to keys with subkeys on hover
- [ ] show lines between namespaces (or display namespaces in a special way, e.g. larger)
- [ ] search without results causes error now

omnidan:
@markus2330 everything should be done now. I will rebuild the docker images when the PR gets merged.

markus2330:
Thank you for the great work!

PR #2125, opened 2018-07-27
YAML CPP: Do Not Add Intermediate Keys 

sanssecours:
# Checklist

- [x] The PR contains updated documentation


sanssecours:
jenkins build elektra please

sanssecours:

sanssecours:
> Ideally we should have guidelines for how parsers should interpret configuration files.

I **might** add documentation about this in a separate PR.

markus2330:
You can also document it in your thesis and we transfer it afterwards. But we need guidelines, the inconsistencies between storage plugins is bothering everyone.

markus2330:
Thank you, great job! It would also be interesting how intermediate keys are represented (if they are stored on purpose, e.g. with empty values).

sanssecours:
> It would also be interesting how intermediate keys are represented (if they are stored on purpose, e.g. with empty values).

This does not work in the moment. Currently I try to fix this problem in the [Directory Value](https://www.libelektra.org/plugins/directoryvalue) plugin, which does not support empty or null keys properly yet.

PR #2127, opened 2018-07-27
YAJL: Use Logger Instead of `printf` Statements

sanssecours:
# Checklist


markus2330:
Thank you so much for taking care of yajl!

PR #2130, opened 2018-07-28
Export opmphm symbols

mpranj:
Export opmphm symbols. Needed for mmapstorage and possibly other plugins.

## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -274,6 +274,7 @@ Thanks to Daniel Bugl.
 ### General
 
 - replaced strdup with elektraStrDup (for C99 compatibility) *(Markus Raab)*
+- Added opmphm to exported symbols. *(Mihael Pranjić)*
 
 - <<TODO>>
 
--- a/src/libs/elektra/libelektra-symbols.map
+++ b/src/libs/elektra/libelektra-symbols.map
@@ -5,6 +5,7 @@
 		kdb*;
 		key*;
 		ks*;
+		opmphm*;
 
 		extern "C++" {
 			*elektra::*;
--- a/src/libs/elektra/libelektra-symbols.map
+++ b/src/libs/elektra/libelektra-symbols.map
@@ -5,7 +5,7 @@
 		kdb*;
 		key*;
 		ks*;
-		opmphm*;
+		opmphmDel;
 
 		extern "C++" {
 			*elektra::*;


markus2330:
It would be better if we do not need to entangle opmphm and the mmap storage.

Do you need anything else other then freeing the opmphm structure?

@KurtMi Can you maybe simplify the structure that a single elektraFree does the job? (It seems like n is static for the lifetime of an OpmphmGraph anyway?)

If reducing to a single elektraFree is not possible, we should open up the needed functions by renaming them to elektra*. (Or wrapping them.)

mpranj:
If opmphm is enabled, I definitely need at least a way to free the structure.

If we want to store the opmphm inside mmap too, it will be inherently entangled and increase the complexity of mmapstorage/cache. Not storing the opmphm will simplify my work, but might be wasteful in the long run. (mmap global cache that always throws away opmphm and the opmphm is then rebuilt every time)

Regarding needed functions: I might benefit from opmphmCopy too.  I don't know if I need anything else yet.

markus2330:
@KurtMi Can you either:

1. provide elektraOpmphmNew/Copy/Del which allocates/copies/frees everything related to opmphm (including the predictor)
2. move everything related to opmph into a single structure on which we can call elektraFree and elektraStrDup.

Option 2. would be faster but we can certainly live very well with 1. But it would be important that we agree on an interface so that @mpranj does not lose even more time on this.

KurtMi:
Hello, since entangling opmphm and the mmap storage is not wanted, I am assuming an ``#include <kdbopmphm.h>`` in the plugin is not an option. I would be nice to store the opmphm is the mmap but I would need more background information. @mpranj I think the fastest way to resolve this issue you just  call me tomorrow after 12pm!

mpranj:
@KurtMi sure we can talk about it. I might have some other small questions about how opmphm works too. 

To persist the opmphm structures in mmapstorage, I will **need** to include at least the struct definitions from `<kdbopmphm.h>`. I don't see a way around that. 

I can think of roughly two directions/approaches we could take here:

1. The minimal "entanglement": KeySets contain a pointer that can just be free()d, **without** persisting the opmphm structures.
2. Medium-entanglement: mmapstorage knows about the opmphm structs and can persist them properly, including how to free and copy the structures.

While the first is probably much easier for me to implement, the second approach should be superior in the long run.

markus2330:
Thank both for you that you are eager to find a solution for that. Of course the persistent feature is more important then coupling. If the symbols are needed, it is okay to use them. But then they should be prefixed with elektra*.

Btw. we will most likely meet in two weeks (9th of August). But if you can find a solution today, then go for it :smile:

KurtMi:
@markus2330 We have discussed the topic on the phone and could not find a proper solution.
It would be best if we three would discuss this topic together, on a firm basis (both of our works in a firm state). We agreed that storing the opmphm in the mmap is the best and the target solution, but for 
now we have found a much simpler and less time consuming solution, that is:
The mmap will for now not store the opmphm components, to prevent memory leaks I will provide an function in the keyset named `elektraFreeOptimizations (Keyset * ks)` that
frees all my components.
I will create an issue for this topic so we will handle it on a later state of our work.
What do you think? I will wait with the creation of the function until you approve it.

markus2330:
> both of our works in a firm state

Is there still something you might need to change in basic data structure?

> for now [...] frees all my components.

I would prefer if we can avoid temporary hacks, especially hacks which require APIs and coordination between you two. If elektraFreeOptimizations is also needed in a final solution, of course you can add it.

@KurtMi I think the most efficient solution would be if you change the layout of your data structure so that everything resides within one block of memory. If the number of vertex/edges is somehow bound that should be relatively easy. @mpranj could then simply fix the pointers and call elektraFree like he does for the key array, the names, the values and so on.

Alternatively, you could provide a elektraOptimizationsSerialize (char *) where you serialize your structure. For this option, you would also need unserialize and free methods. Obviously this would be less efficient.

> I will create an issue for this topic so we will handle it on a later state of our work.

Yes, creating an issue sounds like a good idea. If you find other ways for going forward, please list them.

markus2330:
> Since the hashFunctionSeeds and graph vary in size and are malloced at different places. How should the one block of memory look, just a void * ?

You can split it up in arrays with elements of the same type,

> and is the size of that block also needed?

I think so but let us better confirm this by @mpranj.

> Beside this option would force me to redo the benchmarks with a runtime >24h.

No, you certainly do not need to do this. Benchmarks are obviously only valid for the one commit where you did them.

> shape of my structures

Yes, that why the shape should be as simple as possible, i.e., array.

mpranj:
> I would prefer if we can avoid temporary hacks, [...]

The `get()` function of mmapstorage replaces the KeySet struct members of the user's keyset to point to a mmap region. Therefore anything previously allocated needs to be free()d.

Currently I do a `ksClose() `on the KeySet, and then replace the KeySet struct contents with mmap. This only invalidates the OPMPHM. Similarly, dump does a `ksClear()` in that case, because it needs the allocated key array.

Now that I mention that, maybe this is enough and I don't need to call a free() function on the opmphm at all.

> and is the size of that block also needed?

Yes, I need to know the size of anything I write to a mapped region. I actually need to know the complete size of the mmap region in advance, since mmap resizing is not POSIX.

mpranj:
After trying to work without `opmphmDel()`, it did not work. I use ksDeepDup to copy a KeySet out of an mmap region (unlinking). To avoid `opmphmDel()` I would have to re-write ksDeepDup inside my plugin, or change it in the core (-> additional KS flag to not copy opmphm structures in certain cases only).

We can open the discussion in a new issue if you want.

markus2330:
ksDeepDup is a trivial piece of code. If it does not do what you need, please copy+rewrite it for your plugin as you need it.

I do not understand why you need a ksDeepDup, though. The core already does ksDeepDup for you. And when passing back the KeySet, it should work without ksDeepDup?

mpranj:
I need it to unlink a KeySet. It does most of the copy job for me.

mpranj:
Solved by rewriting ksDeepDup in the plugin.

PR #2134, opened 2018-07-28
 Directory Value: Add Leafs For Empty & Null Keys

sanssecours:
# Checklist

- [x] This PR contains updated documentation.

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -207,6 +207,8 @@ Thanks to Daniel Bugl.
 
 - The [YAJL Plugin](http://libelektra.org/plugins/yajl) now uses the internal logger functionality instead of `printf` statements.
   *(René Schwaiger)*
+- We fixed a problem with negative values reported by the
+  [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html). *(René Schwaiger)*
 
 ### Yan LR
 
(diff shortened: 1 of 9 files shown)


sanssecours:
> Stretch has only yaml-cpp 0.5. Is it possible to somehow support it, e.g. by simply excluding the offending tests?

That is possible, but I do not want to waste time supporting a (slightly broken) version of yaml-cpp. It already took me some time finding the cause of the problems on the build servers.

> Alternatively we could provide a yaml-cpp 0.6 backport for stretch?

Updating the package would be the proper option. Unfortunately it seems like Debian does not offer `yaml-cpp 0.6` yet. The [current Debian version of yaml-cpp](https://packages.debian.org/sid/libyaml-cpp-dev) (version `0.5.2`) was released over [**three years ago**](https://github.com/jbeder/yaml-cpp/releases/tag/release-0.5.2).

markus2330:
>  That is possible, but I do not want to waste time supporting a (slightly broken) version of yaml-cpp. It already took me some time finding the cause of the problems on the build servers.

Yes, I know that such problems are not easy to find. The question is if it would need much *additional* time to disable the tests for 0.5 (and document the limitation).

> Updating the package would be the proper option.

Not in Debian Stable point of view :smile: 

Not supporting 0.5 would mean to not support Debian unless we somehow backport it ourself. And I am afraid that this is much more work than supporting 0.5.

markus2330:
Ok. Thank you for the many fixes!

So we can test yaml-cpp via the Alpine docker images. I won't be able to use the plugin on my own machines, though.

PR #2137, opened 2018-07-29
Documentation: Add Array Tutorial

sanssecours:
# Checklist

- [x] This pull request contains a new Markdown Shell Recorder test.

### DIFF
--- a/doc/tutorials/arrays.md
+++ b/doc/tutorials/arrays.md
@@ -1,0 +1,135 @@
+# Arrays
+
+## Key-Value Pairs
+
+The main building block of Elektra’s database are hierarchical [key-value pairs](https://en.wikipedia.org/wiki/Key-value_database). You can create such a pair using [`kdb set`](../doc/help/kdb-set.md):
+
+```sh
+kdb set user/tests/parent value
+#> Create a new key user/tests/parent with string "value"
+```
+
+. The command above created a key `user/tests/parent` with the value `value`. Since Elektra uses a hierarchical database we can also create keys **below** `user/tests/parent`:
+
+```sh
+# If we do not provide a value, then `kdb set` creates keys with `null` values.
+kdb set user/tests/parent/son
+#> Create a new key user/tests/parent/son with null value
+
+kdb set user/tests/parent/daughter
+#> Create a new key user/tests/parent/daughter with null value
+
+kdb set user/tests/parent/daughter/grandchild
+#> Create a new key user/tests/parent/daughter/grandchild with null value
+```
+
+. We can check the hierarchy of the keys using `kdb ls` and retrieve data using `kdb get`:
+
+```sh
+# Elektra sorts keys alphabetically
+kdb ls user/tests/parent
+#> user/tests/parent
+#> user/tests/parent/daughter
+#> user/tests/parent/daughter/grandchild
+#> user/tests/parent/son
+
+kdb get user/tests/parent
+#> value
+kdb get user/tests/parent/daughter
+#>
+```
+
+.
+
+## Array Keys
+
+Since Elektra sorts keys alphabetically we can use the key-value pair structure described above to store sequences of values (aka. arrays).
+
+### Empty Arrays
+
+For an **empty array** (`[]`) we just add the [meta-key](../help/elektra-metadata.md) `array`:
+
+```sh
+# Create an empty array with the name `user/tests/sequence`
+kdb setmeta user/tests/sequence array ''
+```
+
+.
+
+### Array Elements
+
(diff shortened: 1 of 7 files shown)


doc/tutorials/arrays.md
```diff
@@ -0,0 +1,135 @@
...
+
+ruby -e 'print("#", "_" * (ARGV[0].length - 1), ARGV[0])' 42
+#> #_42
+```
+
+.
+
+### Metadata
+
+To make life easier for users, Elektra’s [storage plugins](plugins.md) should save the [basename of the last key in the array parent](../decisions/array.md). Some plugins do not support this feature. In that case you can add this value manually via `kdb setmeta`:
```
markus2330: I would say it is a bug if they do not do it, not an unsupported feature.

doc/tutorials/arrays.md
```diff
@@ -0,0 +1,135 @@
...
+
+```sh
+# Add array elements
+kdb set user/tests/favorites/superheros/#0 'One-Punch Man'
+kdb set user/tests/favorites/superheros/#1 'Mermaid Man and Barnacle Boy'
+kdb set user/tests/favorites/superheros/#____99999 'The guy with the bow and arrow'
+
+# The metakey `array` should save the basename of the last element.
+kdb setmeta user/tests/favorites/superheros array '#____99999'
```
markus2330: Nice example but the main reason is iteration over all array members.

tests/shell/shell_recorder/tutorial_wrapper/CMakeLists.txt
```diff
@@ -1,13 +1,11 @@
 include (LibAddTest)
 
-add_msr_test (msr_syntax "${CMAKE_SOURCE_DIR}/tests/shell/shell_recorder/tutorial_wrapper/SyntaxCheck.md" REQUIRED_PLUGINS ini)
-add_msr_test (ini_crash_test "${CMAKE_SOURCE_DIR}/src/plugins/ini/crash_test/crash_test.md" REQUIRED_PLUGINS ini)
-add_msr_test (kdb_global_umount "${CMAKE_SOURCE_DIR}/doc/help/kdb-global-umount.md" REQUIRED_PLUGINS spec tracer)
+# ========
+# = Help =
```
markus2330: Do you want to say "help needed" here?

doc/tutorials/arrays.md
sanssecours: In that case our default storage plugin (`dump`) is buggy:

```sh
kdb setmeta user/tests/array array ''
kdb set user/tests/array/#0 first
kdb getmeta user/tests/array array
#>
```

. By the way, since we are already talking about it, I think it would make sense to specify the features a storage plugins supports (e.g. arrays, comments, metadata, …) via the contract of the plugin. We could then add tests for each of these clauses, to check if the plugin supports a feature properly.

doc/tutorials/arrays.md
sanssecours: How does this help us to iterate over array elements (, if we support “empty” array elements)? For example, the array above has only three (non-empty) members, while the size of the array is `100000`. I do not think it makes sense to iterate `100000` times in such a case…

tests/shell/shell_recorder/tutorial_wrapper/CMakeLists.txt
sanssecours: Nope, the commands below this banner generate MSR tests for documents in **`doc/help`**.

doc/tutorials/arrays.md
markus2330: In Elektra 0.7 we tried to specify the features of storage plugins but this was not very useful. It is much too difficult for applications to find out if a storage plugin is suitable for their needs.

Instead plugins should now specify the features they lack in `infos/needs`. And special plugins should work around these problems. (like you did with directoryvalue)

So for the problem you mentioned, a plugin should check for validity of arrays. This is, however, more important for yajl than dump, where you currently can even corrupt configuration files if you mix arrays with other keys. But yes, also for dump we should reject broken arrays (or fix them in simple scenarios like you described).

doc/tutorials/arrays.md
markus2330: Seems like we did not fully thought that through.

Especially if arrays are distributed between different namespaces it gets quite tricky.

sanssecours: > Seems like we did not fully thought that through.

In that case, can I leave this PR as it is or should I update the text?

markus2330: It is a very good starting point. Thanks again for writing the tutorial!

Issue #2142, opened 2018-07-30
pluginprocess: transfer of parentKey


markus2330:
It seems like some arguments are not transfered correctly.

@e1528532 Do you have time for a quick fix? It would be highly appreciated. It should be quite easy to fix and if the problem is like I think it is you will also trap into it with your plugins. It might, however, only a misuse in #2135.

## Steps to Reproduce the Problem

Debug python plugin of PR #2135 and print the name of parentKey.

## Expected Result

That parentKey is fully transmitted to the other process.

## Actual Result

parentKey is missing its name.

## System Information

- Elektra Version: master or PR #2135 to reproduce

## Further Info

Shortly looking at src/libs/pluginprocess/pluginprocess.c line 199 there seems to be the assumption that no key with the name of the parentKey exists in the originalKeySet. This assumption is not valid, it is easily possible that the key also exists in the keyset. /pluginprocess/original only transfers the name, but not the other information of the parentKey.

markus2330:
@e1528532 Do you have time to look at it? A (partial?) fix is done in the python PR #2135

e1528532:
fixed in #2149

PR #2149, opened 2018-08-02
Pluginprocess improvements

e1528532:
## Basics

Improvements to the pluginprocess library as requested in #2135 and #2142 .

## Checklist

- [X] I added unit tests

## Review

@markus2330 please review my pull request. I will try to apply this to the python plugin as well to see if it helps


src/libs/pluginprocess/pluginprocess.c
```diff
@@ -6,6 +6,41 @@
  * Executes plugins in a separate process via fork and uses a simple
  * communication protocol based on the dump plugin via named pipes.
  *
+ * The communication protocol works as follows, where Child and Parent stand
+ * for the child and the parent process:
+ * 1)  Two named pipes are created, called commandPipe and resultPipe
+       commandPipe is used for a keyset containing metainformation about
```
markus2330: use markdown enumeration here (doxygen understands it), so
```
1.  Two named pip...
    - commandPipe
    - resultPipe
```

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -6,6 +6,41 @@
...
+ *     - /pluginprocess/parent
+ *       a copy of the parent key passed to the plugin interface
+ *     - /pluginprocess/parent/name
+ *       the name of the parent key passed to the plugin interface
+ *     - /pluginprocess/command
+ *       the command that should be executed by the child process
+ *     - /pluginprocess/payload/exists
+ *       1 if a keyset will be sent over the resultPipe, 0 otherwise
+ *     - /pluginprocess/version
+ *       a simple string indicating the version of this communication
```
markus2330: Why not simply using a number?

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -71,18 +106,26 @@ void elektraPluginProcessStart (Plugin * handle, ElektraPluginProcess * pp)
...
 	Key * resultPipeKey = keyNew ("/pluginprocess/pipe/result", KEY_VALUE, pp->resultPipe, KEY_END);
-	KeySet * keySet = ksNew (0, KS_END);
 	int counter = 0;
 
 	do
 	{
-		ELEKTRA_LOG_DEBUG ("Child: Wait for a KeySet");
-		elektraInvoke2Args (pp->dump, "get", keySet, commandPipeKey);
-		ELEKTRA_LOG_DEBUG ("Child: We received a KeySet with %zd keys in it", ksGetSize (keySet));
+		KeySet * commandKeySet = ksNew (6, KS_END);
```
markus2330: Why 6? What about sending the size instead of "exists"?

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -6,6 +6,41 @@
...
+       resultPipe is used for transferring the actual payload that gets passed
+       to a plugin
+ * 2)  The parent constructs a keyset called commandKeySet, containing:
+ *     - /pluginprocess/parent
+ *       a copy of the parent key passed to the plugin interface
+ *     - /pluginprocess/parent/name
+ *       the name of the parent key passed to the plugin interface
+ *     - /pluginprocess/command
+ *       the command that should be executed by the child process
+ *     - /pluginprocess/payload/exists
```
markus2330: What about sending the size instead of a bool? See below.

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -396,26 +470,71 @@ int elektraPluginClose (Plugin * handle, Key * errorKey)
...
+ * has to be stored using this function like
+ * @code
+int elektraPluginOpen (Plugin * handle, Key * errorKey)
+{
+	ElektraPluginProcess * pp = elektraPluginGetData (handle);
+	if (pp == NULL)
+	{
+		if ((pp = elektraPluginProcessInit (errorKey)) == NULL) return ELEKTRA_PLUGIN_STATUS_ERROR;
+		ArbitraryPluginData * data = // initialize your plugin data
+		elektraPluginProcessSetData (handle, data);
```
markus2330: Ahh, thank you. That makes it clear how to use it.

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -396,26 +470,71 @@ int elektraPluginClose (Plugin * handle, Key * errorKey)
...
+{
+	ElektraPluginProcess * pp = elektraPluginGetData (handle);
+	if (elektraPluginProcessIsParent (pp)) {
+		ArbitraryPluginData * data = elektraPluginProcessGetData (handle);
+		ElektraPluginProcessCloseResult result = elektraPluginProcessClose (pp, errorKey);
+		if (result.cleanedUp)
+		{
+			elektraPluginSetData (handle, NULL);
+			// cleanup your plugin data here
+		}
```
markus2330: else? Or what do we need the cleanedUp for?

e1528532: cleanedUp indicates when the whole process is over and thus ElektraPluginProcess data got cleared already, it is then time to cleanup any plugin-related data if there is some. The else action is simply doing nothing

src/libs/pluginprocess/pluginprocess.c
e1528532: yes that was the main mistake i've seen in your python plugin, i am currently trying to fix it

src/libs/pluginprocess/pluginprocess.c
e1528532: that forces me to convert the number and that is cumbersome in c :)

src/libs/pluginprocess/pluginprocess.c
e1528532: commandKeySet WILL contain 6 keys in the end. But i agree that i could send over the payload's size for an optimization.

src/libs/pluginprocess/pluginprocess.c
e1528532: it is a number

e1528532:
> It is also interesting if the plugins are able to change the KeySet, including their contract.
Like ordinary plugins, they can do anything with the keyset. i don't see the reason to build any special precautions into pluginprocess for that.

> You might want to add some tests in python or haskell plugins, too.
There are already a lot of haskell plugin tests for the typechecker especially, none of them would work in case pluginprocess doesn't do what its supposed to do as the haskell runtime really cannot be restarted in the same process.

markus2330:
> Like ordinary plugins, they can do anything with the keyset. i don't see the reason to build any special precautions into pluginprocess for that.

My point is that before this PR the root key of the contract was removed. There should be a test checking for this.

> There are already a lot of haskell plugin tests for the typechecker especially, none of them would work in case pluginprocess doesn't do what its supposed to do as the haskell runtime really cannot be restarted in the same process.

Yes, but it seems like there were no tests for keysets which contained the parentkey.

src/libs/pluginprocess/pluginprocess.c
markus2330: Perfect, then only the docu needs to be udpated :smile:

src/libs/pluginprocess/pluginprocess.c
markus2330: Ahh, my fault. Then 6 is perfect.

src/libs/pluginprocess/pluginprocess.c
markus2330: We soon have excellent macros for number conversions. Unfortunately it is not yet merged... #1850

src/libs/pluginprocess/pluginprocess.c
markus2330: Maybe rename cleanedUp to processFinished?

But isn't this wrong if the "close" call did not clean up?

e1528532:
> My point is that before this PR the root key of the contract was removed. There should be a test checking for this.

i can add a test for your peace of mind, but with the new communication protocol the payload / keyset passed to plugins is not being modified, thus it is solely up to the plugin itself what it does with it. Pluginprocess doesn't touch it anymore, it only uses its own commandKeySet.

markus2330:
Yes, I agree that the keyset should now be fully separated. But this does not mean that every future implementation will do it that way. Also a test case for the parentKey would be interesting. In particular also if metadata is passed correctly both ways.

Please make also sure that the PR passes the tests.

e1528532:
> Also a test case for the parentKey would be interesting. In particular also if metadata is passed correctly both ways.

added a test for that case as well (see `test_keysetContainingParentKey()`). Indeed this is an interesting case! Turns out if the parent key is added to the keyset it gets locked, thus keyCopy later on would fail. I've now forbidden to modify parent key names in plugins as imo this shouldn't be done and copy the data manually.

Metadata serialization from parent to child is tested in that case as well. Metadata serialization in the other direction (child -> parent) is already covered in the communication test but checked again to ensure the keyset containment doesn't influence it somehow.
 

> Please make also sure that the PR passes the build server tests.

of course

src/libs/pluginprocess/pluginprocess.c
```diff
@@ -291,9 +293,28 @@ int elektraPluginProcessSend (const ElektraPluginProcess * pp, pluginprocess_t c
 	}
 	else
 	{
+		int parentKeyExistsInOriginalKeySet = keySet != NULL ? ksLookup (originalKeySet, key, KDB_O_POP) != NULL : 0;
 		// Copy everything back into the actual keysets
-		keyCopy (key, parentDeserializedKey);
-		if (keySet != NULL) ksCopy (originalKeySet, keySet);
+		// Unfortunately we can't use keyCopy here as ksAppendKey locks it so it will fail
+		// This is the case if the parent key is also contained in the originalKeySet / has been appended
+		// As an invariant we assume plugins don't change the parent key's name during a plugin call
```
markus2330: Yes, that is a perfectly valid assumption but we should also document this in src/plugins/doc.c

markus2330:
Thank you for the work! Seems like there are still some memleaks.

markus2330:
Btw. the more usual case is not yet tested: Plugins often append the parentKey to be part of the keyset. (So the other way round.)

markus2330:
Thank you for adding the test. Seems like there are still memleaks?

e1528532:
Unfortunately yes, they are quite hard to track down. Asan with Clang still causes issues, asan with gcc is fixed now.

Furthermore it looks like the pipe creation fails on this build job: https://travis-ci.org/ElektraInitiative/libelektra/jobs/411929856 ? Is it possible that i can't write to /tmp/ on that system?

markus2330:
@ingwinlu Do you know why pipe creation fails?

markus2330:
@e1528532 any update here? Would be great to get this merged soon.

About the pipes: Better error messages would help a lot.

e1528532:
Now i see the details. It fails to load dump due to some ASAN error on clang. Apparently that happened before i PR #1825 . Will check how it was solved there

```
buffer is: warnings/#00
number: 1
description: could not load module, dlopen failed
ingroup: modules
module: dl
file: /home/travis/build/ElektraInitiative/libelektra/src/libs/loader/dl.c
line: 89
reason: of module: libelektra-dump.so, because: /home/travis/build/ElektraInitiative/build/lib/libelektra-dump.so: undefined symbol: __ubsan_vptr_type_cache
```

e1528532:
So PR #1825 solved it by preloading the ubsan library. I will try this approach but i wonder why this is no longer present in the current travis file? @sanssecours you often seem to maintain the travis file, do you remember anything about that?

e1528532:
Also do you know how can i add this preloading to our build server for this one test using asan with clang on debian where it fails @ingwinlu ?

ingwinlu:
add it in cmake to make it work everywhere, we use this (for example) to make tests load libs correctly when elektra is not installed:
https://github.com/ElektraInitiative/libelektra/search?l=CMake&q=LD_LIBRARY_PATH

e1528532:
Ok trying to fix the asan issue using cmake.

Furthermore output_warnings was indeed able to cause buffer overflows, fixed those (https://stackoverflow.com/questions/6903997/how-can-i-use-strncat-without-buffer-overflow-concerns#6904044 for reference)

sanssecours:
>  I will try this approach but i wonder why this is no longer present in the current travis file?

As far as I can remember I [deleted the code setting the environment variable](https://github.com/ElektraInitiative/libelektra/commit/43c1a54), since we now remove the affected test if ASAN is enabled.

markus2330:
So we can simply revert 43c1a54921c81a00f19acf51694780b9e35a9706?

e1528532:
I think handling this via cmake is better as it also happens for one of our buildserver tests and not only on travis. In general when using linux + clang as it seems.

e1528532:
@markus Finally i think i have resolved the remaining issues. Any more objections?

markus2330:
No objections! Thank you so much for resolving this so nicely.

PR #2153, opened 2018-08-04
Template for C++ Based Plugins

sanssecours:
# Checklist

- [x] This PR contains additional unit tests.
- [x] I updated documentation affected by this pull request.
- [x] The code of `cpptemplate` contains Doxygen comments.
- [x] The metadata of `cpptemplate` should be more or less correct.


src/plugins/cpptemplate/README.md
```diff
@@ -0,0 +1,17 @@
...
+- infos/placements = getstorage setstorage
+- infos/status = maintained unittest nodep nodoc concept
+- infos/metadata =
+- infos/description = A template for C++ based plugins
+
+# CPP Template
+
+## Introduction
+
+Please use the script [copy-template](../../../scripts/copy-template) to create a new C++ plugin based on this template.
```
markus2330: With the argument `-p`.

src/plugins/cpptemplate/cpptemplate.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+} // end namespace
+
+extern "C" {
+
+typedef Delegator<CppTemplateDelegate> delegator;
+
+/** @see elektraDocOpen */
+int elektraCppTemplateOpen (Plugin * handle, Key * key)
```
markus2330: I think nearly everything of this file below this point could be moved into a macro. No need to duplicate it, the functions are not supposed to be changed by users, aren't they?

And if this file gets much shorter: Is it necessary to split this file with the implementation of the delegator?

src/plugins/cpptemplate/cpptemplate.cpp
```diff
@@ -0,0 +1,111 @@
...
+{
+	// After the call to `delegator::open` you can retrieve a pointer to the delegate via `coderDelegator::get (handle)`
+	return delegator::open (handle, key);
```
markus2330: What happens if delegator::open throws an exception? What about catching them here and set an error for uncaught exceptions?

src/plugins/cpptemplate/cpptemplate.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/** @see elektraDocClose */
+int elektraCppTemplateClose (Plugin * handle, Key * key)
+{
+	return delegator::close (handle, key);
```
markus2330: pass CppKey here (and also in Open)?

src/plugins/cpptemplate/cpptemplate.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+	parent.release ();
+	keys.release ();
+	return updated ? ELEKTRA_PLUGIN_STATUS_SUCCESS : ELEKTRA_PLUGIN_STATUS_NO_UPDATE;
+}
+
+/** @see elektraDocSet */
+int elektraCppTemplateSet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned ELEKTRA_UNUSED, Key * parentKey ELEKTRA_UNUSED)
+{
+	return ELEKTRA_PLUGIN_STATUS_NO_UPDATE;
```
markus2330: No delegator?

src/plugins/cpptemplate/cpptemplate_delegate.cpp
```diff
@@ -0,0 +1,24 @@
...
+using CppKeySet = kdb::KeySet;
+
+/**
+ * @brief This constructor creates a new delegate object used by the `cpptemplate` plugin
+ *
+ * @param config This key set contains configuration values provided by the `cpptemplate` plugin
+ */
+CppTemplateDelegate::CppTemplateDelegate (CppKeySet config ELEKTRA_UNUSED)
+{
+}
```
markus2330: open, close, ... not implemented?

src/plugins/cpptemplate/cpptemplate_delegate.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <kdbplugin.hpp>
+
+namespace elektra
+{
+
+class CppTemplateDelegate
+{
+	using CppKeySet = kdb::KeySet;
+
+public:
```
markus2330: open, close, ... not declared?

src/plugins/cpptemplate/README.md
```diff
@@ -0,0 +1,17 @@
+- infos = Information about the cpptemplate plugin is in keys below
+- infos/author = Author Name <elektra@libelektra.org>
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/recommends =
+- infos/placements = getstorage setstorage
+- infos/status = maintained unittest nodep nodoc concept
```
markus2330: These defaults are not good and inconsistent with the other template.

src/plugins/cpptemplate/cpptemplate_delegate.hpp
```diff
@@ -0,0 +1,29 @@
...
+ * @file
+ *
+ * @brief Delegate definitions for the `cpptemplate` plugin
+ *
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ *
+ */
+
+#ifndef ELEKTRA_CPP_TEMPLATE_DELEGATE_HPP
+#define ELEKTRA_CPP_TEMPLATE_DELEGATE_HPP
```
markus2330: Is CPP_TEMPLATE renamed? Otherwise it is dangerous if different files get the same include guard.

src/plugins/cpptemplate/cpptemplate.cpp
sanssecours: > I think nearly everything of this file below this point could be moved into a macro. No need to duplicate it, the functions are not supposed to be changed by users, aren't they?

While I agree that the functions could be moved into a macro, I think it also makes sense to show the implementation to the plugin developer. If we use a macro here, then a plugin developer might just assume their is some weird magic involved here. As it currently stands a new plugin developer can look at the template code and he will probably know what is going on.

> And if this file gets much shorter: Is it necessary to split this file with the implementation of the delegator?

Not really. However, this template only includes the basic skeleton of a plugin. The code of an actual plugin might be much longer. For this situation it makes sense to keep the code separated in several files.

Anyway, we can always modify the plugin code after we call `copy-template`. It should be easy for a developer to move the whole code into a single file if she or he prefers it that way.

src/plugins/cpptemplate/cpptemplate.cpp
sanssecours: > What happens if delegator::open throws an exception? What about catching them here and set an error for uncaught exceptions?

The code now catches exceptions and adds warning `11`:

```
number:11
description:open of plugin returned unsuccessfully (reason contains plugin, see other warnings for details)
severity:warning
ingroup:kdb
macro:OPEN
```

in that case. I am not sure if warning `11` is really appropriate though. Should I use another warning/error instead?

src/plugins/cpptemplate/cpptemplate.cpp
markus2330: >  I think it also makes sense to show the implementation to the plugin developer. If we use a macro here, then a plugin developer might just assume their is some weird magic involved here.

I cannot follow the argumentation. So we always should prefer to copy code instead of invoking existing code?

> As it currently stands a new plugin developer can look at the template code and he will probably know what is going on.

Why would macros change that?

> Anyway, we can always modify the plugin code after we call copy-template.

Yes, that is why *only* parts that should be changed should be copied around via copy-template.

src/plugins/cpptemplate/cpptemplate.cpp
sanssecours: As far as I can tell that would only work if we change the definition of `open` in `kdbplugin.hpp`? Should I really do that or do? Wouldn’t that break the current C++ plugin API?

src/plugins/cpptemplate/cpptemplate.cpp
sanssecours: Since this is only a template I think this makes sense. The plugin writer can add a call to whatever function of the delegate she or he likes later.

src/plugins/cpptemplate/cpptemplate_delegate.cpp
sanssecours: The `open` and `close` functions are implemented in the delegator ([`kdbplugin.hpp`](https://github.com/ElektraInitiative/libelektra/blob/master/src/bindings/cpp/include/kdbplugin.hpp)).

src/plugins/cpptemplate/cpptemplate_delegate.hpp
sanssecours: We do not need them here. We can add any function we want to the delegate. I feel a little silly telling  you that though, after all you wrote [`kdbplugin.hpp`](https://github.com/ElektraInitiative/libelektra/blob/master/src/bindings/cpp/include/kdbplugin.hpp). I feel like, I might not understand the whole delegator/delegate approach after all…

src/plugins/cpptemplate/cpptemplate_delegate.hpp
sanssecours: Yes.

src/plugins/cpptemplate/README.md
sanssecours: That is not a problem in my opinion since `copy-template` uses `README.md` of the `template` plugin anyway.

src/plugins/cpptemplate/cpptemplate.cpp
sanssecours: > I cannot follow the argumentation. So we always should prefer to copy code instead of invoking existing code?

Nope. in my opinion macros can make code harder to read and to debug (bad error messages). I also think that plugin developers might want to change the definition of `elektraCppTemplateOpen` or `elektraCppTemplateClose` later. 

Anyway, if you want you can still change the template code after it is merged. You can also close this PR if you do not like the code.

> Why would macros change that?

While it is certainly possible to look up macros, I usually do not do that 😊. I also prefer normal functions over macros since macros cause the compiler to print very bad error messages.

> Yes, that is why *only* parts that should be changed should be copied around via copy-template.

While I agree, I also think that a plugin developer still want to adopt `elektraCppTemplateClose` and `elektraCppTemplateClose` her or his liking. Regardless, **please feel free to close this PR** if you do not like the current code.

markus2330: > in my opinion macros can make code harder to read and to debug (bad error messages).

Macros sometimes caused really bad error messages in earlier versions of GCC. But as these macros are only trivial wrappers, they are unlikely to be important for debugging.

> I also think that plugin developers might want to change the definition of elektraCppTemplateOpen or elektraCppTemplateClose later.

I was hoping that it will be enough for C++ plugins to implement the delegator and that the low-level C functions never need to be touched.

> Regardless, please feel free to close this PR if you do not like the current code.

I like the direction of the PR. I only try to push the code a bit further so that plugins that stem from it get easier to maintain. But also in its current form it will help developers.

src/plugins/cpptemplate/cpptemplate.cpp
markus2330: It should be an error because an error status will be returned.

src/plugins/cpptemplate/cpptemplate.cpp
markus2330: You are right, the change does not make much sense if this is hidden in a macro anyway. For user code, however, it would be nice if devs are only confronted with CppKey(Set). In particular code as in src/plugins/type/type.cpp line 95 or src/plugins/ccode/ccode.cpp lines 71-75, 85-88 should be avoided.

Such a change breaks the API but it does not break the ABI because it is header-only. So people need to adapt their code but already compiled code still works.

src/plugins/cpptemplate/cpptemplate.cpp
markus2330: For plugin templates it makes sense that all functions are present and users remove the functions they do not need.

src/plugins/cpptemplate/cpptemplate_delegate.cpp
markus2330: Maybe some comment like "implement open functionality here" would help?

Btw. do you really like the style that everything starts with Cpp? Would you call a ruby function Ruby...?

src/plugins/cpptemplate/cpptemplate_delegate.hpp
markus2330: Sorry for the confusion. I simply forgot what I wrote (and did not see that you include kdbplugin.hpp) and thought that the cpptemplate_delegate.hpp fully implements the delegator.

It makes a lot of sense how it is. You should only mention that the constructor/destructor should contain the open/close functionality.

src/plugins/cpptemplate/cpptemplate_delegate.cpp
sanssecours: > Btw. do you really like the style that everything starts with Cpp?

Not, really. However since `Key`/`KeySet` is usually already reserved for `ckdb::Key`/ `ckdb::KeySet` I tend to just add the `Cpp` prefix. If you have a better naming option in mind, I am open for suggestions. For now I changed the name to the usual `KeySet` in the delegate.

> Would you call a ruby function Ruby...?

Not really, but usually I do not encounter naming conflicts in languages such as Python or Ruby.

src/plugins/cpptemplate/cpptemplate.cpp
sanssecours: Oh, I see. In that case I **might** change the interface in a separate pull request.

src/plugins/cpptemplate/cpptemplate.cpp
sanssecours: In the latest commit I added an example method to the delegate. Since the design of the delegate is not really dependent on the plugin interface, I do not think it makes sense to add a method for each part of the plugin interface. If you think differently, then you are of course welcome to extend the template accordingly.

markus2330: Thank you!

To avoid code duplication from the code generator is more important anyway. Ideally we would also get rid of the manual export of symbols. With supermacros it should be possible to generate the code as needed (for both contract and symbol exporting). It could look like:

```
#define PLUGIN_HAS_OPEN
#define PLUGIN_HAS_GET
#define PLUGIN_HAS_SET

#define PLUGIN_MACRO EXPORT_CONTRACT
#include <kdbplugingen.hpp>

#define PLUGIN_MACRO EXPORT_FUNCTIONS
#include <kdbplugingen.hpp>

#define PLUGIN_MACRO EXPORT_SYMBOLS
#include <kdbplugingen.hpp>
```

By default it generates everything, so the one line `#include <kdbplugingen.hpp>` would be enough.

For more info about supermacros, read: http://s11n.net/papers/supermacros_cpp.html

Similar techniques are not rare, e.g., search for BOOST_SERIALIZATION_NVP

sanssecours: Thank you for the information, but I think I will leave this task up to you (or anybody else who likes to work on this stuff).

sanssecours:
> Can you remove the lower part of doc/todo/PLUGINS?

I did not remove this part of the text, but rather updated it with new information. As far as I can tell the PR does not solve any of the goals mentioned in `doc/todo/PLUGINS`.

markus2330:
Thank you for the improvement!

PR #2158, opened 2018-08-07
Detect nftw

ingwinlu:
Resolve an issue where NFTW was not detected by CMake.
Also adds the ability to ignore aborts when sending status mails.

Closes #2157.

## Basics

Do not describe the purpose here but:

- [ ] Short descriptions should be in the release notes (added as entry in
      doc/news/_preparation_next_release.md which contains `*(my name)*`)
      **Please always add something to the the release notes.**
- [ ] Longer descriptions should be in documentation or in design decisions.
- [ ] Describe details of how you changed the code in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, should be in the commit messages.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests
- [ ] I ran all tests locally and everything went fine
- [ ] affected documentation is fixed
- [ ] I added code comments, logging, and assertions (see doc/CODING.md)
- [ ] meta data is updated (e.g. README.md of plugins)

## Review

Remove the line below and add the "work in progress" label if you do
not want the PR to be reviewed:

@markus2330 please review my pull request

PR #2159, opened 2018-08-07
 CFramework: Improve Detection of `ftw.h`

sanssecours:
# Description

Before this PR the command `kdb_runall` in the Travis build job `🍏 GCC` prints the following text:

```
CYPTO        TESTS
==================


crypto_gcrypt Results: 65 Tests done — 0 errors.
/Users/travis/build/sanssecours/elektra/tests/cframework/tests.c:542: error in clean_temp_home: Could not delete TMPHOME manually
--- running testmod_crypto_openssl ---


CYPTO        TESTS
==================


crypto_openssl Results: 65 Tests done — 0 errors.
/Users/travis/build/sanssecours/elektra/tests/cframework/tests.c:542: error in clean_temp_home: Could not delete TMPHOME manually
```

. After the PR the error messages are gone:

```
CYPTO        TESTS
==================


crypto_gcrypt Results: 65 Tests done — 0 errors.
--- running testmod_crypto_openssl ---


CYPTO        TESTS
==================


crypto_openssl Results: 65 Tests done — 0 errors.
```
.

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -461,6 +461,7 @@ Thanks to Daniel Bugl.
 - The build system now automatically detects Homebrew’s OpenSSL version on macOS. *(René Schwaiger)*
 - We improved the automatic detection of Libgcrypt and OpenSSL. *(René Schwaiger)*
 - Resolved an issue where cmake did not properly set test feature macros to detect and use libc functionality. *(Lukas Winkler)*
+- Improve the detection of `ftw.h`, if the current build use the compiler switch `-Werror`. *(René Schwaiger)*
 
 [Google Test]: https://github.com/google/googletest
 
--- a/tests/cframework/CMakeLists.txt
+++ b/tests/cframework/CMakeLists.txt
@@ -8,7 +8,18 @@ add_definitions (${REQUIRED_FEATURE_MACROS})
 
 cmake_push_check_state (RESET)
 set (CMAKE_REQUIRED_DEFINITIONS ${REQUIRED_FEATURE_MACROS})
+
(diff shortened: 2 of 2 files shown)


markus2330:
Thank you for your relentless pursuit to improve Elektra!

ingwinlu:
As I mentioned in my previous addressing this issue this change is not a good idea. Can we please revert

markus2330:
@sanssecours finished this PR just recently, so I assumed this addressed some additional problem. Can you please comment?

ingwinlu:
It is exactly the same as #2158. Just that this breaks the availability check.

There is no improved detection here. The check simply does not fail anymore even if it is not exported.

sanssecours:
> There is no improved detection here.

As far as I can tell from the output of [build job `🍏 GCC` before](https://travis-ci.org/ElektraInitiative/libelektra/jobs/413232105#L9156-L9170) and [after](https://travis-ci.org/ElektraInitiative/libelektra/jobs/413458677#L9156-L9168) this PR was merged the detection was improved. macOS does provide the function `nftw`, I do not see how choosing GCC instead of Clang should change that, unless GCC does not use the standard macOS C library.

> The check simply does not fail anymore even if it is not exported.

I do not think that is the case. For example, on my machine, if I start with a clean build directory and invoke `cmake` it prints the following messages

```
-- Looking for nftw
-- Looking for nftw - found
```

. If I replaces the line 

```cmake
check_symbol_exists (nftw "ftw.h" HAVE_NFTW)
```

with

```
check_symbol_exists (toBeOrNotToBe "ftw.h" HAVE_NFTW)
```

, clean the project directory and run `cmake`, then the command prints the following text

```
-- Looking for toBeOrNotToBe
-- Looking for toBeOrNotToBe - not found
```

.

PR #2166, opened 2018-08-08
kdb: fixed formatting

Piankero:
Quick fix for the formatting check.


markus2330:
This formatting also does not look right. You have to use the scripts/reformat-source for this purpose.

Issue #1010, opened 2016-10-07
Separators


markus2330:
key-value pairs have a huge  diversity in how keys and values (and key-value pairs) are separated and Elektra obviously needs to support that. Currently, however, even *within* Elektra different ways to separate key-value pairs creep in:

1. plugin config, e.g. for mounting: `key1=value1,key2=value2`
2. embedding key-value pairs within a value of a key: usually space seperated, except
3. enum config uses `'` for values, ',' as separator between key-values: `'abc', 'def'`
4. namespaces: currently no seperator (use the existing `/`, except for obsolete `user:` feature)
5. comment (used new-lines for seperation), already obsolete by using Elektra-arrays
6. infos/status infos/plugins infos/provides: keys are space separated
7. see discussion in #994 we need a separation of tags in infos/provides

Are there any other places missing in the list above?

What is your favourite way?

markus2330:
I think we should avoid key-value pairs within values: an Elektra array would be better here. If we only have keys (e.g. in infos/status) the array, however, seems to be really heavyweight,  space seperation looks nicer there.

markus2330:
Thanks for the feedback! Outcome is:

- #994 stays as is (with `/`)
- #1037 Separate namespace with key path
- #1035 kdb mount: avoid positional arguments

PR #2169, opened 2018-08-08
plugins: Plugin ideas added

Piankero:
I want to propose 3 new plugin ideas. For the recursion plugin I will need some more advise though from more experienced programmers.


src/plugins/port/README.md
```diff
@@ -0,0 +1,27 @@
...
+common for client applications connecting to the server.
+
+## Alternative 1
+
+`check/port/open true` could be used in the metadata to declare that it has to be open. A simple `check/port` will only
+validate the range.
+
+## Alternative 2
+
+Incorporate the port plugin in the network plugin (with the same behaviour). Those plugins basically do different
```
markus2330: This makes a lot of sense. More details please (e.g. extend README.md of network plugin).

Btw. actually ping a server would also be interesting but it is a nice-to-have.

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
+
+
+## Introduction
+
+I took [Relax NG](http://relaxng.org/spec-20011203.html#define-ref)'s `ref` and `define` part and tried to "copy" 
+their approach for elektra's specification.
```
markus2330: I like this idea! Iirc Relax NG is also mathematically sound, could you check if you find a paper with a proof?

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+Until now I haven't really gotten a sensible idea for that though. Maybe we could talk on the next meeting about this.
```
markus2330: I am afraid this is mainly up to you to find out how this works.

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Also I have no idea what your `struct` plugin actually does. There is no example, just some code snippet from
```
markus2330: Thank you for the feedback. The struct plugin was written at a time when no specification existed, so everything was solved via contracts.

The fstab plugin might help you to understand. Important to know is that plugins can read each other's configs/needs.

The idea is that you either extend the struct plugin or you write a new plugin that replaces the struct plugin.

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+I took [Relax NG](http://relaxng.org/spec-20011203.html#define-ref)'s `ref` and `define` part and tried to "copy" 
+their approach for elektra's specification.
+Until now I haven't really gotten a sensible idea for that though. Maybe we could talk on the next meeting about this.
+
+Also I have no idea what your `struct` plugin actually does. There is no example, just some code snippet from 
+the contract which doesn't help understanding the plugin.
+
+## Problems
+1) We want to check if all recursions are non-cyclical
```
markus2330: Details please. Which recursion do you mean? Cycles of links (override/fallback)?

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+2) We want to check if a recursive config actually leads to another correct config and not into NULL for example
+3) Forcing users to define non-recursive settings first on which they can build up 
+recursive settings is inconvenient and could lead to orphan settings.
+4) We can only check right after calling `kdb set` and not a whole config. 
+This means that user can have "unfinished" configurations while setting new values.
```
markus2330: I do not understand this limitation. Within "kdb set" you can check the whole configuration and reject everything that is unfinished.

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Possible Solution
+
+Users have to define recursive setting specifications in the following way:
+
+```
+kdb mount config.dump /recursive dump recursion
+kdb setmeta user/recursive check/recursion/vertex "menu, submenu"
+kdb setmeta user/recursive check/recursion/leaf "param, menupoint"
```
markus2330: Please use Elektra arrays, for example:

```
kdb setmeta user/recursive check/recursion/vertex/#0 menu
kdb setmeta user/recursive check/recursion/vertex/#1 submenu
kdb setmeta user/recursive check/recursion/leaf/#0 param
kdb setmeta user/recursive check/recursion/leaf/#1 menupoint
```

See #1010

A `kdb set-meta-array` would be nice to have.

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+`check/recursion/vertex` basically means that keys containing `menu` or `submenu` after `/recursive`
+will have further configurations below them. So users would build up structures like this:
+`/recursive/<vertex>/#[0-9]+/<vertex>/#[0-9]+/.....`
```
markus2330: Can you give a concrete example? Why is `#[0-9]+` allowed?

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+`check/recursion/leaf` means that there is no subvertex allowed anymore. So this is not allowed:
```
markus2330: What does "param, menupoint" mean?

src/plugins/recursion/README.md
```diff
@@ -0,0 +1,64 @@
...
+kdb set /recursive/menu/#0/menupoint1 menu2
+kdb set /recursive/menu/#1/menupoint1/#0 submenu
+kdb set /recursive/menu/#1/menupoint2/#1 menupoint2
+```
+
+Vertices without leafs are allowed (eg. submenu in this example).
+
+A command will fail if users want to set settings for non-existent vertices, eg. `kdb set /recursive/menu/#5/....`
+
+The validation solely happens on the structure, not on the values of the keys.
```
markus2330: Yes, it makes sense to separate this.

src/plugins/path/README.md
```diff
@@ -24,9 +25,45 @@ is allowed to occur for both device and mountpoint. When checking for
 relative files, it is not enough to look at the first character if it is
 a `/`, because remote file systems and some special names are valid, too.
 
+If `check/permission = [permission], [user]` is also present it will check for the correct permissions
+of the file/directory. `check/permission = rw, tomcat` for example will check if the user `tomcat` has read and
```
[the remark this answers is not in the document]
Piankero: What about `check/permission = rwx `and
`check/permission/user = tomcat` in order to split up the permission and ownership?

When the user types in "current" in the user part, I would simply replace it with the user who caused the "kdb set". Does this make sense for you?

src/plugins/path/README.md
```diff
@@ -24,9 +25,45 @@ is allowed to occur for both device and mountpoint. When checking for
...
 a `/`, because remote file systems and some special names are valid, too.
 
+If `check/permission = [permission], [user]` is also present it will check for the correct permissions
+of the file/directory. `check/permission = rw, tomcat` for example will check if the user `tomcat` has read and 
+write access to the path which was set in `check/path`.
+ 
+ Permissions available:
+ - `r`: **R**ead
+ - `w`: **W**rite
+ - `x`: e**X**ecute
```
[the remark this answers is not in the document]
Piankero: Alright, I will add it in the next commit!

src/plugins/path/README.md
```diff
@@ -24,9 +25,45 @@ is allowed to occur for both device and mountpoint. When checking for
...
+
```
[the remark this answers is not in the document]
Piankero: I will leave this open for now and see in my thesis if it is needed that often. But the idea is good!

src/plugins/path/README.md
```diff
@@ -24,9 +25,45 @@ is allowed to occur for both device and mountpoint. When checking for
...
+#This checks if the file actually exists
+kdb setmeta user/tests/path/value check/path
+
+#This checks if the user has read and write permissions for the application-file.log file
+kdb setmeta user/tests/path/value check/permission "rw, tomcat"
+
+#Generate a file which is only accessable for root
+touch /var/log/application-file-restricted.log
+sudo chmod 700 /var/log/application-file-restricted.log
+sudo chown root:root /var/log/application-file-restricted.log
```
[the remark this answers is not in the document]
Piankero: This doesnt trigger an error. It just generates a file with restrictive permissions. Or am I misunderstanding something?

src/plugins/port/README.md
```diff
@@ -0,0 +1,27 @@
...
+## Purpose
+
+One source of misconfiguration can come from invalid port numbers or from taking used ports such as 22 for ssh.
+This plugin checks if the given port is open as well as the given port lies between 1-65535
+
+## Usage
+
+If `check/port` is specified on a given key, the plugin will validate if the port is a 
+correct number between 1 and 65535.
+Furthermore it will check if the port is open to use.
```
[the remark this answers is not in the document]
Piankero: Good Idea, I will add a section with possible enhancements

src/plugins/port/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+If `check/simpleport` is specified, only the correct port range will be checked. The port may be in use which is
```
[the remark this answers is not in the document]
Piankero: Indeed, but I didn't come up with a better idea after pondering about this to be honest :P . I will change the behavior as follows:

`check/port/open` Range + available
`check/port/range` Range

I think this is better

src/plugins/path/README.md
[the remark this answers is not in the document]
markus2330: > What about check/permission = rwxand
check/permission/user = tomcat in order to split up the permission and ownership?

The ownership has actually little to do with permission.  So maybe `check/owner/user` and `check/owner/group`?

> When the user types in "current" in the user part, I would simply replace it with the user who caused the "kdb set". Does this make sense for you?

This would create a problem if there is a user called "current". What about an empty string? I think this is not allowed as actual user name.

src/plugins/path/README.md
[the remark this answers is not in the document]
markus2330: If you agree and have/will fix it, a "thumbs up" :+1: or similar is enough.

src/plugins/path/README.md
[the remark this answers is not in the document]
markus2330: Yes, making everything complete is not a good idea anyway. It easily happens that stuff is added nobody ever needs. I just mentioned it so that you are aware that such a check might be relevant for some software.

src/plugins/path/README.md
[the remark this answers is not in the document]
Piankero: My question here is, if the owner is actually that much needed.
I just want to check if a certain user can read/write/etc. on a certain file.

If for example user "java" can write to /var/log/abc.log. Why is the owner of the file of any relevance?

src/plugins/port/README.md
```diff
@@ -0,0 +1,27 @@
+## Introduction
+
+This plugin is a check plugin that checks if the port on localhost is open and a valid number.
+
+## Purpose
+
+One source of misconfiguration can come from invalid port numbers or from taking used ports such as 22 for ssh.
```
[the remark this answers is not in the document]
Piankero: I just wanted to provide any example. I will change the sentence so it is clear that any port can be used.

src/plugins/recursion/README.md
Piankero: That's correct, but this would force the user to have a specified ordering when creating the configuration. Either "bottom-up" (you have to create leaf settings before recursive ones) or "top-down" (you have to create top level settings first before adding lower level ones).

Klemens said that some plugins which take a config do not have a specific ordering once you load them.

src/plugins/recursion/README.md
Piankero: I will change the example in the next commit

src/plugins/recursion/README.md
Piankero: I will change the example in the next commit

src/plugins/recursion/README.md
markus2330: Some storage plugins do not remember the order of entries *within the configuration files*. But the order of the configuration structure should have nothing to do with that.

src/plugins/port/README.md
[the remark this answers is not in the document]
markus2330: What about:

`check/port` for the range check (and/or transformation from /etc/services)
`check/port/listen` to check if the port can be listened to (usually for servers)
`check/port/ping` to check if the port can be pinged/reached  (usually for clients)

src/plugins/path/README.md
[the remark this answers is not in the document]
markus2330: I maybe misunderstood this because of the earlier misunderstanding.

src/plugins/path/README.md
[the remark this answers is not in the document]
Piankero: The sticky bit does not change the behavior of the program at all so I think it is better to ignore it.

I am thinking the same about groups. In the end its a a user who executes or uses the file in the setting. So probably its the best to keep it that simple and only provide rwx.

I am unclear now if you really want to split up permission and user as for me it is more intuitive to have them both in one metadata, eg. check/permission "rwx, java" (or maybe without the colon). If check/permission "rwx, " is given, the current user is taken?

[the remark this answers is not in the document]
markus2330: In the implementation it matters: if the file does not exist, it is the sticky bit of the parent directory which may decide if the user can create the file.

Btw. did you check if there an "access (2)" which allows us to say for which user it should be checked? Then the implementation would be trivial.

Maybe there should also be a flag if the file already needs to be present. For log files there are many differences in this area. Actually, if the software does not allow the creation of files, we could simply create it.

> I am unclear now if you really want to split up permission

Yes, the split is a must. Otherwise we get problems if the user names contains ",", spaces or similar.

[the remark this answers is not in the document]
markus2330: >  Unfortunately I do not understand what you mean by this confused

I mean to use the syscall access (2) http://man7.org/linux/man-pages/man2/access.2.html

> logs there are rollovers and this could simply lead to a delayed configuration error which isn't wanted.

Yes, this is certainly not wanted. I think we need to gather more data here for what applications do here.

One more advanced idea is that Elektra already opens a file and then passes /proc/self/fd/<nr> instead of the real file name. Then it is impossible that the file disappears (but it needs additionally resources, namely an open file descriptor). 

> What about

Simply propose it in the PR. We can take a look tomorrow.

src/plugins/port/README.md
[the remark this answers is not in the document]
Piankero: `check/port/ping` sounds promising but I also think that we first should gather data about the usage. Imo it is often the case the client and server configurations are set before the server is actually started (for example I do it for LCDproc like this). This would prohibit the user from setting a potentially correct setting.

I will add it to a potential future work section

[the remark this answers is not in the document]
markus2330: It is also possible and useful to emit warnings.

> I will add it to a potential future work section

If you think it does not work well, it would be rather in the "Discarded Ideas" section.

src/error/specification
```diff
@@ -1209,3 +1209,9 @@ severity:error
 ingroup:plugin
 module:typechecker
 macro:PLUGIN_TYPECHECKER_SET
+
+number:194
+description:The given path does not meet the correct permissions
```
markus2330: We should try to reuse/reduce the errors and make a nice classification system. Hopefully you can contribute to improve the error system.

src/plugins/network/README.md
```diff
@@ -37,3 +39,15 @@ it to implement this plugin.
...
+If `check/port` is specified on a given key, the plugin will validate if the port is a 
+correct number between 1 and 65535.
+
+If `check/port/listen` is specified, the plugin will check if the application can be started
+and listen on the given port.
+
+## Future Work
+
+`check/port/ping` to check if the port can be pinged/reached (usually for clients).
+This potentially could prohibit users from setting correct settings as the server application
```
markus2330: Not if it is a warning.

The problem is rather that a timeout is needed.

src/plugins/path/README.md
```diff
@@ -26,7 +27,7 @@ relative files, it is not enough to look at the first character if it is
 a `/`, because remote file systems and some special names are valid, too.
 
 If `check/permission = [permission], [user]` is also present it will check for the correct permissions
-of the file/directory. `check/permission = rw, tomcat` for example will check if the user `tomcat` has read and 
+of the file/directory. `check/permission = rw` and `check/permission/user = tomcat` for example will check if the user `tomcat` has read and
```
markus2330: There is still no explanation of check/permission, check/permission/user. It is good to have an example but there should also be an explanation without example.

src/plugins/path/README.md
```diff
@@ -48,10 +49,10 @@ sudo kdb mount some_file.ecf /tests/path some_file dump
 kdb set /tests/path/value /var/log/application-file.log
 
 #This checks if the file actually exists
-kdb setmeta user/tests/path/value check/path
+sudo kdb setmeta user/tests/path/value check/path
```
markus2330: "user" needs to be removed, otherwise the change does not make sense.

src/plugins/path/README.md
```diff
@@ -60,10 +61,13 @@ sudo chown root:root /var/log/application-file-restricted.log
 
 #This should trigger the error
 kdb set /tests/path/value /var/log/application-file-restricted.log
-# RET:X
-# ERROR:XXX
+# ERROR:194
+# Reason: User tomcat has no permission to read the given file.
```
markus2330: Reason should state what was expected and what is actual.

src/plugins/recursion/README.md
```diff
@@ -23,25 +23,37 @@ Users have to define recursive setting specifications in the following way:
 
 ```
 kdb mount config.dump /recursive dump recursion
-kdb setmeta user/recursive check/recursion/vertex "menu, submenu"
-kdb setmeta user/recursive check/recursion/leaf "param, menupoint"
+kdb setmeta user/recursive check/recursion/vertex "A, B"
```
markus2330: Still without array syntax.

src/plugins/recursion/README.md
```diff
@@ -23,25 +23,37 @@ Users have to define recursive setting specifications in the following way:
...
+kdb setmeta user/recursive check/recursion/leaf "z, y"
 ```
 
-`check/recursion/vertex` basically means that keys containing `menu` or `submenu` after `/recursive`
-will have further configurations below them. So users would build up structures like this:
+`check/recursion/vertex` basically means that keys containing `A` or `B` after `/recursive`
```
markus2330: The word "basically" is useless. Say it like it is. And again: with+without example.

doc/METADATA.ini
```diff
@@ -679,6 +679,12 @@ type= enum
 	RFC2822
 description= Support of some standards for a date specification.
 
+[check/permission]
+status= idea
+usedby/plugin= path
+type= string
+description= A check to see if a certain user has correct permissions on a file
```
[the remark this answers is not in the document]
Piankero: how to specify any combination or rwx?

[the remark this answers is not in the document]
markus2330: There are many ways, I am sure you will find one :smile:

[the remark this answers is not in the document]
Piankero: well, of course I could find one but I have no idea what this file does or how it is used or if it is used. I mean I could provide a regex under "type" or enumerate all possibilities or give a textual description. But I have no idea if I break anything or if I write something illegal then. Is there any documentation for this?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
 Arbitrary additional data can be saved too such as:
 `kdb set /recursive/define/#2/permission admin`
 
+## A more real example
```
Piankero: @markus2330 I hope that this example is much more illustrative for you. Please take a look at it and tell me anything which is unclear. It really took me quite some time to get this right

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+
+In order to better show how this plugin can be used, a more sophisticated example will be given.
```
markus2330: Its better if the examples show different aspects, not one "less real" and another "more real".

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+Imagine that you want to create your own menu with arbitrary submenus possible. The following example
```
markus2330: Menu like in a restaurant?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+Imagine an editor with a `File` menu with a `Print` option as well as a submenu called `Settings`.
+The `Settings` submenu has `Global Settings` and  `Autocorrect Options`
+Beside the `File` menu we also have a `Tool` menu which also has the same `Settings` submenu but also
+has gotten a `Extension Manager` and a Menu `Macros`. Macros has again a submenu `Organize Macros` which
+again is a menu and has `Macro Basics`.
+
+The following picture illustrates the menu tree:
+![menu-example](example.png?raw=true "Example")
+
+So how can we map this into an elektra configuration?
```
markus2330: configuration or specification?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+
+First we create a config and mount it. Next we will set the specification for all menu entries as following:
+```
+kdb mount config.dump /editor dump recursion
+sudo kdb setmeta /editor check/recursion/vertex "menu"
+sudo kdb setmeta /editor check/recursion/ref "menuref"
```
markus2330: Specification still not explained, what are these two lines for?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+```
+Each key definition in the form of `/editor/menu` will now be treated as recursive entry. You can define arbitrary
```
markus2330: /editor/menu is not a form but rather an example?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+many menues like this: `/editor/menu/#[d]` where `[d]` stands for a positive number.
```
markus2330: It should be in array form, link to array definition.

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+
+Menues can be linked via `menuref` like this: `/editor/menu/#[d]/menuref/#[d]` and is again in the array notation.
+Note that `menu` and `menuref` are our defined keywords in the setmeta command from above.
```
markus2330: That should be said earlier. First explain specification or configuration, do not mix up.

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+
+Next lets create all distinct menu entries such as `File`, `Print` etc. Note that the Settings
```
markus2330: let us?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+ menu appears twice but we will only need to declare it once. The order does not matter:
```
markus2330: order of array entries or order of `kdb set` commands?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+```
+
+Now we link the `Tools` section:
+```
+kdb set /editor/menu/#1/menuref/#0 /editor/menu/#5
+kdb set /editor/menu/#1/menuref/#1 /editor/menu/#3
+kdb set /editor/menu/#1/menuref/#2 /editor/menu/#7
+```
+
+Note that `Settings` is already linked. We do not need to do it for the `Tools` menu again.
```
markus2330: What happens if we link it again?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+```
+
+### Example of wrong settings
+
+In the above example the following settings will be rejected:
+
+```
+#Link "Macro Basics" to the parent menu "Macros" and create an endless loop
+kdb set /editor/menu/#9/menuref/#0 /editor/menu/#7
+ERR: 198
```
markus2330: Please use shell recorder syntax

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+Reason: Cyclical reference detected: "Macro Basics" -> "Macros" -> Organize Macros" -> "Macro Basics"
+
+#Link to something which does not exist
+kdb set /editor/menu/#9/menuref/#0 /editor/menu/#10
+ERR: 199
+Reason: Could not find vertex with #10. Maybe it does not exist yet?
```
markus2330: Why isn't the term also used in the tutorial?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+
+#### LCDproc's current format
+
+[This](https://github.com/lcdproc/lcdproc/blob/master/clients/lcdexec/lcdexec.conf) is the current format in
+lcdexec.
+
+Their approach is not array based but name based. 
+Their reference keyword (`check/recursion/ref`) to other settings is `Entry`.
+
+They do not use an array based approach though but give names instead.
```
markus2330: Can we add a name-based approach or does it confict with the array-based approach?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+#Link the Settings Menu
+kdb set /editor/SettingsMenu/Entry/#0 /editor/GlobalSettings
+kdb set /editor/SettingsMenu/Entry/#1 /editor/AutocorrectOptions
+
+#Give the displaynames
+kdb set /editor/GlobalSettings/DisplayName "Global Settings"
+kdb set /editor/AutocorrectOptions/DisplayName "Autocorrect Options"
+``` 
+
+This would at least guarantee reusability of settings.
```
markus2330: Why "at least"? What is missing?

src/plugins/recursion/README.md
```diff
@@ -81,10 +81,186 @@ kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
...
+
+The plugin could theoretically support an arbitrary name instead of an array. But how does the 
+specification know what is actually a menu entry and what is just some arbitrary other setting.
```
markus2330: Metadata?

Piankero: How could this look like?

src/plugins/recursion/README.md
Piankero: Do you have a link to a tutorial for that?

sanssecours: The following documents should be helpful:

- [ReadMe of the Markdown Shell Recorder (MSR)](https://master.libelektra.org/tests/shell/shell_recorder/tutorial_wrapper/README.md)
- [the MSR syntax check](https://master.libelektra.org/tests/shell/shell_recorder/tutorial_wrapper/SyntaxCheck.md)

. Many of the plugins also contain some basic MSR tests. For example, the ReadMe of [YAML CPP](https://www.libelektra.org/plugins/yamlcpp) contains a few example tests.

src/plugins/recursion/README.md
Piankero: I cannot find anything in this repo where the "array definition" is explained. Can you give me the link please?

src/plugins/recursion/README.md
Piankero: It doesn't conflicht imo, as discussed in your comment below it needs to know if it is a vertex or some arbitrary setting unrelated to the recursive structure. With the array based approach this is clear.

src/plugins/recursion/README.md
sanssecours: The [array tutorial](https://www.libelektra.org/tutorials/arrays) should provide some basic information about the array syntax.

Piankero:
I updated the documentation.
On sunday I will
1) Add src/error/specification for port & recursion
2) write the correct shell recorder syntax

Concerning the "named" recursion it would be interesting to have your feedback.
Also on the `Problem` part in the recursion plugin readme which I updated.

doc/METADATA.ini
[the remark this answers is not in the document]
markus2330: 99440157f06eb600b9a90f873dc2e1595bcfe1b0

src/plugins/recursion/README.md
markus2330: Imho, the array does not make a difference. Without some external knowledge you never know if a string is a reference or a value interpreted otherwise.

src/plugins/recursion/README.md
```diff
@@ -10,18 +10,37 @@ Also I have no idea what your `struct` plugin actually does. There is no example
...
+is a legal configuration. Loading and saving the configuration again could lead to a reversed structure
+
+```
+[a]
+ref = b
+
+[b]
+```
+
+which could cause an error since `a` cannot be loaded as `b` is yet non existent.
```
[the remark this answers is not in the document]
Piankero: Can you explain why? How exactly are keys loaded into the kdb?

I assume that it is loaded in the order of the file, i.e. it would be the same if I call kdb set ... in the same order as the config file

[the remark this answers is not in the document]
markus2330: The plugins receive a whole keyset and can do whatever they want to do with it. So in the implementation you can lookup if something is present, it does not matter where in the keyset it is.

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
+description= A check to see if a certain user has correct permissions on a file. Permissions are
+	either r (read) w (write) x (execute) and any combination of them.
+	 The respective user which must have these permissions is declared in check/permission/user.
+example = "rwx ; rw ; r ; x ; w ; wx"
+
+[check/permission/user]
+status= idea
+usedby/plugin= path
+type= string
+	empty
```
markus2330: empty or string?

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
+description= Used be check/permission to see if the given user has the correct permissions. If empty
```
markus2330: empty or non-existent?

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
+	it takes the current user.
```
markus2330: Can it be used without check/permission? Maybe assume rwx?

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
+
+[check/port]
+status = idea
+usedby/plugin = network
+type = empty
+desciption = If specified, checks if the value of the key is between 0 and 65535.
```
markus2330: As non-root it is 1025-65535. Without checking this it would make little sense to have a separate check? (check/type could be used).

And the translation from port name -> number should also be done (done by getaddrinfo anyway, see below)

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
+type = empty
+desciption = If specified, checks if the given port is open and ready to be listened to.
+
+[check/recursion]
+status = idea
+usedby/plugin = recursion
+type = string
+decription = Used to declare recursive structures like trees or graphs. Every key
+			  below the current key will be treated as vertex from which you can refer to by given string.
+example = eg if 'Entry' was specified, one could refer to other vertices by your/key/Entry/#0 <vertex key>
```
markus2330: Here should be only an example.

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
+usedby/plugin= path
+type= string
+	empty
+description= Used be check/permission to see if the given user has the correct permissions. If empty
+	it takes the current user.
+
+[check/port]
+status = idea
+usedby/plugin = network
+type = empty
```
markus2330: It is a number isn't it?

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
+desciption = If specified, checks if the value of the key is between 0 and 65535.
+
+[check/port/listen]
+status = idea
+usedby/plugin = network
+type = empty
+desciption = If specified, checks if the given port is open and ready to be listened to.
```
markus2330: relation to check/port not documented

src/error/specification
```diff
@@ -1209,3 +1209,39 @@ severity:error
...
+module:path
+
+number:195
+description:Could not find the given user in the system
+severity:error
+ingroup:plugin
+module:path
+
+number:196
+description:The given port is not in a valid range (0-65535)
```
markus2330: The range can be given in the reason.

src/error/specification
```diff
@@ -1209,3 +1209,39 @@ severity:error
...
+severity:error
+ingroup:plugin
+module:network
+
+number:196
+description:The desired port is already in use.
```
markus2330: Please be consistent with the already existing errors if you put a . in the end or not.

src/error/specification
```diff
@@ -1209,3 +1209,39 @@ severity:error
...
+severity:error
+ingroup:plugin
+module:network
+
+number:197
+description:Referring to the desired key would lead to an endless loop
+severity:error
+ingroup:plugin
+module:network
```
markus2330: network or recursive?

src/plugins/network/README.md
```diff
@@ -14,6 +14,8 @@ This plugin is a check plugin that checks if a key contains a valid ip
 address. It uses the `POSIX.1-2001` interface `getaddrinfo()` in order
 to check if an ip address is valid.
 
+The plugin can also check for valid port numbers and if the set port is free to use.
```
markus2330: The port should actually also passed to getaddrinfo. Better to rewrite the intro.

src/plugins/network/README.md
```diff
@@ -37,3 +39,14 @@ it to implement this plugin.
...
+
+If `check/port` is specified on a given key, the plugin will validate if the port is a 
+correct number between 1 and 65535.
+
+If `check/port/listen` is specified, the plugin will check if the application can be started
+and listen on the given port.
+
+## Future Work
+
+`check/port/ping` to check if the port can be pinged/reached (usually for clients).
```
markus2330: maybe connect is better, as ping somehow implies ICMP.

src/plugins/recursion/README.md
```diff
@@ -75,30 +66,33 @@ Users will have to set this structure in the following way:
...
 
 #For A2:
-kdb set /recursive/define/#2/ref/#0 /recursive/define/#4
-kdb set /recursive/define/#2/ref/#1 /recursive/define/#5
+kdb set /recursive/A2/ref/#0 B1
+kdb set /recursive/A2/ref/#1 B2
 ```
 
+It is not needed to add `/recursive` before every key as it is only possible to choose keys
+below this rootkey anyway. Referring to other keys is done by
```
markus2330: directly below?

doc/METADATA.ini
```diff
@@ -679,6 +679,43 @@ type= enum
...
 description= Support of some standards for a date specification.
 
+[check/permission]
+status= idea
+usedby/plugin= path
+type= enum
+description= A check to see if a certain user has correct permissions on a file. Permissions are
+	either r (read) w (write) x (execute) and any combination of them.
+	 The respective user which must have these permissions is declared in check/permission/user.
+example = "rwx ; rw ; r ; x ; w ; wx"
```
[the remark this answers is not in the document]
Piankero: What is a valid example? They are all just some descriptions as i can see. I made it more verbose now

doc/METADATA.ini
Piankero: either or. If empty then the current user is taken, if a string is given it is interpreted as the desired user. It is also written in the description

doc/METADATA.ini
Piankero: Isn't it the same question as above?

doc/METADATA.ini
Piankero: Yes but I think it would lead to more errors since for example `x` is rarely used and attached to a file. 
For a default value I would suggest only `r`.

Still I think it's the best to let the user decide which permissions should be possible as it is very different from file to file. Maybe when we gather more data it is clear?

doc/METADATA.ini
Piankero: What about emitting a warning when trying to set a port below 1025? I think the usability would suffer if users would need to add a separate check just to be assured that a port below 1025 is executed as root.

Also how exactly would you think of check/type? Can you give an example?

doc/METADATA.ini
Piankero: check/port does only a check if the range is correct. A number is not required imo. type is only used to declare the type of the metavalue

src/error/specification
Piankero: Some descriptions have a dot and some do not. Which one should i take now?

src/error/specification
Piankero: woops, copy and paste error

doc/METADATA.ini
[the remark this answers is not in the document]
markus2330: The example needs to be something that actually could be a value of the metavalue.

E.g. "rwx" would be a valid example.

doc/METADATA.ini
markus2330: string automatically includes empty. Elektra does not have sum types, so specifying "string empty" does not make sense. (The result would be "empty")

doc/METADATA.ini
markus2330: Yes, it is the same problem.

src/error/specification
markus2330: Ok, this is something to unify later.

We can also make that sentences get a .

doc/METADATA.ini
```diff
@@ -682,17 +682,24 @@ description= Support of some standards for a date specification.
 [check/permission]
 status= idea
 usedby/plugin= path
-type= enum
+type= r
```
markus2330: enum and then the possibilities

src/error/specification
```diff
@@ -1238,10 +1238,10 @@ number:197
...
-module:network
+module:recursion
 
 number:198
 description:Cannot refer to non-existent recursive entry. Please create it first
 severity:error
 ingroup:plugin
-module:network
\ No newline at end of file
+module:recursion
```
markus2330: Its better to have a new-line in the end of files.

src/plugins/network/README.md
```diff
@@ -14,7 +14,9 @@ This plugin is a check plugin that checks if a key contains a valid ip
 address. It uses the `POSIX.1-2001` interface `getaddrinfo()` in order
 to check if an ip address is valid.
 
-The plugin can also check for valid port numbers and if the set port is free to use.
+Furthermore `getaddrinfo()` is used in `check/port` to resolve a port by its service name
+which is defined under `/etc/services`.
```
markus2330: and it also transforms to the number?

doc/METADATA.ini
markus2330: You are right, the root check only makes sense for servers, not for clients. So there are 3 ways:

1. If a user simply wants a check from 0-65535, she would use `check/type=unsigned_short`.
2. If a user wants to check if a string can be used as port (number 0-65535 or a name from /etc/services), she would use check/port
3. If a user wants to check if a string can be used as port to listen to (as above + not occupied + root for 0-1024), she would use check/port/listen

Btw. is 0 a valid port number?

Piankero:
Hopefully now everything is accordingly.

Could you take a look at the new recursion plugin? It just contains a single check metadata now instead of 2.

kodebach:
@markus2330 I updated the recursion plugin README with results from a talk between me and @Piankero 

I would also suggest renaming the recursion plugin to `reference`, because one major part of the work the proposed plugin would do is checking, whether references are valid. This functionality could also be useful elsewhere without the recursive part.

I propose:
* changing `check/recursion` to `check/reference`
* changing `check/recursion/override` to `check/reference/override`
* changing `check/recursion/restrict` to `check/reference/restrict`
* adding the boolean `check/reference/recursive` to enable/disable the check for loops in the reference graph

An example of such reference checking without recursion would be the references to driver configurations in the lcdproc server configuration.

I would also suggest either closing this PR entirely and opening separate PRs for the proposed plugins, or at least opening a separate PR for the recursion/reference plugin, to keep things a bit more organized.

markus2330:
Yes, please let us open a new PR once we decided on the new name. If the checks are not limited to recursive checks, it makes sense to rename it. How can a user distinguish if a reference check is recursive or only a single check? What if we remove the old "struct" check and call the new plugin also "struct"? Is this plugin able to check some arbitrary recursive structure of configuration?

Btw. if possible we should avoid "override" because this already has a meaning within Elektra (it is a link to be preferred, see doc/METADATA.ini).

kodebach:
> How can a user distinguish if a reference check is recursive or only a single check?
```ini
[/tests/keywithref]
check/reference = ref

[/tests/keywithrec]
check/reference = ref
check/reference/recursive = true
```
Here `/tests/keywithref/ref` would just be a single reference, while `/tests/keywithrec/ref` would be checked recursively. 

An alternative would be the following:

```ini
[/tests/keywithref/ref]
check/reference = single

[/tests/keywithrec/ref]
check/reference = recursive
```
Here the type of `check/reference` would be `enum`, with the values `single` and `recursive` being supported. The name used for recursive checking would be the basename of the key with the `check/recursive` metadata. Then overriding the name could be done through a third enum value.

Personally I like the second option better. It makes more sense to me to specify the metadata on the actual reference key.

> What if we remove the old "struct" check and call the new plugin also "struct"? Is this plugin able to check some arbitrary recursive structure of configuration?

The current proposed plugin would only check references. Checking the remaining structure could simply be done with the `spec` plugin:

```ini
[/tests/base/ref]
check/reference = recursive
check/reference/restrict = typeA

[/tests/base/typeA/_/name]
check/type = string
; etc

[/tests/base/typeA/_/size]
check/type = long
; etc
```
The basic idea is to restrict the references to a certain path `X` and then using `X` as a kind of type identifier. The direct sub-keys of `X` would then be 'instances' of `X`, so by using `X/_/Y` one can specify properties of the field `Y` of type`X`.

markus2330:
> Personally I like the second option better. It makes more sense to me to specify the metadata on the actual reference key.

Yes, I also like the second option better.

> The basic idea is to restrict...

Good example. Could you please incorporate it into the README? But please continue using typeA (if I understood correctly it is changed to X later in the explanation).

Piankero:
@markus2330 
There is a problem with the extension of the `path` plugin and the assumption if the metakey`check/permission/user` is not set, the current user is taken.

The problem lies in the `euidaccess` function and the need to change the egid.

Is is a bit difficult to explain:\
*) `euidaccess` checks for the access with the egid and euid.
*) unfortunately there can only be one egid active at a time when calling `euidaccess()`.
*) So if a user belongs to multiple groups and the filegroup is one of them, I have to actively switch to this group with `setegid()` before calling `euidaccess()`
*) To switch via `setegid()` I have to be running as root even though I am a member of the group I want to "switch" to.
*) When running as root though, the information is lost for which user you really want to check the permission for.

So it is absolutely required that check/permission/user and check/permission/types are both present.
I can only support to assume if check/permission/user is empty the root user is assumed. But checking for file access as root user is a bit fruitless.

I hope I could explain the problem so that it is understandable. Do you have a better solution or do we simply require the coexistence of both metadatas?

markus2330:
What is "check/permission/types"?

> But checking for file access as root user is a bit fruitless.

I am not sure if it is fruitless. There are now many kernel extensions (and also file system features) to disallow root to do something. And I would prefer the "current user" and not root as default. And in the "current user" scenario no switching of egid is necessary.

If two options are absolutely required so that the plugin can do something useful it is of course possible to demand the presence of both options. Afaik this would be the first plugin, though.

Piankero:
> What is "check/permission/types"?

 It is used for the actual permissions, e.g. "rwx"

> And I would prefer the "current user" and not root as default. And in the "current user" scenario no switching of egid is necessary.

I wish for the same but when programming I could not circumvent the problem described above. But I have to switch egid since access only checks with one egid active. If you wish I could also demonstrate it on our next Elektra meeting.

Piankero:
All plugin ideas are now in implementation stage in their respective PR. Closing this PR

PR #2173, opened 2018-08-10
Test: Add C++ Version of `compare_keyset`

sanssecours:
# Checklist

- [x] I added Doxygen comments for the functions added in this pull request.

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -475,7 +475,7 @@ Thanks to Daniel Bugl.
   - conversions to non-pointer type (GCC), and
   - attribute warnings (GCC),
 
-  in the Ruby binding and plugin. *(René Schwaiger)*
+  caused by code generated via [SWIG](http://www.swig.org) in the Ruby binding and plugin. *(René Schwaiger)*
 
 [Google Test]: https://github.com/google/googletest
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -335,7 +335,7 @@ Thanks to Daniel Bugl.
   should be more clear how to store plugin data besides pluginprocess's data structure.
   Tests have been added to the library to ensure its correct functionality. *(Armin Wurzinger)*
 - Anonymous pipes are now used instead of named pipes for the communication as anonymous pipes get
-  terminated by the OS in case a child process dies before writing back data to the parent. 
+  terminated by the OS in case a child process dies before writing back data to the parent.
   Currently the parent process will freeze otherwise attempting to read from the child. *(Armin Wurzinger)*
 
 ## Bindings
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -427,6 +427,7 @@ Thanks to Daniel Bugl.
   test reported [memory leaks](https://issues.libelektra.org/2113) on the latest version of Debian Unstable. *(René Schwaiger)*
 - The [CFramework](https://master.libelektra.org/tests/cframework) macro `compare_keyset` now supports the comparison of two empty key sets. *(René Schwaiger)*
 - The C++ version of the macro `exit_if_fail` now really exits the test progamm if the test fails. *(René Schwaiger)*
+- The C++ testing framework now supports the macro `compare_keyset` that checks if two key sets are equal. *(René Schwaiger)*
 
 [#1887]: https://github.com/ElektraInitiative/libelektra/issues/1887
 [Markdown Shell Recorder]: https://master.libelektra.org/tests/shell/shell_recorder/tutorial_wrapper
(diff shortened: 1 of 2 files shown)


src/bindings/cpp/tests/tests.hpp
```diff
@@ -37,8 +39,147 @@ using namespace kdb;
...
+bool isMetaDataEqual (kdb::Key & key1, kdb::Key & key2)
+{
+	key1.rewindMeta ();
+	key2.rewindMeta ();
+
+	while (key1.nextMeta ())
+	{
+		key2.nextMeta ();
+		if (!key2.currentMeta ()) return false;
+		if (key1.currentMeta ().getName () != key2.currentMeta ().getName ()) return false;
```
markus2330: What if key1 has no meta data?

src/bindings/cpp/tests/tests.hpp
```diff
@@ -37,8 +39,147 @@ using namespace kdb;
...
+/**
+ * @brief This function checks if two keys are equal.
+ *
+ * @param key1 This parameter stores the first key this function compares.
+ * @param key2 This parameter stores the second key this function compares.
+ *
+ * @retval true if all attributes of `key1` and `key2` are equal
+ * @retval false otherwise
+ */
+bool isKeyEqual (kdb::Key & key1, kdb::Key & key2)
```
markus2330: maybe we can have this in the cframework and reuse it?

doc/news/_preparation_next_release.md
```diff
@@ -464,6 +464,14 @@ Thanks to Daniel Bugl.
 - We improved the automatic detection of Libgcrypt and OpenSSL. *(René Schwaiger)*
 - Resolved an issue where cmake did not properly set test feature macros to detect and use libc functionality. *(Lukas Winkler)*
 - Improve the detection of `ftw.h`, if the current build use the compiler switch `-Werror`. *(René Schwaiger)*
+- We now ignore warnings about
```
markus2330: Maybe mention in the beginning that this is mostly about swig generated code?

src/bindings/cpp/tests/tests.hpp
sanssecours: In that case the code should execute line 111:

```cpp
key1.nextMeta () == key2.nextMeta ()
```

, which should evaluate to `false` unless `key2` also does not contain any metadata.

src/bindings/cpp/tests/tests.hpp
sanssecours: While I am definitely a fan of code reuse, there are two major problems here:

- my laziness, and
- the code below is written in C++ (and not C)

.

src/bindings/cpp/tests/tests.hpp
markus2330: But doesn't line 107 already throw a KeyException in that case? (currentMeta() will evaluate to null and then getName is called on a null key, isn't it?)

sanssecours: > But doesn't line 107 already throw a KeyException in that case?

As far as I can tell, the processor should not execute line 107 if `key1` does not contain any meta, since `key1.nextMeta ()` evaluates to `false` in the condition of the `while` loop.

sanssecours: Oh, I see. The function `isKeySetEqual(keys1, keys2)` is broken, since it will return `true` if `keys1` is a proper subset of `keys2`. Thank you for the hint.

src/bindings/cpp/tests/tests.hpp
```diff
@@ -37,8 +39,147 @@ using namespace kdb;
...
+ * @param key This parameter stores the key this function converts to an assertion result.
+ *
+ * @returns An assertion result containing the string representation of `key`
+ */
+testing::AssertionResult & operator<< (testing::AssertionResult & stream, kdb::Key & key)
+{
+	stream << key.getName () << ": " << (key.isString () ? key.getString () : (key.getBinarySize () == 0 ? "NULL" : "BINARY"));
+	key.rewindMeta ();
+
+	key.rewindMeta ();
```
markus2330: rewindMeta twice?

src/bindings/cpp/tests/tests.hpp
```diff
@@ -37,8 +39,147 @@ using namespace kdb;
...
+ *
+ * @retval AssertionSuccess if all keys of `keys1` and `keys2` are equal
+ * @retval AssertionFailure otherwise
+ */
+testing::AssertionResult compareKeySet (kdb::KeySet & keys1, kdb::KeySet & keys2)
+{
+	if (isKeySetEqual (keys1, keys2))
+		return testing::AssertionSuccess ();
+	else
+		return testing::AssertionFailure () << endl
```
markus2330: One endl at the end should be enough, otherwise `\n` is more appropriate (does not flush the stream).

markus2330:
Thank you!

PR #2181, opened 2018-08-14
Directory Value: Support Nested Arrays

sanssecours:
# Checklist

- [x] I added unit tests.
- [x] The updated code contains comments, logging statements, and assertions.
- [x] The metadata of the Directory Value plugin should be up to date.


src/plugins/directoryvalue/directoryvalue.cpp
```diff
@@ -0,0 +1,132 @@
...
+	CppKey parent{ parentKey };
+
+	int status = ELEKTRA_PLUGIN_STATUS_ERROR;
+	try
+	{
+		status = delegator::get (handle)->convertToLeaves (keys);
+	}
+	catch (range_error const & error)
+	{
+		ELEKTRA_SET_ERROR (ELEKTRA_ERROR_DIRECTORY_VALUE_ARRAY, *parent, error.what ());
```
markus2330: status is not changed here?

src/plugins/directoryvalue/directoryvalue.cpp
```diff
@@ -0,0 +1,132 @@
...
+	}
+	catch (exception const & error)
+	{
+		ELEKTRA_SET_ERROR (ELEKTRA_ERROR_UNCAUGHT_EXCEPTION, *parent, error.what ());
+	}
+
+	parent.release ();
```
markus2330: Would be great if we can move this stuff into a shared C++ part.

src/plugins/directoryvalue/directoryvalue_delegate.cpp
```diff
@@ -0,0 +1,524 @@
...
+}
+
+/**
+ * @brief This function splits the given keyset into array leaves (first element of an array parent) and other keys.
+ *
+ * @param input The function searches for array leaves in this key set.
+ *
+ * @return A pair of key sets, where the first key set contains all array leaves and the second key set contains all other keys
+ */
+pair<CppKeySet, CppKeySet> splitArrayLeavesOther (CppKeySet const & arrayParents, CppKeySet const & keys)
```
markus2330: Maybe have some typedef/using for this pair <>

src/plugins/directoryvalue/directoryvalue_delegate.cpp
```diff
@@ -0,0 +1,524 @@
...
+ * @return A copy of the input, where the last level of each key was removed
+ */
+CppKeySet removeBaseName (CppKeySet const & keys)
+{
+	CppKeySet directories;
+
+	for (auto key : keys)
+	{
+		CppKey directory = key.dup ();
+		keySetBaseName (*directory, 0);
```
markus2330: Why using C here? setBaseName exists?

src/plugins/directoryvalue/directoryvalue_delegate.cpp
```diff
@@ -0,0 +1,524 @@
...
+ * @param parent This parameter specifies a possible array parent.
+ * @param keys This variable stores the key set of `parent`.
+ *
+ * @retval true If `parent` is the parent key of an array
+ * @retval false Otherwise
+ */
+bool isArrayParent (CppKey const & parent, CppKeySet const & keys)
+{
+	CppKeySet children = accumulate (keys.begin (), keys.end (), CppKeySet{}, [&parent](CppKeySet collected, CppKey key) {
+		if (key.isBelow (parent)) collected.append (key);
```
markus2330: Its better to have {} around here.

src/plugins/directoryvalue/directoryvalue_delegate.cpp
```diff
@@ -0,0 +1,524 @@
...
+	CppKeySet arrayParents;
+	CppKeySet others;
+
+	keys.rewind ();
+	CppKey previous;
+	for (previous = keys.next (); keys.next (); previous = keys.current ())
+	{
+		bool previousIsArray =
+			previous.hasMeta ("array") ||
+			(keys.current ().isBelow (previous) && keys.current ().getBaseName ()[0] == '#' && isArrayParent (previous, keys));
```
markus2330: Maybe one more bool for this expression (to give it a name).
It seems like the bool can be const.

src/plugins/directoryvalue/directoryvalue.cpp
sanssecours: Yep, but since the initial value is `ELEKTRA_PLUGIN_STATUS_ERROR` this should not be a problem, at least as far as I can tell.

src/plugins/directoryvalue/directoryvalue.cpp
sanssecours: Yeah, that would be great. We only have to find someone willing to do that kind of work 😊…

src/plugins/directoryvalue/directoryvalue_delegate.cpp
sanssecours: Since the statement fits on a single line I like to keep it like it is.

src/plugins/directoryvalue/directoryvalue_delegate.cpp
sanssecours: I tried to use `directory.setBaseName (0)`, but this function call causes segmentation faults. As far as I can tell there is no option to remove the basename in the C++ API. At least I only found the C++ method:

```cpp
void setBaseName (const std::string & baseName);
```

, which takes a string as argument. If I add the following function:

```cpp
inline void Key::setBaseName (char const * baseName)
{
	if (ckdb::keySetBaseName (getKey (), baseName) == -1)
	{
		throw KeyInvalidName ();
	}
}
```

everything works fine.

src/plugins/directoryvalue/directoryvalue_delegate.cpp
markus2330: Ok, because it would make little sense to have the return within the if.

src/plugins/directoryvalue/directoryvalue_delegate.cpp
markus2330: Thank you, there were various errors in this documentation. I tried to fix them all, please open an issue if there are problems in the docu.

I created a new issue #2182 to add a delBaseName(). Please also open issues if the C++ binding does not allow you to do something which is possible in the C API.

markus2330:
Thank you, great job!

But please refrain from rewriting C plugins to C++ or let us at least discuss it beforehand. There are people who do not like C++ deps and it is good for them if we have many plugins in C.

sanssecours:
> There are people who do not like C++ deps and it is good for them if we have many plugins in C.

I definitely can understand that. However, I think I will not write any more plugin code in C, unless I have to.

markus2330:
Of course you can write your (new) plugins in C++. My concern was only about not rewriting all existing C plugins to C++. In most of the cases it should be easier to fix a bug in C than to rewrite it to C++ anyway :smile:

Issue #2180, opened 2018-08-13
How to get the typechecker plugin into kdb


Piankero:
I cannot get the typechecker plugin running.

I used 
```sh
mkdir build;
cd build
ccmake ..
#deleted experimental and added "typechecker". (see screenshot)
cmake ..
make
```
![typechecker](https://user-images.githubusercontent.com/12346621/44024036-6ec630de-9eec-11e8-9a7a-2f744abd1ec7.jpg)

What am I doing wrong?

sanssecours:
As far as I know there is no `typechecker` binding. You need the `typechecker` **plugin**, which requires the `haskell` binding. For an example, on how to get the thing running on macOS (the steps on Linux should be pretty much the same), please take a look at the `🍏 Haskell` build job in [`.travis.yml`](https://github.com/ElektraInitiative/libelektra/blob/master/.travis.yml#L199-L222).

markus2330:
Yes. you need the haskell binding (It has quite some deps: https://www.libelektra.org/bindings/haskell), then the haskell plugin (again deps: https://www.libelektra.org/plugins/haskell) so that you can finally enable the typechecker plugin.

@e1528532 Can you update the docu of the typechecker which parts are needed so that it will work? And why is `infos/needs` in the contract duplicated?

Piankero:
This installation process is extremely cumbersome, especially for linux. You require `GHC (Glasgow Haskell Compiler) > 8.0.1 && < 8.4`. The package manager just has 7.x, stack already gives you 8.4.3.
After tagging a specific version I still receive an error...

1) On various places i get the bug:
```
CMake Error at cmake/Modules/FindHaskell.cmake:61 (string):
  string sub-command REPLACE requires at least four arguments.
Call Stack (most recent call first):
  src/plugins/typechecker/CMakeLists.txt:7 (find_package)
```
2) 
```-- Exclude Binding haskell because ghc:  found, but 8.0.1 required```
When I run `ghc --version` I receive the `The Glorious Glasgow Haskell Compilation System, version 8.2.2`
3) 
```
-- Exclude Plugin typechecker because GHC_RTS_LIB not found
```
What is this? Where can i find it?

e1528532:
[...]

PR #2195, opened 2018-08-18
haskell-xenial

e1528532:
## Basics

add ghc to our xenial docker image to test on ubuntu as there seem to be complications on xenial #2180 

## Checklist

- [ ] I added unit tests

### DIFF
--- a/scripts/docker/ubuntu/xenial/Dockerfile
+++ b/scripts/docker/ubuntu/xenial/Dockerfile
@@ -28,6 +28,8 @@ RUN apt-get update && apt-get -y install \
         valgrind \
         checkinstall \
         swig \
+        libgmp-dev \
+        wget \
     && rm -rf /var/lib/apt/lists/*
 
 # Google Test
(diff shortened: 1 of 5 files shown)


scripts/docker/ubuntu/xenial/Dockerfile
```diff
@@ -56,4 +58,21 @@ RUN useradd \
     --shell "/bin/bash" \
     jenkins
 
+# Handle Haskell dependencies
+# We use ghc8.2.2 from the haskell platform website here, as xenial only provides ghc7
+RUN wget https://haskell.org/platform/download/8.2.2/haskell-platform-8.2.2-unknown-posix--core-x86_64.tar.gz \
```
sanssecours: I assume building Haskell from source will take quite some time. Would it maybe make sense to use the [personal package archive (PPA) for GHC of Herbert V. Riedel](https://launchpad.net/~hvr/+archive/ubuntu/ghc) instead? Since the [official Haskell homepage](https://www.haskell.org/ghc/distribution_packages.html#ubuntu) als mentions this PPA, I guess the support for this third party repo should be quite good.

e1528532: this is the dockerfile so it will be only built once, imo that shouldn't make a big difference.
This branch should mostly show whether its possible on xenial in general and doesn't directly resemble what piankero did no.

e1528532: this is the dockerfile so it will be only built once, imo that shouldn't make a big difference.
This branch should mostly show whether its possible on xenial in general and doesn't directly resemble what piankero did no.

e1528532:
interesting. It compiled successfully (@Piankero i hope this helps you in getting it running), but has a test error in the regex dispatcher tests, will investigate.

markus2330:
Thank you for looking into this! It seems like compatibility between Haskell libraries is really quite neglected.

markus2330:
Not relevant anymore as Haskell support has been removed.

PR #2217, opened 2018-09-01
cmake: add -Werror to a build job

### LINKED ISSUE #1812: Add -Werror to all Build Jobs
```
/home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c: In function ‘elektraOpmphmInvalidate’:
/home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c:48:47: warning: unused parameter ‘ks’ [-Wunused-parameter]
 static void elektraOpmphmInvalidate (KeySet * ks)
                                               ^~
/home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c: In function ‘elektraOpmphmCopy’:
/home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c:65:41: warning: unused parameter ‘dest’ [-Wunused-parameter]
 static void elektraOpmphmCopy (KeySet * dest, const KeySet * source)
                                         ^~~~
/home/markus/Projekte/Elektra/current/src/libs/elektra/keyset.c:65:62: warning: unused parameter ‘source’ [-Wunused-parameter]
 static void elektraOpmphmCopy (KeySet * dest, const KeySet * source)
                                                              ^~~~~~
```

### LINKED ISSUE #2218: On check_symbol_exists and werror
https://cmake.org/Bug/view.php?id=13208

A generated c file that checks for availability of fork:

```
[winlu@virtualArch libelektra]$ cat build/CMakeFiles/CMakeTmp/CheckSymbolExists.c 
/* */
#include <sys/types.h>
#include <unistd.h>

int main(int argc, char** argv)
{
  (void)argv;
#ifndef fork
  return ((int*)(&fork))[argc];
#else
  (void)argc;
  return 0;
#endif
}
```

The error when compiling (with our flags as defined in ElektraCompiling) :
```
build/CMakeFiles/CMakeTmp/CheckSymbolExists.c:9:11: error: ISO C forbids conversion of function pointer to object po
inter type [-Werror=pedantic]
   return ((int*)(&fork))[argc];
           ^
cc1: all warnings being treated as errors
[winlu@virtualArch libelektra]$
```

Hence adding -Werror is not as easy as thought (breaks out cmake). and explaining why https://github.com/ElektraInitiative/libelektra/commit/5b800ceaddcd59b4711e776e256ed8277f906508 had an effect on the detection of functions.
I already argued in https://github.com/ElektraInitiative/libelektra/pull/2159 that the commit is basically snake oil and only hides issues instead of fixing the real issue.
However I could not confirm my original assumption that without -Werror present check_symbol_exists might pass on invalid code since the function might be implicitly declared.

I am unsure on how to best approach on how to reliable determine which symbols/functions are available now across the whole codebase.
I am also not hopeful on an upstream fix as the issue is known for several years now.

One option would be to provide our own implementation which would disable -Wpedantic for all checks.

This blocks efforts that rely on detecting exported symbols such as #2078 or the recently opened #2217 that tries to add -Werror to builds.

ingwinlu:
This PR resolves issues with check_symbol_exists by providing a safe alternative. Also adds -Werror to one of our jenkins build tasks.

Resolve #1812 
Resolve #2218 

## Basics

## Checklist

need to be checked.

- [ ] I added unit tests


markus2330:
Thank you for looking into the `-Werror`!

I do not think that an ENABLE_WERROR makes any sense, there are already several ways to add (any) compiler flags. It looks like you can also simply set COMMON_FLAGS.

ingwinlu:
Does not really matter until #2218 is resolved.

markus2330:
Thank you so much for making this a reality :100:

PR #2233, opened 2018-09-11
network: Changed to allow non-numerical hosts

Piankero:
I fixed the network plugin to allow non-numerical hosts like "localhost" too.


Piankero:
The build says that I leak memory which is very confusing as I just moved a line into an `if` condition. Does anybody has experience with that?

```
 99/198 Test  #99: testmod_network .........................***Failed    0.21 sec
=================================================================
==951==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 65536 byte(s) in 1 object(s) allocated from:
    #0 0x4f0988 in malloc (/home/travis/build/ElektraInitiative/build/bin/testmod_network+0x4f0988)
    #1 0x7f5fb3cd6d05  (/lib/x86_64-linux-gnu/libresolv.so.2+0xbd05)
SUMMARY: AddressSanitizer: 65536 byte(s) leaked in 1 allocation(s).
```

markus2330:
Maybe you need to specify the flag and you are now in undefined behavior?

Piankero:
@tom-wa Did you ever have such a problem with memory leak?
https://travis-ci.org/ElektraInitiative/libelektra/jobs/428298534

sanssecours:
I tried to debug the current code. It seems like setting `hints.ai_flags` to a default value such as `AI_V4MAPPED | AI_ADDRCONFIG` [removes the memory leak](https://travis-ci.org/sanssecours/elektra/builds/428570183). However, afterwards the test

```c
testIPAny ("::1", 1);
```

fails. Since setting `hints.ai_flags` to `AI_V4MAPPED | AI_ALL` [reintroduces the memory leak](https://travis-ci.org/sanssecours/elektra/builds/428581460) (and fixes the failing test), I assume the problem is the `libc` version of Ubuntu Trusty Tahr.

Piankero:
Thank you very much for that non-trivial bug finding! So I guess we now have to decide between
1. Ignoring a Testcase & Addind a default with `AI_V4MAPPED | AI_ADDRCONFIG`
2. Fixing the test with `AI_V4MAPPED | AI_ALL` but accepting a memory leak when using the network plugin

if I did not misunderstand you

sanssecours:
More or less yes. I think **we do not need to set `hints.ai_flags`** (to a default value of `AI_V4MAPPED | AI_ALL`), since in that case the plugin behaves just like it does now. The proper solution is probably disabling the test case that leaks memory, if the systems ships with an old version of `glibc`. I added preliminary code that **might** fix the problem [here](https://github.com/sanssecours/elektra/compare/d1b056ee7ef5...0c4aa602a34a). It will still take some time until [Travis](https://travis-ci.org/sanssecours/elektra/builds/428601077) checks, if the updated code works or not though.

sanssecours:
I just pushed a [commit](https://github.com/ElektraInitiative/libelektra/pull/2233/commits/f0f6bdb8e2c57f249da1fa9e3b8c7a2eedbcf157) that should disable the tests that leak memory (`testIPAny`), if the system uses an old version of `glibc` (version `2.19` or earlier). Please feel free to overwrite or delete the commit, if you do not like it.

markus2330:
Thank you!

PR #2238, opened 2018-09-14
Reference plugin

kodebach:
## Basics

## Checklist

- [ ] I added unit tests


src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
+- infos = Information about the template plugin is in keys below
```
markus2330: template -> reference

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+- infos/author = Klemens Böswirth <k.boeswirth+git@gmail.com>
+- infos/licence = BSD
+- infos/needs =
+- infos/provides =
+- infos/recommends =
+- infos/placements = presetstorage
+- infos/status = nodep libc preview experimental unfinished
+- infos/metadata = check/reference check/reference/restrict
+- infos/description = Useful plugin for validating singular or recursive references
```
markus2330: remove "Useful" (hopefully every plugin is useful)

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Introduction
+
+This plugin serves one single purpose: the validation of reference values inside of keys.
```
markus2330: What are reference values?

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+To specify a key as a reference add the metakey `check/reference`. Currently the metakey 
+supports the three values `single`, `recursive` and `alternative`. While the value 
+`alternative` only makes sense in a certain context inside the `recursive` mode, the other
+two values define to different modes of operation for the plugin. 
+
+If a key has the metakey `check/reference` with value `single`, the plugin reads the value
+of the key and produces an error, if the value is not a valid reference.
```
markus2330: What is a valid reference?

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+If a key has metakey `check/reference` with value `recursive`, the plugin constructs a 
+reference graph starting with the marked key. (for the exact construction see below) 
+In this reference graph each node corresponds to a key in the KDB, while each edge 
+represents a reference between to keys. The plugin will produce an error, if this graph
+is not a directed acyclic graph or contains any invalid references.
+
+Additionally to this basic functionality the plugin allows one to restrict the set of valid
+references using the `check/reference/restrict` metakey. The exact mechanism for this is 
+described below.
```
markus2330: Better to not mention it then?

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Resolution of references
+If the key `A` is interpreted as a reference by the plugin, it has to be an array. Each of
```
markus2330: Why does it need to be an array for the key A? Is this an example?

Having everything as array is a bit tailored towards the INI array feature. Aren't references also useful without arrays?

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Resolution of references
+If the key `A` is interpreted as a reference by the plugin, it has to be an array. Each of 
+the array elements must contain a so called reference value. This reference value will be 
+resolved into a keyname, which will then be validated. The resolution goes as follows:
+
+* If the reference value starts with `/`, it is treated as absolute and interpreted as a 
+  keyname directly.
+* Otherwise the reference value will be interpreted as relative to the **parent** of `A`.
+  That means a reference which does not contain a `/` will also was refer to a sibling of `A`,
```
markus2330: This is done differently in the mathcheck plugin: "Keynames are all either relative to to-be-tested key (starting with `./` or `../`), relative to the parentkey (starting with `@/`) or absolute (e.g. `system/key`).".

It is better if we stay at this syntax.

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+  the reference array key.
+* `.` and `..` will be treated the same way as in UNIX paths. If these are used in an absolute
+  reference, the plugin will emit a warning, because their use in this case is unnecessary and
```
markus2330: It should be the other way round: "syste/key" is more likely to be a mistake then "./syste/key".

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+  That means a reference which does not contain a `/` will also was refer to a sibling of `A`,
+  the reference array key.
+* `.` and `..` will be treated the same way as in UNIX paths. If these are used in an absolute
+  reference, the plugin will emit a warning, because their use in this case is unnecessary and 
+  might be a mistake. The same is true, if `.` is used anywhere, but at the start of the 
+  reference.
+
+### Construction of the reference graph
+
+The process starts with a key `X`, which has the metakey `check/reference` set to the value
```
markus2330: Which one is the key X?

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+`recursive`. This key `X` must be an array key. The basename of `X` will be called the `refname`.
+For each element in the array run the following process:
+
+1. Interpret the current key `K`s value as a referemce and throw an error, if the reference
+   is not valid.
+2. If there is no node for `K` in the reference graph create one. If there is no node for 
+   `R` in the reference graph create one. Insert an edge from `K` to `R` into the graph.
+3. Find a key `K1` directly below `R` that has the `refname` as its basename. Interpret this 
+   key as an array, if found. For each array element, start the process recursively from 1
```
markus2330: no comma before if

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+3. Find a key `K1` directly below `R` that has the `refname` as its basename. Interpret this 
+   key as an array, if found. For each array element, start the process recursively from 1 
+   and then continue with 4.
+4. Find any key `K2` directly below `R` that has the metakey `check/reference` set to 
+   `alternative`. Interpret any such key as an array. For each array element, start the 
+   process recursively from 1, but this time using the basename of `K2` (the of the array)
+   as the `refname`.
+
+Once the whole process is finished, we will have a graph of the whole reference structure 
+stemming from `X`. This graph can then be check for acyclicity.
```
markus2330: the same is needed for the fallback/override links. Maybe do this in the same plugin? Or in a separate plugin reusing the graph code?

src/plugins/reference/README.md
```diff
@@ -0,0 +1,128 @@
...
+be valid.
+
+The restrictions themselves are UNIX-like pathname globbing expression, as defined by 
+[`fnmatch(3)`](http://man7.org/linux/man-pages/man3/fnmatch.3.html) using the `FNM_PATHNAME` 
+option. Path segments equal to `.` and `..` are interpreted like they are in UNIX-paths. If 
+the restriction starts with a `/` it will be treated as an absolute path, and the use of `.` 
+and `..` is unnecessary and discouraged. Therefore a warning will be emitted, if `.` or `..` 
+are encountered. The same is true if a path segment other then the first one is `.`.
+
+If a keyname matches the globbing expression it will be considered a valid reference
```
markus2330: See the spec plugin, Elektra has small differences in globbing characters.

_ matches key names (except arrays) and # matches array entries. See src/libs/elektra/keyname.c

src/plugins/reference/examples/alternative/README.md
```diff
@@ -0,0 +1,36 @@
+# Example: Alternative references
+
+For this example the file [alternative.ini](alternative.ini) will be used.
+We mount it by executing:
+
+```sh
+kdb mount alternative.ini user/tests/reference/alternative ni reference
```
markus2330: It is better to use HERE documents, then this works as shell recorder script.

src/plugins/reference/README.md
kodebach: You are right, we probably should use the same approach as described for `check/reference/restrict`.

src/plugins/reference/README.md
kodebach: I didn't know about these conventions, I will change it.

src/plugins/reference/README.md
kodebach: Neither of these two would produce a warning.
I would only produce a warning if the path was something like "system/key/subkey/../othersubkey",  "system/key/./subkey" or "./somekey/./key". We could also include keys like "../../somekey/../otherkey".

My reasoning was that these kinds of paths can be expressed in another shorter way and there is no need for them to be used other then to make things more complicated.

src/plugins/reference/README.md
kodebach: My plan was to represent the graph as a separate KeySet. So I think it would be easiest to just copy the code to wherever fallback/override links are currently processed.

src/plugins/reference/README.md
kodebach: `X` is the key which has `check/reference` set to `recursive`.

So in the case of `kdb setmeta user/tests/somekey/ref check/reference recursive` the key `user/tests/somekey/ref` will be `X`.

src/plugins/reference/README.md
markus2330: Maybe we can also add it to libmeta, it seems to fit to the topological sort. Or we move elektraSortTopology together with your function to libgraph?

src/plugins/reference/README.md
markus2330: Good question, it was rather by accident that * still works but it somehow makes sense to be not incompatible to glob(3) as people already know it (but only extend it to only match (non-)arrays).

And yes, a good idea to move the matching code. Unfortunately "fnmatch" might cause the library to be removed and there are already quite many libs depending on libease. So maybe it is better to create a libglob?

kodebach:
@markus2330 please review code

The README will need some more work and some code should probably be moved to other places, but all functionality should now be implemented.

src/libs/globbing/globbing.c
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * Valid array items consist of a '#' followed by <code>n</code> underscores ('_'),
+ * followed by <code>n+1</code> digits ('0'-'9'). Additionally the digits describe a
+ * valid 32-bit Integer (i.e. <code>0 <= x < 2^32</code>).
+ *
+ * @param name the string to check
+ * @retval true if @p name is a valid array item
+ * @retval false otherwise
+ */
+bool isArrayName (const char * name) // TODO: move? definitely useful elsewhere, e.g. reference plugin, libease, etc.
```
markus2330: Isn't this the same as elektraArrayValidateName from libease?

If not, putting this to libease makes more sense.

src/include/kdbglobbing.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifdef __cplusplus
+namespace ckdb
+{
+extern "C" {
+#endif
+
+#define ELEKTRA_GLOB_NOMATCH (-1)
+
+int keyGlob (const Key * key, const char * pattern);
```
markus2330: New public names should start with elektra (so it should be elektraKeyGlob)

src/libs/globbing/globbing.c
```diff
@@ -0,0 +1,212 @@
...
+ * @brief checks whether a given Key matches a given globbing pattern
+ *
+ * The globbing patterns for this function are a superset of those from fnmatch(3)
+ * used with the FNM_PATHNAME flag:
+ * <ul>
+ * 	<li> '*' matches any series of characters other than '/'</li>
+ * 	<li> '?' matches any single character except '/' </li>
+ * 	<li> '#', when used as "/#/" (or "/#" at the end of @p pattern), matches a valid array item </li>
+ * 	<li> '_', when used as "/_/"(or "/_" at the end of @p pattern), matches a key part that is <b>not</b> a valid array item </li>
+ * 	<li> if the pattern ends with "&#47;**", matching key names may contain arbitrary suffixes </li>
```
markus2330: Thank you, very nice feature! I was just thinking this would be great. Does this also makes sense within the key name?

And what about using `__` so that it is clear that it is an Elektra extension?

About how it is written down here: Why using '&#47;' instead of /?

src/libs/globbing/globbing.c
```diff
@@ -0,0 +1,212 @@
...
+ * </ul>
+ *
```
markus2330: How to match literal _ and #? (How to escape glob characters?)

src/libs/globbing/globbing.c
```diff
@@ -0,0 +1,212 @@
...
+		digits++;
+	}
+
+	return digits + underscores + 2 <= ELEKTRA_MAX_ARRAY_SIZE && digits == underscores + 1;
+}
+
+/**
+ * @brief checks whether a given Key matches a given globbing pattern
+ *
+ * The globbing patterns for this function are a superset of those from fnmatch(3)
```
markus2330: I think you want to point users to glob(7). If fnmatch is used, is not important.

src/libs/globbing/globbing.c
```diff
@@ -0,0 +1,212 @@
...
+ * used with the FNM_PATHNAME flag:
+ * <ul>
+ * 	<li> '*' matches any series of characters other than '/'</li>
```
markus2330: Please either add character classes here or only describe the Elektra extensions.

src/libs/globbing/globbing.c
```diff
@@ -0,0 +1,212 @@
...
+	int rc = fnmatch (fnmPattern, name, FNM_PATHNAME);
+	elektraFree (fnmPattern);
+
+	if (rc == FNM_NOMATCH)
+	{
+		free (name);
+		return ELEKTRA_GLOB_NOMATCH;
+	}
+
+	const char * ptr = pattern;
```
markus2330: put to a separate function?

src/libs/globbing/globbing.c
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * @param result the KeySet to which the matching keys should be appended
+ * @param input the KeySet whose keys should be filtered
+ * @param pattern the globbing pattern used
+ * @return the number of Keys appended to result or -1,
+ * 	   if @p result, @p input or @p pattern are NULL
+ *
+ * @see keyGlob(), for explanation of globbing pattern
+ */
+int ksGlob (KeySet * result, KeySet * input, const char * pattern)
```
markus2330: Return the result? (Like in ksCut?)

src/libs/globbing/globbing.c
markus2330: Or using _* would be another option. I am also not completely against **. It is, however, already used in zsh. So if we use **, it should have the same meaning.

src/libs/globbing/globbing.c
kodebach: Yes, this is the basically the same as in libease, but I wasn't sure whether a dependency on libease just for this functionality was a good idea. If the libglobbing depends on libease, then I don't see the advantage in having a separate libglobbing at all.

Personally I think that this function should be part of the core, because elektraWriteArrayNumber` is also declared in `kdbhelper.h` and implemented in `keyset.c`

src/libs/globbing/globbing.c
kodebach: Yes you are right we should use `__` or possibly just `...`, because our implementation does not support using this syntax anywhere but at the end of the pattern. To support this behaviour inside the pattern, we would need a non-standard implementation of `fnmatch(3)`. (It might be possible with the GNU extension `FNM_EXTMATCH`)

> Why using '/' instead of /?
If you wanted to know why I used `&#47;`, that is because my IDE (or more specifically clang-tidy) complains about the use of `/*` inside of block comments.

src/libs/globbing/globbing.c
kodebach: It would probably best to just use `FNM_NOESCAPE` and use character classes (which I will add to the documentation) for literal matching.

src/libs/globbing/globbing.c
kodebach: This function is basically the same as `elektraKsFilter` with the filter function (in this case `keyGlob`) inlined.

We should either allocate a new KeySet with `ksNew` and return that or leave it as is, because in the current implementation you could pass a non-empty KeySet as `result` and get a number of appended Keys as the return value. Appending to a non-empty KeySet might be useful, if the result of several globbing operations are needed in a single KeySet, in which case this implementation would be more efficient than allocating multiple KeySets and merging them.

kodebach:
> Please finish up the PR (incl. docu) before requesting reviews

Sorry I didn't expect you to be this quick with your review...

Everything (including your requests from the review) should be done now, except for the stuff regarding dependencies to `libease` and the possibility of moving the graph code into a separate library.

src/libs/globbing/globbing.c
markus2330: > If the libglobbing depends on libease, then I don't see the advantage in having a separate libglobbing at all.

libglobbing uses POSIX, libease not.

> Personally I think that this function should be part of the core

It quite clearly should not be. The core is only for functionality that directly depends on internal structures. It should be as minimal as possible.

> elektraWriteArrayNumber

You are right, elektraWriteArrayNumber should not be in the core. I hope d95edabb20264aef6db35085ad329d43b70e9735 makes it more clear.

markus2330:
Moving code around can be done in a later PR. Can you please make the build server happy and set ready to merge?

src/libs/globbing/globbing.c
kodebach: I know, I just thought having `\` for escaping is just redundant. Or am I missing something?

markus2330: You should at least say that the Elektra extensions follow the same escaping rules as glob(7).

src/libs/globbing/globbing.c
markus2330: Why not using elektraKsFilter? Then the dep to libease should be justified.

Did you measure the performance difference? It would be interesting.

kodebach:
@markus2330 Could you please check the build-server logs and tell me, whether I need to fix something else? I am not sure why some of the builds still fail (seems more like they crashed somehow).

sanssecours:
> Could you please check the build-server logs and tell me, whether I need to fix something else?

It seems you [added a new function to the API](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2238/11/pipeline/221#step-289-log-494). Thank you for that 💖. Please take a look at [`doc/todo/API`](https://github.com/ElektraInitiative/libelektra/blob/master/doc/todo/API). After you checked that you took care of each item mentioned there, please add a new entry for `elektraArrayValidateBaseNameString ` to [`icheck.suppression`](https://github.com/ElektraInitiative/libelektra/blob/master/tests/icheck.suppression).

markus2330:
@sanssecours Thank you for helping out!

Yes, the logs are difficult to read. It is not easy to find out which jobs were aborted and which of the jobs caused the abort. As @sanssecours already said, the API checks failed (it is the last job, called icheck).

kodebach:
I fixed as many errors as I could, but `testshell_markdown_reference` is still failing with ASAN and I am not sure why. It seems the leak comes from inside `elektraJniOpen`, which doesn't really make sense to me.

Also the MinGW build fails because `fnmatch.h` is missing, which is weird because MinGW should have that file.

markus2330:
Where did you see the ASAN failure? In the [latest build I could not see any](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2238/16/pipeline).

Did you see http://mingw.5.n7.nabble.com/can-t-find-fnmatch-h-td2552.html? Maybe only the prototype and linking against the correct lib is needed.

kodebach:
> Where did you see the ASAN failure? In the latest build I could not see any.

https://travis-ci.org/ElektraInitiative/libelektra/jobs/448816201

kodebach:
I did some more research and it seems MinGW actually does not provide fnmatch, only MSYS does. The question is why do none of the plugins using fnmatch produce errors? Are they just not compiled if the header is not found?

markus2330:
We exclude most of the plugins for mingw builds. So the way forward will be to also exclude the reference plugin from the mingw builds.

sanssecours:
> Do you have an idea?

According to the [output of the Travis build job `🐧 GCC ASAN`](https://travis-ci.org/ElektraInitiative/libelektra/jobs/450516235#L2330-L2333):

> STDERR: The command kdb mount terminated unsuccessfully with the info:
> No plugin that provides sync could be found

, `testshell_markdown_reference` requires the Sync plugin. I added `sync` and `dump` as required plugins for the Markdown Shell Recorder test in commit 10da6db. As far as I can tell, this update [should fix the problem](https://travis-ci.org/sanssecours/elektra/builds/450601104), since afterwards the Reference plugin will not be included in the build jobs `🐧 GCC ASAN` and `🍏 Clang ASAN` anymore.

kodebach:
@markus2330 From my side, this PR is now fully ready to be merged.

markus2330:
Thank you! Can you please open an issue that globbing with escaped slashes is not supported?

PR #2241, opened 2018-09-17
Injection plugins

Piankero:
I want to propose a plugin which injects errors in configurations. This is my first idea and some sort of architecture document.

If it is accepted I will also add all information to Metadata.ini and error/specification (so many new metadatas are very timeconsuming).

### DIFF
--- a/src/plugins/injection/README.md
+++ b/src/plugins/injection/README.md
@@ -325,9 +325,13 @@ These errors are set with `inject/limit/min` and `inject/limit/max` and one of b
         contrast = 500
         ```
         
-## Logging of changes
+## Tracing of changes
 
 To see which changes were applied to the whole KeySet, additional metadata will be written to the rootkey.
+If the plugin applies a change to the keyset, it writes it to the next available free array spot under
+`inject/log/#[number]`. So if you have done 10 injections into a KeySet for example, there should be entries
+form `inject/log/#0` to `inject/log/#9`.
+
 Assume you mounted a configuration to `user/my/config` and attached some metakeys to settings.
 All injections which were done are written in array notation to `inject/log/#[number]` and are continuous.
 So for example:
--- a/src/plugins/injection/README.md
+++ b/src/plugins/injection/README.md
@@ -329,22 +329,22 @@ These errors are set with `inject/limit/min` and `inject/limit/max` and one of b
 
 To see which changes were applied to the whole KeySet, additional metadata will be written to the rootkey.
 If the plugin applies a change to the keyset, it writes it to the next available free array spot under
-`inject/log/#[number]`. So if you have done 10 injections into a KeySet for example, there should be entries
-form `inject/log/#0` to `inject/log/#9`.
+`inject/change/#[number]`. So if you have done 10 injections into a KeySet for example, there should be entries
+form `inject/change/#0` to `inject/change/#9`.
 
 Assume you mounted a configuration to `user/my/config` and attached some metakeys to settings.
-All injections which were done are written in array notation to `inject/log/#[number]` and are continuous.
+All injections which were done are written in array notation to `inject/change/#[number]` and are continuous.
 So for example:
 ```sh
 kdb lsmeta user/my/config
-#> inject/log/#0
-#> inject/log/#1
+#> inject/change/#0
+#> inject/change/#1
 ...
 
-kdb getmeta user/my/config inject/log/#0 
+kdb getmeta user/my/config inject/change/#0
 #> "Removed section user/my/config/some/section"
 
-kdb getmeta user/my/config inject/log/#1 
+kdb getmeta user/my/config inject/change/#1
 #> "Changed value "true" to "truGe" on user/my/config/boolean/value"
 ```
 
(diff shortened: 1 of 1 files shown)


src/plugins/injection/README.md
```diff
@@ -0,0 +1,325 @@
+- infos = Information about the injection plugin is below
+- infos/author = Markus Raab <elektra@libelektra.org>
```
markus2330: wrong author.

src/plugins/injection/README.md
```diff
@@ -0,0 +1,325 @@
...
+- infos/licence = BSD
+- infos/needs =
+- infos/provides = inject
+- infos/recommends =
+- infos/placements = presetstorage
+- infos/status = maintained nodep
+- infos/metadata = inject/randSeed inject/structure/sectionRemove inject/structure/sectionDuplicate inject/structure/sectionReallocate inject/semantic inject/resource inject/typo/transposition inject/typo/insertion inject/typo/caseToggle inject/typo/deletion inject/typo/changechar inject/typo/space inject/domain inject/limit/min inject/limit/max
```
markus2330: please avoid camel case, we always use / for separation. So it would be inject/rand/seed, ...

src/plugins/injection/README.md
```diff
@@ -0,0 +1,325 @@
...
+## Introduction
+
+This plugin is used for purposefully alter single or multiple configuration settings in beneath a keyset.
+The plugin searches for metadata 'markers' on every key and depending on the injection, requires additional metadata.
+How the specific injection is carried out can either be random or, if desired, pseudorandom to guarantee reproducability.
+
+## Usage
+
+There up to 6 error types currently with injection types for each.
+1. Structural errors (e.g. missing sections, parameters in wrong sections, omission necessary parameters in section)
```
markus2330: Afaik some markdown renderer need a newline in between so that this enumeration is recognized.

src/plugins/injection/README.md
```diff
@@ -0,0 +1,325 @@
...
+        ```
+        &downarrow;&downarrow;&downarrow;
+        ```
+        a:
+            b2:
+                c2:
+        ```
+
+* Section Reallocator
+    * `inject/structure/sectionReallocate`
```
markus2330: Isn't this the same as rename?

src/plugins/injection/README.md
```diff
@@ -0,0 +1,325 @@
...
+    
+        ```
+        logger = log4j
+        ```
+        &downarrow;&downarrow;&downarrow;
+        ```
+        [logger]
+        inject/semantic/#0 = pyLog
+        inject/semantic/#1 = log4net
+        inject/semantic/#2 = Log4js
```
markus2330: Isn't this a capitalization error?

src/plugins/injection/README.md
```diff
@@ -323,3 +324,23 @@ These errors are set with `inject/limit/min` and `inject/limit/max` and one of b
         ```
         contrast = 500
         ```
+        
+## Logging of changes
+
+To see which changes were applied to the whole KeySet, additional metadata will be written to the rootkey.
```
markus2330: Description is very short, it is not yet clear how this logging should work. In METADATA.ini logging is already proposed, maybe you can extend this work so that logging is not specific to the inject plugin?

Piankero: Maybe `log` is the wrong wording as I just want to backtrace injections which were done (i.e. typo error on setting xy, structure error on key zx, etc.)

markus2330: > I do not see anything concerning logging in the metadata.ini

Sorry, it was removed in 06e10ca90ea129cf0712d8fca40f8f1159616174 and did not really contain useful information.

> Maybe log is the wrong wording as I just want to backtrace injections which were done

Yes, better to change the wording then.

markus2330: And why not add the changes directly to the keys where the changes occurred?

Piankero: > And why not add the changes directly to the keys where the changes occurred?

Because some injections simply remove a whole configuration/ section. It is also much simpler to use when all changes can be found on one key's metadata than searching for all keys if they have an injection on them imo

markus2330:
Please only set "ready to merge" if the PR compiles.

markus2330:
You just replaced log -> change, what should I review here? Please request reviews more carefully.

The logging strategy should have already been a part of the architecture, so it is now quite complicated to give input about the big picture.

@mpranj suggested that we should have information that is passed from plugins to plugins in a separate keyset, similar to the plugin handle and plugin config. Your "logs/changes" would be much better placed there.

Piankero:
> You just replaced log -> change, what should I review here? Please request reviews more carefully.

The only "task" you explicitly gave me is to change the wording. I did that and now want to see if the whole concept is finished. How should I tell you to take a whole look on the concept other than labeling the PR?

> @mpranj suggested that we should have information that is passed from plugins to plugins in a separate keyset, similar to the plugin handle and plugin config. Your "logs/changes" would be much better placed there.

Where did he suggest that?

mpranj:
@Piankero we discussed it in the meeting last week, but I have not written anything about it in doc/decisions yet.

I needed some means of communication between plugins. I proposed to use a KeySet for my scenario and @markus2330 suggested that it can be useful for other plugins too. Plugins would have a handle to a "global" KeySet with the sole purpose of communication between plugins.

PR #2245, opened 2018-09-21
Travis: Fix Compilation With Clang 7 

sanssecours:
# Note

I included commit 1fdb15e, since otherwise the build job `🍏 Clang ASAN` will probably always hit the [timeout limit of public repositories](https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts).


markus2330:
Thank you for taking care of this! You do not need to include 1fdb15eb099192e1690639d598276bb230f6ee1c anymore.

PR #2250, opened 2018-09-25
YAMBi: Execute Unit Tests Again

sanssecours:


markus2330:
Why do you need the main method at all?

sanssecours:
> Why do you need the main method at all?

As far as I can tell, `srcdir_file` does not work properly, if we do not call `init (argc, argv)` beforehand. One function that has access to the command line arguments `argc` and `argv` is `main`. That is the reason why I added a main method that calls `init`.

markus2330:
I see. Maybe we can provide the main (which calls init and also does cleanup afterwards) in a way so that all our google tests automatically use it?

markus2330:
Thank you! I created  #2261.

PR #2251, opened 2018-09-25
YAwn: Initial Version of Storage Plugin

sanssecours:
# Checklist

- [x] I added unit tests.
- [x] I ran all tests locally and every test (except for `testmod_network`) succeeded.
- [x] The pull request also contains a documentation for the plugin.
- [x] The PR contains code comments, logging, and assertions.
- [x] The metadata of the plugin should be up to date.


src/plugins/yawn/listener.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+
+/**
+ * @brief This function converts a given number to an array base name.
+ *
+ * @param index This number specifies the index of the array entry.
+ *
+ * @return A string representing the given indices as Elektra array name.
+ */
+string indexToArrayBaseName (uintmax_t const index)
```
markus2330: What about adding elektraArraySetName (Key, size_t) to libease?

src/plugins/yawn/README.md
```diff
@@ -0,0 +1,74 @@
...
+- infos/placements = getstorage
+- infos/status = maintained unittest preview memleak experimental unfinished nodoc concept discouraged
+- infos/metadata =
+- infos/description = This storage plugin use a Early parser to read YAML files
+
+# YAwn
+
+## Introduction
+
+This plugin uses [YAEP](https://github.com/vnmakarov/yaep) to  parse the [YAML](http://yaml.org) serialization format.
```
markus2330: I would simply say "to  parse [YAML](http://yaml.org)." We did not define what a "serialization format" is and how this differs from a configuration file format."

src/plugins/yawn/lexer.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+#include "input.hpp"
+#include "token.hpp"
+
+// -- Class --------------------------------------------------------------------------------------------------------------------------------
+
+namespace yawn
+{
+
+/** This class implements a basic lexer for the serialization format YAML. */
```
markus2330: again (to parse YAML)

src/plugins/yawn/objstack.h
```diff
@@ -0,0 +1,525 @@
...
+   General Public License as published by the Free Software
+   Foundation; either version 2, or (at your option) any later
+   version.
+
+   This software is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Library General Public License for more details.
+
+   You should have received a copy of the GNU Library General Public
```
markus2330: I am afraid we cannot redistribute this here.

src/plugins/yawn/token.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace yawn
+{
+
+/** This class represents a token emitted by a lexer. */
+class Token
+{
+
+	/** This variable stores the location of this token in the scanned text. */
+	Location _location;
```
markus2330: The C++ standard disallows identifier to start with _. If you used this everywhere now, you can keep it.

src/plugins/yawn/listener.cpp
sanssecours: Sounds like a good idea.

src/plugins/yawn/objstack.h
sanssecours: Where should I put this file then?

src/plugins/yawn/token.hpp
sanssecours: > The C++ standard disallows identifier to start with _.

Isn’t that [only true for the global namespace](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier)?

src/plugins/yawn/objstack.h
markus2330: Isn't this header file present in some package?

If not please ask the author if it is okay to relicense it as BSD.

src/plugins/yawn/token.hpp
markus2330: If you only start with lowercase you should be fine.

src/plugins/yawn/objstack.h
markus2330: Downloading the file causes other problems (in the build server...).

Why isn't this file present in a yaep dev package?

Did you get a response from the author?

sanssecours: > Downloading the file causes other problems (in the build server...).

I do not think that is a problem. If the CMake code is unable to download the file, then it will just remove the plugin.

> Why isn't this file present in a yaep dev package?

I assume, because you can use another function that offers the same functionality (such as `malloc`) instead. That is only an assumption though.

> Did you get a response from the author?

I did not ask him about relicensing the source code.

markus2330:
Do you got an answer if we are not allowed to redistribute it as BSD?

Can you please open a bug report so that they correctly install this header file?

sanssecours:
> Do you got an answer if we are not allowed to redistribute it as BSD?

See: https://github.com/ElektraInitiative/libelektra/pull/2251#discussion_r221878584

> Can you please open a bug report so that they correctly install this header file?

I modified my clone of YAEP accordingly, and opened an pull request for the changes [here](https://github.com/vnmakarov/yaep/pull/15).

src/plugins/yawn/objstack.h
markus2330: > I do not think that is a problem. If the CMake code is unable to download the file, then it will just remove the plugin.

That assumes that the user does not care if the plugin gets added or not.

>  I did not ask him about relicensing the source code.

Please do, I think this is by far the easiest option that does not cause any problems. Iirc no author declined such a request up to now.

markus2330:
> I modified my clone of YAEP accordingly, and opened an pull request for the changes here.

Thank you, that is a very nice long-term solution. What is the situation of YAEP packages? Does any distribution provides one? Is alpine now the only Docker container which builds this plugin on the build server?

src/plugins/yawn/objstack.h
sanssecours: The latest version of the code uses `malloc` as allocation function for the parsing code. While I assume that is not an ideal solution, the update removes the dependency on `objstack.h`.

markus2330: Why is it not an ideal situation? Sounds okay to me?

If a user wants to avoid malloc, he/she will also wants to avoid it at all other places of Elektra, so elektraMalloc need to be changed anyway.

sanssecours:
> What is the situation of YAEP packages? Does any distribution provides one? 

As far as I can tell, [there are no prepackaged versions of YAEP](https://repology.org/metapackage/yaep/versions).

> Is alpine now the only Docker container which builds this plugin on the build server?

Correct. I tried to install the package in one of the Debian images. However, if I add `bison` (a prerequisite for YAEP) to the package list of, then `cabal` is unable to install the prerequisites for the type system.

src/plugins/yawn/objstack.h
sanssecours: > Why is it not an ideal situation? Sounds okay to me?

I assume the author of YAEP would not have written memory management code, if a simple `malloc` would provide the same performance. Anyway, I really do not care if we use `malloc` or something different, as long as the memory management code is reasonable simple to use.

> If a user wants to avoid malloc, he/she will also wants to avoid it at all other places of Elektra, so elektraMalloc need to be changed anyway.

I replaced `malloc` with `elektraMalloc` in the latest commit.

markus2330:
> Correct. I tried to install the package in one of the Debian images. However, if I add bison (a prerequisite for YAEP) to the package list of, then cabal is unable to install the prerequisites for the type system.

I think we should split up our Debian Image and remove the parts we do not need for building the Debian packages (in particular the huge Haskell+Java+GI is not needed for the Debian package.). This would avoid such problems.

@sanssecours @ingwinlu What do you think?

sanssecours:
> What do you think?

I think I am not the right person to ask, since I basically do not know anything about Docker.

markus2330:
Thank you, very good job!

Issue #1789, opened 2018-01-23
reduce error information for common errors


markus2330:

Piankero:
What exactly should I do here? You referenced this issue in #2333 but with a checkmark on the task

markus2330:
This is actually already implemented with the `-d` and `-v` options.

@kodebach once said that it would be much better if we only have `-v`. For me open is a tutorial/docu/guide about how tools should provide `-d` and `-v`. (When to output which parts of warnings/errors).

Piankero:
This issue is unrelated to visible error messages.
As discussed in the Elektra meeting I may unassign myself.

markus2330:
This is very much related to visible error messages but it is already implemented (by you)

PR #2253, opened 2018-09-26
Port plugin extension

Piankero:
## Basics

## Checklist

- [x] I added unit tests

## Review


src/plugins/network/network.c
```diff
@@ -50,6 +50,70 @@ int elektraNetworkAddrInfo (Key * toCheck)
...
+int elektraPortInfo(Key * toCheck, Key * parentKey) {
+	const Key * meta = keyGetMeta (toCheck, "check/port");
+	const Key * listenMeta = keyGetMeta (toCheck, "check/port/listen");
+	if (!meta && !listenMeta) return 0; /* No check to do */
+	char *endptr = NULL;
+	long portNumber = strtol(keyString(toCheck), &endptr, 10);
+	int portNumberNetworkByteOrder;
+
+	if (*endptr == '\0') {
+		if (portNumber < 0 || portNumber > 65535 || *endptr != 0) {
```
sanssecours: I think you can remove the check `*endptr != 0` here. As far as I can tell, `*endptr` should always be `0` in this part of the code, since the if-statement one line above:

```c
if (*endptr == '\0')
```

makes sure that this is the case.

Piankero:
Ty so much for the reformatting. I tried to configure clion to do IT correctly but some options are not available which you require

sanssecours:
> Ty so much for the reformatting. I tried to configure clion to do IT correctly but some options are not available which you require

I would recommend

1. to either run `clang-format` directly on the file you are updating, or
2. to use the script `scripts/reformat-source` before you commit your changes

. Another option is to [run `clang-format` as external tool in CLion](https://stackoverflow.com/questions/34648255/using-clang-format-in-clion). There is also a [CLion plugin](https://plugins.jetbrains.com/plugin/8396-clangformatij) that should be able to reformat code using Clang-Format.

Piankero:
I don't know whats wrong with the build server but it looks like my build is ok. Please someone take a review of it so we can merge it!

markus2330:
@ingwinlu do you know what is wrong with the build server here?

ingwinlu:
looks like the connection to the node was lost during build. simply restart the build next time either via jenkins or via the magic phrase and see if this resolves it.

src/plugins/network/testmod_network.c
```diff
@@ -27,8 +32,53 @@ int main (int argc, char ** argv)
...
+	testPortAny ("https", 1);
+
+	testPortAny ("65536", -1);
+	testPortAny ("-1", -1);
+	testPortAny ("22d", -1);
+	testPortAny ("myInvalidServiceName", -1);
+
+	// These tests aren't portable I guess
+	// testListenPortAny ("http", -1);
+	testListenPortAny ("8080", 1);
```
markus2330: 8080 is quite likely to be in use. There are some reserved ports, maybe let us use?

these https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

src/plugins/network/network.c
```diff
@@ -50,6 +52,75 @@ int elektraNetworkAddrInfo (Key * toCheck)
...
+		if (service == NULL)
+		{
+			ELEKTRA_SET_ERRORF (202, parentKey, "Could not find service with name %s on key %s", keyString (toCheck),
+					    keyName (toCheck));
+			return -1;
+		}
+		portNumberNetworkByteOrder = service->s_port;
+	}
+
+	// Check if we can connect to it
```
markus2330: comment?

src/plugins/network/network.c
```diff
@@ -50,6 +52,75 @@ int elektraNetworkAddrInfo (Key * toCheck)
...
+	sockfd = socket (AF_INET, SOCK_STREAM, 0);
+	if (sockfd < 0)
+	{
+		ELEKTRA_SET_ERROR (204, parentKey, "Could not open a socket");
+	}
+
+	server = gethostbyname (hostname);
+	if (server == NULL)
+	{
+		ELEKTRA_SET_ERRORF (204, parentKey, "Could not connect to %s: No such host", hostname);
```
markus2330: Is "No such host" the actual reason? The reason is in errno, isn't it?

Btw. we should also handle the TRY_AGAIN case somehow.

src/plugins/network/network.c
```diff
@@ -50,6 +52,75 @@ int elektraNetworkAddrInfo (Key * toCheck)
...
+		return -1;
+	}
+
+	bzero ((char *) &serv_addr, sizeof (serv_addr));
+	serv_addr.sin_family = AF_INET;
+	bcopy ((char *) server->h_addr, (char *) &serv_addr.sin_addr.s_addr, server->h_length);
+
+	serv_addr.sin_port = (in_port_t) portNumberNetworkByteOrder;
+	if (connect (sockfd, (struct sockaddr *) &serv_addr, sizeof (serv_addr)) == 0)
```
markus2330: Is there no other way? If you connect to it, you might cause errors in later attempts to connect ("port still in use" errors)

src/error/specification
```diff
@@ -1258,3 +1258,27 @@ severity:warning
...
+module:network
+
+number:202
+description:Could not find service name for port declaration
+severity:error
+ingroup:plugin
+module:network
+
+number:203
+description:Could not use port as it is already in use
```
markus2330: Do we really need so many errors? Aren't the descriptions and the reasons somewhat overlapping?

src/plugins/network/testmod_network.c
Piankero: Do you have a suggestion/preference for which one to use?

src/plugins/network/network.c
Piankero: I read an article about this: [link](https://stackoverflow.com/questions/10294515/how-do-i-find-in-c-that-a-port-is-free-to-use)
Is there a notable difference between bind and connect?

src/error/specification
Piankero: Should I just do an "Port error" and reference to the specific error in the reason metadata?

markus2330: You should generally redesign the errors to be more user friendly. Think about what applications and users care about.

Only if there are situations where applications want to distinguish one error from the other, the error number should be different.

markus2330: Alternatively we could also introduce some inheritance for error messages and something like `kdbContainsError (KDB *key, int whicherror)` (or similar for the higher-level API).

We can discuss this tomorrow if you already want to work on that topic. But maybe you should finish the tasks you started already before.

src/plugins/network/network.c
markus2330: Parsing /proc/net/tcp is also possible. Comparing different ways for how to do validations is part of your work.

src/plugins/network/testmod_network.c
Piankero: I tried nearly all ports from 1-100 and only http does work on my system (but not on the build system) when I try to test for unavaliable ports as root. So what should I do now?

You can give me ports which are reserved on the build server for sure or I cannot test it at all. (a -1 case at least)

src/plugins/network/network.c
Piankero: The code which follows basically does this. I will just move the comment a few lines below so it is more clear

src/plugins/network/testmod_network.c
markus2330: Did you consider that Ports below 1024 can only be used by root?

src/plugins/network/network.c
Piankero: After doing some research there are 2 possibilities:

1. Parse /proc/net/tcp as you suggested
2. Call bind/ connect and see if the port is in use

Doing parsing is very slow and as [this comment](https://stackoverflow.com/a/10311153) suggests is no superior anyway compared to bind. As a result I will stick to the bind approach.

src/error/specification
Piankero: I refactored the errors and are only using a single one which I did not see anywhere else in the specification file. I also reused another error number from the range plugin

src/plugins/network/network.c
Piankero: Fixed errno.

Concerning TRY_AGAIN: Yes we could try to implement a retrying approach but isn't it very unlikely that this happens? Adding this functionality to the code would clearly degrade the readability and maintainability.

src/plugins/network/testmod_network.c
Piankero: Yes I did. I was able to use all ports and `netstat -tulpn` also only showed that port 80 was in use (of all ports below 1000). I am running the tests as root so I can try to bind to the port.

It seems like this is not testable because every build system/ (maybe even distro of linux) uses different ports or uses them just in a very short window (like port 7 `echo`) that it is not deterministic if a port is in use or not.

One possibility is to take the `ssh` port from the build server as it will most likely be in use all the time?

Piankero:
@ingwinlu The build again timed out. Can you restart the job?

@markus2330 Do I have a jenkins account or permission to do that myself?

Piankero:
And another timeout in a test it looks like ....

markus2330:
Please read doc/BUILDSERVER.md which should make clear how to use the buildserver.

src/plugins/network/network.c
Piankero: After digging another time into the internet I was not able to find any other possibility other than trying to connect or parsing proc/net/tcp. Since the first possibility is superior (same disadvantages but faster) I will keep the code as it is

src/error/specification
Piankero: Error messages will be dealt with another issue https://github.com/ElektraInitiative/libelektra/issues/1789

Piankero:
@markus2330 
I wrote text to all of your desired review changes. My personal intuition would say that the PR is ready for merging (if the buildserver does not fail).
I tested everything locally a second time and it works as intended.

If you do not have any further inquiries I would suggest we incorporate the new plugin!
Also @sanssecours could probably take a short look about my code as I am no C coding expert such as u both are?

src/plugins/network/network.c
```diff
@@ -50,6 +52,93 @@ int elektraNetworkAddrInfo (Key * toCheck)
...
+	if (server == NULL)
+	{
+		if (errno == HOST_NOT_FOUND)
+		{
+			ELEKTRA_SET_ERRORF (205, parentKey, "Could not connect to %s: No such host", hostname);
+			return -1;
+		}
+		else
+		{
+			ELEKTRA_SET_ERRORF (205, parentKey, "There was an error when trying to connect to host %s . errno: %s", hostname,
```
sanssecours: Since the error message already uses the human readable form of `errno`, I think it would make sense to not mention the term `errno` at all. How about

```c
"There was an error trying to connect to host %s: %s"
```

instead of

```c
"There was an error when trying to connect to host %s . errno: %s"
```

?

Piankero:
Thank you very much for the feedback!

> Since there was a [release](https://github.com/ElektraInitiative/libelektra/releases/tag/0.8.25) since you opened this pull request: Can you please rebase your changes on `master`?

I am not quite used to such a procedure, usually I branch of and merge back. I also merged again into the master but how exactly does your rebase approach work?

sanssecours:
> I also merged again into the master but how exactly does your rebase approach work?

You can use the steps below to rebase your work on the `master` branch.

1. Switch to the branch you want to rebase: `git checkout port-plugin-extension`
2. Call `git rebase master`
3. While you have a merge conflict
    1. Fix the merge conflict
    2. Call `git rebase --continue`

After you fixed all merge conflicts you should have a branch that applies cleanly on `master`.

Piankero:
> > I also merged again into the master but how exactly does your rebase approach work?
> 
> You can use the steps below to rebase your work on the `master` branch.
> 
> 1. Switch to the branch you want to rebase: `git checkout port-plugin-extension`
> 2. Call `git rebase master`
> 3. While you have a merge conflict
>    
>    1. Fix the merge conflict
>    2. Call `git rebase --continue`
> 
> After you fixed all merge conflicts you should have a branch that applies cleanly on `master`.

But as of now I already have done a merge. Do I need to revert?

sanssecours:
> But as of now I already have done a merge. Do I need to revert?

Nope. Just rebase on `master` and force push your updated branch afterwards.

mpranj:
Looks like you rebased on an old version of master. You need to keep it [up to date](https://help.github.com/articles/syncing-a-fork/), then you won't have merge conflicts here.

Try to make sure that syncing master does not create extra merge commits. It might happen if your git config has disabled fast-forward or you have some unique commits on your master branch.

Piankero:
> Looks like you rebased on an old version of master. You need to keep it [up to date](https://help.github.com/articles/syncing-a-fork/), then you won't have merge conflicts here.
> 
> Try to make sure that syncing master does not create extra merge commits. It might happen if your git config has disabled fast-forward or you have some unique commits on your master branch.

Yea, unfortunately I have seen that a bit too late, I will open a new PR with a single commit for the changes, based on the correct master. This github multi-remote workflow is a bit uncommon to me

mpranj:
I don't think you have to stash everything into a single commit. You can use `git rebase -i`, the interactive mode, to reword some commits messages possibly containing swearwords, or to stash some specific commits together.

PR #2257, opened 2018-09-26
Path access extension

Piankero:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,14 @@
 
 #ignore backups
 *~
+cmake/**
+.idea/**
+CMakeCache.txt
+**/CMakeFiles/**
+DartConfiguration.tcl
+test-symbols.map
+tests/**
+
 
 #ignore debian patches
 /.pc/
--- a/doc/METADATA.ini
+++ b/doc/METADATA.ini
@@ -623,6 +623,30 @@ description= Specifies if multiple, but different, strings are allowed
 	is the separator.
 example= _
 
+[check/permission/types]
+status= idea
+usedby/plugin= path
+type= enum
+    r
+	w
+	x
+	rw
+	rx
+	wx
+	rwx
+description= A check to see if a certain user has correct permissions on a file. Permissions are
+	either r (read) w (write) x (execute) and any combination of them.
+	 The respective user which must have these permissions is declared in check/permission/user.
+	If the user is missing, the current user is taken. (by using this setting, implicitly check/permission/user
+	is also set but remains empty)
+example = rwx
+
+[check/permission/user]
+status= idea
+usedby/plugin= path
+type= string
+description= Used be check/permission to see if the given user has the correct permissions. If empty
+	it takes the current user.
 
 [check/calculate]
 status= implemented
--- a/src/error/specification
+++ b/src/error/specification
@@ -341,7 +341,7 @@ module:path
 
 number:57
 description:could not stat file
-severity:warning
+severity:error
 ingroup:plugin
 module:path
 
--- a/src/error/specification
+++ b/src/error/specification
@@ -1258,3 +1258,21 @@ severity:warning
 ingroup:plugin
 module:gpgme
 macro:GPGME_INVALID_RECIPIENT
+
+number:201
+description:The user does not exist.
+severity:error
+ingroup:plugin
+module:path
+
+number:202
+description:Could not change to user. You might want to try as root (use sudo !!).
+severity:error
+ingroup:plugin
+module:path
+
+number:203
+description:Detected incorrect permissions for file/directory
+severity:error
+ingroup:plugin
+module:path
--- a/src/plugins/path/README.md
+++ b/src/plugins/path/README.md
@@ -26,9 +26,14 @@ is allowed to occur for both device and mountpoint. When checking for
 relative files, it is not enough to look at the first character if it is
 a `/`, because remote file systems and some special names are valid, too.
 
-If `check/permission = [permission], [user]` is also present it will check for the correct permissions
-of the file/directory. `check/permission = rw` and `check/permission/user = tomcat` for example will check if the user `tomcat` has read and
-write access to the path which was set in `check/path`.
+If `check/permission/types = [permission]` is also present it will check for the correct permissions
+of the file/directory. Optionally, you can also add `check/permission/user = [user]"` which then checks the permissions
+for the given user. If `check/permission/user` is not used, the current user will be taken.
+
+ `check/permission/types = rw` and `check/permission/user = tomcat` for example will check if the user
+`tomcat` has read and write access to the path which was set in `check/path`. Please note that the file has to exist already
+and it is not checked if the user has the right to create a file with a specified name
+in the directory.
 
  Permissions available:
  - `r`: **R**ead
@@ -52,7 +57,8 @@ kdb set /tests/path/value /var/log/application-file.log
 sudo kdb setmeta /tests/path/value check/path
 
 #This checks if the user has read and write permissions for the application-file.log file
-sudo kdb setmeta /tests/path/value check/permission "rw, tomcat"
+sudo kdb setmeta /tests/path/value check/permission/user "tomcat"
+sudo kdb setmeta /tests/path/value check/permission/types "rw"
 
 #Generate a file which is only accessable for root
 touch /var/log/application-file-restricted.log
--- a/src/plugins/path/README.md
+++ b/src/plugins/path/README.md
@@ -28,12 +28,12 @@ a `/`, because remote file systems and some special names are valid, too.
 
 If `check/permission/types = [permission]` is also present it will check for the correct permissions
 of the file/directory. Optionally, you can also add `check/permission/user = [user]"` which then checks the permissions
-for the given user. If `check/permission/user` is not used, the current user will be taken.
+for the given user. When calling `kdb set` on the actual key, you have to run as `root` user
+or the file permissions cannot be checked (you will receive an error message).
 
  `check/permission/types = rw` and `check/permission/user = tomcat` for example will check if the user
-`tomcat` has read and write access to the path which was set in `check/path`. Please note that the file has to exist already
-and it is not checked if the user has the right to create a file with a specified name
-in the directory.
+`tomcat` has read and write access to the path which was set for the key. Please note that the file has to exist already
+and it is not checked if the user has the right to create a file in the directory.
 
  Permissions available:
  - `r`: **R**ead
--- a/src/plugins/path/README.md
+++ b/src/plugins/path/README.md
@@ -5,24 +5,35 @@
 - infos/needs =
 - infos/provides = check
 - infos/status = maintained nodep libc nodoc
-- infos/metadata = check/path
-- infos/description = Checks if keys enriched with appropriate metadata contain valid paths as values
+- infos/metadata = check/path check/permission
+- infos/description = Checks if keys enriched with appropriate metadata contain valid paths as values as well
+as correct permissions
 
 ## Introduction
 
-This plugin checks whether the value of a key is a valid file system path. 
+This plugin checks whether the value of a key is a valid file system path and optionally if
+correct permissions are set for a certain user.
 
 ## Purpose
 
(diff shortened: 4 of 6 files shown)


Piankero:
@sanssecours Did you ever have that problem with the "implicit function declaration" despite being declared in the header file in `<unistd.h>`? Honestly stackoverflow does not provide any useful help

markus2330:
@ingwinlu can you maybe help out here?

.gitignore
```diff
@@ -5,6 +5,14 @@
 
 #ignore backups
 *~
+cmake/**
```
ingwinlu: that sounds like a terrible idea?

ingwinlu:
glibc automatically exports all function as long as no feature macros are set. So I have no idea why it does not at least build on glibc tests.

You can use something like https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/passwd/CMakeLists.txt#L8 to test if the function is available in the current environment and change the build process accordingly.

.gitignore
Piankero: wups, I did that in because I always got tons of unneccessary files from this folder to potentially commit and had to singlepick every file. I'll remove that

Piankero:
@sanssecours I included it and now the build process complains:
```
In file included from /home/jenkins/workspace/libelektra_PR-2257-5ERNEW5JQCBG6J7GJCVU5QCJS7S4QVA52KPOOT7HFF6AMH7HOMYQ/src/plugins/path/path.c:9:0:

/home/jenkins/workspace/libelektra_PR-2257-5ERNEW5JQCBG6J7GJCVU5QCJS7S4QVA52KPOOT7HFF6AMH7HOMYQ/src/plugins/path/path.h:12:0: error: "_GNU_SOURCE" redefined [-Werror]

 #define _GNU_SOURCE        // For euidaccess
```

ingwinlu:
`add_definitions (-D_GNU_SOURCE)` already adds it. you could add #ifndef or remove the line completely

sanssecours:
> @sanssecours I included it and now the build process complains:

If it the macro `_GNU_SOURCE` is already defined, then I guess that is not the problem. 

It certainly makes sense to add CMake code to detect if the system provides `euidaccess`, as Lukas already noted before. Does the plugin work on your computer?

src/plugins/path/CMakeLists.txt
```diff
@@ -1,5 +1,18 @@
+include (SafeCheckSymbolExists)
 include (LibAddMacros)
 
+set (plugin path)
+
+if (DEPENDENCY_PHASE)
+	add_definitions (-D_GNU_SOURCE)
+	safe_check_symbol_exists (euidaccess "unistd.h" TEST_EUIDACCESS)
+
+	if (!TEST_EUIDACCESS)
```
sanssecours: I think you should use `NOT` instead of `!` here. As far as I know, CMake does not support the `!` operator.

.gitignore
```diff
@@ -5,6 +5,13 @@
 
 #ignore backups
 *~
+.idea/**
+CMakeCache.txt
+**/CMakeFiles/**
+DartConfiguration.tcl
+test-symbols.map
+tests/**
```
sanssecours: Please remove (at least) the `tests/**` entry here.

ingwinlu:
well once you clean it up it should pass

reported during cmake:
```
-- Looking for euidaccess
-- Looking for euidaccess - found
-- Include Plugin path
```
no errors / warnings during compile on debian-stable.

src/plugins/path/path.c
```diff
@@ -65,6 +90,135 @@ static int validateKey (Key * key, Key * parentKey)
 	return 1;
 }
 
+// I assume the path exists and only validate permission
+static int validatePermission (Key * key, Key * parentKey)
+{
+	__uid_t currentUID = geteuid ();
```
ingwinlu: should be `uid_t`

src/plugins/path/path.c
```diff
@@ -65,6 +90,135 @@ static int validateKey (Key * key, Key * parentKey)
...
+	__gid_t currentGID = getegid ();
```
ingwinlu: should be `gid_t`

src/error/specification
```diff
@@ -1266,7 +1266,7 @@ ingroup:plugin
 module:path
 
 number:202
-description:Could not change to user. You might want to try as root (use sudo !!).
+description:Could not change to user.
```
markus2330: Better to not mix unrelated changes into PRs.

src/plugins/path/path.c
```diff
@@ -206,8 +207,15 @@ static int validatePermission (Key * key, Key * parentKey)
...
 
 	// Change back to initial effective IDs
-	seteuid (currentUID);
-	setegid (currentGID);
+	int euidResult = seteuid (currentUID);
+	int egidResult = setegid (currentGID);
+
+	if (euidResult != 0 || egidResult != 0)
+	{
+		ELEKTRA_SET_ERRORF (202, parentKey, "Could not change back to user before. This case should not happen", name,
```
markus2330: Please go into the users perspective. Users do not know that any user was changed and what they should do now (report a bug at issues.libelektra.org)

.gitignore
```diff
@@ -5,6 +5,12 @@
 
 #ignore backups
 *~
+.idea/**
+CMakeCache.txt
```
markus2330: Better to do out-of-source builds.

.gitignore
```diff
@@ -5,6 +5,12 @@
...
+**/CMakeFiles/**
+DartConfiguration.tcl
+test-symbols.map
```
markus2330: Where is this file from?

doc/METADATA.ini
```diff
@@ -623,6 +623,30 @@ description= Specifies if multiple, but different, strings are allowed
 	is the separator.
 example= _
 
+[check/permission/types]
+status= idea
+usedby/plugin= path
+type= enum
+    r
```
markus2330: formatting problem.

src/error/specification
```diff
@@ -341,7 +341,7 @@ module:path
 
 number:57
 description:could not stat file
-severity:warning
+severity:error
```
markus2330: Why is this an error?

src/plugins/path/README.md
```diff
@@ -5,28 +5,77 @@
...
 case. One situation, however, presents an error: Only an absolute path
-is allowed to occur for both device and mount point. When checking for
+is allowed to occur for both device and mountpoint. When checking for
 relative files, it is not enough to look at the first character if it is
 a `/`, because remote file systems and some special names are valid, too.
 
+If `check/permission/types = [permission]` is also present it will check for the correct permissions
+of the file/directory. Optionally, you can also add `check/permission/user = [user]"` which then checks the permissions
+for the given user. When calling `kdb set` on the actual key, you have to run as `root` user
+or the file permissions cannot be checked (you will receive an error message).
```
markus2330: And how to avoid being root?

src/plugins/path/README.md
```diff
@@ -5,28 +5,77 @@
...
+touch /var/log/application-file-restricted.log
+sudo chmod 700 /var/log/application-file-restricted.log
+sudo chown root:root /var/log/application-file-restricted.log
+
+#This should trigger the error
+kdb set /tests/path/value /var/log/application-file-restricted.log
+# ERROR:194
+# Reason:
+#    Expected: User tomcat has [read/write] permission for the given file /var/log/application-file-restricted.log.
+#    Actual: User tomcat has [] permission for var/log/application-file-restricted.log.
```
markus2330: No slash before /var?

src/plugins/path/path.c
```diff
@@ -9,9 +9,33 @@
...
+
 #include "kdbconfig.h"
+
 #endif
 
+// Used to delete last colon in the error message
+char * lastCharDel (char * name)
+{
+	int i = 0;
+	while (name[i] != '\0')
```
markus2330: Isn't this strlen? (elektraStrLen)

src/plugins/path/path.c
```diff
@@ -65,6 +90,136 @@ static int validateKey (Key * key, Key * parentKey)
...
+					    " Are you running kdb as root?\"",
+					    keyName (key));
+			return -1;
+		}
+	}
+
+	// The following code changes the egid if a group from a user matches the filegroup
+	// TODO: Whats a good default value for ngroups
+	// If the relevant group is ngroups+1 then it wont get recognized
+	int ngroups = 30;
```
markus2330: Avoid this magic constant.

src/error/specification
Piankero: Why is this change unrelated? I created the error 202 and in order to reduce the number of errors I changed the text so I can reuse it for another error which is very similar but not worth distinguishable for the program (root -> user & user -> root changes are both the same problem but the old text above only includes root -> user)

src/plugins/path/path.c
Piankero: I tried to find another way around but the I did not find a way around the "magic constant".
[getgrouplist](http://man7.org/linux/man-pages/man3/getgrouplist.3.html) requires a number in order to know how many groups to fetch.

Does anybody know a better way?

.gitignore
Piankero: Excluded it again, probably was autoadded by intellij

Piankero:
Since this plugin was now untouched for a few month, let me summarize all review comments and unclear results:

> And how to avoid being root?

This was a result of the requirement for being the `root` user when calling `kdb set` for this plugin.

*Reason:*
[setegid](https://linux.die.net/man/2/setegid) and [seteuid](http://man7.org/linux/man-pages/man2/seteuid.2.html) are the reason for this and have to be used in the code. I have to temporarily change to a specific user in the code to check for all relevant privileges via [euidaccess](http://man7.org/linux/man-pages/man3/euidaccess.3.html). Only `root` can change to other users due to possible privilege escalations.
One might ask if it is possible via [stat](https://linux.die.net/man/2/stat) but as [comment#1](https://stackoverflow.com/questions/11375738/how-can-i-check-to-see-if-user-has-execute-permissions#comment14990682_11375738) and [comment#2](https://stackoverflow.com/questions/11375738/how-can-i-check-to-see-if-user-has-execute-permissions#comment14990726_11375854) suggests, euidaccess is much more reliable way due to the complex nature of access rights in *nix systems and ACLs.

Further articles I have read through for checking file permissions (in order to avoid being root):
[#1](https://stackoverflow.com/questions/18726886/checking-file-access-permissions-for-specified-user-in-c-nix), [#2](https://stackoverflow.com/questions/7087958/file-stat-vs-access-to-check-permissions-on-a-directory), [#3](https://stackoverflow.com/questions/20126757/obtaining-permissions-for-a-file-in-c), [#4](https://stackoverflow.com/questions/3617508/determine-file-permissions-for-the-current-user) ... and much more

> Avoid this magic constant.

This was referred by @markus2330 [here](https://github.com/ElektraInitiative/libelektra/pull/2257#discussion-diff-225008362R154) because I set the number of groups to get for a user to 30.

*Reason:*
As [this](https://stackoverflow.com/questions/22104383/proper-way-to-get-groups-of-a-user-in-linux-using-c) article suggests (and various more), one should use [getgrouplist](http://man7.org/linux/man-pages/man3/getgrouplist.3.html) for this purpose. Unfortunately this function requires an argument for the max retrievable groups. Other suggestions are very welcome to counteract this problem.

> Why is this an error?

This refers to a [switch](https://github.com/ElektraInitiative/libelektra/pull/2257#discussion-diff-225008227R344) which changed a warning to an error. 

*Reason:*
This change affects [these](https://github.com/Piankero/libelektra/blob/251cecc702eb3fb58025d8e3d4ee795484e573f5/src/plugins/path/path.c#L64-L74) lines and since the path plugin is used to assert file existence, I changed it from a warning to an error. The added enhancements of the plugin (check for concrete permissions) requires the file/directory to be already existent and also relies on the fact that the file can be stat()'ed.

Piankero:
I would be very grateful if @sanssecours could take a short look again over the plugin since your feedback is very valuable. Apart from the rebasing again, do you have any further suggestions here?

PR #2262, opened 2018-10-02
mmapstorage: change unlinking approach

### LINKED ISSUE #2248: mmapstorage: lock mapped files
## Problem

When mmapstorage returns from the `plugin->kdbGet()` function, the returned keyset is inside a mmap region backed by the configuration/mmap file. There is no specified behaviour, if the file changes after the `mmap()` call. On my system, the changes become visible in the mapped region. Overwriting a currently mapped file therefore destroys old mappings. The previously returned keysets (in elektrified client programs) are therefore invalid and result in SIGBUS/SIGSEGV/unspecified behaviour.

The problem is partly mitigated by mmapstorage remembering current mappings inside the `plugin->data`. Since the data is not shared, this does not work for other plugin instances or other KDB instances.

## Considered Solutions

1. Use `fcntl()` to set advisory locks, so at least two instances of KDB can not overwrite the underlying file while it is mapped. This is already implemented in #2246, but that makes overwriting a mapped config/cache file impossible.
2.  Use a different type of communication to synchronize write access to the mmap file. (shared memory,...?)
3. Use mmap() (or fread/fwrite) only for read/write operations and immediately close the mapping. Return a "normal" malloc()-ed keyset. A quick test suggests that this would be merely 30x faster than dump (instead of the current 67x improvement). I would have to implement it properly to be sure though.

## Implications 

What is more of interest for the team is probably how this effects the global cache.

One possibility is that we could keep the mapped keyset only internally, and expose only malloc()-ed keysets to the user. The downside is that we run into the same problems when a KDB instance wants to overwrite a cache file, which might still be mapped by another KDB instance. (-> we then need a more complex solution)

The more I think about it, the more I lean towards the version (3) where we use mmap() only to read and write.

Any ideas by the team are welcome. @markus2330 you might also want to comment.

mpranj:
Resolves #2248. Unlinking is now implemented by simply calling `unlink()` instead of a custom approach that was error prone.
Includes some other small cleanups and fixes.

## Checklist

need to be checked.

- [x] I added unit tests
- [x] I removed references to old unlinking and locking approaches.

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -40,27 +40,27 @@ You can also read the news [on our website](https://www.libelektra.org/news/0.8.
   The `mmapstorage_crc` variant enables CRC32 checksums for critical data,
   while the `mmapstorage` variant omits the checksum for maximum performance.
 
-  We ran a synthetic benchmark with 127 iterations using 40k keys in a keyset,
+  We ran a synthetic benchmark with 257 iterations using 40k keys in a keyset,
   and compared the performance to the `dump` storage plugin.
 
   Median write time in microseconds:
 
   | Plugin | Time |
   | --- | --- |
-  | `dump` | 63794 |
-  | `mmapstorage` | 4325 |
-  | `mmapstorage_crc` | 9072 |
+  | `dump` | 71079 |
+  | `mmapstorage` | 2964 |
+  | `mmapstorage_crc` | 7644 |
 
   Median read time in microseconds:
 
   | Plugin | Time |
   | --- | --- |
-  | `dump` | 76419 |
-  | `mmapstorage` | 1115 |
-  | `mmapstorage_crc` | 5338 |
+  | `dump` | 82737 |
+  | `mmapstorage` | 1145 |
+  | `mmapstorage_crc` | 5744 |
 
-  In our benchmark, the `mmapstorage` plugin writes more than 14x faster,
-  and reads more than 68x faster than the `dump` storage plugin. *(Mihael Pranjić)*
+  In our benchmark, the `mmapstorage` plugin writes more than 23x faster,
+  and reads more than 72x faster than the `dump` storage plugin. *(Mihael Pranjić)*
 
 
 - Hybrid Search Algorithm for `ksLookup (...)`
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -47,20 +47,20 @@ You can also read the news [on our website](https://www.libelektra.org/news/0.8.
 
   | Plugin | Time |
   | --- | --- |
-  | `dump` | 63692 |
-  | `mmapstorage` | 4338 |
-  | `mmapstorage_crc` | 8813 |
+  | `dump` | 63794 |
+  | `mmapstorage` | 4325 |
+  | `mmapstorage_crc` | 9072 |
 
   Median read time in microseconds:
 
   | Plugin | Time |
   | --- | --- |
-  | `dump` | 74889 |
-  | `mmapstorage` | 1116 |
-  | `mmapstorage_crc` | 5250 |
+  | `dump` | 76419 |
+  | `mmapstorage` | 1115 |
+  | `mmapstorage_crc` | 5338 |
 
   In our benchmark, the `mmapstorage` plugin writes more than 14x faster,
-  and reads more than 67x faster than the `dump` storage plugin. *(Mihael Pranjić)*
+  and reads more than 68x faster than the `dump` storage plugin. *(Mihael Pranjić)*
 
 
 - Hybrid Search Algorithm for `ksLookup (...)`
--- a/src/plugins/mmapstorage/README.md
+++ b/src/plugins/mmapstorage/README.md
@@ -83,5 +83,4 @@ sudo kdb umount user/tests/mmapstorage
 
 ## Limitations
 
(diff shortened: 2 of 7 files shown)


mpranj:
@markus2330 please review my PR.

doc/news/_preparation_next_release.md
```diff
@@ -40,27 +40,27 @@ You can also read the news [on our website](https://www.libelektra.org/news/0.8.
   The `mmapstorage_crc` variant enables CRC32 checksums for critical data,
   while the `mmapstorage` variant omits the checksum for maximum performance.
 
-  We ran a synthetic benchmark with 127 iterations using 40k keys in a keyset,
+  We ran a synthetic benchmark with 257 iterations using 40k keys in a keyset,
```
markus2330: Did you measure the time of all 257 iterations together?

Why increasing this number? 127 is already quite high, usually the iterations are only used to avoid that an outlier is taken.

src/plugins/mmapstorage/README.md
```diff
@@ -80,3 +80,7 @@ kdb rm -r user/tests/mmapstorage
 # Unmount mmapstorage
 sudo kdb umount user/tests/mmapstorage
 ```
+
+## Limitations
+
+Currently mapped files shall not be altered, otherwise the behaviour is undefined.
```
markus2330: You can remove the "Currently".

src/plugins/mmapstorage/mmapstorage.c
```diff
@@ -48,40 +49,39 @@ static MmapMetaData magicMmapMetaData;
...
  */
-static FILE * openFile (Key * parentKey, const char * mode)
+static int openFile (Key * parentKey, int flags)
 {
-	FILE * fp;
+	int fd;
 	ELEKTRA_LOG_DEBUG ("opening file %s", keyString (parentKey));
 
-	if ((fp = fopen (keyString (parentKey), mode)) == 0)
+	if ((fd = open (keyString (parentKey), flags, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH)) == -1)
```
markus2330: Why do you give write permissions? You always unlink the file, so write permissions should be never needed.

doc/news/_preparation_next_release.md
```diff
@@ -40,27 +40,27 @@ You can also read the news [on our website](https://www.libelektra.org/news/0.8.
...
   and compared the performance to the `dump` storage plugin.
 
   Median write time in microseconds:
 
   | Plugin | Time |
   | --- | --- |
-  | `dump` | 63692 |
-  | `mmapstorage` | 4338 |
-  | `mmapstorage_crc` | 8813 |
+  | `dump` | 71079 |
```
markus2330: Any idea why it got slower?

src/plugins/mmapstorage/mmapstorage.c
```diff
@@ -1219,25 +923,17 @@ int ELEKTRA_PLUGIN_FUNCTION (mmapstorage, get) (Plugin * handle, KeySet * ks, Ke
...
-	Key * found = ksLookup (mappedFiles, parentKey, KDB_O_POP);
-	if (found)
-	{
-		unlinkFile (found);
-		keyDel (found);
-	}
-
-	FILE * fp;
-	char * mappedRegion = MAP_FAILED;
+	(void) unlink (keyString (parentKey));
```
markus2330: Please check for errors.

src/plugins/mmapstorage/mmapstorage.c
```diff
@@ -1199,8 +904,7 @@ int ELEKTRA_PLUGIN_FUNCTION (mmapstorage, get) (Plugin * handle, KeySet * ks, Ke
 	}
 
 	if (mappedRegion != MAP_FAILED) munmap (mappedRegion, sbuf.st_size);
-	if (fp) fclose (fp);
-	keyDel (found);
+	if (fd != -1) close (fd);
```
markus2330: Please check for errors.

doc/news/_preparation_next_release.md
mpranj: For the average times I provided, I give the median of 257 measured instances. I did not sum it up, if that is what you mean.

I wanted to get rid of some outliers. I found later that it was because of the turbo boost and cpu freq governor of my cpu. For the updated numbers, turbo was disabled and cpu frequency governor was set to "performance".

src/plugins/mmapstorage/mmapstorage.c
mpranj: @markus2330 would you in general prefer minimal access (only user can read)?

I think the maximum should be user + group access. Preferrably even only S_IRUSR. If others have access, it might leak data that should not be readable by others.

mpranj: I tested it and the minimal access mode bits are `S_IRUSR | S_IWUSR`.

`mmapstorage` works with less, but it introduces problems with other parts of the tests. (1/3 of the tests fail on my machine)

mpranj:
@markus2330 thank you for the review. I am done with the changes.

No mmap structures are needed in `kdbprivate.h` any more, so I removed them in 897eda461dafa6c5c124a1e58bdfa42738b7d68f. Sorry that this is post review, but I figured it fits nicely in this PR.

markus2330:
Thank you so much! Good to see so much progress now!

PR #2263, opened 2018-10-05
Travis: Mitigate Timeout Problems

sanssecours:
# Description

This update should mitigate the timeout problems of some Travis build jobs (`🐧 GCC ASAN`, `🍏 Clang` and `🍏 MMap`) a little bit. From what I can tell, all build jobs should usually [execute in less than 40 minutes](https://travis-ci.org/sanssecours/elektra/builds/437064505) after the update. This update does increase the overall build time a little bit though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,13 @@ before_cache:
 matrix:
   include:
 
+    - os: osx
+      name: 🍏 Check Source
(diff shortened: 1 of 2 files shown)


markus2330:
Thank you for the maintenance!

PR #2275, opened 2018-10-20
CMake: fix BUID_FULL error

kodebach:
## Basics

Fixes #2273 

## Checklist

- [ ] I added unit tests
- [ ] I ran all tests locally and everything went fine
- [ ] affected documentation is fixed
- [ ] I added code comments, logging, and assertions (see doc/CODING.md)
- [ ] meta data is updated (e.g. README.md of plugins)

## Review

@markus2330 please review my pull request

PR #2279, opened 2018-10-25
Travis: Add Jobs With Enabled `BUILD_FULL` & Disabled `BUILD_SHARED`

sanssecours:
# Description

This pull request is a slightly extended version of PR #2275. I mainly added workarounds for problems that occur, if we enable `BUILD_FULL` and disable `BUILD_SHARED`. One puzzling problem is that compiling the unit test of the Base64 plugin works in macOS, but fails on Linux reporting [linker errors](https://travis-ci.org/sanssecours/elektra/jobs/445799127#L2225-L2229):

```
src/plugins_tests/CMakeFiles/testmod_base64.dir/base64/testmod_base64.c.o: In function `test_base64_encoding':
src/plugins/base64/testmod_base64.c:74: undefined reference to `base64Encode'
src/plugins_tests/CMakeFiles/testmod_base64.dir/base64/testmod_base64.c.o: In function `test_base64_decoding':
src/plugins/base64/testmod_base64.c:98: undefined reference to `base64Decode'
src/plugins/base64/testmod_base64.c:112: undefined reference to `base64Decode'
```

. Maybe someone else has an idea why that is the case…


markus2330:
@sanssecours maybe the order of linker flags is wrong (the linker on Linux might be more pedantic about that).

@petermax2 Or do you have an idea?

sanssecours:
>  maybe the order of linker flags is wrong (the linker on Linux might be more pedantic about that).

Commit 601171e should [solve the issue](https://travis-ci.org/ElektraInitiative/libelektra/jobs/446798802#L2933). I do not know if the fix does introduce other problems though.

src/bindings/jna/CMakeLists.txt
```diff
@@ -12,6 +12,12 @@ if (Java_JAVAC_EXECUTABLE)
 	# as 9.x.x > 1.8 and 1.8.x > 1.8 and 1.9.x > 1.8
 	# ~~~
 	if ((${Java_VERSION} VERSION_GREATER "1.8.0") OR (${Java_VERSION} VERSION_EQUAL "1.8.0"))
+		if (NOT BUILD_SHARED)
```
markus2330: Which problem was the reason to add this? Is maybe ONLY_SHARED the better option here?

src/plugins/base64/CMakeLists.txt
```diff
@@ -7,6 +7,6 @@ add_plugin (base64
 		    base64.c
 	    COMPILE_DEFINITIONS ELEKTRA_PLUGIN_NAME=\"base64\"
 				ELEKTRA_PLUGIN_NAME_C=base64
-	    ADD_TEST
 	    TEST_README
+	    ADD_TEST
```
markus2330: Is this change cosmetic (I agree it is nice to have the same order across files) or needed?

src/bindings/jna/CMakeLists.txt
sanssecours: This change was introduced, because I disabled the unit test in the PR, and did not insert `ADD_TEST` at the same location after I re-enabled the test.

src/plugins/base64/CMakeLists.txt
sanssecours: This change was introduced, because I disabled the unit test in the PR, and did not insert `ADD_TEST` at the same location after I re-enabled the test.

src/bindings/jna/CMakeLists.txt
sanssecours: > Which problem was the reason to add this?

The unit test for the [binding fails](https://travis-ci.org/sanssecours/elektra/jobs/445840045) (with a `NoClassDefFoundError`) on Ubuntu, if we do not enable `BUILD_SHARED`.

> Is maybe ONLY_SHARED the better option here?

Yep, that should work too. I will check, if that is a valid option.

sanssecours: Looks like [using `ONLY_SHARED` does not fix the problem](https://travis-ci.org/ElektraInitiative/libelektra/builds/447434672?utm_source=github_status&utm_medium=notification).

PR #2284, opened 2018-10-28
Jenkins: Add Compiler Switch `-Werror` to Debian Unstable Build Job

sanssecours:


src/plugins/simpleini/simpleini.c
```diff
@@ -152,7 +152,7 @@ static char * replaceStringFormatSpec (char * format, const char * replace)
 		// copy pre replacement
 		strncpy (result, format, preReplLen);
 		// copy replacement
-		strncpy (result + preReplLen, replace, replaceLen);
+		strcpy (result + preReplLen, replace);
```
markus2330: Iirc his causes a warning on OpenBSD.

sanssecours: While, this might be the case we do

- not have a build job that uses OpenBSD, and
- the current version of the code causes a warning on a non-outdated version of GCC

. I also do not think using `strcpy` should be a problem, since we already make sure that the buffer filled by the function is large enough.

If you have a better solution to fix the warning, I am also very open to just use that one instead.

markus2330: Thank you for fixing the warning! Unfortunately I have not seen the warning yet, so I do not know what the problem was. What do you think about pasting the warning text into the commit messages?

It is very nice that you have many separate commits.

> not have a build job that uses OpenBSD

Yes, that is the reason why I needed to say it in the review :smile: 

Anyway, sometimes there are situations where it is impossible to fix a warning for every system. So maybe we need to suppress this warning in OpenBSD once somebody takes care of that system.

sanssecours: > Unfortunately I have not seen the warning yet, so I do not know what the problem was.

The build system emits the warning, if you [build Elektra on Debian Unstable](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/master/296/pipeline/420/#step-706-log-379):

```
src/plugins/simpleini/simpleini.c: In function ‘replaceStringFormatSpec’:
src/plugins/simpleini/simpleini.c:155:3: warning: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
   strncpy (result + preReplLen, replace, replaceLen);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/plugins/simpleini/simpleini.c:142:22: note: length computed here
  size_t replaceLen = strlen (replace);
                      ^~~~~~~~~~~~~~~~
```

.

> What do you think about pasting the warning text into the commit messages?

While I am in general in favor of adding information to commit messages, the warning above is quite common. Searching for “string truncation GCC” on the web should return enough information, if you want to read more about the warning.

src/libs/elektra/CMakeLists.txt
```diff
@@ -46,6 +46,10 @@ add_custom_command (OUTPUT exported_symbols.c
 include (LibAddMacros)
 add_headers (HDR_FILES)
 
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
+	set_source_files_properties (keyset.c PROPERTIES COMPILE_FLAGS "-Wno-stringop-truncation")
```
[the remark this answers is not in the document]
sanssecours: > Is it a good idea to silence the warnings?

If there is no reason why we cut the output one character before `strncpy` is able to fill in the terminating `\0` character, then we can also fix the warnings by increasing the size parameter by one. Another option would be to use `memcpy` instead of `strncpy`.

> Shouldn't we at least open an issue or an TODO for that?

I opened pull request #2288, which should fix the problem.

PR #2288, opened 2018-10-30
KeySet: Fix Warnings About Truncated Output

sanssecours:


src/libs/elektra/keyset.c
```diff
@@ -1191,7 +1191,7 @@ KeySet * ksCut (KeySet * ks, const Key * cutpoint)
 		++it;
 	}
 
-	// correct cursor if cursor is in cutted keyset
+	// correct cursor if cursor is in cut keyset
```
markus2330: Thank you! Maybe we could add even a rule to always replace cutted with cut?

src/libs/elektra/keyset.c
```diff
@@ -1113,39 +1113,39 @@ KeySet * ksCut (KeySet * ks, const Key * cutpoint)
 			switch (ns)
 			{
 			case KEY_NS_SPEC:
-				strncpy (newname + 2, "spec", 4);
+				strncpy (newname + 2, "spec", 5);
```
markus2330: We could introduce a convenience macro ELEKTRA_COPY_CONSTANT_STRING which does strncpy(ptr, str, sizeof(str)).

doc/news/_preparation_next_release.md
```diff
@@ -315,19 +315,19 @@ you up to date with the multi-language support provided by Elektra.
 - The build jobs now print less non-relevant output. *(René Schwaiger)*
 - Enable `-Werror` in `debian-stable-full`. *(Lukas Winkler)
 - We added the compiler switch `-Werror` to the build jobs:
```
markus2330: Wasn't it added to all the jobs in the end? Maybe also mention where it is not done yet, if you already know the list.

And you forgot to mention your most important contribution: you did not only add "-Werror" but you also fixed all warnings so that it actually works :heart:

src/libs/elektra/keyset.c
sanssecours: I do not think that does not make much sense currently, since `fix-spelling` does not check mistakes inside source code. I replaced another occurrence of “cutted” with “cut” in commit 49fd0f87.

doc/news/_preparation_next_release.md
sanssecours: > Wasn't it added to all the jobs in the end?

Only the job for Debian “Jessie“ is missing. However, compiling Elektra does not work anymore if we add the compiler switch `-Werror` on that OS. The reason behind that is that `find_package (Threads)` fails, if we use CMake `3.0.2` and add said switch.

markus2330: That is perfectly fine if it does not work for this legacy system, I only wanted to say it would be nice if it is documented in the release notes.

PR #2300, opened 2018-11-20
Array: Change Behavior of `elektraArrayValidateBaseNameString` Slightly

sanssecours:
# Description

This pull request improves the behavior of `elektraArrayValidateBaseNameString`, which now returns the offset to the first digit of the array index for valid array elements:

```c
elektraArrayValidateBaseNameString ("#_10");
//                                     ~~^ Returns `2` (instead of `1`)
 elektraArrayValidateBaseNameString ("#___1337");
//                                    ~~~~^ Returns `4` (instead of `1`)
```

. While the updated behavior is slightly different from the current one, I think the increased utility makes up for breaking (hypothetical existing) user code that compares the return value of the function to `1` using the equality operator (instead of `>=`).

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -94,9 +94,34 @@ compiled against an older 0.8 version of Elektra will continue to work
 - <<TODO>>
 - <<TODO>>
 
-### <<Library1>>
+### Libease
 
-- <<TODO>>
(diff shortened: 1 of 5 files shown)


markus2330:
Thank you!

Maybe we should increase the SO_VERSION next release and release 0.9, let us see which other changes we will have.

PR #2315, opened 2019-01-17
Highlevel-API Improvements

kodebach:
These are further improvements to the highlevel-API as requested in #2293.

### DIFF
--- a/src/include/elektra.h
+++ b/src/include/elektra.h
@@ -30,11 +30,11 @@
 
 #define ELEKTRA_TAG(typeName) struct Elektra##typeName##Tag
 
-#define ELEKTRA_SET_BY_TAG(typeName) _elektraSet##typeName##ByTag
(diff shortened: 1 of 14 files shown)


src/include/elektra.h
```diff
@@ -253,7 +253,7 @@ void elektraFatalError (Elektra * elektra, ElektraError * fatalError);
  *
  **************************************/
 
-const char * elektraGetValue (Elektra * elektra, const char * name);
+const char * elektraGetRawString (Elektra * elektra, const char * name);
```
kodebach: @markus2330 I didn't remove this function, because `elektraSetRawString` (previously `elektraSetValue`) is actually needed for code-generation and having a public setter without the corresponding getter seemed weird. I think `RawString` describes better what the function actually does.

Also I thought it might be useful to have this function, if a user wants to extend the API with a custom type (e.g. something like BigInteger from Java). Users could write their own functions that check the type with `elektraGetType` and the use `elektraGetRawString` followed by some custom conversion.

Issue #1252, opened 2017-01-04
Integrate command-line arguments and environment variables


markus2330:
While code generation already considers command-line (even though it does not use the correct namespace, should be fixed, too), without code generation there is currently no automatic way to facilitate command-line arguments and environment variables.

I propose to implement the function  
`int elektraGetEnv(KeySet *ks, int argc, char ** argv, char ** envp)`

which goes through every key in `spec` and looks for meta-keys `opt`, `opt/long` and `env`. If it finds them, it will try to parse argv/envp. If the command-line argument or environment variable is available, it will add a new key with namespace `proc` accordingly.

For example, if it finds  the key of namespace `spec` (written down in ni syntax):
```
[spec/sw/elektra/restbackend/#0/current/cppcms/service/ip]
check/type=string
description=This option defines the IPv4/IPv6 IP the application should listen on.
opt=i
opt/long=ip
env=CPPCMS_IP
```
it will add `proc/sw/elektra/restbackend/#0/current/cppcms/service/ip` if:

1. `-i` was passed as command-line argument
2. `--ip` was passed as command-line argument
3. `CPPCMS_IP` is available as environment variable (in that order, so env has no effect if command-line were already there)

depending on the type (#1092), it will also use `optarg` for command-line parsing.

The description, type, default, and other info should be displayed when using `--help`.

Furthermore I propose a global plugin that calls `elektraGetEnv` (getting the argc, argv and envp from `/proc/cmdline` and `environ` on Linux, `GetCommandLine` on Windows, `_NSGetArgv` and `_NSGetArgc` on MacOSX) so that it will also work for applications not directly calling `elektraGetEnv`. This idea also plays nicely with order preserving hash map and mmap: We could always add the `proc` keys, but disable them.

Implementation should be in C, in a new library `src/libs/getenv` and a new plugin `src/plugins/getenv`.

Please comment on the proposal.

kodebach:
### Regarding type checking

It absolutely would make sense to check, whether a command-line argument (or environment variable) matches the specified type. But I think we should actually check the whole specification, i.e. also verify `check/path` constraints for example.

The problem is that all these checks are done by plugins in the `kdbSet` stage. Because running `kdbSet` is not feasible for various reasons (expensive, needs `kdbGet` aftwards, `proc` namespace ignored, etc.), I don't see a good way to check the specification.

[...]

PR #2317, opened 2019-01-17
Command line argument parsing

kodebach:
Implementation of #1252. The name was changed to `elektraGetOpts` to avoid confusion with the old [getenv](https://master.libelektra.org/src/bindings/intercept/env).

## Basics

## Checklist

- [x] I added unit tests

### DIFF
--- a/src/error/specification
+++ b/src/error/specification
@@ -1292,3 +1292,33 @@ description:Connection Error occured
 severity:error
 ingroup:plugin
 module:network
+
+number:206
+description:An unknown option was found.
+severity:error
+ingroup:opts
+macro:OPTS_UNKNOWN_OPTION
+
+number:207
+description:An option is missing an argument.
+severity:error
+ingroup:opts
+macro:OPTS_MISSING_ARGUMENT
+
+number:208
+description:The flagvalue metadata can only be used, if the opt/arg metadata is set to 'none'.
+severity:error
+ingroup:opts
+macro:OPTS_FLAGVALUE_ARG
+
+number:209
+description:A non-repeatable option was repeated.
+severity:error
+ingroup:opts
+macro:OPTS_ILLEGAL_REPEAT
+
+number:210
+description:Duplicate option specified.
+severity:error
+ingroup:opts
+macro:OPTS_DUPLICATE
--- a/src/error/specification
+++ b/src/error/specification
@@ -1294,31 +1294,19 @@ ingroup:plugin
 module:network
 
 number:206
-description:An unknown option was found.
+description:An unknown option was found (see reason).
 severity:error
 ingroup:opts
 macro:OPTS_UNKNOWN_OPTION
 
 number:207
-description:An option is missing an argument.
+description:An option was used in an unsupported way (see reason).
 severity:error
 ingroup:opts
-macro:OPTS_MISSING_ARGUMENT
+macro:OPTS_ILLEGAL_USE
 
 number:208
-description:The flagvalue metadata can only be used, if the opt/arg metadata is set to 'none'.
+description:The opts specification contains mistakes (see reason).
 severity:error
 ingroup:opts
-macro:OPTS_FLAGVALUE_ARG
-
-number:209
-description:A non-repeatable option was repeated.
-severity:error
-ingroup:opts
-macro:OPTS_ILLEGAL_REPEAT
-
-number:210
-description:Duplicate option specified.
-severity:error
-ingroup:opts
-macro:OPTS_DUPLICATE
+macro:OPTS_ILLEGAL_SPEC
--- a/src/libs/opts/opts.c
+++ b/src/libs/opts/opts.c
@@ -40,7 +40,8 @@ static Key * splitEnvValue (const Key * envKey);
 
 static KeySet * ksMetaGetSingleOrArray (Key * key, const char * metaName);
 
-static char * generateHelpMessage (KeySet * optionsSpec, const char * progname);
+static char * generateUsageLine (const char * progname, bool hasOpts, bool hasArgs);
+static char * generateOptionsList (KeySet * keysWithOpts);
 
 /**
  * This functions parses a specification of program options, together with a list of arguments
@@ -54,9 +55,15 @@ static char * generateHelpMessage (KeySet * optionsSpec, const char * progname);
  * will be set to a help message describing available options and 1 will be returned. The program
  * name used in this message is taken from `argv[0]`. If it contains a '/' only the part after the
  * last '/' will be used. The help message will contain description for each option. This description
- * is taken from `opt/help`, or if `opt/help` is not specified from `description`. If neither `opt/help`
- * nor `description` are given, or `opt/nohelp` is set to "1", the option will not appear in the help
- * message. The same applies to `env`s.
+ * is taken from `opt/help`, or if `opt/help` is not specified from `description`. The `opt/help` metadata
+ * is specified PER KEY not per option, even if the key has multiple options. This is because all the
+ * options for one key have to be equivalent and therefore only require one description.
+ *
+ * If `opt/nohelp` is set to "1", the option will not appear in the help message. The same applies to
+ * `env`s and `env/nohelp`.
+ *
+ * If `opt/arg/help` is set, its value will be used as the argument name for long options. Otherwise
+ * "ARG" will be used. This value can be set for each option individually.
  *
  * The help message can also be customised in two more ways:
  * 1. By setting the `help/usage` metadata on @p errorKey, you can use a custom usage line for the help
@@ -133,8 +140,13 @@ int elektraGetOpts (KeySet * ks, int argc, const char ** argv, const char ** env
 	KeySet * keysWithOpts = ksNew (0, KS_END);
 	KeySet * usedEnvVars = ksNew (0, KS_END);
 	KeySet * optionsSpec = ksNew (
-		2, keyNew ("/short/h", KEY_META, "hasarg", "none", KEY_META, "kind", "single", KEY_META, "flagvalue", "1", KEY_END),
+		2,
+		keyNew ("/short/h", KEY_META, "hasarg", "none", KEY_META, "kind", "single", KEY_META, "flagvalue", "1", KEY_META, "longopt",
+			"/long/help", KEY_END),
 		keyNew ("/long/help", KEY_META, "hasarg", "none", KEY_META, "kind", "single", KEY_META, "flagvalue", "1", KEY_END), KS_END);
+
+	bool hasOpts = false;
+	bool hasArgs = false;
 
 	cursor_t initial = ksGetCursor (ks);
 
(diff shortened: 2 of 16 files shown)


tests/ctest/test_opts.c
```diff
@@ -0,0 +1,779 @@
...
+
+#include <kdberrors.h>
+#include <kdbopts.h>
+
+#include "tests.h"
+
+#define PROC_BASE_KEY "proc/tests/opts"
+#define SPEC_BASE_KEY "spec/tests/opts"
+
+// version 6 and 7 of clang-format don't agree whether it is supposed to be *[] or * [] so disable it here
```
markus2330: Isn't it easier if we all use the same version (which is now 6)? We can upgrade to 7 when it is in Debian stable.

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+ * There is no `help/suffix` metadata, because you can easily append text to the help message yourself.
+ *
+ * Because "-h" and "--help" are reserved for the help message, neither can be used for anything else.
+ *
+ * To define a command line option set the `opt` meta-key to the short option. Only the first
+ * character of the given value will be used ('\0' is not allowed). You can also set `opt/long`,
+ * to use a long option. Each option can have a short version, a long version or both.
+ * Per default an option is expected to have an argument. To change this behaviour set `opt/arg` to
+ * either 'none' or 'optional' (the default is 'required'). The behaviour of short options, long
+ * options, required and optional arguments is similar to getopt_long(3).
```
markus2330: similar? What are the differences? (as described below?)

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+		Key * key = NULL;
+		KeySet * opts = ksMetaGetSingleOrArray (cur, "opt");
+		if (opts != NULL)
+		{
+			ksRewind (opts);
+			ksNext (opts); // skip count TODO: bug?
+			Key * k;
+			while ((k = ksNext (opts)) != NULL)
+			{
+				char metaBuffer[ELEKTRA_MAX_ARRAY_SIZE + 15]; // 14 = opt//flagvalue (1 extra to circumvent warning)
```
markus2330: you can also write sizeof("opt/flagvalue")? (+comment has double slash between opt and flagvalue)

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+					keyDel (argKey);
+					keyDel (shortOpt);
+					keyDel (optSpec);
+					ksDel (options);
+					return NULL;
+				}
+				keyDel (optSpec);
+
+				bool last = false;
+				if (strcmp (hasArg, "required") == 0)
```
markus2330: elektraStrCmp

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+ * Because "-h" and "--help" are reserved for the help message, neither can be used for anything else.
+ *
+ * To define a command line option set the `opt` meta-key to the short option. Only the first
+ * character of the given value will be used ('\0' is not allowed). You can also set `opt/long`,
+ * to use a long option. Each option can have a short version, a long version or both.
+ * Per default an option is expected to have an argument. To change this behaviour set `opt/arg` to
+ * either 'none' or 'optional' (the default is 'required'). The behaviour of short options, long
+ * options, required and optional arguments is similar to getopt_long(3).
+ *
+ * A short option cannot be set to 'optional', as the only way to give an 'optional' argument is via
```
markus2330: Please also give examples. (Ideally in separate source files included here as snippets.)

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+ * the "--opt=value" syntax of long options. If a long option with an 'optional' argument does not
```
markus2330: Split this documentation here into a user guide/tutorial how to use command-line arguments (next to development docu for the function).

tests/ctest/test_opts.c
```diff
@@ -0,0 +1,779 @@
...
+	if (elektraGetOpts (ks, ARGS ("-ashort", "other", "-a", "short2"), NO_ENVP, errorKey) != 0)
+	{
+		yield_error ("error found");
+		output_error (errorKey);
+	}
+	keyDel (errorKey);
+	succeed_if (checkValue (ks, PROC_BASE_KEY "/apple", "short"), "should have stopped after short option");
+	succeed_if (checkValue (ks, PROC_BASE_KEY "/rest", "#2"), "rest has wrong count");
+	succeed_if (checkValue (ks, PROC_BASE_KEY "/rest/#0", "other"), "rest has wrong value (#0)");
+	succeed_if (checkValue (ks, PROC_BASE_KEY "/rest/#1", "-a"), "rest has wrong value (#1)");
```
markus2330: Shouldn't that be an error? (Specifying -a again without argument?)

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+						"'optional'. (key: %s)",
+						keyName (cur));
+					ksDel (opts);
+					ksDel (optionsSpec);
+					ksDel (keysWithOpts);
+					ksDel (usedEnvVars);
+					return -1;
+				}
+
+				const char * shortOpt = keyString (k);
```
markus2330: refactor a bit more? (function is quite lengthy)

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+ * Lastly, all remaining elements of @p argv will be collected into an array. You can access this array
+ * by specifying `args = remaining` on a key with basename '#'. The array will be copied into this key. As is
+ * the case with getopt(3) processing of options will stop if '--' is encountered in @p argv.
+ *
+ * NOTE: Both options and environment variables can only be specified on a single key. This is because
+ * otherwise e.g. the help message ('opt/help', 'env/help') may be ambiguous. If you need to have the
+ * value of one option/environment variable in multiple keys, you may use fallbacks.
+ *
+ * NOTE: Per default option processing DOES NOT stop, when a non-option string is encountered in @p argv.
+ * If you want processing to stop, set the metadata `posixly = 1` on @p errorKey.
```
markus2330: Do we need this feature? If possible we should avoid hard-coded configuration within applications so that users have the same behavior when using different applications.

markus2330:
Thank you! Maybe we will have a look at this PR together on Tuesday. The PR is very nice, especially if you also add an tutorial :smile:

@ulschaef Maybe you also want to have a look at it, especially if the meta-specification is suitable for your purposes.

src/error/specification
```diff
@@ -1292,3 +1292,21 @@ description:Connection Error occured
...
 module:network
+
+number:206
+description:An unknown option was found (see reason).
+severity:error
+ingroup:opts
+macro:OPTS_UNKNOWN_OPTION
+
+number:207
+description:An option was used in an unsupported way (see reason).
```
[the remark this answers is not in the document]
kodebach: This error is used, if an option is used in a way that doesn't match the specification. e.g. missing argument

It is also used, if two or more options specified on the same key are used, which might make sense in some scenarios (e.g. repeatable options could be merged), but is not supported by the library.

Basically this error code is used for errors that occur during parsing of argv and envp (except for the case of an unknown option). While `OPTS_ILLEGAL_SPEC` is used during parsing of the specification.

src/libs/opts/opts.c
```diff
@@ -0,0 +1,951 @@
...
+ * The options have to be defined in the metadata of keys in the spec namespace. If an option value
+ * is found for any of the given keys, a new key with the same path but inside the proc namespace
+ * will be inserted into @p ks. This enables a cascading lookup to find these values.
+ *
+ * If @p argv contains "-h" or "--help" @p ks will not be changed, instead the value of @p errorKey
+ * will be set to a help message describing available options and 1 will be returned. The program
+ * name used in this message is taken from `argv[0]`. If it contains a '/' only the part after the
+ * last '/' will be used. The help message will contain description for each option. This description
+ * is taken from `opt/help`, or if `opt/help` is not specified from `description`. If neither `opt/help`
+ * nor `description` are given, or `opt/nohelp` is set to "1", the option will not appear in the help
```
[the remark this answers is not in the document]
kodebach: > I think the option should be listed, even if neither opt/help nor description is given.

You are right, `opt/nohelp` should be good enough.

> What is the usecase for opt/nohelp? It is not described above?

 I will add documentation. clang for example has some hidden options, that are only mentioned when `--help-hidden` is used. It could also be used for deprecated options only supported for legacy reasons (i.e. to not break old scripts).

src/libs/opts/opts.c
kodebach: The double slash is intentional. We need one slash before the array index and one after.

tests/ctest/test_opts.c
kodebach: No, this test uses the "posixly" option (same as `POSIXLY_CORRECT` for getopt_long(3)). Therefore option processing should stop before "other".

src/libs/opts/opts.c
kodebach: I included it to mimic `POSIXLY_CORRECT` from getopt_long(3). The feature may be useful for an interface like the one `git` uses. General options like `--version` are specified before the command. The first argument in `argv` that is not an option is used as the command. Everything after the command is treated as arguments to the command.

This is a bit harder to implement with `elektraGetOpts`, because you don't automatically get the index of the command, but would still be possible.

src/libs/opts/opts.c
markus2330: If all differences are listed, then simply say "as described below". If users should not assume that the behavior is as in `getopt_long`, you better not mention it in such specific statements.

tests/ctest/test_opts.c
markus2330: If we really want to support `POSIXLY_CORRECT`, then we maybe should also support it as configuration setting and not let it hard-code within applications.

The question is if we want `POSIXLY_CORRECT` only as environment variable or as Elektra configuration setting (also in config files next to env var).

src/libs/opts/opts.c
markus2330: This can be implemented as later extension. For the `kdb` tool we also need a functionality like this. Would be great if the `kdb` tool also uses your library. And if the OPTIONS sections of the man pages are generated from the --help output. (or a tool directly parsing the specification)

@ulschaef already has some ideas how to extend the specification so that "subcommands" can be supported.

kodebach: > This can be implemented as later extension. 

It is implemented already. The only thing I would (maybe) add to better support "subcommands" is a way to assign the index of the first non-option argument to a key (or maybe even return it directly in some way to avoid conversion to string an back again). Currently you would need to calculate this index from the size of the remaining args array.

The idea would be:
1. call `elektraGetOpts` with the original `argc` and `argv`, in posixly mode.
2. get the index of the first non-option argument
3. call `elektraGetOpts` with changed `argc` and `argv`, so that `argv[0]` is the subcommand.

> And if the OPTIONS sections of the man pages are generated from the --help output.

If you pass `""` as the usage line and `NULL` as prefix to `elektraGetOptsHelpMessage` the result will only be the options list (you might need to trim some whitespace).

tests/ctest/test_opts.c
kodebach: I don't think `POSIXLY_CORRECT` should be user configurable. As the default suggests, not using `POSIXLY_CORRECT` is the way to go in most cases. The non-posix mode is more flexible for the user.

But if you use an interface like `git` uses, you would rely on `POSIXLY_CORRECT`. You program would break, if the user disables it, because there would be no way to know whether an option was given to the main program itself (i.e. before the subcommand in argv) or to the subcommand (i.e. after the subcommand in argv).

src/libs/opts/opts.c
markus2330: Yes, I meant support for subcommands as future extension.

> the result will only be the options list

Nice. Maybe we should also have a `kdb` tool which calls elektraGetOptsHelpMessage for a given specification?

tests/ctest/test_opts.c
markus2330: You mean relying on that `POSIXLY_CORRECT`  is **not** set?

`git -p diff` and `git diff -p` do the same thing, what do you mean with "options passed to subcommand vs. command"? With `POSIXLY_CORRECT` (if git would implement it) users could always do `git -p diff`. It would be strange but it should work?

The whole idea of `POSIXLY_CORRECT` is to have a hack for situations where "POSIX and common sense disagree". I do not know if anyone actually uses this option.

kodebach: > git -p diff and git diff -p do the same thing

That is actually a coincidence. The `-p` before `diff` is short for `--paginate` and pipes all output through `less`. The `-p` after `diff` is short for `--patch`, which is the default option for `diff` and can therefore be omitted.

Try the same with `-u` which is an alternative for `-p`. `git -p diff` will result in "unknown option". As far as I can tell the reasoning behind this behaviour is to allow general `git` options and subcommand options to use the same names.

markus2330:
Seems like the build server did not complete successfully? Please set "ready to merge" if everything is ready.

sanssecours:
> Seems like the build server did not complete successfully?

As far as I can tell, all of the failing build jobs are caused by issue #2328.

markus2330:
Thank you!

I merged this PR, updates please in a new PR.

PR #2327, opened 2019-01-22
Documentation fix for elektraMetaArrayToKS

### LINKED ISSUE #2316: Possible bug in elektraMetaArrayToKS
The documentation for `elektraMetaArrayToKS` gives the following example:

```
* @code
elektraMetaArrayToKS(
	keyNew ("/a", KEY_VALUE, "b, c",
		KEY_META, "dep",    "#1",
		KEY_META, "dep/#0", "/b",
		KEY_META, "dep/#1", "/c", KEY_END),
	"dep");
 * @endcode
 *
 * returns a `KeySet` containing the keys `"dep/#0"` with value `"/b"` and
 * `"dep/#1"` with value `"/c"`.
```

But the actual implementation also includes the key `"dep"` in the result:
```c
KeySet * elektraMetaArrayToKS (Key * key, const char * metaName)
{
	const Key * meta = keyGetMeta (key, metaName);
// [...]
	ksAppendKey (result, keyDup (meta));
// [...]
	return result;
}
```

Is this a bug in the implementation or a mistake in the documentation?

kodebach:
Closes #2316

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -208,7 +208,8 @@ you up to date with the multi-language support provided by Elektra.
 ## Documentation
 
 - We fixed various spelling mistakes. *(René Schwaiger)*
-- <<TODO>>
+- The documentation for `elektraMetaArrayToKS` was fixed. It now reflects the fact
+  that the parent key is returned as well.  *(Klemens Böswirth)*
 - <<TODO>>
 
 ## Tests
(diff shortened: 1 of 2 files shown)


markus2330:
Thank you for fixing the docu!

"with value `#1`" is actually a deprecated way (it should be in meta data "array") but this PR is okay as it documents what it does.

kodebach:
> "with value #1" is actually a deprecated way (it should be in meta data "array")

The function takes the array parent key as-is and appends it to the result. Modifying the resulting keys and adding meta-data onto meta-keys seems a bit unexpected to me.

Also the build server problems are most definitely not caused by any changes made in this PR, so I think we can merge this regardless.

markus2330:
Thank you!

PR #2329, opened 2019-01-23
Network: Add Markdown Shell Recorder Test for Plugin

sanssecours:
This pull request fixes 

- the [Travis build problems](https://issues.libelektra.org/2328) caused by the IO GLIB binding,
- the problems we had with the [tests of the process library on FreeBSD](https://issues.libelektra.org/2323) and
- the [failing tests of the Network plugin](https://issues.libelektra.org/2322) on FreeBSD

.


doc/Doxyfile
```diff
@@ -1034,25 +1034,6 @@ USE_HTAGS              = NO
 
 VERBATIM_HEADERS       = NO
 
-# If the CLANG_ASSISTED_PARSING tag is set to YES, then doxygen will use the
```
markus2330: Why to remove this?

sanssecours: Before this update my version of Doxygen would print the following error message:

```
warning: Tag `CLANG_ASSISTED_PARSING' at line 1046 of file `build/doc/Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
warning: Tag `CLANG_OPTIONS' at line 1054 of file `build/doc/Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
```

. Since we do not use these options at all, I do not think it makes any difference, if we just remove them from the `Doxyfile`.

markus2330: Activating these options might bring an benefit for the documentation of the C++ bindings (if Doxygen is compiled with clang support). But it is a minor thing, so it is okay to remove them so that these warnings are gone.

PR #2331, opened 2019-01-23
High Level API: Minor Mini Updates to Documentation

sanssecours:
While the relatively large amount of commits might tell a different story, I did only find very minor issues in the documentation of the High-Level API. As far as I can tell everything looks good. The API is certainly one of our better documented pieces of code!


markus2330:
Thank you!

PR #2334, opened 2019-01-23
Compiling with clang on Windows

### LINKED ISSUE #2303: Compiling with clang on Windows
As discussed, I tried to compile Elektra on Windows using clang.

I started with the current MinGW config:

```
cmake -DENABLE_TESTING=OFF \
	-DBUILD_TESTING=OFF \
	-DBUILD_STATIC=ON \
	-DBUILD_FULL=ON \
	-DBUILD_SHARED=OFF \
	-DPLUGINS="wresolver;dump;list" \
	-DKDB_DEFAULT_RESOLVER=wresolver \
	-DKDB_DEFAULT_STORAGE=dump \
	-DKDB_DB_SYSTEM=kdb -DKDB_DB_SPEC=spec \
	-DBUILD_DOCUMENTATION=OFF \
	-DCMAKE_PIC_FLAGS="" 
````

I then made the following adjustments:
- I set `-DBINDINGS=-ALL` and `-DTOOLS=""` (getopt is needed for the kdb tool but doesn't exist on windows)
- I also set `-DENABLE_OPTIMIZATIONS=OFF` (not sure why but with optimizations enabled there were linker errors for some of the opmphm functions)
- I manually excluded the benchmarks in `benchmarks` and in `src/libs/tools/benchmarks` (by adding `return()` as the first line of the CMakeLists.txt file). 
- I set `SIZEOF_MODE_T` and `SIZEOF_TV_USEC` in `kdbconfig.h` to `0` so that some `#if`s work (there is probably a better solution in CMake for this)
- I added the following lines to `kdb.h` because these are necessary in a lot of places:
  ```
  #if defined(_MSC_VER)
  #include <BaseTsd.h>
  typedef SSIZE_T ssize_t;
  #define SSIZE_MAX MAXSSIZE_T

  #define strcasecmp _stricmp
  #define strncasecmp _strnicmp
  #endif
  ```
- `sys/time.h` has to be replaced  but according to [this](https://www.codefull.org/2015/12/systime-h-replacement-for-windows/) that should not be too hard.
- `unistd.h` isn't available either but as far as I can tell only `read` is used from this header and that inside a CPP file, so it should be easily replaceable. Also `mode_t` has to be manually defined as `typedef int mode_t`.
- `__attribute__(X)` has to be defined as an empty macro. It seems the windows version of clang does not support this feature.

After these changes, I was able to compile. However a still got a whole lot of linker errors. Basically for every function of Elektra (from `kdbGet` to `keyRewindMeta` and everything else) there was an error `undefined symbol: _XXX` (where `_XXX` stands for the function name prefixed with an `_`).

To sum up: 
- Compiling the Elektra library on native Windows might be possible with some changes and if we find out what causes the linker problems.
- Compiling the kdb tool would either require including (or fetching during CMake configuration) of an implementation of getopt. Alternatively (in my opinion the better option) we could use a getopt-compatible C++ library for command-line arguments (e.g. https://github.com/Taywee/args)

### PR DESCRIPTION
**Pull request by kodebach**
Contains various changes made to allow compilation using clang on Windows.

After the changes in this PR compiling should work with the following CMake configuration (from the MinGW build job, plus a few additions).

```
cmake -DENABLE_TESTING=OFF \
	-DBUILD_TESTING=OFF \
	-DBUILD_STATIC=ON \
	-DBUILD_FULL=ON \
	-DBUILD_SHARED=OFF \
	-DPLUGINS="wresolver;dump;list" \
	-DKDB_DEFAULT_RESOLVER=wresolver \
	-DKDB_DEFAULT_STORAGE=dump \
	-DKDB_DB_SYSTEM=kdb -DKDB_DB_SPEC=spec \
	-DBUILD_DOCUMENTATION=OFF \
	-DCMAKE_PIC_FLAGS=""  \
	-DBINDINGS=-ALL \
	-DTOOLS="" \
	-DENABLE_OPTIMIZATIONS=OFF \
	-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE
```

## TODOs
The things listed below need to be done, in order for the native Windows build to be on par with the MinGW version:
- [ ] Get `ENABLE_OPTIMIZATIONS` working.
- [ ] Once the `kdb` tool doesn't use `getopt` anymore, check whether it compiles.
- [ ] Check for a better way than `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` to suppress the `_` prefix on exported symbols.
- [ ] Try compiling the various bindings.

## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

Closes #2303 

PS. I probably won't do much work on this PR in the foreseeable future. I am just putting out there, in case anybody wants to try their luck.


markus2330:

@FelixResch ping, just for you to know that this exists. But maybe mingw is the better way to go.

PR #2307, opened 2019-01-07
Global plugins: global keyset handle

mpranj:
## Basics

Adds a global keyset handle for global plugins #2270. 
Get return value of global plugin calls.
See docs & release notes of PR.

Do not describe the purpose here but:

- [x] Short descriptions should be in the release notes (added as entry in
      doc/news/_preparation_next_release.md which contains `*(my name)*`)
      **Please always add something to the the release notes.**
- [x] Longer descriptions should be in documentation or in design decisions.
- [x] Describe details of how you changed the code in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, should be in the commit messages.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests
- [x] I ran all tests locally and everything went fine
- [x] affected documentation is fixed
- [ ] I added code comments, logging, and assertions (see doc/CODING.md)
- [ ] meta data is updated (e.g. README.md of plugins)

## Review

Remove the line below and add the "work in progress" label if you do
not want the PR to be reviewed:

PR #2337, opened 2019-01-25
Global KeySet: improvements

mpranj:
Improvements to the global keyset.

- all plugins in the KDB now get a handle to the global keyset
- includes all improvements suggested in #2307

## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

### DIFF
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -46,3 +46,4 @@ section here.
 - [Global Validation](global_validation.md)
 - [Default Values](default_values.md)
 - [High Level API](high_level_api.md)
+- [Global KeySet](global_keyset.md)
--- a/doc/decisions/global_keyset.md
+++ b/doc/decisions/global_keyset.md
@@ -1,0 +1,31 @@
+# Global KeySet
+
+## Problem
+
+Some plugins need to communicate more data than is possible to do with metadata.
+This can limit the functionality of plugins, which need to exchange binary or very
+complex data.
+
+## Constraints
+
+## Assumptions
+
+## Considered Alternatives
+
+## Decision
+
+To make the communication between plugins easier, plugins will additionally
+get a handle to a global keyset. The global keyset is tied to a KDB handle, initialized on kdbOpen() and deleted on kdbClose().
+
+## Rationale
+
+The need for a global keyset arose when developing a global cache plugin. A global cache plugin needs to store internal and binary information for the KDB, which is simply not possible with
+metadata.
+
+## Implications
+
+Plugins are responsible for cleaning up their part of the global keyset.
+
+## Related decisions
+
+## Notes
--- a/doc/decisions/global_keyset.md
+++ b/doc/decisions/global_keyset.md
@@ -15,12 +15,12 @@ complex data.
 ## Decision
 
 To make the communication between plugins easier, plugins will additionally
-get a handle to a global keyset. The global keyset is tied to a KDB handle,
-initialized on kdbOpen() and deleted on kdbClose().
+get a handle to a global keyset via `elektraPluginGetGlobalKeySet()`.
+The global keyset is tied to a KDB handle, initialized on `kdbOpen()`
+and deleted on `kdbClose()`.
 
-The global keyset handle is initialized and accessible for global plugins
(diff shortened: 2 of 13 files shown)


doc/decisions/global_keyset.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Decision
+
+To make the communication between plugins easier, plugins will additionally
+get a handle to a global keyset. The global keyset is tied to a KDB handle,
+initialized on kdbOpen() and deleted on kdbClose().
+
+The global keyset handle is initialized and accessible for global plugins
+and plugins within a backend. It is not available to plugins opened outside
+the context of the KDB, since the global keyset is managed by the KDB instance.
```
mpranj: @markus2330 can you please take a look at this paragraph?

You made the suggestion that all plugins should get the global KS handle, which is a good idea.
The caveat is that it's hard  to do for `elektraPluginOpen`. That is why I propose we do it only for
plugins in a backend. That way the KDB passes the handle when opening a backend.

Would you agree with the solution above or did you want to think of some trickery to somehow make it work
with `elektraPluginOpen()` directly?

markus2330: Can you elaborate which plugins are affected? Is it only about the version plugin? If it is also about default resolver and default storage, this is a call for trouble.

Then this distinction seems to be arbitrary. It would be not so surprising to only pass it to global plugins. (Because of the name of the function. Non-global plugins would get NULL when they try to get the global KeySet).

mpranj: It works for the default backend and everywhere where a KDB instance exists.

Since the KDB instance "manages" the global keyset, it just can not work outside
of the context of a KDB instance. (Example: `elektraPluginOpen` can be done without
having a KDB instance available)

markus2330: I do not understand, `elektraPluginOpen` is used for all mounted plugins (see src/libs/elektra/plugin.c line 185 called from backendOpen in src/libs/elektra/backend.c).

So what exactly is "plugins opened outside the context of the KDB"? Do you mean it only affects plugins opened in test cases?

mpranj: > So what exactly is "plugins opened outside the context of the KDB"? Do you mean it only affects plugins opened in test cases?

Sorry for the misunderstanding, `elektraPluginOpen` is not the problem. The global keyset reference is set after the plugin is loaded. I already implemented it in the backends and in the list plugin to set it correctly.

Any other "custom" loading of plugins, like in the tests, does not support the global keyset.

The alternate approach is to force it by changing the `elektraPluginOpen`. I think that introduces unnecessary complexity though.

markus2330: No, you are right. elektraPluginOpen should not depend on KDB.

It is obvious that for manually instantiated plugins the global KeySet can only be available if it was passed before. Maybe rephrase the sentence so that it is clear that the "not available" only refers to manually created plugins.

mpranj:
I added all suggestions from #2307. @markus2330  please review my pull request.

NB: There is one Cirrus check stuck, please ignore it. It seems to be a bug with Cirrus that sometimes an additional check appears and is never completed. @sanssecours any ideas? When I tried rebuilding it, it simply rebuilds the already green instance.

sanssecours:
> any ideas?

As far as I can tell from the [status page listed under “Details”](https://cirrus-ci.com/build/6259352429133824) all 3 Cirrus checks already ran successfully. It seems the status was just not properly updated here on GitHub.

mpranj:
> all 3 Cirrus checks already ran successfully. 

The problem is that there are 3 checks but this pr lists 4 Cirrus checks.

doc/decisions/README.md
```diff
@@ -46,3 +46,4 @@ section here.
 - [Global Validation](global_validation.md)
 - [Default Values](default_values.md)
 - [High Level API](high_level_api.md)
+- [Global KeySet](global_keyset.md)
```
markus2330: Why is it a draft? It is implemented?

mpranj: I'll update it when we add the cache. This will add info about cached vs volatile parts of the global keyset anyway. Thank you!

PR #2346, opened 2019-01-27
Augeas: Change Default Prefix on macOS

sanssecours:


markus2330:
Thank you for the workaround! When brew delivers correct package-config files, we should remove it, though.

sanssecours:
> When brew delivers correct package-config files, we should remove it, though.

The update changed the default Augeas installation directory that CMake uses on macOS, if it is **not able to find the Augeas prefix via `pkg-config`**. This is for the example the case if `pkg-config` is not installed at all, which is currently the case for the macOS Cirrus build job. As far as I know, the Homebrew version of `augeas.pc` should be just fine.

> Thank you for the workaround!

As explained above, I do not think that this PR contains a workaround.

PR #2350, opened 2019-01-29
Highlevel API improvements

### LINKED ISSUE #2344: highlevel: header file names
Currently, we install the following files:
```
usr/include/elektra/elektra.h
usr/include/elektra/errors.h
usr/include/elektra/types.h
usr/include/elektra/errorcodes.h
usr/include/elektra/conversion.h
```

Thus we usually recommend to put usr/include/elektra into the include path, errors.h might clash with other header files.

We should either:

- implement #817, or
- move them to usr/include/elektra/elektra/errors.h

### LINKED ISSUE #2348: Maximum array size
Currently there is a lot of conflicting information in Elektra regarding the maximum size of an array.

- `ELEKTRA_MAX_ARRAY_SIZE` states "a 32bit number has a maximum of 10 digits" therefore it is implied the largest allowed index is `UINT32_MAX` = 4294967295.
- `elektraWriteArrayNumber` requires the used buffer to be at least `ELEKTRA_MAX_ARRAY_SIZE` in size, but accepts a `kdb_long_long_t` for the index. `kdb_long_long_t` is defined as `long`, if `sizeof(long) == 8`, as `long long`, if `sizeof(long long) == 8` and as `int64_t` in C99+. This implies the largest index is `INT64_MAX` = 9223372036854775807.
- The highlevel API uses `size_t` for array indices, which implies `SIZE_MAX` (at least 65535, but no fixed value) as a maximum index.

Additionally there are a few severe bugs connected to `elektraWriteArrayNumber`:
- `kdb_long_long_t` is not defined, if `SIZEOF_LONG != 8 || !defined(HAVE_SIZEOF_LONG_LONG) || SIZEOF_LONG_LONG != 8`. While this not a huge problem, because you simply couldn't compile Elektra under these circumstances, we should add a custom `#error` like we do for `kdb_long_t`.
- As stated above `elektraWriteArrayNumber` accepts a `kdb_long_long_t` parameter an requires a buffer of minimum size `ELEKTRA_MAX_ARRAY_SIZE`. Which is conflicting information and could lead to a buffer overflow, because `snprintf` is also used incorrectly.

@markus2330 What is the actual intended limit for array sizes?

kodebach:
Closes #2344 and closes #2348


kodebach:
@markus2330 I am not sure how exactly I need to fix the debian branch to make the build work...
Should I just change the header paths in `debian/libelektra-dev.install` and push directly to the branch?
The whole setup seems strange to me because neither the highlevel API nor the kdbopts files are present on the debian branch.

markus2330:
> I am not sure how exactly I need to fix the debian branch to make the build work...
Should I just change the header paths in debian/libelektra-dev.install and push directly to the branch?

Exactly! You could also create a PR for the debian branch but then you could not verify yourself if everything is working together. So it is best if you simply push to the debian branch. Please do it asap, as we should keep the time where the master builds are broken as short as possible (if the debian and master branch are inconsistent, the master builds break).

> The whole setup seems strange to me because neither the highlevel API nor the kdbopts files are present on the debian branch.

Yes, the build server merges master into debian. But only for a release, this merge is actually pushed to the debian branch.

kodebach:
@markus2330 The build server says:
> external_check_example_highlevel did not leave mount config in the same state at is was before!
This means the test itself is flawed or the KDB was externally modified while the tests were running!
You can inspect the original mount config in /tmp/tmp.qoVNv3FHZO/mount.export.dump
compared to situation now in /tmp/tmp.qoVNv3FHZO/mount.check.dump

However I cannot reproduce this locally. I even tried completely removing Elektra rebuilding from a clean state and reinstalling, but `kdb run_all` works fine.

markus2330:
Maybe you did not mark it as KDBTESTS and they erroneously run in parallel?

kodebach:
> Maybe you did not mark it as KDBTESTS and they erroneously run in parallel?

The two external tests now only run via `kdb run_all` because the require an installed instance of elektra, so they should not be run in parallel.

tests/shell/external/check_example_highlevel.sh
```diff
@@ -0,0 +1,118 @@
...
+	succeed_if "application did not print myfloatarray[1]"
+	./application | grep "myfloatarray\\[2\\]: $MYFLOAT2"
+	succeed_if "application did not print myfloatarray[2]"
+	./application | grep "myfloatarray\\[3\\]: $MYFLOAT3"
+	succeed_if "application did not print myfloatarray[3]"
+	./application | grep "myfloatarray\\[4\\]: $MYFLOAT4"
+	succeed_if "application did not print myfloatarray[4]"
+
+	"$KDB" umount "$SPECKEY"
+	"$KDB" umount "$<KEY"
```
markus2330: Is this `<` here on purpose?

kodebach:
@markus2330 The Jenkins build works but is "unstable" (no idea what that means).

On Travis (under Linux) I think `LD_LIBRARY_PATH` (or something similar) needs to be set. Maybe @sanssecours can take a look. Interestingly the example fully work with CMake and can be compiled with pkg-config, but the pkg-config versions cannot be executed.

Sidenote: The external examples don't work with pkg-config when only `BUILD_STATIC` is set. To make that work we would need some Makefile magic, because in that case we need to link with a `g++` or `clang++` because of the `kdb` tools.

Another sidenote: We should maybe move our OSX builds to Cirrus CI, Travis takes forever to build these. Also Cirrus CI only enforces its limits on a [per user basis](https://cirrus-ci.org/faq/#are-there-any-limits), which means two PRs shouldn't block each other.

markus2330:
The highlevel tests should definitely work without any variable. To check if this works, is the purpose of this test.

kodebach:
> The highlevel tests should definitely work without any variable. To check if this works, is the purpose of this test.

I agree, the problem is that the Travis Linux build jobs use a non-standard `CMAKE_INSTALL_PREFIX`. But I will try to emulate whatever CMake does to make it work with pkg-config.

Also the reason this PR is taking so long is that previously external testcase was simply skipped, if pkg-config couldn't find Elektra, which was the case for a lot of our build jobs (the ones that are failing now). I made changes so that it is automatically skipped in `make run_all` but will fail in `kdb run_all` if pkg-config cannot find Elektra, because in this case we know it should be installed.

> canceling jobs on Travis, you do not need

This is by far the most important thing, because Travis doesn't cancel builds if you push a new commit. I think we should mention that somewhere prominent so everyone knows about it.

> I really think our Jenkins build system is severely broken in this regard.

Yes, the current situation is really bad. Is there any reason not to have the files for building the debian package on the master branch, so that each PR could modify them separately?

sanssecours:
> I agree, the problem is that the Travis Linux build jobs use a non-standard `CMAKE_INSTALL_PREFIX`.

If you want you can remove the custom `CMAKE_INSTALL_PREFIX`. We only specify the variable since the builds on Linux failed otherwise, if I remember correctly.

> This is by far the most important thing, because Travis doesn't cancel builds if you push a new commit.

Travis cancels build jobs, but [does so in an intelligent way](https://blog.travis-ci.com/2017-09-21-default-auto-cancellation). If one part of the build already started, it will not cancel the build, unlike our Jenkins server, which will cancel builds, even if they are 99% done. It would be even better, if Travis supported the feature requested in https://github.com/travis-ci/travis-ci/issues/2062 though.

kodebach:
> If you want you can remove the custom `CMAKE_INSTALL_PREFIX`. We only specify the variable since the builds on Linux failed otherwise, if I remember correctly.

Its actually a good test case, because it should work with `CMAKE_INSTALL_PREFIX`.

> Travis cancels build jobs, but does so in an intelligent way. 

As far as I can tell it only cancels whole builds not individual jobs. Yesterday I had to cancel the builds for 4 or 5 commits because the OSX builds were all still waiting.

> But I will try to emulate whatever CMake does to make it work with pkg-config.

Turns out that is not as easy as expected. CMake works arround the problem, by using `-Wl,-rpath`. While this would work with pkg-config as well, it is not very portable and not recommended. Which is why CMake [removes this option](https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling#default-rpath-settings) when installing. Although I think we circumvent that in our root `CMakeLists.txt`.

The last commit hopefully fixes the build, however the pkg-config examples shouldn't be used like this for any kind of release. I added a note to the READMEs so users know about the problems with `-Wl,-rpath`.

kodebach:
It looks like the only remaining problem is that Jenkins reports the build as "unstable". The Travis Builds should all go green soon. (At least they did for [my repo](https://travis-ci.com/kodebach/libelektra/builds/99278720))

@markus2330 Do we know how to fix that, or are we going to ignore it for now?

kodebach:
Both `🍏 FULL` and `🍏 MMap` timed out for the second time in a row now, so I didn't restart them again. Like I said they both ran successfully on my repository. Although one somehow took 54 min, maybe the new travis-ci.com doesn't enforce the timelimit as much.

markus2330:
Thank you a lot for these important fixes!

PR #2357, opened 2019-02-01
Augeas: Change Prefix for Lens Directory on macOS

sanssecours:
It seems `/usr/local` is the correct directory, which is also used by some Linux distributions according to the [documentation of the Augeas plugin](https://master.libelektra.org/src/plugins/augeas/README.md#mounting-and-configuration).

### DIFF
--- a/cmake/Modules/FindAugeas.cmake
+++ b/cmake/Modules/FindAugeas.cmake
@@ -30,7 +30,7 @@ else (LIBAUGEAS_INCLUDE_DIR)
 
 	if (NOT _LIBAUGEAS_PREFIX)
 		if (APPLE)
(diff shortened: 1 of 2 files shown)


markus2330:
There is no "correct" directory. It depends on where an administrator installed something. The purpose of pkg-config is that libs are found regardless of where they were installed.

sanssecours:
> There is no "correct" directory. 

It is the correct directory for the default Homebrew install 😊.

> The purpose of pkg-config is that libs are found regardless of where they were installed.

Not everyone uses `pkg-config`. Some library developers even decide to [not provide `.pc` files for their software](https://dev.gnupg.org/T2037).

markus2330:
Thank you!

Issue #2351, opened 2019-01-30
Jenkins: Build Job for `master` Branch Fails


sanssecours:
# Problem Description

I am not exactly sure why, but since one of the latest updates the [build job for the `master` branch is “unstable”](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/master/367/pipeline). As far as I can tell building the Debian package fails, although all subtasks of the stage “Build artifacts” seem to be successful.

markus2330:
Thank you for reporting this problem! @kodebach is working on this, see #2350

markus2330:
@BernhardDenner @ingwinlu Do you have an idea why this job is unstable?

I thought that unstable refers to failing test cases but this does not seem to be the case?

sanssecours:
Just a minor update: Currently the build for new pull requests is not unstable any more, but [fails printing the following error message](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2356/2/pipeline):

```
make[1]: Entering directory '/home/jenkins/workspace/libelektra_PR-2356-EN6GWJOR5VHPPOXXAYQ7ZKJOW5Z5GVNVM2BBC6LIIINSCUQMGGEQ/libelektra'
dh_install --list-missing
dh_install: Cannot find (any matches for) "usr/include/elektra/elektra/error.h" (tried in "." and "debian/tmp")
dh_install: libelektra-dev missing files: usr/include/elektra/elektra/error.h
dh_install: Cannot find (any matches for) "usr/include/elektra/elektra/errors.h" (tried in "." and "debian/tmp")
dh_install: libelektra-dev missing files: usr/include/elektra/elektra/errors.h
dh_install: Cannot find (any matches for) "usr/include/elektra/elektra/errorcodes.h" (tried in "." and "debian/tmp")
dh_install: libelektra-dev missing files: usr/include/elektra/elektra/errorcodes.h
dh_install: Cannot find (any matches for) "usr/include/elektra/elektra/types.h" (tried in "." and "debian/tmp")
dh_install: libelektra-dev missing files: usr/include/elektra/elektra/types.h
dh_install: Cannot find (any matches for) "usr/include/elektra/elektra/conversion.h" (tried in "." and "debian/tmp")
dh_install: libelektra-dev missing files: usr/include/elektra/elektra/conversion.h
[...]

PR #2359, opened 2019-02-01
Sanitizer: Update Blacklist for GNU C++ Library 8

sanssecours:


markus2330:
Thank you so much! Is this related to what I just reported in #2351?

sanssecours:
> Is this related to what I just reported in #2351?

Yep. I assume the Docker images were for Debian Unstable were upgraded recently and now use a newer version of `libstdc++`.

markus2330:
Thank you for doing this difficult search!

Can you merge this, so that master builds again? The commit about doc publishing can be reverted when #2351 is fixed.

PR #2362, opened 2019-02-03
CMake: Silence Warnings About Missing `pkg-config` Executable

sanssecours:


src/plugins/zeromqrecv/CMakeLists.txt
```diff
@@ -2,7 +2,9 @@ include (LibAddBinding)
 
 if (DEPENDENCY_PHASE)
 	find_package (PkgConfig QUIET)
-	pkg_check_modules (ZMQ QUIET libzmq)
+	if (PKG_CONFIG_FOUND)
```
markus2330: Do we get correct error messages after this PR? Suppose someone has zmq installed but not pkg-config. Correct behavior should be that the error message says that `pkg-config is missing but needed to find zmq`. Is this the case?

But in general I like that the output gets more quiet.

sanssecours:
I just looked at the updates in this pull request again. Since most of the updated code will be overwritten by changes from other PRs by me, I decided to close this.

Issue #2310, opened 2019-01-09
testmod_dbusrecv: sporadic fails


mpranj:
Confirmed sporadic failure on [travis builds](https://travis-ci.org/ElektraInitiative/libelektra/jobs/477332555).

## Log 1

Full log: [log.txt](https://github.com/ElektraInitiative/libelektra/files/2741411/log.txt)

```
 74/225 Test  #74: testmod_dbusrecv .........................***Failed    0.04 sec
DBUSRECV TESTS
==============
testing prerequisites
detecting available bus types - please ignore single error messages prefixed with "connect:"
connect: Failed to open connection to system message bus: Failed to connect to socket /usr/local/var/run/dbus/system_bus_socket: No such file or directory
test commit
../src/plugins/dbusrecv/testmod_dbusrecv.c:180: error in test_commit: string "system/tests/testmod_dbusrecv/changed" is not equal to "user/tests/foo/bar"
test adding keys
../src/plugins/dbusrecv/testmod_dbusrecv.c:228: error in test_keyAdded: string "system/tests/testmod_dbusrecv/added" is not equal to "user/tests/foo"
test adding keys
../src/plugins/dbusrecv/testmod_dbusrecv.c:276: error in test_keyChanged: string "system/tests/testmod_dbusrecv/changed" is not equal to "user/tests/foo/bar"
testmod_dbusrecv Results: 34 Tests done — 3 errors.
```

## Log 2

```
 74/225 Test  #74: testmod_dbusrecv .........................***Failed    0.04 sec
DBUSRECV TESTS
==============
testing prerequisites
detecting available bus types - please ignore single error messages prefixed with "connect:"
connect: Failed to open connection to system message bus: Failed to connect to socket /usr/local/var/run/dbus/system_bus_socket: No such file or directory
test commit
../src/plugins/dbusrecv/testmod_dbusrecv.c:180: error in test_commit: string "system/tests/testmod_dbusrecv/changed" is not equal to "user/tests/foo/bar"
test adding keys
../src/plugins/dbusrecv/testmod_dbusrecv.c:228: error in test_keyAdded: string "system/tests/testmod_dbusrecv/added" is not equal to "user/tests/foo"
test adding keys
../src/plugins/dbusrecv/testmod_dbusrecv.c:276: error in test_keyChanged: string "system/tests/testmod_dbusrecv/changed" is not equal to "user/tests/foo/bar"
testmod_dbusrecv Results: 34 Tests done — 3 errors.
[...]

PR #2364, opened 2019-02-04
highlevel: fix linker flags

### LINKED ISSUE #2363: Linker flags for highlevel API
pkg-config and also `examples/highlevel/README.md` recommend to link against `elektra` but this library is deprecated. (It contains core+kdb for legacy applications that existed before our library split.)

Furthermore, the order of the linker flags seem to be incorrect, they currently are `-lelektra-kdb -lelektra-ease -lelektra-highlevel` but highlevel should be first, and then the libs to provide the symbols for highlevel.

So it is something like `highlevel kdb ease core`?

The "cmake" example does not have these problems.

kodebach:
Closes #2363 

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -74,7 +74,7 @@ types and functions have to be mapped to provide the full functionality.
 
 Take a look at the [README](/src/libs/highlevel/README.md) for more infos.
 
-For examples on how build an application using this API take a look at our [example](/examples/highlevel). *(Klemens Böswirth)*
+For examples on how build an application using this API take a look at our new [example](/examples/highlevel). *(Klemens Böswirth)*
 
 
 ### <<HIGHLIGHT2>>
--- a/examples/highlevel/README.md
+++ b/examples/highlevel/README.md
@@ -4,7 +4,18 @@ This folder contains an example on how to use the high-level API.
 
 The example is provided for CMake and pkg-config build systems, but you can use any build system you like, as long as you setup
 your include directories and linked libraries correctly. The high-level API uses the same include directory as the rest of elektra,
-and you need to link against at least `elektra-highlevel`, `elektra-kdb` and `elektra-ease` (or, alternatively, `elektra-full`).
+and you need to link against at least `elektra-core`, `elektra-highlevel`, `elektra-kdb` and `elektra-ease`.
+
+You should therefore use one of the following lines of linker flags:
+```
+-lelektra-highlevel -lelektra-kdb -lelektra-ease -lelektra-core
+
+# or if Elektra was compiled with BUILD_FULL
+-lelektra-full
+
+# or if Elektra was compiled with BUILD_STATIC (see Limitations below)
+-lelektra-static
+```
 
 ## Setup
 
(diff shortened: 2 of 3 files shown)


markus2330:
It seems like that also LIBRARIES in ./src/libs/highlevel/CMakeLists.txt has the same problem. So #2364 is not fixed by this PR.

kodebach:
The Travis build `🍏 FULL` timed out while building the caches, but run all tests successfully, so I'm marking this as "ready to merge".

PS. I think we should investigate speeding up macOS builds, they are a real bottleneck.

kodebach:
> It seems like that also LIBRARIES in ./src/libs/highlevel/CMakeLists.txt has the same problem.

Not sure what you mean... I changed `LIBRARIES` in this PR so that it uses the correct order of libraries. Although afaik the order should only matter for static libraries.

markus2330:
You are right, #2364 is fixed here.

I was confused by a different problem, which also exists also in the main elektra.pc file, see #2367

kodebach:
> It would be really nice, if someone could look into this and fix the [timeout problems](https://github.com/ElektraInitiative/libelektra/issues/2310) properly.

I just took a quick look and found a few things:
- we are installing packages using `before_install` (why not `install`?), maybe the `apt` and `homebrew` [addons](https://docs.travis-ci.com/user/installing-dependencies/#installing-dependencies-on-multiple-operating-systems) are quicker in some way
- We should definitely cache `/home/travis/.rvm` (or the macOS equivalent if that is different). According to [Travis](https://docs.travis-ci.com/user/caching/) using `rvm install` might take a few minutes.
- Caching the Homebrew directory doesn't seem to speed up `brew install` because our jobs still take 10min+ for installing.
- Using `ninja -j2` or similar might be a good idea  (for build, not necessarily for testing), but I don't know what the best value for the Travis setup would be.
- If all that doesn't help, `ccache` might be worth a try.

sanssecours:
> - we are installing packages using `before_install` (why not `install`?)

I assume we use `before_install`, since the [Travis documentation](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-standard-infrastructure) states that you should install dependencies in this step.

> maybe the `apt` and `homebrew` [addons](https://docs.travis-ci.com/user/installing-dependencies/#installing-dependencies-on-multiple-operating-systems) are quicker in some way

According to the Travis documentation using the Homebrew addon is the preferred way to install macOS dependencies. It might make sense to switch to the new system.

> - We should definitely cache `/home/travis/.rvm` (or the macOS equivalent if that is different). According to [Travis](https://docs.travis-ci.com/user/caching/) using `rvm install` might take a few minutes.

Sounds like a good idea.

> - Caching the Homebrew directory doesn't seem to speed up `brew install` because our jobs still take 10min+ for installing.

Caching the Homebrew installation cache is mainly a workaround against Homebrew starting to build packages from source, if it was unable to download a bottle (binary package) for a formula. I am not sure, if removing this cache would decrease the build time, but it might be worth a try.

> - Using `ninja -j2` or similar might be a good idea  (for build, not necessarily for testing), but I don't know what the best value for the Travis setup would be.

Ninja should (at least in theory) already utilize a multi-core system properly:

```sh
ninja -h
#> ...
#> -j N     run N jobs in parallel [default=10, derived from CPUs available]
#> ...
```

.

> EDIT: I thought you meant the macOS timeout problems on Travis...

Nope 😊, I meant the timeout problems of the notification bindings and plugins. Sorry for not pointing that out clearly. These timeouts happen on all of our build systems (Travis, Jenkins, Cirrus), which makes them especially annoying. Thanks for your input about speeding up the Travis build though 💖.

markus2330:
Thank you for the interesting discussions. If possible, please move discussions to the corresponding issues, or at least link to it. If broad issues like #160 do not fit, please create new issues.

markus2330:
> I'm marking this as "ready to merge".

I assume this should be ready to merge. Now also all builds were successful. Thank you!

PR #2375, opened 2019-02-06
highlevel: rename external tests

### LINKED ISSUE #2372: kdb run_all does not execute external tests

Install Elektra and run:

```sh
kdb stash
kdb run_all
```

## Expected Result

It runs all test cases.

## Actual Result

```
Running check_spec
Running external_check*

The command kdb-full external_check* is not known
[...]
error: external_check*


Following test cases failed: 
testmod_lua
testmod_network
testmod_yaypeg
external_check*

run_all RESULTS: 146 test(s) done 4 error(s).
```

## System Information

- Elektra Version: master
- Debian Stretch

## Further Log Files and Output

`kdb list-tools | grep external`

has no output, so it seems like the external_check is not installed (is not part of the Debian package).

So there are actually two problems:

- [ ] that the external test case is not installed
- [ ] the failures in the case if external test cases are not installed

Why not simply naming it also check_external_... Then e63db7da06154094130dd0c33a56715ceb4f0122 would not have been necessary, which would avoid the second problem.

kodebach:
Closes #2372 

This should fix both problems mentioned in the issue without requiring change to the debian branch.


kodebach:
Apart from timeouts (travis and zeromqsend) everything seems to be fine...

markus2330:

markus2330:
Thank you for the quick fix!

Issue #2371, opened 2019-02-06
yaypeg test fails


markus2330:
## Steps to Reproduce the Problem

Install Elektra and run:

```sh
kdb stash
kdb testmod_yaypeg
```

## Expected Result

Success

## Actual Result

```
Running testmod_yaypeg

[==========] Running 5 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 5 tests from yaypeg
[ RUN      ] yaypeg.basics
/home/jenkins/workspace/libelektra_master-Q2SIBK3KE2NBEMJ4WVGJXAXCSCB77DUBUULVLZDKHQEV3WNDXBMA/libelektra/src/plugins/yaypeg/testmod_yaypeg.cpp:73: Failure
Value of: plugin != __null
  Actual: false
Expected: true
Could not open yaypeg plugin
error: testmod_yaypeg
```

## System Information

- Elektra Version: master
- Debian Stretch

sanssecours:
I can not reproduce this bug on [macOS](https://travis-ci.org/ElektraInitiative/libelektra/jobs/489420151#L2851). On [Ubuntu](https://travis-ci.org/sanssecours/elektra/jobs/488906348#L2124) everything seems to be fine too. I tried the following, using the [Docker image for Debian Stretch](https://github.com/ElektraInitiative/libelektra/blob/master/scripts/docker/debian/stretch/Dockerfile), which also worked:

```sh
mkdir build
cd build
cmake -GNinja                                                  \
      -DPLUGINS=ALL                                            \
      -DBINDINGS=cpp                                           \
      -DCMAKE_INSTALL_PREFIX="/home/jenkins/workspace/install" \
      -DINSTALL_SYSTEM_FILES=OFF                               \
      ..
ninja install
cd /home/jenkins/workspace/install/bin
./kdb testmod_yaypeg
#> [==========] Running 5 tests from 1 test case.
#> [----------] Global test environment set-up.
#> [----------] 5 tests from yaypeg
#> [ RUN      ] yaypeg.basics
#> [       OK ] yaypeg.basics (6 ms)
#> [ RUN      ] yaypeg.empty
#> [       OK ] yaypeg.empty (10 ms)
#> [ RUN      ] yaypeg.scalar
#> [       OK ] yaypeg.scalar (22 ms)
#> [ RUN      ] yaypeg.list
#> [       OK ] yaypeg.list (14 ms)
#> [ RUN      ] yaypeg.map
[...]

PR #2376, opened 2019-02-06
Debian: Update Build Instructions for YAy PEG Plugin

sanssecours:
This update tries to address issue #2371. Markus, can you please check, if this PR really fixes the problem you reported?


markus2330:
Thank you! Unfortunately I cannot test it without merging it. But it looks fine.

PR #2377, opened 2019-02-06
Highlevel

markus2330:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

### DIFF
--- a/src/libs/highlevel/README.md
+++ b/src/libs/highlevel/README.md
@@ -30,15 +30,13 @@ if (elektra == NULL)
 }
 
 int myint = elektraGetLong (elektra, "myint");
+printf ("got long %d\n", myint);
 
-if (myint < 10)
+elektraSetBoolean (elektra, "mybool", true, &error);
+if (error != NULL)
 {
-	elektraSetBoolean (elektra, "smallint", true, &error);
-	if (error != NULL)
-	{
-		printf ("An error occurred: %s", elektraErrorDescription (error));
-		elektraErrorReset (&error);
-	}
+	printf ("An error occurred: %s", elektraErrorDescription (error));
+	elektraErrorReset (&error);
 }
 
 elektraClose (elektra);
@@ -137,6 +135,8 @@ Errors which do not originate inside the high-level API itself are wrapped into 
 `ELEKTRA_ERROR_CODE_LOW_LEVEL`. The high-level Error API provides methods (`elektraKDBError*`) to access the properties of the low-level
 error. You can also access the key to which the error was originally attached, as well as any possible low-level warnings.
 
+TODO
+
 ### Configuration
 
 Currently there is only one way to configure an `Elektra` instance:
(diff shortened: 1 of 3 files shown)


src/libs/highlevel/README.md
```diff
@@ -30,15 +30,13 @@ if (elektra == NULL)
 }
 
 int myint = elektraGetLong (elektra, "myint");
+printf ("got long %d\n", myint);
 
-if (myint < 10)
+elektraSetBoolean (elektra, "mybool", true, &error);
```
markus2330: The idea of this change is that we should not demonstrate dependencies between configs.

src/libs/highlevel/README.md
```diff
@@ -30,15 +30,13 @@ if (elektra == NULL)
 }
 
 int myint = elektraGetLong (elektra, "myint");
+printf ("got long %d\n", myint);
```
markus2330: is "Long" guaranteed to be an int?

src/libs/highlevel/README.md
```diff
@@ -153,8 +153,9 @@ and therefore has no specified default value.
 The handler will also be called whenever you pass `NULL` where a function expects an `ElektraError **`. In this case the error code will be
 `ELEKTRA_ERROR_CODE_NULL_ERROR`.
 
-The default callback simply logs the error with `ELEKTRA_LOG_DEBUG` and then calls `exit()` with the error code of the error. It is highly
-recommended you either use `atexit()` in you application or set a custom callback, to make sure you won't leak memory.
```
markus2330: At exit memory is cleaned up anyway? Why this recommendation?

src/libs/highlevel/README.md
```diff
@@ -165,6 +166,9 @@ The callback must interrupt the thread of execution in some way (e.g. by calling
 The API supports the following types, which are taken from the CORBA specification:
 
 * **String**: a string of characters, represented by `KDB_TYPE_STRING` in metadata
+
+TODO: what does that mean?
```
markus2330: What does "`KDB_TYPE_STRING` in metadata" mean?

src/libs/highlevel/README.md
kodebach: "Long" is guaranteed to be 32 bits that should fit into an int on most modern systems. But you are right using `kdb_long_t` would be safer and better for the example.

src/libs/highlevel/README.md
kodebach: "won't leak memory" should be "dispose of all resources". The application might e.g. use sockets that need to be closed. Even if we assume the user knows that, we should still mention the the default callback calls `exit()`.

src/libs/highlevel/README.md
kodebach: We could be a bit more clear on that yes. We should add something along the lines of:

The API uses the `type` metakey to distinguish types. For each of the following types the corresponding `KDB_TYPE_*` macros defines the value that Elektra looks for in this metakey.

We should probably also insert the values of the macros into the README, i.e.:

* **String**: a string of characters, represented by `KDB_TYPE_STRING` (`"string"`) in metadata

markus2330: Why would a user need these macros?

kodebach: Most users won't, but if they use `elektraGetType` or `elektraSetRawString` they might want to use `KDB_TYPE_*` (they are actually constants not macros) instead of raw strings to prevent typos and enable IDE code completion.

It would probably suffice to mention once, that a `KDB_TYPE_*` constant exists for each type and then just use the string values in the descriptions.

src/libs/highlevel/README.md
```diff
@@ -271,11 +272,12 @@ typedef enum { A, B, C } MyEnum;
...
 
-// Read raw int value
-int value = elektraGetEnumInt (elektra, "message");
+// Read enum
 MyEnum enumValue = elektraGetEnum(elektra, "message", MyEnum);
 
-// enumValue == (MyEnum) value
+// Alternative: Read raw int value and then cast
+int value = elektraGetEnumInt (elektra, "message");
+enumValue == (MyEnum) value
```
kodebach: shouldn't that be a comment, so that the example is valid C code?

markus2330: Thank you, can you please do it? (Is now in master)

src/libs/highlevel/README.md
markus2330: Thank you, can you please do it? (Is now in master)

PR #2387, opened 2019-02-09
 YAy PEG: Fix Segfault Caused by Empty Input

sanssecours:


markus2330:
Thank you for fixing this problem!

PR #2390, opened 2019-02-09
YAwn: Fix Comment Handling & Improve Error Messages

sanssecours:


src/plugins/yawn/README.md
```diff
@@ -69,6 +69,35 @@ kdb rm -r user/tests/yawn
...
+
+# Manually add some data
+printf -- ' - Brutus\n' >  `kdb file user/tests/yawn`
+# Add element with incorrect indentation
+printf -- '- Burst'     >> `kdb file user/tests/yawn`
+
+# Try to retrieve data
+kdb ls user/tests/yawn
+# RET: 5
+# STDERR-REGEX: Reason: .*/config.yaml:2:1: Syntax error on token number 5: “<Token, SEQUENCE_START, SEQUENCE START, 2:1–2:1>”
```
markus2330: It is an improvement but we are still far from where we want to be. Ideally it should be something like "incorrect sequence, one space for indentation missing".

If the only sane way for reporting (regarding effort) is the output of grammar rules, maybe a link to the section of the YAML standard would already help?

markus2330:
Thank you!

PR #2391, opened 2019-02-09
Cirrus: Use Compiler Option `-Werror` on FreeBSD

sanssecours:


markus2330:
Wow, so great to see `-Werror` so widely adopted now.

Issue #1244, opened 2016-12-29
make uninstall does not uninstall everything


markus2330:
- [ ] fix leftovers
- [ ] add check to buildserver, see #160

```
DESTDIR=X make install -j 5
DESTDIR=X make uninstall -j 5 
find X
```

produces following leftovers:

```
X/usr
X/usr/share
X/usr/share/bash-completion
X/usr/share/bash-completion/completions
X/usr/share/zsh
X/usr/share/zsh/vendor-completions
X/usr/local
X/usr/local/lib
X/usr/local/lib/elektra
X/usr/local/lib/elektra/tool_exec
X/usr/local/lib/elektra/tool_exec/ffconfig
X/usr/local/lib/elektra/tool_exec/gen
X/usr/local/lib/elektra/libelektra-resolver.so
X/usr/local/lib/elektra/libelektra-storage.so
X/usr/local/lib/python2.7
X/usr/local/lib/python2.7/dist-packages
X/usr/local/lib/python2.7/dist-packages/elektra_gen-0.8.19.egg-info
X/usr/local/lib/python2.7/dist-packages/support
X/usr/local/lib/python2.7/dist-packages/support/util.pyc
X/usr/local/lib/python2.7/dist-packages/support/c.pyc
X/usr/local/lib/python2.7/dist-packages/support/genopt.py
X/usr/local/lib/python2.7/dist-packages/support/context.py
X/usr/local/lib/python2.7/dist-packages/support/cpp.pyc
X/usr/local/lib/python2.7/dist-packages/support/__init__.py
X/usr/local/lib/python2.7/dist-packages/support/nested.pyc
X/usr/local/lib/python2.7/dist-packages/support/kdb.pyc
X/usr/local/lib/python2.7/dist-packages/support/cpp.py
X/usr/local/lib/python2.7/dist-packages/support/gen.pyc
X/usr/local/lib/python2.7/dist-packages/support/util.py
X/usr/local/lib/python2.7/dist-packages/support/kdb.py
X/usr/local/lib/python2.7/dist-packages/support/__init__.pyc
X/usr/local/lib/python2.7/dist-packages/support/nested.py
X/usr/local/lib/python2.7/dist-packages/support/c.py
X/usr/local/lib/python2.7/dist-packages/support/context.pyc
X/usr/local/lib/python2.7/dist-packages/support/genopt.pyc
X/usr/local/lib/python2.7/dist-packages/support/gen.py
X/usr/local/lib/cmake
[...]

PR #2395, opened 2019-02-10
Cirrus: Make Sure Uninstall Procedure Removes All Installed Files on macOS

sanssecours:
**See also:** [Issue #1244](https://github.com/ElektraInitiative/libelektra/issues/1244)


markus2330:
Thank you so much! It is great to have a mechanism which will avoid regressions in the future.

Issue #2374, opened 2019-02-06
examples/highlevel/README.md not working


markus2330:
## Steps to Reproduce the Problem

Run commands as described in examples/highlevel/README.md

```sh
kdb stash
cd examples/highlevel
sudo kdb mount spec.ini 'spec/sw/example/highlevel/#0/current' ni
sudo kdb import spec/sw/example/highlevel/#0/current ni < spec.ini
sudo kdb spec-mount '/sw/example/highlevel/#0/current'
pkgconfig/application
```

## Expected Result

Successful execution of application.

## Actual Result

```
ERROR: The key '/sw/example/highlevel/#0/current/mystring' has the wrong type (expected 'string' but got '(null)').
```

## System Information

- Elektra Version: master

## Further Log Files and Output

```sh
kdb getmeta /sw/example/highlevel/#0/current/mystring type
#> Metakey not found
kdb getmeta spec/sw/example/highlevel/#0/current/mystring type
#> string
kdb global-mount 
#>
kdb export /sw/example/highlevel/#0/current ni
```

```ini
;Ni1
; Generated by Nickel Plugin using Elektra (see libelektra.org).

myint = 
myfloatarray/# = 
print = 1
mystring = 
mydouble = 
 = 

[myint]
 default = 0
 type = long

[myfloatarray/#]
 default = 0.0
 type = float

[print]
 default = 0
 type = boolean

[mystring]
 default = 
 type = string

[mydouble]
 default = 0.0
 type = double

[]
 mountpoint = highlevelexamples.conf
```

```sh
kdb mount
#> highlevelexamples.conf on /sw/example/highlevel/#0/current with name /sw/example/highlevel/#0/current
#> spec.ini on spec/sw/example/highlevel/#0/current with name spec/sw/example/highlevel/#0/current
```

kodebach:
I can't reproduce this, the only thing I noticed is that `kdb stash` might be broken.
After executing `kdb stash`, `kdb file` still reported the same file as before and `kdb get` produced the same value as before, so the kdb is clearly not in a clean state. Also it removes things like `system/info/elektra/constants` that should clearly always be present.

The only reason for this bug I can think is, that your shell treated the second part of the spec-mount line as a comment, because of the `#` (see syntax highlighting below).

```sh
sudo kdb import spec/sw/example/highlevel/#0/current ni < spec.ini
```

I changed the line to use `'...'` in #2375 to be on the safe side.

markus2330:
[...]

PR #2399, opened 2019-02-11
Highlevel API fixes

### LINKED ISSUE #2381: fix TODOs in src/libs/highlevel/README.md
I added TODOs in the PR #2377 please fix them.

kodebach:
Closes #2381 

Also fixes error message as promised in [#2374](https://github.com/ElektraInitiative/libelektra/issues/2374#issuecomment-462034159)


src/libs/highlevel/README.md
```diff
@@ -49,8 +49,6 @@ kdb setmeta /sw/org/myapp/#0/current/myint type long
 kdb setmeta /sw/org/myapp/#0/current/myint default 5
 ```
 
-TODO: Running application does not work?
```
kodebach: Couldn't find any problems...

src/libs/highlevel/README.md
```diff
@@ -166,32 +164,32 @@ The default callback simply logs the error with `ELEKTRA_LOG_DEBUG` and then cal
 
 TODO: why DEBUG? It should be ERROR? And we should enable logging by default, otherwise users will not see the message?
```
kodebach: The idea was, that most developers would provide their own fatal error handler, because they probably have their own log format and might need to do some cleanup before `exit()`. If we enable logging by default, we need to provide a way to disable it without using any low-level code.

kodebach: I couldn't add a comment in line 145:
```
TODO: explanation of low-level code or removal of this feature for this release?
```

What is that supposed to mean?

kodebach:
@markus2330 please review the changes and comments

src/libs/highlevel/README.md
markus2330: Then simply remove TODO (and add sudo).

src/libs/highlevel/README.md
markus2330: Then please document that it writing an fatal error handler is basically expected.

markus2330: low-level error code APIs are imho not properly explained. You have two options:

1. remove the APIs
2. properly explain them

markus2330:
Sorry for letting you wait! Please reping if something is urgent.

src/libs/highlevel/README.md
```diff
@@ -142,7 +142,31 @@ Errors which do not originate inside the high-level API itself are wrapped into
 `ELEKTRA_ERROR_CODE_LOW_LEVEL`. The high-level Error API provides methods (`elektraKDBError*`) to access the properties of the low-level
 error. You can also access the key to which the error was originally attached, as well as any possible low-level warnings.
 
-TODO: explanation of low-level code or removal of this feature for this release?
+To get the original low-level error code, description, severity, group, module and reason you can use these functions:
+```c
+int elektraKDBErrorCode (const ElektraError * error);
```
markus2330: @Piankero Do you want to change anything of these functions in your new error concept?

src/libs/highlevel/README.md
```diff
@@ -142,7 +142,31 @@ Errors which do not originate inside the high-level API itself are wrapped into
...
+ElektraKDBErrorGroup elektraKDBErrorGroup (const ElektraError * error);
+ElektraKDBErrorModule elektraKDBErrorModule (const ElektraError * error);
+const char * elektraKDBErrorReason (const ElektraError * error);
+```
+
+To iterate over all the warnings use the following to functions:
+
+```c
+int elektraKDBErrorWarningCount (const ElektraError * error);
+ElektraError * elektraKDBErrorGetWarning (const ElektraError * error, int index);
```
markus2330: Do we always use int for index?

src/libs/highlevel/README.md
```diff
@@ -142,7 +142,31 @@ Errors which do not originate inside the high-level API itself are wrapped into
...
 
-TODO: explanation of low-level code or removal of this feature for this release?
+To get the original low-level error code, description, severity, group, module and reason you can use these functions:
+```c
+int elektraKDBErrorCode (const ElektraError * error);
+const char * elektraKDBErrorDescription (const ElektraError * error);
+ElektraErrorSeverity elektraKDBErrorSeverity (const ElektraError * error);
+ElektraKDBErrorGroup elektraKDBErrorGroup (const ElektraError * error);
+ElektraKDBErrorModule elektraKDBErrorModule (const ElektraError * error);
+const char * elektraKDBErrorReason (const ElektraError * error);
```
markus2330: Do we already know if we will merge the error's reason and description?

src/libs/highlevel/README.md
```diff
@@ -142,7 +142,31 @@ Errors which do not originate inside the high-level API itself are wrapped into
 `ELEKTRA_ERROR_CODE_LOW_LEVEL`. The high-level Error API provides methods (`elektraKDBError*`) to access the properties of the low-level
 error. You can also access the key to which the error was originally attached, as well as any possible low-level warnings.
 
-TODO: explanation of low-level code or removal of this feature for this release?
+To get the original low-level error code, description, severity, group, module and reason you can use these functions:
```
markus2330: @kodebach Is this a good design to split the accessors functions of low-level errors and the other errors? I suppose people also want to get the errorcode, severity, ... for high-level error codes?

src/libs/highlevel/README.md
```diff
@@ -191,6 +216,9 @@ to use these functions directly, but they might still be useful sometimes (e.g.
 `elektraGetRawString`). We also provide a `KDB_TPYE_*` constant for each of the types listed above. Again, most users won't use these
 but, if you ever do need to use the raw type metadata using constants enables code completion and protects against typos.
 
+There is also the type `enum` with constant `KDB_TYPE_ENUM`. It is currently neither used nor supported by this API. However, we reserve it
+for a future expansion of this API.
```
markus2330: But it is possible to get the string, isn't it?

src/libs/highlevel/README.md
Piankero: Please give me some time to work out a concept and gather your feedback. I'll try to finish the concept until next weekend at the latest.

kodebach:
@markus2330 please review again

src/libs/highlevel/README.md
```diff
@@ -159,39 +181,43 @@ the most common use case for this callback is inside of functions that do not ta
...
-
-* **String**: a string of characters, represented by `KDB_TYPE_STRING` in metadata
+The API determines the data type of a given key, by reading its `type` metadata. The API supports the following types, 
+which are taken from the CORBA specification:
 
-TODO: please add better explanation, as started in PR #2377
-
-* **Boolean**: a boolean value `true` or `false`, represented by `KDB_TYPE_BOOLEAN` in metadata, in the KDB the raw value `"1"` is
+* **String**: a string of characters, represented by `string` in metadata
+* **Boolean**: a boolean value `true` or `false`, represented by `boolean` in metadata, in the KDB the raw value `"1"` is
```
markus2330: This is not according Elektra's boolean definition. See doc/decisions/bool.md and src/plugins/boolean/README.md

src/libs/highlevel/README.md
```diff
@@ -159,39 +181,43 @@ the most common use case for this callback is inside of functions that do not ta
...
+callback is only viable for very simple applications, because you won't get any indication as to which key caused the error (unless you
+compiled Elektra with debug logging enabled).
 
 <a name="data-types"></a>
 ## Data Types
 
-The API supports the following types, which are taken from the CORBA specification:
-
-* **String**: a string of characters, represented by `KDB_TYPE_STRING` in metadata
+The API determines the data type of a given key, by reading its `type` metadata. The API supports the following types,
```
markus2330: First introduce what metadata and specification is?

src/libs/highlevel/README.md
```diff
@@ -159,39 +181,43 @@ the most common use case for this callback is inside of functions that do not ta
...
-* **(Unsigned) Long Long**: a 64-bit (unsigned) integer, represented by `KDB_TYPE_LONG_LONG` (`KDB_TYPE_UNSIGNED_LONG_LONG`) in metadata
-* **Float**: whatever your compiler treats as `float`, probably IEEE-754 single-precision, represented by `KDB_TYPE_FLOAT` in metadata
-* **Double**: whatever your compiler treats as `double`, probably IEEE-754 double-precision, represented by `KDB_TYPE_DOUBLE` in metadata
-* **Long Double**: whatever your compiler treats as `long double`, not always available, represented by `KDB_TYPE_LONG_DOUBLE` in metadata
+* **Char**: a single character, represented by `char` in metadata
+* **Octet**: a single byte, represented by `octet` in metadata
+* **(Unsigned) Short**: a 16-bit (unsigned) integer, represented by `short` (`unsigned_short`) in metadata
+* **(Unsigned) Long**: a 32-bit (unsigned) integer, represented by `long` (`unsigned_long`) in metadata
+* **(Unsigned) Long Long**: a 64-bit (unsigned) integer, represented by `long_long` (`unsigned_long_long`) in metadata
+* **Float**: whatever your compiler treats as `float`, probably IEEE-754 single-precision, represented by `float` in metadata
```
markus2330: probably? Is CORBA so lax here?

src/libs/highlevel/README.md
```diff
@@ -159,39 +181,43 @@ the most common use case for this callback is inside of functions that do not ta
...
+* **Double**: whatever your compiler treats as `double`, probably IEEE-754 double-precision, represented by `double` in metadata
+* **Long Double**: whatever your compiler treats as `long double`, not always available, represented by `long_double` in metadata
```
markus2330: We should be more strict. It should be 128-bit, otherwise it does not make much sense. Is there an unsigned variant from it?

src/libs/highlevel/README.md
kodebach: No, but in this case `int` is more than enough. The generated functions in `kdbError.h` only work for warnings with index 0-99. After that the functions overwrite the warning at index 0. Also the way the warnings array works means that at that point the warning count is also reset.

If we rework our error concept we should change that but for now I don't think it is a problem because generating more than 100 warnings should be very rare.

src/libs/highlevel/README.md
kodebach: You either have to set `type = string` and use `elektraGetString` or use `elektraGetRawString`. Both are compatible with the `enum` plugin because that only requires `check/enum`.

I will add a note about that.

src/libs/highlevel/README.md
kodebach: You are right, `elektraErrorCode` and `elektraErrorSeverity` exist too, but only `elektraErrorDescription` is mentioned in the README.md

src/libs/highlevel/README.md
kodebach: CORBA specifies IEEE-754 single/double precision for float and double. However, this description is based on the actual code in `kdbtypes.h`. The C standard does not define float/double using IEEE-754 and AFAIK there is no good way to check whether the current architecture uses IEEE-754. In any case, it does not really matter because even embedded systems nowadays use IEEE-754 (maybe some minor features are missing).

src/libs/highlevel/README.md
kodebach: > double-extended (a mantissa of at least 64 bits, a sign bit and an exponent of at least 15 bits) IEEE floating point numbers

Is what the CORBA specification has to say about `long double`. The C standard sadly doesn't provide any guarantees except for `long double` is not smaller than `double`.

> It should be 128-bit

I disagree. Firstly, we should try to keep inline with CORBA spec. Second x86 (or more specifically x87) supports the double-extended type described above natively. (see [Wikipedia](https://en.wikipedia.org/wiki/Long_double)) And last but not least, without an external library only `__float128` would be possible which supported by GCC but clang seems to have problems in some versions.

Also, it seems apart from Microsoft Visual C++ every compiler uses the 80-bit extended type (or something bigger) for `long double`. (see also [Wikipedia](https://en.wikipedia.org/wiki/Quadruple-precision_floating-point_format#Implementations))

> Is there an unsigned variant from it?

An unsigned floating point type? How would that work? If on just wants to prohibit negative values they should use something like the `range` plugin.

If we were to add anything, I think the `fixed` fixed-point type is the most likely, but that would also either require a library or writing our own implementation.

src/libs/highlevel/README.md
markus2330: > If we rework our error concept

I would prefer if we now only publish a minimal error API, so that we will not have any compatibility problems when we rework the API. So I would say that we remove the low-level error API for now.

> because generating more than 100 warnings should be very rare.

That might be. We should be consistent with our types nevertheless.

src/libs/highlevel/README.md
markus2330: As said above, I think we should now publish a minimal API. We can extend it later easily.

src/libs/highlevel/README.md
markus2330: We could output an error for systems not defining __STDC_IEC_559__.

In any case: Elektra's types must be portable and well-specified, otherwise administrators cannot know which values they are allowed to use.

In this case we should write that Elektra uses IEEE-754 single/double.

src/libs/highlevel/README.md
markus2330: Sorry, I added the info to the wrong line. We cannot have any of these types to be compiler dependent. We want different applications to interpret configuration vales exactly the same way, otherwise we would simply move all problems to the applications.

About long double:

We already have defined it in src/include/kdbtypes.h line 123. But we also need to specify maximum numbers, so that we can be sure that any application can read what someone wrote into the KDB.

markus2330:
Do we already have test cases for all border cases of the data types? elektraSet<Type> should only be successful if elektraGet<Type> will also be successful for any compiler.

src/libs/highlevel/README.md
kodebach: > We could output an error for systems not defining **STDC_IEC_559**

Tried that already. Doesn't work on macOS. I think defining `STDC_IEC_559` is not mandatory, even if IEEE-754 is used.

> In this case we should write that Elektra uses IEEE-754 single/double.

I added checks that should ensure, all values allowed by CORBA (IEEE-754 single/double, 80 bit double-extended) can be stored properly. We cannot however guarantee that calculations with these values are done according to IEEE-754 (especially if Inf or NaN is involved). For that we would need much more complicated checks.

src/libs/highlevel/README.md
kodebach: > We already have defined it in src/include/kdbtypes.h line 123.

I made some changes to `kdbtypes.h` in one of my last commits to do some more in depth checks.

src/libs/highlevel/README.md
markus2330: Our goal is not to check if implementations are standard-conform. We "only" need to guarantee that what somewhat wrote into KDB, is identical to what someone will get when reading.

kodebach: > Our goal is not to check if implementations are standard-conform.

I am not checking the compiler implementation, but rather what is used as `float`. This mostly depends on the underlying architecture of the machine. Although the compiler of course decides which available type is used. 

> We "only" need to guarantee that what somewhat wrote into KDB, is identical to what someone will get when reading.

The way floating point numbers work implies there is always a maximum number of decimal digits that can survive a `float -> string -> float` conversion (`std::numeric_limits<T>::digits10` in C++) [1]. Additionally there is a minimum number of decimal digits required so that a `float -> string -> float` always produces the original number (`std::numeric_limits<T>::max_digits10` in C++) [2]. Neither of these guarantee however that the intermediate string is the exact value of the floating point number. That might require hundreds of digits [2].

There are also values like `0.1` that simply have no representation as a (finite-precision binary) floating point number. Therefore `kdb set /some/float 0.1` followed by a call to `elektraGetFloat` will never result in the exact value of `0.1`.

I made some changes to our conversion functions and added some checks that should ensure a floating point value will stay the same through arbitrarily many cycles of read and write by KDB (independent of architectures, starting with a write because of the 0.1 problem) [3,4,5].

[cf6232e](/ElektraInitiative/libelektra/pull/2399/commits/cf6232e9dec2b4c04056f7d5ba37c34de1189c5c) ensures the architecture uses IEEE-754 compatible (i.e. all values can be stored) floating point types.

[2f9076e](/ElektraInitiative/libelektra/pull/2399/commits/2f9076e10ca8d8e456b803bb387813339b13cc5c) changes the conversion functions so that they use 9, 17 and 21 digits of precision, which is enough for IEEE-754 single, double and double-extended precision.

If you want to be even more strict with the checks, please suggest concrete changes.

---
Sources:
[1] [cppreference](https://en.cppreference.com/w/cpp/types/numeric_limits/digits10)
[2] [cppreference](https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10)
[3] [stack overflow](https://stackoverflow.com/a/19897395)
[4] [stack overflow](https://stackoverflow.com/a/21162120)
[5] [C++ standard proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2006/n2005.pdf)

markus2330: Thank you for the added tests! It would be good if we had that information also in our docu. Maybe you can directly put the information in ./src/include/kdbtypes.h?

I was also thinking about the integer functions, which should be much easier to test than floating point. Do you have tests for their limits?

kodebach:
> Do we already have test cases for all border cases of the data types? elektraSet should only be successful if elektraGet will also be successful for any compiler.

Testing `elektraGet/Set` extensively doesn't make much sense because they rely on the conversion functions from `elektra/conversion.h`. If anything these functions should be tested. Currently we don't have these tests, because the `elektraTYPEToString` functions rely on the `printf` family of functions, while `elektraKeyToTYPE` functions rely on `strtol`, `strtoull`, `strtof` etc. These functions should be compatible with each other according to the C Standard. The success of `elektraGet` should therefore not be compiler dependent. The *result*  of `elektraGetFloat` et al. is however implementation dependent, if the key contains a string that is not exactly representable as the chosen type. The C standard only defines that the result of `strtof` et al. must in this case be one of the two closes representable numbers.

markus2330:
> Testing elektraGet/Set extensively doesn't make much sense because they rely on the conversion functions from elektra/conversion.h. If anything these functions should be tested.

High-level tests have their purpose, even if their functionality is already covered in low-level unit tests.

>  The result of elektraGetFloat et al. is however implementation dependent, if the key contains a string that is not exactly representable as the chosen type. 

Yes, bit-identical serialization of float is not trivial. For now, I would be happy if we have tests for the other types.

src/libs/highlevel/README.md
kodebach: For C99+ we use the `(u)intXX_t` types so testing is unnecessary.

For older compilers we use `short` (guaranteed to be at least 16 bits) for `kdb_short_t`. For `kdb_long_t` and `kdb_long_long_t` we do size checks for `int`, `long` and `long long` to find types of the correct size. If no type is found an `#error` is issued.

kodebach:
jenkins build all please

markus2330:
Thank you for the great work!

Is there somewhere an issue to track the left-overs?

- some more tests would be nice (border cases of integral types)
- low-level error handling
- types for array indexes?

kodebach:
> * some more tests would be nice (border cases of integral types)

I added #2430.

> * low-level error handling
> * types for array indexes?

We will need to wait for @Piankero's concept, in order to know what needs to be changed.

markus2330:
We can simply *not release* everything which is unclear. (like the error and array functions).

kodebach:
The array functions should be fine. We consistently use `kdb_long_long_t` for indices apart from the error access functions, which I don't want to change until we know the plan for the future.

Also we need to release at least the error access functions for highlevel errors and a function that provides access to the error Key, in case of a low-level error. Otherwise proper error reporting isn't possible.

markus2330:
> The array functions should be fine. We consistently use kdb_long_long_t for indices apart from the error access functions, which I don't want to change until we know the plan for the future.

Then let us remove this error access function for now. We definitely do not want it to be inconsistent.

> Also we need to release at least the error access functions for highlevel errors and a function that provides access to the error Key, in case of a low-level error. Otherwise proper error reporting isn't possible.

Yes, we only need elektraErrorDescription and elektraErrorReset for now, everything else can be removed for now.

Can you create an issue?

PR #2401, opened 2019-02-11
Tests: Various Minor Improvements 

sanssecours:


src/plugins/yamlcpp/README.md
```diff
@@ -74,16 +74,16 @@ sudo kdb umount /tests/yamlcpp
 YAML CPP provides basic support for Elektra’s array data type.
 
 ```sh
-# Mount yamlcpp plugin to cascading namespace `/tests/yamlcpp`
-sudo kdb mount config.yaml /tests/yamlcpp yamlcpp
+# Mount yamlcpp plugin to `user/tests/yamlcpp`
+sudo kdb mount config.yaml user/tests/yamlcpp yamlcpp
```
markus2330: Why to use user instead of cascading?

markus2330:
Thank you!

src/plugins/yamlcpp/README.md
sanssecours: I used the explicit namespace, since otherwise the output of the command `kdb ls` in line `86` would depend on the (root or non-root) user that executes the test.

markus2330: Thanks for the tip! Maybe we should write that somewhere in the shell recorder tutorial?

sanssecours: I do not think that is necessary. The output of failing Markdown Shell Recorder tests on FreeBSD should already show the issue to someone who might not expect that behavior.

PR #2402, opened 2019-02-11
Libopts improvements

### LINKED ISSUE #2394: command-line-options tutorial
- [ ] Link in "You can find a full working example here" broken (very end)
- [ ] No examples of env

In  #2383 @kodebach wrote:
> Currently the only way environment variables can be configured is via the key name. If it ends with /# the environment variable is split into an array. Technically one could therefore reuse the same environment variable for multiple keys, but for consistency's sake and in case we add configuration options in future it is also forbidden.

I think using a different env-name for a key-name could be useful:

```
[scratchpad]
env = TEMPDIR
```

But if it can be added later without compatibility issues, it is low-priority.

kodebach:
Closes #2394


markus2330:
Thank you!

PR #2429, opened 2019-02-18
Don't reformat gitignored files

kodebach:
A simple PR that changes our reformat scripts so they ignore files that are ignored by git.

Previously they for example also formatted all files in any cmake build directory (really slowed down reformat-cmake).


scripts/reformat-cmake
```diff
@@ -29,18 +29,9 @@ if [ -z "$(which sponge)" ]; then
...
-		-or -path './doc/*' \
-		-or -path './examples/*' \
-		-or -path './scripts/*' \
-		-or -path './src/*' \
-		-or -path './tests/*' \
-		-or -path './CMakeLists.txt' \
-		\)
-); do
-	$CMAKE_FORMAT "$file" | unexpand | sponge "$file"
+directories="benchmarks cmake doc examples scripts src tests"
```
markus2330: Why is directories needed? git-ls-files should already do the correct thing.

scripts/reformat-cmake
```diff
@@ -29,18 +29,9 @@ if [ -z "$(which sponge)" ]; then
...
+files=$(git ls-files $directories CMakeLists.txt | grep '.*CMakeLists.txt')
```
markus2330: The filter is wrong: we also want to reformat files that are called '*.cmake'

So `git ls-files '*.cmake' '*/CMakeLists.txt' ` should return the correct list of files.

scripts/reformat-shfmt
```diff
@@ -15,6 +15,6 @@ if ! which shfmt > /dev/null; then
 	exit 1
 fi
 
-for file in $(shfmt -f scripts tests/shell); do
-	shfmt -s -sr -w "$file"
-done
+shell_files=$(shfmt -f 'scripts' 'tests/shell')
+non_ignored=$(printf "%s\n" "$shell_files" | sed -nE 's/.*/"\0"/p' | xargs git check-ignore -vn | sed -ne 's/^::[ \t]*\(.*\)$/\1/p')
```
markus2330: `git check-ignore` has different semantics than `git ls-files`. Files which are not part of the git repo, but also not ignored, will yield true in `check-ignore` but not in `ls-files`.

scripts/reformat-cmake
kodebach: The old script only checked these directories, but it probably isn't required anymore.

scripts/reformat-shfmt
kodebach: I can't use `git ls-files` with `shfmt -f`. For some reason if you call `shfmt -f` with a filename (not a directory) it always prints that filename even if the file isn't a shell file. For example:
```sh
shfmt -f README.md
#> README.md
```

scripts/reformat-source
```diff
@@ -32,5 +32,5 @@ cd "$SOURCE" || {
 	exit 1
 }
 
-find . \( \( -name '*.[ch]' -or -name '*.[ch]pp' -or -name '*.[ch].in' \) -and -not -regex '^./src/tools/gen.*' \) \
-	-exec "$CLANG_FORMAT" -style=file -i {} +
+source_files=$(git ls-files '*.c' '*.h' '*.cpp' '*.hpp' '*.c.in' '*.h.in' | grep -v '^src/tools/gen.*')
+printf "%s\n" "$source_files" | sed -nE 's/.*/"\0"/p' | xargs "$CLANG_FORMAT" -style=file -i
```
sanssecours: Unfortunately this command does not work on macOS as far as I can tell:

```sh
reformat-source 2>&1 | grep 'No such file or directory' | wc -l | sed -E 's/[ ]+//g'
#> 1026
```

. The problem seems to be `sed -nE 's/.*/"\0"/p'`, which prints `"0"` 1026 😊 times on my machine. If you want to add surrounding `"` chars around each file name, then the following should work:

```
sed -nE 's/(.*)/"\1"/p'
``` 

. I **do not think** that this is a particular safe method to work with the filenames, since it does not handle special characters properly. For our needs it should be enough though, as long as we do not add any source files with names that need escaping.

kodebach: I now use `sed -nE 's/(.*)/'"'"'\1'"'"'/p'` which should surround filenames with `'`. That should be safer.

scripts/reformat-shfmt
```diff
@@ -15,6 +15,6 @@ if ! which shfmt > /dev/null; then
 	exit 1
 fi
 
-for file in $(shfmt -f scripts tests/shell); do
-	shfmt -s -sr -w "$file"
-done
+shell_files=$(shfmt -f 'scripts' 'tests/shell')
+non_ignored=$(printf "%s\n" "$shell_files" | sed -nE 's/(.*)/'"'"'\1'"'"'/p' | xargs git check-ignore -vn | sed -ne 's/^::[ \t]*\(.*\)$/\1/p')
```
sanssecours: Some versions of `sed` [do not support `\t`](https://stackoverflow.com/questions/2610115/why-is-sed-not-recognizing-t-as-a-tab). You can use the [POSIX character class `[:blank:]`](https://www.regular-expressions.info/posixbrackets.html) to solve this problem:

```sh
sed -E 's/^::[[:blank:]]*(.*)$/\1/'
```

.

scripts/reformat-cmake
```diff
@@ -29,18 +29,8 @@ if [ -z "$(which sponge)" ]; then
...
-		-or -path './doc/*' \
-		-or -path './examples/*' \
-		-or -path './scripts/*' \
-		-or -path './src/*' \
-		-or -path './tests/*' \
-		-or -path './CMakeLists.txt' \
-		\)
-); do
-	$CMAKE_FORMAT "$file" | unexpand | sponge "$file"
+files=$(git ls-files '*.cmake' '*/CMakeLists.txt' | grep '.*CMakeLists.txt')
```
markus2330: Why the grep?

markus2330:
@sanssecours Does everything work now? Can you please merge the PR if it does?

markus2330:
Thank you!

PR #2432, opened 2019-02-20
highlevel: remove most of error handling from public API

### LINKED ISSUE #2430: Highlevel API tests
The highlevel API currently only does some basic tests for each of the supported types.

It would be nice, if we added more tests to `testkdb_highlevel.cpp` (especially for border cases).
For example adding a test that checks that we accept `0` and `65535`, but not `-1` and `65536` for the type `unsigned_short`.

kodebach:
As discussed in #2399, this removes most of the current error access functions from the public API.

It also adds border tests for integer types and therefore closes #2430.

### DIFF
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
@@ -524,6 +524,7 @@ static ostream & printPrivate (ostream & os, parse_t & p)
 	   << endl
 	   << "#include <elektra/types.h>" << endl
 	   << "#include <elektra/error.h>" << endl
+	   << "#include <kdbprivate.h>" << endl
 	   << endl
 	   << "#ifdef __cplusplus" << endl
 	   << "extern \"C\" {" << endl
--- a/src/include/elektra/error.h
+++ b/src/include/elektra/error.h
@@ -20,34 +20,10 @@ extern "C" {
 
 typedef struct _ElektraError ElektraError;
 
-typedef enum
-{
-	/**
-	 * Use only, if the error will be raised with elektraFatalError().
-	 */
-	ELEKTRA_ERROR_SEVERITY_FATAL = 0,
-	ELEKTRA_ERROR_SEVERITY_ERROR,
-	ELEKTRA_ERROR_SEVERITY_WARNING
-} ElektraErrorSeverity;
-
-typedef const char * ElektraKDBErrorGroup;
-typedef const char * ElektraKDBErrorModule;
-
 typedef void (*ElektraErrorHandler) (ElektraError * error);
 
 ElektraErrorCode elektraErrorCode (const ElektraError * error);
 const char * elektraErrorDescription (const ElektraError * error);
-ElektraErrorSeverity elektraErrorSeverity (const ElektraError * error);
-
-int elektraKDBErrorCode (const ElektraError * error);
-const char * elektraKDBErrorDescription (const ElektraError * error);
-ElektraErrorSeverity elektraKDBErrorSeverity (const ElektraError * error);
-ElektraKDBErrorGroup elektraKDBErrorGroup (const ElektraError * error);
-ElektraKDBErrorModule elektraKDBErrorModule (const ElektraError * error);
-const char * elektraKDBErrorReason (const ElektraError * error);
-int elektraKDBErrorWarningCount (const ElektraError * error);
-ElektraError * elektraKDBErrorGetWarning (const ElektraError * error, int index);
-Key * elektraKDBErrorKey (const ElektraError * error);
 
 void elektraErrorReset (ElektraError ** error);
 
--- a/src/include/elektra/error.h
+++ b/src/include/elektra/error.h
@@ -22,7 +22,6 @@ typedef struct _ElektraError ElektraError;
 
 typedef void (*ElektraErrorHandler) (ElektraError * error);
 
-ElektraErrorCode elektraErrorCode (const ElektraError * error);
 const char * elektraErrorDescription (const ElektraError * error);
 
 void elektraErrorReset (ElektraError ** error);
--- a/tests/kdb/testkdb_highlevel.cpp
+++ b/tests/kdb/testkdb_highlevel.cpp
@@ -140,7 +140,8 @@ protected:
 		elektraFatalErrorHandler (elektra, &fatalErrorHandler);
 	}
 
-	void setValues (std::initializer_list<kdb::Key> values)
+	template <class T>
+	void setValues (T values)
 	{
 		using namespace kdb;
 		KDB kdb;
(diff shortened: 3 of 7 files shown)


kodebach:
jenkins build all please

kodebach:
jenkins build all please

src/error/exporterrors.cpp
```diff
@@ -524,6 +524,7 @@ static ostream & printPrivate (ostream & os, parse_t & p)
 	   << endl
 	   << "#include <elektra/types.h>" << endl
 	   << "#include <elektra/error.h>" << endl
+	   << "#include <kdbprivate.h>" << endl
```
markus2330: Why is this needed? It would be good if we can avoid it. If it is only about the enum, we better move the enum within here.

src/include/elektra/error.h
```diff
@@ -20,34 +20,10 @@ extern "C" {
...
-	ELEKTRA_ERROR_SEVERITY_ERROR,
-	ELEKTRA_ERROR_SEVERITY_WARNING
-} ElektraErrorSeverity;
-
-typedef const char * ElektraKDBErrorGroup;
-typedef const char * ElektraKDBErrorModule;
-
 typedef void (*ElektraErrorHandler) (ElektraError * error);
 
 ElektraErrorCode elektraErrorCode (const ElektraError * error);
```
markus2330: Remove this also?

tests/kdb/testkdb_highlevel.cpp
```diff
@@ -12,6 +12,7 @@
 
 #include <gtest/gtest-elektra.h>
 #include <kdbhelper.h>
+#include <kdbprivate.h>
```
markus2330: We also should have ABI tests, i.e. tests that only use the public interface.

tests/kdb/testkdb_highlevel.cpp
```diff
@@ -193,6 +200,141 @@ class Highlevel : public ::testing::Test
...
+		makeKey (KDB_TYPE_LONG_LONG, "longlong/above", "9223372036854775808"),
+		makeKey (KDB_TYPE_UNSIGNED_LONG_LONG, "unsignedlonglong/below", "-1"),
+		makeKey (KDB_TYPE_UNSIGNED_LONG_LONG, "unsignedlonglong/min", "0"),
+		makeKey (KDB_TYPE_UNSIGNED_LONG_LONG, "unsignedlonglong/max", "18446744073709551615"),
+		makeKey (KDB_TYPE_UNSIGNED_LONG_LONG, "unsignedlonglong/above", "18446744073709551616"),
+	});
+
+	createElektra ();
+
+	EXPECT_EQ (elektraGetOctet (elektra, "octet/min"), 0) << "octet/min wrong";
```
markus2330: Thank you for adding the tests!

tests/kdb/testkdb_highlevel.cpp
kodebach: The tests themselves do only access the public interface. Some use the low-level API too to check the results, but `kdbprivate.h` is only used for the fatal error handler that outputs error messages. In particular we only use the `elektraError`* and `elektraKDBError*` functions that were part of the public interface before this PR.

src/error/exporterrors.cpp
kodebach: It's needed for `ElektraErrorSeverity`, which is used in the generated function `elektraErrorLowLevel`. The function is used to create the `ElektraError` struct for low-level errors.

I don't think it is a problem to include `kdbprivate.h` here. The generated file is `elektra/errorsprivate.h` and is not intended for external use. It is not even included in any CMake install targets. It is not referenced in any header files, so it basically doesn't exist outside of compiling Elektra.

kodebach:
> Is the high-level API ready to be released after this PR?

I was thinking, now that we decided to use the `specload` plugin for code generation, the `defaults` parameter of `elektraOpen` causes more problems than it solves. We might want to remove it. But I wouldn't actually remove the parameter. I would just not use it and rename it to `config`. That way we are prepared, if we need to pass more configuration parameters to `elektraOpen` in the future (e.g. enabling/disabling certain global plugins, or anything else that has to be done between `kdbOpen` and `kdbGet`).

kodebach:
> That way we are prepared, if we need to pass more configuration parameters to `elektraOpen` in the future

Supporting the notification API through the highlevel API (including code generation) for example would require configuration between `kdbOpen` and `kdbGet`. In this case no reasonable defaults exist, unlike with cmd-line options, so configuration has to be user supplied.

We should also probably provide a `elektraGetKDB` function that returns the `KDB` handle used by the highlevel API, so that it can be combined with the low-level API more easily. Possibly we also expose the internal `KeySet`, but I am not sure about that. Although both of these can be added in a future release.

tests/kdb/testkdb_highlevel.cpp
markus2330: For ABI tests it would be needed that no methods of kdbprivate.h is used at all. Otherwise the ABI tests would also fail on changes within kdbprivate.h.

markus2330:
> I was thinking, now that we decided to use the specload plugin for code generation, the defaults parameter of elektraOpen causes more problems than it solves.

The default parameter is still needed because also kdbOpen() or some unrelated part within kdbGet() can fail. The goal would be that the application is always be able to start, independent of KDB.

> Supporting the notification API through the highlevel API (including code generation) for example would require configuration between kdbOpen and kdbGet.

You mean for elektraIoSetBinding? I don't think that "configuration" would be an elegant solution, it seems to be more straight-forward to call functions because you need to pass function pointers.

> [get internal KDB and KeySet] Although both of these can be added in a future release.

Yes, let us only add it later when it is clear that it will be needed. (e.g. if we decide not to wrap functions like elektraIoSetBinding)

kodebach:
> I don't think that "configuration" would be an elegant solution, it seems to be more straight-forward to call functions because you need to pass function pointers.

Yes, passing any sort of non-string data is very ugly via a KeySet, but I don't see a better solution other than using a custom struct which has to be allocated again. Any configuration before `kdbOpen` or before `kdbGet` has to be done by `elektraOpen`, so the user has to give that information somehow. We also can't delay the `kdbGet` call until the first `elektraGet*` because then `elektraGet*` could fail.

kodebach:
> The default parameter is still needed because also kdbOpen() or some unrelated part within kdbGet() can fail. The goal would be that the application is always be able to start, independent of KDB.

Currently `elektraOpen` always returns `NULL` if `kdbOpen` or `kdbGet` fails.

markus2330:
Yes, for severe problems elektraOpen is allowed to fail. But for a later `elektraGet` also `kdbGet` (or even `kdbOpen` in case of new mountpoints) might fail. But then we should continue to run, as we do not want to fail during run-time.

kodebach:
jenkins build all

kodebach:
jenkins build all please

kodebach:
jenkins build all please

kodebach:
Contrary to what github thinks, all build jobs where successful. One of the Cirrus CI jobs is listed twice, clicking on Details and viewing on Cirrus CI reveals that the job finished.

markus2330:
Thank you for the great work!

Issue #2382, opened 2019-02-07
highlevel: type safety in specification


markus2330:
From  #2374:

> The parameter is intended as an alternative to a specification. If the specification is necessary anyway it is useless.

Yes, it is an alternative. But the user should be safe against wrong changes in the specification, in particular incompatible changes in types.

> Not really. I don't see a situation where the highlevel API would require a 3-way merge.

You are right: with a single key the 3-way merge is trivial.

> we can't avoid this redundant process

The problem is not the redundant conversion but that the user might have messed up the specification. If we check the specification (the types) at startup, this problem could not happen.

So to be more concrete, given following application:

```c
kdb_long_t myint = elektraGetLong (elektra, "myint");
```

the user should be safe against an exit() from Elektra when doing:

```sh
sudo kdb setmeta /sw/org/myapp/#0/current/myint type float
```

Expected behavior: `elektraOpen` should fail, telling that the type of myint does not match with the build-in spec.

kodebach:
> Expected behavior: `elektraOpen` should fail, telling that the type of myint does not match with the build-in spec.

Ideally the `kdb setmeta` call would fail. A user of an application shouldn't be allowed to change the specification of that application at all after it has been installed. How we could implement that, I am not sure. With code generation we could mount the specification as readonly before calling `elektraOpen` and unmount it after `elektraClose`.

markus2330:
> Ideally the kdb setmeta call would fail.

Yes, of course it would fail without sudo. I changed it accordingly.

kodebach:
> it would fail without sudo

That's not what I meant. In an ideal world `kdb setmeta` would fail because `spec/sw/org/myapp/#0/current/myint` is marked as part of the specification for "myapp". Therefore it cannot be changed by the user. But the only way to implement that enforceably is, if the application it self comes with its own readonly plugin that provides the specification. Which would be a possibility for code generation but seems impractical without.

markus2330:
This is clever, I like the idea.

Yes, code generation is useful, otherwise the spec in the application might not describe the configuration actually accessed by the application. But you *could* write the built-in KeySet passed to `elektraOpen` by hand (we obviously do not recommend to do this).

[...]

Issue #2209, opened 2018-08-24
peformance problem with pluginprocess


markus2330:
## Steps to Reproduce the Problem

1. have a huge KeySet
2. globally mount some plugin using gmount and then interact with KDB:
```sh
kdb gmount python script=/path/to/elektra/src/plugins/python/python/python_filter.py
kdb get /something
```
## Expected Result

Should be about same speed.

## Actual Result

It is about 10x slower, from 2sec -> 20sec for my keys.

## System Information

- Elektra Version: master

## Further Input

@mpranj Your global mmap would be the perfect fix because in this case, where I have this problem, basically exclusively `kdb get` is done. (`kdb set` happens only at night in cron jobs.) With global mmap no pluginprocess or python would be involved, and not even the ini plugin, so we should be even below 2 sec.

Nevertheless, I am interested if exchanging the dump plugin in ~e/src/libs/pluginprocess/pluginprocess.c with mmapstorage also improves the situation. 20 -> 10 seconds would already be great :+1: 

I only did a short benchmark and found out that the python plugin causes the overhead and that the python code itself does not make a difference (e.g. if I immediately return some value). So I suspected the serialization. I am, however, not sure about that. So please do not invest too much time into it. If you find some way to finish mmap global plugin in September you can simply forget this task :smile: 

If this improves the situation I have a big motivation to quickly merge your #1990  PR :wink:

mpranj:
#1990 is mmapstorage and not yet tested as a global plugin. It is, however, nearly finished and working well as a default storage plugin. I feel like it can almost be merged, I am just testing a few last things.

I would like to merge #1990 as a storage plugin. After that, the plan is to make minor modifications to the global-plugins and mmapstorage. Mmapstorage only needs a mode where, as a global plugin, it uses a predefined cache file to store the complete KeySet. If everything works as planned, there will be no difference between mmapstorage and the global mmap cache (except the file path).

It will definitely be finished in September.

markus2330:
For what I propose here, it would be used as storage plugin as serialization of KeySet(s) for inter-process communication, not as global plugin.

`sed -i 's/dump/mmap/' src/libs/pluginprocess/pluginprocess.c`

should be the only necessary code modification. You can have this little commit within #1990.

If it does not work or is not faster we might learn something nevertheless. (Maybe it is unsuitable for streaming.)

> If everything works as planned, there will be no difference between mmapstorage and the global mmap cache (except the file path).

[...]

PR #2433, opened 2019-02-20
Specload

kodebach:
This PR implements the plugin conceptualized in #2382.

It also contains a general purpose storage plugin called `quickdump`. The format used is based on the `dump` plugin, but optimized for IPC. This plugin is used to send the specification from the application to the plugin and also for storage of the overridden parts of the specification.

## Basics

## Checklist

- [x] I added unit tests

## Review


mpranj:
You can add quickdump to `tests/ctest/test_storage.c` for some general purpose storage tests, if desired. Just add the plugin name to the array and increment `NUM_PLUGINS_SUGGESTED`.

https://github.com/ElektraInitiative/libelektra/blob/0c750fcb3a3dab73ca14be655b0a478ed8c555ca/tests/ctest/test_storage.c#L21-L24

doc/news/_preparation_next_release.md
```diff
@@ -145,6 +145,19 @@ and be assured that you can safely set a path value to the key. A more detailed
 
 - Added some basic unit tests _(Bernhard Denner)_
 
+### Quickdump
```
markus2330: Does quickdump also work as transport format for pluginprocess? It could be very interesting as we have performance difficulties for large key sets, see #2209.

Btw. it would be great if we find a BSD-licensed alternative to the "portable" library. Why are ntohl (or htobe64 if needed) and similar not enough?

mpranj: after the recently discussed changes to mmapstorage (copy data to non-regular files), it should be compatible too, fyi

kodebach: > Does quickdump also work as transport format for pluginprocess?

(In theory) Yes, quickdump is now 100% equivalent to dump. It even supports shared metakeys (similar to `keyCopyMeta` in dump).

> after the recently discussed changes to mmapstorage (copy data to non-regular files), it should be compatible too, fyi

I think `quickdump` would be more suitable for IPC. If pipes are involved `mmapstorage` has to do a lot of copying and in the end it has to fix all pointers.

> Btw. it would be great if we find a BSD-licensed alternative to the "portable" library.

I will look into that. The current files are simply taken from https://github.com/dvidelabs/flatcc. 

> Why are ntohl (or htobe64 if needed) and similar not enough?

Firstly, as noted in the README, I decided to use little-endian (except for the magic number) to store the data, because of performance advantages.

https://github.com/ElektraInitiative/libelektra/blob/6847f6831adad154c6b03e86ef53ae1a51eb3888/src/plugins/quickdump/README.md#L28

The "library" comes from Flatbuffers (a performance optimized serialization format), it uses the most performant conversion it can find. `htobe64` et al. are non-standard. If they are available (and no compiler built-in can be used) they will be used. In practice of course using a naive implementation with `<<` and `&` would be good enough too, because almost all modern systems are little-endian anyway.

kodebach: I found this public domain endianness helper file: https://gist.github.com/michaeljclark/3b4fd912f6fa8bb598b3 
It covers less cases of optimization, but it should work for our purposes.

I added it as `kdbendian.h` so other plugins/libraries can use it too.

markus2330: > Yes, quickdump is now 100% equivalent to dump.

So it is basically binarydump?

> It covers less cases of optimization, but it should work for our purposes. I added it as kdbendian.h so other plugins/libraries can use it too.

Thank you!

kodebach: > So it is basically binarydump?

Basically yes. Although `dump` technically uses binary files too, because it stores the null terminators of strings.

kodebach:
@mpranj Adding `quickdump` to `tests/ctest/test_storage.c` somehow breaks the `debian-stable-full-i386` job. I cannot reproduce this locally and also the Travis Linux FULL job works. Weirdly it seems that a segfault occurs during the tests for the `dump` plugin.

https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2433/11/pipeline/450

mpranj:
@kodebach that makes no sense to me either. I will try to reproduce and fix it tomorrow!

markus2330:
> I cannot reproduce this locally

Even if you get our docker image, see scripts/docker/README.md? Did you try to compile with 32bit?

mpranj:
> Did you try to compile with 32bit?

I have a 32bit vagrant VM for that.

```
$ valgrind ./bin/test_storage 
==15173== Memcheck, a memory error detector
==15173== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==15173== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==15173== Command: ./bin/test_storage
==15173== 
STORAGE     TESTS
==================

Testing plugin mmapstorage_crc
Tmp-file: /tmp/elektra-test.Qi0Gw2/elektra-tmp.kr7qJ8
Testing plugin mmapstorage
Tmp-file: /tmp/elektra-test.Qi0Gw2/elektra-tmp.kr7qJ8
Testing plugin dump
Tmp-file: /tmp/elektra-test.Qi0Gw2/elektra-tmp.kr7qJ8
Testing plugin quickdump
Tmp-file: /tmp/elektra-test.Qi0Gw2/elektra-tmp.kr7qJ8
src/libs/elektra/internal.c:144:elektraStrNCmp: Assertion `s1 != ((void *)0) && s2 != ((void *)0)' failed: Got null pointer s1: 0x49d17d0 s2: (nil)
==15173== 
==15173== Process terminating with default action of signal 6 (SIGABRT)
==15173==    at 0x40A4EA9: raise (raise.c:54)
==15173==    by 0x40A6406: abort (abort.c:89)
==15173==    by 0x40616C4: elektraAbort (log.c:184)
==15173==    by 0x405E54A: elektraStrNCmp (internal.c:144)
==15173==    by 0x804AC2B: test_keyValue (test_storage.c:722)
==15173==    by 0x804AC2B: main (test_storage.c:1021)
==15173== 
==15173== HEAP SUMMARY:
==15173==     in use at exit: 31,885 bytes in 97 blocks
==15173==   total heap usage: 23,686 allocs, 23,589 frees, 4,279,446 bytes allocated
==15173== 
==15173== LEAK SUMMARY:
==15173==    definitely lost: 0 bytes in 0 blocks
==15173==    indirectly lost: 0 bytes in 0 blocks
==15173==      possibly lost: 0 bytes in 0 blocks
==15173==    still reachable: 31,885 bytes in 97 blocks
==15173==         suppressed: 0 bytes in 0 blocks
==15173== Rerun with --leak-check=full to see details of leaked memory
==15173== 
==15173== For counts of detected and suppressed errors, rerun with: -v
==15173== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
Aborted (core dumped)
```

kodebach:
Thank you for finding the bug, I missed that the build job is 32 bit. I will investigate it further.

> for me the problem definitely occurs at quickdump

Possibly Jenkins terminates the job before it can print the full log.

kodebach:
I am pretty sure I fixed the symptoms but not the actual problem...

We use `size_t` (implementation defined size > 16 bit) for the value size of a key. I used `kdb_unsigned_long_long_t` (always 64 bit) so the value would be portable. When I passed the value to `keyNew`, because of the way `va_list` is implemented and the fact that the system is a little endian 32 bit system. Only the first half of the size was used as such. The second half was used as the next `keyswitch_t action` and interpreted as `KEY_END`.

The problem should be fixed, but we should really only use fixed size types inside `va_list`s to avoid this problem. Especially because problems like this can be very subtle and stay undetected for a long time.

kodebach:
jenkins build all

kodebach:
jenkins build all please

markus2330:
> The problem should be fixed, but we should really only use fixed size types inside va_lists to avoid this problem. Especially because problems like this can be very subtle and stay undetected for a long time.

Yes, I fully agree. Can you make an issue?

kodebach:
jenkins build all please

kodebach:
jenkins build all please

kodebach:
@sanssecours The `mac ASAN...` job on Cirrus CI consistently times out during `testmod_zeromqsend`. Do you know why this might be?

kodebach:
jenkins build all please

sanssecours:
> The `mac ASAN...` job on Cirrus CI consistently times out during `testmod_zeromqsend`. Do you know why this might be?

Not really. It might be a temporary problem though. At least on my computer the test seems to work, if I use the code in this PR and create the build system with

```sh
cmake -GNinja -DBINDINGS=ALL -DPLUGINS=ALL -DENABLE_ASAN=ON ..
```

.

sanssecours:
Looks like this pull request really breaks `testmod_zeromqsend` on macOS, if AddressSanitizer is enabled.

To debug this problem I created a [branch](https://github.com/sanssecours/elektra/tree/👓) that contains the code of this PR, removed the uninteresting build job code, and added the command

```sh
sudo dtruss ctest -V -R testmod_zeromqsend
```

to [the remaining Cirrus build job](https://cirrus-ci.com/build/5672479948275712), which (unfortunately) seems to work without any problems.

kodebach:
I think we might simply be hitting the default Cirrus CI timeout of 60 min. Each run of the job in question fails after exactly `43:08`. Maybe Cirrus CI calculates the timeout interval differently to what it displays in the UI.

sanssecours:
> I think we might simply be hitting the default Cirrus CI timeout of 60 min.

That might be the case. However, a [test time of about 37 minutes](https://github.com/ElektraInitiative/libelektra/pull/2433/checks?check_run_id=67963263) is **quite unusual**. Usually the test stage of `🍎 Clang ASAN` [takes about 10 minutes](https://github.com/ElektraInitiative/libelektra/pull/2438/checks?check_run_id=67955066).

kodebach:
After some more investigation, I think it might be a coincidence that the timeout always occurs after `testmod_zeromqsend`. I think the only running test at that point (I don't know why the remaining ~70 tests are not started) is acutally `testmod_specload`. So I will need to fix something after all, I just need to find out what the actual bug is ...

kodebach:
jenkins build all please

sanssecours:
> I did some experiments on [another branch](https://github.com/kodebach/libelektra/tree/specload-test) and it turns out a simple `sudo` fixes the bug with `testmod_specload`. 

Glad you found a solution to the problem. It would be really nice, if you could find a workaround that does not require `sudo` though. I do not think it does matter that much on the build server, but requiring a user/developer to run the test suite with root privileges does not sound that great in my opinion.

> However it [breaks `testshell_markdown_yanlr`](https://cirrus-ci.com/task/6286963268124672).

I added a fix for the failing test in commit f2b2d3c.

kodebach:
jenkins build all please

kodebach:
jenkins build all please

kodebach:
jenkins build all

kodebach:
jenkins build all please

kodebach:
> but requiring a user/developer to run the test suite with root privileges does not sound that great in my opinion.

I agree, but locally (under Linux) I don't need `sudo` to run the test suite. Neither does Travis. The non-ASAN jobs work everywhere. If someone with a Mac can confirm that this PR actually breaks `run_nocheckshell` with ASAN enabled and without root permissions, then I will look into it further. But if the problem only occurs on the Cirrus, I will simply add `sudo`.

sanssecours:
> If someone with a Mac can confirm that this PR actually breaks `run_nocheckshell` with ASAN enabled and without root permissions, then I will look into it further.

It looks like the tests for `specload` works without `sudo` on my machine. The failure in the Cirrus build job might still show some hidden problems with the current code though, since all other tests did work without `sudo`.

Anyway, in case you have not seen it yet, the UndefinedBehaviorSanitizer does print a warning for `quickdump` on my machine:

```
../src/plugins/quickdump/quickdump.c:507:10: runtime error: negation of 1 cannot be represented in type 'size_t' (aka 'unsigned long')
```
.

kodebach:
> It looks like the tests for `specload` works without `sudo` on my machine

[Travis](https://travis-ci.com/kodebach/libelektra/jobs/180420065) has the same behaviour as Cirrus, so there is definitely something that needs fixing. But if it doesn't even come up a normal Mac, it will take some time...

> the UndefinedBehaviorSanitizer does print a warning for `quickdump` on my machine

Thanks for spotting that 👍

kodebach:
An Update, if anybody has ideas:

- The [test branch](https://github.com/kodebach/libelektra/tree/specload-test) now works under Clang ASAN (local and server):
  - Linux Travis: https://travis-ci.com/kodebach/libelektra/jobs/180929871
  - macOS Cirrus: https://cirrus-ci.com/task/4599170104033280
- It doesn't work under GCC: https://travis-ci.com/kodebach/libelektra/jobs/180929870
  Through local testing I found:
  - When `BUILD_TYPE=Debug`: all tests fail because `elektraModuleLoad` fails somehow when ASAN is on.
  - When `BUILD_TYPE=RelWithDebInfo`:  for some reason only `test_elektra-specload-testapp` fails, but also because `elektraModuleLoad` fails.

src/plugins/specload/testapp.c
```diff
@@ -0,0 +1,83 @@
...
+#else
+#define STDOUT_FILENAME ("/dev/stdout")
+#endif
+
+static int outputKeySet (KeySet * ks)
+{
+	Key * errorKey = keyNew (0, KEY_END);
+
+	KeySet * modules = ksNew (0, KS_END);
+	elektraModulesInit (modules, 0);
```
markus2330: Why is elektraModulesInit needed? That should be done by elektraInvokeOpen

kodebach: Shouldn't really matter considering what `elektraModulesInit` actually does...

But I will remove it anyway

markus2330: You are right, it most likely will not be related to the problem we have here.

kodebach:
jenkins build all please

kodebach:
jenkins build all please

kodebach:
@markus2330 As soon as the build jobs run successfully, this is ready to merge. There shouldn't be any real errors, I just had to increase the timeout because the Jenkins server seems to take much longer for the test runs than I expected.

---

In the last meeting it was noted that quickdump could be improved by reusing buffers. I profiled `elektraQuickdumpGet` with a KeySet of 1,000,000 Keys and found that the direct calls to `elektraMalloc`  /`elektraFree` only account for ~ 1% of time. The whole `readString` method only amounts to < 10%. The bulk of time (~ 80%) is spent in `keyNew` (because of unescaping) and `ksAppendKey` so until these are optimized, I don't see a reason to change quickdump.

[callgrind results](https://user-images.githubusercontent.com/23529132/53637401-58a15200-3c23-11e9-9a97-fc54465f268c.png)

markus2330:
Yes, I agree. ksNew should be optimized first.

src/error/specification
```diff
@@ -1324,3 +1324,17 @@ severity:error
...
+
+number:211
+description:error while writing file
+severity:error
+ingroup:plugin
+module:storage
+macro:WRITE_FAILED
+
+number:212
+description:an error occurred in the specload plugin (see reason)
```
markus2330: This seems to be a very specific (only for specload plugin's errors) and generic (various different errors as stated in the reason) at the same time.

@Piankero What do you think?

src/plugins/quickdump/README.md
```diff
@@ -0,0 +1,114 @@
...
+- infos/provides = storage/quickdump
+- infos/recommends =
+- infos/placements = getstorage setstorage
+- infos/status = tested nodep libc compatible preview
+- infos/metadata =
+- infos/description = much quicker version of dump
+
+## Introduction
+
+`quickdump` is a storage plugin based on the `dump` format. It is a lot quicker than the old `dump` plugin, because it does not use commands
```
markus2330: How much quicker?

src/plugins/quickdump/README.md
```diff
@@ -0,0 +1,114 @@
...
+and stores string lengths as binary data. Through these changes all string comparisons and integer-string conversions can be eliminated,
+which made up for a lot of the time spent by the `dump` plugin.
```
markus2330: Numbers?

src/plugins/quickdump/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+The format is also useful for IPC and streaming, because of this it is used by the `specload` plugin.
```
markus2330: and pluginprocess?

src/plugins/quickdump/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Format
+
+A `quickdump` file starts with the magic number `0x454b444200000001`. The first 4 bytes are the ASCII codes for `EKDB` (for Elektra KDB),
+followed by a version number. This 64-bit is always stored as big-endian (i.e. the way it is written above).
+
+After the magic number the file is just a list of Keys. Each Key consists of a name, a value and any number of metakey names and values.
+Each name and value is written as a 64-bit length `n` followed by exactly `n` bytes of data. For strings we do not store a null terminator.
```
markus2330: Is this compatible to dump? See also #2452

src/plugins/quickdump/README.md
```diff
@@ -0,0 +1,114 @@
...
+Therefore the length also does not account for that. When reading a string, you will need to allocate `n+1` bytes and set the last on to `0`.
```
markus2330: "you will need" -> "the plugin allocates"

src/plugins/quickdump/quickdump.c
```diff
@@ -0,0 +1,558 @@
...
+	elektraFree (metaKeys.array);
+
+	fclose (file);
+
+	ksSetCursor (returned, cursor);
+
+	return ELEKTRA_PLUGIN_STATUS_SUCCESS;
+}
+
+Plugin * ELEKTRA_PLUGIN_EXPORT
```
markus2330: Usually this is the last function.

src/plugins/quickdump/testmod_quickdump.c
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	elektraFree (infile);
+	elektraFree (outfile);
+	ksDel (ks);
+}
+
+int main (int argc, char ** argv)
+{
+	printf ("QUICKDUMP     TESTS\n");
```
markus2330: Did you also try to add your plugin tests/shell/check_kdb_internal_suite.sh

src/plugins/specload/README.md
```diff
@@ -0,0 +1,73 @@
...
+This is were this plugin comes in. Unlike other storage plugins `specload` doesn't use the file given during mounting for most of the
+configuration. Instead `specload` requests the specification with which an application was compiled from the application itself. On top
+of this base specifications the user can make some modifications, which are the only keys stored in the mounted file. All modifications
+made by the user are verified by the plugin. The user can never modify the specification in a way to would break application compatibility.
+
+NOTE: currently the modifications a user can make are very limited. See [Limitations](#limitations) below.
+
+## Usage
+
+To use `specload` first make sure Elektra was compiled with `BUILD_SHARED` enabled, and the `quickdump` plugin is available as well:
```
markus2330: This is dependency, not usage. Maybe we should swap Dependency and Usage sections?

src/plugins/specload/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+kdb list | grep '^quickdump$'
```
markus2330: Why not `kdb list quickdump`?

src/plugins/specload/README.md
```diff
@@ -0,0 +1,73 @@
...
+added by the user, because the application might rely on the default `opt/arg=none`. See also [Limitations](#limitations).
+
+## Dependencies
+
+The plugin relies heavily on the `quickdump` plugin. It is used for storing the overridden values as well as the data transfer between
+`specload` and an application.
+
+## Examples
+
+TODO
```
markus2330: Please add with testapp

src/plugins/specload/specload.c
```diff
@@ -0,0 +1,611 @@
...
+	{
+		ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_SPECLOAD, parentKey, "Could not execute app: %s", strerror (errno));
+		return NULL;
+	}
+	close (stdin_copy);
+
+	return result;
+}
+
+/**
```
markus2330: Please add at least a brief documentation. "Is change allowed" of what?

.idea/vcs.xml
```diff
@@ -3,4 +3,4 @@
   <component name="VcsDirectoryMappings">
     <mapping directory="$PROJECT_DIR$" vcs="Git" />
   </component>
-</project>
+</project>
```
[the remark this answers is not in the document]
kodebach: CLion did that automatically, so if undid it, sooner or later this would probably appear again in a PR.

src/error/specification
kodebach: Because we are changing the error concept, I didn't want to introduce additional errors that had to be removed/changed anyways.

src/plugins/quickdump/README.md
kodebach: As soon as it is implemented (new PR so this can be merged quicker), I will add a note.

src/plugins/quickdump/README.md
kodebach: compatible in what way? The whole format is a slight variation on the dump format, but that is more coincidental than intended.

Also the `dump` format could technically write multiple KeySets into a single file (it has `ksNew` and `ksEnd` commands). `quickdump` has no equivalent of that.

src/plugins/quickdump/testmod_quickdump.c
kodebach: Technically that test should automatically do basic tests with `quickdump`, but I will add a case for additional tests.

src/plugins/specload/README.md
kodebach: Should the example work as a shellrecorder test? Because that might be a bit difficult to achieve.

markus2330: Ideally, it would. If it is not possible, then also a non-shellrecorder example would already help. It does not need to be within this PR.

src/plugins/quickdump/README.md
kodebach: I added some benchmarks.

I think in these benchmarks you can also see, that reusing the buffer would only have a minor impact. In the get only benchmark the `dump` plugin slowly catches up to the speed of the `quickdump` plugin. Most likely this is due too the fact that `quickdump` does a lot more malloc/free. Other than that I don't really see a reason why the factor should change significantly with the size of the keyset (for get only, set is another thing).

src/plugins/specload/README.md
kodebach: Adding an example was acutally a good idea, because it showed some new bugs.

Most importantly the way the resolver works, makes it acutally impossible for `specload` to work. The only way it can be mounted right now is via `noresolver`, which obviously means it will be read-only and without overlay.

src/error/specification
Piankero: Thank you @kodebach. I agree with you

kodebach:
@markus2330 This PR is finally done.

markus2330:
Great work!

src/plugins/quickdump/README.md
markus2330: This is a reasonable choice. Something like ksNew/ksEnd can also be done in a format around individual files/exports. (like HERE documents)

Issue #2406, opened 2019-02-11
QT-GUI gives bad error messages if spec plugins reports an error


Piankero:
When trying to set a key to an invalid value (eg. wrong enum) and pressing the synchronize button in the gui, the error message is not helpful.

Here how it looks in the commandline
```
kdb set '/sw/lcdproc/lcdd/#0/current/curses/background' greeeeen 
#> Using name user/sw/lcdproc/lcdd/#0/current/curses/background
#> The command kdb set failed while accessing the key database with the info:
#> Sorry, the error (#121) occurred ;(
#> Description: validation failed
#> Reason: Validation of key "user/sw/lcdproc/lcdd/#0/current/curses/background" with string "greeeeen" failed.
#> Ingroup: plugin
#> Module: enum
#> At: ....../src/plugins/enum/enum.c:218
#> Mountpoint: user/sw/lcdproc/lcdd/#0/current
#> Configfile: ...../LCDd.conf.25676:1549919217.284067.tmp
```

Here is the error message I receive in the GUI:
![qt-gui-spec-error-message](https://user-images.githubusercontent.com/12346621/52593655-c854c600-2e49-11e9-90a4-3cb60d0c2cb7.png)

#2332

markus2330:
Thank you for finding this tricky problem! I did not know about it.

In which order do you do the `kdb set` and pressing of the synchronization button?

We will improve (rewrite) the 3-way merging soon (), the problem will hopefully fixed with this.

Issue #1473, opened 2017-04-17
Error severity level "normal"


domhof:
In src/error/specification several severity level are used: fatal, error, normal, warning.

What is the difference between error and normal?

markus2330:
Thanks for asking the question!

Currently everything which is not a warning, is an error. So if you clean up, you can fix fatal and normal to be an error.

domhof:
Thanks for the info! I fixed "normal" to be an "error". However I'm not sure if we should do the same to "fatal". In the API we have now "warning", "error" and "fatal" so we can map them 1:1 this way.

markus2330:
I think we should have very little user-visible errors and map all our internal errors to these few.

Basically the user is only interested in:

- validation errors (meaning that the user needs to change the value to succeed)
- conflicts (retrying might be useful, but might also be harmful and cause emergent misbehavior)
- temporary errors (retrying might be useful, e.g. out of disk space, NFS timeouts,..)
- permanent errors (installation broken, admin needs to fix something to make it work again)
- (did I miss something?)

domhof:
I like the idea of having a limited amount of user-visible errors. I could imagine, however, that a mapping can easily get cumbersome. Can you elaborate on how this mapping could be done?

markus2330:
The field `class` and `extends` already were introduced to achieve something like that. I think we can even remove the `extends` and only have 3-4 classes.

To avoid missing classes in future we can change `exporterrors` to abort if `class` is missing.

domhof:
Currently the specification contains the following classes:

FailedLoadModuleError (2 occurrences in specification with similar description: "could not load module...")
FailedResolveSymbolError (2 occurrences in specification with similar description: "could not get pointer to factory...")
NoDefaultBackendError (1 occurrence with description "failed to open default backend")

Is this really all we need and how can we map them (all "permanent errors")? At least for number:30 (conflict) we would need another class.

Who introduced `class` and what was it meant to represent exactly?

markus2330:
You can remove these 3 classes that occur in the files, they were never used.

Instead introduce the errors as mentioned above:
- **validation** errors (meaning that the user needs to change the value to succeed)
- **conflict**s (retrying might be useful, but might also be harmful and cause emergent misbehavior)
- **temporary** errors (retrying might be useful, e.g. out of disk space, NFS timeouts,..)
- **permanent** errors (installation broken, admin needs to fix something to make it work again)

(I marked a word in bold for its potential class name).

[...]

PR #2435, opened 2019-02-22
Error message concept

Piankero:
This PR is done as a basis for discussions on the new error concept and is not intended to be merged.

### DIFF
--- a/doc/ERROR_CONCEPT.md
+++ b/doc/ERROR_CONCEPT.md
@@ -37,10 +37,10 @@ From now on I will take the following error message as example:
 9. Configfile: ...../<file>.25676:1549919217.284067.tmp
 ```
 
+The lines are numbered and will be referred in the following sections.
+
 Generally I would suggest that this error message is way too much information for a user. We can agree on the fact 
 that #4 is by far the most relevant line.
-
-The lines are numbered and will be referred in the following sections.
 
 ### Line 9: Configfile
 
@@ -66,8 +66,8 @@ Reasons against it:
 * I highly doubt that most common users will require this information (unnecessary)
 
 Suggestions:
-//TODO: Validate this
-Remove it completely. If users want to know the location they should use `kdb file` on the concrete setting.
+Remove it for standard error messages. Maybe a `kdb set -v` could include such information for developers, users.
+Alternatively there could be a command for users to see the mountpoint, eg. `kdb get-mountpoint <key>`
 
 ### Line 7: At
 
@@ -142,7 +142,7 @@ if it is on a NFS and it is offline, *temporary* would be correct. If the file h
 appropriate since the admin has to fix it. If the file is currently in use and semlocked by the system, *conflict* would be appropriate.
 * core kdb error/plugin error:
 Not much value for the user
-* differentiation to core features of elektra: specification/ Notification/ KDB Core/ Bindings/ ...
+* differentiation to core features of elektra: Specification/ Notification/ KDB Core/ Bindings/ ...
 An error is again difficult to put into one bucket since some errors might be affected by multiple features (could not stat file for example
 which is used by multiple features).
 This approach would lead to the same problem as we have now (many errors, difficult to maintain)
@@ -234,7 +234,17 @@ No consolidation, splitting, etc. of categories needed in future versions
 No need for splitting the `specification` file because it is too large
 
 Drawbacks:
-* Possible small information loss by removing the description and `At`:
-`At` could be kept for devs by extending the `kdb set` command with an option `-v`.
+* Possible small information loss by removing the description, `At` and `mountpoint`:
+`At` could be kept for devs by extending the `kdb set` command with an option `-v`, `--debug`, etc.
+The same applies for the `mountpoint`.
 Description comes with too many drawbacks to be kept
 * Possible misleading the user with the `Solution` line
+
+The benefits do outweigh the drawbacks. I would suggest an overloaded method like the following:
+```
+// If a solution can be applied
+elektraError(Key parentKey, const * char message, const * char solution)
+
+// If it is not advisable to have a solution
+elektraError(Key parentKey, const * char message)
+```
--- a/doc/ERROR_CONCEPT.md
+++ b/doc/ERROR_CONCEPT.md
@@ -8,6 +8,7 @@ The purpose of this file is solely to better comment on in a Pull Request and ga
 2. Many duplicate or redundant errors
 3. Description/ Reason are very similar
 4. Potential file splitting (1327 LOC)
+5. Verbose error message
 
 ## Some facts
 
@@ -33,36 +34,36 @@ From now on I will take the following error message as example:
 6. Module: enum
 7. At: ....../src/plugins/enum/enum.c:218
 8. Mountpoint: <parentKey>
-9. Configfile: ...../<ile>.25676:1549919217.284067.tmp
+9. Configfile: ...../<file>.25676:1549919217.284067.tmp
 ```
 
-Generally I would suggest that this error message is way too much information for a user. We can agree on that #4 is by far
-the most relevant line.
+Generally I would suggest that this error message is way too much information for a user. We can agree on the fact 
+that #4 is by far the most relevant line.
 
 The lines are numbered and will be referred in the following sections.
 
 ### Line 9: Configfile
 
 Reasons to keep it:
-* If users want to see where the concrete file is located (basically if it is the file itself or some overwritten file)
+* If users want to see where the concrete file is located 
 
 Reasons against it:
 * Very verbose line
-* I highly doubt that most common users will require this information (unneccesary)
+* I highly doubt that most common users will require this information (unnecessary)
 
 Suggestions:
-Remove it completely. If users want to know the location they should use `kdb file` on the concrete setting.
+Remove it completely. If users want to know the location they still can use `kdb file` on the concrete setting.
 
 ### Line 8: Mountpoint
 
 Reasons to keep it:
 * If users want to see the parent key of their configuration
-* If I remember correctly: all error information is saved to this key, so users could investigate further errorinformation which is saved
-to this key (in the metadata)
+* If I remember correctly: all error information is saved to this key, so users could investigate further error
+information which is saved to this key (in the metadata)
 
 Reasons against it:
 * Another extra line in the error message.
-* I highly doubt that most common users will require this information (unneccesary)
+* I highly doubt that most common users will require this information (unnecessary)
 
 Suggestions:
 //TODO: Validate this
@@ -75,10 +76,10 @@ Reasons to keep it:
 
 Reasons against it:
 * Another extra line in the error message.
-* I highly doubt that most common users will require this information (unneccesary)
+* I highly doubt that most common users will require this information (unnecessary)
 
 Suggestions:
-Remove it for standard error messages. Maybe a `kdb set --verbose` could include such information for developers, users.
+Remove it for standard error messages. Maybe a `kdb set -v` could include such information for developers, users.
 Alternatively, another possibility would be to have a command which reveals the last error message with more detail: 
 `kdb last-message`.
 
@@ -92,8 +93,7 @@ Reasons against it:
 * Another extra line in the error message.
 
 Suggestions:
-The need for plugin distinguishment in the error message should not be that high in my opinion. Maybe it would make sense to include it in line #1:
-"The command kdb set failed while accessing the key database *(plugin: <plugin>)* with the info:"
+The need for plugin distinguishment in the error message should not be that high in my opinion. Maybe it would make sense to include it in line #1.
 
 ### Line 5: Ingroup
 
@@ -101,7 +101,7 @@ Reasons to keep it:
 * Another higher layer of distinguishment for the error location
 
 Reasons against it:
-* Extra line & unneccessary
+* Extra line & unnecessary
 
 Suggestion:
 Having both ingroup and module is too much information. Also again in most cases users should get the relevant information out of the reason line.
@@ -110,9 +110,9 @@ As of now there are 6 different "ingroup"s. I would suggest to remove it.
 ### Line 4: Reason
 
 The most relevant line out of all, no question to keep it or not.
-Still I would remove the prefix "Reason: " because of the reduced error message length in the new format. It makes it clear
+Still I would remove the prefix `Reason: ` because of the reduced error message length in the new format. It makes it clear
 that this line is the reason. Probably we could substitute it with the plugin which called the error: 
-"<plugin>: Could not set ......"
+`<plugin>: Could not set ...`
 
 ### Line 3: Description
 
@@ -132,21 +132,22 @@ Suggestion:
 This line is by far the most discussable one. I thought about categorizations such as
 * Permission/ Resource/ Parse/ External(C++, Ruby, Java)/.../ Miscellaneous:
 Most errors are resource errors (could not stat, no connection, could not close/open/find etc.) This again would not give users
-a particularly useful error indication since he/she has to lookup in the reason anyway. Even subdividing the Resource block
+a particularly useful error indication since he/she has to lookup in the reason anyway. Even subdividing the `Resource` category
 would lead to the same problem. No matter how I look at it, the user
-will have to read the concrete reason in most cases which makes the Description line not that valuable.
+will have to read the concrete reason in most cases which makes the Description line not that valuable any longer.
 * temporary/permanent/conflict/validation:
 This seems to be a better approach because it includes a possible solution for the user. How to categorize certain errors though
-seems impossible because there can be multiple reasons which can be categorized in more than one. Take "could not open file" as an example:
-if it is on an NFS and it is offline, "temporary" would be correct. If the file has too restrictive permissions, "permanent" would be more
-appropriate since the admin has to fix it. If the file is currently in use and semlocked by the system, "conflict"  would be appropriate.
+seems impossible because there can be multiple reasons where to categorize an error. Take `Could not open file` as an example:
+if it is on a NFS and it is offline, *temporary* would be correct. If the file has too restrictive permissions, *permanent* would be more
+appropriate since the admin has to fix it. If the file is currently in use and semlocked by the system, *conflict* would be appropriate.
 * core kdb error/plugin error:
 Not much value for the user
 * differentiation to core features of elektra: specification/ Notification/ KDB Core/ Bindings/ ...
-This is again difficult to put into one bucket since some errors might be affected by multiple features (could not stat file for example).
-Would lead to the same problem what is now most likely (many errors, difficult to maintain)
+An error is again difficult to put into one bucket since some errors might be affected by multiple features (could not stat file for example
+which is used by multiple features).
+This approach would lead to the same problem as we have now (many errors, difficult to maintain)
 
-Personally I do not see how the benefit of a categorization outweights the drawbacks because in by far the most cases users will have to look
+Personally I do not see how the benefit of a categorization outweighs the drawbacks because in by far the most cases users will have to look
 into the reason in more detail anyway. It should be enough for users to deal with the error just by the reason.
 
 As a result I would suggest to remove the description line.
@@ -156,7 +157,7 @@ As a result I would suggest to remove the description line.
 Reasons to keep it:
 * Probably could be used to categorize errors such it is done in the http protocol
 * Good to use for shellrecorder scripts to test for errors
-* Possibility for automation (eg. if a certain errornumber occured you could automatically trigger certain actions)
+* Possibility for automation (eg. if a certain error number occurred you could automatically trigger certain actions)
 
 Reasons against it:
 * Extra line
@@ -164,7 +165,7 @@ Reasons against it:
 Suggestion:
 Giving an error a number like it is done in C programs as return code seems to be outdated for elektra, especially when
 the reason is given anyway. Personally I never read the error number because it has not much value.
-The "reasons to keep" will also be very limited when the number of errors are significantly reduced.
+The *reasons to keep* will also be very limited when the number of errors are significantly reduced.
 I would also suggest to remove it.
 
 ### Line 1: General sentence
@@ -175,7 +176,7 @@ I would change it like this:
 "Module <plugin> issued a <error/warning/fatal> while accessing the key database with the info:"
 where plugin should have a different color (blue?) and warning/error/fatal (yellow, red, dark red) too depending on the severity.
 
-With this line 5/6 can still be present but in a much more efficient and succint way.
+With this line 5/6 can still be present but in a much more efficient and succinct way.
 
 ### Error/Warning/Fatal etc. splitting in the specification file
 
@@ -193,11 +194,11 @@ Some errors can also guide users to fix the error by suggesting a possible solut
 Developers though have to be very careful when suggesting solutions because it can easily mislead users. For many cases though, 
 this can help users a lot.
 I would suggest an optional extra line (which is only printed if the developer adds a metafield `solution` to the parentkey):
-"Possible Solution: ......"
+`Possible Solution: ...`
 
 # Conclusion
 
-With the new concept we would not require a `specification` file at all which helps developers immensly but also keeps the
+With the new concept we would not require a `specification` file at all which helps developers immensely but also keeps the
 most relevant information for the users. An error like this:
 
 ```
@@ -209,7 +210,7 @@ most relevant information for the users. An error like this:
 6. Module: enum
 7. At: ....../src/plugins/enum/enum.c:218
 8. Mountpoint: <parentKey>
-9. Configfile: ...../<ile>.25676:1549919217.284067.tmp
+9. Configfile: ...../<file>.25676:1549919217.284067.tmp
 ```
 
 would be changed to this:
--- a/doc/ERROR_CONCEPT.md
+++ b/doc/ERROR_CONCEPT.md
@@ -4,10 +4,17 @@ The purpose of this file is solely to better comment on in a Pull Request and ga
 
 ## Findings/ Motivation
 
-1. Too many errors in the [specificaion](src\error\specification) (210 errors/warnings)
+1. Too many errors in the [specification](src\error\specification) (210 errors/warnings)
 2. Many duplicate or redundant errors
 3. Description/ Reason are very similar
 4. Potential file splitting (1327 LOC)
 
 ## Some facts
 
+1. Some errors have the same text but are duplicated because one is `warning` and the other one is `error`/`fatal` 
+(#153 + #154, #128 + #129, #120 + #41, #44 + 45 , #66 + #67, etc.)
+2. Some errors might be duplicated just because they belong to different plugins (or the dev neglected to lookup a similar in the specification) 
+(#29 + #57 + #72 + #169, #10 + #61, etc.)
+3. Many errors are very similar or could be summarized and are differentiated via the reason field 
+(#54 + #55, #31 - #33, #93 + #7 + #17, #74 + #75, #77 + #10 + #61, etc.)
+4. Some errors are marked as unused (#22 + #48 + #58 + #66 + #67 + #68)
--- a/doc/ERROR_CONCEPT.md
+++ b/doc/ERROR_CONCEPT.md
@@ -18,3 +18,11 @@ The purpose of this file is solely to better comment on in a Pull Request and ga
 3. Many errors are very similar or could be summarized and are differentiated via the reason field 
 (#54 + #55, #31 - #33, #93 + #7 + #17, #74 + #75, #77 + #10 + #61, etc.)
 4. Some errors are marked as unused (#22 + #48 + #58 + #66 + #67 + #68)
+
+## Some considerations
+
+### Error/Warning/Fatal etc. splitting
+
+I would suggest to remove this concept from the specification and let the developer use a respective method to indicate if it is a warning/error/etc.
+The macros ELEKTRA_SET_ERROR, ELEKTRA_ADD_WARNING ... should be enough to make developers cautious of wrong method usage.
+What I have seen is that the return code is the only matter if it is an error or not.
--- a/doc/ERROR_CONCEPT.md
+++ b/doc/ERROR_CONCEPT.md
@@ -1,0 +1,13 @@
+# Error Message Concept
+
+The purpose of this file is solely to better comment on in a Pull Request and gather feedback.
+
+## Findings/ Motivation
+
+1. Too many errors in the [specificaion](src\error\specification) (210 errors/warnings)
+2. Many duplicate or redundant errors
+3. Description/ Reason are very similar
+4. Potential file splitting (1327 LOC)
+
+## Some facts
+
--- a/doc/ERROR_CONCEPT.md
+++ b/doc/ERROR_CONCEPT.md
@@ -18,6 +18,7 @@ The purpose of this file is solely to better comment on in a Pull Request and ga
 3. Many errors are very similar or could be summarized and are differentiated via the reason field 
 (#54 + #55, #31 - #33, #93 + #7 + #17, #74 + #75, #77 + #10 + #61, etc.)
 4. Some errors are marked as unused (#22 + #48 + #58 + #66 + #67 + #68)
+5. In many cases the `Description` is not very relevant at all (#199, #196, #187, etc.)
 
 ## Evaluation
 
@@ -27,12 +28,12 @@ From now on I will take the following error message as example:
 1. The command kdb set failed while accessing the key database with the info:
 2. Sorry, the error (#121) occurred ;(
 3. Description: validation failed
-4. Reason: Validation of key "user/sw/lcdproc/lcdd/#0/current/curses/background" with string "greeeeen" failed.
+4. Reason: Validation of key "<key>" with string "<value>" failed.
 5. Ingroup: plugin
 6. Module: enum
 7. At: ....../src/plugins/enum/enum.c:218
-8. Mountpoint: user/sw/lcdproc/lcdd/#0/current
-9. Configfile: ...../LCDd.conf.25676:1549919217.284067.tmp
+8. Mountpoint: <parentKey>
+9. Configfile: ...../<ile>.25676:1549919217.284067.tmp
 ```
 
 Generally I would suggest that this error message is way too much information for a user. We can agree on that #4 is by far
@@ -82,7 +83,99 @@ Alternatively, another possibility would be to have a command which reveals the 
 `kdb last-message`.
 
 ### Line 6: Module
-...
+
+Reasons to keep it:
+* See the concrete plugin which causes the failure
+* Prohibits error masking (eg. `stat`ing a file happens in more plugins, devs will know where the error occurs)
+
+Reasons against it:
+* Another extra line in the error message.
+
+Suggestions:
+The need for plugin distinguishment in the error message should not be that high in my opinion. Maybe it would make sense to include it in line #1:
+"The command kdb set failed while accessing the key database *(plugin: <plugin>)* with the info:"
+
+### Line 5: Ingroup
+
+Reasons to keep it:
+* Another higher layer of distinguishment for the error location
+
+Reasons against it:
+* Extra line & unneccessary
+
+Suggestion:
+Having both ingroup and module is too much information. Also again in most cases users should get the relevant information out of the reason line.
+As of now there are 6 different "ingroup"s. I would suggest to remove it.
+
+### Line 4: Reason
+
+The most relevant line out of all, no question to keep it or not.
+Still I would remove the prefix "Reason: " because of the reduced error message length in the new format. It makes it clear
+that this line is the reason. Probably we could substitute it with the plugin which called the error: 
+"<plugin>: Could not set ......"
+
+### Line 3: Description
+
+Reasons to keep it:
+* Gives users a quicker overview about the error (eg. permission error/ resource error/ etc. or another type of categorization such as Markus
+suggested with temporary/permanent/conflict/etc.)
+* Can indicate the type of solution a user has to approach
+
+Reasons against it:
+* Extra line
+* What to do with "uncategorizable" errors
+* Maintenance effort: introduce a new category once there are enough similar errors and change all old errors to this category?
+* Awareness of developers: All devs have to know all categories and choose the correct one. Devs should also be prohibited to throw their error
+into the "uncategorized" category due to lazyness.
+
+Suggestion:
+This line is by far the most discussable one. I thought about categorizations such as
+* Permission/ Resource/ Parse/ External(C++, Ruby, Java)/.../ Miscellaneous:
+Most errors are resource errors (could not stat, no connection, could not close/open/find etc.) This again would not give users
+a particularly useful error indication since he/she has to lookup in the reason anyway. Even subdividing the Resource block
+would lead to the same problem. No matter how I look at it, the user
+will have to read the concrete reason in most cases which makes the Description line not that valuable.
+* temporary/permanent/conflict/validation:
+This seems to be a better approach because it includes a possible solution for the user. How to categorize certain errors though
+seems impossible because there can be multiple reasons which can be categorized in more than one. Take "could not open file" as an example:
+if it is on an NFS and it is offline, "temporary" would be correct. If the file has too restrictive permissions, "permanent" would be more
+appropriate since the admin has to fix it. If the file is currently in use and semlocked by the system, "conflict"  would be appropriate.
+* core kdb error/plugin error:
+Not much value for the user
+* differentiation to core features of elektra: specification/ Notification/ KDB Core/ Bindings/ ...
+This is again difficult to put into one bucket since some errors might be affected by multiple features (could not stat file for example).
+Would lead to the same problem what is now most likely (many errors, difficult to maintain)
+
+Personally I do not see how the benefit of a categorization outweights the drawbacks because in by far the most cases users will have to look
+into the reason in more detail anyway. It should be enough for users to deal with the error just by the reason.
+
+As a result I would suggest to remove the description line.
+
+### Line 2: Error code
+
+Reasons to keep it:
+* Probably could be used to categorize errors such it is done in the http protocol
+* Good to use for shellrecorder scripts to test for errors
+* Possibility for automation (eg. if a certain errornumber occured you could automatically trigger certain actions)
+
+Reasons against it:
+* Extra line
+
+Suggestion:
+Giving an error a number like it is done in C programs as return code seems to be outdated for elektra, especially when
+the reason is given anyway. Personally I never read the error number because it has not much value.
+The "reasons to keep" will also be very limited when the number of errors are significantly reduced.
+I would also suggest to remove it.
+
+### Line 1: General sentence
+
+All in all this sentence can stay because it gives the user a first hint on what happened.
+I would though change the wording and include more relevant information in it.
+I would change it like this:
+"Module <plugin> issued a <error/warning/fatal> while accessing the key database with the info:"
+where plugin should have a different color (blue?) and warning/error/fatal (yellow, red, dark red) too depending on the severity.
+
+With this line 5/6 can still be present but in a much more efficient and succint way.
 
 ### Error/Warning/Fatal etc. splitting in the specification file
 
@@ -90,4 +183,57 @@ I would suggest to remove this concept from the specification and let the develo
 The macros ELEKTRA_SET_ERROR, ELEKTRA_ADD_WARNING ... should be enough to make developers cautious of wrong method usage.
 What I have seen is that the return code is the only matter if it is an error or not.
 
+### Additional information which is not present
 
+Some errors can also guide users to fix the error by suggesting a possible solution:
+* Retry as sudo
+* Fix file permissions
+* Check if the file exists
+* etc.
+Developers though have to be very careful when suggesting solutions because it can easily mislead users. For many cases though, 
+this can help users a lot.
+I would suggest an optional extra line (which is only printed if the developer adds a metafield `solution` to the parentkey):
+"Possible Solution: ......"
+
+# Conclusion
+
+With the new concept we would not require a `specification` file at all which helps developers immensly but also keeps the
+most relevant information for the users. An error like this:
+
+```
+1. The command kdb set failed while accessing the key database with the info:
+2. Sorry, the error (#121) occurred ;(
+3. Description: validation failed
+4. Reason: Validation of key "<key>" with string "<value>" failed.
+5. Ingroup: plugin
+6. Module: enum
+7. At: ....../src/plugins/enum/enum.c:218
+8. Mountpoint: <parentKey>
+9. Configfile: ...../<ile>.25676:1549919217.284067.tmp
+```
+
+would be changed to this:
+```
+1. Plugin enum issued a error while accessing the key database with the info:
+2. Validation of key "<key>" with string "<value>" failed.
+```
+
+with optionally a third line indicating a solution. Eg. for a permission related error there would be a third line:
+```
+3. Possible Solution: Retry the command as sudo (sudo !!)
+```
+
+Benefits:
+* Much more relevant information in the error message
+* Ease of development:
+no merge conflicts, no extra line in the `specification` file leading to chaos (such as it is now)
+New plugins will not have to choose an existent error category or implement a new one which could potentially be wrong
+* Less maintenance effort:
+No consolidation, splitting, etc. of categories needed in future versions
+No need for splitting the `specification` file because it is too large
+
+Drawbacks:
+* Possible small information loss by removing the description and `At`:
+`At` could be kept for devs by extending the `kdb set` command with an option `-v`.
+Description comes with too many drawbacks to be kept
+* Possible misleading the user with the `Solution` line
--- a/doc/ERROR_CONCEPT.md
+++ b/doc/ERROR_CONCEPT.md
@@ -19,10 +19,75 @@ The purpose of this file is solely to better comment on in a Pull Request and ga
 (#54 + #55, #31 - #33, #93 + #7 + #17, #74 + #75, #77 + #10 + #61, etc.)
 4. Some errors are marked as unused (#22 + #48 + #58 + #66 + #67 + #68)
 
-## Some considerations
+## Evaluation
 
-### Error/Warning/Fatal etc. splitting
+From now on I will take the following error message as example:
+
+```
+1. The command kdb set failed while accessing the key database with the info:
+2. Sorry, the error (#121) occurred ;(
+3. Description: validation failed
+4. Reason: Validation of key "user/sw/lcdproc/lcdd/#0/current/curses/background" with string "greeeeen" failed.
+5. Ingroup: plugin
+6. Module: enum
+7. At: ....../src/plugins/enum/enum.c:218
+8. Mountpoint: user/sw/lcdproc/lcdd/#0/current
+9. Configfile: ...../LCDd.conf.25676:1549919217.284067.tmp
+```
+
+Generally I would suggest that this error message is way too much information for a user. We can agree on that #4 is by far
+the most relevant line.
+
+The lines are numbered and will be referred in the following sections.
+
+### Line 9: Configfile
+
+Reasons to keep it:
+* If users want to see where the concrete file is located (basically if it is the file itself or some overwritten file)
+
+Reasons against it:
+* Very verbose line
+* I highly doubt that most common users will require this information (unneccesary)
+
+Suggestions:
+Remove it completely. If users want to know the location they should use `kdb file` on the concrete setting.
+
+### Line 8: Mountpoint
+
+Reasons to keep it:
+* If users want to see the parent key of their configuration
+* If I remember correctly: all error information is saved to this key, so users could investigate further errorinformation which is saved
+to this key (in the metadata)
+
+Reasons against it:
+* Another extra line in the error message.
+* I highly doubt that most common users will require this information (unneccesary)
+
+Suggestions:
+//TODO: Validate this
+Remove it completely. If users want to know the location they should use `kdb file` on the concrete setting.
+
+### Line 7: At
+
+Reasons to keep it:
+* Debug purposes
+
+Reasons against it:
+* Another extra line in the error message.
+* I highly doubt that most common users will require this information (unneccesary)
+
+Suggestions:
+Remove it for standard error messages. Maybe a `kdb set --verbose` could include such information for developers, users.
+Alternatively, another possibility would be to have a command which reveals the last error message with more detail: 
+`kdb last-message`.
+
+### Line 6: Module
+...
+
+### Error/Warning/Fatal etc. splitting in the specification file
 
 I would suggest to remove this concept from the specification and let the developer use a respective method to indicate if it is a warning/error/etc.
 The macros ELEKTRA_SET_ERROR, ELEKTRA_ADD_WARNING ... should be enough to make developers cautious of wrong method usage.
 What I have seen is that the return code is the only matter if it is an error or not.
+
+
--- a/doc/decisions/error_message.md
+++ b/doc/decisions/error_message.md
@@ -115,20 +115,20 @@ All "fatal" errors will be converted to "errors" as the distinguishment is not r
 
 Unused marked errors will be removed from the specification.
 
-The remaining ~130 errors will be categorizes into logical groups with subgroups.
-Each group and subgroup will receive a range of numbers such as in the HTTP protocol.
+Errors will be categorizes into logical groups with subgroups.
+Each error will be made up of 5 characters, where the first 2 character indicate the highest level
+and character 3 to 5 will be used for subgrouping.
 
-- Permanent errors (1-500)
-  - Resource (1-50)
-  - Parsing (51 - 100)
-  - Installation (101 - 150)
-  - Logical (151 - 200)
-- Temporary (501 - 1000)
-  - Conflict (501 - 600)
-  - Timeout (601 - 701)
-- Validation (1001 - 1500)
-  - Syntactic (1001 - 1100)
-  - Semantic (1101 - 1200)
+- Permanent errors (01000)
+  - Resource (01100)
+  - Parsing (01200)
+  - Installation (01300)
+  - Logical (01400)
+- Conflict (02000)
+- Timeout (03000)
+- Validation (04000)
+  - Syntactic (04100)
+  - Semantic (04200)
 
 ## Rationale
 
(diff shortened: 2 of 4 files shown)


Piankero:
@markus2330 Should I mark it as "Ready to merge" if want you to take a look at it? (despite it is not intended to be merged)

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
+# Error Message Concept
```
markus2330: Maybe you can use the doc/decision format?

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+The purpose of this file is solely to better comment on in a Pull Request and gather feedback.
+
+## Findings/ Motivation
+
+1. Too many errors in the [specification](src\error\specification) (210 errors/warnings)
+2. Many duplicate or redundant errors
+3. Description/ Reason are very similar
+4. Potential file splitting (1327 LOC)
```
markus2330: The problem is the missing modularity of the file.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+5. Verbose error message
+
```
markus2330: Another problem is that strerror sometimes is not part of the reason, sometimes (not) in quotes, sometimes \n is at the end. (These are basically bugs to be fixed but maybe a better error concept would also avoid such errors.)

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+Reasons to keep it:
+* If users want to see where the concrete file is located 
+
+Reasons against it:
+* Very verbose line
+* I highly doubt that most common users will require this information (unnecessary)
+
+Suggestions:
+Remove it completely. If users want to know the location they still can use `kdb file` on the concrete setting.
```
markus2330: The information was added on request, so I would not remove it completely. Maybe we disable it by default and only show it on request (-v for cmd-line).

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+3. Description: validation failed
+4. Reason: Validation of key "<key>" with string "<value>" failed.
+5. Ingroup: plugin
+6. Module: enum
+7. At: ....../src/plugins/enum/enum.c:218
+8. Mountpoint: <parentKey>
+9. Configfile: ...../<file>.25676:1549919217.284067.tmp
+```
+
+The lines are numbered and will be referred in the following sections.
```
markus2330: Did you compare what qt-gui and web-ui show you?

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+* If users want to see the parent key of their configuration
+* If I remember correctly: all error information is saved to this key, so users could investigate further error
+information which is saved to this key (in the metadata)
+
+Reasons against it:
+* Another extra line in the error message.
+* I highly doubt that most common users will require this information (unnecessary)
+
+Suggestions:
+Remove it for standard error messages. Maybe a `kdb set -v` could include such information for developers, users.
```
markus2330: Yes, I like the idea with -v. But then we should also say that you get more error information with -v?

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+Reasons to keep it:
+* Another higher layer of distinguishment for the error location
+
+Reasons against it:
+* Extra line & unnecessary
+
+Suggestion:
+Having both ingroup and module is too much information. Also again in most cases users should get the relevant information out of the reason line.
+As of now there are 6 different "ingroup"s. I would suggest to remove it.
```
markus2330: Yes, the group concept needs rework.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+Reasons to keep it:
+* Debug purposes
+
+Reasons against it:
+* Another extra line in the error message.
+* I highly doubt that most common users will require this information (unnecessary)
+
+Suggestions:
+Remove it for standard error messages. Maybe a `kdb set -v` could include such information for developers, users.
```
markus2330: Maybe even `kdb <command> -d` (debugging). We should however tell people that they need to use `-d` for bug reports.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+Alternatively, another possibility would be to have a command which reveals the last error message with more detail: 
+`kdb last-message`.
+
+### Line 6: Module
+
+Reasons to keep it:
+* See the concrete plugin which causes the failure
+* Prohibits error masking (eg. `stat`ing a file happens in more plugins, devs will know where the error occurs)
```
markus2330: Not only devs: also users mount some plugin, which might cause unexpected errors later.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+Still I would remove the prefix `Reason: ` because of the reduced error message length in the new format. It makes it clear
+that this line is the reason. Probably we could substitute it with the plugin which called the error: 
+`<plugin>: Could not set ...`
+
+### Line 3: Description
+
+Reasons to keep it:
+* Gives users a quicker overview about the error (eg. permission error/ resource error/ etc. or another type of categorization such as Markus
+suggested with temporary/permanent/conflict/etc.)
+* Can indicate the type of solution a user has to approach
```
markus2330: Maybe only categories should have a description, and not the errors themselves?

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+Reasons against it:
+* Extra line
```
markus2330: I think we should separate the discussions for which information needs to be available and which information should be presented to (end)users.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+* What to do with "uncategorizable" errors
```
markus2330: Examples?

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+* Maintenance effort: introduce a new category once there are enough similar errors and change all old errors to this category?
+* Awareness of developers: All devs have to know all categories and choose the correct one. Devs should also be prohibited to throw their error
+into the "uncategorized" category due to lazyness.
+
```
markus2330: Maybe plugins should even say in the contract which errors they may throw?

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+Suggestion:
+This line is by far the most discussable one. I thought about categorizations such as
+* Permission/ Resource/ Parse/ External(C++, Ruby, Java)/.../ Miscellaneous:
```
markus2330: I agree with "Resource" and "Parsing" Errors.

Permissions is too specific, I would put it to "Resource" and rather add other very general categories like: Validation, Transformation

"External" and "Miscellaneous" I don't like.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+* Gives users a quicker overview about the error (eg. permission error/ resource error/ etc. or another type of categorization such as Markus
+suggested with temporary/permanent/conflict/etc.)
+* Can indicate the type of solution a user has to approach
+
+Reasons against it:
+* Extra line
+* What to do with "uncategorizable" errors
+* Maintenance effort: introduce a new category once there are enough similar errors and change all old errors to this category?
+* Awareness of developers: All devs have to know all categories and choose the correct one. Devs should also be prohibited to throw their error
+into the "uncategorized" category due to lazyness.
```
markus2330: There should be no "other" category.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+Suggestion:
+This line is by far the most discussable one. I thought about categorizations such as
+* Permission/ Resource/ Parse/ External(C++, Ruby, Java)/.../ Miscellaneous:
+Most errors are resource errors (could not stat, no connection, could not close/open/find etc.) This again would not give users
+a particularly useful error indication since he/she has to lookup in the reason anyway. Even subdividing the `Resource` category
+would lead to the same problem. No matter how I look at it, the user
```
markus2330: I think only one division *might* be interesting for a user: Is it something that might get fixed by itself (connection problems) or something that a sysadmin can fix (permission, not enough space, ..)?

But even this division is shaky: some connection problems need to be fixed by admins, or some cron job might remove files (and thus "not enough space" might fix itself)

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+will have to read the concrete reason in most cases which makes the Description line not that valuable any longer.
+* temporary/permanent/conflict/validation:
```
markus2330: I agree, but see temporary/permanent in the discussion above.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+appropriate since the admin has to fix it. If the file is currently in use and semlocked by the system, *conflict* would be appropriate.
+* core kdb error/plugin error:
+Not much value for the user
+* differentiation to core features of elektra: Specification/ Notification/ KDB Core/ Bindings/ ...
+An error is again difficult to put into one bucket since some errors might be affected by multiple features (could not stat file for example
+which is used by multiple features).
+This approach would lead to the same problem as we have now (many errors, difficult to maintain)
+
+Personally I do not see how the benefit of a categorization outweighs the drawbacks because in by far the most cases users will have to look
+into the reason in more detail anyway. It should be enough for users to deal with the error just by the reason.
```
markus2330: The categorization is for applications: if it is a conflict, they can merge and kdbSet again, if the problem is a validation error, they might drop the key and try again; if the error is temporary, they might retry (retry is quite problematic though as it is unclear how often to retry and so on...)

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Line 2: Error code
+
+Reasons to keep it:
+* Probably could be used to categorize errors such it is done in the http protocol
+* Good to use for shellrecorder scripts to test for errors
+* Possibility for automation (eg. if a certain error number occurred you could automatically trigger certain actions)
+
+Reasons against it:
+* Extra line
```
markus2330: It does not need an extra line, it is only a number.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+Suggestion:
+Giving an error a number like it is done in C programs as return code seems to be outdated for elektra, especially when
+the reason is given anyway. Personally I never read the error number because it has not much value.
+The *reasons to keep* will also be very limited when the number of errors are significantly reduced.
+I would also suggest to remove it.
```
markus2330: It is needed for automated checks for specific errors (especially unit and shell recorder tests).

What we could do is to not use numbers but some ID, like the macro name or alternatively something generated (which is guaranteed to be unique). This would avoid conflicts when adding errors and make it easier for external specifications.

Downside: string comparison is needed to check for which error it is. Maybe that is not really a problem.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Line 1: General sentence
+
+All in all this sentence can stay because it gives the user a first hint on what happened.
+I would though change the wording and include more relevant information in it.
+I would change it like this:
+"Module <plugin> issued a <error/warning/fatal> while accessing the key database with the info:"
```
markus2330: Why having fatal? What is the difference?

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+Giving an error a number like it is done in C programs as return code seems to be outdated for elektra, especially when
+the reason is given anyway. Personally I never read the error number because it has not much value.
+The *reasons to keep* will also be very limited when the number of errors are significantly reduced.
+I would also suggest to remove it.
+
+### Line 1: General sentence
+
+All in all this sentence can stay because it gives the user a first hint on what happened.
+I would though change the wording and include more relevant information in it.
+I would change it like this:
```
markus2330: I would not remove the "Sorry,"

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+"Module <plugin> issued a <error/warning/fatal> while accessing the key database with the info:"
+where plugin should have a different color (blue?) and warning/error/fatal (yellow, red, dark red) too depending on the severity.
```
markus2330: Yes, I would like to have colors.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+With this line 5/6 can still be present but in a much more efficient and succinct way.
```
markus2330: I agree!

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Error/Warning/Fatal etc. splitting in the specification file
+
+I would suggest to remove this concept from the specification and let the developer use a respective method to indicate if it is a warning/error/etc.
+The macros ELEKTRA_SET_ERROR, ELEKTRA_ADD_WARNING ... should be enough to make developers cautious of wrong method usage.
```
markus2330: I agree.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+### Additional information which is not present
+
+Some errors can also guide users to fix the error by suggesting a possible solution:
+* Retry as sudo
+* Fix file permissions
+* Check if the file exists
+* etc.
+Developers though have to be very careful when suggesting solutions because it can easily mislead users. For many cases though, 
+this can help users a lot.
+I would suggest an optional extra line (which is only printed if the developer adds a metafield `solution` to the parentkey):
```
markus2330: I like the suggestion! We already have it in one case (Retry as sudo if someone mounts as non-sudo) but there are many cases where the text would be useful but is not printed (e.g. kdb set system/...)

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+5. Ingroup: plugin
+6. Module: enum
+7. At: ....../src/plugins/enum/enum.c:218
+8. Mountpoint: <parentKey>
+9. Configfile: ...../<file>.25676:1549919217.284067.tmp
+```
+
+would be changed to this:
+```
+1. Plugin enum issued a error while accessing the key database with the info:
```
markus2330: Sorry, the plugin "enum" issued error `#<NR>`:

("enum" and  `#<NR>` in colors)

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+2. Validation of key "<key>" with string "<value>" failed.
```
markus2330: Allowed values are missing.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+with optionally a third line indicating a solution. Eg. for a permission related error there would be a third line:
+```
+3. Possible Solution: Retry the command as sudo (sudo !!)
```
markus2330: Possible solution in validation errors is to try a different value.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+Benefits:
+* Much more relevant information in the error message
+* Ease of development:
+no merge conflicts, no extra line in the `specification` file leading to chaos (such as it is now)
```
markus2330: Please elaborate here what exactly the implication would be.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+No need for splitting the `specification` file because it is too large
+
+Drawbacks:
+* Possible small information loss by removing the description, `At` and `mountpoint`:
+`At` could be kept for devs by extending the `kdb set` command with an option `-v`, `--debug`, etc.
+The same applies for the `mountpoint`.
+Description comes with too many drawbacks to be kept
+* Possible misleading the user with the `Solution` line
+
+The benefits do outweigh the drawbacks. I would suggest an overloaded method like the following:
```
markus2330: overloading in C is a bit of an hack (needs macros), so better to give different names or simply return null pointers if there is no possible solution.

doc/ERROR_CONCEPT.md
```diff
@@ -0,0 +1,250 @@
...
+```
+// If a solution can be applied
+elektraError(Key parentKey, const * char message, const * char solution)
```
markus2330: Please see #2432. What you propose is the only function we already have (which is not removed).

doc/ERROR_CONCEPT.md
Piankero: I do not think that such problem is solvable here as in most languages exceptions/error messages are getting simply converted into another one (encapsulation, layering, etc.) and this can simply lead to information loss or formatting problems. It is up to the developer to do it correctly in my opinion.

Piankero: I could just think about another method which takes the strerror() as a parameter and the error logging tool handles it

doc/ERROR_CONCEPT.md
Piankero: Concerning specification related errors yes,
web gui does not show anything at all
qt gui shows a bad message (https://github.com/ElektraInitiative/libelektra/issues/2406)

doc/ERROR_CONCEPT.md
Piankero: Do you mean to include it in the message (eg. extra line saying that you can get more verbose output with `-v`)?
Isn't it common in the unix world to simply use -v when wanting more output (meaning we could potentially spare the extra line)

doc/ERROR_CONCEPT.md
Piankero: Do you want to have `-d` and `-v` both present? I think one would be enough honestly

doc/ERROR_CONCEPT.md
Piankero: But which information should be present also influences the concept in my opinion

doc/ERROR_CONCEPT.md
Piankero: I think I don't know what you mean by that. Can you give me an example how you imagine it to look like?

doc/ERROR_CONCEPT.md
Piankero: They usually are very generic and could have multiple sources of errors:
199, 196, 151, 147

Also kind difficult: 188, 177, 162

doc/ERROR_CONCEPT.md
Piankero: What is the advantage of having it in the contract?
Having a contract may seem to be useful but also causes even more overhead for developers and can cause code complication.
Personally for me it is not very intuitive to code in Elektra (specification, Metadata.ini, release_doc, internal functions, etc) and error contracts also takes it a little step further. But maybe this is common in the C world?

doc/ERROR_CONCEPT.md
Piankero: External is just meant for errors thrown by other programs (or programming languages) and caught be elektra. 
"Miscellaneous" is for errors which are hard to classify

doc/ERROR_CONCEPT.md
Piankero: I agree, I would also like to avoid it if possible.

doc/ERROR_CONCEPT.md
Piankero: I agree on your division but got some practical concerns. Beside some connection/timeout problem I do not see a lot of temporary issues (especially when looking through the specification file which errors could be temporary). And even in theses cases, usually there is a reason behind it that need to be fixed by sysadmins (proxy settings/firewall settings/power outage of router/etc.) In by far the most cases I think the sysadmin has to fix something. For those cases in which it is a real temporary problem, the user will retry it anyway I think

doc/ERROR_CONCEPT.md
Piankero: I think such solution approaches should be indicated by the "Solution: " part in the error message and not via a generic description. Users will have to look into the reason in more detail, no matter the conflict/validation error/ etc.
Therefore I do not see a big advantage of the `Description` field anymore and a possible categorization.

doc/ERROR_CONCEPT.md
Piankero: Where does this message then come from:
`Sorry, the error (#...) occurred ;(`

doc/ERROR_CONCEPT.md
Piankero: I am confused by your message:
Is it absolutely needed or not? As of now the unit tests and shell recorder tests test for a very specific error number. If we reduce the number of errors (which is the main purpose here) we would have to change unit tests anyway.

Does it make such a difference when comparing for a String or Number? If the error message changes, the test too but do error messages get changes that much?

If we could get rid of the `specification` file we can let developers handle errors much more easier.

doc/ERROR_CONCEPT.md
Piankero: There are 7 fatal messages in the specification. I do not know their right to exist in detail but it looks like those cause elektra to malfunction or useless (failed to open backend, could not initialize modules)

Some errors though do not seem to be worth a fatal like "could not compile regex" (you actually commited this 8 years ago)

doc/ERROR_CONCEPT.md
Piankero: see discussion above

doc/ERROR_CONCEPT.md
Piankero: This is a concrete error from the enum plugin, I agree that it should give you all allowed values in the message though.

doc/ERROR_CONCEPT.md
Piankero: I agree, the solution suggestion is up to the developer though

doc/ERROR_CONCEPT.md
Piankero: Which implications do you exactly mean? Those of having no specification file at all?

doc/ERROR_CONCEPT.md
Piankero: I looked through the PR and i am actually more confused than before.
Is it a naming problem or a public api visibility problem or what exactly?

doc/ERROR_CONCEPT.md
kodebach: In my opinion an error code is **absolutely essential**. There has to be a way to infallibly check which error occurred, because some errors may be fixed automatically without user input. The high-level API for example retries kdbSet, if a conflict (error code 30 I think) occurred, with the user not even knowing about it.

Think of it this way:
Most modern languages (e.g. C++ and Java) support exceptions. Certain exceptions (e.g. a timeout for an HTTP request) can be caught and fixed in some way (retry HTTP request with exponential backoff). Others might be totally ignored/suppressed or wrapped in a more general exception. For this reason it is good practice to use different exception types. You could of course always use the same type and compare the error message string, but this is not reliable. A single character changed in the error message breaks the comparison. Not even a typo could be fixed without breaking backwards compatibility. And that doesn't even consider the very common case of dynamic information in an error message. If the error message contains a filename that depends on user input (e.g. mountpoint), your only chance is to partially compare the string. This partial comparison is even more error prone, because it might match other errors as well.

C doesn't have exceptions, so other values (integers in most cases) are used to uniquely identify one type of error. Too guarantee the error code is unique, a centralised specification is the easiest way.

However, I think this specification can be vastly improved:
- Don't specify error codes explicitly. Simply start with 1 for the first error and go from there. This eliminates merge conflicts, as long as we only use macros (or other named symbols) in code.
- Don't use macros. Use `extern const int`. This way the actual error codes we use are hidden from external code and are determined by the version of Elektra used. (@markus2330 please correct me if this is wrong)
- If we also want to guarantee that the actual error codes never change, we could fixate them on each release. Either explicitly, or by using a simple test case that `git pull`s the last release and compares error codes.
- Only put the names for the `extern const int`s and a short description/comment what this error code should be used for into the specification. This allows the error message to be fully customised and an error code would be more or less equivalent to an exception type in Java or C++. The format would basically be:
  ```
  name: FILE
  description: an error related to a file, e.g. missing permissions or missing file
  ```
  From that we would generate `kdberror.h`:
  ```c
  extern const int ELEKTRA_ERROR_FILE;
  ```
  And a source file:
  ```
  #include <kdberror.h>
  
  /**
   * an error related to a file, e.g. missing permissions or missing file
   */
  const int ELEKTRA_ERROR_FILE = 1;
  ```

Piankero: What you suggest though is basically what we have now, hundreds of error codes in a centralized file to distinguish them among each other. Many error codes even share the same text.

So in your opinion this task is just to squash some errors together? This would lead to the same problems again over the long run which we have now. You want to know if there has occurred a very specific error so you can actually do counter measures. My task though was to reduce the high granularity of the error messages.

Where should this be written? 
```
name: FILE
description: an error related to a file, e.g. missing permissions or missing file
```
In the specification file? Or is the kdberror.h the new source of error codes to be checked for rather then the specification.

kodebach: > What you suggest though is basically what we have now

No, I agree with you on all parts of your concept except error codes. Like Markus said above, we don't have to use integers, but we do need some sort of unique ID.

> hundreds of error codes

We don't need hundreds of codes, but we need unique codes for some errors. I personally would group the rest thematically: file errors, broken specification, invalid configuration (i.e. not in accordance to the specification), error in binding (e.g. Java exception) etc.

Looking at the current specification I see only three things that should definitely have separate error codes:
- `CONFLICT` (30): can sometimes be resolved by retrying
- `TIMEOUT` (195): similarly might be resolved by retrying (may be with bigger timeout)
- `NETWORK` (205): An application could suppress this error, wait until a connection is available again and retry then. Of course some sort of message may still be produced to inform the user.

All of these basically come down to the same behaviour: "auto-retry". Especially, if an application is running in the background it is very nice, if the application doesn't require user input for things that might fix themselves. Auto-retry is simply not possible, if you don't have an infallible way of identifying an error.

> in a centralized file to distinguish them among each other.

I really don't care whether the file is centralized, or whether there is a file at all. If you can find a different solution to ensure unique error codes, I'm okay with that too. We could even use code 1 for conflict, 2 for timeout, 3 for network, ... (any future "auto-retry" codes) and -1 for anything else. Basically defining positive error codes are auto-retry-able, negative ones are not. I really don't care, as long as the auto-retry-able errors are infallibly distinguishable from the rest and from each other.

> Where should this be written?

Probably it would be better to just write `kdberror.h` and the corresponding source file by hand. It should only be a few lines anyway.

> This would lead to the same problems again over the long run which we have now.

As long as we make sure that each new error code introduced in a PR is very well justified, I don't think this will happen.

doc/ERROR_CONCEPT.md
markus2330: Yes, in retrospective I think the distinction is arbitrary. I would remove this severity.

doc/ERROR_CONCEPT.md
markus2330: Yes, the line exists now, but it does not have to be that way.

doc/ERROR_CONCEPT.md
markus2330: It is for different purposes: -v is to be more verbose, -d is to debug Elektra.

doc/ERROR_CONCEPT.md
markus2330: Please write this in the text, and in more detail (with examples)

#2406 is one specific problem, it is unfair to generalize this to "qt gui shows a bad message"

doc/ERROR_CONCEPT.md
markus2330: Another method will not help much. The framework should reset errno correctly and it should be part of the error/warning if it was changed by the plugin.

We should also have tests if errno is handled correctly.

doc/ERROR_CONCEPT.md
markus2330: It is not uncommon but still most people would not try it without any indication that it exists (or would waste time to find out).

The text "use `-v` for more information" is not so long and potentially helps a lot.

doc/ERROR_CONCEPT.md
markus2330: Please finish your concept first, then it is easier to discuss about such details.

doc/ERROR_CONCEPT.md
markus2330: Yes of course, but the concept should differentiate between these two cases/viewpoints.

doc/ERROR_CONCEPT.md
markus2330: Yes, but they are problematic as they are not helpful for the API user.

doc/ERROR_CONCEPT.md
markus2330: Yes, it is okay for me that we do not distinguish between temporary and permanent errors. Only the conflict, resource and validation errors need to be definitely different (as different reactions are required on these errors).

doc/ERROR_CONCEPT.md
markus2330: Implications of your whole concept.

doc/ERROR_CONCEPT.md
markus2330: Your concept should describe the whole API.

doc/ERROR_CONCEPT.md
markus2330: @kodebach wrote:
> However, I think this specification can be vastly improved:

I fully agree! These may seem as little low-level problems but they are very annoying in the day-to-day programming.

>  "auto-retry".

I think we should limit "auto-retry" to conflicts (where it is clear what to do). Otherwise auto-retries are better handled by plugins, which have domain specific knowledge (contrary to the applications).

Nevertheless, it is not only useful to distinguish "auto-retry" and "abort" but also between "resource" and "validation" errors: in one case changing the values may help, in the other not.

> Probably it would be better to just write kdberror.h and the corresponding source file by hand. It should only be a few lines anyway.

I do not agree here. I do not see how it could be reduced to "a few lines". Furthermore, the specification will provide immense benefits once we properly support other compiled languages (like C++, Java, Rust, Go, ...). Ideally @Piankero could write a translator to C++ and Java Exception so that we know that the specification has enough information to generate exceptions.

kodebach: > I think we should limit "auto-retry" to conflicts (where it is clear what to do). Otherwise auto-retries are better handled by plugins, which have domain specific knowledge (contrary to the applications).

Just to clarify: "auto-retry" was just an example of why it is sometimes useful for an application to identify the specific error that occurred. Whether or not actually doing an automatic retry makes sense in a certain application and situation is a totally different thing.

> I do not see how it could be reduced to "a few lines".

It would only be a few lines, if we only defined the error codes (plus a code comment for description) in the central file. As soon as we put error codes into categories or add other properties, the specification makes sense again.

> the specification will provide immense benefits once we properly support other compiled languages

I am not sure automatically generating code for other languages is a good solution. For one Rust and Go don't have exceptions and handle recoverable errors more like functional languages (by returning multiple values, a value and/or an error). But even if the language in question has exceptions, auto generating them is not wise. Sometimes using a standard library exception would be preferable. Sometime you might want to subclass another exception type, so you can catch many exceptions or just a special one. And all that aside, to "properly support" a language the binding has to be hand written anyway, otherwise it will always feel clunky and foreign. So you might as well write the exception classes by hand as well (maybe starting from a one-time auto-generated version).

markus2330: > "properly support" a language the binding has to be hand written anyway,

This is not the experience we have had so far. All bindings except of Java are auto-generated, and the Java binding is definitely not the best one.

> So you might as well write the exception classes by hand as well

Of course you can do that but then you risk inconsistencies.

doc/ERROR_CONCEPT.md
markus2330: It is similar to an exception specification: people know which errors they might get.

> also causes even more overhead for developers

I fully agree, so maybe better not have it.

> "and can cause code complication"

I do not understand.

> But maybe this is common in the C world?

No and I hope that your error concept will be easier to use for Elektra's developers.

doc/ERROR_CONCEPT.md
Piankero: Ok, after some longer brainstorming I came to the conclusion that we first should agree on the categories to put all current errors into. For users, these categories do not really have a special meaning because they just care about the concrete error reason anyway. That's why I suggested to remove it. The only reason is keep it is to **differ for programmatic reasons**.

Now the question is what exactly are those categories. For me, every category should dictate another solution"concept" such as, retry, unfixable, try something different, ...
My first proposal is the following:

- **Validation**: Try a different value
- **Temporary**: Retry it
- **Permanent**: Fixing is impossible or a very special case

Is there any other big category anybody knows of which would require a different approach? (@markus2330 , @kodebach , @sanssecours? Since you have the most experience I would value your input very much)

I did not see a specific difference in Temporary and Conflicts as Markus suggested in #1473 because they both are basically retry approaches but one is more "harmful".

The problem here is the granularity as sometimes even small cases such as "could not open file" (which currently would be under permanent errors) can be fixed but then would require String parsing to check for this specific error. If it is really the case that a developer wants to know a very specific error, I think it is acceptable to force String parsing. Also some errors with the same reason could fall under "temporary" (timeout) or permanent ("network driver failure"). The developer has to know in this case which category he has to choose because he has the concrete domain knowledge.

I also thought about auto generated error codes but beside the question if it is even implementable in C, it comes with other drawbacks as developers would then need to check for too many errors (eg. think of specification errors: is it an enum/path/boolean/etc. error.). To me it looks like the error concept is a fight in the level of granularity with each level coming with its own advantages/drawbacks. For me the current approach covers the most central cases with openness to new categories, should they be needed at a later point in time.

I also worked out some sort of migration of the current errors: https://docs.google.com/spreadsheets/d/1-vXNZ7pN9wlMFByIMLbFt_HieyJpop0UyksbgAvmGK4/edit?usp=sharing

I know you are all no fan of google and excel sheets but since this is still very hybrid it would take considerably longer to do it in markdown every time something changes. Once the categories + mappings are clear I will incorporate it into the README. Feel free to comment or change it as I am also not 100% sure if it is a temporary or permanent error in many cases.

markus2330: Thank you for the write-up! I think you can rework the whole error concept now. Maybe start with the goals, which should be very clear now. Furthermore it would be nice if you directly rework the error specification, the spread-sheet already looks good. I think we can split the categories a bit more:

- Validation into syntactic and semantic (and maybe even others)
- Temporary into conflicts and timeouts (watch out: only because "temporary" is in the description does not mean it is a temporary problem: 161 is a resource/permission error)
- Permanent into Resource, Parsing, Installation errors (and maybe even others)

Maybe we want sub-categories, so that applications which e.g. do not care if a validation is syntactic or semantic can write error handling code easier.

I think there is little use of discussing this even more. Simply do it! (I would start with categorizing the errors directly in the error specification. It is a harmless thing and can be merged easily as it will not break anything.)

markus2330: I forgot a very important class of errors: Assertion failures! I.e. logical errors that should never happen and are bugs within Elektra.

kodebach: I think I finally know why we disagree so much. Who is a user in your mind? 
 - A user of Elektra, i.e. a developer using Elektra in their application.

or
 - A user of an application using Elektra (e.g. the `kdb` tool).

The first kind (let's call them "application developers") will probably find specific error codes useful. The more specific an error code is, the easier you can find documentation on how to fix it. While a good description might be enough on its own, an error code always helps, if you need look at the documentation.

The second kind of users (let's call them "end users") won't find that useful. They just want a single line that tells them want they need to do "to make it work". This includes users of `kdb`, so I'm going to use it as an example.

You rightfully pointed out that `kdb`s error messages are quite bad, mainly because of the huge amount of unnecessary (and partly redundant) information. But this has nothing to do with how Elektra itself reports errors to applications. Even if we kept everything the same, `kdb` could just print

```
Error 121: validation failed
Validation of key "<key>" with string "<value>" failed.
```

instead of the 9 lines you mentioned in your example. (`-v` is used, in which case the second *verbose* version makes sense)

---

So I think the problem is twofold and we need two separate decisions here. The questions are:
1. How should Elektra (the library) report errors to applications?
2. How should `kdb` display these errors?

The second question is easy to answer: `kdb` should print one or two lines of error description. `kdb -v` should print **all** the available information, so that diagnosing errors becomes easier.

The first question is not so easy. The important thing here is to get rid of all the duplicate or very similar errors (e.g. the GPG errors), while maintaining enough granularity, so that error codes are useful for testing, diagnosing and actually handling errors. The categories are a good start, but IMO they can't replace actual error codes. We don't need an separate code for each description (in fact that is exactly what we shouldn't do), but we need some distinction.

Another good step to get rid of unnecessary error codes is IMO to not have any kind of specification for warnings. The difference between a warning and an error is that a warning can (more or less) safely be ignored, so I don't see any reason to specify warnings. If we just use `elektraAddWarning(Key * key, const char * format, ...)`, we would get rid of 84 error codes immediately. NOTE: warnings also shouldn't translate into exceptions or similar in bindings no need for specification there either

kodebach: > I think we can split the categories a bit more

> Maybe we want sub-categories

More categories plus sub-categories, might be an alternative to error codes, but I'm not quite convinced it would be a good fit for testing purposes.

> Assertion failures! I.e. logical errors that should never happen and are bugs within Elektra.

Should those even be errors? Shouldn't we just exit if an assertion fails? If this "should never happen," crashing the application shouldn't really be a problem, should it?

markus2330: I fully agree we should distinguish of which error information we emit and which parts of this information is shown to whom.

> Should those even be errors? Shouldn't we just exit if an assertion fails?

The application should exit but a well behaved library should not exit. (The infallible APIs as exception.) So I think we need (assertion) errors for the situations where admins installed Elektra incorrectly or the KDB is completely broken (random stuff in system/elektra).

Piankero:
@markus2330, I rewrote the concept. Please give your final review so I can finally start.
My plan are the following Milestones (also separate PRs to avoid millions of merge conflicts):

- [ ] Remove warnings
- [ ] Add the new Solution API
- [ ] Migrate all errors according to the categories

doc/decisions/error_message.md
```diff
@@ -0,0 +1,133 @@
...
+Warnings will be removed from the specification file as they are not needed there.
+
+All "fatal" errors will be converted to "errors" as the distinguishment is not relevant.
+
+Unused marked errors will be removed from the specification.
+
+The remaining ~130 errors will be categorizes into logical groups with subgroups.
+Each group and subgroup will receive a range of numbers such as in the HTTP protocol.
+
+* Permanent errors (1-500)
```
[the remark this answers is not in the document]
Piankero: Thank you for that input. Your idea also works nice. But if I understood correctly, theses error codes would then be strings because permanent errors would still have digits but something like 0100 for Resource errors. My thaught is that devs check for error codes like >=100 && <200 instead of something like charAt(0) == 0 && charAt(1) == 1 (or startsWith(01)).

Lets wait if @markus2330 has another idea/input on this :). For me, both approaches are equally fine.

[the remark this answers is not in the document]
markus2330: I would prefer if users do not have to fiddle around with numbers. @Piankero did you already did a study what other FLOSS projects are doing? You can write the comparison as thesis text.

In the next meeting we can do a brain storm of which projects have nice error concepts.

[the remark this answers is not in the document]
Piankero: Not yet, do you have a preference towards certain projects? Do they have to be similar to elektra (C/C++, Medium Large, Well Maintained, etc.)?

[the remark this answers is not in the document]
markus2330: I think we should look into projects that also have a generic plugin system where it is not so clear which kind of errors plugins might have to yield, e.g.: GStreamer, Apache http, Jenkins, Databases (MariaDB, PostgresQL).

Also relevant would be standards like SNMP.

Looking at projects that are to some extend alternatives to Elektra, like etcd, Windows Registry or Mac OS X plists also cannot harm.

Of course you should also look into POSIX errors (errno), although we definitely do not want to go back there as it is too primitive. (It does not give any reason or any further helpful information.)

As the error concepts are usually very short or non-existent, I think you can look at more than just a few.

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
+#Error message & handling concept
+
+## Problem
+
+The current error concept has disadvantages in following regards:
+
+- Too verbose error message
+- A lot of redundant errors
+- Hard to manage specification file
```
markus2330: Write more about the goals. What does "Hard to manage" mean? Do we want the specification to be more static or do we want that adding errors is without merge conflicts?

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+- No senseful way for application developers to use error codes from elektra
+
+This concept should improve useability for end users and developers.
```
markus2330: Please also consider administrators, who will also read the error messages and are also capable of resolving resource errors and similar.

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Constraints
+
+- Error numbers must stay because they are more reliable to check against than strings
```
markus2330: There are many more constraints, e.g.:

- supporting multiple programming languages
- constraints from the plugin system
- ...

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Assumptions
+
+## Considered Alternatives
+
+- Removing the specification file without requiring error numbers
```
markus2330: Here can you be much more concrete after you looked into other projects. Basically every project you look into will be a "considered alternative".

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+- Possible variations on what message should be displayed
```
markus2330: This sentence does not say much.

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+```
+Possible Solution: Retry the command as sudo (sudo !!)
+```
+
+To avoid losing information, the user can use the command line argument `-v` (verbose) to show
+`Mountpoint`, `Configfile` in addition to the current error message.
+Furthermore a developer can use the command line argument `-d` (debug)
+to show `At` for debugging purposes.
+
+Warnings will be removed from the specification file as they are not needed there.
```
markus2330: You mean the information if something is a warning or error?

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+
+All "fatal" errors will be converted to "errors" as the distinguishment is not relevant.
+
+Unused marked errors will be removed from the specification.
```
markus2330: It would be nice if we get a checker tool if warnings/errors are used (simple grep of macro should suffice).

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+
+The remaining ~130 errors will be categorizes into logical groups with subgroups.
+Each group and subgroup will receive a range of numbers such as in the HTTP protocol.
+
+- Permanent errors (1-500)
```
markus2330: I like @kodebach approach to use digits for the classification.

But even more important is some underlying logic why the classification is done how it is.

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+  - Resource (1-50)
+  - Parsing (51 - 100)
+  - Installation (101 - 150)
+  - Logical (151 - 200)
+- Temporary (501 - 1000)
+  - Conflict (501 - 600)
```
markus2330: Conflict is not really temporary: it is permanent until the conflict is resolved...

doc/decisions/error_message.md
```diff
@@ -0,0 +1,137 @@
...
+
+The new error message is much more succinct which gives end users more relevant information.
+Furthermore the solution approach still holds all necessary information if requested by users.
+
+A new metakey `error/solution` will be implemented and attached to the parentkey via the new API.
+`Ingroup`, `Description` and `Module` will be removed from the error message as they provide no useful
+information at all.
+
+The grouping of errors will allow developers to filter for specific as well as more general errors to correctly
+react to them programmatically. Even though there are currently just 8 categories they will reserver up to 1500 error numbers.
```
markus2330: reserver?

doc/decisions/error_message.md
Piankero: A few days ago you wrote this:

> I think we can split the categories a bit more:
>...
>* Validation into syntactic and semantic (and maybe even others) 
>* Temporary into conflicts and timeouts...

What exactly should I take now. In one sentence you write I should split it and now you tell me that I should not do it?

doc/decisions/error_message.md
Piankero: Warnings will be removed as a whole from the specification file. Warnings per se will still exist but will not receive error codes. Therefore it does not make sense to include it in the file, an API call is enough

doc/decisions/error_message.md
Piankero: Where exactly is the difference between an administrator and an end user? They both use elektra via gui/cmd/etc and fix a problem once it occurs. A developer reacts to error codes or uses the elektra API to emit an error.

doc/decisions/error_message.md
markus2330: I do not see the contradiction you see here.

"[split] Temporary into conflicts and timeouts" means that there should be two categories: conflicts and timeouts.

doc/decisions/error_message.md
markus2330: Of course there is overlap but I would not see an end user in the position to fix most resource error (e.g. resizing disks, changing permissions). So an application might add the text "please contact your system administrator" if an resource error occurs.

Having resource/conflict/temporary errors in different categories, however, is maybe the only requirement we need to add in order to consider system administrators.

doc/decisions/error_message.md
markus2330: Unfortunately you were not here today to discuss this. I currently do not understand why we should throw away the categorization of warnings. It would destroy the ability to convert errors to warnings (and vice versa). I think we should keep warning and error compatible (or even have them more compatible: every error would be a valid warning and vice versa).

doc/decisions/error_message.md
Piankero: How does it exactly differ when supporting multiple programming languages?
What does the plugin system exactly constraint in this case?

doc/decisions/error_message.md
Piankero: I understood it the way that I should make a further subdivision but keep the overall category. I understand it now

doc/decisions/error_message.md
markus2330: > How does it exactly differ when supporting multiple programming languages?

The errors should be in a way so that the specification can be mapped to exceptions (or other ways different programming languages handle errors).

> What does the plugin system exactly constraint in this case?

In many ways, for example, multiple plugins might try to set an error and the system needs to decide which is the error to be shown to the user. Hopefully you learn such constraints from related work.

doc/decisions/error_message.md
```diff
@@ -65,6 +65,17 @@ Various FLOSS projects:
...
+* [Windows Registry](https://docs.microsoft.com/en-us/windows/desktop/sysinfo/registry):
+    The registry does not use any specific error concept but takes the standard [Win32 Error Codes](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/18d8fbe8-a967-4f1c-ae50-99ca8e491d2d). These are neither hierarchical nor have any special ordering. Basically it is the same as elektra has now except for no duplicated
+    errors.
+* Mac OS X plist:
+    Just like Windows, plist uses standard Mac OS X errors which is a [huge catalog](http://krypted.com/lists/comprehensive-list-of-mac-os-x-error-codes/) of unordered
+    return codes as integers.
+* [SNMP Standard](http://www.snmp.com/protocol/):
+    Being a standard network protocol, error codes are very specific to the domain itself. A list can be found [here](https://docs.microsoft.com/en-us/windows/desktop/snmp/snmp-error-codes) and would not meet the needs of elektra at all.
```
markus2330: Where do you see that the error codes are specific to the domain? With a few exceptions (e.g. SNMP_ERRORSTATUS_AUTHORIZATIONERROR) I think most of the error codes have some equivalent in Elektra.

Piankero: UNDOFAILED, WRONGLENGTH, TOOBIG, WRONGENCODING, INCONSISTENTNAME, ... they might all could be mapped to some elektra class but in my opinion these are completely different domains. While SNMP emphasizes on errors in packets, elektra does on handling key value pairs.

Piankero:
I reworked the concept as you desired @markus2330 and also investigated other projects. I liked the SQLSTATES and incorporated it into the concept. The categories as of now did not change because I did not see the need for it when going through other error categories.
If we ever see the need for another category it is very easy to change.

doc/decisions/error_message.md
Piankero: I am a bit confused concerning "Conflicts" even though when reading through #1473. What kind of errors do you imagine under this category? Can you give some examples?

doc/decisions/error_message.md
```diff
@@ -0,0 +1,215 @@
...
+
+The classification as of now is done with the following concept in mind:
+
+- Permanent errors
+    Generally errors which require changes by a user which are not getting fixed by eg. retrying.
+- Resource
+    The class of errors which require certain permissions (opening a file) or where some directory/file is missing.
+    One possible reaction to these kind of errors is to try a different directory/file or create a missing resource.
+- Parsing
+    Errors while parsing files such as ini/yml/etc.
```
markus2330: yml?

doc/decisions/error_message.md
```diff
@@ -0,0 +1,215 @@
...
+
+```
+// Already present
+ELEKTRA_SET_ERROR (nr, parentKey, message);
+ELEKTRA_SET_ERRORF(nr, parentKey, message, ...);
+ELEKTRA_ADD_WARNING(nr, parentKey, message);
+ELEKTRA_ADD_WARNINGF(nr, parentKey, message, ...)
+
+// Extended API
+ELEKTRA_SET_ERROR_WITH_SOLUTION(nr, parentKey, solution, message)
```
markus2330: With this approach we would not have format specifiers for solutions and it would also blow up the size of kdberrors.h (maybe you even find a way how to make the file shorter).

What about having a macro `ELEKTRA_ADD_SOLUTION` which adds a solution message to the last error/warning? This would also allow something like src/tools/kdb/mountbase.cpp very bottom.

markus2330:
Btw. the related work is nicely done!

doc/decisions/error_message.md
```diff
@@ -0,0 +1,215 @@
...
+
+All "fatal" errors will be converted to "errors" as the distinguishment is not relevant.
+
+Unused marked errors will be removed from the specification.
+
+Errors will be categorizes into logical groups with subgroups.
+Each error will be made up of 5 characters, where the first 2 character indicate the highest level
+and character 3 to 5 will be used for subgrouping.
+
+- Permanent errors (01000)
```
kodebach: Use `"01000"` instead of `01000`. Otherwise it looks like a C-style octal number.

doc/decisions/error_message.md
```diff
@@ -0,0 +1,215 @@
...
+Description: validation failed
+Reason: Validation of key "<key>" with string "<value>" failed.
+Ingroup: plugin
+Module: enum
+At: ....../src/plugins/enum/enum.c:218
+Mountpoint: <parentKey>
+Configfile: ...../<file>.25676:1549919217.284067.tmp
+```
+
+The new default error message will look like this:
```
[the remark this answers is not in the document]
Piankero: I slightly changed the message. Warnings and Errors are as u suggested interchangeable and should therefore also share the same message format

doc/decisions/error_message.md
```diff
@@ -0,0 +1,215 @@
...
+
+```
+Sorry, plugin <PLUGIN> issued error #<NR>:
```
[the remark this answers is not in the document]
Piankero: Since the new error is a String I would not prefix it with a letter.
Why not take a more straightforward approach and use `... issued error code XXXXX`?

doc/decisions/error_message.md
Piankero: yml and yaml are more or less interchangeable. But the official documentation suggests yaml so I will take it

doc/decisions/error_message.md
Piankero: Concerning the "blow up" part I already talked about with @kodebach. With some refactoring we could drastically reduce the code duplication generated by the cpp file.

Yes, we could add another macro which could do that but would also break the atomicity then. Eg. plugin a and b throw an independent error. Plugin a also wants to add a solution. First plugin a writes to the parent key. Then plugin b follows immediately. Plugin a then adds its solution which would not fit together then

kodebach: > maybe you even find a way how to make the file shorter

I got some ideas about this while working on the highlevel code generation. I already sent them to @Piankero, but I will post it here again, in case you have any input:

**Solution 1:** 1 macro + 1 short static inline function per error code

We start with a header file that contains the basic macros (`ELEKTRA_SET_ERROR` etc):
```c
#define ELEKTRA_SET_ERROR(code, key, message, <...>) elektraSetError##code (key, message, <...>)
```

For each error code we append something like the following:
```c
#define ELEKTRA_ERROR_SOMETHING 01001

static inline void elektraSetError01001 (Key * key, const char * message, <...>) {
    keySetMeta(key, "error", "01001");
   [...]
    elektraSetErrorInformation (message, <...>);
}
```

`elektraSetErrorInformation` sets the common attributes of an error. It should be declared in the generated header (or one that is included in the generated file), and defined in a separate C file, so that we don't duplicate the code everytime `kdberrors.h` is included.

The `<...>` part is a placeholder for the remaining dynamic parameters needed for error creation. The `[...]` part is a placeholder for the remaining static properties of an error given in the specification.

**Solution 2:** 2(?) macros per error code

By abusing the preprocessor a bit, we could also do everything without the inline functions.
The reason the inline functions exist in the current (AFAIK) and the above solution, is so that we get compiler errors, if `ELEKTRA_SET_ERROR` is used with unspecified error codes. The following fulfills the same purpose:

Like above we start with a header containing the basic macros. But instead of simply calling the inline function we do something like:

```c
#define ELEKTRA_ERROR_PROPS(code) ELEKTRA_ERROR_PROPS_##code

#define ELEKTRA_SET_ERROR(code, key, message, <...>) \
    do {                                             \
        ELEKTRA_ERROR_PROPS(code);                   \
        elektraSetErrorInformation(message, <...>);  \
   } while(0)
```

```c
#define ELEKTRA_ERROR_SOMETHING 01001

#define ELEKTRA_ERROR_PROPS_01001(key)     \
        keySetMeta(key, "error", "01001"); \
        [...]
```
We might need additional helper macros so the substitution is done correctly

Beacuse `ELEKTRA_ERROR_PROPS_##code` only exists for specified we will still get a compiler error for undefined error codes. The advantage of using macros is that they are definitely inlined, although the static inline functions should also be inlined in optimized builds.

> What about having a macro ELEKTRA_ADD_SOLUTION which adds a solution message to the last error/warning?

I think that is better than having `ELEKTRA_SET_ERROR_WITH_SOLUTION`,  but I would prefer a function (should reduce binary size, if not static) instead of a macro, unless you know a reason why a function wouldn't work.

kodebach:
> For the API there are still discussions (SOLUTION) and extensions (high-level API?) needed.

The current error codes of the high-level API could simply integrated into the new concept. Either as their own category or otherwise marked with some attribute. The `ElektraError` struct of the high-level API would be replaced by one that contains all the information of the new errors. For all errors belonging to the high-level API we would also generate functions that directly create the appropriate `ElektraError` struct.

This would also be good test for whether the concepts is usable, if a different error reporting framework (other than an `errorKey`) is used.

doc/decisions/error_message.md
kodebach: > Eg. plugin a and b throw an independent error.

Plugins are always called sequentially by Elektra, so this situation cannot normally arise. Elektra is also explicitly not thread safe, so if parallelism occurs the developer has to handle all problems that may arise.

As long as the plugin doesn't try to do something stupid like setting an error in kdbGet and the solution in kdbSet, everything should be fine.

markus2330: > Plugins are always called sequentially by Elektra, so this situation cannot normally arise. 

There are some situations, e.g. if the plugins within the "error" position emit further errors or multiple notification fail (which happens after commits).

Of course we can prohibit to emit such errors in such situations but that should be explained in guidelines hopefully written by @Piankero

markus2330:
I approve the categories for now. Please split up the decision in further files in the next PR.

You can start working recategorize all errors.

Piankero:
> > For the API there are still discussions (SOLUTION) and extensions (high-level API?) needed.
> 
> The current error codes of the high-level API could simply integrated into the new concept. Either as their own category or otherwise marked with some attribute. The `ElektraError` struct of the high-level API would be replaced by one that contains all the information of the new errors. For all errors belonging to the high-level API we would also generate functions that directly create the appropriate `ElektraError` struct.
> 
> This would also be good test for whether the concepts is usable, if a different error reporting framework (other than an `errorKey`) is used.

After categorizing some errors I would kind of tend to have an own category because all errors from other languages could also be included there (python, java, etc.)

Maybe we could call is `Language error` or `API error`?

markus2330:
There are already high-level API specific errors and other APIs should reuse these errors. Why do you want to also have API errors in the low-level errors?

Piankero:
Ok, well then the concept of the high level API is not clear yet to me how it works or is implemented.

I still wonder though where to put language errors into which come from java/python/lua/etc.
They do not fit anywhere as of now

markus2330:
Please specify what you mean with "language errors". In which language a plugin is implemented should not matter at all. And everything else is API?

Piankero:
Can elektra (or C?) differentiate between the kind of errors that are thrown from eg. java/python/etc.?
Imagine 2 python plugins written in python. One throws an error for resource permission while the other one throws an exception for validation.

markus2330:
No, you cannot distinguish and I do not see why this should be wanted. If plugins are implemented correctly (according the metadata specification) they should behave the same (with the same errors) and the implementation language should be irrelevant.

Piankero:
I agree but as of now there are several errors which are very generic, eg 111:python error.
Where do I put those into? Installation Error?

markus2330:
Thank you, now I see what you mean.
These are either:

- implementation issues of the plugin (basically a failed ASSERT), or
- plugin configuration errors/installation errors, i.e., the script (which might be part of Elektra's installation) is not present, or
- (in extremely rare cases) out-of-memory errors

We should separate at least the first two kind of errors, as they belong to different categories.

markus2330:
@kodebach wrote:

> The current error codes of the high-level API could simply integrated into the new concept. Either as their own category or otherwise marked with some attribute. The ElektraError struct of the high-level API would be replaced by one that contains all the information of the new errors. For all errors belonging to the high-level API we would also generate functions that directly create the appropriate ElektraError struct.

> This would also be good test for whether the concepts is usable, if a different error reporting framework (other than an errorKey) is used.

I fully agree!

@Piankero the discussion here is already very confusing. Can you please make a new issue out of all these points within this PR so that you forget no part of it?

markus2330:
Alternatively you can make your own Project about "Error codes" and create issues for all (major) tasks to be done.

doc/decisions/error_message.md
markus2330: I like @kodebach idea to reduce the size of kdberrors.h. This should also be part of the new concept. But as said below, we need a project or some other means to keep track of these proposals.

doc/decisions/error_message.md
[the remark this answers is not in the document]
markus2330: Because people copy&paste only the XXXXX and sometimes the information if it was a warning/error gets lost then. Please also create an issue about that.

doc/decisions/error_message.md
markus2330: I heard of yml being a file extension (maybe to be compatible to DOS 8.3 file names) but never that yml is used to describe the file format.

doc/decisions/error_message.md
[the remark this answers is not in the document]
Piankero: Well, the concept relies on the fact that errors and warning are interchangeable like you requested. It also does not make a considerable difference if it is a Resource "Warning" or "Error" for example. The message itself states if it was an error or warning and is even color highlighted. Honestly I would not prefix a code with "W" or "E" just because user could potentially lazy copy an error code. The error code itself also is primarily used for programmatic reaction rather than for users to copy. Users should rely on the message. This is what we have talked all about in this PR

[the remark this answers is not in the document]
markus2330: Ok.

doc/decisions/error_message.md
markus2330: Conflict is one kind of error: when kdbSet() fails because some external modification was done. The solution is that the application needs to do kdbGet() afterwards.

The conflict happens because we are optimistic: we assume we are the only one working with the configuration files. If this assumption does not hold, a conflict occurs.

PR #2443, opened 2019-02-27
 Yan LR: Check if ANTLR 4 Version Supports ANTLR’s C++ Runtime

sanssecours:
Before this update building Elektra would fail, if you included the [Yan LR](https://www.libelektra.org/plugins/yanlr) plugin and installed a version of ANTLR 4 that does not support the C++ runtime. Thank you to Klemens (@kodebach), who reported this bug in today’s meeting.


markus2330:
Thank you!

PR #2445, opened 2019-02-28
Markdown Link Converter: Change Location Style

sanssecours:


markus2330:
The previous style was created for some other editor. But I am okay with this change if it works better for you.

sanssecours:
> The previous style was created for some other editor.

Interesting. Do you remember the name of the editor? The location style

    file:line:column

should be pretty much the standard format. For example, it is also the default style for error messages, if you compile some files with `clang` or `gcc`.

markus2330:
> For example, it is also the default style for error messages, if you compile some files with clang or gcc.

Yes, this was the reasons. Isn't it better to keep it, then?

sanssecours:
> Yes, this was the reason. Isn't it better to keep it, then?

Your version of `clang` and `gcc` uses

```
filename|line col 0|
```

as default error style? At least 

- [Clang on FreeBSD](https://cirrus-ci.com/task/4923839935938560), 
- [Clang on macOS](https://cirrus-ci.com/task/4783102447583232), 
- [GCC on macOS](https://travis-ci.org/ElektraInitiative/libelektra/jobs/499855597), 
- [Clang on Ubuntu](https://travis-ci.org/ElektraInitiative/libelektra/jobs/499855600), and
- [GCC on Debian](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2445/4/pipeline)

seem to use

```
file:line:column
```

by default, according to various build jobs.

doc/news/_preparation_next_release.md
```diff
@@ -120,7 +120,22 @@ you up to date with the multi-language support provided by Elektra.
...
+  filename:line:0
+  ```
+
+  instead of
+
+  ```
+  filename|line col 0|
+  ```
+
+  to show the location data for broken links. This has the advantage, that certain tools such as [TextMate](https://macromates.com) are
```
markus2330: As discussed: please write here that this is a default format for gcc, clang, ...

src/libs/highlevel/README.md
```diff
@@ -9,7 +9,7 @@ should be hard to use it in a wrong way. This tutorial gives an introduction for
 using the high-level API.
 
 The API supports all CORBA Basic Data Types, except for `wchar`, as well as the `string` type
-(see also [Data Types](../high-level-api#data-types) below).
+(see also [Data Types](#data-types) below).
```
markus2330: This does not work on the homepage (also not the previous version)

https://www.libelektra.org/tutorials/high-level-api

markus2330:
Thank you! Can you open an issue or PR for the broken links please?

PR #2446, opened 2019-02-28
examples: added some specification examples

Piankero:
As promised I add my written specifications for cassandra and lcdproc.
@kodebach thankfully wrote lcdexec.ini

Please note that some checks are intentionally commented out because of the max plugin error.
With this they can at least still be spec-mounted.

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -183,7 +183,7 @@ you up to date with the multi-language support provided by Elektra.
   location data, providing additional features, such as clickable links to the error source. _(René Schwaiger)_
 
 - We added a badge for [LGTM](https://lgtm.com) to the [main ReadMe file](https://master.libelektra.org/README.md). _(René Schwaiger)_
-- Added [LCDproc](http://lcdproc.omnipotent.net/) and [Cassandra](http://cassandra.apache.org/) specification examples. These examples provide a good   guideline for writing specifications for configurations. _(Michael Zronek)_
+- Added [LCDproc](../../examples/spec/lcdproc) and [Cassandra](../../examples/spec/cassandra.ini) specification examples. These examples provide a good   guideline for writing specifications for configurations. _(Michael Zronek)_
 
 - <<TODO>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -183,7 +183,7 @@ you up to date with the multi-language support provided by Elektra.
   location data, providing additional features, such as clickable links to the error source. _(René Schwaiger)_
 
 - We added a badge for [LGTM](https://lgtm.com) to the [main ReadMe file](https://master.libelektra.org/README.md). _(René Schwaiger)_
-- Added mores specification examples. _(Michael Zronek)_
+- Added [LCDproc](http://lcdproc.omnipotent.net/) and [Cassandra](http://cassandra.apache.org/) specification examples. These examples provide a good guideline for writing specifications for configurstions. _(Michael Zronek)_
 
 - <<TODO>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -183,7 +183,7 @@ you up to date with the multi-language support provided by Elektra.
   location data, providing additional features, such as clickable links to the error source. _(René Schwaiger)_
 
 - We added a badge for [LGTM](https://lgtm.com) to the [main ReadMe file](https://master.libelektra.org/README.md). _(René Schwaiger)_
-- Added [LCDproc](http://lcdproc.omnipotent.net/) and [Cassandra](http://cassandra.apache.org/) specification examples. These examples provide a good guideline for writing specifications for configurstions. _(Michael Zronek)_
+- Added [LCDproc](http://lcdproc.omnipotent.net/) and [Cassandra](http://cassandra.apache.org/) specification examples. These examples provide a good   guideline for writing specifications for configurations. _(Michael Zronek)_
 
 - <<TODO>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -183,7 +183,8 @@ you up to date with the multi-language support provided by Elektra.
   location data, providing additional features, such as clickable links to the error source. _(René Schwaiger)_
 
 - We added a badge for [LGTM](https://lgtm.com) to the [main ReadMe file](https://master.libelektra.org/README.md). _(René Schwaiger)_
-- Added [LCDproc](../../examples/spec/lcdproc) and [Cassandra](../../examples/spec/cassandra.ini) specification examples. These examples provide a good   guideline for writing specifications for configurations. _(Michael Zronek)_
+- Added [LCDproc](../../examples/spec/lcdproc) and [Cassandra](../../examples/spec/cassandra.ini) specification examples. These examples 
+  provide a good guideline for writing specifications for configurations. _(Michael Zronek)_
 
 - <<TODO>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -183,7 +183,7 @@ you up to date with the multi-language support provided by Elektra.
   location data, providing additional features, such as clickable links to the error source. _(René Schwaiger)_
 
 - We added a badge for [LGTM](https://lgtm.com) to the [main ReadMe file](https://master.libelektra.org/README.md). _(René Schwaiger)_
-- Added [LCDproc](../../examples/spec/lcdproc) and [Cassandra](../../examples/spec/cassandra.ini) specification examples. These examples 
+- Added [LCDproc](../../examples/spec/lcdproc) and [Cassandra](../../examples/spec/cassandra.ini) specification examples. These examples
   provide a good guideline for writing specifications for configurations. _(Michael Zronek)_
 
 - <<TODO>>
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -184,6 +184,7 @@ you up to date with the multi-language support provided by Elektra.
 
 - We added a badge for [LGTM](https://lgtm.com) to the [main ReadMe file](https://master.libelektra.org/README.md). _(René Schwaiger)_
 - Added mores specification examples. _(Michael Zronek)_
+
 - <<TODO>>
 
 ## Tests
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -183,6 +183,7 @@ you up to date with the multi-language support provided by Elektra.
   location data, providing additional features, such as clickable links to the error source. _(René Schwaiger)_
 
 - We added a badge for [LGTM](https://lgtm.com) to the [main ReadMe file](https://master.libelektra.org/README.md). _(René Schwaiger)_
+- Added mores specification examples. _(Michael Zronek)_
 - <<TODO>>
 
 ## Tests
--- a/examples/spec/cassandra.ini
+++ b/examples/spec/cassandra.ini
@@ -1,14 +1,21 @@
-# Cassandra storage config YAML
+[]
+mountpoint = cassandra.ini
+infos/plugins = dump
 
-# NOTE:
-#   See http://wiki.apache.org/cassandra/StorageConfiguration for
-#   full explanations of configuration directives
-# /NOTE
 
+[/cluster_name]
+;type = string
+check/type = string
+default = "Test Cluster"
+description = "
 # The name of the cluster. This is mainly used to prevent machines in
-# one logical cluster from joining another.
-cluster_name: 'Test Cluster'
+# one logical cluster from joining another."
 
+[/num_tokens]
+;type = unsigned_long
+check/type = unsigned_long
+default = 1
+description = "
 # This defines the number of tokens randomly assigned to this node on the ring
 # The more tokens, relative to other nodes, the larger the proportion of data
 # that this node will store. You probably want all nodes to have the same number
@@ -22,8 +29,13 @@ cluster_name: 'Test Cluster'
 #
 # If you already have a cluster with 1 token per node, and wish to migrate to 
 # multiple tokens per node, see http://wiki.apache.org/cassandra/Operations
-num_tokens: 256
+"
 
+[/allocate_tokens_for_keyspace]
+;type = string
+check/type = string
+default = "KEYSPACE"
+description = "
 # Triggers automatic allocation of num_tokens tokens for this node. The allocation
 # algorithm attempts to choose tokens in a way that optimizes replicated load over
 # the nodes in the datacenter for the replication strategy used by the specified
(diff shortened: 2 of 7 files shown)


examples/spec/cassandra.ini
```diff
@@ -0,0 +1,1978 @@
+[]
+mountpoint = cassandra.ini
+infos/plugins = dump
```
markus2330: dump format for an ini file?

examples/spec/cassandra.ini
```diff
@@ -0,0 +1,1978 @@
...
+
+
+[/cluster_name]
```
markus2330: beginning slash?

examples/spec/cassandra.ini
```diff
@@ -0,0 +1,1978 @@
...
+;type = string
```
markus2330: Why?

examples/spec/cassandra.ini
```diff
@@ -0,0 +1,1978 @@
...
+default = AES/CBC/PKCS5Padding
+
+[/transparent_data_encryption_options/key_alias]
+;type = string
+check/type = string
+default = testing:1
+
+[/transparent_data_encryption_options/key_provider]
+
+[/transparent_data_e
```
markus2330: Seems to be incomplete?

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
+[]
+mountpoint = LCDd.conf
+;infos/plugins = dump validation enum range ipaddr path network type
+infos/plugins = dump validation enum
```
markus2330: dump would be completely incompatible to current format

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
...
+
+;=============================================================================
+;=============================================================================
+
+;Affecting all subsettings, eg curses/Contrast, CFontz/Contrast, etc.
+;------------------------------------
+
+[/_/contrast]
+type = unsigned_short
+check/type = unsigned_short
```
markus2330: needed?

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
...
+[/_/offbrightness]
+type = unsigned_short
+check/type = unsigned_short
+check/range = 0-1000
+
+[/_/reboot]
+type = boolean
+check/type = boolean
+default = no
+description = Reinitialize the LCD`s BIOS normally you shouldn`t need this
```
markus2330: Two sentences?

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
...
+
+[/_/offbrightness]
+type = unsigned_short
+check/type = unsigned_short
+check/range = 0-1000
+
+[/_/reboot]
+type = boolean
+check/type = boolean
+default = no
```
markus2330: Needs boolean plugin?

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
...
+    g15, glcd, glcdlib, glk, hd44780, icp_a106, imon, imonlcd,, IOWarrior,
+    irman, joy, lb216, lcdm001, lcterm, linux_input, lirc, lis, MD8800,
+    mdm166a, ms6931, mtc_s16209x, MtxOrb, mx5000, NoritakeVFD,
+    Olimex_MOD_LCD1x9, picolcd, pyramid, rawserial, sdeclcd, sed1330,
+    sed1520, serialPOS, serialVFD, shuttleVFD, sli, stv5730, svga, t6963,
+    text, tyan, ula200, vlsys_m428, xosd, yard2LCD"
+
+[/server/bind]
+type = string
+;check/ipaddr =
```
markus2330: say why it is commented out

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
...
+check/type = string
+default = 127.0.0.1
+description = Tells the driver to bind to the given interface
+
+[/server/port]
+type = unsigned_short
+;check/port/listen =
+check/type = unsigned_short
+default = 13666
+description = Listen on this specified port. [default: 13666]
```
markus2330: metadata duplicated

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
...
+
+[/server/reportlevel]
+type = unsigned_short
+check/type = unsigned_short
+check/range = 0-5
+default = 2
+description = TSets the reporting level, defaults to warnings and errors only.
```
markus2330: TSets?

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2752 @@
...
+type = boolean
+check/type = boolean
+default = no
+description = Should we report to syslog instead of stderr?
+
+[/server/user]
+type = string
+check/type = string
+default = nobody
+description = "ser to run as.  LCDd will drop its root privileges and run as this user instead. [default: nobody]"
```
markus2330: ser?

examples/spec/cassandra.ini
Piankero: I think I copied this from somewhere here, it has actually no effect I think

examples/spec/cassandra.ini
Piankero: Does is make a difference? actually it doesn't affect the code but I can remove it

examples/spec/cassandra.ini
Piankero: From the Pull Request Description:
> Please note that some checks are intentionally commented out because of the max plugin error.
With this they can at least still be spec-mounted.

examples/spec/lcdproc/LCDd.ini
Piankero: see reason above, I changed it

examples/spec/lcdproc/LCDd.ini
Piankero: Contrast can only range from 1-1000. If the range plugin automatically checks for the type too it doesn't matter i guess.

examples/spec/lcdproc/LCDd.ini
Piankero: Copied from the original config was like this:
```
# Reinitialize the LCD's BIOS [default: no; legal: yes, no]
# normally you shouldn't need this
```
I removed the `[..]` part because it is actually visible in other metadata anyway.
After reevaluation I think it is still good to have it in the description.

examples/spec/lcdproc/LCDd.ini
Piankero: It does not load the boolean plugin? I want the type plugin solely:

> The type checker plugin supports all basic CORBA types: ... , boolean, ... When checking any it will always be successful, regardless of the content.

examples/spec/lcdproc/LCDd.ini
Piankero: From the Pull Request Description:

> Please note that some checks are intentionally commented out because of the max plugin error.
> With this they can at least still be spec-mounted.

examples/spec/lcdproc/LCDd.ini
Piankero: Honestly I kept it there because users can instantly see by the description what the default value is and might not need to look for the default value metadata. 
Concerning maintenance of course it is better to remove it because you cannot forget to update the description

examples/spec/cassandra.ini
Piankero: actually not:
The description of the yaml contained lines affecting multiple subsettings
```
# .... some long description affecting all subsettings and explaining everything
transparent_data_encryption_options:
    enabled: false
    chunk_length_kb: 64
    cipher: AES/CBC/PKCS5Padding
    key_alias: testing:1
    # CBC IV length for AES needs to be 16 bytes (which is also the default size)
    # iv_length: 16
    key_provider: 
      - class_name: org.apache.cassandra.security.JKSKeyProvider
        parameters: 
          - keystore: conf/.keystore
            keystore_password: cassandra
            store_type: JCEKS
            key_password: cassandra
```
With this method I have the description for the subsettings too without duplication
Unfortunately you only see the description until you set a value (#2448)

examples/spec/cassandra.ini
markus2330: As said in the PR description, such information should be within the repo, not in Pull Request Descriptions.

examples/spec/cassandra.ini
markus2330: It is style, so it makes a pragmatic difference.

Unfortunately the style is not described in doc/CODING.md yet but you can see the other examples.

examples/spec/lcdproc/LCDd.ini
markus2330: Without the boolean plugin, only "1" and "0" are valid.

Piankero: You are right, still I cannot load the boolean plugin though (#2451)

Piankero:
@markus2330 I changed everything accordingly, I hope this time it is fine

examples/spec/lcdproc/LCDd.ini
Piankero: When looking into the plugin it does not state that "no/yes" are actually valid options.
Only `on/true`, `on/false`, `on/invalid`, `on/invalid/warning,`true`,`false`.
So I have to change the yes/no to enums I think

markus2330: You can also introduce yes/no as valid values in the boolean plugin. It seems like these values make sense, too.

markus2330:
The build server fails and "ready to merge" is not set.

examples/spec/lcdproc/LCDd.ini
```diff
@@ -0,0 +1,2712 @@
...
+type = boolean
+check/type = boolean
+default = no
+description = The server will stay in the foreground if set to yes
+
+[server/hello]
+type = string
+check/type = string
+default = builtin
+description = "Hello message: each entry represents a display line; default: builtin"
```
markus2330: still duplicated defaults

Please review yourself properly.

examples/spec/lcdproc/LCDd.ini
markus2330: `type` matters, as it will be the type @kodebach gets. `check/type` should not be needed here.

kodebach: One note on types: the highlevel API only uses `type`. The `type` plugin meanwhile will prefer `check/type` but falls back to `type`. This means in most cases just `type` is enough, but `check/type` can be used for overrides (e.g. for enums without codegen)

doc/news/_preparation_next_release.md
```diff
@@ -183,6 +183,7 @@ you up to date with the multi-language support provided by Elektra.
   location data, providing additional features, such as clickable links to the error source. _(René Schwaiger)_
 
 - Added mores specification examples. _(Michael Zronek)_
+
```
markus2330: Please describe which examples you added (and link to them). "Added mores specification examples." is not very helpful.

examples/spec/lcdproc/LCDd.ini
```diff
@@ -2308,17 +2298,16 @@ description = "Specifies the displaytype.
...
 
 [serialvfd/custom-characters]
 type = unsigned_long
 check/type = unsigned_long
 check/validation = ^([1-9]\d*|0)$
 check/validation/match = LINE
 check/validation/message = Must be any positive number including 0
-default = 0
-description = "Number of Custom-Characters. default is display type dependent"
+default = ;default is display type dependent
```
markus2330: can you calculate the default?

kodebach: Each type has its own value that is more or less arbitrary, so you would need some sort of switch statement or chained if-else statements.

But the current code uses the value -83 as an indicator that the default should be used, so we might just set that as a default for now. Also if the value is above 99 a warning is displayed and the default is used, so the allowed value range is only 0-99 as well as -83.

https://github.com/lcdproc/lcdproc/blob/master/server/drivers/serialVFD_displays.h

markus2330: Yes, default value of -83 makes sense, then.

But ideally we would use the conditional plugin to actually calculate the number.

kodebach: > But ideally we would use the conditional plugin to actually calculate the number.

Not sure whether the syntax is correct, but I think the mapping from types to values is correct:
```
assign/condition/#0 = ( ./ == '0' ) ? '1' : '0'
assign/condition/#1 = ( ./ == '1' ) ? '31' : '0'
assign/condition/#2 = ( ./ == '2' ) ? '16' : '0'
assign/condition/#3 = ( ./ == '3' ) ? '3' : '0'
assign/condition/#4 = ( ./ == '4' ) ? '10' : '0'
assign/condition/#5 = ( ./ == '5' ) ? '3' : '0'
assign/condition/#6 = ( ./ == '6' ) ? '0' : '0'
assign/condition/#7 = ( ./ == '7' ) ? '16' : '0'
assign/condition/#8 = ( ./ == '8' ) ? '0' : '0'
```

examples/spec/lcdproc/LCDd.ini
```diff
@@ -2308,17 +2298,16 @@ description = "Specifies the displaytype.
...
-description = "'no' if display connected serial, 'yes' if connected parallel.
-I.e. serial by default"
+description = "'no' if display connected serial, 'yes' if connected parallel."
 
 [serialvfd/custom-characters]
 type = unsigned_long
 check/type = unsigned_long
 check/validation = ^([1-9]\d*|0)$
 check/validation/match = LINE
 check/validation/message = Must be any positive number including 0
```
markus2330: Why not use range instead of this validation?

@Piankero In the error concept, it would be nice to always have something like the "check/validation/message", so that spec writers can improve the error messages.

kodebach: > In the error concept, it would be nice to always have something like the "check/validation/message", so that spec writers can improve the error messages.

This is especially true, if the global plugin is used for command line options. We have to position it before all of the validation plugins, so the error message will be about some `proc` key being invalid. And the user would have to know which key corresponds to which command-line option.

Piankero: > it would be nice to always have something like the "check/validation/message", so that spec writers can improve the error messages.

What exactly do you mean? A metadata which is added to the parentkey which solves as the optional "solution" which I suggested in the concept?

kodebach: As I understood it, we introduce a metakey e.g. `info/error`. We then provide a way of specifying which key caused an error, and the name of this key and its `info/error` would be stored together with the error.

`info/error` would be given by application developers writing a specification, a "solution" can only be given by developers of Elektra and its plugins IMO. If the `check/path/mode` validation fails, a solution is e.g. "Ensure you have write access to file XYZ". This can't be known when writing the specification, but in most cases "solution" should be obvious from the error description.

`info/error` wouldn't really be useful for `path`. It is more useful in cases where the specification is complicated and needs an explanation. For example if regexes are involved, or if the `conditional` plugin is used, e.g. "If /some/key is set to XYZ, this must not be set to ABC".

`info/error` should be an explanation of what the specification is meant to enforce, while `solution`  describes the simplest way to get rid of the error.

---

Note: If we can specify in a standardized way which key caused an error, it would be nicer to store the command-line options of this key with error. Then it wouldn't be necessary to add `info/error` to every key with a command-line option.

examples/spec/lcdproc/LCDd.ini
Piankero: @kodebach: I think we need to change it to enum rather than boolean because the boolean plugin would convert it to 0 and 1 and LCDproc checks for "yes/no", "on/off". Is this correct?

kodebach: No, definitely don't convert it to enum. If it is a boolean than declare it as a boolean. The old LCDproc does accept yes/no, on/off and a few other things for booleans but that should handled by Elektra and LCDproc should only expect 0 and 1 for booleans.

examples/spec/cassandra.ini
Piankero: Actually this works better than the ini plugin itself. There is currently a crucial problem for the ini plugin #2467

kodebach: I think the problem is more that the file extension `.ini` suggest an actual INI file.

Piankero: It actually takes the storage format from infos/plugins and not from the extension

examples/spec/lcdproc/LCDd.ini
Piankero: Thank you @kodebach for that input, I will consider it in the concept.

> Why not use range instead of this validation?

The range plugin has no possibility to just set a minimum value but no maximum. 
Theoretically I can delete the validation anyway and just let the type plugin check for it. What do you say about that?

examples/spec/cassandra.ini
kodebach: Yes,that is exactly the point you are creating a `dump` file with `.ini` extension. That will confuse users when they try to look at the file in an external editor.

Change the plugin to one of the ini plugins (`ni` works quite well) or change the extension to `.ecf` or something else that doesn't suggest a common file format.

examples/spec/lcdproc/LCDd.ini
Piankero: Unfortunately this did not work. The README.md does not show any array example at all. In the code (conditional.c:953) I am not even sure if this is intended to work like this

kodebach: https://github.com/ElektraInitiative/libelektra/blob/b635bc3f69621d7f23e4ff3e3785fbae2b57504f/src/plugins/conditionals/README.md#L54-L55

I just based my solution on these lines in the README, but the plugin implementation is very confusing and not at all documented, so I don't know whether or not this should work somehow.

markus2330:
Thank you for this large examples!

@e1528532 Can you type check these examples?

examples/spec/cassandra.ini
markus2330: For the LCDproc files we should use "ini" so that old config files will still work (or switch to YAML if @sanssecours is ready, but then we need some migration path).

examples/spec/lcdproc/LCDd.ini
markus2330: Did you try it? @tom-wa sometimes has undocumented features, maybe "0-" works. If it works, please document it.

markus2330: > Theoretically I can delete the validation anyway and just let the type plugin check for it. What do you say about that?

The type plugin might also not provide the best error message. Hand-written error messages in the specification will often outperform generic error messages. I hope I will read an analysis of this in your thesis.

examples/spec/lcdproc/LCDd.ini
markus2330: @Piankero so you do not need "check/type" if it is the same as "type". Can you also add this as documentation to the "type" plugin?

examples/spec/cassandra.ini
Piankero: If we use the ini format then most things will not work because of multiline descriptions.

markus2330: Yes, it is a bug to be fixed. When using dump, people will not be able to edit files by hand, which is a much bigger usability problem. (It breaks their current workflow)

examples/spec/lcdproc/LCDd.ini
Piankero: https://github.com/ElektraInitiative/libelektra/pull/2475

Issue #2439, opened 2019-02-25
Build Jobs: Parts of Test Suite Fail Regularly


sanssecours:
# Description

I opened this as an issue to keep track all of the **temporary** test failures in one of the build jobs. The main reasons for the build failures are 

- the [tests for the notification bindings/plugins](https://github.com/ElektraInitiative/libelektra/issues/2310), and
- the [tests for the crypto plugin](https://github.com/ElektraInitiative/libelektra/issues/2341) (this only applies to the Jenkins build server)

. In a [recent PR](https://github.com/ElektraInitiative/libelektra/pull/2436) I had to restart the Jenkins build job 5 times before everything worked. In the [PR after that](https://github.com/ElektraInitiative/libelektra/pull/2438) I restarted the Jenkins build job thrice, as far as I can remember. Anyway, the failure rate is much too high in my opinion.

# Failures

| Location | Failed Tests | Build Job |
|----------|-------------|-----------|
|     [`master`](https://github.com/ElektraInitiative/libelektra/tree/master)     |       `testmod_gpgme` (1)      |   `debian-stable-full`        |
|     [`master`](https://github.com/ElektraInitiative/libelektra/tree/master)     |       `testmod_gpgme` (1), `testmod_zeromqsend` (1)      |   `debian-stable-full-ini`        |
|     [`master`](https://github.com/ElektraInitiative/libelektra/tree/master)     |       `testmod_crypto_botan` (1), `testmod_fcrypt` (1), `testmod_gpgme` (2), `testmod_zeromqsend` (1)    |   `debian-stable-full-mmap`        |
|     [`master`](https://github.com/ElektraInitiative/libelektra/tree/master)     |       `testmod_crypto_botan` (1), `testmod_fcrypt` (2)      |   `debian-unstable-full`        |
|     [`master`](https://github.com/ElektraInitiative/libelektra/tree/master)     |       `testmod_crypto_botan` (2), `testmod_crypto_openssl` (3), `testmod_fcrypt` (1)       |   `debian-unstable-full-clang`        |
|     [`PR #2442`](https://github.com/ElektraInitiative/libelektra/pull/2442)     |       `testmod_crypto_openssl` (1), `testmod_gpgme` (1)      |   `debian-stable-full-ini`        |
[...]

PR #2447, opened 2019-03-01
Jenkins: Disable Failing Tests

sanssecours:
This update should get rid of most of the temporary test failures reported in issue [#2439](https://issues.libelektra.org/2439).

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -154,7 +154,18 @@ you up to date with the multi-language support provided by Elektra.
 
 ### Jenkins
 
-- <<TODO>>
+- We disabled the tests:
+
+  - `testmod_crypto_botan`,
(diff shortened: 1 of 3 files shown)


markus2330:
Thank you! Shouldn't we disable it within our CMake files so that they do not fail for anyone?

markus2330:
And please add the disabled tests in doc/todo/TESTING

sanssecours:
> Shouldn't we disable it within our CMake files so that they do not fail for anyone?

I leave that decision up to you. On my machine the tests work without any problems. They will fail sometimes on Cirrus and Travis, but I feel it is still acceptable to restart failed build jobs, if they do.

Anyway, if you want to disable the tests everywhere, please just open a pull request containing the relevant changes.

markus2330:
Thank you!

Issue #2270, opened 2018-10-15
Global Cache


mpranj:
Meta issue for the global cache.

- [ ] two stage pointer update: 1) only global keyset (timestamps), 2) "real" keyset
- [x] resolver: check if update is needed #2249
- [ ] support other resolvers #2598
- [ ] cache: transient plugins #2269
- [ ] document limitations w.r.t. #2269
- [ ] crypto plugins + caching #2613
- [x] global KeySet handle
- [x] use new plugin name / variant API #2360
- [ ] problems with non-file-based storage plugins #2454
- [x] put cache into release note highlights and add benchmarks
- [ ] rename debugGlobalPositions to something better (cacheHitBypass?) as discussed in #2643
- [x] Add tools for kdb cache {enable,disable,clear}. #2831

markus2330:
Thank you for tracking this!

mpranj:
@markus2330 as I mentioned earlier, the very rough numbers for the cache are: 10x faster than no cache (with dump and no weird edge cases).

I investigated, why the improvement is not at its highest potential (we saw mmapstorage read is >70x faster than dump). It seems that other global plugins take up **most** of the time. (`spec`, `list` and stuff in PREGETSTORAGE, PROCGETSTORAGE, POSTGETSTORAGE).

When disabling these positions which are always called, even with cache, we would get an improvement of **>80x**. Maybe it would be interesting to think about going this direction.

markus2330:
Yes, we absolutely should do that. In my mind the cache plugin would always immediately exit for cache hits and not call any further plugins, maybe except some unlock, logging or notification plugins. But spec definitely is not needed when we have a cache hit.

markus2330:
A big thanks that you investigated this!

mpranj:
Nice, I think that will be a huge performance improvement then.

The drawback is that we'll need additional position(s) for that case. Currently PREGETSTORAGE, PROCGETSTORAGE and POSTGETSTORAGE are always called (whether cache hit, or no update as it was before we had a cache). I'm just not sure what this means for the semantics.

markus2330:
Yes, the idea is to bypass everything that builds up the KeySet on a cachehit. As logging/notification for kdbGet is not really important and locking currently is not implemented (semlock is disabled) you can simply start by ignoring the semantics of the global plugins on cache hits (not executing anything for now).

mpranj:
Two things came up regarding this: 

[...]

PR #2457, opened 2019-03-04
mmapcache

mpranj:
Initial implementation of #2270. Some improvements are left for a second PR.
## Basics

## Checklist

need to be checked.

- [x] I added unit tests


src/include/kdbplugin.h
```diff
@@ -86,6 +86,9 @@ typedef enum
 /** Everything went fine and the function **did not** update the given keyset/configuration */
 #define ELEKTRA_PLUGIN_STATUS_NO_UPDATE 0
 
+/** Everything went fine and we have a cache hit */
+#define ELEKTRA_PLUGIN_STATUS_CACHE_HIT 2
```
markus2330: Is it necessary to distinguish between NO_UPDATE and CACHE_HIT?

src/libs/elektra/kdb.c
```diff
@@ -809,10 +924,90 @@ int kdbGet (KDB * handle, KeySet * ks, Key * parentKey)
...
+			goto cachefail;
+		}
+	}
+
+cachefail:
 	// Check if a update is needed at all
 	switch (elektraGetCheckUpdateNeeded (split, parentKey))
 	{
+	case -2: // We have a cache hit
+		ELEKTRA_LOG_DEBUG ("CACHE HIT");
```
markus2330: put this to a function?

src/plugins/cache/README.md
```diff
@@ -0,0 +1,54 @@
...
+- infos/description = caches keysets from previous `kdbGet()` calls
+
+## Introduction
+
+This caching plugin stores keysets from previous `kdbGet()` calls
+to improve performance when reading configuration files.
+
+## Usage
+
+You can use `scripts/copy-cache`
```
markus2330: Please update.

mpranj: Yep. Sorry, but it was not yet ready for review. Some problems which I did not encounter came up on the build servers.

src/include/kdbplugin.h
mpranj: Yes. If all backends return:

- NO_UPDATE: nothing at all to do.
- CACHE_HIT: need to fetch from cache (invoke cache, update all pointers)
- SUCCESS: need to fetch from storage

NO_UPDATE is faster than fetching the cache. I need to implement delaying the pointer update though.

src/plugins/cache/README.md
markus2330: I only looked over it now so that we hopefully can immediately merge it once the build server says okay. The obvious TODOs, like the code to be removed, I did not mark, though.

markus2330:
This pull request **fixes 1 alert** when merging c5d8a991241cc5403978552ea15cdeb01ed5deef into 0f9f89493d150093a5d57319aadfe47e62619557 - [view on LGTM.com](https://lgtm.com/projects/g/ElektraInitiative/libelektra/rev/pr-ae93a323e20f4471aaaeafec3f840c8ee4e936c3)

**fixed alerts:**

* 1 for Missing header guard

---

markus2330:
Any update here? It would be great to have this!

```
kdb ls system  7,74s user 0,59s system 40% cpu 20,734 total
kdb ls system > /dev/null 2> /dev/null  7,77s user 0,16s system 98% cpu 8,075 total
```

mpranj:
I'm still working on it. Some problems appeared on the build server. I have not been able to reproduce the problems locally, which makes it tedious to debug.

[One test](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2457/24/pipeline/457) fails because it apparently reorders some ini ordering. I don't know why this happens yet.

```diff
--- a/user.check.dump.txt
+++ b/user.export.dump.txt
@@ -5,7 +5,7 @@ user/sw/tests/jna/1
 keyMeta 22 3
 internal/ini/key/last #0
 keyMeta 19 3
-internal/ini/order #2
+internal/ini/order #3
 keyMeta 20 5
 internal/ini/parent user
 keyMeta 21 1
@@ -16,7 +16,7 @@ user/tests
 keyMeta 22 3
 internal/ini/key/last #0
 keyMeta 19 3
-internal/ini/order #3
+internal/ini/order #2
 keyMeta 20 5
 internal/ini/parent user
 keyMeta 21 1
´´´

**Comment by markus2330** [comment]
> I have not been able to reproduce the problems locally, which makes it tedious to debug.

Even not with docker?

> One test fails because it apparently reorders some ini ordering.

We were thinking to throw this INI tests as we anyway want YAML and not INI as default (see #2499 for a previous situation where this test caused problems). But in your case it might be that it actually unveiled a bug?

If this is the only problem, please simply exclude your plugin in the INI tests, create an issue that we do not forget about it, and let us merge this.

**Comment by mpranj** [comment]
> Even not with docker?

I was able to reproduce it today with docker, but the root cause is still unclear to me.

> If this is the only problem, [...]

There are [other test failures](https://travis-ci.org/ElektraInitiative/libelektra/jobs/510366439) and I think this has unveiled a real bug. It's weird that the same tests run fine on multiple other instances.

**Comment by mpranj** [comment]
Ok, so part of the problem is that older macOS (using HFS+) is missing nanosecond timestamp information. We basically have this problem in our resolver already, which could be shown with a test case. The cache simply triggers the bug more often, since this timestamp information is persisted.

I think the best would be to disable the cache for HFS+ filesystems (where tv_nsec is always 0).

**Comment by markus2330** [comment]
> I think the best would be to disable the cache for HFS+ filesystems (where tv_nsec is always 0).

Yes, I agree. Once we fixed  #2533, we can activate the cache again.

File: doc/news/_preparation_next_release.md [hunk 4]
```diff
@@ -155,6 +155,10 @@ The following section lists news about the [modules](https://www.libelektra.org/
 
 - We fixed an incorrect format specifier in a call to the `syslog` function. _(René Schwaiger)_
 
+### Cache
+
+- [cache](https://www.libelektra.org/plugins/cache) is a new global caching plugin. It uses [mmapstorage](https://www.libelektra.org/plugins/mmapstorage) as its storage backend and lazily stores keysets from previous ´kdbGet()´ calls. _(Mihael Pranjić)_
```
markus2330: Please highlight it and also put benchmarks here (does not need to be within this PR, though)

mpranj:
@markus2330 I'm afraid the reported problem with ini is not yet fixed. The tests succeeded because of an error where the cache was effectively disabled (always a cache miss).

With the cache disabled, the key `sw/tests/jna/1` does not exist (is deleted). With the cache enabled, the key remains in the config. I'd like to fix this before we merge.

mpranj:
@markus2330 I implemented a check where we fetch from the backends **and** from the cache to see what is going on with this last problem. Doing some tests by hand it seems it only fails with ini and not with dump.

After a bit more investigation I noticed that ini returns different `internal/ini/order` metakeys for exactly the same key (e.g. `user/test/foo`) depending on the parentKey from the invocation. I did not expect this behaviour and it is definitely incompatible with the current design of the cache.

@tom-wa can you confirm my understanding og the `internal/ini/order` key?

If I am not missing something here, we should simply make the cache incompatible with ini. Either not compile it at all, or simply put in the README, or some other solution. I think a dynamic check whether ini is in a backend is maybe too much overhead.

markus2330:
>  I implemented a check where we fetch from the backends and from the cache to see what is going on with this last problem.

This sounds like a very handy debugging feature. Maybe you can leave this code or at least leave the commit in the history so that we can run this check again if we once again run into such a problem.

>If I am not missing something here, we should simply make the cache incompatible with ini.

Yes, as already said I am fine with this solution! Simply document that INI as default and mmap cache do not play well together. (Maybe with a bit of description what is the worst thing that could happen.) And obviously disable your plugin in the INI-as-default test case.

It would be great to have this merged asap.

mpranj:
> This sounds like a very handy debugging feature.

The way I implemented it is not really suitable because it clutters kdbGet a lot but I'll think of a nice way to add it. If I don't find a nice way I'll just leave the dirty way in the history.

> Simply document that INI as default and mmap cache do not play well together.

I think ini makes problems even if it is just mounted **anywhere**. It keeps track of global ordering which is different when kdbGet is invoked with different parentKeys. I think the problem/effect occurs when ini is present in multiple backends.

Unfortunately I really thought it was a bug in the cache at first, which wasted a lot of time. From my current understanding it's really just an incompatibility issue.

> It would be great to have this merged asap.

I should be able to finish it up tomorrow, I can't work on it today.

markus2330:
>   The way I implemented it is not really suitable because it clutters kdbGet a lot but I'll think of a nice way to add it. If I don't find a nice way I'll just leave the dirty way in the history.

leaving in the history is enough, do not spent extra time here.

markus2330:
> I think ini makes problems even if it is just mounted anywhere. It keeps track of global ordering which is different when kdbGet is invoked with different parentKeys. I think the problem/effect occurs when ini is present in multiple backends.

Can you create a new issue where you write what you found out?

And please document that it should not be used together (as long as the bug is not fixed)

mpranj:
@markus2330 this is ready. 🎉

The cache debug code was left in c31979d142eb51052b6344d97a66c190f1a4db5f and then reverted.  As discussed, some improvements are left for a second PR. The cache is disabled dynamically whenever we encounter the ini plugin, otherwise it just works.

A best-scenario (cache hit) benchmark is available in `./bin/benchmark_kdb` and shows roughly a 10x improvement over dump without cache. There might be room for improvement. (Concrete numbers and highlight in the news coming in a second PR)

markus2330:
Thank you! I am looking forward using it tomorrow (I'll run tomorrow a script which issues *many* `kdb` commands.)

PR #2458, opened 2019-03-05
Minor addition to specload

kodebach:
This is a minor addition to the specload plugin, to abstract over the dependency on `quickdump`.

### DIFF
--- a/src/plugins/specload/README.md
+++ b/src/plugins/specload/README.md
@@ -54,6 +54,26 @@ kdb mount specload.eqd spec/tests/specload/example specload 'app=/usr/bin/exampl
 
 To inhibit the default `--elektra-spec` argument and call an application without any arguments use `'app/args='`.
 
+The app must only output the expected base specification to `stdout` and then exit, when called by `specload`. We use the `quickdump` plugin
+for communication, so the application should call `elektraQuickdumpSet`. Because this dependency may change in future, it is recommended
+you use the function `elektraSpecloadSendSpec` exported as `"system/elektra/modules/specload/exports/sendspec"`:
+
+```c
+Key * errorKey = keyNew (0, KEY_END);
+
+// add 'system/module' key to suppress checking the 'app' key in elektraSpecloadOpen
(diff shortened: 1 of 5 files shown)


src/plugins/specload/README.md
```diff
@@ -54,6 +54,26 @@ kdb mount specload.eqd spec/tests/specload/example specload 'app=/usr/bin/exampl
...
 
+The app must only output the expected base specification to `stdout` and then exit, when called by `specload`. We use the `quickdump` plugin
+for communication, so the application should call `elektraQuickdumpSet`. Because this dependency may change in future, it is recommended
+you use the function `elektraSpecloadSendSpec` exported as `"system/elektra/modules/specload/exports/sendspec"`:
+
+```c
+Key * errorKey = keyNew (0, KEY_END);
+
+// add 'system/module' key to suppress checking the 'app' key in elektraSpecloadOpen
+KeySet * specloadConf = ksNew (1, keyNew ("system/module", KEY_END), KS_END);
```
markus2330: This looks like a hack. What about using /sendspec to put the module into the mode as needed for invoking sendspec?

markus2330:
Thank you! I merged it. One minor detail which can be fixed later when the high-level API adds support for invoking sendspec.

src/plugins/specload/README.md
kodebach: It is a bit of a hack, but `kdb mount` opens the plugin at least once with a config only containing `system/module`, so we have to accept this anyways.

It's also not really a problem, `kdbGet` must expect that the `/app` key is invalid, in case the executable was removed between `kdbOpen` and `kdbGet` for example. The only reason to have the check in `kdbOpen` at all, is so we can catch problem as early as possible.

markus2330: The purpose of /module is to open plugins in a way so that they "always work" and can return their contract (to be shown in system/elektra/modules). To use the same config key for other purposes, seems like a hack for me.

PR #2459, opened 2019-03-05
move reference resolving into libease

kodebach:
This simply moves reference resolution from the reference plugin into libease. That way it can be used by other parts of Elektra, in particular code generation.

### DIFF
--- a/src/libs/ease/reference.c
+++ b/src/libs/ease/reference.c
@@ -1,0 +1,85 @@
+/**
+ * @file
+ *
+ * @brief Reference methods.
+ *
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ */
+
+#include "kdbease.h"
+
+#include "kdbhelper.h"
+
+#include <string.h>
+
+
+/**
+ * Check whether a reference is redundant (i.e. it can be expressed in less characters) or not.
+ *
+ * @param reference the reference to check
+ * @retval 1 if the reference is redundant
+ * @retval 0 otherwise
+ */
+int elektraIsReferenceRedundant (const char * reference)
+{
+	const char * cur = reference;
+	while (strncmp (cur, "../", 3) == 0)
+	{
+		cur += 3;
+	}
+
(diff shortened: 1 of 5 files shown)


src/libs/ease/reference.c
```diff
@@ -0,0 +1,85 @@
...
+
+
+/**
+ * Check whether a reference is redundant (i.e. it can be expressed in less characters) or not.
+ *
+ * @param reference the reference to check
+ * @retval 1 if the reference is redundant
+ * @retval 0 otherwise
+ */
+int elektraIsReferenceRedundant (const char * reference)
```
markus2330: Do we really want to export this API call?

Can you give an example how it can be used by an user?

And some tests would be nice.

markus2330:
Thank you! Again some minor details which can be resolved later.

src/libs/ease/reference.c
kodebach: It isn't really useful for applications, but it can be used by plugins and other parts of elektra.
The `reference` plugin for example produces a warning, if a reference is redundant, because this is probably a mistake. The same could be done by any other future plugin that uses `references` (default value calculation, better conditionals, etc.).

> And some tests would be nice.

I thought about adding test, but we would basically just test `keySetName/keyAddName`. The 3 `if`s inside `elektraResolveReference` directly correspond to the documentation above and `elektraIsReferenceRedundant` is also very very simple.

Also, `testmod_reference` from the `reference` plugin will fail, if this method is broken. Because the methods originally where contained in the plugin, all currently possible cases should be covered already.

coveralls also reports [100% coverage](https://coveralls.io/builds/22003479/source?filename=src/libs/ease/reference.c)

markus2330: > It isn't really useful for applications, but it can be used by plugins and other parts of elektra...

I did not doubt that this is useful to have in a lib. Only the docu did not explain what you are writing here.

> coveralls also reports 100% coverage

Ok, then it is called indirectly.

Issue #2451, opened 2019-03-01
How to activate the boolean plugin in a specification file


Piankero:
## Steps to Reproduce the Problem

The boolean plugin does not correctly convert to a canonicalizes value
Mount the specification with the following content:

```ini
[]
mountpoint = test-spec.conf
infos/plugin = ini boolean

[/setting/enabled]
check/type = boolean
```

```sh
kdb mount <path>/test.ini spec/test ni                     
kdb spec-mount /test                                                    
kdb set /test/setting/enabled true                                      
#> Using name user/test/setting/enabled
#> The command kdb set failed while accessing the key database with the info:
#> Sorry, the error (#52) occurred ;(
#> Description: could not type check value of key
#> Reason: The type boolean failed to match for user/test/setting/enabled with string: true
#> Ingroup: plugin
#> Module: type
#> At: ..../src/plugins/type/type.cpp:31
#> Mountpoint: user/test
#> Configfile: .../user/test-spec.conf.11243:1551466614.338725.tmp
```
## Expected Result

true gets accepted for the type plugin

## Actual Result

true gets rejected

## System Information

- Elektra Version: master (0.8.26)
```
cat /etc/os-release
#> NAME="Linux Mint"
#> VERSION="18.2 (Sonya)"
#> ID=linuxmint
#> ID_LIKE=ubuntu
#> PRETTY_NAME="Linux Mint 18.2"
#> VERSION_ID="18.2"
#> HOME_URL="http://www.linuxmint.com/"
#> SUPPORT_URL="http://forums.linuxmint.com/"
#> BUG_REPORT_URL="http://bugs.launchpad.net/linuxmint/"
#> VERSION_CODENAME=sonya
#> UBUNTU_CODENAME=xenial
```

Notes:
How exactly can I see what plugins are currently active? (There is a certain key which holds those 5 plugin I remember)

markus2330:
Thank you for reporting this problem!

> How exactly can I see what plugins are currently active? (There is a certain key which holds those 5 plugin I remember)

Please ask questions in separate issues, then they are easier to find in future. `kdb mountpoint-info` tells you which plugins exist for a mountpoint. With `kdb find-tools -t mount` you would have been able to find this tool.

Please also always add the issues you create to LCDproc and assign yourself if you will work on it.

Now to the issue itself:

[...]

PR #2473, opened 2019-03-10
boolean: added ordering

Piankero:
A PR showing my changes to the boolean plugin so it gets executed before the the type plugin

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -104,6 +104,11 @@ The following section lists news about the [modules](https://www.libelektra.org/
   problem with the resolver this cannot be used right now (see [limitations](https://www.libelektra.org/plugins/specload)).
 - We also export `elektraSpecloadSendSpec` to abstract over the `quickdump` dependency. _(Klemens Böswirth)_
 
+### Boolean
+
+- Added `infos/ordering = type` so that the plugin gets loaded before the `type` plugin. This is needed because values have to be canonicalized
+  before the type plugin does its checks.
+
 ## Libraries
 
 The text below summarizes updates to the [C (and C++)-based libraries](https://www.libelektra.org/libraries/readme) of Elektra.
--- a/src/plugins/boolean/README.md
+++ b/src/plugins/boolean/README.md
@@ -36,6 +36,18 @@ kdb rm -r user/tests/boolean
 sudo kdb umount user/tests/boolean
 ```
 
+```sh
+sudo kdb mount config.ecf user/tests/boolean dump boolean type
+kdb set user/tests/boolean/truthiness 0
+kdb setmeta user/tests/boolean/truthiness type boolean
+
+kdb set user/tests/boolean/truthiness yes
+
+# Undo changes
+kdb rm -r user/tests/boolean
+sudo kdb umount user/tests/boolean
+```
+
 ## Configuration
 
 - `on/true` specifies the normalized value for `TRUE`. Default: `1`
--- a/src/plugins/boolean/README.md
+++ b/src/plugins/boolean/README.md
@@ -42,6 +42,7 @@ kdb set user/tests/boolean/truthiness 0
 kdb setmeta user/tests/boolean/truthiness type boolean
 
 kdb set user/tests/boolean/truthiness yes
+# RET: 0
 
 # Undo changes
 kdb rm -r user/tests/boolean
--- a/src/plugins/boolean/README.md
+++ b/src/plugins/boolean/README.md
@@ -6,6 +6,7 @@
 - infos/recommends =
 - infos/placements = postgetstorage presetstorage
 - infos/status = recommended productive maintained reviewed conformant compatible coverage specific unittest tested nodep configurable
+- infos/ordering = type
 - infos/metadata =
 - infos/description = canonicalizes boolean values
 
--- a/src/plugins/boolean/boolean.c
+++ b/src/plugins/boolean/boolean.c
@@ -72,25 +72,9 @@ int elektraBooleanClose (Plugin * handle ELEKTRA_UNUSED, Key * errorKey ELEKTRA_
 	return 1; // success
 }
 
-static int isTrue (const char * value, const char ** trueValues)
+static int arrayContains (const char ** array, const char * value)
 {
-	char ** ptr = (char **) trueValues;
-	int retVal = 0;
-	while (*ptr)
-	{
-		if (!strcasecmp (value, *ptr))
-		{
-			retVal = 1;
-			break;
-		}
-		++ptr;
-	}
-	return retVal;
-}
-
-static int isFalse (const char * value, const char ** falseValues)
-{
-	char ** ptr = (char **) falseValues;
+	const char ** ptr = array;
 	int retVal = 0;
 	while (*ptr)
 	{
@@ -255,14 +239,14 @@ int elektraBooleanGet (Plugin * handle, KeySet * returned, Key * parentKey)
 		{
 			const char * value = keyString (key);
 
-			if (isTrue (value, trueValues))
+			if (arrayContains (trueValues, value))
 			{
 				keySetMeta (key, "internal/boolean/origvalue", keyString (key));
 				keySetMeta (key, "internal/boolean/normvalue", data->trueValue);
 				ELEKTRA_LOG_DEBUG ("Convert “%s” to “%s”", value, data->trueValue);
 				keySetString (key, data->trueValue);
 			}
-			else if (isFalse (value, falseValues))
+			else if (arrayContains (falseValues, value))
 			{
 				keySetMeta (key, "internal/boolean/origvalue", keyString (key));
 				keySetMeta (key, "internal/boolean/normvalue", data->falseValue);
@@ -335,16 +319,16 @@ int elektraBooleanSet (Plugin * handle, KeySet * returned, Key * parentKey)
 				continue;
 			}
 
-			if (isTrue (value, trueValues))
+			if (arrayContains (trueValues, value))
 			{
-				if (originalValue != NULL && isTrue (originalValue, trueValues))
+				if (originalValue != NULL && arrayContains (trueValues, originalValue))
 				{
 					keySetString (key, originalValue);
 				}
 			}
-			else if (isFalse (value, falseValues))
+			else if (arrayContains (falseValues, value))
 			{
-				if (originalValue != NULL && isFalse (originalValue, trueValues))
+				if (originalValue != NULL && arrayContains (falseValues, originalValue))
 				{
 					keySetString (key, originalValue);
 				}
--- a/src/plugins/boolean/boolean.c
+++ b/src/plugins/boolean/boolean.c
@@ -315,7 +315,7 @@ int elektraBooleanSet (Plugin * handle, KeySet * returned, Key * parentKey)
 			const Key * origValue = keyGetMeta (key, "internal/boolean/origvalue");
 			const char * originalValue = origValue == NULL ? NULL : keyString (origValue);
 			const Key * norm = keyGetMeta (key, "internal/boolean/normvalue");
-			const char * normValue = origValue == NULL ? NULL : keyString (norm);
+			const char * normValue = norm == NULL ? NULL : keyString (norm);
 
 			if (norm == NULL && strcmp (value, data->invalidValue) == 0)
 			{
@@ -326,7 +326,7 @@ int elektraBooleanSet (Plugin * handle, KeySet * returned, Key * parentKey)
 			}
 
 
-			if (strcmp (value, normValue) == 0)
+			if (normValue != NULL && strcmp (value, normValue) == 0)
 			{
 				// unchanged valid (no warning) normalized value
 				keySetString (key, originalValue);
--- a/src/plugins/boolean/boolean.c
+++ b/src/plugins/boolean/boolean.c
@@ -12,37 +12,41 @@
 #include <kdberrors.h>
 #include <kdbhelper.h>
 #include <kdblogger.h>
+#include <stdbool.h>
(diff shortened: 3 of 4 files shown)


markus2330:
Ok, looks good. Can you add a shell recorder test case which shows type+boolean used together?

Piankero:
I added this PR because of a request by you in #2451 
> Please create PRs if you change something. Otherwise it is unclear where exactly you added what.

markus2330:
> I added this PR because of a request by you in #2451

Yes, and everything is fine with the PR. So now we need some way to trigger the bug, ideally within this PR so that I do not need to reproduce it locally.

markus2330:
Thank you for this demonstration! The boolean plugin seems indeed to have a severe flaw.

Piankero:
@kodebach As long as this bug is not fixed, the specification for LCDproc will not work correctly. Should we change to the enum plugin temporary?

markus2330:
Please do not make such workarounds. It is much easier to fix the boolean plugin than to rewrite all specifications twice.

kodebach:
I think we should just abandon the `boolean` plugin, and integrate it into the `type` plugin. It doesn't make a whole lot of sense to use either of them separately.

The `boolean` plugin checks a subset of the keys checked by `type`. By default it sets any key with boolean type that is not a valid boolean value to the true value, when in fact it should really raise an error. It can be configured to use values other than `1` and `0` for true and false, which goes against [`/doc/decisions/bool.md`](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/bool.md). We should normalize to 0 and 1 on kdbGet and normalize to user defined values on kdbSet.

To emulate the current `type` plugin, the user would simply set the allowed true values to just `1` and the allowed false values to only `0`. Emulating the current `boolean` plugin wouldn't be possible, but I don't think this matters, because it is obviously very broken.

---

PS. the actual bug seems to be that the boolean plugin assumes that if the value of a Key is changed the whole Key is replaced (and therefore the internal metadata removed). Which of course is not at all a valid assumption, because `keySetString` is a thing.

kodebach:
Actually this might be a good moment to rewrite the `type` plugin in C, because there is really no reason why it has to use C++. IMO the use of C++ and templates just complicates the plugin unnecessarily.

markus2330:
> I think we should just abandon the boolean plugin, and integrate it into the type plugin. It doesn't make a whole lot of sense to use either of them separately.

I agree, the modularity is a bit overdone in this case.

> We should normalize to 0 and 1 on kdbGet

I fully agree. Most of the configuration the boolean plugin has, is not needed or violates our decisions. We should remove them.

> and normalize to user defined values on kdbSet.

I would prefer if it remembers the value as it was.

> Actually this might be a good moment to rewrite the type plugin in C

Of course it would be nice to streamline this plugin and have nice and clean C code. However, there are quite a lot open tasks to elektrify LCDproc and a rewrite of the type plugin would not bring us closer to the goal.

> IMO the use of C++ and templates just complicates the plugin unnecessarily.

Yes. The implementation tried to be extensible so that users can add new types. But it is now quite clear that this is useless and instead people want a fixed set of data types which are available everywhere and consistent across Elektra's whole stack.

kodebach:
I fixed the bug in the boolean plugin, but there is a larger problem here. The interaction between `type` and `boolean` is more or less impossible:
1. on kdbGet the order has to be `boolean` -> `type` (to normalize values), but on kdbSet it has to be the reverse (to restore original values), which is not how `infos/ordering` works AFAIK
2. even if the order was correct, `boolean` would need to normalize values on kdbSet. But that means `kdb set /x/y yes` would result in `x/y = 1` being stored. But editing the file so that `x/y = yes` would keep the value `yes` set in that file. This seems a bit weird. 

There is also a weird edge case I can't get rid of:
- Assume `user/booleankey = yes`, `yes` is a valid true value and `1` is the normalized true value
- Now call kdbGet on a file like this.
- The boolean plugin will normalize `user/booleankey` to `1`.
- Call `keySetString()` to set `user/booleankey` to `1` and then call kdbSet.
- You would expect that the value `1` is stored, but actually the `boolean` plugin can't ever now that you called keySetString and restores the original value of `yes`.

To get rid of all this weirdness, I think we need integrate `boolean` into type. Also I think we need to normalize to user defined values on kdbSet and to `0`/`1` on kdbGet. Actually I am not sure that even covers all problems, but it should cover most of them.

markus2330:
> but on kdbSet it has to be the reverse (to restore original values)

This should be the default, is it not? (see infos/stacking)

> There is also a weird edge case I can't get rid of

Thank you for looking into so deeply. This is actually a bug. keySetString should remove original values in that case! (It already removes the information that a key is binary so it would not be a big change). Isn't it even the case if you make keySetString("0") that still "yes" would be serialized?

> To get rid of all this weirdness, I think we need integrate boolean into type.

I already agreed that this makes sense. But it will not solve the keySetString problem, will it?

kodebach:
> This should be the default, is it not? (see infos/stacking)

Ah, okay, I missed that. But, see below for other issues.

> Isn't it even the case if you make keySetString("0") that still "yes" would be serialized?

No, I worked around that by storing the normalized value in metadata as well. Only if the new normalized value is the same as the one in metadata the bug occurs.

> keySetString should remove original values in that case! (It already removes the information that a key is binary so it would not be a big change).

That means we would have a metakey "origvalue" that **all** normalization plugins **must** use to store the original value. That also means that a normalization plugin is not allowed to touch a key that already has the metadata "origvalue" because it would break the previous plugin that normalized the key.

Another (less complicated and more reliable) solution would be to always normalize to user defined values on `kdbSet` instead of restoring the original. IMO this makes sense, because then all booleans for one mountpoint would use the same syntax, instead of possible mixing `yes`/`no` `true`/`false` etc in one file. The problems below would however still persist.

> But it will not solve the keySetString problem, will it?

No, it wouldn't actually help much at all, which is why I didn't go down this route.

If we integrate the `boolean` plugin (normalization) into the `type` plugin (validation) the positioning of the `type` plugin becomes problematic. On the one hand normalization plugins should be executed first, so that other plugins don't have to care about whether values are normalized or not. On the other hand validation plugins should be executed last, so that it is ensured no plugin broke a constraint.

Lets say for example a key has `type = long` and the `conditionals` plugin sets it to `"cheese"` (because of some weird setup). Obviously one would expect an error from the `type` plugin, but that is not the case, if it is mounted first to ensure normalized boolean values for other plugins. If it is mounted last, plugins cannot rely on boolean values being either `1` or `0`.

On the `kdbSet` side this problem gets even more complicated. For example, if a user sets a key with `type = boolean` to `yes`, we need to normalize it to `1` then execute all the plugins and in the end turn it back into `yes` (but only if no plugin modified it) so that we preserve the users preferred way of writing booleans. This actually can't even be solved by keeping `boolean` and `type` separate. Because, if they are separate the `type` plugin requires normalized boolean values, so the `boolean` plugin must the executed first, but then we would always store `0` or `1` for booleans, so the whole `yes` `no` etc. thing becomes useless. This would also not be solved by normalizing to user defined values on `kdbSet`, unless the user chooses to go with `0` and `1`.

All this is so complicated that I am starting to think the plugin system needs a major overhaul. IMO plugins shouldn't be defined by their placement, but by what the do. `storage` plugins, `normalization` plugins, `validation` plugins, etc. The core would then decide based on that in what order to execute these plugins (possibly multiple times). I also think that the generic plugin interface of just to methods is bad (e.g. with 3 validation plugins you iterate the whole KDB 3 times, even though it would work in 1 iteration too).

The only other solution I can think of is integrating the type system deeply into the core of Elektra. Then normalization, validation etc. could be done at the appropriate moments. We could also store the actual boolean/integer/float values to avoid multiple conversions. However, this is obviously not a solution right now, because of the effort and compatibility issues involved.

markus2330:
> That means we would have a metakey "origvalue" that all normalization plugins must use to store the original value.

Yes, of course. Plugins are not allowed to use any metakeys but only these which are defined in doc/METADATA.ini. For restoring original values currently nothing is defined but using origvalue makes sense (as origname already exists). I would prefer original/value and original/name though. (Needs changes in the rename plugin, though.)

> If we integrate the boolean plugin (normalization) into the type plugin (validation) the positioning of the type plugin becomes problematic.

Yes, this is maybe the best solution for now. Is it possible to move the boolean normalization code to the type plugin without rewriting everything?

If not: we could make the "type" plugin not check booleans if the boolean plugin is mounted. To do this the boolean plugin can write something to the  common config (via config/needs), read by the type plugin.  See be878fb398ed62c8737e5fcfaf64b67c967dfeba for how something similar was done with the dump and fcrypt plugin.

> All this is so complicated that I am starting to think the plugin system needs a major overhaul.

Yes, we are working on this right now. If you have any input for this task, please do not hesitate to create an issue with proposals. But I am afraid there are only two ways:

1. either all plugins do something in the same way (like only updating origvalue if origvalue was not there before)
2. or we make frameworks **specific** for validation and transformation, see #1373 for a sketch of how type checking could work. Unfortunately, nobody is working on this as @Piankero refocused on the error handling.

> The only other solution I can think of is integrating the type system deeply into the core of Elektra.

I am not a fan of this. In most cases the modular type checking works very well, only when you want to reuse the same metadata for several types it fails. (Like with the `type` metadata.) Having a `boolean` metadata for booleans (and not `type = boolean`) would also solve the problem btw.

markus2330:
And an important general principle: We should always honor what people wrote in configuration files. There might have been a good reason why it is "yes" and not "1" and why it is "0x01" and not "1". Thus we should always preserve what users wrote, not what we think of "normal".

kodebach:
> Yes, this is maybe the best solution for now. 

My point was that there is no solution... We would have to normalize at the start of kdbSet and restore at the end of kdbSet, which is not possible at the moment.

> In most cases the modular type checking works very well

The type system could still be modular. We would just need a well defined interface for a type checking and normalization plugin (i.e. something like `elektraCheckType (Key * key)`, `elektraNormalizeValue (Key * key)`, `elektraRestoreValue (Key * key)`). This interface would be implemented via a plugin. That way the core can normalize, restore, type check whenever it is needed possibly even multiple times during `kdbGet`/`kdbSet`.

markus2330:
> My point was that there is no solution...

Of course there is:

1. integrating boolean normalization into type
2. removing type checking of boolean from type, and making the "boolean" type specification differently (`boolean` instead of `type = boolean`)

I agree with your reasoning that the modularity in this case is much more than needed, so "1." is a good and compatible solution, especially if we aim at a rewrite of the type plugin in C  in long-term (i.e., when everything else works very nicely for LCDproc :+1:). The API and the user experience for LCDproc users is clearly more important than the validation.

That the same metadata is used in different plugins for different purposes is not supported. This was a conceptual mistake of the boolean plugin from the beginning.

> The type system could still be modular. We would just need a well defined interface for a type checking ...

This is exactly what #1373 tries to do.

kodebach:
> This is exactly what #1373 tries to do.

But that is still just a normal plugin isn't it? I would have done something more similar to resolver plugins, i.e. there has to be a special plugin that implements the type system, which acts differently to other plugins and has other constraints. The default type system plugin, would just a plugin that implements a single type that accepts any value.

> Of course there is:

Let me give you a more detailed example, to show that neither of those (and nothing else) solves the following problem.

Assume the following config is mounted with `ni` at `user/tests/broken` with the plugin(s) necessary for boolean values and the `conditionals` plugin:

```ini
[booleankey]
type = boolean

[otherkey]
assign/condition = (booleankey == '1') ? ('then') : ('else')
```

Before we continue, a few assumptions:
- `condtionals` works correctly
- `assign/condition` is evaluated during `kdbGet` and during `kdbSet`, so that the values of `otherkey` are persisted. It doesn't matter whether the `conditionals` plugin is intended to work this way or whether there are different solutions. We just need a plugin that evaluates a boolean key on `kdbSet`.


If we call `kdb set user/tests/broken/booleankey 1`, the value of `otherkey` will be set to `then` and if we call `kdb set user/tests/broken/booleankey 0` it will be `else`. So far so good.

Now we call `kdb set user/tests/broken/booleankey yes`. The question now is: What happens? Ideally I would expect the value of `otherkey` to be `then`.

This will be the case, if we check the value with `kdb get` and the plugin doing boolean normalization (`type` or `boolean`) is mounted before `conditionals`. But weirdly the actual mounted file, will contain the value `else` for `otherkey`. This is simply because boolean normalization doesn't happen on `kdbSet` (in fact the opposite happens), so that the user chosen value is stored.

To solve this problem the value of `booleankey` has to be normalized before `conditionals` is executed, and (optionally) restored afterwards. But **that** is not possible. Not with a a modified `type` plugin, not with `type` + `boolean`, not with separating the boolean type into its own metadata. The problem here is not a limitation on interaction between plugins, the problem is that a plugin can only be executed once during `kdbSet`. The **only** solution that is kind of (*) possible with the current plugin framework would be to have two plugins that communicate with each other. One that normalizes boolean values at the beginning of `kdbGet` and the beginning of `kdbSet`. And another one that restores boolean values at the end of `kdbSet`.

(*) kind of, because I'm not sure plugins can be ordered this way

markus2330:
Yes, if also integrate the problem of plugins which modify keys you would need to find a total ordering of all modifications, see #907 for a partial solution (only for conditionals but not for more complex interactions as you described).

But type checkers usually do not make any changes to keys, or only normalizations which do not change the semantics of the value. In *this* case interactions do not matter because normalizations should be idempotent. (Of course they *do* matter if plugins do not understand the semantics, as you described in your example.)

It is a very interesting problem you described and you are welcome to work on this for diploma thesis. For now we need to restrict ourselves to validiation (checking) and normalization which creates no problem as long as only one plugin is involved. As @Piankero only makes trivial validation (without dependences), the merge of the type and boolean plugin should suffice for LCDproc.

markus2330:
Can I close this PR?

PR #2483, opened 2019-03-11
CM: Edit installation guide -- explicitly mention Windows, etc.

alexsaber:
Mtrnr: 01263258


markus2330:
Thank you for improving the installation guide!

It might also worth mentioning that it is possible to download Windows binaries from https://build.libelektra.org/jenkins/job/libelektra/job/master/lastSuccessfulBuild/artifact/artifacts/debian-stable-mingw-w64/

Please also add an entry into the release notes to make the build server happy.

I added you to the private repo. Matrnr is not necessary in the PR description.

Issue #1271, opened 2017-01-12
Introduce symbol versioning


markus2330:
API breaks are much less painful if old software continues to work (ABI stays compatible). Symbol versioning provides that. See `doc/todo/FUTURE` for planned changes that require introduction of symbol versioning.

It would be better to also always stay API compatible, but the proposal #600 was rejected as being too ambitious.

kodebach:
If anybody wants to tackle this, the solution that FreeBSD uses is quite neat. It is described [here](https://people.freebsd.org/~deischen/symver/library_versioning.txt). The "magical awk script" that is mentioned is [this one](https://github.com/freebsd/freebsd/blob/1d6e4247415d264485ee94b59fdbc12e0c566fd0/share/mk/version_gen.awk), as far as I can tell. The script is invoked [here](https://github.com/freebsd/freebsd/blob/1d6e4247415d264485ee94b59fdbc12e0c566fd0/share/mk/bsd.symver.mk). The master version script would be `libelektra-symbols.map` in our case.

The biggest hurdle for this issue IMO is testing. We would need separate tests for each of the symbol versions to make sure we don't accidentally break something.

markus2330:
Thank you so much for your input here!

> The biggest hurdle for this issue IMO is testing. We would need separate tests for each of the symbol versions to make sure we don't accidentally break something.

This is already done by me manually. I run the latest elektra libraries against the unit tests of the previous release (which were compiled with Elektra of the previous version).

To automate this would of course be very nice, then we would be very near to continuous releases (releasing a new version of Elektra by pressing a button) because the ABI tests are currently the point which eat up the most time during releases.

markus2330:
Finished thanks to @kodebach

PR #2498, opened 2019-03-14
Highlevel codegen

kodebach:
## Basics

## Checklist

- [x] I added unit tests

## Review

@markus2330 This PR is technically still WIP, but apart from documentation and minor changes (among others integration of the options plugin) everything is finished.

For some examples take a look at `tests/shell/check_gen.sh` and the test data in `tests/shell/gen/elektra`.

## Other notes

For now I just (rather crudely) disabled the stuff related to the old `kdb gen`. The reason I didn't remove it is, that the new `kdb gen` doesn't support the contextual values part of the old version. It is possible to support it as is, but I think there may be a better solution involving less template magic and more code generation.


markus2330:

kodebach:
If anybody has input as to why as to why the mac build jobs seem to have problems with `inttypes.h` (@sanssecours ?), I would be very thankful.

sanssecours:
> If anybody has input as to why as to why the mac build jobs seem to have problems with `inttypes.h`, I would be very thankful.

Honestly, I do not know why the test fails. It **does works on my machine**. Maybe there is something wrong with the Xcode installation on the Cirrus CI macOS images? My suggestions would be to

- test the code on Travis, and
- check if one of the answers of [this StackOverflow question](https://stackoverflow.com/questions/52509602) solves the problem

.

kodebach:
@markus2330 do you have any ideas how to get rid of the ubsan problems with the ni plugin?

markus2330:
Where do I see these problems? I only see a failing testscr_check_gen

markus2330:
Btw. we can rename the old gen to "python-gen"

kodebach:
> Where do I see these problems? I only see a failing testscr_check_gen

Yes, and the test fails because there is a ubsan problem in the nickel library used by ni:
```
../src/plugins/ni/nickel-1.1.0/src/hash.c:451:6: runtime error: unsigned integer overflow: 2575944922 + 1852143981 cannot be represented in type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/plugins/ni/nickel-1.1.0/src/hash.c:451:6 in 
```
from https://travis-ci.org/ElektraInitiative/libelektra/jobs/507313602#L2692

markus2330:
Seems to be a valid complain. Why is this triggered now?

@KurtMi: can we maybe reuse our hashing function also there? This function is a pain to maintain.

kodebach:
> Why is this triggered now?

Probably, because this one path of the hash function using values like these never came up before...
But because it is a hash function it might actually be intended that the values wrap around, which for unsigned integers is the overflow behaviour specified by the standard [since at least C99](https://stackoverflow.com/q/18195715).

markus2330:
I can only guess. Based on the code comments this second code path is actually only there to make sanitizers happy. So if @KurtMi does not have another idea, we should go for suppression.

KurtMi:
> @KurtMi: can we maybe reuse our hashing function also there? This function is a pain to maintain.

I introduced the `ELEKTRA_NO_SANITIZE_UNDEFINED` and `ELEKTRA_NO_SANITIZE_INTEGER` macros see in [opmphm.c](https://github.com/ElektraInitiative/libelektra/blob/master/src/libs/elektra/opmphm.c) line 647 for usage.

In the mentioned opmphm.c file is the hashlitte implemented, in the nickel 1.1 are more but I am not sure what functions nickel really uses.

markus2330:
@KurtMi: So we can safely suppress these errors?

KurtMi:
> @KurtMi: So we can safely suppress these errors?

Well, this is a hash function, I guess some overflows are wanted.

markus2330:
@kodebach so simply suppress these errors with the macro  ELEKTRA_NO_SANITIZE_INTEGER.

Or should we make a new macro ELEKTRA_NO_SANITIZE_UNSIGNED_INTEGER?

kodebach:
We may also try this https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#silencing-unsigned-integer-overflow

But I am not sure how to enable this feature only for certain files or parts of files...

markus2330:
I think it is better to suppress too much locally than to suppress anything globally. Creating ELEKTRA_NO_SANITIZE_UNSIGNED_INTEGER shouldn't be too difficult, see 8987165dd4064f9815a2c8be95c8fccefd431b74

kodebach:
> I think it is better to suppress too much locally than to suppress anything globally. Creating ELEKTRA_NO_SANITIZE_UNSIGNED_INTEGER shouldn't be too difficult, see [8987165](https://github.com/ElektraInitiative/libelektra/commit/8987165dd4064f9815a2c8be95c8fccefd431b74)

It would have to be ELEKTRA_NO_SANITIZE_UNSIGNED_INTEGER_OVERFLOW.
`-fsanitize=unsigned-integer-overflow` is part of the group `-fsanitize=integer`, but there is no group for unsigned integers


https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#available-checks

kodebach:
@markus2330 IMO the only thing missing here are documentation and more automatic tests. Could you please check, if there are any major features missing before I do that? (so that I don't have to update the tests over and over again)

To see how the code-generation works and what its current features are, you can take a look at the two new examples. The READMEs should tell you how to use them. (Apart from the fact that, because of the `spec` plugin's problem, you may need to define a lot of keys that should have default values)

examples/codegen/econf/spec.ini
```diff
@@ -58,11 +58,11 @@ gen/enum/type = Charset
 
 [format/#/trim]
 type = boolean
-default = true
```
markus2330: Is this not working?

kodebach: I didn't try it, the type plugin would have to be mounted of course. Whether that is automatically done by `spec-mount` I don't know.

markus2330: Yes, spec-mount should mount the type plugin if a "type" specifier is present. But if the default is transformed, I do not know: the key is added by the spec plugin, is the type plugin then executed later?

kodebach:
@markus2330 is there any reason why `kdbmacros.h` isn't installed?

kodebach:
While improving/fixing my tests, I had many problems with the `spec` plugin. Mostly it just didn't do the globbing correctly and didn't copy the metadata as expected. (Don't know the issue number right now, but I am pretty sure there already is one...)

I also increasingly think that the spec plugin is just not powerful enough, and specifications have to be integrated deeper into Elektra. Take for example this simple specification:

```ini
[mykey/_]
default = 5
```

Since `_` matches everything but array elements, one would expect that `kdb get /.../mykey/test` would return `5`. But it doesn't. The spec plugin has no way of knowing about `/.../mykey/test`. It has no value and therefore doesn't exist in the KDB. I thought maybe calling `kdb setmeta user/.../mykey/test dummy 123` might trigger the `spec` plugin to copy metadata, but it doesn't really. The result `kdb get /.../mykey/test` is an empty string in this case...

markus2330:
> @markus2330 is there any reason why kdbmacros.h isn't installed?

No, I assume it was forgotten.

> While improving/fixing my tests, I had many problems with the spec plugin. Mostly it just didn't do the globbing correctly and didn't copy the metadata as expected. (Don't know the issue number right now, but I am pretty sure there already is one...)

Please report. It is easier to mark issues as duplicate as to run into the same issue twice.

> I also increasingly think that the spec plugin is just not powerful enough

Yes, I agree. Please create new issues for proposals/bugs. (What you described with kdbset sounds like a bug). I created a meta-issue for spec features in #2632.

kodebach:
> What you described with kdbset sounds like a bug

I am not so sure... This part
>> I thought maybe calling kdb setmeta user/.../mykey/test dummy 123 might trigger the spec plugin to copy metadata, but it doesn't really. 

probably is a bug. But the original problem is an intrinsic problem of the `spec` plugin. The globbing expressions are resolved in `kdbGet`. For arrays you know the size and can create keys for all elements, but for the `_` glob you just can't do anything for keys that don't exist yet.

IMO the only real solution is to integrate the globbing functionality into (a version of) `ksLookup`. It is the only place, where you can detect that `test/a` should use the data from `test/_` as a fallback.

markus2330:
> IMO the only real solution is to integrate the globbing functionality into (a version of) ksLookup

we already had a ksLookupRE which worked with regular expressions (it actually still exists as part of the validation plugin). Afaik it was never used by anyone.

But we will hopefully soon get a new query layer, let us see how this works.

kodebach:
@markus2330 After multiple very long debugging sessions of trying to figure out, why the examples don't work, I can report this:

I was trying to find out, why the code-generation example "econf" got a key `/sw/example/econf/current/root` without the metakey `type` while calling `kdb getmeta /sw/example/econf/current/root type` prints `boolean` as expected.

- The `spec` plugin definitely has **many (!)** problems (e.g. processes the key `/sw/example/econf/current/spec/sw/example/econf/current/format/#/trim`, if `spec/sw/example/econf/current/format/#/trim` is part of the spec).
- In this case, however, I am not sure the `spec` plugin is at fault. It actually adds the key cascading key `/sw/example/econf/current/root` to the returned KeySet. Everything is as expected, including the metadata.
- The `ksLookup` code for cascading lookups is incredibly complicated and I couldn't figure out whether there is a bug in there or not. The only thing I can say that this is the line, where everything breaks:
https://github.com/ElektraInitiative/libelektra/blob/fe106c02b4680ec35e49ccfd9c99e03e0a58b0e7/src/plugins/spec/spec.c#L767
  Before this line, the key returned by a `ksLookup` contains all the metadata, afterwards it is gone.
- I am also fairly certain that this problem only occurs, if `elektraEnsure` is involved. The highlevel API calls `kdbGet` once in `elektraOpen`. A call to `elektraEnsure` first calls `kdbEnsure` and then calls `kdbGet` again. This second `kdbGet` call uses the same KeySet as the first call. It seems to me, that this is the problem. However:
  - I am not sure whether this is a bug. The documentation for `kdbGet` says:
    https://github.com/ElektraInitiative/libelektra/blob/b9d3289a900f529d73bf35926457c189689a0285/src/libs/elektra/kdb.c#L864-L866
    This makes it seem, like my scenario should work. So, _if_ there is a bug somewhere, I suspect it is in the cascading part of `ksLookup`.
  - I can work around this problem, if this behaviour is (somehow) expected, but that would require keeping a copy of the defaults KeySet passed to `elektraOpen`.

markus2330:
> The spec plugin definitely has many (!) problems

Yes, we need to reimplement it. It will hopefully be done beginning with July.

> The only thing I can say that this is the line, where everything breaks:
> https://github.com/ElektraInitiative/libelektra/blob/fe106c02b4680ec35e49ccfd9c99e03e0a58b0e7/src/plugins/spec/spec.c#L767

Looks ok, only keyDup+keyDel (next line) is unnecessary.

> Before this line, the key returned by a ksLookup contains all the metadata, afterwards it is gone.

ksLookup should have the following behavior (when unable to find a key otherwise), see elektraLookupBySpecDefault:

1. if it finds exactly the cascading key, it will immediately return it (optimization: here step 2 was already done before)
2.  it will create and return a new cascading key with the "default" data

In step 2 ksLookup modifies the keySet which is unwanted because of the hash map. Thus the spec plugin already creates all cascading keys with the default data, so that later ksLookups do not have to do it.

So with the spec plugin, ksLookup should only do step 1 (with all metadata from spec) and never need to do step 2. In the long run we should remove the step 2 in ksLookup.

> A call to elektraEnsure first calls kdbEnsure and then calls kdbGet again.

This could be the problem: kdbGet is not designed for changes in the plugins. It assumes that the plugins stay identically as long as the KDB handle lives. 

Why do you use kdbGet before calling kdbEnsure? In my understanding kdbEnsure needs to be the first call after kdbOpen and is not allowed to be called again later. In the long run we want to integrate it into kdbOpen (as suggested by you). Maybe we should introduce kdbOpen2 now and never release kdbEnsure? Especially if a later call to kdbEnsure might cause hard-to-find bugs.

With symbol versioning #1271 we could even add an argument to kdbOpen without breaking old clients.

I would definitely not introduce elektraEnsure but integrate the feature in elektraOpen. In the high-level API there is no need to repeat legacy mistakes that happened in the low-level API.

> This second kdbGet call uses the same KeySet as the first call.

kdbGet actually should not modify its parameters if the config did not change.

Maybe you can split this PR? It becomes very long...

kodebach:
First of all: I found a seemingly unrelated bug and fix it in https://github.com/ElektraInitiative/libelektra/pull/2498/commits/8b82d954dd9d8a5e0d8431a9ec8c66a32284a132. After that the bug described above went away. I am not sure know how or why, but it did. My best guess is, that now we always use the Keys from the `defaults` KeySet and basically circumvent `spec`.

> kdbGet is not designed for changes in the plugins. It assumes that the plugins stay identically as long as the KDB handle lives.

This is probably the problem, yes.

> Why do you use kdbGet before calling kdbEnsure?

`elektraOpen` calls `kdbGet`, because after `elektraOpen` the `Elektra` instance must be fully usable. But `elektraOpen` is the only call that can create an instance of `Elektra`.

> I would definitely not introduce elektraEnsure but integrate the feature in elektraOpen. In the high-level API there is no need to repeat legacy mistakes that happened in the low-level API.

That would be preferable yes, but it also means a breaking change, since the highlevel API was released already.

> Maybe you can split this PR? It becomes very long...

I am not sure that would make sense. It basically means merging code, of which we know it doesn't work correctly. Splitting out the parts that do work, would be a lot of work, since the parts that do and don't work are very intertwined. Also, the remaining bugs should all be very easy to fix.

markus2330:
> That would be preferable yes, but it also means a breaking change, since the highlevel API was released already.

You can also use symbol versioning for the high-level API.

> I am not sure that would make sense. It basically means merging code, of which we know it doesn't work correctly. Splitting out the parts that do work, would be a lot of work, since the parts that do and don't work are very intertwined. Also, the remaining bugs should all be very easy to fix.

Ok, then I am looking forward to see a "ready to merge" here soon!

kodebach:
> You can also use symbol versioning for the high-level API.

For now, I will make it work without that. Once the PR is merged, we can still decided to change over to a solution using symbol versioning before the next release. A quick google search reveals that FreeBSD seems to have a nice (but a bit complicated) solution for symbol versioning...

kodebach:
Hm, ok @markus2330 I think I need help again...

On Jenkins the [`oldstable` build tells me to use the `-std=c99` option](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2498/45/pipeline#step-698-log-233). I added it in https://github.com/ElektraInitiative/libelektra/pull/2498/commits/9538d37669531acff7d5968fc4f3486d90af2595, but the server still complains...

On Travis there are a few different problems:
- Both the Linux GCC and the Linux FULL job seem to have linker problems with `check_external_pythongen` (see [here](https://travis-ci.org/ElektraInitiative/libelektra/jobs/528497727#L2221) and [here](https://travis-ci.org/ElektraInitiative/libelektra/jobs/528497724#L2205)). I cannot reproduce either of them locally.
- The Haskell job is simply missing plugins. The question here is, whether I should add all the necessary plugins, or just disable the code-generation test for this build job (because of the time constraints)?

markus2330:
> On Jenkins the oldstable build tells me to use the -std=c99 option. I added it in 9538d37, but the server still complains...

You can use `VERBOSE=1 make` (instead of make) to see which compiler flags are used during compilation.

> Travis

@sanssecours do you have an idea here?

sanssecours:
> do you have an idea here?

Since the Clang Linux builds work fine, it might be a problem of the GCC version of the build jobs `🐧 FULL` and `🐧 GCC`. I tried [unsetting the environment variables `CC` and `CXX`](https://travis-ci.org/sanssecours/elektra/builds/528773592), which unfortunately did not help.

kodebach:

kodebach:
@markus2330 in case you missed it, this is now "ready to merge"

markus2330:
Thank you for the reminder! @sanssecours can you maybe also take a short look?

examples/codegen/tree/README.md
```diff
@@ -0,0 +1,65 @@
...
+# execute in the current directory or replace $PWD accordingly
+DIR=$PWD
+
+# for CMake
+LOC="cmake/build"
+
+# for Pkgconfig
+# LOC="pkgconfig"
+
+"$DIR/$LOC/application"
```
sanssecours: Please also add some `kdb` commands to generate an example config. The output of the command with an empty `kdb`:

```
(empty)
```

 is pretty boring in my opinion. I assume that most people that use this tutorial will just copy and paste the commands here. Hence they probably will see the same output.

examples/codegen/tree/README.md
```diff
@@ -0,0 +1,65 @@
+# Tree example
+
+This example application reads its configuration from the KDB using code-generated functions.
+It then prints a tree defined inside the KDB.
```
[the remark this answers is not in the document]
markus2330: The tree is not the problem: that it is unclear what you want to show with this example is a problem.

kodebach:
> Maybe you can create an issue to collect left-overs from here.

See #2680

kodebach:
> Maybe you can create an issue to collect left-overs from here.

See #2498

PR #2499, opened 2019-03-15
Let kdb get -v display if value comes from default

### LINKED ISSUE #2474: kdb get -v should also display default values.
It is beneficial for users to see where a value from `kdb get` actually comes from. This is especially useful when `default` values are taken. 

Unforuntately `kdb get -v` does not indicate anything like that:
```sh
kdb get -v '/sw/lcdproc/lcdd/#0/current/curses/foreground'
#> got 660 keys
#> searching spec/sw/lcdproc/lcdd/#0/current/curses/foreground, found: 
#> spec/sw/lcdproc/lcdd/#0/current/curses/foreground, options: KDB_O_CALLBACK
#> The resulting keyname is /sw/lcdproc/lcdd/#0/current/curses/foreground
#> The resulting value size is 5
#> blue
```

Proposal:
```sh
kdb get -v '/sw/lcdproc/lcdd/#0/current/curses/foreground'
#> got 660 keys
#> searching spec/sw/lcdproc/lcdd/#0/current/curses/foreground, found: 
#> spec/sw/lcdproc/lcdd/#0/current/curses/foreground, options: KDB_O_CALLBACK
#> The key was not found in any other namespace, taking the default from the metadata
#> The resulting keyname is /sw/lcdproc/lcdd/#0/current/curses/foreground
#> The resulting value size is 5
#> blue
```

e01306821:
src/tools/kdb/get.cpp
I've added the check after the final key-lookup inside the execute-method. If the printTrace-method is the better place for this please let me know and I'll find a way to retrieve the information there.
Close #2474

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

### DIFF
--- a/doc/help/kdb-get.md
+++ b/doc/help/kdb-get.md
@@ -51,17 +51,48 @@ This command will return the following values as an exit status:
 
 ## EXAMPLES
 
-To get the value of a key:<br>
-`kdb get user/example/key`
+```sh
+# Backup-and-Restore: user/tests/get/examples
 
-To get the value of a key using a cascading lookup:<br>
-`kdb get /example/key`
+# Create the keys we use for the examples
+kdb set user/tests/get/examples/kdb-get/key myKey
+kdb setmeta /tests/get/examples/kdb-get/anotherKey default defaultValue
 
-To get the value of a key without adding a newline to the end of it:<br>
-`kdb get -n /example/key`
+# To get the value of a key:
+kdb get user/tests/get/examples/kdb-get/key
+#> myKey
 
-To explain why a specific key was used (for cascading keys):<br>
-`kdb get -v /example/key`
+# To get the value of a key using a cascading lookup:
+kdb get /tests/get/examples/kdb-get/key
+#> myKey
+
+# To get the value of a key without adding a newline to the end of it:
+kdb get -n /tests/get/examples/kdb-get/key
+#> myKey
+
+# To explain why a specific key was used (for cascading keys):
+kdb get -v /tests/get/examples/kdb-get/key
+#> got 3 keys
+#> searching spec/tests/get/examples/kdb-get/key, found: <nothing>, options: KDB_O_CALLBACK
+#>     searching proc/tests/get/examples/kdb-get/key, found: <nothing>, options: 
+#>     searching dir/tests/get/examples/kdb-get/key, found: <nothing>, options: 
+#>     searching user/tests/get/examples/kdb-get/key, found: user/tests/get/examples/kdb-get/key, options: 
+#> The resulting keyname is user/tests/get/examples/kdb-get/key
+#> The resulting value size is 6
+#> myKey
+
+# Output if only a default value is set for a key:
+kdb get -v /tests/get/examples/kdb-get/anotherKey
+#> got 3 keys
+#> searching spec/tests/get/examples/kdb-get/anotherKey, found: spec/tests/get/examples/kdb-get/anotherKey, options: KDB_O_CALLBACK
+#> The key was not found in any other namespace, taking the default from the metadata
+#> The resulting keyname is /tests/get/examples/kdb-get/anotherKey
+#> The resulting value size is 13
+#> defaultValue
+
+kdb rm user/tests/get/examples/kdb-get/key
+kdb rmmeta spec/tests/get/examples/kdb-get/anotherKey default
+```
 
 To use bookmarks:<br>
 `kdb get +kdb/format`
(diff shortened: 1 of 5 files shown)


e01306821:
Thanks for your review! I'll address your findings

e01306821:
I've added shell recorder tests for /doc/help/kdb-get.md which resulted in reformatting the existing examples.
The formatting-check fails because "kdb get -v" prints a trailing space-character after options:-output

Example-output:
...
    searching dir/tests/get/examples/kdb-get/key, found: <nothing>, options: 
...

Therefore this test fails without this character. Should I ignore this failure in this case?

markus2330:
I think we can safely remove the space from the output (or even options: if there are none).

@sanssecours is this a bug in the Markdown formatter? Imho it should not reformat any source code.

sanssecours:
> Should I ignore this failure in this case?

I think it should work, if you 

- add the trailing space to the Shell Recorder test, and then
- [enclose the whole test in `<!-- prettier-ignore-start -->` and `<!-- prettier-ignore-end -->` tags](https://github.com/ElektraInitiative/libelektra/blob/master/doc/CODING.md#markdown-guidelines)

.

markus2330:
Alternatively, we could simply not output "options" if there are none.

e01306821:
Thanks for your help!

I've removed the options-output if there are no options to print.

sanssecours:
> is this a bug in the Markdown formatter?

I do not think so.

>  Imho it should not reformat any source code.

In my opinion removing trailing whitespace should be a standard feature of every formatting tool 😊. The nice thing is, that `prettier` even reformats fenced code blocks, for supported languages such as YAML and JavaScript.

> please review my pull request

I do not know, if you noticed it yet, but the test `testshell_markdown_kdb_get` [fails in the build job `debian-stable-full-ini`](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2499/4/pipeline/456). The cause of is the INI plugin, which [adds keys in between levels](https://github.com/ElektraInitiative/libelektra/blob/master/doc/tutorials/storage-plugins.md#dont-add-additional-keys). I think there are multiple options to fix the failing test:

1. exclude the test `testshell_markdown_kdb_get`, if INI is the default storage plugin,
2. mount a plugin that “behaves properly”, or
3. adapt the MSR test in another way

.

scripts/jenkins/Jenkinsfile
```diff
@@ -1199,7 +1199,18 @@ def cpuCount() {
 /* Run ctest with appropriate env variables
  * @param target What target to pass to ctest
  */
-def ctest(target = "Test") {
+def ctest(target = "Test", testName = "Testname") {
+
+  /*
+  	 We disable the test `testshell_markdown_kdb_get` if the storage plugin INI is used, because it adds keys in between levels which
```
sanssecours: I think the proper location to disable the test is the CMake code, otherwise the test will still fail, if someone uses INI as default storage plugin and runs the test suite locally. Sorry if that was not clear from the suggestions in [my previous comment](https://github.com/ElektraInitiative/libelektra/pull/2499#issuecomment-474394136).

markus2330: I am also happy with the solution in this PR (if it works). I think people should not get the impression that the INI plugin works flawlessly, as it does not. Afaik nobody (except us on the build server) currently runs the tests with INI as default storage. I think soon it will be even okay to completely remove the INI-as-default tests, as we decided to use YAML and not INI as future default storage (for good reasons).

e01306821: Thanks for your feedback!
First I'll try to figure out why my test still gets executed on debian-stable-full-ini. If I don't find my error, I'll look into the CMake code.

markus2330: The fix in the CMake code most likely is much easier as many such fixes already exist.

See doc/todo/TESTING for a list of disabled tests. Please also add your test into this list.

e01306821:
The test now requires the dump storage plugin to run. Is that an acceptable solution?
I'm however not sure why testmod_dbus fails now.

sanssecours:
> The test now requires the dump storage plugin to run. Is that an acceptable solution?

It is, in my opinion 😊. It would be great, if you could mount the `dump` plugin at the start of the MSR test and unmount it at the end. This way you make sure that the test uses the correct storage plugin. For an example, on how to do this please take a look [here](https://github.com/ElektraInitiative/libelektra/blob/master/doc/help/kdb-find.md#examples).

> I'm however not sure why testmod_dbus fails now.

That is a [known problem](https://github.com/ElektraInitiative/libelektra/issues/2439). I restarted the failed build jobs.

markus2330:
Thank you very much for your contribution! Everything is now nice and clean!

Issue #423, opened 2015-12-12
mounting: avoid missing sync plugin


markus2330:
Currently, when you change `/sw/kdb/current/plugins` you might forget to add `sync` which will might lead to data loss on later crashes (no fsync() after writing to configuration file).

We could either warn when it is not added or always add it irrelevant to the content of `/sw/kdb/current/plugins`?

Additionally, sync plugin is not available for the default backend (at least when nothing was mounted to `/` or `system/elektra/mountpoints`).

markus2330:
sync is already recommended, so we should this bug in a more general: warn if recommended plugins are missing. Lets rethink this for the next milestone.

markus2330:
There is another issue that even if something already provides sync #1394, sync still gets added.

A very simple solution would be to always add recommended plugins. Then sync would get added, too.

petermax2:
> A very simple solution would be to always add recommended plugins. Then sync would get added, too.

Does this mean that if some other plugin provides sync, then sync itself is not added?

markus2330:
>  Does this mean that if some other plugin provides sync, then sync itself is not added?

Yes, then we would not need sync to be part of `/sw/elektra/kdb/#0/current/plugins`, which always gets added. Recommendations are only added, if they were not fulfilled earlier.

kodebach:
@markus2330 

Why don't we just call `fsync()` in `resolver`? I assume it is because, there could be further modifications in `postcommit`. But (at least in `new-backend`) it would be pretty easy to enforce that the resolver plugin is always called as the last `postcommit` plugin (where it could call `fsync()`).

markus2330:
The idea is that there are different ways of implementing sync (fdatasync) and the resolver is already complicated enough as is. But yes, I am not totally against implementing it in the resolver, as it would also optimize the number of syscalls (one open/close less). This is, however, totally OT here, the issue here is about something else. I hide these two comments.

kodebach:
@markus2330 making it the resolvers obligation to sync isn't really off topic. The resolver is already required (without one nothing works), so by moving the sync there, it couldn't be missing.

PR #2501, opened 2019-03-15
quickdump v2: relative key names, reused buffer, etc.

kodebach:
A small PR with some changes to quickdump. We now only store relative key names, to enable relocation. We also reuse the buffer on `kdbGet`, for a minor speed-up.


markus2330:
Thank you for the fix! It is much better to have relative key names, see also #51.

kodebach:
From https://travis-ci.org/ElektraInitiative/libelektra/jobs/507354465#L1741:
> No plugin that provides sync could be found

@markus2330 How should I fix that? The `sync` plugin is excluded for the Travis GCC ASAN job (I don't  know why), should I re-include it or should I exclude `quickdump`? (locally GCC ASAN works, if the sync plugin is there) Or is there something I can do, so that `sync` isn't required?

markus2330:
Sync is one of the essential plugins which should always be present. In #423 we discussed about how to improve without risking data loss. So at the moment it would be best if the sync plugin is included in every build job.

@sanssecours Do you know why the sync plugin is not included?

sanssecours:
> Do you know why the sync plugin is not included?

Yes. I [excluded sync and all other non-essential plugins to improve the runtime of `🐧 GCC ASAN`](https://github.com/ElektraInitiative/libelektra/commit/1fdb15eb099192e1690639d598276bb230f6ee1c).

kodebach:
> Yes. I [excluded sync and all other non-essential plugins to improve the runtime of `🐧 GCC ASAN`](https://github.com/ElektraInitiative/libelektra/commit/1fdb15eb099192e1690639d598276bb230f6ee1c).

If we want to only include essential plugins in `🐧 GCC ASAN` shouldn't we use:
```
-DPLUGINS='dump;fstab;haskell;jni;journald;list;lua;python;python2;regexdispatcher;resolver;ruby;simpleini;spec;typechecker;xerces'
```
(the list of plugins available at the time of your commit which you didn't exclude)

Instead of starting with `ALL` and excluding a whole bunch of plugins?

sanssecours:
> …Instead of starting with `ALL` and excluding a whole bunch of plugins?

I explicitly excluded the old (and tested) plugins, since I wanted to make sure newly added plugins (that do not depend on any of the excluded plugins) are tested by `🐧 GCC ASAN`.

markus2330:
Thank you!

PR #2505, opened 2019-03-17
New type plugin

kodebach:
I first tried to add the boolean functionality to `type`, but that wasn't as easy as I thought, so I just rewrote the whole thing in C using the conversion functions from `elektra/conversion.h`.

I also added in the enum functionality, so it cooperates correctly with the highlevel API code generation.

The markdown shell tests and unittests are very closely based on the ones from the old plugins.

I am not sure whether we want to keep the old `type` plugin or not. Personally, I think we should rename this from `ctype` to `type` and rename `type` to `oldtype`, or remove it.

## Basics

## Checklist

- [x] I added unit tests

## Review


markus2330:
Thank you for this clean and beautiful solution!

You can downgrade the old plugins in infos/status even more. They are not recommended or productive anymore. Or do you see still value in them?

doc/news/_preparation_next_release.md
```diff
@@ -134,6 +134,10 @@ The following section lists news about the [modules](https://www.libelektra.org/
 
 - We fixed an incorrect format specifier in a call to the `syslog` function. _(René Schwaiger)_
 
+### CType
```
markus2330: CType -> NewType.

It can be an highlight.

src/plugins/newtype/README.md
```diff
@@ -0,0 +1,185 @@
...
+    check/enum = #3
+    check/enum/#0 = small
+    check/enum/#1 = middle
+    check/enum/#2 = large
+    check/enum/#3 = huge
+
+Only the values listed in this array will be accepted. The array indices don't have to be continuous, using e.g. only `#1`, `#2` and
+`#4` is also allowed. Just make sure `check/enum` is set to the largest index in the array.
+
+Furthermore `check/enum/multi` may contain a separator character, that separates multiple allowed occurrences.
```
markus2330: Maybe rename "multi" to "separator" or "delimiter"?

src/plugins/newtype/README.md
```diff
@@ -0,0 +1,185 @@
...
+- `string` matches any non-empty key value.
+- `empty` only matches empty key values.
+- `octet` and `char` are equivalent to each other.
+- `enum` will do enum checking as described below.
+- To use `wchar` and `wstring` the function `mbstowcs(3)` must be able convert the key value into a wide character string. `wstring`s can
+  be of any non-zero length, `wchar` must have exactly length 1.
+
+## Enums
+
+If a key is set to the type `enum` the plugin will look for the metadata array `check/enum/#`.
```
markus2330: Do the enums also support normalization? (To the representing int value.)

If not, why including it into one plugin?

src/plugins/newtype/types.c
```diff
@@ -0,0 +1,361 @@
+/**
+ * @file
+ *
+ * @brief Only a destructor
```
markus2330: Copy and paste error?

doc/METADATA.ini
```diff
@@ -1163,3 +1163,11 @@ default = dec
 usedby/plugins = hexnumber
 description = used to specify the base of an integer value. Currently only the hexnumber plugin supports this metadata. It converts any value marked as "hex" from
     hexadecimal into decimal when the configuration is read, and back when it is written. The hexnumber plugin also ensures that a value marked as "hex" starts with "0x" or "0X".
+
+[origvalue]
+status = implemented
+type = string
+usedby/plugins = ctype boolean
```
markus2330: ctype -> newtype

markus2330:
Thank you! Please open a new PR for renaming the plugins.

kodebach:
> I agree that the old plugin should be renamed (I would use "cpptype") and name the new plugin "type".

Renaming the plugin to `type` would actually be a breaking change.

src/plugins/newtype/README.md
kodebach: No that isn't supported. We could implement it, but the ultimate goal should be that applications use the highlevel API with code generation, which solves this problem differently and probably more efficiently than we could inside the plugin.

Its also not really that useful, because the int value will be stored as a string, so instead of doing a few `strcmp` (or nested switches like the highlevel API) you will have to do `strtol`. The only case where this would have significant benefits is when using the highlevel API without code generation.

markus2330: Yes exactly: everything we have in the plugins works for everyone, also those who do not use the high-level API.

markus2330:
What does "newtype" do incompatible compared to "type"? The name alone does not make it a breaking change if you make the "cpptype" providing "type".

kodebach:
There are at least these incompatibilities:
- `newtype` doesn't support `FSType` (`empty` is still supported, but only because I forgot to remove it)
- `newtype` supports `enum`, `wchar` and `wstring` as types, whereas `type` would throw an error for these
- `newtype` does not support `check/type/min` and `check/type/max`

markus2330:
I think it is okay to have these incompatibilities as they are marked to be deprecated already very long time ago. Simply describe these changes in "Compatibility" of doc/news/_preparation_next_release.md

Issue #2490, opened 2019-03-13
allow customization of error messages


markus2330:
The error messages of the some plugins are often hard to read (the syntax is not very intuitive).

A custom error message text similar to validation would be very helpful.

markus2330:
Some specifier to refer to other parts of the specification would be very helpful.

E.g. the "example" or the "description":

```
[_/term]
description=The current term of this lecture.
example=2019s
default=2019s
check/validation=[0-9]{4}[sw]
check/validation/message=%description% Should be 4 digits and then s or w (match to %check/validation%), e.g. %example%.
```

markus2330:
It would be useful to also have specifiers for the regex and the key name, see #2660

kodebach:
> also have specifiers for the regex

Wouldn't that just be the value of `check/validation` (so `%check/validation%` in your example)?

markus2330:
You are of course right. We only need for key-name and key-value special specifiers.

The question is if we should misuse the parentKey to pass all this information (including meta-data) or we should add another parameter to ELEKTA_SET_ERROR.

@Piankero what are your thoughts on this issue?

Piankero:
> It would already help if the key can be printed, then one could do the lookup in the spec.

So why don't we statically add it to the error message itself (`The key %s does not confirm to %s.`) rather than giving the user a complicated and easy to fail possibility? I would not let the user add arbitrary keys (such as the description) to the message because the consistency of error messages suffers. E.g., like in cassandra one decides to expand the description with more detail (even with multiple lines and sentences) the actual `Reason` will get lost in much text or does not conform with the `check/validation/message` anymore that it forms a sensible sentence.

markus2330:
For the regex compilation error and the failed regex I agree with @Piankero: It makes more sense to have some static text, and the message is simply added (if available and only for regex does not match errors). This way the messages are more consistent and obviously the placeholders cannot be used wrongly. (It is less flexible but the flexibility most likely is not needed as it is very unlikely that specification writers will invest a lot of time to make their error messages perfect.)

>   We could in theory completely get rid of all *_ERRORF macros.

Yes, you are right. It would be more elegant to have no _*F macros but this would be a quite big change. So I would say it is rather low priority.

> We could even detect the "%s" argument, not do the printf stuff and just use the second argument as the message.

This is most likely not needed as compilers already do such things. https://developers.redhat.com/blog/2019/08/12/efficient-string-copying-and-concatenation-in-c/

[...]

Issue #123, opened 2014-10-05
kdb get fails to output multiline values


fberlakovich:
kdb get only outputs the first line of a multiline value. This issue may also be bash related because kdb get works fine in the Eclipse console.

PR #2549, opened 2019-03-27
WIP: Hexcolor plugin

PhilippGackstatter:
## Basics

## Checklist

need to be checked.

- [x] I added unit tests

## Review


markus2330:
Thank you, you already have a lot of progress!

What your plugin currently does is basically equivalent to:
```
kdb setmeta /some/key check/validation "#([0-9a-fA-F]{3}|[0-9a-fA-F]{6})$"
kdb setmeta /some/key check/validation/message "Validation of key %s with value %s failed."
```

(at least once #2490 is implemented)

So to make it useful enough to be added to Elektra it should do more than checking for a regular expression. This extra could be:

- that it normalizes to some color format, even if given in some other format (e.g. `#123` -> `123`)
- that it has support for specifying some colors by name

PhilippGackstatter:
I would like to add specifying colors like 'red', 'blue', 'white' and have them transformed to their hex equivalent by the plugin. This could be the [Extended Color keywords list of CSS3](https://www.w3.org/TR/css-color-3/#color0) found in section 4.3. 

I could also add a transformation between two or all of the following formats.
`255,0,0`
`100%,0%,0%`
`#ff0000`

I guess at this point I could rename it to just the color plugin.

> that it normalizes to some color format, even if given in some other format (e.g. `#123` -> `123`)

I'm not sure what you mean by normalization in this case.

kodebach:
> I'm not sure what you mean by normalization in this case.

Normalization is basically just what you described as transformations. To make it proper normalization, we would decide on a standard format for colors used by all of Elektra. Your plugin would than normalize (i.e. transform) all the different representations of one color into this standard format during the `kdbGet` call and restore the original format during `kdbSet`. The `type` pluign for example does this for booleans.

Whether the `#RRGGBBAA` (we should support alpha channel as well) format is the best choice for such a standard format I am not sure. It still requires some parsing and deconstruction to actually use the color (unless your library uses strings for color, which in C seems unlikely). It might be better to convert the color into a 32-bit integer. Converting red into 4294901760 (= 0xFFFF0000, ARGB) is weird, but the color can be read easily with high-level API by just using `elektraGetLong` (if the `type = long` is set as well). After that you can just use bit manipulation to extract the individual channels.

For starters, you can implement a conversion for the colors. (The [Basic color keywords](https://www.w3.org/TR/css-color-3/#html4) would be enough I think). Whether you convert to the `#RRGGBBAA` format or to 32 bit integers doesn't really matter for now. That shouldn't be too hard to change later on. The actual hard part of all of this is correct normalization and restoring (without creating weird edge cases where the plugin doesn't quite behave like you expect). You can take a look at the `type` plugin to see how this can be achieved. The plugin implemented in a very generic way, you don't need to do that, just look at what happens to keys with type `boolean` or `enum`.

PhilippGackstatter:
Thank you for your answer! I agree that converting to an integer makes more sense. Should be more efficient and easier to handle in C. I'll implement the conversion from color names to integers next.

src/plugins/hexcolor/README.md
```diff
@@ -0,0 +1,54 @@
...
+  explanation of what your plugin does
+
+## Dependencies
+
+None.
+
+## Examples
+
+```sh
+sudo kdb mount /tests/color.ini /tests/ipaddr ini hexcolor
```
markus2330: ipaddr? "ini" not needed, you can omit it.

src/plugins/hexcolor/README.md
```diff
@@ -0,0 +1,54 @@
...
+#> Mount a config file with the hexcolor plugin
+
+kdb setmeta /tests/color/hex check/hexcolor any
+#> Check the /tests/color/hex key for validity
+
+kdb set /tests/color/hex "#fff"
+#> Suceeds, since the value is a valid hexcolor. Quotes are important!
+
```
markus2330: Also show what `kdb get` returns.

src/plugins/hexcolor/README.md
```diff
@@ -0,0 +1,54 @@
...
+```sh
+sudo kdb mount /tests/color.ini /tests/ipaddr ini hexcolor
+#> Mount a config file with the hexcolor plugin
+
+kdb setmeta /tests/color/hex check/hexcolor any
+#> Check the /tests/color/hex key for validity
+
+kdb set /tests/color/hex "#fff"
+#> Suceeds, since the value is a valid hexcolor. Quotes are important!
```
markus2330: This is not valid for shell recorder, see tests/shell/shell_recorder/tutorial_wrapper/README.md

src/plugins/hexcolor/hexcolor.c
```diff
@@ -0,0 +1,82 @@
...
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ *
+ */
+
+#include <kdberrors.h>
+#include "hexcolor.h"
+#include <regex.h>
+#include <kdbhelper.h>
+
+int is_valid_key(Key * key, Key * parentKey)
```
markus2330: can be static

src/plugins/hexcolor/testmod_hexcolor.c
```diff
@@ -0,0 +1,64 @@
...
+
+	ksDel (ks);
+	keyDel (parentKey);
+
+	PLUGIN_CLOSE ();
+}
+
+static void test_hexcolor(void) 
+{
+	test_color ("#0fb", 1);
```
markus2330: Good that there are already tests!

markus2330:
> I could also add a transformation between two or all of the following formats.

Sounds good.

> I guess at this point I could rename it to just the color plugin.

Is ok from my point. Although with the current planned feature set it would be rgbcolor. But it makes sense to have all color related functionality inside of one plugin.

> The actual hard part of all of this is correct normalization and restoring

It would be great to have a lib that does that or a tutorial describing how to do it.

> This could be the Extended Color keywords list of CSS3 found in section 4.3.

Yes, that would be nice. It should be only copy&paste anyway.

kodebach:
> It would be great to have a lib that does that or a tutorial describing how to do it.

If I find time, I might write a tutorial. A library is not really the right solution, but a special plugin template might work. Also we might need to extend the `origvalue` metadata, to support the case where multiple plugins one after another change a keys value.

markus2330:
> If I find time, I might write a tutorial.

That would be nice, as it seems to be a common problem.

> but a special plugin template might work

We should avoid templates (to be copied around) as much as possible. If they are not trivial, they cause huge problems in maintenance.

> Also we might need to extend the origvalue metadata, to support the case where multiple plugins one after another change a keys value.

We could introduce a plugin `original/<pluginname>/value`. But isn't the strategy that simply the first/last wins the easier one? (To only change `original/value` if it was not set before.)

kodebach:
> But isn't the strategy that simply the first/last wins the easier one? (To only change original/value if it was not set before.)

Yes currently it is required that  `origvalue` is only set, if it doesn't exist yet. However, this limits how well plugins can be isolated. For example, you can have the color plugin normalize into hexadecimal (because it is arguably easier) and let another plugin (e.g. hexnumber) deal with converting it to decimal, so that it can be read via the high-level API. It basically means that once a plugin sets `origvalue`, the key's value must be in a standard Elektra format.

PhilippGackstatter:
> Your plugin would than normalize (i.e. transform) all the different representations of one color into this standard format during the `kdbGet` call and restore the original format during `kdbSet`

If I store the original representation in the database, and only normalize during `kdbGet`, then why do I have to restore anything durgin `kdbSet`?

This is how I understand the plugin should work.
```sh
kdb mount config.ecf user/tests/rgbcolor dump hexcolor
kdb set user/tests/rgbcolor "#aabbccdd"
#> Create a new key user/tests/rgbcolor with string "#aabbccdd"
# I'll change that syntax
kdb setmeta user/tests/rgbcolor check/hexcolor any
# Normalizes to hex
kdb get user/tests/rgbcolor
#> 0xaabbccdd
kdb set user/tests/rgbcolor green
#> Set string to "green"
# Normalizes green to hex
kdb get user/tests/rgbcolor
#> 0x00ff00ff
```

So this is basically what I have implemented (color names are not in yet), only that I'm normalizing during `kdbSet` and store it as binary internally. So `kdbGet` currently returns `\xaa\xbb\xcc\xdd` for the first example.

So where and why would I need to restore anything?

kodebach:
The KeySet that is returned from kdbGet() has to be passed to kdbSet() and you have to call kdbGet() before kdbSet(). So any changes a plugin makes during kdbGet() will be persisted during the next kdbSet() call, unless the plugin undoes them during that call.

This doesn't become apparent in your example, because you only use a single key. If you use `kdb set user/tests/rgbcolor2 green` instead of `kdb set user/tests/rgbcolor green`, both keys will be changed in the underlying file. This is unexpected behaviour.

If my explanation is to convoluted, just mount an ini file (makes it easier to see changes), play around with getting/setting multiple keys and look at how the file changes.

PhilippGackstatter:
Ah yes, I understand that now, thanks.

src/plugins/rgbcolor/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Suceeds, since the value is a valid rgbcolor. Quotes are important!
+kdb set user/tests/color/hex "#a1C2b3"
+
+# Suceeds, since the value is a valid RGBA rgbcolor. Quotes are important!
+kdb set user/tests/color/hex "#aabbccdd"
+
+# Colors are normalized to bytes
+kdb get user/tests/color/hex
+#> \xaa\xbb\xcc\xdd
```
PhilippGackstatter: @markus2330 
Is it okay to return the bytes like that? I suppose this is good for any program that wants to read the values, but it's not so nice for the enduser.

kodebach: I think it is a bit weird to parse this string inside a C program. 

IMO we should either convert to decimal (i.e. `2864434397` in your case), to ensure optimal compatibility with the highlevel API. Or just convert to the standard `#aabbccdd` representation, as extending the highlevel API is not really hard.

markus2330: I agree with @kodebach. I would prefer the decimal variant.

PhilippGackstatter:
I have now added named colors and normalization. See the Readme for working examples. If there's anything left that I can add, let me know. Other than that, I think this is ready for a review.

The only two checks failing are broken links and testmod_dbus, which I don't really know why that's happening as it seems unrelated to my plugin. Maybe someone can help me understand that, thanks.

kodebach:
> testmod_dbus

You can ignore that one. It is very time sensitive and fails, if there is high load on the build server. I triggered a re-run.

> broken links

Those links seem fine to me... @sanssecours any idea?

src/plugins/rgbcolor/rgbcolor.c
```diff
@@ -0,0 +1,379 @@
...
+	snprintf (colorStr, 11, "%u", c);
+
+	ELEKTRA_LOG_DEBUG ("Set %s to integer %s", keyName (key), colorStr);
+	keySetString (key, colorStr);
+	keySetMeta (key, "type", "unsigned_long");
+}
+
+static char * elektraColorExpand (const char * str, ColorVariant colVar)
+{
+	// Expand #abcd to #aabbccdd or #abc to #aabbccff
```
kodebach: The documentation for the function should use doxygen. So something like:

```c
/**
 * Expand #abcd to #aabbccdd or #abc to #aabbccff
 */
static char * elektraColorExpand (const char * str, ColorVariant colVar)
```

Also, I think "Expands all color variants to the full RRGGBBAA variant." better describes what happens here. Since the returned string is allocated with `elektraMalloc`, the documentation should also state that it has to be freed with `elektraFree`.

The `malloc` could also be avoided by passing in a `char *` similarly to `elektraWriteArrayNumber`.

src/plugins/rgbcolor/rgbcolor.c
```diff
@@ -0,0 +1,379 @@
...
+
+static void elektraColorNormalizeHexString (Key * key, ColorVariant colVar)
+{
+	const char * str = keyString (key);
+	char * origvalue = elektraStrDup (str);
+
+	kdb_unsigned_long_t color;
+	if (colVar == NAMED_COLOR)
+	{
+		const NamedColor * namedColor = findNamedColor (str);
```
kodebach: If you can somehow find a way to avoid calling `findNamedColor` a second time that would be nice.

doc/news/_preparation_next_release.md
```diff
@@ -357,6 +357,11 @@ removed due to:
 - no time in future to maintain
   _(Kurt Micheli)_
 
+### RGBColor
+
+- New plugin to validate hex formatted colors (e.g. #fff or #abcd) and normalize them to rgba (#ffffffff and #aabbccdd respectively). It also has support for named colors according to the [extended color keywords](https://www.w3.org/TR/css-color-3/#svg-color) from CSS3.
```
kodebach: This is not up-to-date anymore.

src/plugins/rgbcolor/rgbcolor.c
```diff
@@ -0,0 +1,379 @@
...
+}
+
+static void elektraColorSetInteger (Key * key, kdb_unsigned_long_t c)
+{
+	char colorStr[11];
+	snprintf (colorStr, 11, "%u", c);
+
+	ELEKTRA_LOG_DEBUG ("Set %s to integer %s", keyName (key), colorStr);
+	keySetString (key, colorStr);
+	keySetMeta (key, "type", "unsigned_long");
```
kodebach: Don't set this. It only causes problems:
- in kdbSet you have to remove metadata you added in kdbGet
- you must only set the metadata, if it isn't set to some other value
- etc.

It's best, if you ignore `type` altogether. Instead add `infos/ordering = type` to the README, the the `type` plugin will always be executed after you normalized the plugin. That way the user can safely set `type = unsigned_long` themselves, if the need it.

mpranj:
The links were fixed on master. You'll need to rebase in order to fix the build.

sanssecours:
> Those links seem fine to me... @sanssecours any idea?

The test just shows temporary connection issues with those links. I restarted the build job. Now it only shows  two broken links to:

- https://github.com/sanssecours/rawdata/blob/master/YAML/keyframes.yaml

, which I unfortunately introduced somewhere in the last days. Anyway, I fixed the problem in commit bda33c3b.

src/plugins/rgbcolor/rgbcolor.c
```diff
@@ -0,0 +1,379 @@
...
+{
+	// set all keys
+	// this function is optional
+	Key * cur;
+	ksRewind (returned);
+	while ((cur = ksNext (returned)) != NULL)
+	{
+		const Key * meta = keyGetMeta (cur, "check/rgbcolor");
+		if (!meta) continue;
+		ColorVariant colVar = is_valid_key (cur, parentKey);
```
kodebach: I think you don't need this first call, if you call `elektraColorRestore` unconditionally.

doc/METADATA.ini
```diff
@@ -1060,6 +1060,14 @@ description= who should see this configuration setting?
 	- developer: settings only relevant for developers of the application
 	- internal: settings which should not be seen by anyone. (e.g. intermediate steps for calculations)
 
+[check/rgbcolor]
+type= enum
+	any
```
kodebach: AFAIK this means that the only valid value for `check/rgbcolor` is `any`. Use `type= any` instead.

markus2330: Or type=string. Or do we really want to support binary data?

kodebach: The code below only checks that the metakey exists, so the value actually doesn't matter.

markus2330: Then it should be "empty". (like [binary], see #2705)

doc/news/_preparation_next_release.md
PhilippGackstatter: I updated the text to indicate that the plugin normalizes to decimal rrggbbaa instead of hex. Is that what you meant?

src/plugins/rgbcolor/rgbcolor.c
PhilippGackstatter: I removed it.

src/plugins/rgbcolor/rgbcolor.c
PhilippGackstatter: Great catch, I removed the first call.

doc/METADATA.ini
PhilippGackstatter: The enum was still leftover from an earlier version. As @kodebach pointed out, the keys value doesn't matter, only its existence. I set the type to empty.

src/plugins/rgbcolor/rgbcolor.c
```diff
@@ -0,0 +1,379 @@
...
+	regmatch_t offsets;
+	int compile_failure = regcomp (&regex, regexString, REG_NOSUB | REG_EXTENDED | REG_NEWLINE);
+	if (compile_failure) return -1;
+	// regexec returns 0 on a successful match
+	int match = !(regexec (&regex, value, 0, &offsets, 0));
+	regfree (&regex);
+
+	const NamedColor * namedColor = findNamedColor (value);
+
+	if (namedColor != NULL)
```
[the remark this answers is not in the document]
PhilippGackstatter: I moved the check before the hex-check. I assume you suggested this because it is faster than the regex check?

src/plugins/rgbcolor/rgbcolor.c
```diff
@@ -0,0 +1,379 @@
...
+	}
+	return NULL;
+}
+
+static ColorVariant is_valid_key (Key * key, Key * parentKey)
+{
+	const Key * meta = keyGetMeta (key, "check/rgbcolor");
+	if (!meta) return 1;
+	const char * value = keyString (key);
+	const char * regexString = "^#([0-9a-fA-F]{3,4}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})$";
```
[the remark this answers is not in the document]
PhilippGackstatter: Great suggestion, thank you!

src/plugins/rgbcolor/rgbcolor.c
```diff
@@ -0,0 +1,379 @@
...
+	case 4:
+		return HEX_THREE;
+	case 5:
+		return HEX_FOUR;
+	case 7:
+		return HEX_SIX;
+	case 9:
+		return HEX_EIGHT;
+	default:
+		return COLOR_INVALID;
```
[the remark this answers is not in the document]
PhilippGackstatter: I removed the default case entirely.

src/plugins/rgbcolor/rgbcolor.c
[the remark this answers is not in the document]
kodebach: > I assume you suggested this because it is faster than the regex check?

Not necessarily faster, but you always computed the regex result, even if you didn't use it.

PS. You don't need to answer all the comments, if you don't have questions and just did want I suggested a thumbs up (or no reaction at all) is fine.

src/plugins/rgbcolor/rgbcolor.c
markus2330: Is this done now?

markus2330:
Amazing job, is this ready to merge?

@kodebach  I wonder if this plugin works together with hexnumber?

src/plugins/rgbcolor/rgbcolor.c
PhilippGackstatter: I haven't found a reasonably simple way to avoid this. I think this would require a rewrite of how the validity of a key is checked and wherever it is called, which is a fairly big part of the plugin. I would leave it as is, if that's fine.

markus2330: Yes, of course it is fine.

PhilippGackstatter:
> Amazing job, is this ready to merge?

@markus2330 
From my side this is ready :slightly_smiling_face:

kodebach:
> I wonder if this plugin works together with hexnumber?

The plugin now converts to decimal, so there is no need for that.

markus2330:
Thank you!

Issue #2456, opened 2019-03-03
kdb ls produces additional keys


kodebach:
## Steps to Reproduce the Problem

Have the spec plugin (global-)mounted and run:

```sh
kdb ls spec/tests/example
```

## Expected Result

Only keys below `spec/tests/example` should be listed.

## Actual Result

Cascading keys below `/tests/example` are also listed.

## Further Comments

IMO `kdb ls` should never produce cascading keys, even if it is called with a cascading parent key. If I call `kdb ls`, I want to know about the actual keys present.

markus2330:
Thank you for reporting this problem. I fully agree!

Only with `-v` these keys should be part of the output.

DaHopeman:
I would like to do this as my first issue @markus2330

markus2330:
Good choice, this should be very easy. I assigned you.

DaHopeman:
@markus2330 
I just want to make sure I understood the problem correctly:
using `kdb ls` should just list those keys that are directly in the given path and not "deeper down", right?
like this:
![Screenshot from 2019-03-27 19-28-34](https://user-images.githubusercontent.com/11979093/55102418-a2684580-50c6-11e9-9e6a-f056d1b78730.png)

markus2330:
No, "deeper down" is a separate feature using -m and -M switches (man kdb ls). You should only filter out the keys which start with / and not filter the correct keys.

You get such keys by setting defaults:
```sh
kdb setmeta spec/tests/example default something
kdb ls spec/tests
#> spec/tests/example
```

Wanted output:
```
spec/tests/example
```

Current output:
```
/tests/example
```

Or also (maybe a different bug?):
```sh
kdb setmeta spec/tests/example default something
kdb ls spec/tests/example
#> spec/tests/example
```

Please also add test cases (the code above is already a shell recorder test, see tests/shell/shell_recorder/tutorial_wrapper/README.md)

DaHopeman:
I have modified the `printResults` method in my fork to filter out the keys which start with /.
I did not create a pull request however, because I am not sure if this is really the problem for this bug.

Using the input you mentioned above:
```
kdb setmeta spec/tests/example default something
kdb ls spec/tests
```

I get the following output:
```
/tests/example
spec/tests/example
```

I get both versions every time. Is this intentional or could there be a problem in the `cut` function, which is used to provide the Keyset for `printResults`?

kodebach:
> could there be a problem in the cut function

[...]

Issue #972, opened 2016-09-19
Lookup should not modify KeySet


markus2330:
Currently, ksLookup modifies the KeySet when a not-present default is returned. This would be harmful with a hashmap present, so we need alternative solutions:

1.) Already add the default keys during `kdbGet` and only calculate the values during `ksLookup`.
2.) Put the default keys at the end/beginning of the keyset and exclude them from the hashmap.

markus2330:
Should be fixed via default value calculation.

markus2330:
Now we only need to remove the code creating the keys within ksLookup to complete this proposal.

markus2330:
Now we only need to remove the code creating the keys within ksLookup to complete this proposal.

KurtMi:
> Now we only need to remove the code creating the keys within ksLookup to complete this proposal.

I assume code creating the keys refers to the `elektraLookupBySpecDefault (...)` function.
Does the default value creation needs to be switched on in some way? Because removing the function leads to failing test cases.

markus2330:
The spec plugin needs to be present (which should be the case by default). Maybe the spec plugin is buggy and does not always create default keys or it creates them by looking them up. Seems like we need a change in the spec plugin, too.

Is it possible for you to invalidate the hash map during ksLookup? Such a feature seems to be necessary anyway (for KDB_O_POP) and might also be interesting for user-supplied hooks. For performance reasons users obviously should always avoid changes of KeySets in ksLookup hooks but it seems to be expensive to guarantee that a hook did not modify the KeySet (at least if we would use run-time checks).

PR #2576, opened 2019-04-03
keyhelper: Fix calculation of relative path

### LINKED ISSUE #2460: kdb import broken in spec namespace

```sh
echo -e "[test]\n default = value" | kdb import spec/tests/import ni

echo -e "[test]\n default = value" | kdb import spec/tests/import ni
```

## Expected Result

The second line doesn't change anything about the KDB (except file timestamp, etc.)

## Actual Result

Something inside of `kdb import` (most likely the `spec` plugin, as this only occurs with the spec namespace) butchers the key name and the following error is shown:

```
The command bin/kdb import terminated unsuccessfully with the info:
the supplied key spec/tests/importt is not below the old parent spec/tests/import
```

Note: `spec/tests/import/test` somehow became `spec/tests/importt`.

## Further Information

- `kdb import -v` doesn't provide any additional information.
- The parent key or added keys don't matter as long as the spec namespace is used.
- It is always the `/` and the first 3 characters after the parent key name that go missing.

e01306821:
src/plugins/spec/spec.c

This PR prevents canonical keys to be added to the returned KeySet in elektraSpecGet.
Close #2460
 
## Basics

- [X] Short descriptions should be in the release notes (added as entry in

## Checklist

need to be checked.

- [ ] I added unit tests

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -281,7 +281,7 @@ you up to date with the multi-language support provided by Elektra.
 - `kdb get -v` now displays if the resulting value is a default-value defined by the metadata of the key. _(Thomas Bretterbauer)_
 - `kdb cp` now succeeds if target-keys already have the same values as the source-keys. _(Thomas Bretterbauer)_
 - `web-ui` issue #2441 is fixed in form of a workaround _(Josef Wechselauer)_
-- <<TODO>>
+- `kdb import` does not fail anymore if executed more than once with the same target in the spec-namespace. _(Thomas Bretterbauer)_
 - <<TODO>>
 - <<TODO>>
 
--- a/src/libs/tools/src/helper/keyhelper.cpp
+++ b/src/libs/tools/src/helper/keyhelper.cpp
@@ -40,7 +40,17 @@ string rebasePath (const Key & key, const Key & oldParent, const Key & newParent
 		throw InvalidRebaseException ("the supplied key " + key.getName () + " is not below the old parent " +
 					      actualOldParent.getName ());
 
-	string relativePath = oldKeyPath.substr (actualOldParent.getName ().length (), oldKeyPath.length ());
+	string relativePath;
+	if (oldKeyPath[0] == '/')
+	{
+		string actualOldParentName = actualOldParent.getName ();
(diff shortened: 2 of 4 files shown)


markus2330:

markus2330:
Yes, this filtering only makes sense in `kdb ls` (#2456) but for the spec plugin it is a feature that these keys are added in the spec plugin (for #972).

Please document, test and cleanup in the spec plugin while you are researching for the solution of this issue. This would be badly needed. This would be much more welcomed then yet another validation plugin.

kodebach:
Should we maybe use `kdbEnsure` to unmount `spec` in `kdb ls`, or is there a case where `spec` is needed in `kdb ls`?

markus2330:
> Should we maybe use kdbEnsure to unmount spec in kdb ls, or is there a case where spec is needed in kdb ls?

Nice idea but, I think we should list these keys if `-v` is given. Imho it makes sense to list all (default) keys. Furthermore: I would prefer to have `spec` available, because if `spec` messes something up, then at least we see it.

e01306821:
Okay, sorry, it seems that I misinterpreted some things while debugging. Both keys (with and without spec-prefix) are returned from elektraSpecGet, which is, if I am not mistaken, the expected behavior. Therefore I think that the error lies in src/tools/kdb/import.cpp, because for other namespaces no cascading names for keys are returned except for spec. And this cascading name seems to be the issue when merging the KeySets in import.cpp which means that import.cpp does not handle this case.
Therefore I don't know how to fix this without filtering out these cascading names. Or should merging take care of this case?

kodebach:
If you try to find the error message that is output in this case, you will notice that it comes from the `tools` library, not from `kdb`. This leads me to suspect that the bug might actually be in our merge code.

Maybe you can debug the process and find out were exactly the keyname `spec/tests/import/test` turns into `spec/tests/importt`. Somewhere it obviously has to be broken apart and put together wrong.

e01306821:
src/bindings/cpp/include/keyset.hpp

There the calculation of "relativePath" in the method "rebasePath" leads to this error if the key starts with "/".
I'll add a fix there.

markus2330:
Great to have this fix!

@e01306821 It would be good to always have test cases that demonstrate that the problem was fixed. Can you maybe add such a test case in ~e/src/libs/tools/tests/testtool_keyhelper.cpp?

PR #2575, opened 2019-04-03
JNA: Improve Maven Integration Into Build System

sanssecours:

Issue #1075, opened 2016-11-10
cmake: changes README.md do not trigger rebuild


markus2330:
Cmake is not called automatically when README.md are changed

kodebach:
Pretty sure this link is solved by line 629 in LibAddMacros.cmake (at least for those README.md files that use generate_readme):

https://github.com/ElektraInitiative/libelektra/blob/ab01f77af560dd98c424e2e3df2ffeed4f3d2730/scripts/cmake/Modules/LibAddMacros.cmake#L628-L629

markus2330:
Jup, worked for me in the dbus plugin.

PR #2588, opened 2019-04-06
WIP Try to set up the new merge tool

Chemin1:
This commit doesn't compile. I just wanted to verify that this basically the structure that we want before I dig deeper into why the linker does not find the function definition for mymerge() in the .c file.

What I stumpled upon was:

`We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. 
`
https://cmake.org/cmake/help/v3.14/command/file.html

Is there a specific reason why we use GLOB often?

### DIFF
--- a/src/libs/tools/include/CMakeLists.txt
+++ b/src/libs/tools/include/CMakeLists.txt
@@ -1,5 +1,6 @@
 file (GLOB HDR_FILES
-	   *.hpp)
+	   *.hpp
+	   newmerge.h)
 
 install (FILES ${HDR_FILES} DESTINATION include/${TARGET_INCLUDE_FOLDER})
 
--- a/src/libs/tools/include/newmerge.h
+++ b/src/libs/tools/include/newmerge.h
@@ -1,0 +1,15 @@
+#ifndef NEWMERGE_H_
+#define NEWMERGE_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+int mymerge (void);
+int test (void);
+
+#ifdef __cplusplus
+}
+#endif
+#endif
--- a/src/tools/kdb/factory.hpp
+++ b/src/tools/kdb/factory.hpp
@@ -47,6 +47,7 @@
 #include <metaset.hpp>
 #include <mount.hpp>
 #include <mv.hpp>
+#include <newmerge.hpp>
 #include <remount.hpp>
 #include <rm.hpp>
 #include <set.hpp>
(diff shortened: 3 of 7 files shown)


markus2330:
> Is there a specific reason why we use GLOB often?

Because it is much easier to write and maintain CMake files that contain globs. You do not need to update the CMake file for every added file.

Currently, there are many bugs that rebuilds are not triggered on some changes, e.g. see #1075. In #2575 we recently avoided globbing so that rebuilds work properly.

To avoid such rebuild bugs, it is certainly better to list the files.

src/libs/tools/include/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 file (GLOB HDR_FILES
-	   *.hpp)
+	   *.hpp
+	   newmerge.h)
```
markus2330: kdbmerge.h can be directly in src/include (it will be an independent lib, unrelated to libtool)

src/libs/tools/include/newmerge.h
```diff
@@ -0,0 +1,15 @@
+#ifndef NEWMERGE_H_
```
markus2330: please also prefix (see other header files in src/include)

src/libs/tools/include/newmerge.h
```diff
@@ -0,0 +1,15 @@
...
+#define NEWMERGE_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+int mymerge (void);
```
markus2330: Please add real API now. It is always good to think about the API asap.

src/tools/kdb/factory.hpp
```diff
@@ -110,6 +111,7 @@ class Factory
 		m_factory.insert (std::make_pair ("file", std::make_shared<Cnstancer<FileCommand>> ()));
 		m_factory.insert (std::make_pair ("sget", std::make_shared<Cnstancer<ShellGetCommand>> ()));
 		m_factory.insert (std::make_pair ("merge", std::make_shared<Cnstancer<MergeCommand>> ()));
+		m_factory.insert (std::make_pair ("newmerge", std::make_shared<Cnstancer<NewMergeCommand>> ()));
```
markus2330: Maybe call it "cmerge". In any case this is temporary, the goal is to have all current commands replaced with your merging.

src/libs/tools/include/CMakeLists.txt
Chemin1: This was a really helpful hint into the right direction. Thank you!

Chemin1:
> To avoid such rebuild bugs, it is certainly better to list the files.
Sounds good for me. 

I already included your suggestions. Linking the header files is not working yet. I hope to be able to resolve this soon.

markus2330:
Good to hear that there is progress.

> Linking the header files is not working yet.

Header files do not need to be linked.

Issue #2572, opened 2019-04-02
Error Concept Improvements Metaissue


Piankero:
Summary of improvement TODOs from PR https://github.com/ElektraInitiative/libelektra/pull/2435

- [x] Write guidelines how to categorize errors.
- [x] Please put memory allocation errors in its own category so that they are easy to find in future.
- [ ] Reduce the size of kdberror.h by refactoring (https://github.com/ElektraInitiative/libelektra/pull/2435#discussion_r269786368)
- [ ] Possibly add a macro `ELEKTRA_ADD_SOLUTION` which can serve as new method of supplying a solution (https://github.com/ElektraInitiative/libelektra/pull/2435#discussion_r269769574)
- [x] Error code format changes for Warning/Errors (https://github.com/ElektraInitiative/libelektra/pull/2435#discussion_r269767654)
- [ ] Make a concept for the API (https://github.com/ElektraInitiative/libelektra/pull/2435#pullrequestreview-219722372)
- [ ] Errno should be handles by error framework
- [ ] Checker tool to see for unused errors/warnings (https://github.com/ElektraInitiative/libelektra/pull/2435#discussion_r267290712)
- [ ] Optimize the dynamic dispatching in the `kdberrors.h` to not `strcmp` for the full code but for the first digits and then for the full string. This reduces the `strcmp` calls significantly if there are lots of error codes. This performance optimization though should be more relevant if we have much more error codes (currently only 8) (Idea from @kodebach )
- [x] Differentiate the error by the name rather than by the parameter, eg. `ELEKTRA_ERROR_RESOURCE(...)` instead of `ELEKTRA_ERROR(RESOURCE,...)`. (https://github.com/ElektraInitiative/libelektra/pull/2589#discussion_r273174729)
- [ ] Change the warning key structure to not be limited to 100 entries but instead use arrays (https://github.com/ElektraInitiative/libelektra/pull/2589#discussion_r272829754)
- [x] Incorporate an `INFO` level (https://github.com/ElektraInitiative/libelektra/issues/2610)
- [ ] Add ElektraError accessor functions to highlevel API (@kodebach idea)

markus2330:
Thank you for starting collecting the issues!

Please collect all problems/open discussions, even if you maybe will not solve them. E.g. "errno should be handled by error framework".

Piankero:
In #2435 we never talked about errno handling, I think it was another discussion which I do not remember unfortunately.
[...]

PR #2589, opened 2019-04-07
New error codes

Piankero:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

## Merging


Piankero:
Some remarks about the PR:

1. Unfortunately I was not able to fix some test cases, especially `/media/wespe/extended/repository/libelektra/tests/kdb/testkdb_error.cpp`. My C++ skills are not sufficient enough for those because of types. The new errors are Strings but C++ treats them as octal. I was not able to use `EXPECT_STREQ`. Also line 114-131 should not work anymore if I am correct because they assume consecutive error codes. Could anybody help me in this regards?
2. @kodebach, can you check if your highlevel implementation is still working? Especially the part where you take the lowlevel error and take it to the highlevel. I am not sure if it takes it as string.
3. I migrated all errors but I was not always 100% sure if they are in the correct category. In case of doubt I gave them into the `Installation` category. Please let us not discuss to long about that because nearly every commit will cause a merge conflict.
4. In some cases I incorporated the description from the `specification` into the reason because the text would otherwise be useless.
5. Kudos to @sanssecours and @kodebach for their error messages. They were self explanatory and I rarely had to change stuff there :)

.gitignore
```diff
@@ -5,6 +5,8 @@
 
 #ignore backups
 *~
+.idea/vcs.xml
```
Piankero: I know that these files should be checked in from intellij but mine is a bit different and forces me to commit it before I checkout a different branch. If someone *really* wants to check in their vcs.xml he should simply force add it.

Also .vscode always adds the folder everywhere (also subfolders). I added it in this regards

src/libs/highlevel/elektra.c
```diff
@@ -216,7 +216,7 @@ void elektraSaveKey (Elektra * elektra, Key * key, ElektraError ** error)
 				return;
 			}
 
-			if (elektraKDBErrorCode (kdbSetError) != 30) // ELEKTRA_ERROR_CONFLICT = 30
+			if (elektraKDBErrorCode (kdbSetError) != 03000) // CONFLICT = 03000
```
Piankero: @kodebach, can you help me fix this problem? It is treated as octal and I do not know what to change for your high level api

.gitignore
kodebach: In what way is it different? If the IDE adds one or more lines like this:
```
    <mapping directory="$PROJECT_DIR$/build/googletest-src" vcs="Git" />
```
simply remove them and maybe mark the directory `$PROJECT_DIR$/build` (`build` may be something different) as excluded. This happens, because our CMake checks out the googletest source (recommended this way). The IDE thinks, we might want to commit to this repo, but we don't so we can just remove the line.

src/libs/highlevel/elektra.c
kodebach: 1. Is the error code `03000` used exclusively for the cases were `30` was used before? If not please create a separate error code, otherwise this cannot ever work.
2. Please use the macro instead of the explicit code. The reason we didn't use the macro before is that there was no macro when the code was written.
3. You need to change `elektraKDBErrorCode` to return a `const char *` and use `strcmp` here, now that error codes are strings.

src/error/specification
```diff
@@ -1,5 +1,5 @@
 number:0
-description:a test for an exception specification
+description:A test for an exception specification
 :new lines to the same entry can be started with a colon
 :the number is sequential and has no special meaning
```
markus2330: Not updated.

src/error/specification
```diff
@@ -17,1331 +17,34 @@ ignore:true
...
 
-number:1
-ingroup:modules
-description:could not load module, dlopen failed
-module:dl
-severity:warning
-class:FailedLoadModuleError
-extends:ModuleError
+number:01100
+description:Resource
```
markus2330: Please properly describe what the error/warning is for.

src/error/specification
```diff
@@ -17,1331 +17,34 @@ ignore:true
...
+macro:RESOURCE_CODE
```
markus2330: The previous system was better. Here should be just one word, and then some common prefix should make that the macros do not clash with other macros.

src/error/parser.cpp
```diff
@@ -49,8 +49,7 @@ parse_t parse (std::string const & file)
 			int cmpNumber;
 			std::istringstream istr (text);
 			istr >> cmpNumber;
-
-			if (number != cmpNumber) throw parse_error ("Given number for that entry wrong", linenr);
+			// std::cout << cmpNumber << endl;
```
markus2330: Please check if the given number is as it should be.

src/libs/highlevel/elektra.c
kodebach: Also `doc/decisions/error_message.md` says [Conflict is 02000](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/error_message.md#decision).

src/error/specification
Piankero: Do you have an suggestion for the prefix? Warning/Error should not be in the prefix

kodebach: Why should error not be in the prefix? It is an error code after all...

Piankero: Because it can also be used as a warning

src/error/exporterrors.cpp
```diff
@@ -102,201 +102,186 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
...
-			continue;
-		}
 
-		if (p[i]["severity"] == "warning")
+		for (int f = 0; f < 2; ++f)
 		{
-			for (int f = 0; f < 2; ++f)
+			if (f == 0)
+			{
+				os << "static inline void elektraAddWarningf" << p[i]["number"] << "(Key *warningKey, const char *reason,"
```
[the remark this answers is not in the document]
Piankero: Markus requested that they stay the same as errors and warnings are interchangeable

src/error/exporterrors.cpp
```diff
@@ -307,64 +292,35 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
...
-		if (p[i]["unused"] == "yes")
-		{
-			continue;
-		}
-
-		os << "		keyNew (\"system/elektra/modules/error/specification/" << i << "\"," << endl
+		os << "		keyNew (\"system/elektra/modules/error/specification/" << p[i]["number"] << "\"," << endl
 		   << "			KEY_END)," << endl
-		   << "		keyNew (\"system/elektra/modules/error/specification/" << i << "/description\"," << endl
+		   << "		keyNew (\"system/elektra/modules/error/specification/" << p[i]["number"] << "/description\"," << endl
```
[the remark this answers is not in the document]
Piankero: This pr is Just for Error codes. The New Message With no description will be done in a separate PR

src/error/specification
kodebach: But as you said yourself below, errors and warnings should be interchangeable. So essentially a warning is just a special kind of error. If you really want to keep the semantics neat, just generate to macros one with the prefix `ELEKTRA_ERROR_` and one with `ELEKTRA_WARNING_`.

markus2330: Alternatively we could encode the error directly in the issuing macro, e.g. `ELEKTRA_ADD_RESOURCE_WARNING` or `ELEKTRA_SET_RESOURCE_ERROR`. This macro would simply have one parameter less.

As there are so few errors/warnings now, which all can be described with one/two words, this should work?

I am not sure though if this will conflict with the improvements of a smaller kdberrors.h as suggested by @kodebach.

src/error/exporterrors.cpp
```diff
@@ -307,64 +292,35 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
...
 		   << "			KEY_VALUE, \"" << p[i]["description"] << "\", KEY_END)," << endl
-		   << "		keyNew (\"system/elektra/modules/error/specification/" << i << "/ingroup\"," << endl
+		   << "		keyNew (\"system/elektra/modules/error/specification/" << p[i]["number"] << "/ingroup\"," << endl
```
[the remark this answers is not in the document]
Piankero: ingroup will be left completely, but module (or "plugin") will stay. When I am thinking about that I will need to add an additional parameter to the triggerError/Warning to pass in the plugin name in order to show it. :/ Or is there any other option you could think of?

src/error/specification
Piankero: Just because the number of errors are so few at the moment, it does not mean that this will not grow again. I also do not think that this is more intuitive for devs. The parameter omission comes with the price for knowing the exact macro name.

Piankero:
> Thank you for the large PR! Good to have progress here.
> 
> Why do you prefix all numbers with 0 if you have troubles with octal numbers?

The error code system is a String system like SQLStates starting from 0 to 9 and then from A to Z. By having 2 characters left for the "category" we can very easily extend the categories or even think of a naming convention like Postgres did by saying that new errors to the SQLStates generated from Postgres should include a `P`, eg. `01P43`.

.gitignore
Piankero: Thank you for this workaround.
I have no problem with doing that but every developer who joins elektra will always need to do that which is kind of unnecessary in my opinion, especially if the vcs.xml does not change at all anyway in the repository. What do you think @markus2330

src/libs/highlevel/elektra.c
Piankero: >1. Is the error code 03000 used exclusively for the cases were 30 was used before? If not please create a separate error code, otherwise this cannot ever work.

Do you really require a completely separate highlevel-only error code for this conflict? Which error from which file do you explicitly want to catch with this codeblock?

src/error/exporterrors.cpp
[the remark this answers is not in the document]
kodebach: Module/plugin will stay, yes but it won't come from the specification, so these lines here are wrong.

Also for plugins we inject a macro via CMake that contains the plugin's name. Take a look at kdbplugin.h to see how you can use it. You could use a similar method to inject a general macro everywhere that contains the module, i.e. plugin name, library name or just kdb/core.

.gitignore
kodebach: I just saw you didn't remove the file from the repository, only added it to .gitignore. That is definitely a bad solution. We should remove it completely, if we don't want the file.

src/libs/highlevel/elektra.c
kodebach: This code catches the error that has code 30 in the old system. It is used in the resolver, if the storage file is opened exclusively by another process (in which case this code amounts to busy waiting). It is also used, if the file was modified by another process between kdbGet and kdbSet. This normally requires a merge. But the high-level API only changes one key at a time. Therefore we can just repeat the set procedure for the same result.

src/error/exporterrors.cpp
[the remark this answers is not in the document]
markus2330: Yes, injecting the plugin name within the macro is an excellent idea.

src/error/exporterrors.cpp
```diff
@@ -102,201 +102,186 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
...
+			}
+			else
+			{
+				os << "static inline void elektraSetError" << p[i]["number"] << "(Key *errorKey, const char *reason,"
+				   << endl
+				   << "	const char *file, const char *line)" << endl;
+			}
+			os << "{" << endl
+			   << "	if (!errorKey) return;" << endl
+			   << "	char buffer[25] = \"warnings/#00\";" << endl
```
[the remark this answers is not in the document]
markus2330: Yes, we should use proper Elektra arrays here without any limit.

src/error/specification
markus2330: I hope they will not grow again!

.gitignore
markus2330: Of course we want to make the entrance of new developers as smooth as possible. @kodebach is right about .gitignore.

Piankero: By removing this file from the project the same problem will remain. Whoever uses CLion has to manually exclude the .idea folder to avoid git always trying to add the file. The file being checked in is ok as [jetbrains](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems) suggests. But any changes to it should not be checked in anymore

src/error/specification
markus2330: > The parameter omission comes with the price for knowing the exact macro name.

Does it really matter if you need to know the error macro code name or error macro name?

The main difference is that not two but only one macro need to be prefixed, leading to an overall shorter call:

```
ELEKTRA_SET_RESOURCE_ERROR(...
vs.
ELEKTRA_SET_ERROR(ELEKTRA_RESOURCE_CODE, ...
```

The last time we discussed it, however, there were people against this approach. One disadvantage is, that the code cannot be calculated. This should be very minor, though.

Please do not forget to write design decisions about these topics, so that we have it documented. Please update #2572.

kodebach: > One disadvantage is, that the code cannot be calculated

The current code also doesn't allow calculated codes. In fact they have to be known to the preprocessor not just to the compiler. If codes aren't statically known, you have to use the `elektraTriggerError` function, but that means you don't get compile time error for unspecified error codes anymore.

src/error/parser.cpp
```diff
@@ -52,6 +52,7 @@ parse_t parse (std::string const & file)
 			bool isHighlevelFile = (file.find ("highlevel") != string::npos);
 			if (!std::regex_match (text, codeRegex) && !isHighlevelFile)
 				throw parse_error ("Error code does not match regular expression [0-9A-Z]{5,5}", linenr);
+			text = "C"+text;
```
Piankero: @markus2330 is it ok to deal it in that way or should we enforce it in the specification file for consistency (wysiwyg)

.gitignore
```diff
@@ -68,3 +69,5 @@ nbproject/
 modules.xml
 .idea/misc.xml
 *.ipr
+**/.vscode/*
+c.tmp
```
kodebach: what is `c.tmp`?

Piankero: vscode generates it in the root folder

kodebach: I have never see that while working with vscode... Normally vscode should write all its files to .vscode (or somewhere in your home directory)

If you are absolutely sure that this file has to be gitignored and it wasn't just a one time thing, please add a comment saying what `c.tmp` is and why it is ignored.

Also the suggested gitignore for vscode is:
```gitignore
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
```
https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore

Piankero:
@markus2330 : I added a commit indicating unsure categorizations: https://github.com/ElektraInitiative/libelektra/pull/2589/commits/058d194c21efeeceed183d8f4a2724fa6d58a128

Many of those are very similar. Thanks for your help

src/error/parser.cpp
markus2330: I do not know what you try to make here. Yes, in general consistency should be enforced.

Piankero: Either we force the user to write in the specification file `01000` and the parser prepends the `C` or we force the user to write `C01000` into the specification file. We agrees in the meeting that error codes should have a prepended `C`.

Piankero:
Thank you for looking into the errors @markus2330.
I see that you add various suggestions for subcategories (API error, process error, memory allocation, etc.)
Those errors seem to be nice concerning categorization but do they really force a different behavior programmatically? Does it make a difference if it is an API error or any other internal bug? Also concerning parsing: Transformation or anything else? We seem again to bloat up the categories unnecessarily. The question for every new category should be: "Can I react differently if I would know every error X in category Y? (try different value, automatic permission grant, etc.) For most of these cases I do not see a valid point. Also what comes when trying to subcategorize an existing category is that you need to find a category for the other errors. Take `Transformation` errors as you suggested. So parsing errors now would be `Transformation Errors` and ...:`General Parsing Errors`? You wanted to avoid such "Miscellaneous" categories.

kodebach:
> The question for every new category should be: "Can I react differently if I would know every error X in category Y? (try different value, automatic permission grant, etc.)"

I don't think this should be the only question we ask ourselves. I think another question should be: "Would a new developer be confused by the category chosen for error X?" For example assume there is category P, which is only used for file permission errors so far. Now we put a memory allocation error into this category, because there isn't any other category that fit better. So a new developer might think: "Well an integer overflow is kind of like a memory allocation error, let's put that into category P". At what point do we create a new category? Creating a new category only once the assignments become ridiculous enough, also means huge refactorings every time a new category is created.

Also, if there are many different errors in the same category, you cannot google the error code anymore.

src/error/parser.cpp
markus2330: I see. Yes, better let the user write the complete code in the specification and enforce the C in the beginning.

Piankero:
> So a new developer might think: "Well an integer overflow is kind of like a memory allocation error, let's put that into category P"

This more or less seems to be a naming problem rather than a categorization problem. Fail categorization can happen with any system. It's up to the maintainers too to enforce consistency.
For me the errors only differ concerning possible reactions, not the type of error itself. 

> At what point do we create a new category?

If there is the need for a different behavior for this category. @markus2330 said that memory allocation errors are treated differently and even the macro should do additional things. This is a perfect use case for a new category.

> ... also means huge refactorings every time a new category is created.

That is true. But categories should be very stable and huge refactorings should rarely be the case. This here should be the last refactoring of this size anyway

> Also, if there are many different errors in the same category, you cannot google the error code anymore.

Is it that common that you google for such specific elektra errors? Neither Github nor Google ever helped me in these regards.

src/error/specification
```diff
@@ -1,1347 +1,44 @@
-number:0
-description:a test for an exception specification
+number:00000
```
sanssecours: I definitely like the shortness of this file. However, while I think I can grasp what a

- parsing,
- conflict,
- timeout, and
- validation

error is, I am not sure about the categories. Can you please add a short description and one example for the other errors? Otherwise, I do not think I will be able to add any useful comment to commit 058d194c.

Piankero: I am currently working on a description which will come in the next commit! :)

kodebach:
> This more or less seems to be a naming problem rather than a categorization problem.

You think we should categorise by possible reaction. So I'll just ask you this:

_What are your proposed reactions to the current categories in this PR?_

For example what is the difference in reaction between "validation semantic" and "validation syntactic"? IMO in both cases the some kind of specification or configuration has to be changed.

Also "resource" error is mostly used for file permission and access stuff. So I thought the reaction to a "resource" error is "check that the file exists and the permissions are correct". 

But then I saw this one (ignoring the malloc errors that will be changed):
https://github.com/ElektraInitiative/libelektra/blob/b3a5c75c11c05635ef90f9e1e20fff9a8aa854f9/src/plugins/jni/jni.c#L88-L94

No file or memory allocation involved here... I am not even sure what is going in this code, but I am fairly certain that checking some file for the correct permissions won't solve a JNI problem.

> Is it that common that you google for such specific elektra errors?

**I** don't but a user you encountered the error might try that and good error codes should be useful too users as well as developers. Otherwise why print them in non-verbose mode.

Piankero:
I added my guidelines which I had in mind when creating the categories along with Markus:
https://github.com/ElektraInitiative/libelektra/pull/2589/commits/2cfd3ed91d8d2a924f10123eb788ce90c726e79f

> But then I saw this one (ignoring the malloc errors that will be changed):

You are right that this error is wrongly categorized. It belongs to an installation problem and not to a resource one. Usually Resource errors also indicate missing resources where a Key obviously is no resource but actually belong to Elektra's internals.

> For example what is the difference in reaction between "validation semantic" and "validation syntactic"? IMO in both cases the some kind of specification or configuration has to be changed.

They differ in the kind of reaction the application should do. Either try it with a different format (syntactic) or try a completely different value if the given one does not conform (semantic)

> I don't but a user you encountered the error might try that and good error codes should be useful too users as well as developers. Otherwise why print them in non-verbose mode.

I see your point. I think that a good error message though is more useful for googling rather than a simple code.

Please take a short look into the error-categorization.md.
It is very had to get *all* kind of errors right in the first place.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+Applications cannot handle such errors themselves other than reporting it to Elektra and hope
+that the developers investigate and fix the issue.
+
+### Conflict ("02000")
+
+`Conflict Errors` are errors which indicate temporary problems which can be handled by handled by
+updating the current state. Examples are the need for calling `kdbGet` before `kdbSet`, or pulling changes
+from git before pushing them. These kind of errors are usually in resolver plugins when the state of the file
+has changed without the system knowing. Applications interacting with such an error usually know
+how to handle them once approached such as the `highlevel API`.
```
kodebach: > such as the `highlevel API`.

This is not correct. The high-level API could only deal so easily with the old error code 30, because that was  a very specific error and the high-level API provides very specific circumstances. Most applications can only **attempt** a merge, there is no guarantee there won't be merge conflicts. Please mention that.

Also, if we categorise everything that fits the above description as `C02000`, the high-level API won't be able to handle this error anymore. (Doesn't really matter, but we should mention it here in that case).

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+`Logical Errors` indicate a bug in Elektra such as internal errors, assertion failures or errors
+which should not happen such as going into a `default` branch when you are assured that all cases
+are covered. Usually such errors come with a message to report such failures to Elektra's bugtracker.
+Applications cannot handle such errors themselves other than reporting it to Elektra and hope
+that the developers investigate and fix the issue.
+
+### Conflict ("02000")
+
+`Conflict Errors` are errors which indicate temporary problems which can be handled by handled by
+updating the current state. Examples are the need for calling `kdbGet` before `kdbSet`, or pulling changes
```
kodebach: What is "the current state"? A configuration file is technically also part of the current state, but that obviously doesn't fit this case.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+
+If not it will most likely fit into an existing category. Categories are leaf based, so you cannot
+put an error into a node such as `Permanent errors` (see below).
+
+## Error categorization Guideline
+
+Now we will investigate each category in more detail and when to put an error/warning in there.
+
+The current structure looks like this:
+- Permanent errors ("01000")
```
kodebach: Should use the C prefix, i.e. "C01000"

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+  - Installation ("01300")
+  - Logical ("01400")
+- Conflict ("02000")
+- Timeout ("03000")
+- Validation ("04000")
+  - Syntactic ("04100")
+  - Semantic ("04200")
+### Permanent errors ("01000")
+
+The main category `Permanent Errors` refer to such errors which cannot be fixed by retry
```
kodebach: Timeouts are the only ones that might be fixed by retrying without any changes.

Resource errors can definitely be fixed by retrying (after changing the permissions).

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+
+`Resource Errors` are all kinds of errors which are either permission related, existence related
+or missing resources such as memory, RAM, etc.
+Examples are missing files, insufficient permissions or out of memory.
+This category forces users or applications to provide additional resources, create a missing file/ directory etc
+or provide the relevant permissions for Elektra to work.
+
+#### Parsing ("01200")
+
+`Parsing Errors` are reserved for parsers for xml/yaml/ini etc. and indicate errors which
```
kodebach: What is the difference between a parsing and a validation syntactic error?

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+at all. `Permanent Errors` are subdivided into
+- Resource
+- Parsing
+- Installation
+- Logical
+#### Resource ("01100")
+
+`Resource Errors` are all kinds of errors which are either permission related, existence related
+or missing resources such as memory, RAM, etc.
+Examples are missing files, insufficient permissions or out of memory.
```
kodebach: Memory allocation errors are very different and should not be resource errors. They require very special behaviour. Once a malloc fails, you basically cannot call malloc again or you might just again end up where you started (with a memory allocation error).

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+were investigated during parsing. Such errors show unexpected encounters like missing line endings,
+illegal characters or wrong encodings.
+Users or applications will have to sanitize the input and retry again.
+
+#### Installation ("01300")
+
+`Installation Errors` are those errors which are related to Elektra's installation such as
+wrong plugin names, missing backends, initialization errors, misconfiguration of Elektra etc.
+Users will have to reconfigure, reinstall, recompile (with other settings) Elektra in order to
+get rid of this error. Applications will most likely not be able to handle such errors by themselves.
```
kodebach: A missing backend could easily be added by most applications. Especially, if a specification is used and even more so, if the `specload` plugin is used.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### Logical ("01400")
+
+`Logical Errors` indicate a bug in Elektra such as internal errors, assertion failures or errors
```
kodebach: I think this should be two separate categories.

One for errors actually internal to Elektra like assertion failures (although assertion use `abort(3)` right now I think). These are the ones that come with a message to report the error to our bug tracker. 

And another one for logical interface errors. E.g. when an application passes a NULL pointer to `kdbGet`. In this case reporting the error to Elektra's bug tracker doesn't make sense, but an application might want to add a line telling user to notify the developer about the bug.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Timeout ("03000")
+
+`Timeout Errors` are easy to categorize as the reaction simply suggests to wait for a short period and retry again.
+Examples are lost connections to a server.
+
+### Validation ("04000")
+
+`Validation Errors` are heavily used for Elektra's `configuration specification` feature and
+should tell users that their given input does not match a certain pattern/ type/ expected semantic/ etc.
```
kodebach:
```suggestion
should tell users that their given input does not match a certain pattern / type / expected semantic etc.
```

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+
+Validation errors can either be syntactic or semantic.
+
+#### Syntactic ("04100")
+
+`Syntactic Errors` are errors which tell users or applications that the current format is not valid.
+Examples are wrong date formats or invalid notations such as polish prefix notation. Also path related errors
+like missing slashes come into this category. Applications should try a different format and retry pushing
```
kodebach: > Applications should try a different format and retry

You would definitely have to parse the error message to do any kind of automatic handling of this kind of error. So I think you should write "Users" instead of "Applications".

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+it to Elektra.
+
+#### Semantic ("04200")
+
+`Semantic Errors` are a bit more tricky to solve programmatically. Examples are if users provide a `String`
+where Elektra expects an `Integer` or are given server port is already in use in which the specification says
+it has to be available. Applications should try a different value and retry again.
```
kodebach: same as above, automatic changes are impractical here

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+#### Syntactic ("04100")
+
+`Syntactic Errors` are errors which tell users or applications that the current format is not valid.
+Examples are wrong date formats or invalid notations such as polish prefix notation. Also path related errors
+like missing slashes come into this category. Applications should try a different format and retry pushing
+it to Elektra.
+
+#### Semantic ("04200")
+
+`Semantic Errors` are a bit more tricky to solve programmatically. Examples are if users provide a `String`
```
kodebach: I think you need to clarify the string/integer example. It is only a semantic problem, if a key is marked with `type = string`, when an application expected the type to be e.g. `long`. If however the type is correctly set, and the value just isn't a valid integer, it is technically a syntax problem.

And if a port is already in use, it is clearly a resource problem.

I think a good semantic error example is, if the `reference` plugin detects that a referenced key doesn't exist. Syntactically everything is fine the plugin can read everything it wanted to read. But semantically there is a problem. Setting a reference implies that the key should exist, but it doesn't.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+
+`Validation Errors` are heavily used for Elektra's `configuration specification` feature and
+should tell users that their given input does not match a certain pattern/ type/ expected semantic/ etc.
+
+Validation errors can either be syntactic or semantic.
+
+#### Syntactic ("04100")
+
+`Syntactic Errors` are errors which tell users or applications that the current format is not valid.
+Examples are wrong date formats or invalid notations such as polish prefix notation. Also path related errors
```
kodebach: > or invalid notations such as polish prefix notation

That seems like you want to say "polish prefix notation is never syntactically correct". I think it is much more understandable example to say: "or a missing closing bracket `]` inside of a regex".

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Categorization Mindset
+
+Errors along with their unique code only exist because they can be **reacted
+differently on programmatically**. Imagine a method call which returns a `Timeout` error. Naturally
+a senseful reaction would be to retry at a later point in time. So
+the reaction here would be to retry with a time based approach. On the other hand
+it does not make sense to differentiate between `No Write Permission` and `No Read Permission` as
+the application just knows that it simply cannot access the desired resource and tells
+the user to grant it. One could argue that an application could always react differently
```
kodebach: > One could argue that an application could always react differently

You provide a counter argument to your own categorisation method, before you even introduced the method... At least move the question above this explanation.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,111 @@
...
+based on the error (eg. give yourself write or read permission) but such detailed granularity
+is rarely needed and would come with a lot of maintenance effort. In the previous versions of Elektra
+we were facing many duplicated errors because too many existed and developers simply cannot know each
+and every code.
+
+So if you feel the need for a new category you should first ask yourself the following question:
```
kodebach: If I already feel the need for a new category, I looked at the descriptions below  and decided my error doesn't fit any of these categories. Your question will not help me decide into which of the categories I should put my error. So my only way out is either creating a new category nonetheless, or somewhat randomly choose a category that fits a bit.

doc/dev/error-categorization.md
Piankero: They are very similar, thats correct. Parsing though has more focus on Parsers where syntactic validation has more focus on specification validation. I think though we could merge those categories into the syntactic error.

doc/dev/error-categorization.md
Piankero: @markus2330 requested this category and I am also not 100% sure. I assume that it is an error where you try to modify another state but see that it differs from what you expect (git push, file changed externally, kdb get before kdb set) .. does this make more sense to you?

doc/dev/error-categorization.md
Piankero: What are your proposed changes here for the categorization? Should we introduce a missing backend Error or should I reword the `Installation Error` here?

doc/dev/error-categorization.md
Piankero: Should I write "cannot be fixed by a simple retry"? Or what is your suggested solution here?

doc/dev/error-categorization.md
Piankero: What is your suggested way of finding an existing category or create a new one?

Piankero:
I reworked the error categories a bit. This would be the new categorization:
- Permanent errors
  - Resource (New subcategory)
    - Memory Allocation
    - General Resource (Permission, Not-Found)
  - ~~Parsing~~ (Merged into Syntactic Validation)
  - Installation
  - Logical (New subcategory)
    - Assertion ("Please report to Elektra" bugs)
    - Interface (Provide additional message to user to notify dev)
- Conflict
- Timeout
- Validation
  - Syntactic
  - Semantic

Please let us not discuss endlessly about these because after merging it is still very easy to change categories

doc/dev/error-categorization.md
kodebach: If you just remove "Applications will most likely not be able to handle such errors by themselves." it would be fine. Of course then you don't really categorise by reaction anymore.

Creating a separate category just for missing backend seems to miss the point of reducing the granularity. And the name "Installation error" definitely fits for a missing backend as backends should be created during installation.

doc/dev/error-categorization.md
kodebach: That still doesn't address the fact that "conflict" and "validation" errors also cannot be fixed by a "simple retry".

doc/dev/error-categorization.md
kodebach: I would say "an error where the current state is incompatible with the attempted operation".

The incompatibility is what causes the problem. Whether this incompatibility is temporary or not or what needs to change shouldn't really matter I think.

Piankero:
Thank you @sanssecours !
I have a bit of troubles concerning out-of-range errors on where to categorize them. As of now I put it into its own category.

sanssecours:
> I have a bit of troubles concerning out-of-range errors on where to categorize them. As of now I put it into its own category.

That makes sense. Looks like [C++ also uses a top-level category for those kind of errors](https://en.cppreference.com/w/cpp/error) (see header “Exception categories”). C++ also seems to differentiate between `range_error` and `out_of_range`, which seems too fine grained for Elektra in my opinion.

kodebach:
@Piankero Not sure, why Github renders the comments above without context. Maybe because they are not on the latest commit.

You can view all the comments with context here: https://github.com/ElektraInitiative/libelektra/commit/058d194c21efeeceed183d8f4a2724fa6d58a128

markus2330:
> Please let us not discuss endlessly about these because after merging it is still very easy to change categories

I think there would be much less discussions if there were very clear descriptions of what the categories actually are and clear guidelines when to use what.

Piankero:
>I think there would be much less discussions if there were very clear descriptions of what the categories actually are and clear guidelines when to use what.

Many of these categories were actually suggested by you without any clear description at all. I just tried to grasp the concept you were thinking of. You did not even answer on some follow up [questions](https://github.com/ElektraInitiative/libelektra/pull/2435#discussion_r269700432)

markus2330:
> Many of these categories were actually suggested by you without any clear description at all.

Yes, to help you. I did not say much you could not also have found out by yourself by looking at the code.

> I just tried to grasp the concept you were thinking of.

Yes, this is good. You need to really understand what the errors are about, how to trigger them, how users can react on it, ... Your explanation of the errors and the guidelines are the most important documents so that we will all get a common understanding.

> You did not even answer on some follow up questions

It was not on purpose, there are so many discussion threads with so many questions. It would be good if you separate the PRs (the .gitignore, the guidelines, and the design decisions could be in different PRs.)

Besides that: It is very clear what the original "conflict error" does. There are many tests for it [0] and the code that issues the error is not so complicated. Nevertheless,  I answered it now.

[0] https://github.com/ElektraInitiative/libelektra/blob/115d932222013e1feacb29fcf86fc7bc1cb566c2/tests/kdb/testkdb_conflict.cpp#L138-L150

[1] https://github.com/ElektraInitiative/libelektra/blob/115d932222013e1feacb29fcf86fc7bc1cb566c2/src/plugins/resolver/resolver.c#L847-L849

Piankero:
@markus2330 @kodebach @sanssecours : Thank you very much vor your valuable input.
I updated the guideline documentation and would be glad to have another check if this is now senseful to all

Piankero:
@sanssecours : i added the out of range errors in https://github.com/ElektraInitiative/libelektra/pull/2589/commits/46a903fbabbc88db0c9175031870b8ff213b7d69

Are there any other errors like this you know of?

sanssecours:
>  i added the out of range errors in 46a903f

Thank you.

> Are there any other errors like this you know of?

Nope. I never did much work concerning Elektra’s error concept or error categorization in general, though.

Piankero:
@e1528532 can you help me to fix the compile errors (and test errors) for your haskell plugin? The errors are now strings instead of integers and I tried to change it in the Typechecker.hs:96 but now receive the following error message from the build server: 
``` • No instance for (Data.String.IsString Int)

        arising from the literal ‘"C04200"’

    • In the first argument of ‘triggerError’, namely ‘"C04200"’

      In the expression: triggerError "C04200" k errStr

      In a case alternative: Set -> triggerError "C04200" k errStr
```

Piankero:
@markus2330 Do you know the reason for the build server failure?
https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2589/49/pipeline

markus2330:
Seems like GitHub did not respond. Simply restart in such situations.

markus2330:

Piankero:
Closing due to splitup in #2686 and #2682

PR #2599, opened 2019-04-08
YAML CPP: Convert Boolean Values

sanssecours:


src/plugins/yamlcpp/read.cpp
```diff
@@ -96,7 +96,14 @@ Key createLeafKey (YAML::Node const & node, string const & name)
 	Key key{ name, KEY_BINARY, KEY_END };
 	if (!node.IsNull ())
 	{
-		key.setString (node.as<string> ());
+		try
+		{
+			key.set<bool> (node.as<bool> ());
+		}
+		catch (YAML::BadConversion const &)
```
markus2330: This might be very slow with many booleans. Isn't there an API to ask if it is an boolean without that an exception is thrown?

markus2330:
Thank you so much for extending the tutorial!

src/plugins/yamlcpp/read.cpp
sanssecours: > Isn't there an API to ask if it is an boolean without that an exception is thrown?

None that I know of. At least the author of the library writes that this is [a valid option on how to react to different types of scalars](https://stackoverflow.com/questions/19994312/obtain-type-of-value-stored-in-yamlnode-for-yaml-cpp).

Anyway, my local copy of YAML CPP now compares the string value of the node to `true` and `false`, which should work too.

markus2330: >At least the author of the library

Seems like performance is not a priority.

>now compares the string value of the node to true and false, which should work too.

Are you sure? `"true"` and `true` is something different, isn't it?

sanssecours: > Are you sure? "true" and true is something different, isn't it?

Yep, I forgot to consider quoted scalars. In that case I do not know a method to check for a boolean without using exceptions 😢.

markus2330: If the author of the lib recommended this approach (without giving alternatives!) then he/she must live with a performance comparison where this approach is used.

PR #2614, opened 2019-04-10
WIP: Units plugin

goJoWe16:
**Pull Request for Midterm submission of the Homework for CM 2019 SS. Work in Progress.**

For now, only the correctness of given values + prefixes + SI-units is checked. Prefixes (such as *k* or *µ*) are not transformed until now. Furthermore, the units and prefixes are stored as they are (e.g. value + prefix + unit is stored). This is also intended to be changed in further steps, such that only the value is stored, the prefix is considered to transform the value to it's SI-Base unit and the SI-Unit is discarded.

Regarding issue #1398:
I am not sure if this is the intended behaviour of this plugin, as the correctness of SI-Base Units will be ignored in this case. For example, *1m* would be correct when checking against the unit without a prefix (m interpreted as meter). In the example of issue #1398 , *1m* could be interpreted as 1 milli.

@markus2330 could you give me some explanation of how you would expect to distinguish between the to possibilities in this case?
As far as i envisioned this plugin, a value with an optional prefix and a mandatory SI-unit has to be entered.

## Basics

- [X] Short descriptions should be in the release notes (added as entry in

## Checklist

need to be checked.

- [ ] I added unit tests
- [X] meta data is updated (e.g. README.md of plugins and doc/METADATA.md)

## Review


kodebach:
Take a look at the example in #1398:

> ```
> [key]
> type = long
> unit/base/#0 = decimal
> unit/base/#1 = hexadecimal
> unit/notation/#0 = metric
> unit/notation/#1 = e
> unit/measurement = s
>```

The names aren't that important (but I think they work), but the basic idea is. The plugin would use the metakey `unit/measurement` to set which Unit to use. It also doesn't have to be SI, no reason to not allow fun units like the [Barn-megaparsec](https://en.wikipedia.org/wiki/List_of_humorous_units_of_measurement#Barn-megaparsec) (bMpc). So `1m` is 1 meter if `unit/measurement = m` and 0.001 if `unit/measurement` is unset (or empty), otherwise it is an error.

I would also be nice to have some way of configuring how the values are converted. The default would be normalizing to the base unit (so `km` to `m` etc.). This might however require using floating point numbers, which could introduces errors. I think it would be good to have a second mode, where the plugin converts to the prefix given in e.g. `unit/scale`. This way you can control the precision of the floating point value. You could also add an option to always produce an integer and just ignore the decimal part, so that the `type` plugin can validate correctly.
(Note: it should be possible to convert the number *without* actually converting it from a string to an integer type, you just need to shift the decimal point)

For the whole procedure of normalizing to a number without a unit and restoring back to number + unit, you can look at the `type` plugin (especially the boolean and enum parts).

The other parts of  #1398 (different bases, different notations) are nice to have, but its probably best, if you only implement the actual unit and SI-prefix parsing. You should however keep in mind, that we also want to implement the other stuff in future and write your code accordingly. So while using regex might be nice for the units and prefixes, I think it would get very complicated once the other stuff comes in (although it would probably work).

src/plugins/units/README.md
```diff
@@ -0,0 +1,56 @@
...
+#> Check the /tests/units key for validity
+
+kdb set /tests/units "1m"
+#> Suceeds, since the value is a valid decimal with SI-unit representation.
+
+kdb set /tests/units "35453.327468 µH"
+#> Suceeds, since the value is a valid decimal with SI-unit representation.
+
+kdb set /tests/units "32"
+#> Throws an error: value of key is not a representation for a value with a SI-unit
```
markus2330: Which error? (See shell recorder tests/shell/shell_recorder/tutorial_wrapper/README.md  for how to specify errors)

src/plugins/units/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+kdb set /tests/units "12 n"
+#> Throws an error: value of key is not a representation for a value with a SI-unit
+#> This could be changed in a further step (see issue #1398) to be valid, as for example the user wants to enter 1k to represent 1000 without an SI-Unit. 
+#> For now, this is ignored as the entered units should be checked for correctness
+
+kdb set /tests/units 432 km
+#> Throws an error: quotes are expected.
```
markus2330: expected -> needed because of the shell

src/error/specification
```diff
@@ -1345,3 +1345,9 @@ description:the contract was malformed
 severity:error
 ingroup:kdb
 macro:MALFORMED_CONTRACT
+
+number:214
+description:value of key is not a valid SI-unit
+severity:error
+ingroup:plugin
+module:units
```
markus2330: add newline

doc/news/_preparation_next_release.md
```diff
@@ -235,6 +239,11 @@ The following section lists news about the [modules](https://www.libelektra.org/
 
 - [cache](https://www.libelektra.org/plugins/cache) is a new global caching plugin. It uses [mmapstorage](https://www.libelektra.org/plugins/mmapstorage) as its storage backend and lazily stores keysets from previous ´kdbGet()´ calls. We added initial support for the default resolver and multifile resolver. _(Mihael Pranjić)_
 
+### units
```
markus2330: twice?

src/plugins/units/units.c
```diff
@@ -0,0 +1,88 @@
...
+#include <string.h>
+#include <stdlib.h>
+
+int is_valid_key(Key * key, Key * parentKey)
+{
+	const Key * meta = keyGetMeta(key, "check/units");
+	if (!meta) return 1;
+	const char* value = keyString(key);
+	const char * prefix = "(Y|Z|E|P|T|G|M|k|h|da|d|c|m|µ|n|p|f|a|z|y)";
+    const char * unit = "?(m|g|s|A|K|mol|cd|rad|sr|Hz|N|Pa|J|W|C|V|F|Ω|Ohm|S|Wb|T|H|C|lm|lx|Bq|Gy|Sv|kat|l|L)";
```
markus2330: formatting broken

src/plugins/units/units.c
```diff
@@ -0,0 +1,88 @@
...
+
+#include <kdberrors.h>
+#include "units.h"
+#include <regex.h>
+#include <kdbhelper.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+
+int is_valid_key(Key * key, Key * parentKey)
```
markus2330: can be static

src/plugins/units/units.c
```diff
@@ -0,0 +1,88 @@
...
+	const char* value = keyString(key);
+	const char * prefix = "(Y|Z|E|P|T|G|M|k|h|da|d|c|m|µ|n|p|f|a|z|y)";
+    const char * unit = "?(m|g|s|A|K|mol|cd|rad|sr|Hz|N|Pa|J|W|C|V|F|Ω|Ohm|S|Wb|T|H|C|lm|lx|Bq|Gy|Sv|kat|l|L)";
+    char regexString[256];
+    snprintf(regexString, sizeof regexString, "%s%s%s%s%s", "^([0-9]+)\\.?([0-9]*)\\s*", prefix, "?", unit, "$");
+	
+	regex_t regex;
+	regmatch_t offsets;
+	int compile_failure = regcomp (&regex, regexString, REG_NOSUB | REG_EXTENDED | REG_NEWLINE);
+	if (compile_failure) return -1;
```
markus2330: No error set when the compilation fails.

markus2330:
@goJoWe16 any progress here? This is quite similar to #2744

PR #2617, opened 2019-04-10
New java unique values plugin

dmoisej:
It is template for the new Java plugin, which saves unique values to the mounted endpoint.


markus2330:
I do not understand what "saves unique values to the mounted endpoint" mean. Can you please elaborate on the functionality?

dmoisej:
Let's imagine, that we have for instance some mount point like: "user/test".

The purpose is, that under user/test there can be no such key-value and key1-value1 pairs, that 
key != key1 and value=value1. 

I wanted to look more into the JNI right now but I will try to implement for this plugin something in the nearest time.

markus2330:
Ok, so this should be a validation plugin that makes values constant (unique is misleading). This makes sense and currently does not exist.

I suppose it could work like this:
```
kdb mount a.cfg /tests const
kdb setmeta /tests/a default 80
kdb setmeta /tests/a const    # make constant
kdb get /tests/a   # always returns 80
#> 80
kdb set /tests/a 80    # set always fails
```

dmoisej:
Thank you very much for help. I understood the purpose and will be implementing this const functionality.

kodebach:
If I understood both of your comments correctly, I think you described two completely different, but equally valid plugins.

@markus2330 You described a plugin that creates constants. This is a good idea, but probably should be part of the spec plugin.
@dmoisej Described a plugin that ensures no two different keys contain the same value. This would  be useful for arrays that are used with set semantics.

markus2330:
Yes, you are right within arrays the uniqueness property might be useful. @kodebach Thank you for pointing that out.

@dmoisej Did you have a concrete scenario where you needed this?

If not, I would recommend to implement a plugin that implements the `restrict/*` properties as already specified in doc/METADATA.ini (const is actually equivalent to restrict/write, so we should use this as metadata and not introduce a new one. I updated the example above.). Btw. the `restrict/*` properties are actually already implemented in the Web-UI (it is not only an idea), so the documentation of them is not up-to-date.

PR #2618, opened 2019-04-10
YAML CPP: Improve Support for Sparse Key Sets

sanssecours:


markus2330:
Thank you very much for being so attentive to all semantical details!

PR #2621, opened 2019-04-10
WIP: Plugin for validating MAC-Addresses

e01306821:
WIP for CM2019-homework.

This plugin validates MAC-addresses. The following MAC-address-formats are allowed for now:
- XX-XX-XX-XX-XX-XX
- XX:XX:XX:XX:XX:XX
- XXXXXX-XXXXXX

If "set" is used, the value will be transformed to the format of the middle one.

TODO:
- Write tests
- Update readme
- Better error-messages
- Maybe also support formats "XXXXXXXXXXXX" and "XXXX.XXXX.XXXX"


kodebach:
You should also transform the MAC-address in `kdbGet` and in this case probably also restore to the original format in `kdbSet`. You can look at the `boolean` and `enum` parts of the `type` plugin, to get an idea of how such a behaviour can be implemented.

It would also be nice, if the plugin could transform the MAC-address into a decimal number. Then the high-level API could more easily be used to read the value (as a 64 bit integer).

e01306821:
Thanks for your review!
@kodebach You mean transform the address in `kdbGet` instead of `kdbSet`?
Or should I rather return a 64 bit integer on `kdbGet` instead of returning a "standardized" MAC-string? If yes just so that I am understanding it correctly, using `kdbGet` shall return an integer-MAC instead of a "conventional" MAC-address (using my supported formats as checks if the read address is valid) and `kdbSet` shall check and transform an integer-value back to the original format if there was any (if not, then one of my supported formats will be written)?

kodebach:
Probably  @markus2330 should decide what the standard format for MAC addresses will be. MAC addresses are 48 bits long, so a 64 bit integer could store them. Which could then be retrieved using `elektraGetUnsignedLongLong`.

The transformation should be done in `kdbGet`. In addition you store the original value in the `origvalue` metakey. Then in `kdbSet` you transform any key that is marked as a MAC address (`check/macaddr` set), but not those with `origvalue`. If the key has `origvalue`, it wasn't changed (ensured by the framework) so you simply set the key to the original value and remove `origvalue`. The keys that were transformed are immediately restored to there original value as well (you can also just do the format check and skip the transformation/restore if possible).

The point of doing this is that the user always sees their preferred representation (since we always restore) and the program using Elektra always sees the standard form.

One additional note: Right now only one plugin can normalize a keys value, so if for some reason `origvalue` is already set, you shouldn't change it or the keys value. Instead you should raise an error in this case, noting the multiple plugins tried to normalize the key.

markus2330:
Yes, 48bit in a 64bit integer sounds fine.

markus2330:
This pull request **introduces 1 alert** when merging 42eb9cb2ab2fe7d2abca971d3189d98e9a752200 into 27b31c2c69819d739fc301853d45e6ca75f8e149 - [view on LGTM.com](https://lgtm.com/projects/g/ElektraInitiative/libelektra/rev/pr-683cfe4ace873ec10d75ea5fc0a9f915ac1e6594)

**new alerts:**

* 1 for Returning stack\-allocated memory

---

markus2330:
@kodebach there are some bugs in functionality (even simple kdb get/set do not work). Can you help @e01306821?

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,271 @@
...
+		if (origValue)
+		{
+			keySetString (cur, keyString (origValue));
+			return ELEKTRA_PLUGIN_STATUS_SUCCESS;
+		}
+		keySetMeta (cur, "origvalue", keyString (cur));
+
+		if (checkIntMac (keyString (cur)))
+		{
+			transformMac (cur);
```
kodebach: You shouldn't transform in `kdbSet`. If `origvalue` is present, restore the value. Then validate the value (whether restored or not) and store it. This ensures that we always use the format that the user chose when setting a value.

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,271 @@
...
+	for (int i = 2; i <= 14; i += 3)
+	{
+		mac[i] = ':';
+	}
+	mac[17] = '\0';
+}
+
+void transformMac (Key * key)
+{
+	const char * macKey = keyString (key);
```
kodebach: I am not sure I understand what this function does, but from the way it is used, I inferred that you are trying to transform the mac address from the supported formats into a standard format.

My suggestion would be to do something like this (since you already used a regex to validate the string):
```
output = malloc a 12 character string
outputIndex = 0

foreach char c in input:
    if c is a separator:
        skip
    else:
        output[outputIndex] = c
        outputIndex++
```

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,271 @@
...
+	}
+}
+
+void transformToInt (Key * key)
+{
+	const char * mac = keyString (key);
+
+	char * macReduced = elektraMalloc (12);
+
+	removeStandardizedSeparator (mac, macReduced);
```
kodebach: If you remove the separators in transfromMac you don't need to here (see above)

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,271 @@
...
+			return ELEKTRA_PLUGIN_STATUS_ERROR;
+		}
+		int rc = validateMac (cur);
+		if (rc)
+		{
+			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_INVALID_FORMAT, parentKey, "%s is not in a supported format.", keyString (cur));
+			return ELEKTRA_PLUGIN_STATUS_ERROR;
+		}
+
+		if (checkIntMac (keyString (cur)))
```
kodebach: You already call `checkIntMac` inside `validateMac`, if you use a different return code, for the case where an integer mac was given, you won't have to call it again here.

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,271 @@
...
+		{
+			transformMac (cur);
+			transformToInt (cur);
```
kodebach: since you don't need the separate in `transfromMac` in `kdbSet` (see below), you can merge these to functions

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,271 @@
...
+			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_INVALID_FORMAT, parentKey, "%s is not in a supported format.", keyString (cur));
+			return ELEKTRA_PLUGIN_STATUS_ERROR;
+		}
+		const Key * origValue = keyGetMeta (cur, "origvalue");
+		if (origValue)
+		{
+			keySetString (cur, keyString (origValue));
+			return ELEKTRA_PLUGIN_STATUS_SUCCESS;
+		}
+		keySetMeta (cur, "origvalue", keyString (cur));
```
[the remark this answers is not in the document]
e01306821: After restoring the original value of the key this function already returns. Therefore `origvalue` gets only set if `origvalue` was not set before. Or shouldn't it be set here at all?

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,271 @@
...
+
+	if (reg == REG_NOMATCH)
+	{
+		return 1;
+	}
+	else
+	{
+		return 0;
+	}
+	// return reg == REG_NOMATCH ? 1 : 0;
```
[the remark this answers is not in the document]
e01306821: I once planned to add code to one of the if-branches. So yes, since it didn't happen these branches do not make much sense.

e01306821:
Thanks for your review!

Running a manual test locally seems to work now. I've created a shell script containing:
```
./kdb mount macconf.ecf /tests/mac macaddr
./kdb setmeta /tests/mac/mac1 check/macaddr ""
./kdb setmeta /tests/mac/mac2 check/macaddr ""
./kdb setmeta /tests/mac/mac3 check/macaddr ""
./kdb setmeta /tests/mac/mac4 check/macaddr ""
./kdb set /tests/mac/mac1 00:A0:C9:14:C8:29
./kdb set /tests/mac/mac2 00-A0-C9-14-C8-29
./kdb set /tests/mac/mac3 00A0C9-14C829
./kdb set /tests/mac/mac4 17661175009296
./kdb set /tests/mac/mac1 00:G1:C9:14:C8:29
./kdb set /tests/mac/mac1 00:E1:C914:C8:29
./kdb set /tests/mac/mac4 281474976710656
./kdb get /tests/mac/mac1
./kdb get /tests/mac/mac2
./kdb get /tests/mac/mac3
./kdb get /tests/mac/mac4
./kdb umount /tests/mac
./kdb rmmeta spec/tests/mac/mac1 check/macaddr
./kdb rmmeta spec/tests/mac/mac2 check/macaddr
./kdb rmmeta spec/tests/mac/mac3 check/macaddr
./kdb rmmeta spec/tests/mac/mac4 check/macaddr
```
Executing it inside my build-folder returns values I expect for each command. But the shellrecorder test containing the exact same commands is not successful. Does anyone have an idea what the problem might be here?

kodebach:
> Executing it inside my build-folder returns values I expect for each command. But the shellrecorder test containing the exact same commands is not successful. Does anyone have an idea what the problem might be here?

Try using the `user` namespace for all keys (i.e. `user/tests/mac` instead of `/tests/mac`), there are certain problems with cascading keys in shellrecorder tests.

e01306821:
Thanks for your suggestion. Unfortunately it didn't solve the problem locally. I'll try again tomorrow (or so)

markus2330:
> Unfortunately it didn't solve the problem locally.

Please what you did, what you expect and what happened so that we can help you.

e01306821:
Oh, yes, sry.

I've changed all key namespaces to `user` in such a way to resemble the data initialization part of the README.md from the plugin `mathcheck`.
It now looks like this:
```
# Backup-and-Restore: user/tests/mac

# Mount `macaddr` plugin
sudo kdb mount macconf.ecf user/tests/mac macaddr

# Check the validity of the following MAC addresses
kdb setmeta user/tests/mac/mac1 check/macaddr ""
kdb setmeta user/tests/mac/mac2 check/macaddr ""
kdb setmeta user/tests/mac/mac3 check/macaddr ""
kdb setmeta user/tests/mac/mac4 check/macaddr ""
...
```

The shell recorder test fails already at `kdb setmeta user/tests/mac/mac1 check/macaddr ""` with error code 5:
```
kdb setmeta user/tests/mac/mac1 check/macaddr ""

ERROR - RET:
Return value “5” does not match “0”
```

Since running all commands of my test using a script works (they return all what I expect as stated in my README-file) I suspect that I have to initialize all keys another way to make the test also work. I am currently thinking about dropping the shell recorder tests and instead write only testmod-tests but I did not have sufficient time yet, hopefully tomorrow or at the weekend.

Thanks!

markus2330:
Thank you!

Because you rewrote to user/ namespaces you also need to do other changes: you first need to write valid mac addresses, and later on set the meta data. Otherwise your plugin will try to check if an empty key is a valid MAC address, which it is not.

@kodebach Can you help rewriting the shell recorder tests?

markus2330:
There is a memory problem in transformMac (macaddr.c:46) see https://build.libelektra.org/jenkins/job/libelektra/job/PR-2621/23/artifact/artifacts/debian-stable-full/DynamicAnalysis.xml

Is macWithoutSeparators properly terminated?

e01306821:
This PR went already successful through all checks once (commit: 8834219) and the last commit just removed unused code and removed/added lines, so it should still pass all tests.
So if you are satisfied with the functionality and code of this plugin, this PR can be merged.

src/plugins/macaddr/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Introduction
+
+This plugin validates MAC-addresses. The following MAC-address-formats are supported:
+
+    XX-XX-XX-XX-XX-XX
+    XX:XX:XX:XX:XX:XX
+    XXXXXX-XXXXXX
+
+If `set` is used, the value will be transformed to the format of the middle one.
```
markus2330: Is this still true? I thought you used a decimal number now?

src/plugins/macaddr/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+`kdbGet` returns an integer-representation of these values.
+
+TODO:
```
markus2330: Please resolve these TODOs.

doc/METADATA.ini
```diff
@@ -959,6 +959,12 @@ type= long
 status= proposal
 description= Add the error but do not fail (to test robustness of framework).
 
+[check/macaddr]
+status= implemented
+type= empty
+usedby/plugins= macaddr
+description = Marks a key as MAC address
```
markus2330: very short description

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,210 @@
...
+	if (errno == EINVAL || errno == ERANGE || *endptr != '\0') return VALIDATION_ERROR;
+	if (ret > MAXMACINT) return VALIDATION_ERROR;
+
+	return VALIDATION_ISINT;
+}
+
+/**
+ * Checks if the supplied MAC key is a valid MAC address
+ * @param key the key containing the MAC address
+ * @return VALIDATION_SUCCESS if MAC address is valid, VALIDATION_ISINT if MAC address is a 64 bit integer and valid, else VALIDATION_ERROR
```
markus2330: You could use retval here.

markus2330:
@kodebach: do you still request changes?

markus2330:
Thank you for checking again!

@e01306821 could you please correct these issues?

src/plugins/macaddr/README.md
e01306821: You are right, this isn't true anymore. But I also do not transform anything anymore on `set`, just validating. `get` now transforms a given address to an integer representation.

src/plugins/macaddr/macaddr.c
```diff
@@ -160,6 +160,7 @@ int elektraMacaddrGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned, Key *
 
 		if (rc != VALIDATION_ISINT)
 		{
+			keySetMeta (cur, "origvalue", keyString (cur));
```
kodebach: You need to set `origvalue` after calling `transformMac`, because `transformMac` calls `keySetString` which resets `origvalue`.

To detect this problem in future, see the comments in `testmod_macaddr.c`

src/plugins/macaddr/testmod_macaddr.c
```diff
@@ -0,0 +1,158 @@
...
+	ksDel (testKs);
+}
+
+static void testAddressesSetGet (const char * keyValue, unsigned long long longValue)
+{
+	char intChar[21];
+	KeySet * testKs = ksNew (10, keyNew ("user/tests/mac/addr", KEY_VALUE, keyValue, KEY_META, META, "", KEY_END), KS_END);
+	setKey (testKs);
+	convertLong (intChar, longValue);
+	succeed_if (!strcmp (getKeyString (testKs, "user/tests/mac/addr"), intChar), "error");
```
kodebach: Please check here that `origvalue` was set to `keyValue`.

src/plugins/macaddr/testmod_macaddr.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+static void testAll (void)
+{
+	testAddressesStandardColons ();
+	testAddressesStandardHyphens ();
+	testAddressesSingleHyphen ();
+	testAddressesNumber ();
+	testAddressesReturn ();
+}
```
kodebach: You should also add a test case that calls both `kdbGet` and `kdbSet` and checks that after `kdbSet` the value is correctly restored to the value before `kdbGet`.

e01306821:
Thanks! I hope my last changes is what you've meant.

src/plugins/macaddr/macaddr.c
```diff
@@ -0,0 +1,210 @@
...
+		{
+			ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_INVALID_FORMAT, parentKey, "%s is not in a supported format.", keyString (cur));
+			return ELEKTRA_PLUGIN_STATUS_ERROR;
+		}
+
+		if (rc != VALIDATION_ISINT)
+		{
+			const char * origvalue = elektraStrDup (keyString (cur));
+			transformMac (cur);
+			keySetMeta (cur, "origvalue", origvalue);
```
kodebach:
```suggestion
			keySetMeta (cur, "origvalue", origvalue);
			elektraFree (origvalue);
```

e01306821: Oh, yes, thanks again!

e01306821:
I think this PR is now ready to merge

markus2330:
Great work! Thank you very much!

PR #2637, opened 2019-04-16
Logger: Fix Minor Problems

sanssecours:
After this update the logger

- adds debug messages to the system log, if you enable the CMake variable `DEBUG`
- also logs messages at the specified minimum log level
- does not truncate relative file paths incorrectly anymore

.


markus2330:
Thank you very much!

PR #2648, opened 2019-04-26
Style: Improve cmake format Install Command

sanssecours:


markus2330:
Thank you so much for the maintenance work!

PR #2649, opened 2019-05-01
YAML Smith: Fix Conversion of Keys with Common Prefix

sanssecours:


markus2330:
Thank you for the fix! Doesn't it make sense that our YAML output and prettier settings are consistent?

sanssecours:
> Doesn't it make sense that our YAML output and prettier settings are consistent?

That would make sense. However, currently YAML Smith is rather experimental and does not produce pretty output. There are even situations where the plugin does not produce valid output at all, since it does not support escaping yet.

PR #2650, opened 2019-05-01
YAML CPP: Improve Support for Mixed Data

sanssecours:


src/plugins/yamlcpp/read.cpp
```diff
@@ -21,6 +20,25 @@ using namespace kdb;
...
 namespace
 {
+/**
+ * @brief This function converts a given number to an array base name.
+ *
+ * @param index This number specifies the index of the array entry.
+ *
+ * @return A string representing the given indices as Elektra array name.
+ */
+string indexToArrayBaseName (uintmax_t const index)
```
markus2330: We should avoid to have too many of these implementations. Can we maybe move this to the cpp binding?

markus2330:
Thank you very much!

src/plugins/yamlcpp/read.cpp
sanssecours: > We should avoid to have too many of these implementations.

I totally agree.

> Can we maybe move this to the cpp binding?

That would make sense. I think the right place to implement this functionality might even be [`libease`](https://github.com/ElektraInitiative/libelektra/blob/master/src/libs/ease/array.c).

PR #2659, opened 2019-05-02
Website: Fix Convert Engine Test

sanssecours:


sanssecours:
jenkins build homepage please

markus2330:
Thank you so much for fixing the homepage!

Btw. I think for readers of the release notes it is more interesting that JSON now round-trips (and before it did not). Where exactly which test case was fixed is imho not of interest.

PR #2662, opened 2019-05-02
Yan LR: Fix Conversion of Empty Values

sanssecours:


src/plugins/yanlr/listener.cpp
```diff
@@ -132,6 +132,7 @@ void KeyListener::enterPair (PairContext * context)
 	{
 		// Add key with empty value
 		// The parser does not visit `exitValue` in that case
+		child.setBinary (NULL, 0);
```
markus2330: Maybe we even add a convenience function setNull in the C++ binding?

markus2330:
Thank you, very clean and beautiful PR, as always.

src/plugins/yanlr/listener.cpp
sanssecours: I do not know, if it makes that much sense to add a new method, since we might not use it that often. On the other hand, I would certainly have used the method `setNull`, if it would exist.

PR #2667, opened 2019-05-03
Benchmark: Extend YAML Benchmark Script

sanssecours:
I also added a generated file, created via

```sh
scripts/generate_yaml.py > benchmarks/data/generated.yaml
```

to this pull request. While this file is relatively large (about 500kB) I think it makes sense to add it, since this will improve the reproducibility of the YAML plugin benchmark.


markus2330:
Please use https://github.com/ElektraInitiative/rawdata to add data that you need for reproducibility. You can also add your result data there.

markus2330:
Thank you very much for this update!

PR #2682, opened 2019-05-09
Error categorization guidelines & design decision splitup

Piankero:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

## Merging

### DIFF
--- a/doc/dev/error-categorization.md
+++ b/doc/dev/error-categorization.md
@@ -1,0 +1,162 @@
+# Error Categorization
+
+In this document we start to explain how to categorize errors.
+Many errors are unique and can be difficult to categorize so this
+document will act as a guideline.
+
+## Categorization Mindset
+
+Errors along with their unique code only primarily exist because they can be **reacted
+differently on programmatically**. One could argue that an application could always react differently
+based on the error (eg. give yourself write or read permission) but such detailed granularity
+is rarely needed and would come with a lot of maintenance effort. In the previous versions of
+Elektra we were facing many duplicated errors because too many existed
+and developers simply cannot know each and every code.
+
+To get an idea of programmatic reactions take a method call which returns a `Timeout` error. Naturally
+a senseful reaction would be to retry at a later point in time. So
+the reaction here would be to retry with a time based approach. On the other hand
+it does not make sense to differentiate between `No Write Permission` and `No Read Permission` as
+the application just knows that it simply cannot access the desired resource and tells
+the user to grant it.
+
+Categories are leaf based, so you cannot put an error
+into a node (branch) such as `Permanent errors` (see below). If you feel for a new category,
+please forge a design decision document and make a PR to Elektra's repo.
+
+## Error categorization Guideline
+
+Now we will investigate each category in more detail and when to put an error/warning in there.
+
+The current structure looks like this:
+
+- Permanent errors
+  - Resource
+    - Memory Allocation
+    - General Resource
+  - Installation
+  - Logical
+    - Assertion
+    - Interface
+    - Broken plugin
+- Conflicting State
+- Timeout
+- Out-Of-Range
+- Validation
+  - Syntactic
+  - Semantic
+
+### Permanent errors ("C01000")
+
+The branch category `Permanent Errors` refer to such errors which cannot be fixed by retry
+at all. `Permanent Errors` are subdivided into
+
+- Resource
+- Installation
+- Logical
+
+#### Resource ("C01100")
+
+`Resource Errors` as a branch category are all kinds of errors which are either permission related, existence related
+or missing resources such as memory, RAM, etc.
+Examples are missing files, insufficient permissions or out of memory.
+This category forces users or applications to provide additional resources, create a missing file/ directory etc
+or provide the relevant permissions for Elektra to work.
+
+##### Memory Allocation ("C01110")
+
+`Memory Allocation Errors` are errors which come from failed `malloc` calls primarily as no
+more memory could be allocated for the application. Such errors will gain special handling
+in future releases.
+
+##### General Resource ("C01120")
+
+`General Resource Errors` are all kind of permission and existence errors. Examples are
+missing files/ directories or insufficient permission to execute certain commands (eg. you
+would require sudo permissions). Reactions are fixing the permissions or creating the file/directory
+and retry the operation.
+
+#### Installation ("C01300")
+
+`Installation Errors` are those errors which are related to a wrong installation such as
+wrong plugin names, missing backends, initialization errors, misconfiguration of Elektra etc.
+Installation errors might also be non-Elektra specific but also from dependent library/applications
+such as gpg.
+Also plugin configuration errors belong to `Installation Errors` as they this happens during
+mounting.
+Users will have to reconfigure, reinstall, recompile (with other settings) Elektra in order to
+get rid of this error or fix the installation of the corresponding library/application.
+
+#### Logical ("C01410")
+
+`Logical Errors` is another branch category in which you indicate a bug in Elektra
+such as internal errors, assertion failures or errors
+which should not happen such as going into a `default` branch when you are assured that all cases
+are covered. Usually such errors come with a message to report such failures to Elektra's bugtracker.
+Applications cannot handle such errors themselves.
+
+##### Assertion ("C01410")
+
+`Assertion Errors` are such errors which indicate a flaw in the internal logic such as going into a `default`
+branch which you do never expect to happen.
+This category might be used in the future to automatically
+issue a bug to our bugtracker. As of now if you have to use this error please add a message
+indicating that this bug should be reported.
+
+##### Interface ("C01420")
+
+`Interface Errors` errors indicate a wrong usage of Elektra's API. An example would be to pass a NULL pointer to
+`kdbGet`. Also violations of the backend belong into this category. Compared to semantic validation errors,
+this category has its focus on detecting wrong usages of the API instead of a "retry with a different value" approach.
+
+##### Broken Plugin ("C01430")
+
+`Broken Plugin Errors` errors indicate that a plugin does not behave in an intended way. Unrecognized commands,
+unkown return codes, plugin creation errors, etc. belong to this category. Also uncaught exceptions belong here because
+Elektra expects all exceptions to be caught.
+
+### Conflicting State ("C02000")
+
+`Conflicting State Errors` are errors where the current state is incompatible with the attempted operation.
+Examples are the need for calling `kdbGet` before `kdbSet`. Another example would be to try to push you changes
+into a git repository where the remote branch has already changed.
+These kind of errors are usually in resolver plugins when the state of the file
+has changed without the system knowing. Such errors demand a special synchronization action before retrying.
+
+### Timeout ("C03000")
+
+`Timeout Errors` are easy to categorize as the reaction simply suggests to wait for a short period and retry again.
+Examples are lost connections to a server.
+
+### Validation ("C04000")
+
+`Validation Errors` are heavily used for Elektra's `configuration specification` feature and
+should tell users that their given input does not match a certain pattern/type/expected semantic etc.
+
+Validation errors can either be syntactic or semantic.
+
+#### Syntactic ("C04100")
+
+`Syntactic Errors` are errors which tell users or applications that the current format is not valid.
+Examples are wrong date formats or missing closing brackets `]` inside of a regular expression. Also path related errors
+like missing slashes come into this category. Also parsing errors fall under syntactic errors such as
+unexpected encounters like missing line endings, illegal characters or wrong encodings. Also transformation
+and conversion errors are to be categorizes here because the format of the given input does not allow such
+actions. Users should try a different format and retry pushing it to Elektra.
+
+#### Semantic ("C04200")
+
+`Semantic Errors` are a bit more tricky to solve programmatically. Examples are references to
+non-existent keys (`reference` plugin), setting values to keys which require to be empty (`required` plugin),
+wrong provided values in a specification if you restricted the values (`enum` plugin), etc.
+
+The provided value differs from the developers/ administrators (specification writer) intention,
+even though syntactically everything is valid. Users should try a different value and retry again.
+
+### Out of Range ("C05000")
+
+`Out of Range Errors` are operations which attempt to be past a valid range. Examples are trying to set
+an Elektra Array entry to `-1` or trying to set a port outside of a valid range. There is quite some overlap
+in semantic/syntactic/resource errors but these errors are primarily focused on Elektra arrays. It's use is for callers
+who are iterating (or continually incrementing/ decrementing) through a space and can easily detect when they are
+done by looking for this error.
--- a/doc/dev/error-categorization.md
+++ b/doc/dev/error-categorization.md
@@ -15,8 +15,10 @@ make sense to differentiate between `No Write Permission` and `No Read Permissio
 as the application just knows that it simply cannot access the desired resource
 and tells the user to grant it.
 
-Categories are leaf based, so you cannot put an error into a node (branch) such
-as `Permanent errors` (see below).
+Categories are hierarchically structured. In some categories you cannot put an error such
+as `Permanent errors` (see below) because they are too general and developers should choose
+a more specific category. Please choose the most specific category as possible when trying
+to assign an error to a category.
 
 If you feel for a new category, please forge a design decision document and make
 a PR to Elektra's repo.
--- a/doc/dev/error-categorization.md
+++ b/doc/dev/error-categorization.md
@@ -1,37 +1,38 @@
 # Error Categorization
 
-This document explains how to categorize errors and should act as a form of guideline.
-The categorization aims to reduce
-duplicated errors and the maintenance effort for errors.
+This document explains how to categorize errors and should act as a form of
+guideline. The categorization aims to reduce duplicated errors and the
+maintenance effort for errors.
 
 ## Categorization Mindset
 
-Errors along with their unique code only primarily exist because they can be **reacted
-differently on programmatically**.
-To get an idea of programmatic reactions take a method call which returns a `Timeout` error. Naturally
-a senseful reaction would be to retry at a later point in time. So
-the reaction here would be to retry with a time based approach. On the other hand
-it does not make sense to differentiate between `No Write Permission` and `No Read Permission` as
-the application just knows that it simply cannot access the desired resource and tells
-the user to grant it.
+Errors along with their unique code only primarily exist because they can be
+**reacted differently on programmatically**. To get an idea of programmatic
+reactions take a method call which returns a `Timeout` error. Naturally a
+senseful reaction would be to retry at a later point in time. So the reaction
+here would be to retry with a time based approach. On the other hand it does not
+make sense to differentiate between `No Write Permission` and `No Read Permission`
+as the application just knows that it simply cannot access the desired resource
+and tells the user to grant it.
 
-Categories are leaf based, so you cannot put an error
-into a node (branch) such as `Permanent errors` (see below).
+Categories are leaf based, so you cannot put an error into a node (branch) such
+as `Permanent errors` (see below).
 
-If you feel for a new category,
-please forge a design decision document and make a PR to Elektra's repo.
+If you feel for a new category, please forge a design decision document and make
+a PR to Elektra's repo.
 
 ## Error categorization Guideline
 
-Now we will investigate each category in more detail and when to put an error/warning in there.
+Now we will investigate each category in more detail and when to put an
+error/warning in there.
 
-For a complete structural overview please visit the corresponding
-[Design decision Document](../decisions/error_codes.md)
+For a complete structural overview please visit the corresponding [design
+decision document](../decisions/error_codes.md)
 
 ### Permanent errors ("C01000")
 
-The branch category `Permanent Errors` refer to such errors which cannot be fixed by retry
-at all. `Permanent Errors` are subdivided into
+The branch category `Permanent Errors` refer to such errors which cannot be
+fixed by retry at all. `Permanent Errors` are subdivided into
 
 - Resource
 - Installation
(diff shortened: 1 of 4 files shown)


doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+- Installation
+- Logical
+
+#### Resource ("C01100")
+
+`Resource Errors` as a branch category are all kinds of errors which are either permission related, existence related
+or missing resources such as memory, RAM, etc.
+Examples are missing files, insufficient permissions or out of memory.
+This category forces users or applications to provide additional resources, create a missing file/ directory etc
+or provide the relevant permissions for Elektra to work.
```
markus2330:
```suggestion
or to provide the relevant permissions.
```

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+
+##### Memory Allocation ("C01110")
+
+`Memory Allocation Errors` are errors which come from failed `malloc` calls primarily as no
```
markus2330:
```suggestion
`Memory Allocation Errors` are errors which come from failed `elektraMalloc` calls primarily as no
```

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Installation ("C01200")
+
+`Installation Errors` are those errors which are related to a wrong installation such as
+wrong plugin names, missing backends, initialization errors, misconfiguration of Elektra etc.
+Installation errors might also be non-Elektra specific but also from dependent library/applications
+such as gpg.
+Also plugin configuration errors belong to `Installation Errors` as they this happens during
+mounting.
+Users will have to reconfigure, reinstall, recompile (with other settings) Elektra in order to
```
markus2330: Very good explanation! Actually you can make this structure everywhere within this document:

1. describe which errors belong the the category
2. which errors do not belong there
3. comparison with other categories
4. examples of this error
5. what users can do about it

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+get rid of this error or fix the installation of the corresponding library/application.
+
+#### Logical ("C01300")
+
+`Logical Errors` is another branch category in which you indicate a bug in Elektra
```
markus2330: It is the only branch (up to now) which indicates a bug in Elektra.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+such as internal errors, assertion failures or errors
```
markus2330: not-implemented features

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+which should not happen such as going into a `default` branch when you are assured that all cases
+are covered. Usually such errors come with a message to report such failures to Elektra's bugtracker.
+Applications cannot handle such errors themselves.
+
+##### Assertion ("C01310")
+
+`Assertion Errors` are such errors which indicate a flaw in the internal logic such as going into a `default`
+branch which you do never expect to happen.
+This category might be used in the future to automatically
```
markus2330: We only describe here how it is, not what might happen in the future.

This category actually could be a sub-category of Logical.

But for now, we do not need this category, unless someone rewrites the macro ELEKTRA_ASSERT to add an error.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+
+##### Interface ("C01320")
+
+`Interface Errors` errors indicate a wrong usage of Elektra's API. An example would be to pass a NULL pointer to
+`kdbGet`. Also violations of the backend belong into this category. Compared to semantic validation errors,
+this category has its focus on detecting wrong usages of the API instead of a "retry with a different value" approach.
+
+##### Plugin is Broken ("C01330")
+
+`Broken Plugin Errors` errors indicate that a plugin does not behave in an intended way. Unrecognized commands,
```
markus2330: I do not like the name. What about "Plugin Interface"?

Or even "Interface" with two subcategories:

- API
- Plugin

It should also be made more clear what to do with "Version" errors.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+unkown return codes, plugin creation errors, etc. belong to this category. Also uncaught exceptions belong here because
+Elektra expects all exceptions to be caught.
+
+### Conflicting State ("C02000")
+
+`Conflicting State Errors` are errors where the current state is incompatible with the attempted operation.
```
markus2330: Explanation could be improved.

It should start with an explanation like: "Elektra has opportunistic locking...."

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+Examples are the need for calling `kdbGet` before `kdbSet`. Another example would be to try to push you changes
+into a git repository where the remote branch has already changed.
+These kind of errors are usually in resolver plugins when the state of the file
+has changed without the system knowing. Such errors demand a special synchronization action before retrying.
+
+### Timeout ("C03000")
+
+`Timeout Errors` are easy to categorize as the reaction simply suggests to wait for a short period and retry again.
```
markus2330: All errors are easy to categorize when you know the category. (Please rewrite according to structure above)

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+
+`Validation Errors` are heavily used for Elektra's `configuration specification` feature and
+should tell users that their given input does not match a certain pattern/type/expected semantic etc.
+
+Validation errors can either be syntactic or semantic.
+
+#### Syntactic ("C04100")
+
+`Syntactic Errors` are errors which tell users or applications that the current format is not valid.
+Examples are wrong date formats or missing closing brackets `]` inside of a regular expression. Also path related errors
```
markus2330: If the regular expression is checked or if the regular expression is used?

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+like missing slashes come into this category. Also parsing errors fall under syntactic errors such as
+unexpected encounters like missing line endings, illegal characters or wrong encodings. Also transformation
+and conversion errors are to be categorizes here because the format of the given input does not allow such
+actions. Users should try a different format and retry pushing it to Elektra.
```
markus2330:
```suggestion
actions. Users should try a different value and retry setting it with Elektra.
```

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Semantic ("C04200")
+
+`Semantic Errors` are a bit more tricky to solve programmatically. Examples are references to
```
markus2330: Isn't semantic validation now the same as resource errors?

You need a more careful distinction here.

I would say:

- C01100: an essential resource, needed for Elektra to operate is missing (e.g. cannot write to config file)
- C04200: a resource, which is marked essential by the configuration specification is missing (e.g. log file required by application not there)

The handling, however, is usually the same: provide the resource. But if the admin does not want to provide the resource, the handling is different:

- C01100: remount
- C04200: change spec of application OR retry with different value

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+non-existent keys (`reference` plugin), setting values to keys which require to be empty (`required` plugin),
+wrong provided values in a specification if you restricted the values (`enum` plugin), etc.
+
+The provided value differs from the developers/ administrators (specification writer) intention,
+even though syntactically everything is valid. Users should try a different value and retry again.
+
+### Out of Range ("C05000")
+
+`Out of Range Errors` are operations which attempt to be past a valid range. Examples are trying to set
+an Elektra Array entry to `-1` or trying to set a port outside of a valid range. There is quite some overlap
```
markus2330: "There is quite some overlap" then we should not have the error. Try to properly distinguish.

If it is only about Elektra's arrays, Validation/Out of Range would be better?

Or we create Validation/Structure for problems in the structure of the KeySet?

@sanssecours Can you maybe help out?

sanssecours: > Can you maybe help out?

Nope, sorry. I do not know anything about error categorization and I am also currently not interested in learning more about the topic.

markus2330: Then I think we should remove this category.

Validation/Structure might be useful to pursuit, though.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+in semantic/syntactic/resource errors but these errors are primarily focused on Elektra arrays. It's use is for callers
+who are iterating (or continually incrementing/ decrementing) through a space and can easily detect when they are
+done by looking for this error.
+
+## Related
```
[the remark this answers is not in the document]
Piankero: This is no "decision" in a narrow sense but a guideline and also therefore I purposefully deleted "Decision"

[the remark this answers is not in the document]
Piankero: Also it is under a different path

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,167 @@
...
+and developers simply cannot know each and every code.
+
+To get an idea of programmatic reactions take a method call which returns a `Timeout` error. Naturally
+a senseful reaction would be to retry at a later point in time. So
+the reaction here would be to retry with a time based approach. On the other hand
+it does not make sense to differentiate between `No Write Permission` and `No Read Permission` as
+the application just knows that it simply cannot access the desired resource and tells
+the user to grant it.
+
+Categories are leaf based, so you cannot put an error
```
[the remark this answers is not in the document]
Piankero: > Why having this restriction? This restriction forces us to introduce otherwise not needed categories such as "General Resource".

To avoid that users put it in very general categories

> And the even bigger problem is that it would disallow to add leaves to an already in-use category.

Why? Extending an existing non-leaf category is easy. Just splitting requires more attention. This makes sense in my opinion because if someone thinks of a special case which deserves special handling it deserves its own category and should be clearly distinguishable. New developers might just put it in a non-leaf category despite it deserves this special handling

doc/dev/error-categorization.md
Piankero: It is a subcategory of Logical. Where would you then put errors such as going into a default branch despite you expect this never to happen?

doc/dev/error-categorization.md
Piankero: I would suggest "Plugin misbehavior". Your suggested splitting would not be very intuitive for uncaught exceptions for example.

doc/dev/error-categorization.md
Piankero: Can you please help me in this regards? I have not that deep insight on how opportunistic locking in elektra works and why this can cause an error. Can you roughly explain it to me?

doc/dev/error-categorization.md
Piankero: I will merge it back to validation. It was initially here because some errors suggested by sanssecours were difficult to get right at first. This category would be a very clear and intuitive category despite some overlap. 

Such overlaps are btw now uncommon in other error frameworks too such as [gRPC](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md)

doc/dev/error-categorization.md
Piankero: Probably "Internal" would be a better naming for assertion errors as also used in other protocols like http, grpc, etc.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: Ok! "Related" by itself is, however, unclear.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: We could give some categories numbers (like "Resource") and others not (like "Validation). Then we could decide on case-per-case if the category is "too general to be used".

> Why?

Let us assume that we need a further category below "Timeout", e.g. "Resolvable By Automatic-Retry". Then we would only move the errors that belong there to this category. Others we would keep directly in "Timeout" (these errors do not need to be touched). I think you understand that going through errors is not something everyone wants to do.

doc/dev/error-categorization.md
markus2330: We do not need assertion errors (as our assertions do not set errors). Protocol errors are something else.

doc/dev/error-categorization.md
markus2330: Yes, Plugin misbehavior is okay. It is only too long for the macro.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
Piankero: > We could give some categories numbers (like "Resource") and others not (like "Validation). Then we could decide on case-per-case if the category is "too general to be used".

We could do that but then developers will use too broad error categories instead of good specific ones. It is really against the mindset of the decision. The branches are 
only used so that api users can react to whole categories of errors instead of specific errors. It also increases your effort to tell devs in every PR if they might want to use a better specific code.

> Others we would keep directly in "Timeout" (these errors do not need to be touched). I think you understand that going through all errors is not something everyone wants to do.

These errors still do not need to be touched because one could rename "Timeout" to another category and change the error code. It is a matter of seconds to do that. (eg. Timeout before: C03000 ... now General Timout: C03100 & "Resolved By Automatic-Retry" C03200)

doc/dev/error-categorization.md
Piankero: I will abbreviate the macro to "PLUGIN_MISBHV".

doc/dev/error-categorization.md
Piankero: Maybe the name assertion errors are misleading in this case then and "Internal error" is more fitting. If the category is not existent: where do you categorize errors such as [this](https://github.com/ElektraInitiative/libelektra/pull/2686/files#diff-839ad0c7bdbbc07424ca015b000030a8R1496) one?

doc/dev/error-categorization.md
Piankero: I tried to apply this form to all leaf categories even though 2 and 3 are very difficult to apply to all categories.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: > We could do that but then developers will use too broad error categories instead of good specific ones.

I fully agree here. You are welcome to prohibit to emit errors in some of the (very general) categories.

> These errors still do not need to be touched because one could rename "Timeout" to another category and change the error code.

No, we cannot do that. Because clients rely that they will get the same error category for the same error. If you design it well, it will be possible to make errors more specific. But changing the category, will always destroy the expectations of clients.

Please always remember that Elektra is a library and the *only* purpose is to make the life easier for clients (applications using Elektra).

Such breaking changes (like you do now with the redesign of the error concept) need to be rare exceptions. The concept now needs to be designed that this should never be needed again.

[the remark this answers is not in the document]
Piankero: > But changing the category, will always destroy the expectations of clients.

The design itself allows exactly for splitting a category without harming existing clients. Clients itself should check for the "branch" (e.g., .startsWith("C03")) and will always be safe no matter we split the categories they receive the same errors.
As in the example above changing e.g., C03000 to C03100 & C03200 (add C03200, change C03000 to C03100 and rename) will not break anything from clients, instead it gives them another ability to only catch more specific errors if they desire (.startsWith("C031")). The "rename and change category" approach I suggested is just a shortcut.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,173 @@
...
+misconfiguration of Elektra etc. Installation errors might also be non-Elektra
+specific but also from dependent library/applications such as gpg. Also plugin
+configuration errors belong to `Installation Errors` as this happens during
+mounting. Users will have to reconfigure, reinstall, recompile (with other
+settings) Elektra in order to get rid of this error or fix the installation of
+the corresponding library/application.
+
+#### Logical ("C01300")
+
+`Logical Errors` is a branch category in which you indicate a bug in Elektra
```
markus2330: "Interface" is exactly not a bug in Elektra (but wrong usage of Elektra).

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,173 @@
...
+`Installation Errors` are those errors which are related to a wrong installation
+such as wrong plugin names, missing backends, initialization errors,
+misconfiguration of Elektra etc. Installation errors might also be non-Elektra
+specific but also from dependent library/applications such as gpg. Also plugin
+configuration errors belong to `Installation Errors` as this happens during
+mounting. Users will have to reconfigure, reinstall, recompile (with other
+settings) Elektra in order to get rid of this error or fix the installation of
+the corresponding library/application.
+
+#### Logical ("C01300")
```
markus2330: Maybe the top-level category is called "Programming". And "Logical" is the subcategory of bugs in Elektra.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,173 @@
...
+which should not happen such as going into a `default` branch when you are
+assured that all cases are covered. Usually such errors come with a message to
+report such failures to Elektra's bugtracker. Applications cannot handle such
+errors themselves.
+
+##### Internal ("C01310")
+
+`Internal Errors` are such errors which indicate a flaw or bug in the internal
+logic of Elektra. Examples are going into a `default` branch which you do never
+expect to happen. This category might be used in the future to automatically
```
markus2330: We do not plan to automatically report issues.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,173 @@
...
+please use `Installation` errors. Users can try to remount, recompile the plugin
+under different options or use a different plugin (e.g., switching to `mini`
+instead of `ini`).
+
+### Conflicting State ("C02000")
+
+`Conflicting State Errors` are errors where the current state is incompatible
+with the attempted operation. These kind of errors are usually in resolver
+plugins when the state of the file has changed without the system knowing.
+Examples are the need for calling `kdbGet` before `kdbSet`. Another example
```
markus2330: `kdbGet` before `kdbSet` is not an example. It is what this error tells you to do.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,173 @@
...
+
+`Semantic Errors` are errors which indicate a misunderstanding of the intended
+meaning between a user's/developer's/administrator's way of seeing a setting and
+the application's one. The main focus of this category is to enforce
+specifications compared to other categories. So if users provide input which do
+not match prespecified criterias even though syntactically everything is valid,
+this category should be used. Examples are references to non-existent keys
+(`reference` plugin), setting values to keys which require to be empty
+(`required` plugin), wrong provided values in a specification if you restricted
+the values (`enum` plugin), etc.
```
markus2330: Where are structural validation errors now?

Btw. you deviate here from literature. In literate semantic errors are only errors which represent domain-specific knowledge. Enum would be syntactically.

For users it might be important to be able to distinguish:

- errors where value(s) needs to be fixed to be valid again
- errors where key(s) need to be added/removed to be valid again

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: Good to hear that you considered this already. Please also write a tutorial how to react to errors correctly. Btw. the text in the description of kdbSet also needs to be updated.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,173 @@
...
+
+##### Memory Allocation ("C01110")
+
+`Memory Allocation Errors` are errors which come from failed `elektraMalloc`
+calls primarily as no more memory could be allocated for the application. Errors
+with not enough hard disc space do not belong here but into `General Resource`.
+Such errors will gain special handling in future releases and users cannot deal
+with such errors as of now.
+
+##### General Resource ("C01120")
```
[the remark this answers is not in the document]
Piankero: Alright, since Elektra is your project and more or less belongs to you I will also incorporate branches as categories despite it is really against my desired option.

doc/dev/error-categorization.md
```diff
@@ -0,0 +1,173 @@
...
+#### Logical ("C01300")
+
+`Logical Errors` is a branch category in which you indicate a bug in Elektra
+such as internal errors, assertion failures, not implemented features or errors
+which should not happen such as going into a `default` branch when you are
+assured that all cases are covered. Usually such errors come with a message to
+report such failures to Elektra's bugtracker. Applications cannot handle such
+errors themselves.
+
+##### Internal ("C01310")
```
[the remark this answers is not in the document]
Piankero: Will be merged to `Logical` since branches are now allowed now

doc/dev/error-categorization.md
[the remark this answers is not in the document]
Piankero: What is the reason for this if it is already part of the guideline documentation?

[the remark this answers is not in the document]
markus2330: There are two different views: people developing Elektra (who set errors) and users (who receive errors). It seems like that the user view is essential to understand how the system will work (e.g. that extending errors works, all users are only allowed to compare with "startsWith").

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: We are searching for the best option, not for a personal opinion. See above: the user view was not clear to me, thus I did not see how the extension will work.

If you find a good name for "General Resource", I'll happily accept your proposal.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: See above. And we need a distinction between errors that are bugs in core, in plugins and API use.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
Piankero: Unfortunately there are many good options and there is no clear "best" option. So it boils down to some sort of personal opinion in the end. You want selectable branch categories because it is sometimes easier to categorize and name whereas I want leafs only because I assume that developers will be too lazy to put errors correctly (it honestly is a boring task). Both has its pros and cons and you insisted multiple times on branch categories so it is ok for me.

> If you find a good name for "General Resource", I'll happily accept your proposal.

No because there is no self explaining name for "Non-memory allocation failed errors but resource related errors" as far as I know.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
Piankero: Why do we suddenly need the distinction? The reaction does not differ.

The categorization was more or less clear and stable as I changed it very often due to the previous PR and I was only expecting minor changes but suddenly you want to completely restructure everything again? The `assertion` errors by the way were initially suggested by you and made sense but suddenly you want them to be removed? This is getting very confusing as this proceeds. Please let us not reinvent the wheel twice. The categorization will not be perfect, I honestly expect it to take several month to years to be near to perfect as we also do not have notable experience on what users of elektra expect of errors (or developers) and it will change in future versions no matter how often I rejig this.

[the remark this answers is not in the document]
Piankero: > Maybe we find a better name than "Internal". Errors by design are never internal (if we can handle them internally, we would never create an error. See above for suggestion.).

I honestly find the name internal absolutely correct because `Assertion` or `Internal` errors are such errors which you simply do not expect but have to deal with. As written in 2 other posts: default branches which should not be taken are such examples. You cannot "handle" these errors internally because you are sure they do not happen and if they do you still cannot say why and what to do.

Also the programmatic reaction is clear because it clearly indicates a bug where the source was not the user but a misimplemented feature.

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: > Both have its pros and cons and you insisted multiple times on branch categories so it is ok for me.

For such situations we should write design decisions. The 3rd way is that there is some way to specify which categories can have errors and which are too generic (It seems like "Resource" seems to be the only branch that should contain errors).

> No because there is no self explaining name for "Non-memory allocation failed errors but resource related errors" as far as I know.

Yes, that is why I thought would make sense to directly have errors in "Resource". Because the "General Resource" is actually "Resource" \ "Memory Errors" (\ as "set difference").

doc/dev/error-categorization.md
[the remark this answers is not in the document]
markus2330: > Why do we suddenly need the distinction? The reaction does not differ.

I would not suggest it if I saw no difference. For bugs in plugins, the user can simply mount another plugin which behaves better, for bugs in the core this is not possible.

But I agree it is a subtle difference (in particular: what is about default cases in plugins).

> The categorization was more or less clear and stable as I changed it very often due to the previous PR and I was only expecting minor changes but suddenly you want to completely restructure everything again?

I do not understand where you see a "complete restructure" if I only see the need of clarification of two categories (resource and logic).

> also do not have notable experience on what users of elektra expect of errors (or developers)

That is why this process takes so long. It is *essential* that you get expert on your topic.

> and it will change in future versions no matter how often I rejig this

Of course there might be some need to extend it later. But as we saw in PostgreSQL: in well designed error categories this is possible without breaking anything.

> [the name internal errors]

Ok, then internal it is.

Piankero:
> Did you go over all errors? Is it now clear for you where each error should be?

Yes, I also migrated some errors already and found some inconsistencies.
There are still though some errors which I am unsure as they are usually related to encryption and are very specific why a certain error occurred. You can take a look [here](https://github.com/ElektraInitiative/libelektra/pull/2686/commits/6a79f5bbcb6773f417d1b9876c1b5c434bc2dfe6)

Interestingly after looking through all errors, there are no TIMEOUT errors anymore. I could remove them but I think the category is still worth having because the reaction is unique and not unlikely to be used in the future.

If there are still wrong categorized errors I will most likely have had a different understanding of the code compared to the author.

doc/dev/error-categorization.md
Piankero: > Btw. you deviate here from literature. In literate semantic errors are only errors which represent domain-specific knowledge. Enum would be syntactically.

I get your point! But are enum not also some sort of "domain-specific knowledge"? Eg. if you have a setting with `red`, `green`, `blue` as background setting and a user puts in `yellow`? The domain itself though is limited to these three colors.

doc/dev/error-categorization.md
Piankero: I thought about that but I do not think that "Programming" is a good self explanatory name. Also a introduction for another branch category does not yield a considerable benefit. I rewrote the logical section to not include "bugs" but only the subcategory. This also makes much more sense now.

markus2330:
>    Yes, I also migrated some errors already and found some inconsistencies.

Thank you, great you found them now.

> There are still though some errors which I am unsure as they are usually related to encryption and are very specific why a certain error occurred. You can take a look here

It says "We went looking everywhere, but couldn’t find those commits." maybe you rebased? I think its best that you make a new PR with *only* the errors where you are unsure.

> Interestingly after looking through all errors, there are no TIMEOUT errors anymore. I could remove them but I think the category is still worth having because the reaction is unique and not unlikely to be used in the future.

I never liked this category, so I would be very happy if we can get rid of it.

> If there are still wrong categorized errors I will most likely have had a different understanding of the code compared to the author.

It would be interesting (for your work) how likely such errors actually occur. Then we could spend more time on improving errors that users actually see. (And also have test cases specifically for them.) I

From my experience following errors are very often:

- permission errors (unable to write to config files), in particular this often happens with mounting
- validation errors (in particular I use types, regex and ranges very often)
- sometimes I see the installation error that some essential lib is missing (usually about that dlopen of the resolver failed)


You could also dig through issues and/or open a new issue and ask all Elektra developers which errors they see often.

Piankero:
@markus2330 https://github.com/Piankero/libelektra/commit/8fb0e4aa0b045f6a269798ea8c9e2fea7010ac2f

markus2330:
> @markus2330 Piankero@8fb0e4a

Thank you! But I think we should avoid to commenting commits. It is better to have a PR (with only these errors, TODO is then not needed).

Piankero:
> Thank you! But I think we should avoid to commenting commits. It is better to have a PR (with only these errors, TODO is then not needed).

They are not that much anymore. And I could only PR to my own branch which would not be visible here.

markus2330:
You can easily pick some commits to some other branch (cherry-pick) and make a PR out of this branch.

markus2330:
Btw. here is a nice guideline on how to choose the correct error code (for http): https://www.codetinkerer.com/2015/12/04/choosing-an-http-status-code.html

Maybe you can come up with similar graphs for our guideline?

Piankero:
> Btw. here is a nice guideline on how to choose the correct error code (for http): [codetinkerer.com/2015/12/04/choosing-an-http-status-code.html](https://www.codetinkerer.com/2015/12/04/choosing-an-http-status-code.html)
> 
> Maybe you can come up with similar graphs for our guideline?

Thank you for the link. I like this kind of visualization too. I would prefer to do such visual categorization help after the merge of https://github.com/ElektraInitiative/libelektra/pull/2686 because if there still are some changes it is very time consuming to update the graphics.

Piankero:
> It would be interesting (for your work) how likely such errors actually occur. Then we could spend more time on improving errors that users actually see. (And also have test cases specifically for them.) I
> 
> From my experience following errors are very often:
> 
> * permission errors (unable to write to config files), in particular this often happens with mounting
> * validation errors (in particular I use types, regex and ranges very often)
> * conflict error (that another process wrote something in between)
> * sometimes I see the installation error that some essential lib is missing (usually about that dlopen of the resolver failed)
> 
> You could also dig through issues and/or open a new issue and ask all Elektra developers which errors they see often.

Thank you for this suggestion and your first experience feedback. I will consider this in my thesis and might also ask my colleagues at work (ops).

markus2330:
Thank you!

> also ask my colleagues at work (ops)

Excellent idea. (Maybe as structured interview? This you could add to your thesis.)

doc/dev/error-categorization.md
markus2330: I do not see a clear separation from enum to regex or ranges. (As a matter of fact they can substitute each other in many cases.)

Issue #99, opened 2014-08-26
keyIsInactive does not work for inactive directories


fberlakovich:
Despite the current testcases, `keyIsInactive` should consider a key below an inactive key as inactive too (as mentioned in #9). Example: `user/.hidden/key` should be inactive, but currently is not.

Issue #10, opened 2014-07-11
VC++ support broken


markus2330:
Try to build whole Elektra with C++ (even the C files) to support VC++ (that does not support C99)

PR #2685, opened 2019-05-09
error: warning array concept added

Piankero:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

## Merging

### DIFF
--- a/doc/decisions/warning-array.md
+++ b/doc/decisions/warning-array.md
@@ -6,12 +6,7 @@ Currently multiple warnings are saved in an elektra non-conforming array
 notation which is limited to 100 entries. The notation of `#00` is against
 the design [decision made](array.md).
 
-Furthermore if the number of warnings is exceeded, the earliest entries are overwritten
-which might yield the most useful warning messages.
-
 ## Constraints
-
-- The number of entries should stay limited but may be more than 100 entries.
 
 ## Assumptions
 
@@ -20,14 +15,15 @@ which might yield the most useful warning messages.
 ## Decision
 
 The format should be aligned with the correct array notation,
-starting with `#0`. The maximum number of warnings will be expanded to
-1001 entries (`#0` - `#1000`).
-
-If more warnings occur, warning `#1000` is overwritten each time.
+starting with `#0`. The maximum number of warnings will stay at
+100 entries (`#0` - `#_99`).
 
 ## Rationale
 
 ## Implications
+
+To keep the ordering we add an underscore `_` once we go to
+beyond 10 warnings (`#_10`).
 
 ## Related Decisions
 
--- a/doc/decisions/warning-array.md
+++ b/doc/decisions/warning-array.md
@@ -1,0 +1,36 @@
+# Arrays
+
+## Problem
+
+Currently multiple warnings are saved in an elektra non-conforming array
+notation which is limited to 100 entries. The notation of `#00` is against
+the design [decision made](array.md).
+
+Furthermore if the number of warnings is exceeded, the earliest entries are overwritten
+which might yield the most useful warning messages.
+
+## Constraints
+
+- The number of entries should stay limited but may be more than 100 entries.
+
+## Assumptions
+
+## Considered Alternatives
+
+## Decision
+
+The format should be aligned with the correct array notation,
+starting with `#0`. The maximum number of warnings will be expanded to
+1001 entries (`#0` - `#1000`).
+
+If more warnings occur, warning `#1000` is overwritten each time.
+
+## Rationale
+
+## Implications
+
+## Related Decisions
+
+- [Array](array.md)
+
+## Notes
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -40,6 +40,7 @@ section here.
 - [Vendor Spec](vendor_spec.md)
 - [Deferred Plugin Calls](deferred_plugin_calls.md)
 - [Error codes](error_codes.md)
+- [Array for Warnings](warning-array.md)
 
 ## Drafts
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -118,7 +118,7 @@ you up to date with the multi-language support provided by Elektra.
 
 ## Documentation
 
-- <<TODO>>
+- Added decision for array concept of warnings. _(Michael Zronek)_
 - <<TODO>>
 - <<TODO>>
 


doc/decisions/warning-array.md
```diff
@@ -0,0 +1,36 @@
+# Arrays
+
+## Problem
+
+Currently multiple warnings are saved in an elektra non-conforming array
+notation which is limited to 100 entries. The notation of `#00` is against
+the design [decision made](array.md).
```
markus2330:
```suggestion
a [design decision](array.md).
```

doc/decisions/warning-array.md
```diff
@@ -0,0 +1,36 @@
...
+
+Furthermore if the number of warnings is exceeded, the earliest entries are overwritten
```
markus2330: This is not a problem?

doc/decisions/warning-array.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Assumptions
+
+## Considered Alternatives
+
+## Decision
+
+The format should be aligned with the correct array notation,
+starting with `#0`. The maximum number of warnings will be expanded to
+1001 entries (`#0` - `#1000`).
```
markus2330: Actually, there is no reason to change the current value of 100, nobody complained that it is too low. It only should not be hard-coded  that it must be 100.

```suggestion
1001 entries (`#0` - `#___1000`).
```

doc/decisions/warning-array.md
```diff
@@ -0,0 +1,36 @@
...
+
+If more warnings occur, warning `#1000` is overwritten each time.
```
markus2330: This might be a problem as one kdbGet/kdbSet sometimes yield more than one warning. Isn't it better to drop old warnings?

doc/decisions/warning-array.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Rationale
+
+## Implications
```
markus2330: Missing.

doc/decisions/warning-array.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Decision
+
+The format should be aligned with the correct array notation,
+starting with `#0`. The maximum number of warnings will be expanded to
+1001 entries (`#0` - `#1000`).
+
+If more warnings occur, warning `#1000` is overwritten each time.
+
+## Rationale
```
markus2330: Missing. (To confirm to the array notation.)

Is there any reason why it should be 1001?

doc/decisions/warning-array.md
Piankero: We discussed this with @kodebach in an elektra meeting that this is a problem.

doc/decisions/warning-array.md
Piankero: We agreed in the elektra meeting to hardcap the limit in case of recursive functions which issue warnings and never end.

doc/decisions/warning-array.md
Piankero: 1001 is just because it is more natural (to me) to run from 0-1000 instead of 0-999. I have no problem to limit it somewhere else

doc/decisions/warning-array.md
markus2330: The problem are situations like this:
```
#01: the newest warning
#02: the oldest warning
...
#99:  the second newest warning
```

So that new/old warnings might get mixed up.

There are several ways how to avoid it, one way is to simply continue counting up and start dropping the warnings with the lowest index.

I remember that @kodebach said that we should also add a warning when we started doing something like this.

doc/decisions/warning-array.md
markus2330: Yes, we agreed on that. But not on a concrete number.

If you give any reason for why it should be 1000, it might be ok.

doc/decisions/warning-array.md
Piankero: How can this happen?
What do you mean by "dropping the warnings with the lowest index"?
Should it be possible to have only entries from `#3` to `#99` for example because 0-2 is dropped?

doc/decisions/warning-array.md
Piankero: Actually I remember that we agreed on the number 1000 but any other number is fine too. But I think we need a hard-cap somewhere to avoid filling up the whole system with warning messages if someone accidentally introduces an endless loop.

doc/decisions/warning-array.md
markus2330: Exactly. (`#99` would be `#_99`, though)

kodebach: As far as I remember we agreed on this:

Handle the first `n-1` warnings as normal. Instead of the `n`-th warning we issue a special warning that the limit has been reached and ignore the actual warning. From then on all warnings are dropped. Additionally we could simply log all warnings above the limit, so that in case of debugging the warnings are not completely gone.

The reason for this behaviour is that often the first warning created is the most important to fix the problem. Later warnings may be caused by earlier ones, if the earlier warning is overwritten it becomes hard to diagnose.

Piankero: > Exactly. (#99 would be #_99, though)

Why exactly the underscore? This is really confusing.
I am also confused why we should only partially drop entries. What I have seen in the code is that you iterate over an elektra array and also save the current array number under `warnings/number`.

> Instead of the n-th warning

What exactly is `n` here? @markus2330 does not want a hard cap but this obviously requires one.

markus2330: > Additionally we could simply log all warnings above the limit, so that in case of debugging the warnings are not completely gone.

In debugging you usually only have short sessions (not many kdbGet in a long-running loop) thus it is rather unlikely you will reach the limit.

> The reason for this behaviour is that often the first warning created is the most important to fix the problem.

This is true for warnings within a single execution of `kdb*`-call (e.g. `kdbSet`) In applications many warnings might get collected until finally an error occurs and everything is presented to the user. Then the *most recent* warnings are relevant. Not what happened a long time ago.

Alternatively, we could simply clean all warnings before a `kdb*`-call. Then the behavior as described by @kodebach is best (only keep the earliest warnings). This decision does not mention this, however.

> Later warnings may be caused by earlier ones, if the earlier warning is overwritten it becomes hard to diagnose.

Yes, within a single kdb*-call it is like this.

> Why exactly the underscore?

It is needed for correct sorting.

> This is really confusing.

You are free to implement #2698. It would be highly appreciated.

Piankero: I am much more confused now than before. I still don't know why you simply drop complete warnings when the current warning number is saved in the parent meta (`warning/number`). When and how to use these weird underscores? Is there any document which explains this? I will stick to the array number of 100 because there is neither better a reason to use 1000 nor to use 10.
What exactly is the desired behavior now when `n` is exceeded?
I feel like consuming more time from you rather if someone else implements this with more insight in Elektra ...

kodebach: > When and how to use these weird underscores?

Simply put: Elektra sorts lexicographically using ASCII codes. `_` has a higher ASCII code than the digits `0`-`9`. Therefore by prefixing the index with `_` we ensure that the sort order is `#0`, `#1`, `#2`, ..., `#9`, `#_10` instead of `#0`, `#1`, `#10`, `#2`, ..., `#9`. You need more underscores, the longer the number is because otherwise the same problem occurs for 10 and 100 instead of 1 and 10.

> Is there any document which explains this?

Yes, https://github.com/ElektraInitiative/libelektra/blob/master/doc/tutorials/arrays.md. Feel free to improve it, if it isn't clear enough.

> What exactly is the desired behavior now when n is exceeded?

For now, I think we should stick to the old behaviour of overwriting old warnings (so basically the index is always modulo `n`), because of the problem with multiple kdb*-calls.

markus2330:
Why did you close this?

Piankero:
Because I will not implement it anyway. But I guess the decision itself can be finished

markus2330:
Actually I thought you already implemented it. Are there any other leftovers of things you have already started? Dropping implementation work that was already started should have a very good reason.

For the arrays, simply change `#00` to `#0` and `#10` to `#_10` and put the last entry as value so that they are proper Elektra (meta-)arrays and leave the logic otherwise as-is. This should be acceptable effort (no discussion needed, you only implement what is already decided about arrays).

And I would also like to merge this PR, simply update the decision to leave it as-is (except of array entry key names).

markus2330:
@Piankero Thank you! I am looking forward to the implementation of the non-generated macros that honor Elektra arrays.

PR #2686, opened 2019-05-09
Error codes implementation

Piankero:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests
- [x] meta data is updated (e.g. README.md of plugins and doc/METADATA.md)

## Review

## Merging

### DIFF
--- a/src/bindings/cpp/include/kdbplugin.hpp
+++ b/src/bindings/cpp/include/kdbplugin.hpp
@@ -76,7 +76,6 @@ private:
 		catch (const char * msg)
 		{
 #ifdef KDBERRORS_H
-			// TODO: Correct?
 			ELEKTRA_ADD_PLUGIN_MISBHV_WARNINGF (errorKey, "Could not create C++ plugin: %s", msg);
 #endif
 			return -1;
--- a/src/bindings/cpp/include/kdbplugin.hpp
+++ b/src/bindings/cpp/include/kdbplugin.hpp
@@ -77,7 +77,7 @@ private:
 		{
 #ifdef KDBERRORS_H
 			// TODO: Correct?
-			ELEKTRA_ADD_BROKEN_PLUGIN_WARNINGF (errorKey, "Could not create C++ plugin: %s", msg);
+			ELEKTRA_ADD_PLUGIN_MISBHV_WARNINGF (errorKey, "Could not create C++ plugin: %s", msg);
 #endif
 			return -1;
 		}
--- a/src/bindings/cpp/include/kdbplugin.hpp
+++ b/src/bindings/cpp/include/kdbplugin.hpp
@@ -76,7 +76,7 @@ private:
 		catch (const char * msg)
 		{
 #ifdef KDBERRORS_H
-			ELEKTRA_ADD_PLUGIN_MISBEHAVIOR_WARNINGF (errorKey, "Could not create C++ plugin: %s", msg);
+			ELEKTRA_ADD_PLUGIN_MISBEHAVIOR_WARNINGF (errorKey, "Could not create C++ plugin. Reason: %s", msg);
 #endif
 			return -1;
 		}
--- a/src/bindings/cpp/include/kdbplugin.hpp
+++ b/src/bindings/cpp/include/kdbplugin.hpp
@@ -76,7 +76,7 @@ private:
 		catch (const char * msg)
 		{
 #ifdef KDBERRORS_H
-			ELEKTRA_ADD_PLUGIN_MISBHV_WARNINGF (errorKey, "Could not create C++ plugin: %s", msg);
+			ELEKTRA_ADD_PLUGIN_MISBEHAVIOR_WARNINGF (errorKey, "Could not create C++ plugin: %s", msg);
 #endif
 			return -1;
 		}
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
@@ -249,7 +249,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 				os << "	va_list arg;" << endl
 				   << "	va_start(arg, module);" << endl
 				   << "	char * r = elektraVFormat(reason, arg);" << endl
-				   << " 	if(meta)" << endl
+				   << " 	if (meta)" << endl
 				   << "			keySetMeta(errorKey, buffer, r);" << endl
 				   << " 	else" << endl
 				   << "			keySetMeta(errorKey, \"error/reason\", "
@@ -260,7 +260,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			}
 			else
 			{
-				os << " 	if(meta)" << endl
+				os << " 	if (meta)" << endl
 				   << "			keySetMeta(errorKey, buffer, reason);" << endl
 				   << " 	else" << endl
 				   << "			keySetMeta(errorKey, \"error/reason\", reason);" << endl;
@@ -293,7 +293,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 	os << "static inline void elektraTriggerWarnings (const char *nr, Key *parentKey, const char *message)" << endl << "{" << endl;
 	for (size_t i = 1; i < p.size (); ++i)
 	{
-		os << "	if(strcmp(nr, \"" << p[i]["number"] << "\") == 0)" << endl << "	{" << endl;
+		os << "	if (strcmp(nr, \"" << p[i]["number"] << "\") == 0)" << endl << "	{" << endl;
 		os << "		ELEKTRA_ADD_" << p[i]["macro"] << "_WARNING (parentKey, message);" << endl
 		   << "		return;" << endl
 		   << "	}" << endl;
@@ -302,7 +302,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 	os << "static inline void elektraTriggerError (const char *nr, Key *parentKey, const char *message)" << endl << "{" << endl;
 	for (size_t i = 1; i < p.size (); ++i)
 	{
-		os << "	if(strcmp(nr, \"" << p[i]["number"] << "\") == 0)" << endl << "	{" << endl;
+		os << "	if (strcmp(nr, \"" << p[i]["number"] << "\") == 0)" << endl << "	{" << endl;
 		os << "		ELEKTRA_SET_" << p[i]["macro"] << "_ERROR (parentKey, message);" << endl
 		   << "		return;" << endl
 		   << "	}" << endl;
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
@@ -298,7 +298,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 		   << "		return;" << endl
 		   << "	}" << endl;
 	}
-	os << "	ELEKTRA_ADD_INTERNAL_WARNING (parentKey, \"Unkown warning code\");" << endl << "}" << endl << "" << endl;
+	os << "	ELEKTRA_ADD_INTERNAL_WARNINGF (parentKey, \"Unkown warning code %s\", nr);" << endl << "}" << endl << "" << endl;
 	os << "static inline void elektraTriggerError (const char *nr, Key *parentKey, const char *message)" << endl << "{" << endl;
 	for (size_t i = 1; i < p.size (); ++i)
 	{
@@ -307,7 +307,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 		   << "		return;" << endl
 		   << "	}" << endl;
 	}
-	os << "	ELEKTRA_SET_INTERNAL_ERROR (parentKey, \"Unkown error code\");" << endl << "}" << endl;
+	os << "	ELEKTRA_SET_INTERNAL_ERRORF (parentKey, \"Unkown error code %s\", nr);" << endl << "}" << endl;
 
 	os << "#endif" << endl;
 	return os;
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
@@ -129,7 +129,9 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "	keySetMeta(warningKey, buffer, \"" << p[i]["number"] << "\");" << endl
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/description\");" << endl
 			   << "	keySetMeta(warningKey, buffer, \"" << p[i]["description"] << "\");" << endl
-			   << "	buffer[12] = '\\0'; strcat(buffer, \"/module\");" << endl
+			   << "	buffer[12] = '\\0'; strcat(buffer, \"/module\");"
+			   << endl
+			   // TODO: Not the best implementation, fix if better way is found
 			   << "	if (strcmp(module, \"ELEKTRA_MODULE_NAME\") == 0) module = \"kdb\";" << endl
 			   << "	keySetMeta(warningKey, buffer, module);" << endl
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/file\");" << endl // should be called sourcefile
@@ -203,7 +205,9 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "		keySetMeta(errorKey, buffer, \"" << p[i]["number"] << "\");" << endl
 			   << "		buffer[12] = '\\0'; strcat(buffer, \"/description\");" << endl
 			   << "		keySetMeta(errorKey, buffer, \"" << p[i]["description"] << "\");" << endl
-			   << "		buffer[12] = '\\0'; strcat(buffer, \"/module\");" << endl
+			   << "		buffer[12] = '\\0'; strcat(buffer, \"/module\");"
+			   << endl
+			   // TODO: Not the best implementation, fix if better way is found
 			   << "		if (strcmp(module, \"ELEKTRA_MODULE_NAME\") == 0) module = \"kdb\";" << endl
 			   << "		keySetMeta(errorKey, buffer, module);" << endl
 			   << "		buffer[12] = '\\0'; strcat(buffer, \"/file\");" << endl // should be called sourcefile
@@ -222,7 +226,9 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "number description  module file line function reason"
 			   << "\");" << endl
 			   << "		keySetMeta(errorKey, \"error/number\", \"" << p[i]["number"] << "\");" << endl
-			   << "		keySetMeta(errorKey, \"error/description\", \"" << p[i]["description"] << "\");" << endl
+			   << "		keySetMeta(errorKey, \"error/description\", \"" << p[i]["description"] << "\");"
+			   << endl
+			   // TODO: Not the best implementation, fix if better way is found
 			   << "		if (strcmp(module, \"ELEKTRA_MODULE_NAME\") == 0) module = \"kdb\";" << endl
 			   << "		keySetMeta(errorKey, \"error/module\", module);" << endl
 			   << "		keySetMeta(errorKey, \"error/file\", "
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
@@ -130,6 +130,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/description\");" << endl
 			   << "	keySetMeta(warningKey, buffer, \"" << p[i]["description"] << "\");" << endl
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/module\");" << endl
+			   << "	if (strcmp(module, \"ELEKTRA_MODULE_NAME\") == 0) module = \"kdb\";" << endl
 			   << "	keySetMeta(warningKey, buffer, module);" << endl
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/file\");" << endl // should be called sourcefile
 			   << "	keySetMeta(warningKey, buffer, file);" << endl
@@ -203,6 +204,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "		buffer[12] = '\\0'; strcat(buffer, \"/description\");" << endl
 			   << "		keySetMeta(errorKey, buffer, \"" << p[i]["description"] << "\");" << endl
 			   << "		buffer[12] = '\\0'; strcat(buffer, \"/module\");" << endl
+			   << "		if (strcmp(module, \"ELEKTRA_MODULE_NAME\") == 0) module = \"kdb\";" << endl
 			   << "		keySetMeta(errorKey, buffer, module);" << endl
 			   << "		buffer[12] = '\\0'; strcat(buffer, \"/file\");" << endl // should be called sourcefile
 			   << "		keySetMeta(errorKey, buffer, file);" << endl
@@ -221,6 +223,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "\");" << endl
 			   << "		keySetMeta(errorKey, \"error/number\", \"" << p[i]["number"] << "\");" << endl
 			   << "		keySetMeta(errorKey, \"error/description\", \"" << p[i]["description"] << "\");" << endl
+			   << "		if (strcmp(module, \"ELEKTRA_MODULE_NAME\") == 0) module = \"kdb\";" << endl
 			   << "		keySetMeta(errorKey, \"error/module\", module);" << endl
 			   << "		keySetMeta(errorKey, \"error/file\", "
 			   << "file"
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
@@ -128,7 +128,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "	strcat(buffer, \"/number\" );" << endl
 			   << "	keySetMeta(warningKey, buffer, \"" << p[i]["number"] << "\");" << endl
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/description\");" << endl
-			   << "	keySetMeta(warningKey, buffer, \"" << p[i]["description"] << "\");" <<  endl
+			   << "	keySetMeta(warningKey, buffer, \"" << p[i]["description"] << "\");" << endl
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/module\");" << endl
 			   << "	keySetMeta(warningKey, buffer, module);" << endl
 			   << "	buffer[12] = '\\0'; strcat(buffer, \"/file\");" << endl // should be called sourcefile
@@ -220,7 +220,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 			   << "number description  module file line function reason"
 			   << "\");" << endl
 			   << "		keySetMeta(errorKey, \"error/number\", \"" << p[i]["number"] << "\");" << endl
-			   << "		keySetMeta(errorKey, \"error/description\", \"" << p[i]["description"] << "\");" <<  endl
+			   << "		keySetMeta(errorKey, \"error/description\", \"" << p[i]["description"] << "\");" << endl
 			   << "		keySetMeta(errorKey, \"error/module\", module);" << endl
 			   << "		keySetMeta(errorKey, \"error/file\", "
 			   << "file"
@@ -271,7 +271,8 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 		   << "			KEY_END)," << endl
 		   << "		keyNew (\"system/elektra/modules/error/specification/" << p[i]["number"] << "/description\"," << endl
 		   << "			KEY_VALUE, \"" << p[i]["description"] << "\", KEY_END)," << endl;
-		//		   << "		keyNew (\"system/elektra/modules/error/specification/" << p[i]["number"] << "/severity\"," <<
+		//		   << "		keyNew (\"system/elektra/modules/error/specification/" << p[i]["number"] << "/severity\","
+		//<<
 		// endl
 		//		   << "			KEY_VALUE, \"" << p[i]["severity"] << "\", KEY_END)," << endl
 		//		   << "		keyNew (\"system/elektra/modules/error/specification/" << p[i]["number"] << "/module\"," <<
--- a/src/error/exporterrors.cpp
+++ b/src/error/exporterrors.cpp
@@ -298,7 +298,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 		   << "		return;" << endl
 		   << "	}" << endl;
 	}
-	os << "	ELEKTRA_ADD_ASSERTION_WARNING (parentKey, \"Unkown warning code\");" << endl << "}" << endl << "" << endl;
+	os << "	ELEKTRA_ADD_INTERNAL_WARNING (parentKey, \"Unkown warning code\");" << endl << "}" << endl << "" << endl;
 	os << "static inline void elektraTriggerError (const char *nr, Key *parentKey, const char *message)" << endl << "{" << endl;
 	for (size_t i = 1; i < p.size (); ++i)
 	{
@@ -307,7 +307,7 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
 		   << "		return;" << endl
 		   << "	}" << endl;
 	}
-	os << "	ELEKTRA_SET_ASSERTION_ERROR (parentKey, \"Unkown error code\");" << endl << "}" << endl;
+	os << "	ELEKTRA_SET_INTERNAL_ERROR (parentKey, \"Unkown error code\");" << endl << "}" << endl;
 
 	os << "#endif" << endl;
 	return os;
--- a/src/error/specification
+++ b/src/error/specification
@@ -33,7 +33,7 @@ macro:INTERFACE
 
 number:C01430
 description:Plugin Misbehavior
-macro:PLUGIN_MISBHV
+macro:PLUGIN_MISBEHAVIOR
 
 number:C02000
 description:Conflict
--- a/src/error/specification
+++ b/src/error/specification
@@ -24,8 +24,8 @@ description:Installation
 macro:INSTALLATION
 
 number:C01410
-description:Assertion
-macro:ASSERTION
+description:Internal
+macro:INTERNAL
 
 number:C01420
 description:Interface
--- a/src/error/specification
+++ b/src/error/specification
@@ -19,10 +19,6 @@ number:C01120
 description:General Resource
 macro:RESOURCE
 
-number:C01200
-description:Parsing
-macro:PARSING
-
 number:C01300
 description:Installation
 macro:INSTALLATION
--- a/src/error/specification
+++ b/src/error/specification
@@ -4,6 +4,11 @@ description:A test for an exception specification
 :the code is similar to SQLSTATES and corresponds
 :to a specific category
 :
+:For a categorization guideline please visit
+:doc/dev/error-categorization.md
+:The related decision can be found in
+:doc/decisions/error_codes.md
+:
 :there are no comments possible in this file
 ignore:true
 :a generator should always ignore this first entry
--- a/src/error/specification
+++ b/src/error/specification
@@ -11,13 +11,13 @@ ignore:true
 :It is for errors which are not relevant anymore
 :an empty line will start the next error specification
 
+number:C01100
+description:Resource
+macro:RESOURCE
+
 number:C01110
 description:Out of Memory
 macro:OUT_OF_MEMORY
-
-number:C01120
-description:General Resource
-macro:GENERAL_RESOURCE
 
 number:C01300
 description:Installation
@@ -32,25 +32,17 @@ description:Interface
 macro:INTERFACE
 
 number:C01430
-description:Broken plugin
-macro:BROKEN_PLUGIN
+description:Plugin Misbehavior
+macro:PLUGIN_MISBHV
 
 number:C02000
 description:Conflict
 macro:CONFLICTING_STATE
 
-number:C03000
-description:Timeout
-macro:TIMEOUT
-
-number:C04100
+number:C03100
 description:Validation Syntactic
 macro:VALIDATION_SYNTACTIC
 
-number:C04200
+number:C03200
 description:Validation Semantic
 macro:VALIDATION_SEMANTIC
-
-number:C05000
-description:Out of Range
-macro:OUT_OF_RANGE
--- a/src/error/specification
+++ b/src/error/specification
@@ -17,7 +17,7 @@ macro:OUT_OF_MEMORY
 
 number:C01120
 description:General Resource
-macro:RESOURCE
+macro:GENERAL_RESOURCE
 
 number:C01300
 description:Installation
--- a/src/error/specification
+++ b/src/error/specification
@@ -50,3 +50,7 @@ macro:VALIDATION_SYNTACTIC
 number:C04200
 description:Validation Semantic
 macro:VALIDATION_SEMANTIC
+
+number:C05000
+description:Out of Range
+macro:OUT_OF_RANGE
--- a/src/error/specification
+++ b/src/error/specification
@@ -37,7 +37,7 @@ macro:BROKEN_PLUGIN
 
 number:C02000
 description:Conflict
-macro:CONFLICT
+macro:CONFLICTING_STATE
 
 number:C03000
 description:Timeout
--- a/src/include/elektra/error.h
+++ b/src/include/elektra/error.h
@@ -9,6 +9,7 @@
 #ifndef ELEKTRA_ERROR_H
 #define ELEKTRA_ERROR_H
 
+#include "types.h"
 #include <kdb.h>
 #include <stdbool.h>
 
@@ -27,6 +28,10 @@ const char * elektraErrorCode (const ElektraError * error);
 const char * elektraErrorDescription (const ElektraError * error);
 
 void elektraErrorReset (ElektraError ** error);
+
+// for code-generation
+ElektraError * elektraErrorConversionToString (KDBType sourceType, const char * keyname);
+ElektraError * elektraErrorConversionFromString (KDBType targetType, const char * keyname, const char * sourceValue);
 
 #ifdef __cplusplus
 }
--- a/src/include/kdbmacros.h
+++ b/src/include/kdbmacros.h
@@ -37,13 +37,13 @@
 	do                                                                                                                                 \
 	{                                                                                                                                  \
 		if (errno == EACCES)                                                                                                       \
-			ELEKTRA_SET_RESOURCE_ERRORF (                                                                                      \
+			ELEKTRA_SET_GENERAL_RESOURCE_ERRORF (                                                                                      \
 				parentKey,                                                                                                 \
 				"Insufficient permissions to open configuration file for reading. Reason: %s. You might want "             \
 				"to retry as root or change access using chmod.",                                                          \
 				strerror (errno));                                                                                         \
 		else                                                                                                                       \
-			ELEKTRA_SET_RESOURCE_ERRORF (parentKey, "Could not open configuration file for reading. Reason: %s",               \
+			ELEKTRA_SET_GENERAL_RESOURCE_ERRORF (parentKey, "Could not open configuration file for reading. Reason: %s",               \
 						     strerror (errno));                                                                    \
 	} while (0)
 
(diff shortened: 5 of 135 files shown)


markus2330:
I think we should not merge this except shortly before the release. After the release everyone needs to rebase anyway.

sanssecours:
> I think we should not merge this except shortly before the release.

In that case I think we can merge #2690 right now.

markus2330:
Sorry, with "this" I was referring to #2690. The error codes should be first (if finished before doing the release).

Piankero:
@sanssecours can you help us debug the highlevel test which causes a [SEG fault](https://travis-ci.org/ElektraInitiative/libelektra/jobs/532370865)? It seems to be an apple-only bug and as far as I remember you have an apple

sanssecours:
> @sanssecours can you help us debug the highlevel test which causes a [SEG fault](https://travis-ci.org/ElektraInitiative/libelektra/jobs/532370865)? It seems to be an apple-only bug and as far as I remember you have an apple

Sorry, but as far as I know the Haskell binding does not compile on my version of macOS (`10.14.5`) anymore.

Piankero:
@sanssecours 
> Sorry, but as far as I know the Haskell binding does not compile on my version of macOS (10.14.5) anymore.

Could you do me a favor and check if the `testkdb_highlevel` run on your system without a fault on this branchPR? It seems to be a problem only on this specific travis job and seems to be not mac specific

sanssecours:
> Could you do me a favor and check if the `testkdb_highlevel` run on your system without a fault on this branchPR?

The test runs successful on my machine. However, I do not think that is a big surprise, considering that my configuration does not include the Haskell binding.

kodebach:
> However, I do not think that is a big surprise, considering that my configuration does not include the Haskell binding.

You can actually see [here](https://travis-ci.com/Piankero/libelektra/builds/113214655) that the Haskell binding doesn't cause the problem.

Piankero:
It seems to be a macOs version problem. when updating to 10.2 it works
https://travis-ci.com/Piankero/libelektra/jobs/203370290

sanssecours:
>  is there a particular reason why the macOS version is 10.1 for the HASKELL build?

According to the [Travis documentation](https://docs.travis-ci.com/user/reference/osx/#macos-version) the actual macOS version used for `🍏 Haskell` is macOS 10.13. The value `10.1` is the [Xcode](https://developer.apple.com/xcode) version used for that specific build job.

> Can we upgrade it to 10.2?

I do not think so. The problem is already mentioned inside [`.travis.yml`](https://github.com/ElektraInitiative/libelektra/blob/master/.travis.yml#L196):

https://github.com/ElektraInitiative/libelektra/blob/54a4c0194946917b7d093e0777f465619b2f3d6f/.travis.yml#L196

. You can see that the build job fails, if we use macOS 10.14 (and Xcode 10.2) [here](https://travis-ci.org/sanssecours/elektra/jobs/538177568).

Piankero:
> You can see that the build job fails, if we use macOS 10.14 (and Xcode 10.2) here.

I see, is there a good way to exclude these tests just on that specific xcode version? Do you have some experience in this regard? Or is it better to make something like `#ifndef HASKELL`?

sanssecours:
> I see, is there a good way to exclude these tests just on that specific xcode version?

I think the CMake variable [`XCODE_VERSION`](https://cmake.org/cmake/help/v3.0/variable/XCODE_VERSION.html) might be helpful.

> Do you have some experience in this regard?

I already wrote CMake code that excludes tests, if the current CMake configuration does not meet specific criteria. However, I did not exclude a test, because of a specific Xcode version yet.

> Or is it better to make something like `#ifndef HASKELL`?

That does not sound like a good idea in my opinion, especially since [Klemens already mentioned](https://github.com/ElektraInitiative/libelektra/pull/2686#issuecomment-496287743) that the Haskell binding/plugin does not seem to be the cause of the segmentation fault.

Piankero:
> I think the CMake variable [`XCODE_VERSION`](https://cmake.org/cmake/help/v3.0/variable/XCODE_VERSION.html) might be helpful.

Thank you very much for that suggestion.
As a Linux user it is very difficult for me to get mac specific code correct in cmake.

Do you think that the following code can work for the [CMakeLists.txt](https://github.com/ElektraInitiative/libelektra/blob/cb74c10bb16fcf6445f692ea3f6e03b84eb1058e/tests/kdb/CMakeLists.txt)?
```
if (XCODE)
	if (!XCODE_VERSION MATCHES "10.1.13")
		set_source_files_properties (testkdb_highlevel PROPERTIES COMPILE_FLAGS -Wno-sign-promo)
		add_kdb_test (highlevel LINK_ELEKTRA elektra-highlevel REQUIRED_PLUGINS error)
	endif ()
else()
	set_source_files_properties (testkdb_highlevel PROPERTIES COMPILE_FLAGS -Wno-sign-promo)
	add_kdb_test (highlevel LINK_ELEKTRA elektra-highlevel REQUIRED_PLUGINS error)
endif ()
```

kodebach:
Instead of using `MATCHES` with a regex, you could use CMake's `VERSION_LESS`.

Piankero:
> `VERSION_LESS`

Thank you for the suggestion! I edited my comment from above

sanssecours:
> Do you think that the following code can work for the [CMakeLists.txt](https://github.com/ElektraInitiative/libelektra/blob/cb74c10bb16fcf6445f692ea3f6e03b84eb1058e/tests/kdb/CMakeLists.txt)?

I do not think you can use `!` in CMake code. The correct keyword should be `NOT`, as far as I know. Otherwise the code looks okay to me. Below I wrote down some suggestions, which would improve the code in my opinion.

- Please do not use copy and paste. You can also write code that does the same, but only uses `add_kdb_test` and `set_source_files_properties` once.
- Please use `VERSION_EQUAL` instead of `VERSION_LESS`, unless you are sure that the test also fails on Xcode versions before Xcode 10.1.
- I think you should also include the CMake variable [`APPLE`](https://cmake.org/cmake/help/v3.0/variable/APPLE.html) in the `if`-statement. It does not make sense to check the Xcode version, unless the the current system runs macOS.
- Please do not use nested `if`-statements, if you do not have to. As far as I know one `if`-statement should be enough.

Piankero:
Unfortunately the approach does not work :(

```
if (NOT
(APPLE
	AND XCODE
	AND XCODE_VERSION
	VERSION_EQUAL
	10.1))
	message(STATUS =============/ ${APPLE} /================)
	message(STATUS =============/ ${XCODE} /================)
	message(STATUS =============/ ${XCODE_VERSION} /================)
	set_source_files_properties (testkdb_highlevel PROPERTIES COMPILE_FLAGS -Wno-sign-promo)
	add_kdb_test (highlevel LINK_ELEKTRA elektra-highlevel REQUIRED_PLUGINS error)
endif ()
```
returns
```
-- =============/1/================
-- =============//================
-- =============//================
```

sanssecours:
> Unfortunately the approach does not work :(

Look like [`XCODE_VERSION`](https://cmake.org/cmake/help/v3.0/variable/XCODE_VERSION.html) is only set, if you use Xcode as generator. I wrote a CMake function that should fix this issue:

```cmake
# This function checks information about the currently selected version of
# Xcode.
#
# The function exports the variables described below.
#
# * XCODE_INSTALLED: This variable specifies, if the current system uses Xcode
#   or not.
# * XCODE_VERSION: This variable stores the version string of the selected Xcode
#   version. If Xcode is not installed, then it stores the value `0`.
#
function(check_xcode)
  set(XCODE_INSTALLED OFF PARENT_SCOPE)
  set(XCODE_VERSION 0 PARENT_SCOPE)

  if(NOT APPLE)
    return()
  endif(NOT APPLE)

  execute_process(COMMAND xcode-select --print-path
                  RESULT_VARIABLE XCODE_SELECT_UNSUCCESSFULL
                  OUTPUT_VARIABLE XCODE_PATH
                  OUTPUT_STRIP_TRAILING_WHITESPACE)

  if(XCODE_SELECT_UNSUCCESSFULL)
    return()
  endif(XCODE_SELECT_UNSUCCESSFULL)

  execute_process(COMMAND "${XCODE_PATH}/usr/bin/xcodebuild" -version
                  RESULT_VARIABLE XCODE_BUILD_UNSUCCESSFULL
                  OUTPUT_VARIABLE XCODE_VERSION_STRING
                  OUTPUT_STRIP_TRAILING_WHITESPACE)

  if(XCODE_BUILD_UNSUCCESSFULL)
    return()
  endif(XCODE_BUILD_UNSUCCESSFULL)

  string(REGEX
         REPLACE "Xcode ([0-9]+\\.[0-9]+\\.[0-9]+).*"
                 "\\1"
                 XCODE_VERSION_STRING
                 ${XCODE_VERSION_STRING})

  set(XCODE_INSTALLED ON PARENT_SCOPE)
  set(XCODE_VERSION "${XCODE_VERSION_STRING}" PARENT_SCOPE)
endfunction(check_xcode)
```

. To use the function above, call it, and check the value of the variable `XCODE_VERSION` afterwards:

```cmake
check_xcode()
if(NOT ${XCODE_VERSION} VERSION_EQUAL 10.1)
  message("Include test")
endif(NOT ${XCODE_VERSION} VERSION_EQUAL 10.1)
```

. I uploaded a simple example project that uses `check_xcode` [here](https://github.com/ElektraInitiative/libelektra/files/3236281/test.zip). If I invoke `make` in the root of  the unexpanded archive, then `cmake` prints the following text on my machine 

```
…
XCODE_INSTALLED: ON
XCODE_VERSION:   10.2.1
Include test
…
```

.

Piankero:
Thank you very much for your help!
I tried this on my branch with with https://github.com/Piankero/libelektra/pull/3/commits/1dee7acc1fd6be47c8b2071d65cbc2e891fceba3 but unfortunately without success.
I also tried another approach though which worked in https://github.com/Piankero/libelektra/pull/3/commits/6b468e3e1d3e18efb2ffb80d98a52abcca1fee64 and also makes sense. Would this be a valid approach too for you?

sanssecours:
> I tried this on my branch with with Piankero@1dee7ac but unfortunately without success.

Seems the version string returned by Xcode does not always contain the patch version number. I [adapted the code accordingly](https://github.com/sanssecours/elektra/commit/1b785bdfabfe9fe9493dac589a5f7ab5425d5b7a):

```cmake
string (REGEX
	REPLACE ".*Xcode[ ]*([0-9]+(\\.[0-9]+)?(\\.[0-9]+)?).*"
		"\\1"
		XCODE_VERSION_STRING
		${XCODE_VERSION_STRING})
```

. After that update [`testscr_check_gen` fails with a segmentation fault](https://travis-ci.org/sanssecours/elektra/builds/539198538#L2030) on my [modified test branch](https://github.com/sanssecours/elektra/tree/🏗).

I also applied my changes on [a copy of this PR](https://github.com/sanssecours/elektra/tree/🗄). The Travis build results for this branch are available [here](https://travis-ci.org/sanssecours/elektra/builds/539204941).

By the way, if Xcode 10.1 really is the problem, why does `testkdb_highlevel` [work without any problems on Cirrus CI](https://cirrus-ci.com/task/6260837438717952), which also uses Xcode 10.1 (on macOS 10.14):

https://github.com/ElektraInitiative/libelektra/blob/7da63e6f736dca2425c1945609eb82011ad406ea/.cirrus.yml#L57

?

> Would this be a valid approach too for you?

It would also work. However, I am not a fan of CI specific changes. I am pretty sure a new Elektra developer (that develops on macOS 10.13 and uses Xcode 10.1) would not know he has to turn on `EXCLUDE_HIGHLEVEL_TEST`.

Piankero:
Thank you very much for your help here. It is very frustrating to debug something just via CI logs that takes 30 minutes for a response.

> By the way, if Xcode 10.1 really is the problem, why does `testkdb_highlevel` [work without any problems on Cirrus CI](https://cirrus-ci.com/task/6260837438717952), which also uses Xcode 10.1 (on macOS 10.14):

@kodebach even tried to investigate this problem on a mac book of his friend and was not able to find the concrete cause. He assumes some travis specific magic that causes that trouble.

kodebach:
> @kodebach even tried to investigate this problem on a mac book of his friend and was not able to find the concrete cause.

I did use Xcode 10.2, however. I only later noticed the Xcode version. But I followed the Travis log step-by-step, which lead me to the conclusion that either Travis itself, or the Xcode/macOS version causes the problem.

> By the way, if Xcode 10.1 really is the problem, why does testkdb_highlevel work without any problems on Cirrus CI, which also uses Xcode 10.1 (on macOS 10.14):

It that case probably either macOS 10.13 or the combination of macOS 10.13 and Xcode 10.1 is the problem. It could also be pos

sanssecours:
> It that case probably either macOS 10.13 or the combination of macOS 10.13 and Xcode 10.1 is the problem. 

I think we should also check for the macOS version then, and only disable the test on the system combination macOS 10.13/Xcode 10.1. CMake should also print a message for disabled tests. Something like 

```cmake
# See: https://issues.libelektra.org/2686
message (STATUS "Exclude KDB high-level test, because it fails with a segmentation fault on the current system (macOS 10.13/Xcode 10.1)")
```

would make sense, in my opinion.

kodebach:
I just ran into the same mysterious segfault that only occurs on our Haskell Travis build in #2707. I will look into it further as it appeared differently, but maybe there actually is a bug here.

Piankero:
The build fails but that is ok since we will need to delete 
usr/include/elektra/elektra/errorcodes.h from debian/libelektra-dev.install
from the `debian` branch before we decide to merge.
The remaining ~11 TODO's concerning categorization we could do afterwards.

src/bindings/cpp/include/kdbplugin.hpp
```diff
@@ -76,7 +76,7 @@ class Delegator
 		catch (const char * msg)
 		{
 #ifdef KDBERRORS_H
-			ELEKTRA_ADD_WARNING (69, errorKey, msg);
+			ELEKTRA_ADD_PLUGIN_MISBHV_WARNINGF (errorKey, "Could not create C++ plugin: %s", msg);
```
markus2330: Please write the full name, not MISBHV

src/error/exporterrors.cpp
```diff
@@ -38,57 +38,32 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
...
-	   << "	(key, text, file, #line, __VA_ARGS__); } while (0)" << endl
-	   << endl
-	   << endl;
+	   << "using namespace ckdb;" << endl
+	   << "#endif" << endl;
+	os << endl << endl;
+
+	for (size_t i = 1; i < p.size (); ++i)
+	{
+		os << "#define ELEKTRA_SET_" << p[i]["macro"] << "_ERROR(key, text) \\" << endl
```
markus2330: Maybe it is too late now but actually it would make a lot of sense to rewrite this to use mustache (it is C++ anyway and would be much cleaner then).
@kodebach what do you think? Would it be easy to share the mustache header?

src/error/exporterrors.cpp
```diff
@@ -301,247 +276,88 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
...
 	{
-		if (p[i]["unused"] == "yes")
-		{
-			continue;
-		}
-
-		if (p[i]["severity"] != "warning") continue;
-		os << "		case " << i << ": ELEKTRA_ADD_WARNING (" << i << ", parentKey, message);" << endl
-		   << "			break;" << endl;
+		os << "	if(strcmp(nr, \"" << p[i]["number"] << "\") == 0)" << endl << "	{" << endl;
```
markus2330: Are we sure that the strcmp is safe? Coding guidelines are violated here: space after if.

src/error/exporterrors.cpp
```diff
@@ -301,247 +276,88 @@ static ostream & printKDBErrors (ostream & os, parse_t & p)
...
-	   << "}" << endl;
-
+	os << "	ELEKTRA_SET_INTERNAL_ERROR (parentKey, \"Unkown error code\");" << endl << "}" << endl;
 
 	os << "#endif" << endl;
 	return os;
 }
 
-static ostream & printCreatorSignature (ostream & os, map<string, string> & data)
+static string macroCaseToPascalCase (const string & s)
```
markus2330: What is PascalCase? You mean CamelCase?

src/error/specification
```diff
@@ -1,1354 +1,48 @@
-number:0
-description:a test for an exception specification
+number:C00000
+description:A test for an exception specification
 :new lines to the same entry can be started with a colon
-:the number is sequential and has no special meaning
+:the code is similar to SQLSTATES and corresponds
+:to a specific category
```
markus2330: Link to your tutorials and decisions.

src/error/specification
```diff
@@ -1,1354 +1,48 @@
...
 
-number:1
-ingroup:modules
-description:could not load module, dlopen failed
-module:dl
-severity:warning
-class:FailedLoadModuleError
-extends:ModuleError
+number:C01100
+description:Resource
```
markus2330: This description does not help.

src/error/specification
```diff
@@ -1,1354 +1,48 @@
...
+description:Interface
+macro:INTERFACE
 
-number:6
-description:Could not call function exported by ELEKTRA_PLUGIN_EXPORT
-ingroup:kdb
-severity:warning
+number:C01430
+description:Plugin Misbehavior
+macro:PLUGIN_MISBHV
```
markus2330: See above: please avoid this name.

src/include/elektra/error.h
```diff
@@ -22,10 +22,17 @@ typedef struct _ElektraError ElektraError;
 
 typedef void (*ElektraErrorHandler) (ElektraError * error);
 
+ElektraError * elektraErrorPureWarning (void);
```
markus2330: What is PureWarning? Design decision for API missing.

src/include/kdbmacros.h
```diff
@@ -37,24 +37,33 @@
 	do                                                                                                                                 \
 	{                                                                                                                                  \
 		if (errno == EACCES)                                                                                                       \
-			ELEKTRA_SET_ERROR (109, parentKey, strerror (errno));                                                              \
+			ELEKTRA_SET_RESOURCE_ERRORF (                                                                                      \
+				parentKey,                                                                                                 \
+				"Insufficient permissions to open configuration file for reading. Reason: %s. You might want "             \
```
markus2330: for reading what?

src/include/kdbmacros.h
```diff
@@ -37,24 +37,33 @@
...
+				"to retry as root or change access using chmod.",                                                          \
```
markus2330: Somehow mark this (maybe with TODO SOLUTION): This is a solution and should be separated.

src/include/kdbmacros.h
```diff
@@ -37,24 +37,33 @@
...
+				strerror (errno));                                                                                         \
 		else                                                                                                                       \
-			ELEKTRA_SET_ERROR (110, parentKey, strerror (errno));                                                              \
+			ELEKTRA_SET_RESOURCE_ERRORF (parentKey, "Could not open configuration file for reading. Reason: %s",               \
```
markus2330: Which file?

src/include/kdbmacros.h
```diff
@@ -37,24 +37,33 @@
...
 
 #define ELEKTRA_SET_ERROR_SET(parentKey)                                                                                                   \
 	do                                                                                                                                 \
 	{                                                                                                                                  \
 		if (errno == EACCES)                                                                                                       \
-			ELEKTRA_SET_ERROR (9, parentKey, strerror (errno));                                                                \
+			ELEKTRA_SET_RESOURCE_ERRORF (                                                                                      \
+				parentKey,                                                                                                 \
+				"Insufficient permissions to open configuration file for writing. You might want to retry as "             \
+				"root. Errno: %s",                                                                                         \
```
markus2330: Also here is a solution.

src/include/kdbmacros.h
```diff
@@ -73,21 +82,30 @@
 			Key * c = ksNext (info);                                                                                           \
 			if (!c)                                                                                                            \
 			{                                                                                                                  \
-				ELEKTRA_SET_ERRORF (84, error, "the key %s (value %s) was added", keyName (k), keyString (k));             \
+				/* TODO: Correct? Maybe Interface? */                                                                      \
+				ELEKTRA_SET_INSTALLATION_ERRORF (                                                                          \
+					error, "Read only plugin, kdbSet not supported but the key %s (value %s) was added", keyName (k),  \
```
markus2330: but it was tried ... (because it did not succeed?)

src/include/kdbmacros.h
```diff
@@ -73,21 +82,30 @@
...
+					keyString (k));                                                                                    \
 				ksDel (info);                                                                                              \
 				return -1;                                                                                                 \
 			}                                                                                                                  \
 			if (strcmp (keyName (k), keyName (c)) || strcmp (keyString (k), keyString (c)))                                    \
 			{                                                                                                                  \
-				ELEKTRA_SET_ERRORF (84, error, "the key %s (expected %s) was modified to %s (expected %s)", keyName (k),   \
-						    keyName (c), keyString (k), keyString (c));                                            \
+				/* TODO: Correct? Maybe Interface? */                                                                      \
+				ELEKTRA_SET_INSTALLATION_ERRORF (                                                                          \
```
markus2330: I think it could be a validation error (read-only keys). Definitely not installation or interface.

tests/ctest/test_splitget.c
```diff
@@ -470,7 +470,7 @@ static void test_sizes (void)
 	succeed_if (output_error (parentKey), "error found");
 	// there will be a warning
 	// succeed_if (output_warnings(parentKey), "warning(s) found");
-	succeed_if_same_string (keyString (keyGetMeta (parentKey, "warnings/#00/number")), "79") // drop key
+	succeed_if_same_string (keyString (keyGetMeta (parentKey, "warnings/#00/number")), ELEKTRA_ERROR_INTERFACE) // drop key
```
markus2330: Very elegant that we now also can use macros to compare.

src/plugins/yamlcpp/README.md
```diff
@@ -186,34 +184,6 @@ kdb rm -r user/tests/yamlcpp
 sudo kdb umount user/tests/yamlcpp
 ```
 
-### Sparse Arrays
```
markus2330: Why are you removing this?

src/error/exporterrors.cpp
kodebach: https://stackoverflow.com/questions/41768733/camel-case-and-pascal-case-mistake

src/error/exporterrors.cpp
kodebach: > @kodebach what do you think?

I expected the new version of `exporterrors.cpp` to be much smaller, otherwise I would have suggested it already.

Since there the template is very static (only a few things change for each error), it should be fairly straightforward to switch to mustache. Although this could easily be done in a follow up PR.

> Would it be easy to share the mustache header?

We could simply copy it. Putting it into a common location would work, might be awkward since the only sensible location would be `src/include`.

markus2330: > Since there the template is very static (only a few things change for each error), it should be fairly straightforward to switch to mustache. Although this could easily be done in a follow up PR.

It is much easier to review mustache code than this print code.

>  We could simply copy it. Putting it into a common location would work, might be awkward since the only sensible location would be src/include.

Copying the copy is awkward. Maybe we create a src/include/external which contains header files we stole from other projects?

kodebach: > Maybe we create a src/include/external which contains header files we stole from other projects?

Maybe we should use `external/mustache/include` instead, i.e. one folder under `external` for each external dependency. Then it would also be easier to find and update external code. Right now some plugins (e.g. `ni`) also include source copies of libraries.

src/include/elektra/error.h
Piankero: This change was actually done by @kodebach to make the highlevel conform with the new error code system. CLion tells me it is unused though.

kodebach: https://github.com/ElektraInitiative/libelektra/blob/cbd85e11236b1d0c0eadb77d3cd0f2a8709470be/src/libs/highlevel/elektra_error.c#L208-L214

The doxygen documentation of `elektraErrorPureWarning` describes exactly what this is used for. We need it because the highlevel API only has the `ElektraError` struct and no separate `ElektraWarning` struct.

> CLion tells me it is unused though.

That is a mistake on my part, it should be used here (instead of the `elektraErrorCreate` call):

https://github.com/ElektraInitiative/libelektra/blob/cbd85e11236b1d0c0eadb77d3cd0f2a8709470be/src/libs/highlevel/elektra_error.c#L98-L101

Piankero: Fixed in https://github.com/ElektraInitiative/libelektra/pull/2686/commits/4966b8a830880bd39bffd6f036936933976f04e3

Piankero: Fixed in https://github.com/ElektraInitiative/libelektra/pull/2686/commits/9df13eb2affcc573879c005f066939abf1a81210

src/error/exporterrors.cpp
Piankero: We must compare a string here, what else would you suggest?

src/error/specification
Piankero: What exactly do you mean by "does not help"? Should I write something like `Resource related` or a whole label list picked from the categorization guideline, eg. `False permissions, missing files/ directory, not enough space, resource, etc.`? The description here is just a very short headline of the underlying error

src/include/kdbmacros.h
Piankero: I would need to look through all errors again anyway so no need to add an extra TODO here

src/include/kdbmacros.h
Piankero: What is unclear here? I think it is a decent error message. This macro was implemented by you 3 years ago, I guess you know it better what should be here

src/include/kdbmacros.h
Piankero: We agreed that eventual error message improvements will be done in another PR. This PR here has already again lots of comments

src/include/kdbmacros.h
Piankero: See above

src/error/specification
kodebach: IMO a link to the decision document containing the full descriptions should be good enough.

src/include/kdbmacros.h
Piankero: Validation error does not fit in this context. Installation might not fit because I assume that a different installation or compilation of elektra will suddenly make the plugin also support writing of keys. But since interface errors indicate a `misuse` as written in the guideline I think it actually fits here because someone misuses the plugin for setting keys. `Compared to semantic validation errors, this category
has its focus on detecting wrong usages of the API`. I think using a plugin in a not intended way fits into this category

src/plugins/yamlcpp/README.md
Piankero: This looks like rebase mistake from my side. I changed it accordingly

src/plugins/yamlcpp/README.md
```diff
@@ -71,7 +71,7 @@ sudo kdb umount /tests/yamlcpp
 
 ## Arrays
 
-YAML CPP provides support for Elektra’s array data type.
+YAML CPP provides basic support for Elektra’s array data type.
```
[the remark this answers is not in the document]
Piankero: This looks like rebase mistake from my side. I changed it accordingly

src/error/specification
Piankero: This is exactly what I have thought of. I added the link to the top of the file

src/include/kdbmacros.h
Piankero: See above

src/include/kdbmacros.h
markus2330: I thought that this PR is already the other PR. If you set "ready to merge" I expect **ALL** the code in the PR to be ready to merge. Please do not submit such large PRs for review if you do not want them to be reviewed, this wastes a lot of time on both sides.

markus2330:
Just to be sure it will not be overseen in the wall of comments:

If you set "ready to merge" I expect ALL the code in the PR to be ready to merge. Please do not submit such large PRs for review if you do not want them to be reviewed, this wastes a lot of time on both sides.

src/include/kdbmacros.h
Piankero: We agreed in the elektra meeting to keep PR's as little as possible. If I also fix *every of the 650* messages this will PR will get again out of hand ... We *definitely* agreed that this PR does only affect the error categorizations, not the message in addition

Piankero:
> If you set "ready to merge" I expect **ALL** the code in the PR to be ready to merge. Please do not submit such large PRs for review if you do not want them to be reviewed, this wastes a lot of time on both sides.

See my comment ... this is *exactly* what I intend to do ... Please review **ALL** the code because I fixed every of your concerns or wrote a comment to it when I do not intend to fix it in this PR. I even asked Rene to look over the misrebased yaml files again which he also approved. What else do you want?

src/include/kdbmacros.h
markus2330: The PR is not getting bigger if the messages are better. But ok, if you prefer to make it in a different PR, ok.

Piankero: I **highly** doubt this because up until now each and every PR which I had done with even minor subjective matters in it like wording, adding extra information, removing adding "sorry", etc gained tons of comments. I will not do that again.

markus2330:
> What else do you want?

The problem for me was that you introduce text for the error messages (maybe copied from description) and I saw improvements for the text.

What advantage do you see to now introduce wrong text and to fix the text in a later PR?

Piankero:
> The problem for me was that you introduce text for the error messages (maybe copied from description) and I saw improvements for the text

I copied the description if it was very unclear what the text should be but left out descriptions which are not very useful like "General Parse Error". So actually I have just restructured the content. Not deleted or added anything.

But alright, if it is your profound wish to do it in this PR I will do that, but please still review all of the categorizations and messages so that I can do it in one go and not for every 10 errors.

src/include/kdbmacros.h
markus2330: > removing adding "sorry"

This is not a good example as on this topic there are papers.

> I will not do that again.

I would very much welcome if the quality of your PRs would improve even more.

markus2330:
> But alright, if it is your profound wish to do it in this PR I will do that, but please still review all of the categorizations and messages so that I can do it in one go and not for every 10 errors.

Ok, thank you. Tell me when all messages are ready.

src/include/kdbmacros.h
Piankero: I changed the text to include the filename. It now gives a message as following: ```Resource: Insufficient permissions to open configuration file <path> for reading. Reason: Permission denied. You might want to retry as root or change access using chmod.```

markus2330: Thank you! Btw. the solution you do not need to add in the text as it will be added differently later. (Otherwise we mind and up with many duplicates of the text "You might want to retry...".)

Only make sure that everything visible in the PR has very good quality.

src/include/kdbmacros.h
Piankero: > but it was tried ... (because it did not succeed?)

Actually after thinking twice about the sentence I am not sure what exactly you mean. Reading through the code actually shows me that setting a value actually succeeded but shouldn't?

src/error/exporterrors.cpp
markus2330: In ni, the header file is not reused outside of the plugin.

src/include/elektra/error.h
markus2330: Why is this change in the PR of @Piankero then?

src/error/exporterrors.cpp
markus2330: strncmp

src/include/elektra/error.h
kodebach: Because the changes to low-level errors required that the high-level errors API is changed as well.

src/error/specification
markus2330: Ok, good solution!

Do we need this "description:" in this file then at all?

src/include/kdbmacros.h
markus2330: No, these read-only parts are always read-only, it is not dependent on compilation or installation. It only depends on the specification, like other validation.

src/include/kdbmacros.h
markus2330: not supported but the key %s (value %s) was added" -> not supported but the key %s (value %s) was tried to be added"

src/plugins/blockresolver/blockresolver.c
```diff
@@ -367,7 +369,7 @@ int elektraBlockresolverSet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned E
 		fin = fopen (data->tmpFile, "r");
 		if (!fin)
 		{
-			ELEKTRA_SET_RESOURCE_ERRORF (parentKey, "Couldn't open %s for reading", data->tmpFile);
+			ELEKTRA_SET_RESOURCE_ERRORF (parentKey, "Couldn't open %s for reading Reason: %s", data->tmpFile, strerror (errno));
```
markus2330: Because I saw it just now: Please make sure to also have guidelines (also written down) of how the error texts should be formatted. E.g. should there be a dot after a sentence? Or a comma and continue "reason" with small letters? Or should it read `"for reading because of '%s'"`? And how to enclose strings so that it is clear where they start and where they end?

Your work now is the best opportunity to make this consistent.

If you see that macros would be helpful please introduce them, see also #2697. Otherwise guidelines will be essential.

Piankero: I added a short guideline. Please take a look if it fits the needs. https://github.com/ElektraInitiative/libelektra/pull/2686/commits/405167cbad8753f33f5bb6fae7eea250d3603a83

> If you see that macros would be helpful please introduce them
Honestly this would make coding in elektra even more frustrating for newcomers because it is even more to know which is not self explanatory but gets rejected in a new PR.
I also think error messages are in general too heterogeneous to put into a macro but if I see something I will add it.

doc/dev/error-message.md
```diff
@@ -0,0 +1,14 @@
+# Error message
+
+Elektra has a lot of developers and hence the error messages vary in how they are written.
+This guideline should unify the message format so that users will get consistent error messages.
+
+- All error messages start with a capital letter even if you reference a method name such as "KdbSet() not supported". The only exception are quoted messages at the start such as "'xxxx-xxxx' is an invalid syntax". Try to avoid variable strings at the start of the sentence ("%s"-strings) if possible because you cannot guarantee capital casing.
```
markus2330: What about also quoting method names? Naming them wrongly in the error messages does not seem right.

doc/dev/error-message.md
```diff
@@ -0,0 +1,14 @@
...
+- **Always** provide the errno if possible.
+- If you can provide an errno, write `"Reason: %s", strerror(errno)` at the end of the error message. Also end the preceding sentence with a dot. Example: `"XY failed. Reason: %s", strerror(errno)`. Alternatively you can also embed it into a standard text: `"XY failed because of %s", strerror(errno)`.
```
markus2330: We should not recommend to call strerror(errno) as it is not thread-safe. strerror_r should be used instead. It would be better if we can embed the strerror in a generic way, so that not every caller needs to worry about that.

doc/dev/error-message.md
```diff
@@ -0,0 +1,14 @@
...
+- The last sentence (or single sentence of your message if you just provide one) must not end with a dot. This should encourage users to continue reading
+if other messages appear.
+- If your message has multiple sentences, separate them with dots and start with a capital letter like you would do in the normal english language.
```
markus2330: I would also say that multiple short sentences should be preferred.

doc/dev/error-message.md
```diff
@@ -0,0 +1,14 @@
...
+- Avoid abbreviations because not everybody will know them (eg. "No Byte Order Mark found" instead of "No BOM found").
```
markus2330: What about abbreviations in ()?

doc/dev/error-message.md
Piankero: > It would be better if we can embed the strerror in a generic way, so that not every caller needs to worry about that.

This is already written in https://github.com/ElektraInitiative/libelektra/issues/2572

Piankero:
> Thank you for adding the guidelines! It makes sense to have separate PRs for that.

https://github.com/ElektraInitiative/libelektra/pull/2793

doc/dev/error-message.md
Piankero: > We should not recommend to call strerror(errno) as it is not thread-safe.

This seems a bit problematic: http://www.club.cc.cmu.edu/~cmccabe/blog_strerror.html
Always allocating and freeing memory just for an error message seems very cumbersome.
Also what you mentioned multiple times now is that elektra is not intended to be used with multi threading. So why now introduce thread safety just for error messages?

Piankero:
jenkins build libelektra

Piankero:
@markus2330 I corrected and standardized all error messages. If possible please look over everything and not partly. Thank you!

src/plugins/conditionals/conditionals.c
```diff
@@ -209,7 +209,7 @@ static CondResult evalCondition (const Key * curKey, const char * leftSide, Comp
 			}
 			if (elektraRealloc ((void **) &compareTo, endPos - rightSide) < 0)
 			{
-				ELEKTRA_SET_ERROR (87, parentKey, "Out of memory");
+				ELEKTRA_SET_OUT_OF_MEMORY_ERROR (parentKey, "Out of memory");
```
markus2330: Maybe we make ELEKTRA_SET_OUT_OF_MEMORY_ERROR to not accept arguments: or we pass what was resized with which size.

src/plugins/fcrypt/fcrypt.c
```diff
@@ -244,18 +249,21 @@ static int fcryptGpgCallAndCleanup (Key * parentKey, KeySet * pluginConfig, char
 		shredTemporaryFile (tmpFileFd, parentKey);
 		if (unlink (tmpFile))
 		{
-			ELEKTRA_ADD_WARNINGF (ELEKTRA_WARNING_FCRYPT_UNLINK, parentKey, "Affected file: %s, error description: %s", tmpFile,
-					      strerror (errno));
+			ELEKTRA_ADD_RESOURCE_WARNINGF (
+				parentKey,
+				"Failed to unlink a temporary file. WARNING: unencrypted data may leak! Please try to delete "
+				"the file manually. Affected file: %s, error description: %s",
```
markus2330: Is not according to our rules (should be Reason: instead of error description:)?

src/error/specification
markus2330: Still open: why would we need "description:" at all here?

tests/shell/check_error.sh
```diff
@@ -91,7 +91,7 @@ if is_plugin_available error; then
 	#
 	#CONTENT=`cat $TMPFILE`
```
markus2330: This is still open (can be done in separate PR)

src/plugins/wresolver/wresolver.c
```diff
@@ -163,21 +164,21 @@ static void elektraResolveDir (resolverHandle * p, Key * warningsKey)
...
 			       NULL);
-		ELEKTRA_ADD_WARNINGF (ELEKTRA_WARNING_NOCWD, warningsKey, "GetCurrentDirectory failed: %s, defaulting to /", buf);
+		ELEKTRA_ADD_RESOURCE_WARNINGF (warningsKey, "GetCurrentDirectory failed: %s, defaulting to /", buf);
 		dir[0] = 0;
 	}
 	else if (dwRet > MAX_PATH)
 	{
-		ELEKTRA_ADD_WARNINGF (ELEKTRA_WARNING_NOCWD, warningsKey, "GetCurrentDirectory failed, buffer size too small, needed: %ld",
-				      dwRet);
+		// TODO: Solution? Compile with higher MAX_PATH?
```
markus2330: The solution is to submit a bug in the distribution's bug tracker to do that?

src/plugins/type/type.c
```diff
@@ -109,14 +109,14 @@ bool elektraTypeValidateKey (Plugin * handle, Key * key, Key * errorKey)
...
-		ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_TYPE, errorKey, "Unknown type '%s' for key '%s'", typeName, keyName (key));
+		ELEKTRA_SET_VALIDATION_SEMANTIC_ERRORF (errorKey, "Unknown type '%s' for key '%s'", typeName, keyName (key));
 		return false;
 	}
 
 	if (type->normalize != NULL && !type->normalize (handle, key))
 	{
-		ELEKTRA_SET_ERRORF (ELEKTRA_ERROR_TYPE, errorKey, "The value '%s' of key %s could not be normalized (type is '%s')",
-				    keyString (key), keyName (key), typeName);
+		ELEKTRA_SET_VALIDATION_SEMANTIC_ERRORF (errorKey, "The value '%s' of key '%s' could not be normalized (type is '%s')",
```
markus2330: Isn't this a syntactic problem? (Some string is in a form that it cannot be normalized?)

src/plugins/shell/shell.c
```diff
@@ -70,15 +70,16 @@ int elektraShellGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned, Key * pa
 		int retVal = executeCommand (keyString (cmdKey));
 		if (retVal == -1)
 		{
-			ELEKTRA_SET_ERRORF (144, parentKey, "launching childprocess failed with %s\n", strerror (errno));
+			ELEKTRA_SET_PLUGIN_MISBEHAVIOR_ERRORF (parentKey, "Launching childprocess failed. Reason: %s", strerror (errno));
```
markus2330: I think these are resource errors (forking process usually fails because of too little resources).

src/plugins/jni/jni.c
```diff
@@ -218,14 +218,14 @@ int elektraJniOpen (Plugin * handle, Key * errorKey)
 	elektraFree (classpath);
 	if (res < 0)
 	{
-		ELEKTRA_SET_ERROR (102, errorKey, "Cannot create Java VM");
+		ELEKTRA_SET_INSTALLATION_ERROR (errorKey, "Cannot create Java VM");
```
markus2330: This could basically have any reason. Doesn't JNI give us some hints?

src/plugins/fcrypt/fcrypt.c
```diff
@@ -127,7 +127,10 @@ static int shredTemporaryFile (int fd, Key * errorKey)
 
 	if (fstat (fd, &tmpStat))
 	{
-		ELEKTRA_SET_ERROR (ELEKTRA_ERROR_FCRYPT_TMP_FILE, errorKey, "Failed to retrieve the file status of the temporary file.");
+		ELEKTRA_SET_RESOURCE_ERRORF (
+			errorKey,
+			"Failed to overwrite the temporary data. Cannot retrieve file status Unencrypted data may leak. Errno: %s",
```
markus2330: Sentence? Why _U_nencrypted?

src/plugins/conditionals/conditionals.c
Piankero: I have thought about that too since since there is already a macro in [kdbmacros.h](https://github.com/ElektraInitiative/libelektra/blob/d3d98a8216430e40bdcc09004c668f49e38f3e92/src/include/kdbmacros.h#L64). I don't think though that does yield a considerable benefit (consistency, additional information) but might confuse developers that all macros behave identically but OUT_OF_MEMORY suddenly does not.

tests/shell/check_error.sh
Piankero: I remember that you wanted me to 'cleanup' the shell script but how exactly do you want me to do that? Should I remove the commented code?

src/plugins/wresolver/wresolver.c
Piankero: I will change the text that either the user/dev can compile it himself or issue a bug so we do it for him in the main repo. If a single has a hilarious long path it might not be sensible to change the default.

Piankero:
> Still open: why would we need "description:" at all here?

I think this is much more readable if you look at the file compared to the bulky upper cased macros with underscores. I can easily remove it if you wish though.

src/plugins/conditionals/conditionals.c
markus2330: Yes, if OUT_OF_MEMORY would be the only one it might be confusing. But I think it is okay in general for macros of different categories to behave differently (or to have specializations that behave differently, see #2697). To pass an argument, that does not provide any information, does not make sense. As alternative we could pass as argument how much memory we tried to allocate: but this is a lot of work with little benefit.

tests/shell/check_error.sh
markus2330: The idea is of course to make the commented-out code work again.

markus2330:
Thank you for the explanation, it is okay as is.

PR #2691, opened 2019-05-10
Include Common: Fix Broken Shell Variables

sanssecours:


markus2330:
Thank you for the fix!

PR #2704, opened 2019-05-18
Error message implementation

Piankero:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

## Merging

### DIFF
--- a/doc/decisions/error_message_format.md
+++ b/doc/decisions/error_message_format.md
@@ -53,12 +53,12 @@ Configfile: ...../<file>.25676:1549919217.284067.tmp
 The new default message will look like this:
 
 ```
-Sorry, module <MODULE> issued [error|warning] code <NR>:
-<ERROR_CODE_DESCRIPTION>: Validation of key "<key>" with string "<value>" failed.
+Sorry, module `MODULE` issued [error|warning] `NR`:
+`ERROR_CODE_DESCRIPTION`: Validation of key "<key>" with string "<value>" failed.
 ```
 
-The <NR> will be the color red in case of an error or yellow in case of a warning
-while <MODULE> will be the color blue.
+The `NR` will be the color red in case of an error or yellow in case of a warning
+while `MODULE` will be the color blue.
 
 Optionally a third line indicating a solution can be added. Eg. for a permission related error there would be a third line:
 
--- a/doc/decisions/error_message_format.md
+++ b/doc/decisions/error_message_format.md
@@ -53,12 +53,12 @@ Configfile: ...../<file>.25676:1549919217.284067.tmp
 The new default message will look like this:
 
 ```
-Sorry, plugin <MODULE> issued [error|warning] code <NR>:
+Sorry, module <MODULE> issued [error|warning] code <NR>:
 <ERROR_CODE_DESCRIPTION>: Validation of key "<key>" with string "<value>" failed.
 ```
 
 The <NR> will be the color red in case of an error or yellow in case of a warning
-while <PLUGIN> will be the color blue.
+while <MODULE> will be the color blue.
 
 Optionally a third line indicating a solution can be added. Eg. for a permission related error there would be a third line:
 
--- a/doc/help/kdb-set.md
+++ b/doc/help/kdb-set.md
@@ -30,7 +30,9 @@ To set a key to a negative value, `--` has to be used to stop option processing.
 - `-C`, `--color <when>`:
   Print never/auto(default)/always colored output.
 - `-v`, `--verbose`:
-  Explain what is happening.
+  Explain what is happening. Also shows Configfile and Mountpoint in case of an error/warning
+- `-d`, `--debug`:
+  Shows the line at which an error happened in case an error or warning is issued
 - `-q`, `--quiet`:
   Suppress non-error messages.
 - `-N`, `--namespace=NS`:
--- a/doc/help/kdb-set.md
+++ b/doc/help/kdb-set.md
@@ -30,9 +30,7 @@ To set a key to a negative value, `--` has to be used to stop option processing.
 - `-C`, `--color <when>`:
   Print never/auto(default)/always colored output.
 - `-v`, `--verbose`:
-  Explain what is happening. Also shows Configfile and Mountpoint in case of an error/warning
-- `-d`, `--debug`:
-  Shows the line at which an error happened in case an error or warning is issued
+  Explain what is happening.
 - `-q`, `--quiet`:
   Suppress non-error messages.
 - `-N`, `--namespace=NS`:
--- a/doc/help/kdb.md
+++ b/doc/help/kdb.md
@@ -48,9 +48,11 @@ Every core-tool has the following options:
 Most tools have the following options:
 
 - `-v`, `--verbose`:
-  Explain what is happening.
+  Explain what is happening. Also shows Configfile and Mountpoint in case of an error/warning
 - `-q`, `--quiet`:
   Suppress non-error messages.
+- `-d`, `--debug`:
+  Shows the line at which an error happened in case an error or warning is issued
 
 ## KDB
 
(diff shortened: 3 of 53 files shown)


Piankero:
I added a first implementation of the new error message format for you to review (test case fixes excluded). I especially have a question concerning these lines:
https://github.com/Piankero/libelektra/blob/error-message-implementation/src/tools/kdb/main.cpp#L231-L234
Is there a better way to pass down those parameters? 

Also there are some cases where I cannot pass down the command line options but take them as "true":
https://github.com/Piankero/libelektra/blob/error-message-implementation/src/bindings/cpp/include/kdbexcept.hpp#L56-L57

doc/decisions/error_message_format.md
```diff
@@ -53,12 +53,12 @@ Configfile: ...../<file>.25676:1549919217.284067.tmp
...
 
 ```
-Sorry, plugin <MODULE> issued [error|warning] code <NR>:
+Sorry, module <MODULE> issued [error|warning] code <NR>:
 <ERROR_CODE_DESCRIPTION>: Validation of key "<key>" with string "<value>" failed.
 ```
 
 The <NR> will be the color red in case of an error or yellow in case of a warning
-while <PLUGIN> will be the color blue.
+while <MODULE> will be the color blue.
```
markus2330: This might be misinterpreted as html in markdown. Please write as `code`.

markus2330:
> Is there a better way to pass down those parameters?

You can also simply overload what. C++ is like Java in this regards. Additionally, you can let it even default values to true, true (then you have a single implementation, unlike Java).

> Also there are some cases where I cannot pass down the command line options but take them as "true":

At least this situation could be avoided with default parameters.

Piankero:
Thank you, I will implement it with the default method as you suggested.
@sanssecours: Do you have an idea why exactly a single test case fails on a special debian distribution? These kind of errors are so weird to me

Piankero:
@omnidan: I have a short question concerning the web ui. The error message format has changed and I tried to fix it in https://github.com/ElektraInitiative/libelektra/pull/2704/commits/66e4161e33a1ad915ca92f61f7c8ae8a9fa4e69f for the gui.
Is there anything else I need to consider? I haven't found the usage of "parseError" in my IDE.

As a short introduction to you:
The old format
``
The command kdb set failed while accessing the key database with the info:
Sorry, the error (#121) occurred ;(
Description: validation failed
Reason: Validation of key "<key>" with string "<value>" failed.
Ingroup: plugin
Module: enum
At: ....../src/plugins/enum/enum.c:218
Mountpoint: <parentKey>
Configfile: ...../<file>.25676:1549919217.284067.tmp
```

will change to this one (just an example)

```
Sorry, module `MODULE` issued [error|warning] <NR>:
`ERROR_CODE_DESCRIPTION`: Validation of key "<key>" with string "<value>" failed.
```
`At`, `Mountpoint` and `Configfile` can optionally be added via cmdline parameters

**Comment by markus2330** [comment]
> At, Mountpoint and Configfile can optionally be added via cmdline parameters

In the Web-UI it obviously should be a button. Such a button already exists.

**Comment by Piankero** [comment]
> Additionally, you can let it even default values to true, true (then you have a single implementation, unlike Java).

If I do that the compiler complains about overloaded-virtual.

> In the Web-UI it obviously should be a button. Such a button already exists.

Then it should work as before because I did not change these lines

**Comment by markus2330** [comment]
>     If I do that the compiler complains about overloaded-virtual.

https://stackoverflow.com/questions/18515183/c-overloaded-virtual-function-warning-by-clang

> Then it should work as before because I did not change these lines

But please test if the error messages now look nicely. As you most likely already know, you can easily trigger errors via the error-trigger plugin, which is called "error".

**Comment by Piankero** [comment]
> stackoverflow.com/questions/18515183/c-overloaded-virtual-function-warning-by-clang

I have read this post but the problem ist that there is an ambiguous declaration then because what() is also inherited from Exception and if no parameters are given the compiler does not know if it should take the what() function or what(arg1, arg2) function with default params. I even tried to hide the exception what() by doing 
```
private:
	using Exception::what;
```

**Comment by Piankero** [comment]
> But please test if the error messages now look nicely. As you most likely already know, you can easily trigger errors via the error-trigger plugin, which is called "error".

web-ui-error
Looks good to me!

**Comment by markus2330** [comment]
> function or what(arg1, arg2) function with default params

Of course you cannot have default parameters and a second function without parameters at the same time. What about:

- having a what without args, calling the other what with (true, true)
- having a second what with two args, one default arg true

having a differently called "what*" is also okay. Duplicating the implementation, however, is not ok.

> Looks good to me!

Of what use is a "details" which shows the same information?

- "KDBError" is not informative
- "Command failed..." I do not think this is relevant for the average user
- Error Details: here I would expect all details (-v and -d) and which command the GUI tried to execute

**Comment by Piankero** [comment]
New Web-UI error now looks as follows:
no-detail
and by clicking on detail:
with-detail

**Comment by Piankero** [comment]
I also removed some indeed unused code in the web ui. I hope we can now merge this PR @markus2330

**Comment by markus2330** [comment]
This pull request **fixes 1 alert** when merging 33962c06a89291ffe216f45fee680175bfddd479 into 4a3c1634be7a732b00aacc9c5ed6dc46d33652c2 - [view on LGTM.com](https://lgtm.com/projects/g/ElektraInitiative/libelektra/rev/pr-75545e6e14ba7d5269b83d3c6949c99e504fdc74)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

File: doc/help/kdb-set.md [hunk 2]
```diff
@@ -30,7 +30,9 @@ To set a key to a negative value, `--` has to be used to stop option processing.
 - `-C`, `--color <when>`:
   Print never/auto(default)/always colored output.
 - `-v`, `--verbose`:
-  Explain what is happening.
+  Explain what is happening. Also shows Configfile and Mountpoint in case of an error/warning
+- `-d`, `--debug`:
```
markus2330: You do not need to document this for every command individually. Please document it once in doc/help/kdb.md

Please also make sure that the behavior when errors/warnings are shown is identical for every program (with tests).

Piankero: > Please also make sure that the behavior when errors/warnings are shown is identical for every program (with tests).

What is "every program"?
Also which kind of tests do you think of? shell recorder?

markus2330: > What is "every program"?

Web-UI you already showed. (Thank you, looks very nice) Then also manual tests for qt-gui. And then for `kdb` suite (complete convert cp editor export file find fstab get globalmount import ls metaget metals metaremove metaset mount mv remount rm set sget (should never print an error) shell specmount umount validation). For merge tools you do not need to make tests, as they are reworked now anyway.

> Also which kind of tests do you think of? shell recorder?

For kdb tools shell recorder. web-ui and qt-gui manual tests are enough, as automation of these tests would be too much effort.

markus2330: Snippet-conversion (our web-page) a manual test is also enough.

Piankero: Ok, do you know an easy way how to trigger an error/warning for each of these commands? Or is this case dependent?

markus2330: With the trigger/error (error plugin) you should be always able to trigger an error (even when only getting config). Maybe you can write a single loop to test most of the programs but then you cannot use shell recorder. I think its better if you start with simple individual shell-recorder tests and then progress to a loop similar to tests/shell/check_import.sh but iterating over commands and not plugins.

Btw. please bring also tests/shell/check_error.sh up-to-date.

Piankero:
qt-gui manual error format: 
qt-gui-error
Absolutely fine imo.

markus2330:
This pull request **fixes 1 alert** when merging 38980e8a12cc9e64f937cae8cd41692b6978dc9e into efa5ec48f77cf73a64c2d6b67e29ca09d7eb57ac - [view on LGTM.com](https://lgtm.com/projects/g/ElektraInitiative/libelektra/rev/pr-569190a72b09517dafe7a07db0ca535042c68805)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

markus2330:
> Absolutely fine imo.

The problem is that the standard-text "Synchronized failed" is completely unrelated to the actual error. It would be nice to have the same behavior as in Web-UI.

markus2330:
Thank you very much!

qt-gui doesn't have the highest priority and also would make more sense in a different PR. Can you create an issue for that? (Maybe the initial author is interested in doing this.)

PR #2711, opened 2019-05-23
Yan LR: Convert Empty File to Empty Parent Key

sanssecours:


markus2330:
Great work! Did you already update the tutorial?

Don't you think it is already time to only update the best plugin? Is your comparison between the plugins finished? I'd love to read it. (I am curious if I will come to the same conclusion.)

sanssecours:
> Did you already update the tutorial?

I do not know, if converting an empty file to an empty key makes sense for all file formats. It certainly does in the case of YAML, since the following:

```yaml

```

this:

```yaml
~
```

, and this

```yaml
null
```

describe the same content (“nothing”).

> Don't you think it is already time to only update the best plugin?

I think this is the last time I update all plugins. I only wanted to make sure, that the comparison of the YAML output for an empty file is the same.

> Is your comparison between the plugins finished?

Nope. I am still benchmarking the runtime of the plugins. The graph [here](https://github.com/ElektraInitiative/libelektra/files/3216153/BenchmarkResultGenerated.pdf) shows the runtime of the plugins for the first 1, 5, 10, 50, 100, 500, 1000, 5000, 10000, 50000 and 100000 lines of a generated YAML file. As you can see the runtime of the plugins on Linux and macOS is sometimes quite different. All of the plugins show the expected linear behavior for a line number of about 1000 lines and higher. The constant difference for a file can be relatively large though (~ 5), as the graph [here](https://github.com/ElektraInitiative/libelektra/files/3216181/BenchmarkResultGeneratedBarGraph.pdf) – for another generated file with 10 000 lines – shows.

markus2330:
> I do not know, if converting an empty file to an empty key makes sense for all file formats.

I do not think we need to have the claim that the tutorial works for every format. (The KeySet itself, however, does make the claim.) As YAML will be our main format, the documentation for the mapping between KeySet and YAML is also the most important.

>  I think this is the last time I update all plugins. I only wanted to make sure, that the comparison of the YAML output for an empty file is the same.

Ok!

> Nope. I am still benchmarking the runtime of the plugins.

Do not overbenchmark. Runtime is only one criteria of many and not even the most important one.

> The graph here

Amazing graphs.

PR #2734, opened 2019-05-30
Reformat All: Support Spaces Inside Format Script Paths

sanssecours:


markus2330:
Thank you! Do we still don't have a build server which has spaces in the source/build dir?

sanssecours:
> Do we still don't have a build server which has spaces in the source/build dir?

Yes, we do:

https://github.com/ElektraInitiative/libelektra/blob/092648cfa2fc89e860efc12b2b3f1a47a06875a2/scripts/jenkins/Jenkinsfile#L854

. However, the build server does not use `reformat-all`. Another problem is that only the build directory path contains spaces, while the filepath of the cloned repository does not.

Issue #2727, opened 2019-05-26
keyIsBelow and friends


kodebach:
A simple question: Should `keyIsBelow` and friends work, if the first argument is a cascading key?

markus2330:
I think it safe to say "no" as it is not specified how it should work and there are hardly any tests for it. Some support was implemented by @tom-wa in 36e0ed568689a5e0f9778bb44d2c8621912a423f and 52db645bf456f15200a441a14aa0ec045c1799ef (iirc, the commit message does not say much). But as you know, the spec plugin is quite buggy and that keyIsBelow is not specified properly is one of the main reasons.

kodebach:
Why is the code even so complicated? Shouldn't it be a simple `memcmp` for the non cascading case? After all we are just doing a prefix check. If `key` is cascading it should be enough to ignore the first part of both keys. So the following should work:

```c
int keyIsBelow (const Key * key, const Key * check)
{
	const void * above = keyUnescapedName (key);
	const void * below = keyUnescapedName (check);
	
	size_t size1 = keyGetNamespace (key);
	size_t size2 = keyGetNamespace (check);

	size_t size = size1 < size2 ? size1 : size2;

	if(above[0] == '\0')
	{
		// cascading
		below = strchr (below, '\0');
	}
	
	return memcmp (above, below, size) == 0;
}
```


Or am I missing something here?

markus2330:
> Why is the code even so complicated?

It was written before unescaped names existed and it was not simplified after it was introduced.

I fully support your new implementation, please try in a PR if it works.

> just noticed the above code would be for keyIsBelowOrSame

It does not matter: then simply call keyIsBelowOrSame within keyIsBelow.

markus2330:
(In the PR it is also easier to give feedback to the individual lines.)

kodebach:
I just found `testabi_key.c`, it seems the behaviour `keyIsBelow` is indirectly defined by those tests. But I am not sure these tests are correct. They contain for example:

https://github.com/ElektraInitiative/libelektra/blob/54a4c0194946917b7d093e0777f465619b2f3d6f/tests/abi/testabi_key.c#L1362-L1365

markus2330:
> But I am not sure these tests are correct.

Nobody is because there is no specification how it should be. (cascading as "match all" vs. cascading as normal namespace).

[...]

Issue #2455, opened 2019-03-02
optimize ksAppend()


mpranj:
## Motivation
When `kdbGet()` has a cache hit but the supplied keyset was not empty, we need to append the cached keyset to the non-empty keyset. In cases where the appended keyset is much larger than the destination it could be beneficial to append the smaller keyset to the larger one. (less ksAppendKey operations).

## Proposal
Add a `ksMerge()` function, which returns only one resulting keyset and `ksDel()`etes the other keyset. This is for the cases, where the other keyset would be thrown away anyway.

markus2330:
Thank you for writing the proposal!

> When kdbGet() has a cache hit but the supplied keyset was not empty

The same situation occurs when the command-line arguments are appended to proc/ in the global plugin done by @kodebach, see #2317 

> In cases where the appended keyset is much larger than the destination

For this specific situation, I think an `if` that swaps the arguments of `ksAppend` in the user code is acceptable.

I would like to avoid to have many API calls that appear to be highly optimized but actually only contain trivial (slow) operations. I would very much see optimizations for `ksAppend`. In its current form, it was maybe a mistake that we provide `ksAppend`.

mpranj:
I agree that improving ksAppend is more important than adding another API function.

> For this specific situation, I think an if that swaps the arguments of ksAppend in the user code is acceptable.

The problem is that the semantics of ksAppend do not allow for this optimization. ksAppend has to append source to destination. Swapping the arguments would not result in an equivalent output, because we must not append destination to source.

Therefore the idea only works for a new API function, that only promises to merge two sets to one, and throws away the other set.

kodebach:
> I would very much see optimizations for ksAppend

Two things that I think are quite simple to implement, but might make a huge difference:

1. `ksAppendKey` calls `ksSearchInternal`, which does a complete binary search. This can be optimized by first comparing against the last Key and immediately returning, if the new Key has to be inserted at the end. This would make `ksAppendKey` calls much faster, when Keys are inserted in the correct order. (Currently we do `log(ksSize)` Key comparisons to arrive at this result.)

2. `ksAppend(ks1, ks2)` should basically operate like this:
     - Create an array of the correct size
     - Iterate over both `ks1->array` and `ks2->array`, always appending the "smaller" (according to `keyCmp`) Key to the new array and advancing the corresponding iterator.
    - Replace the array of `ks1` with the new one.

[...]

PR #2736, opened 2019-05-30
Spec improvements

### LINKED ISSUE #2634: spec: globbing problems
The `spec` plugin doesn't do globbing correctly: 

- The `#` glob should query the array size and should copy metadata to non-existent keys within the array bounds as well. Otherwise something like `kdb get array/#0` won't work, even if the specification contains a default value for `array/#`. Setting the `fallback` metadata might also be a solution.
- The same problem exists for `_`, however, I don't really see a solution there, because you can't create all possible keys.

Where possible, the `spec` plugin should also use the `globbing` library.

### LINKED ISSUE #2700: spec: does not remove copied metadata before writing config
I think this is a show-stopper for using in LCDproc or machinekit (who want INI as first step for smooth migration)

- mount a specification
- mount an INI file
- do a `kdb set` 

```sh
kdb mount wrong-copy.ini /tests ini
kdb setmeta /tests/some-key some-meta some-value
kdb set /tests/some-key x
```

## Expected Result

To have a file

```ini
some-key = x
```

## Actual Result

```ini
#@META some-meta = some-value
some-key = x
```

## System Information

- Elektra Version: master

## Implementation Hints

The spec plugin does not have the positioning to remove its meta-data before writing to the storage. See discussion below: the KeySet passed back from `kdbSet` **must not** be altered. Only the meta-data within `split` is allowed to be removed.

So the easiest way is to call the spec plugin for every part in the split plugin separately. (As foreach plugin.)

kodebach:
**NOTE:** Technically there are some breaking changes in this PR. However since, every breaking change that I know of concerns parts of `spec` that can already be considered broken, I don't think that is a problem.

---

Closes #2634 (only array part, new issue for _ needed)

Also improves `ksIsBelow` and friends (see #2727) and `ksSearchInternal` (see #2455). Changing `ksAppend` is more complicated, because we need to make sure the OPMPHM is not invalidated, when we append a subset of the current KeySet (i.e. no new keys).

`elektraUnescapeKeyName` now only does the full escaping procedure, if needed and `ksCut` utilises `ksSearchInternal` to find the start point of the cut.

## Basics

## Checklist

- [x] I added unit tests
- [x] meta data is updated (e.g. README.md of plugins and doc/METADATA.md)

## Review

### DIFF
--- a/doc/AUTHORS.md
+++ b/doc/AUTHORS.md
@@ -195,3 +195,11 @@ development of the mmapstorage plugin
 - email: mpranj@limun.org
 - github user: [mpranj](https://github.com/mpranj)
 - devel/test on: Fedora, Debian, macOS
+
+## Klemens Böswirth
+
+development of the highlevel API and code-generation; various other things
+
+- email: k.boeswirth+git@gmail.com
+- github user: [kodebach](https://github.com/kodebach)
+- devel/test on: Fedora
--- a/src/libs/elektra/internal.c
+++ b/src/libs/elektra/internal.c
@@ -715,6 +715,21 @@ size_t elektraUnescapeKeyName (const char * source, char * dest)
 
 	ELEKTRA_ASSERT (sp != NULL && dp != NULL, "Got null pointer sp: %p dp: %p", (void *) sp, (void *) dp);
 
+	// if there is nothing to unescape, just make a copy and replace / with \0
+	if (strpbrk (sp, "\\%") == NULL)
+	{
+		strcpy (dest, sp);
+		char * last = dp;
+		while ((dp = strchr (dp, '/')) != NULL)
+		{
+			*dp = '\0';
+			++dp;
+			last = dp;
+		}
+		// add 1, if we didn't end with \0 already
+		return last - dest + strlen (last) + (*last != '\0');
+	}
+
 	if (*sp == '/')
 	{
 		// handling for cascading names
--- a/src/plugins/spec/README.md
+++ b/src/plugins/spec/README.md
@@ -1,75 +1,137 @@
 - infos = Information about the spec plugin is in keys below
-- infos/author = Thomas Waser <thomas.waser@libelektra.org>
+- infos/author = Thomas Waser <thomas.waser@libelektra.org>, Klemens Böswirth <k.boeswirth+git@gmail.com>
 - infos/licence = BSD
 - infos/needs =
 - infos/provides = check apply
 - infos/placements = postgetstorage presetstorage
-- infos/status = maintained nodep configurable global preview
-- infos/description = copies metadata from spec namespace to other namespaces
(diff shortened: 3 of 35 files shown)


markus2330:
This pull request **introduces 4 alerts** when merging a4437ed459052d348509f1198aed2c0cd7fd6e25 into 5f7f855c79753f734946ef1a964570cba311e189 - [view on LGTM.com](https://lgtm.com/projects/g/ElektraInitiative/libelektra/rev/pr-f9e7e4689a2c91ceb9a0913a704020ee6d92f8c7)

**new alerts:**

* 4 for Array argument size mismatch

---

kodebach:
I decided not to fix #2700 in this PR, because that needs more considerations (see https://github.com/ElektraInitiative/libelektra/issues/2700#issuecomment-497977597).

doc/AUTHORS.md
```diff
@@ -195,3 +195,11 @@ development of the mmapstorage plugin
 - email: mpranj@limun.org
 - github user: [mpranj](https://github.com/mpranj)
 - devel/test on: Fedora, Debian, macOS
+
+## Klemens Böswirth
+
+development of the highlevel API and code-generation; various other things
```
markus2330: Good that you finally add yourself! :heartbeat:

src/libs/elektra/internal.c
```diff
@@ -715,6 +715,21 @@ size_t elektraUnescapeKeyName (const char * source, char * dest)
 
 	ELEKTRA_ASSERT (sp != NULL && dp != NULL, "Got null pointer sp: %p dp: %p", (void *) sp, (void *) dp);
 
+	// if there is nothing to unescape, just make a copy and replace / with \0
+	if (strpbrk (sp, "\\%") == NULL)
```
markus2330: Is this a performance improvement? Why not put it into a separate function?

src/plugins/spec/README.md
```diff
@@ -1,76 +1,138 @@
...
+### Array Size Validation
+
+As hinted to above, we validate array sizes. If a spec key `x/#` is given, and the spec key `x` has the metakey `array/min` or `array/max`
+set, we validate the array size (given as metakey `array` on `x`) is within the limits of `array/min` and `array/max`. Both `array/min` and
+`array/max` have to be valid array-elements similar to `array`. If the array size is out of bounds, this causes a `range` conflict.
+
+We also check that the array only contains actual array elements. If this not the case, this causes a `member` conflict.
 
-### Per Key Configuration
+Note: We don't actually validate that the array doesn't contain elements above the given array size. This is because it doesn't have anything
```
markus2330: Good that you wrote the note, this is unexpected.

src/libs/elektra/internal.c
kodebach: Yes, this should improve performance. The function was so short already, it didn't seem necessary to split it up.

Issue #2561, opened 2019-03-29
kdb set user/ and -N user should behave the same, -f for old behavior


markus2330:
@sanssecours Can you maybe help define the behavior here?

sanssecours:
If you mean that something like

```sh
kdb get user/tests/get
```

should behave the same as

```sh
kdb get -N user /tests/get
```

, then I think that sounds reasonable. What exactly is the old behavior? As far as I can tell `kdb get` does not support an argument `-N` yet:

```sh
kdb 2>&1 get -N
#> kdb: invalid option -- N
#> Sorry, I could not process the given options (see errors above)
#> 
#> Usage: kdb get <name>
#> 
#> Get the value of an individual key.
#> When the key starts with / a cascading lookup will be done.
#> 
#> Example:
#> 
#>    kdb get system/elektra/version/constants/KDB_VERSION
#> 
#> -a --all                 Consider all of the keys.
#> -H --help                Show the man page.
#> -n --no-newline          Suppress the newline at the end of the output.
#> -p --profile <name>      Use a different profile for kdb configuration.
#> -v --verbose             Explain what is happening.
#> -V --version             Print version info.
#> -C --color <when>       Print never/auto(default)/always colored output.
```

.

markus2330:
> If you mean that something like

exactly, but I meant `kdb set`

> What exactly is the old behavior?

That one way supports the validation and the other does not.

sanssecours:
> That one way supports the validation and the other does not.

Yes, it make sense that in the following situation:

```sh
kdb mount tutorial.dump /tests/spec dump validation
kdb setmeta spec/tests/spec/test check/validation '[1-9][0-9]*'
kdb setmeta spec/tests/spec/test check/validation/match LINE
kdb setmeta spec/tests/spec/test check/validation/message 'Not a number'
```

, the command

```sh
kdb set user/tests/spec/test 'incorrect value'
```

fails, just as

```sh
kdb set -N user /tests/spec/test 'incorrect value'
```

and

```sh
kdb set /tests/spec/test 'incorrect value'
```

do. Using the option `-f` (force) to change the value, even if the validation fails sounds like a good choice.

markus2330:
Exactly! Thank you for the description of the error. Maybe we should also add `-N` to `kdb get` so that the same syntax can be used for every `kdb` command.

markus2330:
Maybe we can fix this together with #2724. Then scripts do not need to be adapted twice.

[...]

PR #2746, opened 2019-06-04
Add example how to test rest api on localhost

dmoisej:
## Basics

## Checklist

need to be checked.

- [ ] I added unit tests

## Review

## Merging


markus2330:
Thank you very much for improving the docu!

Can you maybe use `curl` for the example, so that your examples can be actually executed?

dmoisej:
@markus2330 yes, no problem, will add either today evening or tomorrow.

dmoisej:
@markus2330 could you please help me and tell why some of the checks have failed? I just added changes to documentation and the new line to the changelog. Did I miss something?

markus2330:
The markdown formatting check fails, see: https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2746/2/pipeline/223

@sanssecours is there really no way to turn off code-formatting inside of Markdown? The suggested change for the JSON is ugly.

sanssecours:
> @sanssecours is there really no way to turn off code-formatting inside of Markdown?

You can disable the formatting using the tags `<!-- prettier-ignore-start -->` and `<!-- prettier-ignore -->` as specified in [`CODING.md`](https://github.com/ElektraInitiative/libelektra/blob/master/doc/CODING.md#prettier).

> The suggested change for the JSON is ugly.

I think the problem here is that `prettier` treats the JSON content as normal text and not as code, which is correct according to the [Markdown specification](https://daringfireball.net/projects/markdown/syntax). The HTML output below shows the rendering of the JSON content using GitHub’s Markdown converter.

---

{
    "exists": true,
    "name": "test",
    "path": "user/test",
    "ls": [
        "user/test"
    ],
    "value": "5",
    "meta": ""
}

---

If you use a [fenced code block](https://help.github.com/en/articles/creating-and-highlighting-code-blocks), then `prettier` will format the code correctly and GitHub’s Markdown renderer will also produce nicer looking output.

---

```json
{
  "exists": true,
  "name": "test",
  "path": "user/test",
  "ls": ["user/test"],
  "value": "5",
  "meta": ""
}
```

---

markus2330:
CODING.md does not mention that some code within Markdown also gets reformatted and it seems like that people do not expect this. Can you write the tip you gave here into CODING.md? (Btw. we should also consider to split up CODING.md. The instructions how to install the tools somehow do not belong there.)

sanssecours:
> Can you write the tip you gave here into CODING.md?

The coding guidelines already specify that you should use fences for code blocks:

https://github.com/ElektraInitiative/libelektra/blob/a03fdfddd540552f41b9f8496f4bd10c1b0a65f7/doc/CODING.md#L373-L374

.

markus2330:
yes, but in which cases code might be reformatted is not explained. Or is it not reformatted in fences and fences is the only allowed option?

sanssecours:
> yes, but in which cases code might be reformatted is not explained. Or is it not reformatted in fences and fences is the only allowed option?

As far as I can tell, `prettier` only changes code sections, if

- you specify the language of a code block, and
- it knows how to parse the specified language

.

dmoisej:
@markus2330 I updated the example with using curl and I used [fenced code block](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) to show the result of the request.

markus2330:
@sanssecours can you maybe help out here?

src/tools/web/README.md
```diff
@@ -114,6 +114,43 @@ Elektra web consists of multiple components:
...
+will return the value of path key, which is stored in the database.
+
+```sh
+kdb set /test 5
+```
+
+curl http://localhost:33333/kdb/user/test
+will return 
+
+```json
```
sanssecours: Is this the exact output from `curl`? If it is, then please add the comment `<!-- prettier-ignore-start -->` before the code block and `<!-- prettier-ignore-end -->` after the code block. For an example, please take a look at the ReadMe of Yan LR:


https://github.com/ElektraInitiative/libelektra/blob/733643dd00054f2e93fd09303fca9bd35c8e9fcd/src/plugins/yanlr/README.md#L228-L234

.

src/tools/web/README.md
```diff
@@ -114,6 +114,43 @@ Elektra web consists of multiple components:
...
+- `npm start` (replaces `kdb run-elektrad`)
+
+Now the server is runing on [http://localhost:33333](http://localhost:33333). After that you can test API with help of Postman or other tool, which allows to send REST API requests.
+
+As an example:
+
+GET: http://localhost:33333/kdb/path
+
+will return the value of path key, which is stored in the database.
+
```
sanssecours: I think part of an explanatory text is missing here. How about something like:

> After you changed the value `user/test` with the command
>
> ```sh
> kdb set user/test 5
> ```
>
> , the command
>
> ```sh
> curl http://localhost:33333/kdb/user/test
> ```
> 
> will return …

.

src/tools/web/README.md
```diff
@@ -114,6 +114,43 @@ Elektra web consists of multiple components:
...
+```sh
+kdb set /test 5
```
sanssecours: It might make sense to use `user/test` instead of `/test` here. This will make sure someone who reads this text does not have to know how cascading keys work.

markus2330: Once #2561 is fixed this should be identical anyway.

dmoisej:
Please see https://github.com/ElektraInitiative/libelektra/pull/2770

PR #2769, opened 2019-06-10
Travis: Simplify Mac Setup

sanssecours:


markus2330:
Is it a good idea to not cache Library/Caches/Homebrew?

sanssecours:
> Is it a good idea to not cache Library/Caches/Homebrew?

According to the [Travis documentation](https://docs.travis-ci.com/user/caching/#things-not-to-cache):

> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source…

it makes sense to not cache Homebrew packages.

markus2330:
Thanks a lot for the maintenance work.

PR #2809, opened 2019-06-25
Symbol Versioning

kodebach:
This introduces a setup for proper symbol versioning. I hope I listed all symbols correctly.

- Stuff that isn't part of any header is not listed at all, even if it was exported until now. This only affects a few `elektra*` helper functions.
- Stuff that is listed in a non-installed header like `kdbprivate.h` and actually used in another compilation unit (i.e. must be available during linking), is listed in `libelektraprivate_1.0` so that it gets exported.
- Everything else, i.e. the symbols that are actually public API is listed in `libelektra_0.8`. Symbols that were not part of the last release should be listed in `libelektra_0.9`. Right now that is only the case for the highlevel API for everything else I still need to check what was released and what wasn't.

Like stated in #1271 the basis of the whole system is the FreeBSD source.

## Basics

## Checklist

- [ ] I added unit tests
- [ ] I added code comments, logging, and assertions (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] meta data is updated (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))

## Review

## Labels

- Add the "ready to merge" label **if the build server is happy** and also you


kodebach:
Sidenote: The reason for this PR is so that `elektraOpen` can be changed to include `kdbEnsure` functionality.

markus2330:
Amazing that you work on it!

Btw. you also should ABI tests to see if test cases compiled against the old Elektra still work after this PR.

kodebach:
> Btw. you also should ABI tests to see if test cases compiled against the old Elektra still work after this PR.

You mean compile the libraries, `git checkout` the last release and then compile its test cases against the new libraries?

kodebach:
Also a separate question:

`libtools` has a separate version script already. Should that be integrated or is there any special reason that it is separate?

markus2330:
> You mean compile the libraries, git checkout the last release

Yes.

> and then compile its test cases against the new libraries?

No, not compiling against the new libraries. Compile both versions separately and then simply replace the files by coping over them (dynamic linking will do the rest).

kodebach:
> libtools has a separate version script already. Should that be integrated or is there any special reason that it is separate?

this questions is still open

> Btw. you also should ABI tests to see if test cases compiled against the old Elektra still work after this PR.

Automating this will be problematic. Ideally we would execute the test cases from the last release against the current source code to ensure backwards compatibility. However, that causes certain problems, e.g. #2821.

Some of these problems are not easily solved. For example:

- If internal functions not part of the public API are changed or removed tests will fail (e.g. the `elektraGenInternal*` functions have been removed since the last release). Such test failures are not a problem, but impossible to detect automatically and ignoring a test case entirely might hide other incompatibilities (e.g. ignoring `testkdb_highlevel` because `elektraGenInternal*` was removed, hides all problems with the highlevel API).
- Running the plugin tests is a good way to cover a lot of the API surface, but if a plugin changed some of its functionality since the last release, the test cases will fail. e.g. `boolean` had some changes since the last release that required changes in `keySetString` (namely the `origvalue` stuff).

We could of course just run a small subset of tests that should always work (e.g. the `testabi_*`)  tests, but I think the best solution is to just do manual ABI compatibility testing.

kodebach:
@markus2330 since we decided in our meeting to leave `libtools` as is, I think we can merge this now

kodebach:
@markus2330 As you can see in #2805 (which incorporates this PR) symbol versioning seems to create more problems than anticipated. First of all, it is not supported on macOS. Secondly it doesn't compile at all on the Jenkins server, while it works fine on Travis and my local machine.

IMHO the best way forward is to not use symbol versioning and simply allow breaking changes once in a while. Unless of course somebody with a deeper understanding of linking and symbol versioning not just on Linux, but on macOS and FreeBSD can help out here.

markus2330:
> symbol versioning seems to create more problems than anticipated.

which ones?

> IMHO the best way forward is to not use symbol versioning and simply allow breaking changes once in a while.

This is okay for 0.9 but not okay during 1.0 maintenance. So it would be great if we can get this PR merged.

What about enabling symbol versioning only for Linux for now?

kodebach:
> which ones?

I mentioned some above and in general it seems to be very very hard to do right. Apart from the glibc I didn't find any examples of cross-platform libraries that use it. However, I found lots of libraries that decided to not use symbol versioning (because of the difficulties involved) and to accept breaking changes whenever necessary. Some even decided against version scripts altogether not just against having multiple versions of the same symbol.

> This is okay for 0.9 but not okay during 1.0 maintenance. So it would be great if we can get this PR merged.

We could just adopt [semantic versioning](https://semver.org). It is not perfect, because of Elektra's global nature, but after 1.0 our API should be fairly stable and no major changes should be necessary.

Semantic versioning would also be more compatible with the way macOS does library versioning. (Basically during compilation we can define a version of our library and the minimum compatible version.)

> What about enabling symbol versioning only for Linux for now?

I am working on that right now. But it's not really a solution, but rather a crude work around. It might also require a more complicated setup than we have right now.

kodebach:
There is also another solution that doesn't use symbol versioning (at least not in the conventional sense) at all. We could simply generate a header that contains macros overriding each symbol name with a versioned variant e.g. `#define kdbOpen kdbOpen_v1`.

I don't know whether that would solve all the same problems as proper symbol versioning, but it works on every platform. For the problem of hiding some symbols, we can use the same approach as libssh with their [`LIBSSH_API` macro](https://github.com/substack/libssh/blob/master/include/libssh/libssh.h).

I created a test project based on this idea [over here](https://github.com/kodebach/symver-test)

markus2330:
> I mentioned some above

Do you mean "macOS not supported" and that "Jenkins report a problem"?

> Apart from the glibc I didn't find any examples of cross-platform libraries that use it.

ncurses: https://invisible-island.net/ncurses/ncurses-mapsyms.html

> to accept breaking changes whenever necessary.

This moves all difficulties to the clients (and it was very challenging to use Elektra across versions before 0.8). What we could do is to simply not change the API within 1.0 as we did during the whole 0.8 series. But as you saw with elektraOpen sometimes it is *really nice* to be able to change arguments. If it creates many complications and problems, however, we can live without it.

> in general it seems to be very very hard to do right.

That is why we should try it before 1.0.

> We could just adopt semantic versioning.

I planned to use it anyway. Before 1.0 semantic versioning allows anything. Semantic versioning, however, does not really help for the SOVERSIONs. Symbol versioning helps: it allows you to (nearly) never increment the SOVERSION.

> Therefore finishing this PR now would consume a lot of time that would be better spent on the code generation and LCDproc.

I see: yes, if you cannot find a quick way to get it running for Linux, let us rest this idea for now.

kodebach:
> Do you mean "macOS not supported" and that "Jenkins report a problem"?


> ncurses: https://invisible-island.net/ncurses/ncurses-mapsyms.html

Yes, I have found that article as well, since yesterday. However, I think the author is also not a fan of symbol versioning. Also, as far as I can tell, `ncurses` has no symbols with multiple versions. Github can't find any mention of `__asm__` or `.symver` in the source and the version script contains no symbol twice.

> This moves all difficulties to the clients (and it was very challenging to use Elektra across versions before 0.8). What we could do is to simply not change the API within 1.0 as we did during the whole 0.8 series. But as you saw with elektraOpen sometimes it is _really nice_ to be able to change arguments. If it creates many complications and problems, however, we can live without it.

I could solve some of the problems (see below), but macOS will still be a problem. The problem with `elektraOpen` could have been avoided, if we had more carefully planned ahead (`kdbEnsure` was planned for a long time already).

The best solution is probably to declare new APIs as experimental for some time and only after a few minor releases guarantee stability. In fact the real problem with `elektraOpen` was that we released it as a stable API, before finishing the whole highlevel API including code-generation.

> I planned to use it anyway. Before 1.0 semantic versioning allows anything. Semantic versioning, however, does not really help for the SOVERSIONs. Symbol versioning helps: it allows you to (nearly) never increment the SOVERSION.

`SOVERSION`s are still required (to some extent) on macOS.

> I see: yes, if you cannot find a quick way to get it running for Linux, let us rest this idea for now.

I think I fixed the problems in #2805 (there it can actually be tested). There is just one bug with spaces in paths left that I need to fix, then I will move the code into this PR.

markus2330:
> The best solution is probably to declare new APIs as experimental for some time and only after a few minor releases guarantee stability.

We already do this since a long time (kdbproposal.h) but this also has limitations because sometimes important extensions are not seen.

> I think the author is also not a fan of symbol versioning.

I am also not a fan of it. But it solves an important problem: It allows you to add arguments without breaking ABI and without exposing macros (which are ugly from programmers point of view as they disallow the use of the name at all).

> SOVERSIONs are still required (to some extent) on macOS.

And also musl-based systems.

But if a system does not support symbol versioning they obviously will not benefit from it. The respective maintainers need to take care of these problems then (like SOVERSION updates or recompilation of applications).

> I think I fixed the problems in #2805

Great! I am looking forward to merge this for 0.9 then!

kodebach:
> We already do this since a long time (kdbproposal.h) but this also has limitations because sometimes important extensions are not seen.

`kdbproposal.h` is a bad solution. It is a different header file and therefore not forward compatible. Also all of the functions declared there are defined in the separate library `elektra-proposal` (needs to be linked separately), apart from `elektraKeySetName` which is part of `elektra-core`. The whole thing is just confusing to me.

A better solution would be to define experimental symbols inside a `#ifdef ELEKTRA_EXPERIMENTAL` section (or use an integer for different versions like `_POSIX_C_SOURCE`). Then the user has to explicitly enable experimental stuff and if a symbol becomes stable everything still works.

markus2330:
> A better solution would be to define experimental symbols inside a #ifdef ELEKTRA_EXPERIMENTAL section (or use an integer for different versions like _POSIX_C_SOURCE). Then the user has to explicitly enable experimental stuff and if a symbol becomes stable everything still works.

Yes, this would be a practical solution. (Although it is technically wrong: `#ifdef` is on source-level, changes in the signatures of the functions would still break ABI.) Can you make a separate issue for that? In any case we should get rid of libproposal before 1.0.

kodebach:
> #ifdef is on source-level

Isn't `#include` to include a special header also source-level?

> Can you make a separate issue for that?

See #2840

markus2330:
> Isn't #include to include a special header also source-level?

Of course. But with libproposal the ABI changes happened in a library people normally did not link to.

> See #2840

Thank you!

kodebach:
I have no idea whatsoever, why the MinGW build fails. As far as I can tell, it is because the `elektra-full` library is linked against the C++ stdlib and when gcc tries to link `elektratools-full` against `elektra-full`, we get duplicate symbols because of the static linking of the stdlib. What I cannot explain is why this happens now, but doesn't happen on `master`.

I tried the whole thing locally in docker and interestingly `elektratools-static` works.

markus2330:
What about disabling symbol versioning for mingw?

kodebach:
It is already disabled, because symbol versioning is not supported on mingw

kodebach:
@markus2330 This should now finally be fixed. Turns out the mingw compiler does support the `--version-script` argument and we need to use it. We just have to disable the `.symver` pseudo assembler directives.

markus2330:
Great job! Thank you so much!

PR #2819, opened 2019-07-09
Docker: Add Images for Debian buster

sanssecours:


scripts/docker/debian/buster/Dockerfile.doc
```diff
@@ -0,0 +1,54 @@
+FROM debian:buster
```
markus2330: Shouldn't it be doc.Dockerfile? See https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles

(Same problem also in stretch)

src/plugins/yamlcpp/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ if (DEPENDENCY_PHASE)
 	     AND
 	     ${CMAKE_CXX_COMPILER_VERSION}
 	     VERSION_LESS
-	     7)
+	     9)
```
markus2330: Thanks for maintaining ASAN. What happens with this gcc version?

scripts/docker/debian/stretch/Dockerfile.doc
```diff
@@ -31,12 +30,17 @@ RUN mkdir -p ${GTEST_ROOT} \
     && tar -zxvf gtest.tar.gz --strip-components=1 -C ${GTEST_ROOT} \
     && rm gtest.tar.gz
 
+# ronn
+# Unfortunately the package [`ruby-ronn`](https://packages.debian.org/buster/all/ruby-ronn/filelist) does not include the `ronn` executable.
```
markus2330: They have split the package, it is now called ronn https://packages.debian.org/buster/ronn

scripts/docker/debian/stretch/Dockerfile.doc
```diff
@@ -31,12 +30,17 @@ RUN mkdir -p ${GTEST_ROOT} \
...
+RUN gem install ronn
+
 # Create User:Group
 # The id is important as jenkins docker agents use the same id that is running
 # on the slaves to execute containers
 ARG JENKINS_GROUPID
 RUN groupadd \
     -g ${JENKINS_GROUPID} \
+    -f \
```
markus2330: Why is this needed?

sanssecours: Otherwise building the image on macOS will not work (see also commit https://github.com/ElektraInitiative/libelektra/pull/2819/commits/8b916927d9b459711813bde3192bee0ba85eff48).

src/plugins/yamlcpp/CMakeLists.txt
sanssecours: As far as I can tell, there seems to be a problem in the code of the header `shared_ptr_base.h` in `libstdc++`, reported by ASan enabled builds:

```
/usr/include/c++/8/bits/shared_ptr_base.h:735:4: runtime error: member call on address 0x603000000010 which does not point to an object of type '_Sp_counted_base'
0x603000000010: note: object is of type 'std::_Sp_counted_ptr_inplace<antlr4::CommonTokenFactory, std::allocator<antlr4::CommonTokenFactory>, (__gnu_cxx::_Lock_policy)2>'
 01 00 80 4c  40 9f 25 fb 87 7f 00 00  01 00 00 00 01 00 00 00  10 9f 25 fb 87 7f 00 00  00 be be be
              ^~~~~~~~~~~~~~~~~~~~~~~
```

. In Clang ASan builds we just ignore this error using a [blacklist](https://clang.llvm.org/docs/SanitizerSpecialCaseList.html):

https://github.com/ElektraInitiative/libelektra/blob/d95c76004518d1b7db1f7b069c4f804db951afed/tests/sanitizer.blacklist#L1-L5

. Since GCC [does not support blacklists](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59454) we disable the plugin for all versions of `libstdc++` that still contain the problematic code. I added a comment about this to the CMake code in commit 99e56a1.

markus2330:
Great job!

PR #2848, opened 2019-08-02
Build: Do Not Install System Files by Default

sanssecours:
It would be great, if we could include this PR in the next release. This way issue #2835 should be resolved pretty quickly, after we update the Homebrew formula for Elektra `0.9`.


markus2330:
Ok, I'll add it. The only sys files we have is GI files (which are obsolete) and the shell completions (which do not work well, except of fish).

markus2330:
Seems like we also need to fix debian/control before we can merge this. I did it in 6c3a23d2e3b3c18f90eeae277583ab18945e8f49

markus2330:

markus2330:
Thank you!

Issue #2697, opened 2019-05-13
unify error messages via templates


markus2330:
At the moment very similar error/warnings have irrelevant differences or are missing important information (e.g. #2660).

I propose that we get templates (macros with predefined text) for some of the error categories. E.g. for validation errors, in nearly all cases, the error message template could look like:

```
Sorry, the validation of the key '<key>' failed because the value '<value>' is not one of the allowed values '<allowed values>'.
E.g. for the range plugin:
Sorry, the validation of the key '/somewhere' failed because the value '5' is not one of the allowed values '7-9'.
In the code one would use:
ELEKTRA_SET_VALIDATION_ERROR_T(key, value, allowed_values)
```

Of course it would not work with every error message but it would be helpful in common cases.

kodebach:
It is a nice idea, but the use of this macro would be complicated. The `allowed_values` probably has to be constructed out of other parts. Without the macro you can use something like `ELEKTRA_SET_VALIDATION_SEMANTIC_ERRORF(key, "[...] of the allowed values '%d-%d'", min, max)`. With the macro that would require additional malloc/free.

Also I think your template is to restrictive. Many validation plugins just have a condition that the value has to meet (reference to existing key, must be integer, match regex, etc.). Something like this may work better:

```
Sorry, the validation of key `<key>` with value '<value>' failed, because <reason>.
```

Where `<reason>` could be "it doesn't reference an existing key", "the value is not an integer" or "the value didn't match the regex [...]". But this still has the problem of complicated use in many cases.

markus2330:
> The allowed_values probably has to be constructed out of other parts.

As said: not in every situation it is best to use the macro. If you have exactly these arguments, they will help. If one argument is missing, you will hopefully rewrite the code so that you deliver the needed information. (Which is the main purpose of these templates: to give a guideline which information should be present.)

We already have such macros, e.g. ELEKTRA_MALLOC_ERROR and ELEKTRA_SET_ERROR_READ_ONLY, see src/include/kdbmacros.h
They were introduced to make the error handling more consistent.

My proposal here is to unify them and to match them with the categories we have.

> Also I think your template is to restrictive. [...] , because <reason>

We decided to drop the separation of "description" and "reason". Furthermore, "reason" would be typically constructed out of other parts.

markus2330:
For resource errors, it might make sense to have something like:

`ELEKTRA_SET_SYSCALL_ERROR (errno, resource, msg, ..`

[...]

PR #2854, opened 2019-08-03
Jna error code and plugin usage

Piankero:
## Basics

## Checklist

need to be checked.

- [x] I added unit tests
- [ ] I added code comments, logging, and assertions (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] meta data is updated (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))

## Review

- [x] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [x] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)
- [x] Examples are well chosen and understandable

## Additional Notes

I completely overhauled the jna binding and also removed the strange code examples for a plugin since it did nothing useful than logging stuff to the console without any interaction to elektra. The new Plugin class is actually usable.


markus2330:
Thank you for the PR! I'll look at it after the release. Please make it pass the build server.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/plugin/PropertiesStorage.java
```diff
@@ -1,74 +0,0 @@
-package org.libelektra.plugin;
```
markus2330: Why is this plugin removed here? The PR is vey difficult to read due to reformatting of all files. Please describe some formatting for Java in doc/CODING.md (can be separate to Groovy, or also reformat scripts/jenkins/Jenkinsfile)

Piankero: I did not reformat anything. I deleted a lot of files which basically had no senseful meaning like the Echo.java or the PropertiesStorage.java because:

1. I want to use the native elektra plugin from is implemented by the new Plugin class and used by the ErrorCodeTest (I use the `error` plugin there)
2. This would be in a conflicting name with the current `Plugin` interface which more or less just gave 4 methods to implement and basically had no real connection to KDB at all
3. The "implemented Plugins" in Java did not interact with KDB at all except for reading something from the parent key. Isn't it the meaning of elektra to load all into a keyset and handle the operations there rather than doing it with the native java code that directly manipulates the files? The specification for example could never be used with a java plugin. 

Due to all of these reasons I decided to remove them because they might even mislead new developers in implementing plugins in java

markus2330: > I did not reformat anything.

Please look at the diff in GitHub. E.g. src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java had tabs before and in this PR it now uses spaces.

> The "implemented Plugins" in Java did not interact with KDB at all except for reading something from the parent key. Isn't it the meaning of elektra to load all into a keyset and handle the operations there rather than doing it with the native java code that directly manipulates the files? The specification for example could never be used with a java plugin.

These plugins are actual Elektra plugins which can be mounted (and used as storage plugin). Please read src/plugins/jni/README.md to understand the plugin interface.

I am very glad if you improve the plugin interface (even beyond the introduction of the new error concept) but:

- there needs to be documentation on how to use it
- you need to update existing documentation (this PR does not touch documentation at all)
- there need to be examples
- there need to be test cases

Piankero: > Please look at the diff in GitHub. E.g. src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java had tabs before and in this PR it now uses spaces

On my GitHub I do not see a difference between tabs and spaces, do you have a special plugin or setting which shows this?

Apart from that: It is the default setting in intellij is to use spaces and it sometimes even suggests to use them instead of tabs if it sees it is used in a file. It is in general much more common to use spaces in java. I can revert that back to tabs if you like but imho it does not make a lot of difference at all

> These plugins are actual Elektra plugins which can be mounted (and used as storage plugin). Please read src/plugins/jni/README.md to understand the plugin interface.

I readded them and renamed the plugin to nativeplugin which was used by myself to for the test case.

markus2330: You see tabs/spaces by trying to mark them.

It is commons sense that mixing tabs and spaces is the worst of all possibilities. So yes, please revert to tabs. We can convert everything to spaces in a later PR to have more Java-style. In any case, it is good if this is documented in doc/CODING.md. Even better would be to automatically reformat with a tool (and on the build server).

Piankero: > You see tabs/spaces by trying to mark them.

Honestly I have seen problem with line endings but never with tabs/spaces. Nontheless I think there is a good plugin in case you are using firefox or chrome for that https://github.com/glebm/render-whitespace-on-github. I used it now to quickly see tabs/spaces.

I fixed the tab/spaces stuff. Please review the rest of the PR :)

Piankero:
@markus2330 I have absolutely no idea why on that one build it does not find the error plugin and returns null. I will exclude let the test pass in case the plugin returns null

doc/news/_preparation_next_release.md
```diff
@@ -99,6 +99,9 @@ you up to date with the multi-language support provided by Elektra.
 ### Java
 
 - Upgraded maven dependencies for java binding _(Michael Zronek)_
+- Completely overhauled the java binding to be able to use Elektra plugins directly. For an example see the [test case](src/bindings/jna/libelektra4j/src/test/java/org/libelektra/ErrorCodeTest.java). _(Michael Zronek)_
```
markus2330:
```suggestion
- Completely overhauled the Java binding to be able to use Elektra plugins directly. For an example see the [test case](src/bindings/jna/libelektra4j/src/test/java/org/libelektra/ErrorCodeTest.java). _(Michael Zronek)_
```

doc/news/_preparation_next_release.md
```diff
@@ -99,6 +99,9 @@ you up to date with the multi-language support provided by Elektra.
...
+- The java binding now supports the [error codes](doc/decisions/error_codes.md) in a native way. All exceptions contain the necessary information. For
```
markus2330:
```suggestion
- The Java binding now supports the [error codes](doc/decisions/error_codes.md) in a native way. All exceptions contain the necessary information. For
```

doc/news/_preparation_next_release.md
```diff
@@ -99,6 +99,9 @@ you up to date with the multi-language support provided by Elektra.
...
+  an example please take a look at the [test case](src/bindings/jna/libelektra4j/src/test/java/org/libelektra/ErrorCodeTest.java). _(Michael Zronek)_
```
markus2330: Twice the same example?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/NativeElektraPlugin.java
```diff
@@ -0,0 +1,118 @@
...
+import org.libelektra.exception.InstallationException;
+import org.libelektra.exception.InternalException;
+import org.libelektra.exception.KDBException;
+import org.libelektra.exception.mapper.ExceptionMapperService;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class NativeElektraPlugin {
```
markus2330: No documentation for this class and its methods

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
```diff
@@ -0,0 +1,16 @@
+package org.libelektra.exception;
```
markus2330: These files should be generated from the specification.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
```diff
@@ -0,0 +1,74 @@
...
+		return errorKey.getMeta("error/module").getString();
+	}
+
+	@Override
+	public String getLocalizedMessage() {
+		return getMessage();
+	}
+
+	@Override
+	public String getMessage() {
```
markus2330: Shouldn't that print the whole message?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
```diff
@@ -0,0 +1,49 @@
...
+			Key temporaryError = Key.create("user/temporary/errorkey");
+			temporaryError.setMeta("error/number", InternalException.errorCode());
+			temporaryError.setMeta("error/reason", "Sorry, Elektra returned an error/warning but no " +
+					"information along it, please report this incident at https://issues.libelektra.org/");
+			return new InternalException(temporaryError);
+		}
+
+		String errorCode = k.getMeta("error/number").getString();
+
+		if (errorCode.equals(ResourceException.errorCode())) {
```
markus2330: Also this here should be generated.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class ExceptionMapperService {
+
+	public static KDBException getMappedException(Key k) {
+
+		// At least one error or warning should be there
+		if (k.getMeta("error/number").isNull() && k.getMeta("warnings/#00/number").isNull()) {
+			Key temporaryError = Key.create("user/temporary/errorkey");
+			temporaryError.setMeta("error/number", InternalException.errorCode());
+			temporaryError.setMeta("error/reason", "Sorry, Elektra returned an error/warning but no " +
```
markus2330: Undocumented hack to rewrite the reason? Why not append the keys when returning the message?

markus2330:
Thank you! Looks much better now :sparkle: The class design is much nicer.

Without docu, however, it is really hard to understand how a developer can use your classes.

doc/news/_preparation_next_release.md
Piankero: Yes, one shows how to use the `NativePlugin` and one shows how Exceptions are used. They are both  used in the test

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
Piankero: I fear that autogeneration is not possible.
The main reason is the class hierarchy which maps to the error code.

I have also mapped invisible classes (eg. InstallationException) which do not appear at all in the specification file. These classes are neccessary though so developers can catch the more generic exception which was one of the main intentions behind the error code design.

Furthermore the autogeneration would need to know which error code is below which for correct inheritence. So there would be error prone recursive parsing of the individual digits.
Additionally we would completely deny ourself special implementation in subclasses, e.g., the ConflictingState/ OutOfMemory Exception which have more of a special meaning. All classes would be equally with autogeneration.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: This is usually just 1-2 sentences and not a big bunch of additional information. It is a bit weird to try to marry C kind of error with java exceptions. I personally would only want to get the Reason message and try to avoid the other verbose and eventually unneccessary output. I can though print everything if you want.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
Piankero: See above

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
Piankero: > Undocumented hack to rewrite the reason?

It is not to "rewrite" the reason but to actually give one because Elektra did not provide one. E.g., you make a call to Elektra, it returns -1 but the errorkey is empty. This should never happen in reality but I provided a fallback so the user would not receive a NullPointer/ Empty Message

Piankero: If you wish I can add another constructor which expects (String errorCode, String reason) as paramters

doc/news/_preparation_next_release.md
markus2330: It would be good to have more than one example. And examples should not try to to much. I still do not understand what the differences in the new plugin system is.

If it is better, we should migrate the old plugins (which are only small examples anyway) to the new system.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
markus2330: > I fear that autogeneration is not possible.

Then the specification is not like it should be. We want the specification useful for several languages.

> All classes would be equally with autogeneration.

It is always possible to specialize, like I also suggested in  #2697

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: It is not about what I want but it is about consistency of error messages. Why should a Java-GUI print different error messages than the Qt-GUI?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
markus2330: I do not think that another constructor solves the problem.

> you make a call to Elektra, it returns -1 but the errorkey is empty.

Then this is the wrong place to fix this. This code must be within the `kdb*` functions.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
Piankero: The specification is useful for all languages. It is just that the specification *file* does not have all errors.
Autogeneration of C code to an exact equivalent optimized java code is very error prone. Especially when it comes to inheritance. There are actually [companies](https://www.mtsystems.com/) who specialize in this matter and still are not able to do it perfectly. Even from java to kotlin or vice versa it does not work well and they both rely on the jvm.

Implementing your desired autogeneration would be worth at least another bachelor thesis.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: The QT-GUI also just prints the Reason message and on expansion prints everything. I do not see why they are different

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
Piankero: This is not "fix" per se but a safeguard. There may always be changes in the kdb functions which may cause no error to be returned. The java binding would at least catch this error and not return a null pointer. I already had this behavior when trying to load a plugin which was not existent on the build server.

Piankero: > I do not think that another constructor solves the problem.

The additional constructor would be just some better usability like a constructor with a predefined size I have done in the KeySet class

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
markus2330: > The specification is useful for all languages.

This is to be shown by you.

> It is just that the specification file does not have all errors.

Then how can it be useful?

> Especially when it comes to inheritance.

You maybe need to add the relevant information to the specification.

> Autogeneration [...]

You are talking something very offtopic here and use wrong terms. Nobody requested to write a transpiler.

The code (the exceptions) should be generated from the (one and only) specification src/error/specification

> Implementing your desired autogeneration would be worth at least another bachelor thesis.

If this would be the case then something somewhere went terribly wrong. In one bachelor thesis two complete bindings were written. The new specification should make it easier, not harder.

We also need to write some generator for Go and Rust.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: I expected that "getReason" will return the reason and "getMessage" will return "everything".

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
markus2330: https://de.wikipedia.org/wiki/Design_by_contract

Clients should be able to rely on that they get an error message if they get `-1`. If this is not the case, we have a bug, please report it.

Piankero: I know that it is design by contract. If the contract is violated though it is still good working code. I will simply write "please report a bug to libelektra..." instead.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: getReason is elektra specific while getMessage is java specific. I simply wanted to have both working. But I will change it "everything"

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
Piankero: > This is to be shown by you.

It is shown here that it works for both java and C very well which can be seen in this PR. It is just that autogeneration is very complicated and error prone.

> Then how can it be useful?

Because developers are only permitted to use those error codes which are relevant and not those which are too generic. On could also write down additional codes with a metatag that says it is not allowed to use but only for "documentation" but this is really just unnecessarily bloating of the specification file.

> The code (the exceptions) should be generated from the (one and only) specification src/error/specification

Lets talk about this in another elektra meeting as it will simply be too time intensive writing here all the time.

> If this would be the case then something somewhere went terribly wrong. In one bachelor thesis two complete bindings were written. The new specification should make it easier, not harder.
We also need to write some generator for Go and Rust.

Did these bindings include autogeneration and exception mapping for a language which has the same sort of inheritance as java? Go and Rust have a different form of inheritence which is easier to model

Piankero: Lets talk about this in an elektra meeting as this would otherwise be more time intensive than needed

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
markus2330: > I know that it is design by contract. If the contract is violated though it is still good working code.

Then please also describe your interface according design-by-contract. I did not see that you try to avoid to return null pointers.

> I will simply write "please report a bug to libelektra..." instead.

Yes, better.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: All these discussions would not be needed if you had some documentation about what these methods are supposed to do. It is completely unclear what "I simply wanted to have both working" means. What is a working method? One that does not throw exceptions?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
markus2330: Just to clarify what I am talking about: The specification in src/error/specification should be complete enough, that the relation (or subtyping, inheritance, ..) between the error codes can easily be known within a tool that generates exception classes (or needs to know if one error is subtype of another). If you find the encoding as error code letters too difficult yourself, why not encode it in an easier form?

> Did these bindings include autogeneration and exception mapping for a language which has the same sort of inheritance as java?

Yes, it used code generation far beyond of what we are talking here. Generating these few exception classes is trivial compared to that. Feel free to have a look at src/bindings/swig/python/kdb.i

> Lets talk about this in an elektra meeting as this would otherwise be more time intensive than needed

We can also talk about it but we also should document how we came to the conclusion.

And it is an "urgent" matter because:

1. you want to finish
2. the decision is needed for Go and Rust bindings which are finished soon, too

I started  #2871.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: If someone comes from C (or more specific Elektra) and uses the java binding he would expect "getReason" to be the method with the most "useful" String in it.
If however someone comes from Java he would expect the "getMessage()" method to be the most relevant one.
I wanted to supply both so either of them is satisfied. But you wish for the whole message in the reason part which is also perfectly fine

markus2330: I expected the getMessage() to be the most relevant one. So I wondered why it did not contain the whole error message.

> I wanted to supply both so either of them is satisfied.

Of course it makes sense to have getters for the individual parts but it also makes sense to have a getter for the standardized message format as specified by you in doc/decisions/error_message_format.md

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
Piankero: > Yes, better.

Actually I see that the current error message is already in this form (did not see it in the marked PR line immediately) `Sorry, Elektra returned an error/warning but no information along it, please report this incident at https://issues.libelektra.org/`

doc/news/_preparation_next_release.md
Piankero: I added another example in the README.md
I migrated the old plugins to the new Interface with added methods. Also NativeElektraPlugins implements it and works.
I removed Template&ExceptionTest as it is not really valuable

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
Piankero: Since this is discussed in #2871 it can stay now as it is

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: You can see an example message now in the README.md

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
```diff
@@ -14,7 +22,7 @@
 	 *            a key
 	 * @return the plugin's return value for open
 	 */
-	int open(KeySet conf, Key errorKey);
+	int kdbOpen(KeySet conf, Key errorKey);
```
markus2330: Why do you rename here? The prefix seems unnecessary (both for Plugin and PluginLoader).

It would be great if you can add a test where a Java plugin is loaded and tested.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/ConflictingStateException.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.libelektra.Key;
+
+public class ConflictingStateException extends KDBException {
+
+	private static String errorCode = "C02000";
```
markus2330: final?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
```diff
@@ -0,0 +1,90 @@
+package org.libelektra.exception;
+
+import org.libelektra.Key;
+import org.libelektra.exception.model.WarningEntry;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * Custom KDB exception class being used for I/O errors
```
markus2330: for I/O errors?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public abstract class KDBException extends java.io.IOException {
+
+	private static final long serialVersionUID = 1L;
```
markus2330: ?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
```diff
@@ -0,0 +1,90 @@
...
+		for (int i = 0; i <= nr; i++) {
+			warnings.add(new WarningEntry(k, nr));
+		}
+	}
+
+	public Key getErrorKey() {
+		return errorKey;
+	}
+
+	public String getErrorCode() {
```
markus2330: There is a huge mess with code/number. Please use always the same.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: Thank you!

But the methods are still not documented.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/LogicalException.java
```diff
@@ -0,0 +1,16 @@
...
+
+public abstract class LogicalException extends PermanentException {
+
+	private static String errorCode = "C01300";
+
+	public LogicalException(Key k) {
+		super(k);
+	}
+
+	public static String errorCode() {
```
markus2330: Why no @Override here?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/plugin/NativeElektraPlugin.java
```diff
@@ -0,0 +1,169 @@
...
+import org.libelektra.exception.KDBException;
+import org.libelektra.exception.mapper.ExceptionMapperService;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class can be used to load native C/C++ written Elektra Plugins to be used by Java directly
+ */
+public class NativeElektraPlugin implements Plugin {
```
markus2330: See above, another name would fit better.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/plugin/NativeElektraPlugin.java
```diff
@@ -0,0 +1,169 @@
...
+import org.libelektra.Plugin;
+import org.libelektra.exception.InstallationException;
+import org.libelektra.exception.KDBException;
+import org.libelektra.exception.mapper.ExceptionMapperService;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class can be used to load native C/C++ written Elektra Plugins to be used by Java directly
```
markus2330:
```suggestion
 * This class can be used to load native Elektra Plugins in Java directly
```

src/bindings/jna/libelektra4j/src/test/java/org/libelektra/ErrorCodeTest.java
```diff
@@ -0,0 +1,77 @@
...
+			errorPlugin.kdbSet(ks, parentKey);
+		} catch (KDBException e) {
+			assertEquals(e.getWarnings().size(),1);
+			assertTrue(e instanceof ResourceException);
+			assertEquals(e.getWarnings().iterator().next().getWarningCode(), SemanticValidationException.errorCode());
+			return;
+		}
+		throw new RuntimeException("Exception did not trigger");
+	}
+}
```
markus2330: Not many tests here. No newline at the end.

src/bindings/jna/README.md
```diff
@@ -113,6 +113,64 @@ Most likely, you have to include this dependency as well:
 
 [here](../../examples/external/java/read-keys-example/pom.xml) you can find a fully working example of the pom file.
 
+### Using Elektra plugins directly
+
+The `NativeElektraPlugin` can be used to load a native Elektra plugin which is written in C/C++.
```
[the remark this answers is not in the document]
Piankero: I don't think "loader" is correct as the class itself then acts as the plugin.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/KeySet.java
```diff
@@ -43,7 +43,7 @@ protected KeySet(final long p) {
 	 *
 	 * @param p Pointer to another KeySet
 	 */
-	protected KeySet(final Pointer p) {
+	public KeySet(final Pointer p) {
```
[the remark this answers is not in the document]
Piankero: Because I moved the NativeElektraPlugin into the plugin package and it makes usage of this method. It does not harm the development experience though if this is public.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
Piankero: The names are the equivalents to those in the C written implementation. I wanted it to be consistent but if you wish I can refactor it.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: This documentation was copied from the old Exception class which I patched. I was also wondering why `I/O` errors are meant but thought it will have some special meaning behind it. I can remove it

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: Same reason as above

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: What do you mean with "mess"? Do I sometimes write "code" and sometimes "number"?

src/bindings/jna/libelektra4j/src/test/java/org/libelektra/ErrorCodeTest.java
Piankero: It tests the basic functionality of the errors with success. Which other tests do you have in mind?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
Piankero: > It would be great if you can add a test where a Java plugin is loaded and tested.

I do not have written those plugins and also do not have the time to test them unfortunately.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: Yes, see the line below.

src/bindings/jna/libelektra4j/src/test/java/org/libelektra/ErrorCodeTest.java
markus2330: There are so many things not tested (related to errors). I somewhat understand that you do not want to make so many tests in C and shell but that you also write so little tests in Java is surprising for me. As you removed the asserts in two tests, you literally wrote a single test. If you really do not know what to test: You can easily find them via code coverage.  E.g. warnings, warnings/errors from native plugins, ... are obvious parts.

src/bindings/jna/README.md
```diff
@@ -113,6 +113,64 @@ Most likely, you have to include this dependency as well:
...
+It will behave exactly like a normal Elektra plugin. The main
+advantage is the isolation from the rest of Elektra as well as
+the circumvention of the "too many plugins" [error](https://github.com/ElektraInitiative/libelektra/issues/2133).
+You can load a Plugin like the following:
+
+```java
+Key parentKey = Key.create("user/tests/javabinding");
+String errorPluginName = "error";
```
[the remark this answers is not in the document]
markus2330: But only once?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
markus2330: > The names are the equivalents to those in the C written implementation. I wanted it to be consistent but if you wish I can refactor it.

It is astonishing that such important decisions are not mentioned at all. **Please** document such things, it would have saved a lot of time for both of us.

src/libs/tools/include/plugin.hpp is not really very good design, but it also avoid the "kdb" prefixes.

Instead of the modules as parameter it is nicer to have a PluginLoader, which is a very usual thing in Java.

src/bindings/jna/README.md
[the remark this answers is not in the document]
markus2330: It is also weird that both things are married in one class. See in other comment about design.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
markus2330: Furthermore, it would be nice if one gets a "Plugin" in both cases, as the NativePlugin and the current "Plugin" obviously should be subclass of a common class.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: Where above?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: Please do not copy text you do not understand. I do not think that any part of that text is correct for what you implemented here. (It is not "Custom", and not used for I/O errors.)

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
Piankero: > It is astonishing that such important decisions are not mentioned at all. Please document such things, it would have saved a lot of time for both of us.

Why is it an important decision how I name a method? I will simply revert it to the  non-prefix version.

> Instead of the modules as parameter it is nicer to have a PluginLoader, which is a very usual thing in Java.

You mean you want some sort of "Factory"? The current NativePlugin is just a proxy to abstract the pointer logic away and implement the Plugin Interface correctly.

> Furthermore, it would be nice if one gets a "Plugin" in both cases, as the NativePlugin and the current "Plugin" obviously should be subclass of a common class.

Currently the `Plugin` class is an interface and the NativePlugin along with the other plugins implement it. I do not know what you mean here.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/KeySet.java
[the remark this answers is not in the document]
Piankero: I did not check this, I can throw a RunTime Exception if the pointer is null but apart from that I do not see any other way around making the constructor public so that the NativePlugin can load

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: > Please do not copy text you do not understand. I do not think that any part of that text is correct for what you implemented here. (It is not "Custom", and not used for I/O errors.)

Neither was it before then obviously but you usually complain if I delete any comments so I kept them in case the author before me know more than I did. I will remove them

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
Piankero: The previous author added this for serialization versioning (which was also not very intuitive for me). I will just delete it

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
markus2330: >  Why is it an important decision how I name a method? I will simply revert it to the non-prefix version.

It is not only about the method names: it was completely unclear me what the whole NativeElektraPlugin is about. The renaming of the public methods is also problematic: it breaks all already written plugins.

> You mean you want some sort of "Factory"?

Yes, something like https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html

> The current NativePlugin is just a proxy to abstract the pointer logic away and implement the Plugin Interface correctly.

Which does not say anything about if this design is good.

>  Currently the Plugin class is an interface and the NativePlugin along with the other plugins implement it. I do not know what you mean here.

I was talking about the ClassLoader: it could be a secret that such a thing like a NativePlugin exists at all.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/KeySet.java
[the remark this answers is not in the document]
markus2330: Not ideal but if your design needs it you can add the check. Please document invariants.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: Because both is wrong:

1. having wrong comments
2. having no comments

Please properly document your methods according design-by-contract.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/KDBException.java
markus2330: Yes, I do not know which use case the previous author had in mind for serializing the errors.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
Piankero: I am still not sure how exactly how I should name the NativeElektraPlugin as it has to have a name. I can make a PluginLoader but still I need to have name for the class that loads the NativePlugins from elektra and not those implemented in java

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/LogicalException.java
Piankero: Because it is a static method?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Plugin.java
markus2330: NativePlugin or Plugin in another package should be ok.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/LogicalException.java
markus2330: Exactly, wouldn't that possibly return wrong errorCodes in catch blocks?

Piankero: Why should it? The error code can be returned via `e.getErrorNumber()` which is in the KDBException and actually calls `errorKey.getMeta("error/number").getString()`. The static method is never used in a catch block

markus2330: It would be far more efficient if you improve the docu/code and avoid starting discussions. (Do you really want me to explain why someone might invoke a public method?)

Anyway: If a user should not use the method, why is it there, then? Docu does not give a glue. Furthermore, we again have a mess in code/number.

Piankero: It would not even come into my mind that something like this should be documented because it is a standard java coding style.

```java
try {
//code that causes exception
} catch (KDBException e) { \\or any subexception
	e.getErrorNumber();
}
```
Why should somebody call a static method here?

The static method is just useful so one does not have to provide a no-arg constructor (which would even be misleading and not even be possible for the abstract classes) and use `new XYException().getErrorNumber()` or lookup the error number in the design decisions.

Piankero:
@markus2330 could you please take a look now. I will update the Readme and test cases once you agree on the core structure of the java binding (PluginLoader, method naming, etc.)

markus2330:
I do not know what to say, most of what was discussed is still not fixed.

Piankero:
> I do not know what to say, most of what was discussed is still not fixed.

1. I added a `PluginLoader`
2. I renamed everything from `errorCode` to `errorNumber`
3. I added javadoc to nearly all methods

What exactly is not fixed listing? I am not yet 100% sure of your plugin loader implementation wish so I asked you to take a look at it now

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.libelektra.exception.InstallationException;
+import org.libelektra.plugin.Echo;
+import org.libelektra.plugin.NativePlugin;
+import org.libelektra.plugin.PropertiesStorage;
+import org.libelektra.plugin.Return;
+
+/**
+ * This class can be used to load Plugins from Elektra.
+ * It also loads self implemented Java plugins.
```
markus2330: Here I would have the expectation that also my newly implemented Java plugin can be loaded (which obviously is not the case as the loading is hard-coded).

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
```diff
@@ -0,0 +1,40 @@
...
+public class PluginLoader {
+
+	private Key errorKey;
+
+	public PluginLoader(Key errorKey) {
+		this.errorKey = errorKey;
+	}
+
+	/**
+	 * This plugin loads a Java/Native Elektra Plugin.
```
markus2330:
```suggestion
	 * This class loads a Java or native Elektra plugins.
```

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
```diff
@@ -0,0 +1,40 @@
...
+	 *
+	 * @param name the plugin name
+	 * @return the Plugin
+	 * @throws InstallationException if the plugin does not exist
+	 */
+	public Plugin loadPlugin(String name) throws InstallationException {
+		if (name.equals(new Echo().getName())) {
```
markus2330: I am not sure we need direct instantiation of Java plugins. But if we do it should be done in a proper way. The current implementation instantiated every plugin, even if for none of them the name matches.

doc/news/_preparation_next_release.md
```diff
@@ -98,7 +98,9 @@ you up to date with the multi-language support provided by Elektra.
 
 ### Java
 
-- Upgraded maven dependencies for java binding _(Michael Zronek)_
+- Upgraded maven dependencies for Java binding _(Michael Zronek)_
+- Completely overhauled the Java binding to be able to use Elektra plugins directly. For an example see the [test case](../../src/bindings/jna/libelektra4j/src/test/java/org/libelektra/ErrorCodeTest.java). _(Michael Zronek)_
```
markus2330: Here it is still called ErrorCode

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * This class can be used to load Plugins from Elektra.
+ * It also loads self implemented Java plugins.
+ */
+public class PluginLoader {
+
+	private Key errorKey;
+
+	public PluginLoader(Key errorKey) {
```
markus2330: Javadoc still missing. Why is this constructor needed? You can also instantiate the errorKey internally?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
```diff
@@ -0,0 +1,40 @@
...
+		if (name.equals(new Echo().getName())) {
+			return new Echo();
+		}
+		if (name.equals(new PropertiesStorage().getName())) {
+			return new PropertiesStorage();
+		}
+		if (name.equals(new Return().getName())) {
+			return new Return();
+		}
+		return new NativePlugin(name, errorKey);
```
markus2330: One more important thing: The PluginLoader should be responsible for the modules. The NativePlugin should not reopen the modules every time (such a constructor should not be offered as it would be unnecessarily slow).

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
Piankero: I could use reflection but this comes with several drawbacks.
From java 9 on Reflection is not very favorable anymore
Furthermore one would need to force a default constructor so every plugin can be instantiated via reflection.

I will add it to the README that it also has to be added to the PluginLoader implementation.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
Piankero: Developers should also have the possibility to define the errorKey themselve in my opinion to access its properties.

Should I add a no-arg constructor which with the default error key name `system/jna/default/errorkey`?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
markus2330: Yes, I also think the readme how to add the plugin would be enough.

What is not so ideal in the current solution that the Java plugins shadow other plugins. If there is also a native "echo" plugin, it would not be possible to load it.

Maybe we drop the idea to "load" Java plugins altogether.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
markus2330: > Should I add a no-arg constructor which with the default error key name system/jna/default/errorkey?

Yes, but I would leave the name empty.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/PluginLoader.java
Piankero: > What is not so ideal in the current solution that the Java plugins shadow other plugins. If there is also a native "echo" plugin, it would not be possible to load it.

I agree with you. I separate the methods into a "loadNativePlugin" and "loadJavaPlugin" to make it clear.

markus2330: good idea :+1:

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/exception/mapper/ExceptionMapperService.java
```diff
@@ -0,0 +1,49 @@
...
+
+	public static KDBException getMappedException(Key k) {
+
+		// At least one error or warning should be there
+		if (k.getMeta("error/number").isNull() && k.getMeta("warnings/#00/number").isNull()) {
+			Key temporaryError = Key.create("user/temporary/errorkey");
+			temporaryError.setMeta("error/number", InternalException.errorNumber());
+			temporaryError.setMeta("error/reason", "Sorry, Elektra returned an error/warning but no " +
+					"information along it, please report this incident at https://issues.libelektra.org/");
+			return new InternalException(temporaryError);
```
PhilippGackstatter: Just a question here: If you return this exception and later someone calls `exception.getModule()`, would that fail? Because the `error/module` key is not set here and `getModule` doesn't do any checks and doesn't throw exceptions either: 
```java
public String getModule() [ getModule()]  {
		return errorKey.getMeta("error/module").getString();
	}
```

I'm implementing the kdb errors for the Rust binding atm, so I'm looking at this PR for some inspiration and noticed that. So to fully handle the case where an error key returned from a kdb function is empty, shouldn't all the `error/*` metakeys be set?

markus2330: Yes, I also found the code weird, see my comments two lines above.

From my perspective, the `kdb*` calls must properly set the errors. It would be wrong to fix problems from the core in every binding separately.

> So to fully handle the case where an error key returned from a kdb function is empty, shouldn't all the error/* metakeys be set?

This case does not need to be handled, as it would be a bug in the core and needs to be fixed there. And it is actually quite simple to fix all such errors: for every `return -1` in the `kdb` functions, we simply need to check if an error was correctly set, and set some error (similar to the code above) if it was not set.

Piankero:
@markus2330 the build fails: 
```genhtml: ERROR: cannot write HTML data (No space left on device)```

**Comment by Piankero** [comment]

**Comment by markus2330** [comment]
The build server fails, so it cannot be "ready to merge".

**Comment by Piankero** [comment]

**Comment by Piankero** [comment]
The build server fails quite often and just hang on the last step which was docker related (https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-2854/44/pipeline/). I think it is still ready to merge because all tests ran through

File: src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java [hunk 29]
```diff
@@ -489,10 +469,10 @@ public int copyMeta(final Key source, final String metaName) {
 	/**
 	 * Helper function to copy all meta information from a source key to this key
 	 *
-	 * @param source
-	 *            Key object that is used as source
-	 * @return 1 if meta was successfully copied, 0 if source doesn't contain any meta and nothing had to be done, -1 in case of an error or
-	 *         if the source parameter was null
+	 * @param source Key object that is used as source
+	 * @return 1 if meta was successfully copied, 0 if source doesn't contain any meta and nothing had to be done, -1
```
markus2330: should be @retval

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
```diff
@@ -472,12 +453,11 @@ public Key currentMeta() {
 	/**
 	 * Helper function to copy some meta information from a source Key to this key.
 	 *
-	 * @param source
-	 *            Key object that is used as source
-	 * @param metaName
-	 *            Key name of the meta to be copied
-	 * @return 1 if meta was successfully copied, 0 if source doesn't contain the required meta and nothing had to be done, -1 in case of an
-	 *         error or if the source parameter was null
+	 * @param source   Key object that is used as source
```
markus2330: Did you try to build the docu? Does it look nice? https://doc.libelektra.org/api/current/html/classorg_1_1libelektra_1_1KeySet.html

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
```diff
@@ -472,12 +453,11 @@ public Key currentMeta() {
...
+	 * @param metaName Key name of the meta to be copied
+	 * @return 1 if meta was successfully copied, 0 if source doesn't contain the required meta and nothing had to be
```
markus2330: should be @retval here

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/KeySet.java
```diff
@@ -40,10 +40,14 @@ protected KeySet(final long p) {
...
 	/**
 	 * Helper constructor for duplication by pointer
+	 * The pointer may not be null or an IllegalArgumentException is thrown
 	 *
 	 * @param p Pointer to another KeySet
 	 */
-	protected KeySet(final Pointer p) {
+	public KeySet(final Pointer p) {
+		if (p == null) {
+			throw new IllegalArgumentException("Passed pointer may not be null");
```
markus2330: Should be NullPointerException?

```suggestion
			throw new IllegalArgumentException("Passed pointer should not be null");
```

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/plugin/NativePlugin.java
```diff
@@ -0,0 +1,182 @@
...
+	private ElektraPlugin elektraPlugin;
+
+	/**
+	 * Constructor for loading an elektra plugin
+	 * @param pluginName The plugin name
+	 * @param errorKey The errorKey
+	 * @throws InstallationException if the plugin does not exist
+	 */
+	public NativePlugin(String pluginName, Key errorKey) throws InstallationException {
+		KeySet modules = KeySet.create();
```
markus2330: modules are still recreated every time, this should be in the module loader.

markus2330:
> The build server fails quite often

Yes, please improve #2855 

> think it is still ready to merge because all tests ran through

There are also later checks (webui, Debian packages, ...).

I'll merge it for now. Please further improve the JNA binding in a new PR.

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
Piankero: @*RetVal* is not a standard javadoc annotation (https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html). Do you still want it nonetheless?

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
Piankero: To me it looks nice:

src/bindings/jna/libelektra4j/src/main/java/org/libelektra/Key.java
markus2330: This is parsed by doxygen, so I do not see why javadoc is relevant here. Or do you want to build this with javadoc instead?

PR #2903, opened 2019-08-22
Draw Plugins: Support Space Characters in Repository Path

sanssecours:
I think this should be ready to merge. I do have one question though. Is it expected behavior that the generated image look differently, even if I use Docker and execute the current version of the script from the `master` branch?

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -115,7 +115,7 @@ you up to date with the multi-language support provided by Elektra.
 ## Scripts
 
 - The script [run_icheck](../../scripts/run_icheck) now also work correctly, if the last entry of [`icheck.suppression`](../../tests/icheck.suppression) does not end with a newline character. _(René Schwaiger)_
-- <<TODO>>
+- The script [`draw-all-plugins`](../../scripts/draw-all-plugins) now also works properly, if the repository path contains space characters. _(René Schwaiger)_
 - <<TODO>>
 
 ## Documentation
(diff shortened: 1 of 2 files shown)


markus2330:
I never tried to create the images with something else then with my computer. If your image looks better, we can use it.

sanssecours:
>  If your image looks better, we can use it.

I do not think it looks better. On my system Graphviz just seems to place the nodes slightly different. The result is an image that is taller, but less wide:

comparison

. I do not think that is a problem, though. On the other hand, the PDF is [cropped on my machine](https://github.com/ElektraInitiative/libelektra/files/3529925/cropped.pdf). Using the command:

```sh
fdp -Tpdf "${OUTPUT}.flatten" -o "$SOURCE/doc/images/plugins.pdf"
```

instead of Ghostscript would [fix the problem](https://github.com/ElektraInitiative/libelektra/files/3529922/fixed.pdf). However, the comment:

```sh
# bad fonts
```

seems to suggest that there are problems with the generated PDF. For me the font does looks okay. Does the PDF output of `fdp` still look bad on your machine?

markus2330:
Different fonts could explain the different result.

I like taller results more, as they better fit to https://www.libelektra.org/plugins/readme

sanssecours:
> Different fonts could explain the different result.

I just checked. The font seems to be the same. 

Comparison

The only difference is probably that your machine (right image) uses [sub-pixel antialiasing](https://en.wikipedia.org/wiki/Subpixel_rendering), while mine does not (left image).

Issue #2913, opened 2019-08-25
Example Web UI Instance is Down


sanssecours:
# Steps to Reproduce the Problem

Open the following link:

- https://webdemo.libelektra.org

.

# Expected Result

The URL above should show an example instance of the [Web UI](https://master.libelektra.org/src/tools/web/README.md).

# Actual Result

The web server reports the error “502 Bad Gateway”.

kodebach:
The server no longer returns a 502 Error. Instead the request just times out.

Since our link-checker script didn't use a timeout and instead our build jobs timed out, I added the timeout in #2914.

kodebach:
After further investigation it turns out the docker registry https://hub-public.libelektra.org is also down. As far as I can tell they are on the same server (at least via the same IP), so probably the whole server crashed.

kodebach:
> the docker registry https://hub-public.libelektra.org is also down

The Jenkins server also doesn't work because of that.

markus2330:
Thank you for reporting these problems!

I think "only" a7 is down, all the services you mentioned here are on a7. I asked our admin to restart the machine.

kodebach:
Did the restart already happen? If so, it didn't help...

markus2330:
The admin did not write me anything yet :cry:

markus2330:
a7 is up again. But I can reproduce now that https://webdemo.libelektra.org/ stopped working. I am afraid I caused it by #2907 :cry: 

Luckily @Mistreated will now work on this topic again (or more precise on #2530)

@kodebach: Can you check if https://hub-public.libelektra.org/ works again for you?

kodebach:
> Can you check if https://hub-public.libelektra.org/ works again for you?

Yes, it seems to be working again.

kodebach:
@markus2330 The server is down again. Maybe revert #2907, since it may be the cause for the crashes.

Chemin1:
https://hub-public.libelektra.org/ as well as https://webdemo.libelektra.org are down for me, too (connection timed out).

markus2330:
> @markus2330 The server is down again. Maybe revert #2907, since it may be the cause for the crashes.

The server crashes because the kernel/hardware is unstable. It has nothing to do with the code it compiles and runs. It occurs since we upgraded to Buster, so maybe the Buster kernel has some regression on this hardware.

> https://hub-public.libelektra.org/ as well as https://webdemo.libelektra.org are down for me, too (connection timed out).

[...]

PR #2915, opened 2019-08-26
Type: disabling restore

kodebach:
## Basics

## Checklist

- [x] I added unit tests
- [x] I added code comments, logging, and assertions (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [x] meta data is updated (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))

## Review

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -50,11 +50,10 @@ The following section lists news about the [modules](https://www.libelektra.org/
 
 - We removed the plugin in favor of [Yan LR](../../src/plugins/yanlr/README.md). _(René Schwaiger)_
 
-### <<Plugin2>>
+### Type
 
-- <<TODO>>
-- <<TODO>>
-- <<TODO>>
+- We added an option to disable the restoring of boolean values. This useful for storage formats like YAML that have
+  native boolean types. _(Klemens Böswirth)_
 
 ### <<Plugin3>>
 
--- a/src/plugins/type/README.md
+++ b/src/plugins/type/README.md
@@ -115,17 +115,22 @@ You can also change how values shall be restored in `kdbSet`:
 sudo kdb mount typetest.dump user/tests/type dump type booleans=#0 booleans/#0/true=true booleans/#0/false=false boolean/restoreas=#0
 ```
 
-The config key `boolean/restoreas` must be a valid index of the `booleans` array. If `boolean/restoreas` was set the chosen boolean pair
-will be used when values are restored in `kdbSet`. So in the above example the plugin accepts `1`, `true`, `0` and `false` as boolean values,
-on `kdbGet` it turns `true` into `1` and `false` into `0` and on `kdbSet` it turns `1` into `true` and `0` into `false`.
+The config key `boolean/restoreas` must be a valid index of the `booleans` array or the special value `none`. If
+`boolean/restoreas` was set to an index, the chosen boolean pair will be used when values are restored in `kdbSet`. So
+in the above example the plugin accepts `1`, `true`, `0` and `false` as boolean values, on `kdbGet` it turns `true` into
+`1` and `false` into `0` and on `kdbSet` it turns `1` into `true` and `0` into `false`.
 
-If no `booleans` array was given the allowed values for `boolean/restore` are:
+If no `booleans` array was given the allowed values for `boolean/restoreas` are:
 
 - `#0` for `yes`/`no`
 - `#1` for `true`/`false`
 - `#2` for `on`/`off`
 - `#3` for `enabled`/`disabled`
 - `#4` for `enable`/`disable`
+
+The special value `boolean/restoreas=none` completely disables the restore procedure. In other words, `kdbSet` will
+always return either `0` or `1` for boolean values. This is useful, if a storage format with buitlin support for boolean
+values is used.
 
 ### Enums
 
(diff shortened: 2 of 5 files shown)


kodebach:
@sanssecours I hope this is what you need for you YAML plugins.

sanssecours:
> @sanssecours I hope this is what you need for your YAML plugins.

Yep, looks good as far as I can tell. Thank you 💖.

doc/news/_preparation_next_release.md
```diff
@@ -50,11 +50,10 @@ The following section lists news about the [modules](https://www.libelektra.org/
 
 - We removed the plugin in favor of [Yan LR](../../src/plugins/yanlr/README.md). _(René Schwaiger)_
 
-### <<Plugin2>>
+### Type
 
-- <<TODO>>
-- <<TODO>>
-- <<TODO>>
+- We added an option to disable the restoring of boolean values. This useful for storage formats like YAML that have
```
markus2330: You should either:

- mention here that the option was renamed
- not rename it, I think boolean/restore is still perfectly suitable even if it gets the additional option "none".

src/plugins/type/README.md
```diff
@@ -115,18 +115,23 @@ You can also change how values shall be restored in `kdbSet`:
 sudo kdb mount typetest.dump user/tests/type dump type booleans=#0 booleans/#0/true=true booleans/#0/false=false boolean/restoreas=#0
 ```
 
-The config key `boolean/restoreas` must be a valid index of the `booleans` array. If `boolean/restoreas` was set the chosen boolean pair
-will be used when values are restored in `kdbSet`. So in the above example the plugin accepts `1`, `true`, `0` and `false` as boolean values,
-on `kdbGet` it turns `true` into `1` and `false` into `0` and on `kdbSet` it turns `1` into `true` and `0` into `false`.
+The config key `boolean/restoreas` must be a valid index of the `booleans` array or the special value `none`. If
```
markus2330: What is the default?

src/plugins/type/README.md
```diff
@@ -115,18 +115,23 @@ You can also change how values shall be restored in `kdbSet`:
...
+If no `booleans` array was given the allowed values for `boolean/restoreas` are:
 
 - `#0` for `yes`/`no`
 - `#1` for `true`/`false`
 - `#2` for `on`/`off`
 - `#3` for `enabled`/`disabled`
 - `#4` for `enable`/`disable`
 
+The special value `boolean/restoreas=none` completely disables the restore procedure. In other words, `kdbSet` will
+always return either `0` or `1` for boolean values. This is useful, if a storage format with buitlin support for boolean
```
markus2330: buitlin?

src/plugins/type/types.c
```diff
@@ -104,11 +104,11 @@ bool elektraTypeNormalizeBoolean (Plugin * handle, Key * key)
 	const char * origTrueValue;
 	const char * origFalseValue;
 
-	bool restore = data->booleanRestore >= 0;
+	bool restoreAs = data->booleanRestore >= 0;
```
markus2330: Inconsistent naming: restoreAs on the left side, Restore on the right side.

src/plugins/type/type.c
```diff
@@ -203,10 +203,15 @@ static kdb_long_long_t readBooleanRestore (KeySet * config)
 
 	const char * restoreString = keyString (restore);
 
+	if (strcmp (restoreString, "none") == 0)
+	{
+		return -2;
```
markus2330: It would be nice if readBooleanRestore has some docu, e.g. from the meaning of the return values.

doc/news/_preparation_next_release.md
kodebach: The option was always called `boolean/restoreas`. There was just a typo in the README.

The relevant line in the code remained unchanged:

https://github.com/ElektraInitiative/libelektra/blob/9420c349af265cc427a98fad30466fc0d696c619/src/plugins/type/type.c#L198

markus2330: Ahh, ok. Then it is okay as-is. A bit more info in the release notes would not harm.

Nevertheless, I would prefer boolean/restore.

src/plugins/type/README.md
kodebach: The default is described above the code snippet. By default we restore the value the user chose,  `boolean/restoreas` is only used to modify the behaviour.

doc/news/_preparation_next_release.md
kodebach: I chose the name `restoreas`, because we always restore (except with the new `none` option). The config key just changes _how_ we restore.

Since it would be a breaking change to rename it now and there is no actual need for it, we will leave it as is.

kodebach:

doc/news/_preparation_next_release.md
markus2330: The exception is essential: now the option decides whether to restore at all, so "boolean/restore" would fit better.

> Since it would be a breaking change to rename it now and there is no actual need for it, we will leave it as is.

Within 0.9 we can do such changes. A simple sed should do, doesn't it?

kodebach:
@markus2330 I think this can be merged, the link-checker won't work because of #2913 (without a rebase, it will also not timeout until Cirrus kills it)

PR #2939, opened 2019-09-08
YAMBi: Remove Plugin

sanssecours:


markus2330:
Why do you want to remove this plugin? I thought the top candidates stay?

sanssecours:
> Why do you want to remove this plugin? I thought the top candidates stay?

It is not my top candidate, but of course I can leave it as is too.

PR #2916, opened 2019-08-26
Move conversion functions from highlevel to ease

kodebach:
## Basics

- [x] Short descriptions should be in the release notes (added as entry in
      `doc/news/_preparation_next_release.md` which contains `_(my name)_`)
      **Please always add something to the the release notes.**
- [x] Longer descriptions should be in documentation or in design decisions.
- [x] Describe details of how you changed the code in commit messages
      (first line should have `module: short statement` syntax)
- [x] References to issues, e.g. `close #X`, should be in the commit messages.

## Checklist

- [ ] I added unit tests
- [x] I ran all tests locally and everything went fine
- [ ] affected documentation is fixed
- [ ] I added code comments, logging, and assertions (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] meta data is updated (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))

## Review

- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)
- [ ] Documentation is concise and good to read
- [ ] Examples are well chosen and understandable

PR #2944, opened 2019-09-09
Codegen: Ignore missing required keys in help mode

kodebach:
If the user starts an application that has required keys with `-h`/`--help`, it would result in an error and the help message wouldn't be printed. This changes that.

It also adds the regenerated man page that is missing from #2942 (should've been part of 
b18447e).

## Basics

## Checklist

## Review

- [ ] Documentation is introductory, concise and good to read

### DIFF
--- a/doc/decisions/highlevel_help_message.md
+++ b/doc/decisions/highlevel_help_message.md
@@ -1,0 +1,54 @@
+# High-level API Help Message
+
+This decision _does not_ assume code-generation is used. For the case of code-generation see the [Notes](#notes) section.
+
+## Problem
+
+We want to allow to print the help message no matter what errors happened in `kdbOpen` or `kdbGet`.
+
+## Constraints
+
+- `elektraOpen` should not return a broken `Elektra` instance.
+- The help message can only be printed, if `elektraOpen` returns an `Elektra` instance and no `ElektraError`.
+
+## Assumptions
+
+- We assume that the application in question was correctly installed.
+- We assume `gopts` was mounted.
+- We assume the application was called in _help mode_, i.e. with `-h` or `--help`.
+  Otherwise printing the help message is not possible, anyway.
+
+## Considered Alternatives
+
+- Ignore all errors (in help mode):
+  Not a feasible solution, because there may have been problems when reading the storage file and therefore,
+  the help message may be broken or incomplete.
+- Ignore all errors (in help mode), which occurred after the `gopts` plugin ran:
+  Complicated to implement (we need to know about plugin order, etc.).
+  Not actually necessary (see [Rationale](#rationale)).
+
+## Decision
+
+Ignore missing `require`d keys (in help mode), but fail for every other error.
+
+## Rationale
+
+Required keys **must** be provided by the user/admin and cannot come from another source (Elektra, app developer, etc.).
+Therefore they will be missing until the user makes changes to the KDB. Before that, no other error should occur (we
+assumed a correct installation). If a user runs `app` for the first time and receives an error about a missing required
+key, they will:
+
+- know what to do and add the key, thereby fixing the problem.
+- try `app -h`/`app --help` to find out more. The help message may or may not contain useful information. If not they may try 3.
+- read some other documentation to find out more. Ideally this leads them to 1.
+
+In any case after this the user definitely know how to interact with the KDB. Since we assumed that there won't be any
+errors before the KDB was changed, we can assume that the user caused other errors by changing the KDB.
+
+## Notes
+
+If code-generation is used, the situation is a little different. If the parameter `embedHelpFallback` is set to `1`, a
+fallback help message will be created from the specification originally passed to the code-generator and embedded into
+the application. The parameter also changes, how help mode is detected and ultimately allows the help message function
+(`printHelpMessage` by default) to always print a help message. Although it may not reflect changes the user made to the
+specification.
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -38,7 +38,7 @@ If you specifically want to use it with the High-Level API take a look at [this 
 
 We also created a new CMake function that will be available, if you include Elektra via CMake's
 `find_package`. The function is called `elektra_kdb_gen` and can be used to tell CMake about files
-that are generated with `kdb gen`. _(Klemens Böswirth)_
+that are generated via `kdb gen`. _(Klemens Böswirth)_
 
 ### <<HIGHLIGHT2>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -80,6 +80,10 @@ The following section lists news about the [modules](https://www.libelektra.org/
 
 - We now treat relative paths as relative to `KDB_DB_SPEC` instead of the current working directory. _(Klemens Böswirth)_
 
+### Spec
+
+- There is now the config key `missing/log` that allows logging of all missing `require`d keys. _(Klemens Böswirth)_
+
 ## Libraries
 
 The text below summarizes updates to the [C (and C++)-based libraries](https://www.libelektra.org/libraries/readme) of Elektra.
--- a/src/libs/ease/symbols.map
+++ b/src/libs/ease/symbols.map
@@ -10,7 +10,9 @@ libelektra_0.8 {
 	elektraKeyGetRelativeName;
 	elektraKsFilter;
 	elektraResolveReference;
+};
 
+libelektra_0.9 {
 	## ToString;
 	elektraBooleanToString;
 	elektraCharToString;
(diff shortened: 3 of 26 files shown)


kodebach:

doc/news/_preparation_next_release.md
```diff
@@ -80,6 +80,10 @@ The following section lists news about the [modules](https://www.libelektra.org/
 
 - We now treat relative paths as relative to `KDB_DB_SPEC` instead of the current working directory. _(Klemens Böswirth)_
 
+### Spec
+
+- There is now the config key `missing/log` that allows logging of all missing `require`d keys. _(Klemens Böswirth)_
```
markus2330: Shouldn't we fail if a required key is missing? Why do we need the log as this should already be part of the error message?

kodebach: 1. `spec` can be configured to ignore some or all errors. I don't really know why, but it's always been the case.
2. This is needed for the case where a program using the codegen API is called in "help mode", i.e. `app -h` or `app --help`. In such a case, we want to ignore missing required keys and simply print the help message.
How the `missing/log` array helps detect this case, is documented in the relevant code:
https://github.com/ElektraInitiative/libelektra/blob/ea74e58203edeaf9175d65f091753d286887da45/src/libs/highlevel/elektra.c#L143-L158
There could technically be other problems from the `spec` plugin as well, but since `gopts` runs first and was successful, there can't be any errors that would break the help message.

markus2330: > spec can be configured to ignore some or all errors. I don't really know why, but it's always been the case.

This is one of spec's anti-features. Thank you for documenting it but I would be in favor of removing these anti-features. If validation fails, the plugin should return an error.

> This is needed for the case where a program using the codegen API is called in "help mode"

Ok. But wouldn't it be better to directly ask for a meta-key which says that we are in the help mode? We want the help to be printed no matter what errors happened in `kdbOpen` or `kdbGet`.

kodebach: > We usually only add warnings during kdbGet()

There are many cases, where `kdbGet()` should return an error. The best example would be a broken storage file. The storage plugin should return an error and `kdbGet()` should abort.

For the `spec` plugin, it probably makes sense to have to option of switching between warnings and errors. In `kdb` (and other configuration tools) warnings would be preferred, as a `kdbGet()` is required before a `kdbSet()` could fix the problems. In applications on the other hand errors are the better solution. They are easier to detect, since `kdbGet()` uses a different return code for errors. The handling of actual errors is also easier in the high-level API (see #2952).

> Can you make a decision of what you wrote here?

Will do.

markus2330: > There are many cases, where kdbGet() should return an error.

I agree. I made a different proposal: #2955

> For the spec plugin, it probably makes sense to have to option of switching between warnings and errors.

I hope this will not be needed once we have #2955 implemented

>   Will do.

:sparkling_heart:

kodebach:
Side note: The fallback help message only works with `app -h` or `app --help`, but not with combined short options like `app -fh`. The reason is simply, that without the spec we don't know which option takes arguments and which doesn't (`app -fh` might be the same as `app -f -h` or `app -f h`). I think this limitation is acceptable.

doc/decisions/highlevel_help_message.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Constraints
+
+- `elektraOpen` should not return a broken `Elektra` instance.
+- The help message can only be printed, if `elektraOpen` returns an `Elektra` instance and no `ElektraError`.
+
+## Assumptions
+
+- We assume that the application in question was correctly installed.
+- We assume `gopts` was mounted.
```
markus2330: Is this already default?

src/libs/ease/symbols.map
```diff
@@ -10,7 +10,9 @@ libelektra_0.8 {
 	elektraKeyGetRelativeName;
 	elektraKsFilter;
 	elektraResolveReference;
+};
 
+libelektra_0.9 {
```
markus2330: Why this change?

src/plugins/gopts/gopts.c
```diff
@@ -43,6 +72,7 @@ int elektraGOptsGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned, Key * pa
 			ksNew (30, keyNew ("system/elektra/modules/gopts", KEY_VALUE, "gopts plugin waits for your orders", KEY_END),
 			       keyNew ("system/elektra/modules/gopts/exports", KEY_END),
 			       keyNew ("system/elektra/modules/gopts/exports/get", KEY_FUNC, elektraGOptsGet, KEY_END),
+			       keyNew ("system/elektra/modules/gopts/exports/ishelpmode", KEY_FUNC, elektraGOptsIsHelpMode, KEY_END),
```
markus2330: isinhelpmode or simply helpmode?

src/plugins/gopts/gopts.c
```diff
@@ -34,6 +34,35 @@ static void cleanupEnvp (char ** envp);
...
+	int argc = loadArgs (&argv);
+
+	if (argv == NULL)
+	{
+		return -1;
+	}
+
+	for (int i = 0; i < argc; ++i)
+	{
+		if (strcmp (argv[i], "-h") == 0 || strcmp (argv[i], "--help") == 0)
```
markus2330: I do not like that this is hardcoded here. `-H` is very common (and also used in the `kdb` tool). `-h` has man other meanings.

Furthermore this is not properly parsed, e.g. `-vh` would not be recognized.

src/libs/ease/symbols.map
kodebach: It should have been part of #2916. The functions were not part of any 0.8.x release, so they should not be versioned as `libelektra_0.8`.

If we won't have another 0.9.x release and go straight to 1.0.0, they should actually be in `libelektra_1.0`.

markus2330:
Thank you!

doc/decisions/highlevel_help_message.md
kodebach: `gopts` is not mounted by default no. However, the file generated by the `highlevel` template, include a `kdbEnsure` contract, that mounts it, if it is missing.

src/plugins/gopts/gopts.c
kodebach: > I do not like that this is hardcoded here. `-H` is very common (and also used in the `kdb` tool). `-h` has man other meanings.

Using `-h`/`--help` for help mode is already hard-coded in `elektraGetOpts` (there it is properly parsed).

This function is not actually intended to accurately detect help mode (for that one should still check for the presence of `proc/elektra/gopts/help`. It is only meant in case `kdbGet()` failed.

Apart from `kdb` tool, I have never seen `-H` used in favour of `-h`. The `kdb` tool also takes a very unusual approach, since it opens the man-page instead of printing to stdout/stderr.
In fact most tools I know, use only `--help` (if they support long options).

> Furthermore this is not properly parsed, e.g. `-vh` would not be recognized.

See above https://github.com/ElektraInitiative/libelektra/pull/2944#issuecomment-531577466.

src/plugins/gopts/gopts.c
kodebach: Since you already merged this PR: Do you still want the change?

doc/decisions/highlevel_help_message.md
markus2330: Please copy this sentence into the decision.

src/libs/ease/symbols.map
markus2330: We will definitely have a 0.9 release before 1.0.

src/plugins/gopts/gopts.c
markus2330: It was only a question. If you like ishelpmode best, we can keep it.

src/plugins/gopts/gopts.c
markus2330: > Apart from kdb tool, I have never seen -H used in favour of -h.

Because of how the cmdline options in Elektra are implemented every option needs to have a short option. `-H` was the most natural one as `-h` is already "human readable".

> The kdb tool also takes a very unusual approach, since it opens the man-page instead of printing to stdout/stderr.

It makes a lot of sense to not have a --help output and a separated man page. E.g. git also does it this way.  

> In fact most tools I know, use only --help (if they support long options).

I would prefer if we only hardcode `--help` and drop support for `-h`.

markus2330:
> Side note: The fallback help message only works with app -h or app --help, but not with combined short options like app -fh. The reason is simply, that without the spec we don't know which option takes arguments and which doesn't (app -fh might be the same as app -f -h or app -f h). I think this limitation is acceptable.

Please make sure that such notes are documented in the documentation. Here nobody will read it.

kodebach:
It is documented in the doxygen comment for `elektraGOptsIsHelpMode`, I just added the note here, in case you miss the comment.

markus2330:
Sorry, I do not see it. You mean src/plugins/gopts/gopts.c line 37-44?

kodebach:
> You mean src/plugins/gopts/gopts.c line 37-44?

Yes

Issue #2963, opened 2019-09-15
backend plugin


vLesk:
Building upon the implementation of `elektraPluginCommit` in #2798, I would like to reimplement the `Backend` structure as a plugin. That would bring the following benefits:

- Greater detachment from the core of `libelektra`, as `KDB` would only need to call the respective plugin functions,
- Nesting plugins would be possible as backends could contain further backends.

Another problem I would like to solve with this implementation is that backends have a limited fixed amount of plugins that they can contain. 

To solve this, I propose restructuring the plugin arrays (`getplugins`, `setplugins` and `errorplugins`) to arrays of linked lists. That way, plugins assigned to a specific role are added to the end of the list at the position corresponding to the role, with the possibility of a flexible number of further plugins being added to that role.

Due to complexity issues with operations such as file resolving and change recognition, the following roles will still only contain a single slot in their respective arrays:

- `getresolver`
- `getstorage`
- `rollback`
- `setresolver`
- `setstorage`
- `commit`

Implementing backends as plugins has already been discussed in #973 and #1274.

markus2330:
Thank you for the summary!

markus2330:
@vLesk What is the current status?

vLesk:
As far as I can see, the implementation should be done (the `ensurePluginUnmounted` function in `kdb.c` is commented in though as we discussed a few months ago, additional work would need to be done for that to be supported), I'm still working on fixing the bugs. I've made changes to `mount.c` and further changes to the backend-related .cpp files. I'm currently working full time but I have some days off I can use, and in September I'll be working reduced hours so it should be easier to work out the rest of the bugs.

markus2330:
Thank you, it would be so great to get this finished. :sparkling_heart: 

It is really one of the core features of Elektra and we constantly run into the too many plugin problem. The ensurePluginUnmounted is not a big deal, as discussed before. Maybe you can throw a not_implemented error there?

PR #2969, opened 2019-09-16
backend plugin

vLesk:
Implementing the `backend` plugin. Still under construction :)

Discussed in #2963

## Basics

## Checklist

need to be checked.

## Review

- [ ] Documentation is introductory, concise and good to read

## Labels

- Add the "ready to merge" label **if the build server is happy** and also you


doc/decisions/backend_plugin.md
```diff
@@ -0,0 +1,45 @@
...
+## Constraints
+
+- It should be possible for all existing plugins to run normally
+
+## Assumptions
+
+## Considered Alternatives
+
+- Multiple storage plugins within a single backend
+- Plugin containing more plugin slots
```
markus2330: You can link to the list plugin here.

doc/decisions/backend_plugin.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Rationale
+
+- Making backends plugins themselves detaches their implementation from the core of Elektra, making it possible to develop new kinds of backends
+  without major changes to the core itself.
+- As plugins, backends can contain further backends, making it possible to nest plugins and enabling new kinds of plugin combinations such as
+  fallback storage options.
+
+## Implications
+
```
markus2330: One important implication is that we need a hard-coded default backend, similar to the default storage and resolver we already have.

vLesk:
I'm currently working on moving backendOpen in backend.c to elektraBackendOpen in the backend plugin, one notable function which is called there is elektraProcessPlugins, which builds up the plugin arrays. Since this is implementation specific for backends (correct me if I'm wrong), the idea would be to move the function to the backend plugin and call it there (modified to work with the new array structure). Since elektraPluginOpen and elektraProcessPlugin are exported, they can be called there as well. 

Most of the arguments the function needs can be accessed from within the plugin as well:
-plugins: is the plugin array and is hence accessible
-reference plugins: given to the function as an empty keyset
-config & systemConfig: given to the plugin
-global: given to the plugin
-errorKey: given as argument to elektraBackendOpen

The only argument I'm not sure about is the KeySet modules, since it is neither given to the plugin nor as an argument, but is necessary to call elektraPluginOpen. 

My idea would be to give the keyset to the plugin using the data field which is available in the plugin struct and make it possible to access the keyset that way. Is that a valid way to solve this or is this too much of a workaround?

markus2330:
> Since this is implementation specific for backends

exactly

> the idea would be to move the function to the backend plugin and call it there

Maybe we should get rewrite of this code. If you add any number of plugins in the middle of this code, it might be very confusing.

Did you already design how the new system/elektra/mountpoints will look like?

> My idea would be to give the keyset to the plugin using the data field which is available in the plugin struct and make it possible to access the keyset that way. Is that a valid way to solve this or is this too much of a workaround?

It is a good idea but use a new `KeySet * modules` in _Plugin and not the data field. Loading of modules is a common task for many plugins (multiresolver, language binding plugins, ...). So if the modules are somehow accessible from within plugins, it is a good thing.

vLesk:
The new mountpoints will have the mountpoint key moved under config/mountpoint, like config/path, that way they can be assigned to the plugins and read out. The easiest solution would be to add the mountpoint variable to the plugin struct, but I'm not sure if it is useful enough for other plugins like the modules keyset.

Another thing that needs to be changed in the mountpoint configuration is the allocation of plugins in the arrays, since array numbers are currently being used to define a plugin's respective slot.  I would like to change the numbers to the actual names of the roles. For example:

`setplugins/#3#plugin#name`

is changed to 

`setplugins/setstorage/#plugin#name`.

That way it is possible to have more flexible numbers of plugins. This means that, among other things, I need to change the default configuration, as well as the configuration created when new backends are mounted. However, I am not sure where I can find these configurations, I was able to find `elektra.ecf` in /etc, but I'm not sure where it is created from.

markus2330:
Thank you, can you please write this in the documentation?

> setplugins/setstorage/#plugin#name.

Do we need the `#` here? I think that it would be better if we have `#` only for arrays, e.g. `setplugins/presetstorage/#0/plugin`?

> This means that, among other things, I need to change the default configuration, as well as the configuration created when new backends are mounted. However, I am not sure where I can find these configurations, I was able to find elektra.ecf in /etc, but I'm not sure where it is created from.

It is created from the `kdb` tool which uses the libtools. In the source it is src/libs/tools/src/backend.cpp

With the tests it should be easy to understand what the code does. outputGTest allows you to generate regression tests. Current regression tests (you will need to fix them) are in src/libs/tools/testtool_backend.cpp

vLesk:
Thank you for the answers! I agree that the `#` should be used only for arrays, I can implement the configuration that way. What I need to do in that case is find a new place for plugin labels. A good solution in my opinion would be something like `setplugins/presetstorage/#0/plugin/label`.

markus2330:
Can you write up everything including a full example? It is easier to see how it looks like then.

vLesk:
I'll use the mount point configuration from `doc/dev/architecture.md` as an example. The changed configuration would then look as follows:

```
system/elektra/mountpoints
system/elektra/mountpoints/fstab
system/elektra/mountpoints/fstab/config
system/elektra/mountpoints/fstab/config/mountpoint
system/elektra/mountpoints/fstab/config/path=fstab
system/elektra/mountpoints/fstab/config/struct=list FStab
system/elektra/mountpoints/fstab/config/struct/FStab
system/elektra/mountpoints/fstab/config/struct/FStab/device
system/elektra/mountpoints/fstab/config/struct/FStab/dumpfreq
system/elektra/mountpoints/fstab/config/struct/FStab/mpoint
system/elektra/mountpoints/fstab/config/struct/FStab/options
system/elektra/mountpoints/fstab/config/struct/FStab/passno
system/elektra/mountpoints/fstab/config/struct/FStab/type
system/elektra/mountpoints/fstab/errorplugins
system/elektra/mountpoints/fstab/errorplugins/rollback
system/elektra/mountpoints/fstab/errorplugins/rollback/#0
system/elektra/mountpoints/fstab/errorplugins/rollback/#0/label (="resolver")
system/elektra/mountpoints/fstab/errorplugins/rollback/#0/name (="resolver")
system/elektra/mountpoints/fstab/getplugins
system/elektra/mountpoints/fstab/getplugins/getresolver
system/elektra/mountpoints/fstab/getplugins/getresolver/#0
system/elektra/mountpoints/fstab/getplugins/getresolver/#0/ref (="resolver")
system/elektra/mountpoints/fstab/getplugins/getstorage
system/elektra/mountpoints/fstab/getplugins/getstorage/#0
system/elektra/mountpoints/fstab/getplugins/getstorage/#0/label (="fstab")
system/elektra/mountpoints/fstab/getplugins/getstorage/#0/name (="fstab")
system/elektra/mountpoints/fstab/setplugins
system/elektra/mountpoints/fstab/setplugins/commit
system/elektra/mountpoints/fstab/setplugins/commit/#0
system/elektra/mountpoints/fstab/setplugins/commit/#0/ref (="resolver")
system/elektra/mountpoints/fstab/setplugins/presetstorage
system/elektra/mountpoints/fstab/setplugins/presetstorage/#0
system/elektra/mountpoints/fstab/setplugins/presetstorage/#0/label (="struct")
system/elektra/mountpoints/fstab/setplugins/presetstorage/#0/name (="struct")
system/elektra/mountpoints/fstab/setplugins/presetstorage/#1
system/elektra/mountpoints/fstab/setplugins/presetstorage/#1/label (="type")
system/elektra/mountpoints/fstab/setplugins/presetstorage/#1/name (="type")
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2/label (="path")
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2/name (="path")
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2/config
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2/config/path/allow=proc tmpfs none
system/elektra/mountpoints/fstab/setplugins/setresolver
system/elektra/mountpoints/fstab/setplugins/setresolver/#0
system/elektra/mountpoints/fstab/setplugins/setresolver/#0/ref (="resolver")
system/elektra/mountpoints/fstab/setplugins/setstorage
system/elektra/mountpoints/fstab/setplugins/setstorage/#0
system/elektra/mountpoints/fstab/setplugins/setstorage/#0/ref (="fstab")
```
The notable changes are the following:

- The key containing the mountpoint has been moved from `system/elektra/mountpoints/backendname/mountpoint` to `system/elektra/mountpoints/backendname/config/mountpoint`. That way, the mountpoint of the backend can still be read out in the core from the backend plugin's plugin configuration. 
- Plugin roles are no longer displayed as array slots, but actually by their names. For example, `system/elektra/mountpoints/fstab/getplugins/#0` is now `system/elektra/mountpoints/fstab/getplugins/getresolver`.
- Each plugin role consists of a linked list containing the plugins fulfilling this role. In the configuration, the position of the plugin in the linked list is shown using an array. For example, the key `system/elektra/mountpoints/fstab/setplugins/presetstorage/#2` means that the plugin (in this case `path`) belongs to the second position of the linked list belonging to the `presetstorage` role.
- The name, reference name and the label of a plugin are now stored in separate keys to avoid using the `#` symbol for something other than arrays. An example from the `path` plugin:

```
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2/label
system/elektra/mountpoints/fstab/setplugins/presetstorage/#2/name
```
That way, the `path` plugin is opened and stored for later use with the defined label. If it were to be used later, it would be referenced by adding `ref` instead of `label` and `name`.

vLesk:
I made a rough implementation of elektraBackendOpen. I am currently working on implementing the other plugin functions, as well as writing appropriate tests and documentation.

markus2330:
Can you put all the text you put here directly as text into the docu? It is much easier to review then (can also be in a separate PR).

> system/elektra/mountpoints/fstab

fstab is not a valid mountpoint. It must be \\/fstab, user\\/fstab or similar (at least one \\/ must be present to be valid)

> system/elektra/mountpoints/fstab/config/path=fstab

Why is =fstab here not in ()?

> struct

The struct plugin was removed and fstab is not finished, so better you use some other example (e.g. hosts with glob).

> That way, the mountpoint of the backend can still be read out in the core from the backend plugin's plugin configuration.

Very nice idea.

> Plugin roles are no longer displayed as array slots, but actually by their names. For example, system/elektra/mountpoints/fstab/getplugins/#0 is now system/elektra/mountpoints/fstab/getplugins/getresolver.

Maybe we can get rid of some redundancy here. `system/elektra/mountpoints/fstab/get/resolver` should be enough (getplugins -> get and getresolver -> resolver).

> The name, reference name and the label of a plugin are now stored in separate keys to avoid using the # symbol for something other than arrays.

This is a big step forward. Did you also consider to completely separate the definitions of plugins and their use in the positions?

> it would be referenced by adding ref instead of label and name.

I think @sanssecours will prefer reference instead of ref here :wink: 

@kodebach do you also want to share some thoughts here?

kodebach:
Everything here seems good to me.

> One important implication is that we need a hard-coded default backend, similar to the default storage and resolver we already have.

We should also ensure in CMake that there is no way to build Elektra without the default storage, resolver and backend plugins. I think right now, you could exclude them from the plugin list and get a broken build. At the least there should be an obvious warning from CMake.

markus2330:
Yes, I agree that CMake should not succeed until the default-backend plugin is included. For resolver/storage this should already be the case, see src/plugins/CMakeLists.txt lines 67 and 76. Please open an issue if you find a way how to fail during compilation.

kodebach:
> For resolver/storage this should already be the case, see src/plugins/CMakeLists.txt lines 67 and 76. Please open an issue if you find a way how to fail during compilation.

Those checks are only applied during `cmake` configure, not during `make`. They ensure that a `make` (or `make all`) works, but a `make kdb` will not (re)build the default storage/resolver, since there is no dependency in the `kdb` target for this.

I think adding something like `elektra-${DEFAULT_STORAGE}` and `elektra-resolver` here: 
https://github.com/ElektraInitiative/libelektra/blob/4db328f823bc86fe5d43df405329f103afd55789/src/tools/kdb/CMakeLists.txt#L30-L33

(and a few lines below for `kdb-full` and `kdb-static`) might be enough.

IMO `make kdb` should also build a functional Elektra. In fact it should build the absolute minimal version of Elektra.

markus2330:
> IMO make kdb should also build a functional Elektra. In fact it should build the absolute minimal version of Elektra.

I would not spent too much time on the build system. The sub-targets are for expert use only as we do not mention or encourage them anywhere.

But of course you are free to send patches :wink:

kodebach:
> The sub-targets are for expert use only as we do not mention or encourage them anywhere.

The [CLion tutorial](https://github.com/ElektraInitiative/libelektra/blob/master/doc/tutorials/contributing-clion.md) kind of mentions them. But yes, its a minor issue.

markus2330:
Thank you, I was not aware of that. Any specific reason why this is mentioned there?

kodebach:
> Any specific reason why this is mentioned there?

It actually just says that you can run individual tests by executing "any of the testmod_* or testkdb_*targets". In most cases that should work fine (especially `testmod_*`), at least after running `make` once to build everything. The changed parts should be rebuilt by CLion automatically.

markus2330:
>  It actually just says that you can run individual tests by executing "any of the testmod_* or testkdb_*targets".

At least on the command-line it is not like this. When I execute `make testmod_mini` it builds the target testmod_mini but it **does not** run the test.

> In most cases that should work fine (especially testmod_*), at least after running make once to build everything. The changed parts should be rebuilt by CLion automatically.

For me there is no noticeable difference between building everything vs. building a single target if everything was already build before. So why not simply always call the `all` target. Does someone use such a slow computer that this matters?

kodebach:
> When I execute make testmod_mini it builds the target testmod_mini but it does not run the test.

CLion automatically creates run configurations for all (executable) CMake targets. For target `X` these run configurations build `X` (i.e. `make X`) and then run the resulting executable. In the case of `testmod_mini` this is more or less like running `make testmod_mini && ./bin/testmod_mini`.

> For me there is no noticeable difference between building everything vs. building a single target if everything was already build before.

Depends what you modified. Modifications to some CMake files for example invalidate a big chunk of the project. If you use `make testmod_mini` only `testmod_mini` and its dependencies are built, `make all` meanwhile would rebuild everything that is out dated including entirely unrelated plugins.

> So why not simply always call the all target.

As stated above, the CLion run configurations are created automatically. You would have to manually change them to do that (or not use them).

> Does someone use such a slow computer that this matters?

Even on my Ryzen 7 3700X system (using `make -j 16`), I see a small difference between `make all` and `make kdb`, if for example I change something in `kdbGet()`.

Besides that, using e.g. `make testmod_mini` instead of `make all` may uncover unknown dependencies.

markus2330:
I moved the discussion to  #3055.

@vLesk do you already have some docu about the new mountpoint configs so that we can review that?

vLesk:
I wrote documentation on the new configs. I also implemented the other plugin functions. 

One issue I have right now is integrating the global plugins into the plugin functions. In order to access them I would need access to the KDB handle within the backend plugin itself. I can think of either passing the pointer to the handle using the global keyset, or adding it to the Plugin struct. I personally think that using the global keyset is better practice since it is tied to the KDB handle.

kodebach:
> One issue I have right now is integrating the global plugins into the plugin functions. 

As far as I understand, there will be exactly one instance of a backend plugin per mountpoint. Global plugins are called once for the whole KDB. Therefore they cannot be integrated into the backend plugin.

vLesk:
>As far as I understand, there will be exactly one instance of a backend plugin per mountpoint. >Global plugins are called once for the whole KDB. Therefore they cannot be integrated into the >backend plugin.

I didn't phrase that right, I'll try explaining it in a better way. What I meant was that, assuming I understood this correctly, certain global plugins in `foreach` positions are called while `kdbGet` or `kdbSet` is carried out in the backend. This means that I cannot call them before or after calling `kdbGet/kdbSet` within the backend plugin. This means that I need to access whatever instance of `KDB` called `kdbGet/kdbSet` in the backend plugin and call `kdbGet/kdbSet` in the global plugins which are supposed to be called at that position. Currently, this is done in `elektraSetPrepare` and `elektraGetDoUpdateWithGlobalHooks` in `kdb.c`.

kodebach:
> What I meant was that, assuming I understood this correctly, certain global plugins in `foreach` positions are called while `kdbGet` or `kdbSet` is carried out in the backend. [...]

Ah yes, that is true. But there is still a problem, even if you give the backend plugin access to the global plugins.

Let's take a `postgetstorage` `foreach` global plugin as an example. Since it is in the `foreach` position, it should be called for every mountpoint. Since it is a global plugin, it should always receive the full KDB KeySet, not just the KeySet for the current mountpoint. This means we must call all getplugins up to and including the `getstorage` position for all mountpoints. Then we must merge the keysets for all mountpoints. Next for each mountpoint, we call our global plugin with the merged keyset and then we extract the part of the keyset for the current mountpoint and call the non-global plugins. As you said, this is done in `elektraGetDoUpdateWithGlobalHooks` right now.

The best solution I see for this problem, is to call `elektraBackendGet` once per position instead of only once per mountpoint. The iteration over the positions will then happen in `kdbGet` as before. We should pass the current position to the plugin (e.g. via the `Plugin` handle), instead of relying on some mechanism inside the plugin (e.g. counting the `elektraBackendGet` calls).

markus2330:
@mpranj can you also take a look at the concept here (in particular for the global plugins as your plugins are the main customer of the global plugin interface)?

markus2330:
> In order to access them I would need access to the KDB handle within the backend plugin itself. 

Of course it would be nice (more modular) if the plugins would not need KDB. But we already had several occasions where the KDB handle would have been very handy. So if you do not find an elegant way to implement the backend plugin without the KDB handle, then you are allowed to pass the handle to the plugins. Necessary restrictions should be documented then (to avoid endless recursion and so on).

doc/decisions/backend_plugin.md
```diff
@@ -0,0 +1,45 @@
...
+- The current backend implementation will be redeveloped into a backend plugin. That way, the core of Elektra will only access backends through
+the standard plugin interface. 
+- The `getplugins`, `setplugins` and `errorplugins` arrays will be changed into arrays of linked lists. Each time a plugin is added to a specific
+slot, it will be added at the end of the linked list.
+
+## Rationale
+
+- Making backends plugins themselves detaches their implementation from the core of Elektra, making it possible to develop new kinds of backends
+  without major changes to the core itself.
```
mpranj: Can you give an example of a different kind of backend, which would be desirable to have?

vLesk:
After some consideration, I decided that I'm going to go with the proposal from @kodebach. If I access the global plugins from within the backend plugin, some roundabout solutions are going to have to be implemented, since I will also have to be able to recognize whether a backend is being used as a nested plugin and not execute the global plugins then and different KDBs will have to be set based on the user. I find it less of a compromise to define the behavior of kdbGet and kdbSet in the backend plugin to be called once per position and not have to deal with internal core issues.

doc/decisions/backend_plugin.md
vLesk: One example would be a backend using a different kind of structure for the plugin arrays, for example using arrays of arrays instead of arrays of linked lists, which would improve plugin search time but make the number of usable plugins in the backend fixed.

vLesk:
> I reviewed only the concept, not the code as the builds are not yet passing. Ping me if a code review is needed.
Thank you for the review!

vLesk:
I made modifications to the testtool_backend tests and the affected cpp files. I'm currently working on fixing my errors and adding things I may have missed. One problem I'm currently faced with (should be reproducable by running the testtool_backend tests with the changes I just pushed) is that most of the tests show the error message "unknown file: Failure", does that mean that there are files somewhere that I still need to modify or does it mean something else? Is anybody familiar with this error?

markus2330:
> show the error message "unknown file: Failure", does that mean that there are files somewhere that I still need to modify or does it mean something else? Is anybody familiar with this error?

I do not know (and as also nobody else replied it seems to be a new error) two guesses:

1. it might be a temporary error of the build server, try a restart of the build job
2. it might be a problem because files are conflicting, maybe a rebase to master helps
3. it might be a bug in GitHub, creating a new PR should help then

markus2330:
Looking at the current error of Jenkins again: now it seems to be the problem of "Conflicting files"

vLesk:
I tried rebasing the branch on master, it did not work, the problem definitely seems to be in my changes. I reimplemented the changes and still get the same error. Apparently the parse function in plugin.cpp does not add the "error" symbol for certain plugins, throwing a MissingSymbol exception when tryPlugin is called in plugins.cpp. I think I will set this aside for now, finish implementing the other parts and come back to it.

vLesk:
I also get the same error when running the test locally, so it is probably not a problem with GitHub or the build server.

kodebach:
You are missing the contract part of your plugin's get function:

https://github.com/ElektraInitiative/libelektra/blob/b75bc76133b5fb8dde3f4113a25262d0bd7dc91a/src/plugins/template/template.c#L33-L51

This declares all the symbols your plugin exports and this is why `plugin.cpp` cannot find the symbol called "error".

markus2330:
@kodebach thank you for helping out!

vLesk:
An update on my progress, I modified most of the backend functions in `libs/elektra/backend.c` as well as related parts of `split`, `trie`, `mount` and `plugin.c`. I am currently dealing with an issue in `split.c` and `backend.c`. 

At many points, the functions directly access the variables `specsize`, `dirsize`, `usersize` and `systemsize` contained within the `Backend` struct, meaning that I need some way of reading and modifying them from outside the `backend` plugin. A clean solution to this problem in my opinion is storing these variables as keys in the backend configuration. I would store them under `config`. An example, only showing the `config` part of the configuration:

```
system/elektra/mountpoints/\/hosts
system/elektra/mountpoints/\/hosts/config
system/elektra/mountpoints/\/hosts/config/dirsize
system/elektra/mountpoints/\/hosts/config/glob/set/#0
system/elektra/mountpoints/\/hosts/config/glob/set/#1
system/elektra/mountpoints/\/hosts/config/glob/set/#2
system/elektra/mountpoints/\/hosts/config/glob/set/#3
system/elektra/mountpoints/\/hosts/config/glob/set/#4
system/elektra/mountpoints/\/hosts/config/glob/set/#4/flags
system/elektra/mountpoints/\/hosts/config/mountpoint
system/elektra/mountpoints/\/hosts/config/path
system/elektra/mountpoints/\/hosts/config/specsize
system/elektra/mountpoints/\/hosts/config/systemsize
system/elektra/mountpoints/\/hosts/config/usersize
```
An issue I find it more difficult to find a clean solution to is changing the `ensurePluginUnmounted` function in `kdb.c`, which searches for a certain plugin in a backend and removes only it by directly accessing and modifying the plugin arrays, which is no longer possible with the `backend` plugin.

One solution I can think of is triggering this behavior in the set function of the `backend` plugin if a specific key is passed as an argument, but this is not the cleanest solution since it adds more state to the plugin functions of the `backend`. Does anybody have any better ideas as to how to modify this function?

kodebach:
> A clean solution to this problem in my opinion is storing these variables as keys in the backend configuration. I would store them under config. An example, only showing the config part of the configuration:

I'd use the global KeySet of the KDB handle not the actual configuration. However, I think we don't have to get that complicated. The sizes are actually of no interest to the plugin itself. AFAIK they are just used to quickly check, whether keys were added/removed between `kdbGet` and `kdbSet`.

I see that you removed the `struct _Backend` (which had the fields for the sizes) and `struct  _Split` now directly uses a `Plugin **`. If you look closely at where the sizes were used, you'll that they are only used in combination with the `struct _Split`. So we can just move the data there by adding five new arrays to the `struct _Split`

```c
ssize_t * specsize;
ssize_t * dirsize;
ssize_t * usersize;
ssize_t * systemsize;
```

Then you should just be able to replace e.g. `split->handles[i]->specsize` with `split->specsize[i]`. (You also have to make sure that the new arrays are always allocated to the same size as the `Plugin **` and the other arrays).

> An issue I find it more difficult to find a clean solution to is changing the ensurePluginUnmounted function in kdb.c, which searches for a certain plugin in a backend and removes only it by directly accessing and modifying the plugin arrays, which is no longer possible with the backend plugin.

That functionality hasn't existed for a very long time and I doubt, anybody is actually using it. I think, we could just remove it until we find a solution. We can certainly disable/comment out the one test case for it and implement it in another PR later on.

The cleanest solution I can think of, is requiring backend-type plugins to export one or more functions for manipulating the backend, e.g. `mountPlugin` and `unmountPlugin`. These would be exported like the `checkfile` function in resolver-type plugins.

markus2330:
> At many points, the functions directly access the variables specsize, dirsize, usersize and systemsize contained within the Backend struct, meaning that I need some way of reading and modifying them from outside the backend plugin.

Why is this a problem? The _Backend structure can be private to the backend plugin, can't it?

>  A clean solution to this problem in my opinion is storing these variables as keys in the backend configuration. I would store them under config. An example, only showing the config part of the configuration:

Yes, if this information needs to be shared this makes sense. As @kodebach said it could be in the global KeySet.

vLesk:
First of all, thank you for the reply, it's really helpful!

> I see that you removed the `struct _Backend` (which had the fields for the sizes) and `struct _Split` now directly uses a `Plugin **`. If you look closely at where the sizes were used, you'll that they are only used in combination with the `struct _Split`. So we can just move the data there by adding five new arrays to the `struct _Split`
> 
> ```c
> ssize_t * specsize;
> ssize_t * dirsize;
> ssize_t * usersize;
> ssize_t * systemsize;
> ```
> 
> Then you should just be able to replace e.g. `split->handles[i]->specsize` with `split->specsize[i]`. (You also have to make sure that the new arrays are always allocated to the same size as the `Plugin **` and the other arrays).

That's a very good solution! It also removes the need for toying around with the backend config.

> That functionality hasn't existed for a very long time and I doubt, anybody is actually using it. I think, we could just remove it until we find a solution. We can certainly disable/comment out the one test case for it and implement it in another PR later on.
> 
> The cleanest solution I can think of, is requiring backend-type plugins to export one or more functions for manipulating the backend, e.g. `mountPlugin` and `unmountPlugin`. These would be exported like the `checkfile` function in resolver-type plugins.

I guess commenting out the functionality is a valid short term solution for this PR, but I'll leave it in so that a more proper solution can be implemented in another PR.

vLesk:
> Why is this a problem? The _Backend structure can be private to the backend plugin, can't it?
Yes, it is private to the plugin, but up until now, the variables were in the structure and used directly by the core from the outside, which is why a solution was needed to access these variables without making the structure public.

vLesk:
I'm currently working on getting the `testmod_backend` tests to work, and I just came across a problem I haven't been able to fix yet. I modified the other tests so that the program compiles, but they are not expected to pass yet, as I am working on getting the backend tests to work.

I am getting the following output when running the `testmod_backend` tests:

```
UpdateCTestConfiguration  from :/home/toasterpro/CLionProjects/libelektra/build/DartConfiguration.tcl
Parse Config file:/home/toasterpro/CLionProjects/libelektra/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/toasterpro/CLionProjects/libelektra/build/DartConfiguration.tcl
Parse Config file:/home/toasterpro/CLionProjects/libelektra/build/DartConfiguration.tcl
Test project /home/toasterpro/CLionProjects/libelektra/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 37
    Start 37: testmod_backend

37: Test command: /home/toasterpro/CLionProjects/libelektra/build/bin/testmod_backend "/home/toasterpro/CLionProjects/libelektra/src/plugins/backend"
37: Environment variables: 
37:  LD_LIBRARY_PATH=/home/toasterpro/CLionProjects/libelektra/build/lib
37: Test timeout computed to be: 1500
1/1 Test #37: testmod_backend ..................***Exception: SegFault  0.14 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.15 sec

The following tests FAILED:
         37 - testmod_backend (SEGFAULT)
Errors while running CTest
```
I was able to narrow down the scope of the problem by commenting out all tests other than the first part of `test_simple`. The error occurs when `open_backend` is called, which attempts to open the `backend` plugin and call its `elektraBackendOpen` function.

Within the function, I noticed that by commenting out everything from line 601 forward (therefore only processing `getplugins`), the tests in their current state pass. Uncommenting only the part where `setplugins` or `errorplugins` is processed makes the error occur again. 

I tried opening the backend by giving an empty keyset as config, in that case the processing of `getplugins` fails as well.

Also, if I comment out the code from line 609 forward, the test passes again, meaning that in this state, the segfault seems to occur between line 609 and 620

The cause of the segfault is probably due to my limited experience with C, but help or at least debugging tips would be very much appreciated.

mpranj:
I don't have much time to check it out currently, but I'll give a very generic debugging tip.

I would use `valgrind` here. It should give you a trace and tell you _exactly_ where the segfault happens without having to comment out other code. This should save you quite some time.

Run your plugin test from the build directory similar to this: `valgrind ./bin/testmod_backend`.
If you're stuck with the output, post it here again.

vLesk:
> I don't have much time to check it out currently, but I'll give a very generic debugging tip.
> 
> I would use `valgrind` here. It should give you a trace and tell you _exactly_ where the segfault happens without having to comment out other code. This should save you quite some time.
> 
> Run your plugin test from the build directory similar to this: `valgrind ./bin/testmod_backend`.
> If you're stuck with the output, post it here again.

I was able to find the problem and fix it by using `valgrind`, thank you very very much!

vLesk:
I would like to ask for your help yet again, another segfault appears when running the `testmod_backend` test. 

However, when I run it with valgrind, the `elektraBackendOpen` function in `backend.c` is not even called when calling `elektraPluginOpen` in line 85, setting the backend as 0 and thus makes the first test in line 152 fail, not even reaching the point of the segfault. I tried debugging with `gdb` as, with the same results. 

If I run the test with `ctest` however, the function is called as it should. The segfault seems to occur when building up the linked list of plugins in the function `processRole`, specifically in line 294 of `backend.c`:

```
curSlot = curslot->next;
``` 
The segfault is most likely due to not handling the pointers of the linked list in the function properly, but I am not 100% sure since I cannot check the stack trace.

How can I configure `valgrind` properly to be able to view the stack trace? Did I make an error while handling the pointers in the function?

kodebach:
> How can I configure `valgrind` properly to be able to view the stack trace? Did I make an error while handling the pointers in the function?

When using valgrind, the program doesn't behave 100% the same. AFAICT this is one such case. In `processGetPlugins`, `processSetPlugins` and `processErrorPlugins` you are using `elektraMalloc` to allocate the slots. This is not a problem itself, but under certain circumstances you read `slots` before writing to it. This is what this `valgrind` message tells you:

```
==12238== Conditional jump or move depends on uninitialised value(s)
==12238==    at 0x10C01A: check_null_in_slot (testmod_backend.c:127)
==12238==    by 0x10C1DD: test_simple (testmod_backend.c:157)
==12238==    by 0x1115E0: main (testmod_backend.c:356)
==12238== 
```

The easiest solution is to use `elektraCalloc` instead of `elektraMalloc`. Then the allocated memory is guaranteed to only contain zero-bytes.

I also noticed that lines 160 and 162 in `testmod_backend.c` are the same. I assume one of them is wrong...
And in line 188, `keyString` should probably be `keyName`.

src/plugins/backend/testmod_backend.c
```diff
@@ -0,0 +1,363 @@
...
+		      keyNew ("system/elektra/mountpoints/simple/set", KEY_END),
+		      keyNew ("system/elektra/mountpoints/simple/set/presetstorage", KEY_END),
+		      keyNew ("system/elektra/mountpoints/simple/set/presetstorage/#0", KEY_END),
+		      keyNew ("system/elektra/mountpoints/simple/set/presetstorage/#0/name", KEY_VALUE, KDB_DEFAULT_STORAGE, KEY_END),
+		      KS_END);
+}
+
+KeySet * set_default (void)
+{
+	return ksNew (30, keyNew ("default", KEY_END), keyNew ("default/config", KEY_END),
```
kodebach: None of the keys in this key set have valid names, therefore the key set will be empty.
The names either have to start with a `/` or a valid namespace, which `default` is not (yet).

src/plugins/backend/backend.c
```diff
@@ -0,0 +1,949 @@
...
+
+	KeySet * systemConfig = ksRenameKeys (config, "system");
+	ksDel (config);
+
+	return systemConfig;
+}
+
+int linkedListPosition (Key * cur, Key * errorKey)
+{
+	char * name = elektraMalloc (sizeof (keyBaseName (cur)));
```
kodebach: You are using `sizeof` wrong (not just here, but in many places). In C `sizeof` is evaluated by the compiler, so it cannot evaluate to the length of a string. Instead `sizeof (keyBaseName (cur)) == sizeof (char *)`, i.e. the size of a pointer (4 or 8 bytes depending on 32-bit or 64-bit machine).

In general you should use `strlen (myString) + 1` instead of `sizeof (myString)`, but when dealing with keys there is often a better solution, size `Key`s store there sizes. For example there are `keyGetBaseNameSize`, `keyGetNameSize` and `keyGetValueSize` (they all count the terminating zero-byte, so are the same as strlen + 1).

markus2330: Thank you for helping out.

vLesk:
Thank you very much for your help! Your review helped me solve my issue yesterday. The tests pass now.

> When using valgrind, the program doesn't behave 100% the same. AFAICT this is one such case.

One issue that still persists is that `elektraPluginOpen` only opens the `backend` plugin when the tests are run through `ctest`. If I run them through the binaries `./bin/testmod_backend`, the plugin is not opened.

I started writing some benchmarks (`benchmarks/backend.c`) and am facing the same problem. In both cases, `elektraPluginOpen` is called in the `open_backend` function. 

My guess would be that I'm missing some `CMake` configuration to correctly link to the plugin, but I'm not sure since it does work when using `ctest`.

Thank you for answering my flood of questions!

kodebach:
AFAIK the only difference between running with `ctest` and directly executing the binaries, is that `ctest` passes the source folder as the first argument. But since you don't use `srcfile` in your test that shouldn't matter.

But you can still try calling `bin/testmod_backend $PWD/../src/plugins/backend` that should be what crest does.

markus2330:
ctest with some option also prints what it does (maybe `-V` is enough). But it is also not a big deal if you run the benchmarks via ctest for now, simply document it in the method section.

Important is that you submit your thesis today!

markus2330:
@vLesk any progress here? What is still missing?

mpranj:
@vLesk is there any update to this PR? Don't hesitate to ask here if you're stuck with something.

vLesk:
I have a question regarding kdbEnsure, is this functionality planned to be implemented completely or will it stay the way it is? In previous versions of my PR, I commented it out as a short-term solution since it worked with backends directly and I would have to modify it. If it is still to be used I will find a way to make it compatible.

kodebach:
I think you can remove the part of `kdbEnsure` that deals with non-global plugins. I don't think it is used anywhere. AFAIK `kdbEnsure` is only used to mount `gopts`.

markus2330:
I agree, we should remove this functionality for now. If it will be needed later, we can reintroduce it later.

markus2330:
Good to see activity here :sparkling_heart:

vLesk:
I have a question regarding a memory leak I have been unable to identify. When running `testmod_backend`, I get the following reply from the leak sanitizer:

```
Direct leak of 64 byte(s) in 1 object(s) allocated from:
35:     #0 0x7f9b48fe3d28 in __interceptor_calloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xded28)
35:     #1 0x7f9b49f980bd in elektraCalloc /home/weed/CLionProjects/libelektra/src/libs/elektra/internal.c:290
35:     #2 0x7f9b49f9c90c in keyVNew /home/weed/CLionProjects/libelektra/src/libs/elektra/key.c:227
35:     #3 0x7f9b49f9dc3c in keyNew /home/weed/CLionProjects/libelektra/src/libs/elektra/key.c:179
35:     #4 0x55c12766b908 in set_default /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:55
35:     #5 0x55c1276722f6 in test_default /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:264
35:     #6 0x55c12766b1e4 in main /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:397
35:     #7 0x7f9b47bf5b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)

```

The leak seems to stem from the keyset I generate in set_default(), yet as far as I can see, it gets freed when the backend is closed. If I add an extra ksDel in the test, I get this message, which I take to mean that the memory has already been freed:

```
 ==31763==ERROR: AddressSanitizer: heap-use-after-free on address 0x606000003db8 at pc 0x7fa530960ec1 bp 0x7fff694d1800 sp 0x7fff694d17f0
35: WRITE of size 8 at 0x606000003db8 thread T0
35:     #0 0x7fa530960ec0 in ksRewind /home/weed/CLionProjects/libelektra/src/libs/elektra/keyset.c:1340
35:     #1 0x7fa5309687fd in ksClose /home/weed/CLionProjects/libelektra/src/libs/elektra/keyset.c:2554
35:     #2 0x7fa5309690cf in ksDel /home/weed/CLionProjects/libelektra/src/libs/elektra/keyset.c:443
35:     #3 0x5648d7e305b2 in test_default /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:279
35:     #4 0x5648d7e291e4 in main /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:396
35:     #5 0x7fa52e5aeb96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
35:     #6 0x5648d7e293d9 in _start (/home/weed/CLionProjects/libelektra/build/bin/testmod_backend+0x223d9)
35: 
35: 0x606000003db8 is located 24 bytes inside of 64-byte region [0x606000003da0,0x606000003de0)
35: freed by thread T0 here:
35:     #0 0x7fa52f99c7a8 in __interceptor_free (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xde7a8)
35:     #1 0x7fa53096922e in ksDel /home/weed/CLionProjects/libelektra/src/libs/elektra/keyset.c:459
35:     #2 0x7fa530a3b9ad in elektraPluginClose /home/weed/CLionProjects/libelektra/src/libs/elektra/plugin.c:142
35:     #3 0x5648d7e30572 in test_default /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:276
35:     #4 0x5648d7e291e4 in main /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:396
35:     #5 0x7fa52e5aeb96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
35: 
35: previously allocated by thread T0 here:
35:     #0 0x7fa52f99cb40 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb40)
35:     #1 0x7fa530951015 in elektraMalloc /home/weed/CLionProjects/libelektra/src/libs/elektra/internal.c:275
35:     #2 0x7fa5309680b9 in ksVNew /home/weed/CLionProjects/libelektra/src/libs/elektra/keyset.c:249
35:     #3 0x7fa530968663 in ksNew /home/weed/CLionProjects/libelektra/src/libs/elektra/keyset.c:231
35:     #4 0x5648d7e29d13 in set_default /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:55
35:     #5 0x5648d7e302f2 in test_default /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:262
35:     #6 0x5648d7e291e4 in main /home/weed/CLionProjects/libelektra/src/plugins/backend/testmod_backend.c:396
35:     #7 0x7fa52e5aeb96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)

```

Since I am still rather new to these tools, does anyone notice any obvious mistakes I may have made when creating and allocating the keyset?

markus2330:
Thank you for asking!

Do you see the same error with valgrind? Sometimes valgrind gives the better error messages.

The second error says that it was used after it was freed, it does not necessarily say that it was freed twice. So maybe your fix is simply incorrect and you delete it at the wrong place (too early)?

> yet as far as I can see, it gets freed when the backend is closed.

Can you point me to the place where the freeing is implemented?

markus2330:
@vLesk see above, can you check if valgrind also find this leak?

markus2330:
@sanssecours is it possible and sensible to blacklist this ASAN error? Is there a good description of how to add ASAN errors to the blacklist?

mpranj:
This does not look like an ASAN error or false positive to me. I would take the ASAN erros more like hints where to look now.

It seems that the lines mentioned do not match what I see in the PR.
@vLesk can you push your latest version? Maybe even rebase from master if it is not too problematic. Make a backup and try rebasing in a copy so that you don't lose everything if the rebase fails.

vLesk:
Thanks to both of you for helping! I will try it with valgrind in about two hours when I'm at my PC again.

vLesk:
As for the specific ASAN errors, the problem seemed to be that I did nto handle the keysets I passed to the backend plugin correctly. I now patched up the ASAN errors, I will try it with valgrind afterwards.

vLesk:
> @vLesk can you push your latest version? Maybe even rebase from master if it is not too problematic. Make a backup and try rebasing in a copy so that you don't lose everything if the rebase fails.

All my commits should be pushed, I will try rebasing from master.

sanssecours:
> is it possible and sensible to blacklist this ASAN error?

Just [like Mihael](https://github.com/ElektraInitiative/libelektra/pull/2969#issuecomment-611998545), I think those errors are probably caused by our code. Otherwise there would probably be more calls of code that are not part of Elektra’s codebase. Usually AdressSanitizer [should also not produce false positives](https://clang.llvm.org/docs/AddressSanitizer.html#issue-suppression):

> AddressSanitizer is not expected to produce false positives. If you see one, look again; most likely it is a true positive!  

> Is there a good description of how to add ASAN errors to the blacklist?

The syntax of entries is described [here](https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-errors-in-recompiled-code-blacklist).

kodebach:
@vLesk It seems to me that the problem is that, while the config `KeySet` from `set_default` is freed by `elektraPluginClose`, not all the `Key`s in the `KeySet` are freed. Are you copying/moving keys to other keysets somewhere?

One instance where I found one such leak (not the one from your logs):

- `elektraBackendOpen` cuts out a part of the config and passes it to `processConfig`:
https://github.com/ElektraInitiative/libelektra/blob/b013dee7a95d6be9e8d75803debf1a9170e01569/src/plugins/backend/backend.c#L578-L589
- `processConfig` only calls `ksDel` if it is successful:
https://github.com/ElektraInitiative/libelektra/blob/b013dee7a95d6be9e8d75803debf1a9170e01569/src/plugins/backend/backend.c#L45-L56
- `setMountpoint` also doesn't call `ksDel`:
https://github.com/ElektraInitiative/libelektra/blob/b013dee7a95d6be9e8d75803debf1a9170e01569/src/plugins/backend/backend.c#L10-L43
- Therefore at least in this one error case part of the config `KeySet` leaks.

Also `open_backend` `ksDup`s the config `KeySet` before calling `elektraPluginOpen` and `ksDel`s the original one, meaning the `KeySet` from `set_default` can never leak (only the `Key`s within) and adding a `ksDel (defaultKeyset)` to `test_default` will always be a double-free.

---

One idea for debugging this, if `valgrind` doesn't help. Try to fix other memory leaks first (if there are others), maybe this one is automatically fixed in the process. Another idea would be to find out the name of the key that leaks, e.g. by attaching a debugger and noting the addresses of all the `Key`s in the `set_default` `KeySet` and the comparing them with the ASAN output. With the keyname you'll probably know where to look for a leak.

The [Clang Analyzer](https://clang-analyzer.llvm.org/) might also find something, if you manage to get it to work (its a bit fiddly with CMake). Although I doubt give any direct hints for the memory leak.

vLesk:
@kodebach thank you for the recommendations! These would have been especially difficult to find.

I rebased my PR on the newest version of master. After some tinkering, I get no more complaints from ASAN. I will still have to fix the leaks you showed me though.

As for valgrind, am I using it correctly if I run it with the generated binaries? So for example, if I want to run `testmod_backend` I run the command  `valgrind ./bin/testmod_backend`?

mpranj:
> So for example, if I want to run testmod_backend I run the command valgrind ./bin/testmod_backend?

Correct. 

When you get to the point that you want more feedback from the build servers add some release notes, like it's suggested by jenkins:

```
For example, if Sandra Power fixed a typo, she adds to doc/news/_preparation_next_release.md:
- typo fixed _(Sandra Power)_
```

The build servers will also not run all tests if the code is not correctly formatted. You can check the output of the build server on how to fix the formatting or just use what the [restyled bot did](https://github.com/ElektraInitiative/libelektra/pull/3039).

kodebach:
> As for valgrind, am I using it correctly if I run it with the generated binaries? So for example, if I want to run testmod_backend I run the command valgrind ./bin/testmod_backend?

You might want to add the options that `valgrind` suggest at the end when you run it without any options. Alternatively you can also use `ctest -R testmod_backend -T memcheck --output-on-failure`.

mpranj:
@vLesk if you get not complaints from ASAN maybe you already fixed the leaks. I would be surprised if valgrind found a leak which was not shown by ASAN.

src/libs/elektra/split.c
```diff
@@ -168,6 +184,12 @@ ssize_t splitAppend (Split * split, Backend * backend, Key * parentKey, int sync
 	split->parents[n] = parentKey;
 	split->syncbits[n] = syncbits;
 
+	// FIXME: somehow find correct sizes
+	split->systemsizes[n] = 0;
+	split->usersizes[n] = 0;
+	split->specsizes[n] = 0;
+	split->dirsizes[n] = 0;
```
kodebach, line 191: Theses values were part of the `Backend` struct passed into this function before. I don't know what these sizes are used for, so I am not sure how to solve this now.

kodebach:
I have now rebased this PR and fixed some of the most obvious problems. I think it should now at least compile on all build servers.

However, there are still some memory problems and also quite a few other bugs (e.g. [this one](https://github.com/ElektraInitiative/libelektra/pull/2969/files#r483989030)). It might also be the case that some of the tests (e.g. `test_mount`) don't really make sense in their current form anymore.

src/libs/elektra/split.c
markus2330: In the previous implementation there were similar sizes (the names were singular, though) in Split which were used to track how many keys in the respective namespace exist.

markus2330:
> I have now rebased this PR and fixed some of the most obvious problems. I think it should now at least compile on all build servers.

Thank you very much! :rocket: 

It seems like release notes are missing?

src/libs/elektra/split.c
kodebach: Like I said, previously these values where stored *within* the `Backend` struct. Therefore, they where automatically passed around.

In `splitBuildup` the `mountGetBackend` function used to return a `Backend *` containing these values. This pointer was then passed to `splitAppend` and simply stored in the `Split` struct.

However, now `mountGetBackend` returns a `Plugin *`, meaning the sizes are not available inside `splitAppend`.

In general, I am not sure we even need these fields at all. If they are supposed to store the number of keys in the respective namespace, then all except the one matching the namespace of `split->parent[n]` should be zero. The one with the matching namespace should just be `ksGetSize (split->keysets[n])`, if I understand the concept correctly.

markus2330: Without checking the code: The different sizes for different namespaces are needed in the Backend struct as cascading backends might have keys across namespaces. And this is probably also the situation where they differ between the Backend and Split struct.

mpranj: I tried to get around this in the cache implementation (using `ksGetSize()`) and failed. I didn't want to dive too deep into the split code, so now the cache simply persists these values.

@markus2330 could the values of `split->systemsizes[n]` etc. be reconstructed somehow? It would be nice to get rid of all those `split->zzzsizes[n]` members.

mpranj:
@kodebach thank you so much for rebasing and fixing the problems! 🚀 

@vLesk are there any bugs/TODOs you know of which are not mentioned in code or issues? If there are could you document them somewhere so it would be easier to help out with this PR.

src/libs/elektra/split.c
markus2330: With the current logic (making `kdbSet` a NOP for the backends where the keys are not changed) the sizes are needed somewhere. We use the sizes to check if the same amount of unsynced keys would be written (NOP situation). To have the sizes both per Backend and Split, depending on the implementation, might be unnecessary. It might be, however, a major rewrite with little benefits. (iirc there are many tests that check these sizes.)

Having new higher-level tests that directly check the outcome of the whole `kdbSet` (which plugins are called with which keys) would be very beneficial. @PhilippGackstatter @mpranj: What do you think?

mpranj: Sounds like it makes no sense to rewrite the `split->zzzsizes[n]` code then.

Higher level tests make sense to me. Being able to track which plugins are called with which keys would be great. I needed this quite extensively myself when debugging problems.

PhilippGackstatter: Having these kinds of tests sounds good from my end. It would probably be benficial for a reimplementation. It also sounds like a lot of work though, so it needs to make sense from other points of views as well.

kodebach: I might quite sure I understand everything, but to me it sounds like these `size` variables do in fact have a use. However, it also sounds like the could be maintained within the `backend` plugin itself. The NOP case could then be detected within the `elektraBackendSet`, which would return ASAP.

markus2330:
How is it going on here? Did you find some concrete problem where someone might be able to help you?

vLesk:
@markus2330 it is going quite well. An error keeps occurring in most of the tests when `kdbGet` is called on the cache plugin, concretely when it looks for the mountpoint in the parent key (line 288 in `cache.c` to be precise). The parent key which is given does not contain the mountpoint of the backend. Hence, calling `strcmp` on it throws an error. However, I was able to single down the error to the mountpoint not being set correctly in `backendGetMountpoint` in `backend.c`. As far as I've looked, it seems like a simple fix, but I'll write if greater changes are needed. 

Another error pops up in tests like `test_splitget`, there, it seems like split operations don't work correctly yet, it's probably due to the problems @kodebach talked about.

markus2330:
> cache plugin

Maybe @mpranj has some idea?

mpranj:
Line 288 in `cache.c` is where the contract is set up. This is more or less identical code to 90% of Elektra plugins, so I suspect that would cause problems with any plugin.

The cache plugin, however, relies on the mountpoint being set correctly in the parentKey. This is specific to the plugin, so the cache would not work if the mountpoint is not set.

I'll take a closer look what's going on here. If the cache plugin is an obstacle for further development I encourage you to disable it for this PR.

markus2330:
Thank you @mpranj for looking :sparkling_heart: 

> If the cache plugin is an obstacle for further development I encourage you to disable it for this PR.

This is an amazing offer. :rocket:

vLesk:
Thank you for the comment, @mpranj! I think disabling the cache plugin will not be necessary for this specific issue, the errors pointed me to a general implementation problem of the backend plugin.

markus2330:
Good to see that you fixed a bug! :sparkling_heart: 

What is the next step?

src/libs/elektra/split.c
kodebach: I may have found a way of avoiding this problem by rethinking splits entirely. See #3492.

vLesk:
I see the tests don't produce the correct values, I'm not sure what the cause is, but I think it might be due to the size arrays mentioned by @kodebach, I'll try fixing that first since I need to do it anyway.

kodebach:
> I'll try fixing that first since I need to do it anyway.

Before you start working on the sizes, we should discuss #3492. If we are going to get rid of the `struct _Split`, it would be unnecessary work fix the sizes with the current setup.

markus2330:
@vLesk are there also other tasks to be done or are the sizes the last problem? Do all other test cases already succeed?

kodebach:
```
C++ exception with description "8 Warnings were issued:
	Sorry, module resolver_fm_hpu_b issued the warning C01100:
	Resource: Could not create configuration file '(null)'. Reason: Bad address
	Mountpoint: user
	Configfile: 
```

Looks like the `resolver` plugin receives a NULL pointer somewhere. I think 

Also in other tests I see:
```
 Sorry, 1 warning was issued ;(
	Sorry, module backend issued the warning C01330:
	Plugin Misbehavior: The resolver is not defined properly, the backend was not initialized correctly!
```
Which seems to be this error message from the `backend` plugin:
https://github.com/ElektraInitiative/libelektra/blob/25734d31f30dd59f34904a61d410c48cc87989a8/src/plugins/backend/backend.c#L814-L815

You changed how the `system/elektra/mountpoints` configuration worked, so maybe you missed some of the hardcoded bootstrap configuration? For some tests we might also have to update the tools library. Normally a debugger like `gdb` would help here, but the tests with those errors are shell tests, so a debugger doesn't work.



You could also try to add some logging statements, like I did in #3447 with this line:
https://github.com/ElektraInitiative/libelektra/blob/615975f4485a466004804cc60f716a769ba2d050/src/libs/elektra/key.c#L316

For that to work, you will have to pass `-DENABLE_LOGGER=ON` (`-DENABLE_DEBUG=ON` can also help) to `cmake`.


Finally there is this error message:
```
fallback for bootstrapping: you might want to run `kdb upgrade-bootstrap` (in kdbOpen at src/libs/elektra/kdb.c:366)
```

That means `kdbOpen` doesn't work correctly (which explains why nothing really works). Sadly there isn't really a simple C test for `kdbOpen` AFAIK. I suggest you try to step through the execution of e.g. `kdb ls system` with a debugger. If you get to line 366 in src/libs/elektra/kdb.c and you can stop, the bug has to be before that line. **Note:** make sure your KDB is empty otherwise you will have problems with old configs (e.g. by calling `cmake` with `-DKDB_DB_SPEC="$PWD/config/spec" -DKDB_DB_SYSTEM="$PWD/config/system" -DKDB_DB_USER="cdbg-1/.config"` and ensuring these three directories are empty)

For the C++ tests to work, you will probably have to update the tools library at src/libs/tools.

markus2330:
> fallback for bootstrapping

Please remove this fallback code. It is not useful anymore.

> update the tools library

This is probably quite a big task to be done?

vLesk:
@kodebach thank you very much for the recommendations! I'll look into it. 

@markus2330 Regarding the concrete implementation, what remains are the sizes, writing the documentation, there is also a part in the plugin where I still need to implement opening the missing plugin in case of errors, but that shouldn't take long. Otherwise there may still be minor bugs which are not yet visible.

vLesk:
@kodebach, I was able to get the correct test values by clearing my kdb, thank you for your recommendation! I'm working on solving the resolver issue and getting kdb to run now.

vLesk:
> For the C++ tests to work, you will probably have to update the tools library at src/libs/tools.

I already updated the tools library to work with the new mountpoint configuration, but it's possible that I may have missed something.

vLesk:
I have a question regarding the logger, where are the logs normally stored?

kodebach:
> I have a question regarding the logger, where are the logs normally stored?

Depends on your configuration, by default (if logging is enabled) log output will go to `stderr` and to the syslog. By default `stderr` only contains messages with level `ELEKTRA_LOG_LEVEL_WARNING` or greater.

For more info see: https://github.com/ElektraInitiative/libelektra/blob/master/doc/dev/logging.md

(Note: if you do change any of the parameters in `kdblogger.h`, please remember to change them back)

You can view the relevant syslog with e.g. `journalctl -t Elektra` or a GUI syslog viewer. All of Elektras messages are tagged with the identifier `Elektra`.

vLesk:
Thank you!

vLesk:
I'm currently working on getting `kdb` to work, when calling `kdb ls system` I get the following response:
```
Add Warning C01200: Dlopen failed. Could not load module libelektra-backend.so. Reason: libelektra-backend.so: cannot open shared object file: No such file or directory (in elektraModulesLoad at src/libs/loader/dl.c:90)
Failed to load plugin backend@ (in elektraPluginOpen at src/libs/elektra/plugin.c:128)
Add Error C01200: Could not open default backend. See other warning or error messages for concrete details (in kdbOpen at src/libs/elektra/kdb.c:330)
 Sorry, 1 warning was issued ;(
	Sorry, module kdb issued the warning C01200:
	Installation: Dlopen failed. Could not load module libelektra-backend.so. Reason: (null)
Sorry, module kdb issued the error C01200:
Installation: Could not open default backend. See other warning or error messages for concrete details

``` 
The error occurs in line 90 of `dl.c`, after calling `dlopen`. This occurs when calling `elektraPluginOpen` on the `backend` plugin in line 208 of `elektra/backend.c`. As `kdbOpen` has not been called on the `backend` plugin yet, I'm not sure where this error comes from. `libelektra-backend.so` is present in my build directory, even when setting it up again completely. Is there something else I may have missed in the configuration? I made sure to include all plugins.

markus2330:
You might not use Elektra from the build dir but your installed Elektra. To run Elektra from the build dir, use `source ~e/scripts/dev/run_env` from the build dir. `~e` must be the source dir. (It adds builddir/lib to LD_LIBRARY_PATH, which might be the cause for the problem you have.)

vLesk:
Thank you very much! I'll try it out.

vLesk:
Opening backends is working now AFAIK. There is one case when calling `kdbGet` in `elektraGetCheckUpdateNeeded` where it fails and the resolver reports an error, I checked the config of the erroneous backend when that happens, it is just one key with name `system/mountpoint` and string `system/elektra/version`. This is curious as opening and calling the `version` backend already seems to occur correctly  The `kdbGet` calls work for all backends except for this one. Does anybody know where this backend comes from?

src/libs/elektra/kdb.c
```diff
@@ -505,24 +506,26 @@ static int elektraGetCheckUpdateNeeded (Split * split, Key * parentKey)
...
 
-		Plugin * resolver = backend->getplugins[RESOLVER_PLUGIN];
-		if (resolver && resolver->kdbGet)
+		if (backend && backend->kdbGet)
 		{
 			ksRewind (split->keysets[i]);
 			keySetName (parentKey, keyName (split->parents[i]));
 			keySetString (parentKey, "");
-			ret = resolver->kdbGet (resolver, split->keysets[i], parentKey);
+			ret = backend->kdbGet (backend, split->keysets[i], parentKey);
```
kodebach, line 517: This should not do a full `kdbGet`. You can either let this code access the `resolver` directly (like it was before) or you (probably better) set a special metakey on `parentKey` (e.g. `backend/resolveronly`) to tell the `backend` plugin that is should only call the `resolver` and nothing else.

AFAIK the purpose of this call here is to:
1. let the resolver find the filename for a mountpoint and save it in the split
2. Detect if the file exists at all. If the file doesn't exist `resolver` returns `ELEKTRA_PLUGIN_STATUS_NO_UPDATE` and we can ignore this mountpoint in `kdbGet`.
3. Detect if the cache can be used (not 100% sure how this works)

@mpranj probably knows more, since he had to adapt this to for the cache

kodebach:
>  There is one case when calling kdbGet in elektraGetCheckUpdateNeeded where it fails and the resolver reports an error

See the code comment I made. Not sure, about the relation to `version`, but maybe this is just the first mountpoint that gets checked.

src/libs/elektra/kdb.c
mpranj: Good catch, @kodebach already said what is important. This should not be a full kdbGet.

The resolver is called first to determine whether we need to call the storage plugin at all. (obviously it also resolves the filenames etc.) The resolver already has similar optimizations without the cache, it's just that the cache is able to persist this information. The default resolver simply remembers the modification timestamp of the file, but this could be different for other resolvers.

vLesk:
Thank you for your help, @kodebach and @mpranj! The handling of the resolver did end up being the issue, but for a different reason. I already took into account that only the resolver is called at first, but there was an issue when calling `kdbGet` on the `version` plugin. The backend that the latter is using contains only the plugin itself. This didn't use to be an issue, but the `backend` plugin was implemented to return an error code if there is no resolver present. 

I now reimplemented that aspect so that the lack of a resolver is not considered to be an error by itself. Could this be an issue in certain cases? 

So far, things seem to be working fine with that implementation. Later on, I get the following error from the `cache` plugin:

```
Sorry, 12 warnings were issued ;(
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/constants not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/constants/KDB_VERSION not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/constants/KDB_VERSION_MAJOR not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/constants/KDB_VERSION_MINOR not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/constants/KDB_VERSION_PATCH not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/constants/SO_VERSION not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/infos not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/infos/author not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/infos/description not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/infos/licence not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
        Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system/elektra/version/infos/version not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint
```

@mpranj since the cache is mentioned, do you have any idea what this could be? It's most likely another problem with the `version` backend...

vLesk:
I played around with `kdb ls` some more, calling `kdb ls system` seems to work and always return the keys now, but sometimes with and sometimes without these errors...

vLesk:
Simple `kdb get` and `kdb set` calls seem to be working as well

kodebach:
> I now reimplemented that aspect so that the lack of a resolver is not considered to be an error by itself. Could this be an issue in certain cases?

Most plugins probably expect that the `parentKey` contains a resolved filename.
If it works, it would be better to add the `noresolver` plugin as an resolver in the `version` backend.

> So far, things seem to be working fine with that implementation. Later on, I get the following error from the cache plugin:

That is actually a somewhat known bug, see #3299. The bug itself is almost certainly unrelated to your changes. It can appear, if certain tests fail.

As a workaround you can delete the `elektra.ecf` file with the broken data (as described in #3299). Then run `make run_all` again to see which test first produces these errors. Either this test or the last one that touched the KDB before that fails. Then you can disable the problematic test to get proper results from the other ones (how to disable depends on the test).

Alternatively, you can run tests separately with `ctest --output-on-failure -R <REGEX>` where `<REGEX>` is a regular expression that will be matched against the test names. For `ctest --output-on-failure -R testmod_` will run all containing `testmod_` i.e. all plugin tests.

You can also use `ctest` calls to narrow down which test causes the problem.

If you can't find the problematic test, try to fix the other ones first maybe the bug goes away in the process.

mpranj:
> `Postcondition of backend was violated: drop key system/elektra/version not belonging to "system/elektra/version" but instead to "system/elektra/modules/cache" because it is hidden by other mountpoint`

Yes, please ignore this at least until #3299 is resolved. If it goes away by disabling the cache, simply disable the cache for now. It is actually very likely that the cache does not work as expected with such huge changes to the backends, even if we disregard #3408 

If you want I can push changes to this branch which should definitely disable the cache, and we will take a look together once everything else is resolved. I think this makes the most sense. It will be very hard to debug things if there are errors in the core (or backend implementation) and in the cache as well. I think it's better to focus on the core and backend plugin right now.

vLesk:
@mpranj that would make things a lot easier, thank you! That way I can clear up the errors on my side.

mpranj:
I disabled it now as it would be disabled on the master branch. I don't know if there are any other changes which I am not aware of.

Whenever you are unsure, you can also use -DPLUGINS="ALL;-cache" when running cmake, which is an alternative approach to what I pushed.

vLesk:
Thank you very much!

vLesk:
I was held up in traffic on my way back from Slovenia, so I won't be able to make it to the meeting today, unfortunately. 

The plugin is working now, I can do basic `kdb` commands, the default backend is also mounted correctly. The only issue that persists is the issue with the `version` backend we talked about. It works normally if I do not throw an error if a backend does not have a resolver, but if I use the `noresolver` plugin as @kodebach suggested, the Slot containing it is null when carrying out `kdbGet`, so it gets lost somewhere along the way. Resolvers in other backends work properly otherwise.

I am working on getting all the tests running, and I noticed that the `testshell` tests do not work, and after running them, the other tests as well as `kdb` no longer work until I clear `default.ecf`. Where can I find the `testshell` tests so that I can fix them?

I work full time from November onwards, so it will be difficult for me to find time for Elektra after that. That is why I will take next week and the weekend off from work to focus on the project and get as much work done as possible. I intend to finish my work during that week, but it could also happen that unforeseen issues pop up and I cannot finish all the work in time. In November, I will be unable to invest greater amounts of time into the project, but I will be available to help out in case of questions and provide minor fixes.  

However, I do intend to finish my work next week, and I hope that this won't be an issue in the first place.

markus2330:
Thank you, yes, I also think there is enough time in October. Would be amazing to get this merged soon! :sparkling_heart:

markus2330:
@vLesk how is it going on? Do you need some specific help?

vLesk:
> @vLesk how is it going on? Do you need some specific help?

@markus2330 thanks for asking! I have a question, `kdb` seems to be working, but the `testshell` tests produce configuration which breaks it, leading to the following error which persists until I clear `default.ecf`:

```
Sorry, module dump issued the error C01100:
Resource: Could not open configuration file  for reading. Reason: No such file or directory

fatal: Could not export system config
```
My question in order to find and fix the problem is, where can I find the `testshell` tests and how do they work?

kodebach:
> My question in order to find and fix the problem is, where can I find the `testshell` tests and how do they work?

`testshell` is used as a prefix for a few different tests. The most common case is `testshell_markdown_[plugin]`. These are generated from code blocks in the README.md files of plugins. For example `testshell_markdown_type` is generated from `src/plugins/type/README.md`. The tests run all code blocks that are marked as `sh`. For `type` this means theses blocks:

https://github.com/ElektraInitiative/libelektra/blob/4146061ddefbc7c3d48327f61a5e7a716e4b115f/src/plugins/type/README.md#L179-L291

The tests are quite hard to debug. But if all of them fail, there is probably a quite severe bug. You should be able to narrow things down by manually executing the shell commands one after another.

The exact format used in the comments of these shell scripts is documented in [`tests/shell/shell_recorder/tutorial_wrapper/README.md`](https://github.com/ElektraInitiative/libelektra/blob/master/tests/shell/shell_recorder/tutorial_wrapper/README.md)

vLesk:
Thank you @kodebach! It seems that the issue is not with the tests themselves, but with mounting in general, the config does not seem to be stored correctly, I'm looking into it.

markus2330:
@vLesk any update? Can you describe us what does not work? :sparkling_heart:

markus2330:
I hope we can get this into 0.9.5. @vLesk do you have any status update? Can you find some time? Now would be a good time for a rebase, we upgraded to clang-format-11.

kodebach:
I'm closing this PR, since it takes for every to load on GitHub. I will open a new PR with a rebased version of this, once #3651 is merged.

markus2330:
Thank you so much! :sparkling_heart: 

I just read yesterday in #3461 that it would be great to have this backend plugin finally fixed.

PR #2975, opened 2019-09-17
Plugins: Unify Name of Configuration Check Function

sanssecours:
May I ask why we now [use the name `checkConf`](https://github.com/ElektraInitiative/libelektra/commit/3478694d) instead of `checkConfig`? I liked the old name better, since I am usually not a fan of abbreviated names. It also looks like we still use the similar name `kdbGenConfig` in the [ReadMe of all plugins](https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/README.md#c-interface). Do we use this function anywhere? At least searching the code base for `GenConfig` did only show two occurrences of the string, one in the ReadMe of all plugins, and one in the man page generated from said ReadMe file.

### DIFF
--- a/doc/man/man7/elektra-plugins.7
+++ b/doc/man/man7/elektra-plugins.7
@@ -36,7 +36,7 @@ All plugins implement the same interface:
 \fBkdbClose()\fR makes sure that plugins can finally free their own resources in \fBelektraPluginClose()\fR\.
 .
 .IP "\(bu" 4
-\fBkdbCheckConfig()\fR can be called manually to ensure a plugin is configured properly\.
+\fBkdbCheckConf()\fR can be called manually to ensure a plugin is configured properly\.
 .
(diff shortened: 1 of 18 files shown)


markus2330:
Thank you for improving the consistency!

> May I ask why we now use the name checkConf instead of checkConfig?

The relevant name (in the exported symbols) always was checkconf since the introduction from @petermax2. How the C functions are called is only an interna and, from the interface perspective, irrelevant.

>  I liked the old name better, since I am usually not a fan of abbreviated names

I agree. Especially if they are not always abbreviated it gets confusing. I consider other problems much more important, though.

> It also looks like we still use the similar name kdbGenConfig in the ReadMe of all plugins. Do we use this function anywhere?

Actually these two sentences are wrong, I clarified them in ca7cc623bca3e2d72a86dda7051a42a3ac12a025

PR #2985, opened 2019-09-19
Fix Spelling: Move Script to Correct Location

sanssecours:


markus2330:
Thank you very much for this!

I wonder if reformat-all should also execute fix-spelling. It seems like this script is often forgotten and the fix-spelling is also kind-of reformatting.

Or we introduce a new script that:

- calls reformat-all
- calls fix-spelling
- calls the link checker

What do you think?

sanssecours:
> What do you think?

I am fine with the situation as it is now 😊. If you want to add another script, I do not think that is a problem. I do not know how useful it will be though, since the build servers already 

- run all checks,
- informs us what went wrong, and 
- tell us how to fix the situation.

markus2330:
Yes, you are right. Even more scripts will not help in the zoo of scripts.

PR #3010, opened 2019-09-26
Kdb showmeta

Piankero:
## Basics

Check relevant points but **please do not remove entries**.
Do not describe the purpose of this PR in the PR description but:

- [ ] Short descriptions should be in the release notes (added as entry in
      `doc/news/_preparation_next_release.md` which contains `_(my name)_`)
      **Please always add something to the the release notes.**
- [ ] Longer descriptions should be in documentation or in design decisions.
- [ ] Describe details of how you changed the code in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, should be in the commit messages.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests for my code
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

Reviewers will usually check the following:

- [ ] Documentation is introductory, concise and good to read
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

- Add the "work in progress" label if you do not want the PR to be reviewed yet.
- Add the "ready to merge" label **if the build server is happy** and also you
  say that everything is ready to be merged.

PR #3013, opened 2019-09-27
CMake: Print Warning About Missing Man Pages

sanssecours:


scripts/cmake/Modules/LibAddMacros.cmake
```diff
@@ -572,6 +573,15 @@ function (generate_manpage NAME)
 			add_dependencies (man man-${NAME})
 		endif (RONN_LOC)
 
+		if (NOT EXISTS "${OUTFILE}")
+			message (WARNING "\nThe file “${MAN_PAGE_LOCATION}” does currently not exist. \
```
markus2330: Please use error, we want to always have all man pages.

sanssecours: As far as I can tell, we can not use `ERROR` here, since then generating the man page – via the current build system (e.g. `make`, `ninja`) – will not be possible any more. If we do not add the man page, then the installation will fail with an error anyway, just like it does now.

markus2330: Ahh, ok. As long as the build server fails, I am good.

markus2330:
Why does https://github.com/ElektraInitiative/libelektra/pull/3010 succeed now? Shouldn't it fail during installation?

sanssecours:
> Why does #3010 succeed now? Shouldn't it fail during installation?

The reason is commit https://github.com/ElektraInitiative/libelektra/pull/3010/commits/87bcf4836114bc3b220686307e64e122d5038adc, which contains an updated `if`-statement:

```cmake
if (INSTALL_DOCUMENTATION AND EXISTS "${OUTFILE}")
  install (FILES ${OUTFILE} DESTINATION share/man/man${SECTION})
endif ()
```

.

markus2330:
I see now, I wrongly assumed that this PR contains the same change.

sanssecours:
> I see now, I wrongly assumed that this PR contains the same change.

Technically you are right 😊, this PR also contained the overwritten changes. I removed the relevant commit.

markus2330:
Thank you!

PR #3100, opened 2019-10-19
Cirrus: Add ASan Enabled FreeBSD Build Job

sanssecours:


src/plugins/python2/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ if (DEPENDENCY_PHASE)
 		remove_plugin (python2 "swig_python2 binding is required")
 	elseif (NOT PLUGINPROCESS_FOUND)
 		remove_plugin (python "Elektra's pluginprocess library is required")
+		remove_plugin (python2 "Elektra's pluginprocess library is required")
```
markus2330: Is this fix correct? Shouldn't *only* python2 be removed (why remove the python plugin here?) It might not really harm if removal of plugins is idempotent.

markus2330:
Thank you so much!

src/plugins/python2/CMakeLists.txt
sanssecours: > Is this fix correct?

I am not sure. If I do not add this, then building the Python 2 plugin will fail on FreeBSD, as far as I can remember.

Issue #3102, opened 2019-10-19
Keyname overhaul


kodebach:
There are a few issues already concerning key names:

- #2983: add `meta` namespace
- #3084: non-alphabetical ordering of namespaces
- #2698: automatically prefix array indices with underscores
- #3082: add `default` namespace

In addition `elektraKeySetName` is known to be a bottleneck and should therefore be optimized.

The goal of this issue is to define the structure of key names, so that we can reimplement `elektraKeySetName`.

kodebach:
## Escaped Key Name Structure

The structure of escaped key names is defined by the following ANTLRv4 grammar.

```antlr
grammar EscapedKeyName;

keyName
    :   namespacedName
    |   cascadingName
    ;

namespacedName
    :   Namespace ':' keyPath
    ;

cascadingName
    :   keyPath
    ;

keyPath
    :   ('/' keyPart)+
    |   '/'
    ;

keyPart
    :   inactivePart
    |   activePart
    ;

inactivePart
    : '.' canonicalPart
    | '.' UnprefixedArrayIndex
    ;
 
activePart
    :   canonicalPart
    |   nonCanonicalPart
    ;

canonicalPart
    :   namePart
    |   ArrayPart
    |   emptyPart
    ;

nonCanonicalPart
    :   '.'
    |   '..'
    |   
    |   UnprefixedArrayIndex
    ;

Namespace
    :   'spec'
    |   'dir'
    |   'user'
    |   'system'
    |   'meta'
    |   'default'
    |   'proc'
    ;

namePart
    : partStartChar partChar*
    | DotSequence
    ;

partStartChar
    :   EscapedStartChar
    |   EscapedSpecialChar
    |   NormalChar
    |   ':'
    ;
partChar
    :   EscapedSpecialChar
    |   StartChar
    |   NormalChar
    |   ':'
    |   '%'
    ;

emptyPart : '%' ;

ArrayPart : '#' PrefixedNumber ;
UnprefixedArrayIndex : '#' Number ;
DotSequence : '..' '.'+ ;

fragment Number : NonZeroDigit Digit* ;
fragment NonZeroDigit : [1-9] ;
fragment Digit : [0-9] ;

fragment PrefixedNumber
    :   Digit
    |   PrefixedNonZeroNumber
    ;

fragment PrefixedNonZeroNumber
    :   NonZeroDigit
    |   '_' PrefixedNonZeroNumber Digit
    ;

StartChar : [.%#@] ;
[...]

Issue #3029, opened 2019-10-02
Each Error should be mapped to a different error code


PhilippGackstatter:
As mentioned in #2979 many functions return -1 for multiple errors. Take `keySetName` as an example. The documentation states

**Return values**
```
size | in bytes of this new key name including ending NULL
0    | if newName is an empty string or a NULL pointer (name will be empty afterwards)
-1   | if newName is invalid (name will be empty afterwards)
-1   | if key was inserted to a keyset before
```

Two different errors are mapped to -1, making them indistinguishable for the caller. This is especially problematic for bindings that map these error codes to Exceptions or some Error structure. The bindings might then throw an `InvalidName` Exception because -1 was returned, but the real problem was that the key was in a keyset. This only leads to confusion for developers.

Error codes should probably use -1, -2, and so on (or any other mechanism that allows for explicitness) to make exact matching for the caller possible.

markus2330:
Thank you for the proposal!

markus2330:
I agree that mapping to different negative numbers should have no negative side-effect (comparing using `< 0` on "any error" is trivial) but helps in some cases (at least in the case you had in the binding and a similar case I already had in the python binding) where the different error matters.

How many places are effected (next to keySetName)?

PhilippGackstatter:
> How many places are effected (next to keySetName)?

Also affected
- keyGetMeta
- keyAddBaseName
- keyAddName
- keySetBaseName
- keyGetBinary
- keyGetString
- keySetBinary

Affected, if null pointer and memory errors should be mapped to different error codes
- keyCopyAllMeta
- keyCopyMeta

Maybe affected, needs discussion
- keyGetFullName (returns -1 on Null pointers or if maxSize is 0 or larger than SSIZE_MAX)
- keyGetBaseName (same as above)
- keyGetName (same as above)
- keyString, maybe? It could return error codes instead of "(null)" and "(binary").

markus2330:
> Also affected

We should fix all of them (with consistent error values).

> Affected, if null pointer and memory errors should be mapped to different error codes

I think there is little benefit from this.

> returns -1 on Null pointers or if maxSize is 0 or larger than SSIZE_MAX

Is both API misuse, so same return value is ok.

> keyString, maybe? It could return error codes instead of "(null)" and "(binary").

This is on purpose so that a `printf("%s", keyString(k))` is safe.

markus2330:
@kodebach did you have some new insight here while you did #3102?

[...]

Issue #2983, opened 2019-09-19
allow access to KeySet of meta-data


markus2330:
When meta-data was introduced it was not so clear which datastructure will be the ideal one for meta-data. As the KeySet now is also a hash-map, the KeySet seems to be the best candidate.

Because of that, it is not needed to hide anymore which datastructure is used for metadata. Thus I propose to directly expose the Meta-KeySet to API users. We would only need to implement one run-time restriction for meta-data keysets: non-meta data keys cannot be added (this implies that all keys inside would be read-only).

Furthermore, the old `keyMeta*` should be moved to the libmeta library.

Possible advantages:

- same code to manipulate KeySets and Meta-Data. This is particularly useful for 3-way merges (@Chemin1) or storage plugins (@sanssecours)
- bindings can reuse iterator logic for KeySet and meta data (@PhilippGackstatter @raphi011)
- availability for all the methods for keysets (like cutting, serializing)

What do you think?

markus2330:
@kodebach wrote in #2979:
> If we return an actual KeySet that would open up the possibility of metakeys having metakeys of their own. This also opens up the possibility of cycles in the references. I don't think this is a good idea...

The keyset would, of course, be read-only. It would only be allowed to add/del new meta-keys. The keys itself are already read-only. So the change within `ks*` would not be too big. If the meta-functions are moved to libmeta, then the client-code should also run unmodified.

kodebach:
> The keyset would, of course, be read-only. It would only be allowed to add/del new meta-keys.

That's a contradiction... "read-only" != "add / del new meta-keys"

Also, how would this be enforced? I assume through some internal flags. IMO that would be a bad API, since you cannot tell easily tell whether a KeySet can be used with a certain function or not.

> If the meta-functions are moved to libmeta, then the client-code should also run unmodified.

Clients would, however, be required to link against `libelektra-meta`.

IMO we should just add a `const char * keyMeta (Key * key, const char * metaName);` and `KeySet * keyMetaKeySet (Key * key);` to `libelektra-core`. These are pretty essential functions. Their code should also be extremely small, so binary size impact should also be irrelevant.

Ideally `keyMetaKeySet` would return a `const KeySet *`, but that is problematic because of the internal iterators.

PhilippGackstatter:
So does that mean the old API vanishes completely or not? 
[...]

PR #3115, opened 2019-10-25
Keyname overhaul

### LINKED ISSUE #2698: arrays with natural sorting
As discussed in #182 and #2388 we could avoid the `_` for array numbers if we internally escape `#<nr>` to `#_..<nr>` (elektraUnescapeKeyName):

> As already discussed shortly, we could also extend the escaping logic for key names and avoid the _ prefixes. So it would be #10, and the (internally) unescaped version (see keyUnescapedName) would be #_10 for correct sorting.

Then the sorting would be still correct and the user would not be bothered by `___` (as mentioned recently in #2685)

Furthermore, we maybe should change the character for arrays to not be `#` as `#` is a shell comment (see #2388).

### LINKED ISSUE #3050: remove elektraKeySetName
The idea of elektraKeySetName was to have flags which change the restrictions on the names (to allow cascading, meta, and empty key names). Now:

1. cascading names are allowed by default
2. against empty key names we are already protected by `ksAppendKey`
3. against meta/non-meta key names we will get protection with #2983

So we actually do not need the argument for elektraKeySetName. So I propose to remove elektraKeySetName and always allow every name. The introspection which key name was used can be done with keyGetNamespace (see also #3049).

This change has effect on tooling: the error message will not be thrown at `keySetName` but later at `ksAppendKey`.

### LINKED ISSUE #3082: default namespace
As already discussed in #2723, #2983 and others (older issues I could not find about the spec plugin) it would be clearer if we would not abuse cascading keys to also represent default keys.

## Current Situation

Currently, the spec plugin creates cascading keys for all `default` it does find. `ksLookup` uses these keys (instead of creating new cascading keys, see #972) if they are found.

## Wanted Situation

Not cascading keys but `default:` keys should be added in the spec plugin (and within `ksLookup` if we cannot fix #972)

## Advantages

Following advantages are expected:

- ksAppendKey could reject cascading keys (only default keys would be allowed)
- The `default` key would make clearly visible what kind of key it is (not clear if `cascading` keys are reused for different purposes.

### LINKED ISSUE #3084: order of namespaces in KeySet
Currently keys are added in alphabetical order, thus the namespaces would be in following order:

1. default
1. dir
1. proc
1. spec
1. system
1. user

More logical, however, would be to have the keys in order of preference:

1. spec
1. proc
1. dir
1. user
1. system
1. default

This could be implemented similar to #2698: Within the unescaped key names there could be some prefix that brings the keys to the desired order.

kodebach:
Implements #3102 

Closes #3050
Closes #3084
Closes #2698
Closes #3082

## Basics

      contains `_(my name)_`)

## Checklist

      (not in the PR description)

## Review


kodebach:
@markus2330

Before we get to the questions, one important thing:

I added a separate `ukey` field for the unescaped name to `struct _Key`. This simplifies a lot of things. I hope this is not a problem. (@mpranj I also hope I modified `mmapstorage` correctly, please check.)

I also removed `elektraFinalizeName`, so any modification to the key name has to be made via the proper functions. This resulted in some changes to `keyset.c`, most cases are covered by the newly added `keySetNamespace` (not working correctly yet) function that changes the namespace of a key without touching the rest of the name. `ksCut` was an exception. There I introduced a small hack, by only changing the namespace in the unescaped name, since the escaped name is not relevant inside of `ksCut` (or its callees).

---

Now the many questions:

This is still very much broken (as you can see), but there are a lot of open questions:

First the more concrete questions:

1. There are some weird escaping tests in `data_escape.c`. I have no idea, why the old escaping function worked this way. Can you maybe shed some light on that?

https://github.com/ElektraInitiative/libelektra/blob/2f20bd4acb740996658cd46b9d0e5c05bd36981e/tests/data/data_escape.c#L26-L27

2. In `test_key.c` there are some test cases that are wrong IMO. Adding e.g. `../../a` to `/b` should be an error, since we cannot go two levels up. Do you agree?

https://github.com/ElektraInitiative/libelektra/blob/2f20bd4acb740996658cd46b9d0e5c05bd36981e/tests/ctest/test_key.c#L1403-L1406

3. Is there any reason that `keyCompareByName` et al. in `keyset.c` use `const void *` arguments instead of `const Key *`? I know that this makes it compatible with things like `qsort`, but in that case why doesn't the public `keyCmp` use `const void *`?

And now some more conceptual things:

1. There are many functions that return `ssize_t`, but could return `size_t` and use 0 as an error case. `keyGetNameSize` is such a function. The return value can never be 0, since we include the terminating null byte in the size and explicitly state that 1 is return, if no key name exists. Should we make this change?
2. Currently there is the possibility of passing a key like `user:hello/my/key` to `keyNew`. This results in a key with name `user/my/key` and metadata `owner = hello`. There is a comment in the function `elektraHandleUserName` that indicates this is legacy behaviour:
https://github.com/ElektraInitiative/libelektra/blob/2f20bd4acb740996658cd46b9d0e5c05bd36981e/src/libs/elektra/keyname.c#L380
 I assume this means, I don't need to handle that case and can remove the function?
3. If I am not mistaken, the `options` argument of `elektraKeySetName` has no use with the new key names. Am I missing something, or can we remove the argument and therefore re-integrate `elektraKeySetName` into `keySetName`?
4. `keySetName` and `keyAddName` currently don't distinguish between different kinds of errors, is that something we want? If so how much distinction should there be?
5. I noticed that `keyNew` creates a key without a name, if `elektraKeySetName` fails (e.g. because of an invalid name). I changed it so that `keyNew` (and `keyVNew`) return `NULL` in this case. This also required changing `keyVInit` so that it returns a value.

And last a few things that are still completely open:

1. Right now both `keyNew (0, KEY_END)` and `keyNew ("", KEY_END)` create a key with name `""`. IMO one of the two should fail. The current (somewhat broken) solution in this PR is that `keyNew (0)` creates a key with no name (i.e. `key->key == NULL`), but externally nothing changes, `keyName (keyNew (0))` still returns `""`. However, `keyNew ("", KEY_END)`  fails because `""` is not a valid key name. Note: this behaviour is implemented in `elektraKeySetName`, so `keySetName (key, 0)` free the allocated name buffers.
2. What about backwards compatibility? The introduction of the colon `:` after the namespace already means that many `keyNew` and `keySetName` calls are broken. And I don't think we should provide any backwards compatibility for that in general. However, we might want to add an `#ifdef ELEKTRA_KEYNAME_COLON_COMPAT` (like in https://github.com/ElektraInitiative/libelektra/issues/2983#issuecomment-534949292) that allows use using old key names without the colon. The `#ifdef` would enable code in `elektraKeyNameCanonicalize` that translates the old keys into new ones. I would however only add compatibility for the colon. Any other incompatibilities should not be addressed since they should be rare.
3. Should the functions `elektraKeyNameValidate`, `elektraKeyNameCanonicalize`, `elektraKeyNameUnescape` and `elektraKeyNameEscapePart` be made public? `elektraKeyNameValidate` and `elektraKeyNameEscapePart` don't make any assumptions about the name/part given, but `elektraKeyNameCanonicalize` assumes a name that is valid according to `elektraKeyNameValidate` and `elektraKeyNameUnescape` assumes a name that is canonical according to `elektraKeyNameCanonicalize`.

src/libs/elektra/key.c
```diff
@@ -403,6 +419,7 @@ int keyCopy (Key * dest, const Key * source)
 
 	// free old resources of destination
 	if (!test_bit (dest->flags, KEY_FLAG_MMAP_KEY)) elektraFree (destKey);
+	if (!test_bit (dest->flags, KEY_FLAG_MMAP_KEY)) elektraFree (destUKey);
```
mpranj: The `KEY_FLAG_MMAP_KEY` is a flag for the Key->key member. If I understand this correctly `ukey` is a member which is separately allocated. Then we need to add a separate flag for this field.

src/plugins/mmapstorage/mmapstorage.c
```diff
@@ -966,6 +966,10 @@ static void updatePointers (MmapMetaData * mmapMetaData, char * dest)
 		{
 			key->meta = (KeySet *) ((char *) (key->meta) + destInt);
 		}
+		if (key->ukey)
+		{
+			key->ukey = key->ukey + destInt;
+		}
```
mpranj: Sounds like the right direction, but there are other places we need to change. We will also need to copy this data into the `mmap` region.

markus2330:
> I added a separate ukey field for the unescaped name to struct _Key. This simplifies a lot of things. I hope this is not a problem. (@mpranj I also hope I modified mmapstorage correctly, please check.)

It is a performance thing: ideally `keyNew` would be two mallocs: one for the struct (so that the identity will stay the same throughout the lifetime) and one for all dynamic parts that might need resizing. One more separate field brings us one step further from this situation. But this optimizations can also be done after 1.0.

> I have no idea, why the old escaping function worked this way. Can you maybe shed some light on that?

If your formalization brings different results in this cases, it is okay to change this. Actually the best test would be to escape und unescape random byte sequences and see if you receive the same bits.

> I assume this means, I don't need to handle that case and can remove the function?

Yes, please remove this owner stuff. It is not working anyway and it is unlikely we can have it for 1.0.

> keySetName and keyAddName currently don't distinguish between different kinds of errors, is that something we want? If so how much distinction should there be?

@PhilippGackstatter didn't you already make this suggestion? Do you already work on this?

> What about backwards compatibility?

Please do not provide any. Elektra 1.0 should only understand key names with the colon. We make this change either fully or not.

> Right now both keyNew (0, KEY_END) and keyNew ("", KEY_END) create a key with name ""

What about making this fail? (letting keyNew return NULL) Is there some code that really needs these empty key names?

From the user view it would be nice to have an invariant that keys have a valid name. (Especially, if you already changed that wrong names let keyNew fail. The empty name would simply be yet another wrong name.)

> Should the functions elektraKeyNameValidate, elektraKeyNameCanonicalize, elektraKeyNameUnescape and elektraKeyNameEscapePart be made public? 

Maybe somewhere in the future during 1.0. At the moment we make as little as possible public.

kodebach:
> It is a performance thing: ideally keyNew would be two mallocs: one for the struct (so that the identity will stay the same throughout the lifetime) and one for all dynamic parts that might need resizing. One more separate field brings us one step further from this situation. But this optimizations can also be done after 1.0.

I don't think blindly aiming for less mallocs is a good idea. Having two smaller buffers instead of one big one might for example affect cache behaviour. (e.g. `ukey` is accessed more so it is kept in cache, but `key` is not accessed as much). This is pure speculation and without really good benchmarks I would prefer we decide based on code simplicity and readability.

Apart from that using only 2 mallocs is pretty hard (and definitely was not the case for the old `keyNew`) when array index prefixing is involved. AFAIK the only way to calculate the name buffer size is to iterate the whole string to check for array indices.

> Actually the best test would be to escape und unescape random byte sequences and see if you receive the same bits.

There is a test that checks all 2 byte sequences (with an extra flag it checks all 4 byte sequences). But I can add some more edge cases. Actually random tests are not good for reproducibility and just generating some fixed strings and using them doesn't really make for good tests, as most strings are won't involve any escaping.

> @PhilippGackstatter didn't you already make this suggestion?

He did yes, my question was more about the "how much granularity do we want".

> Is there some code that really needs these empty key names?

There are definitely some tests that create a key with an empty name and then use `keyAddName`. There might also be some other places where something like this is used. There may also be some cases where you need a `Key`, but don't care about its name and want to avoid the malloc. For example if the `Key` is used for error handling.

> From the user view it would be nice to have an invariant that keys have a valid name. (Especially, if you already changed that wrong names let keyNew fail. The empty name would simply be yet another wrong name.)

I agree. The easy solution to this would be to allow `keyName` to return `NULL` (and `keyGetNameSize` to return 0). Then the invariant would be `keyName` returns a valid name or `NULL`, if the key doesn't have a name. We could also use `""` instead of `NULL`, which would avoid segfaults.

src/libs/elektra/key.c
kodebach: I don't think we do. `key->key` and `key->ukey` are always changed at the same time. Either both are `NULL` or both are correctly allocated.

PhilippGackstatter:
> @PhilippGackstatter didn't you already make this suggestion? Do you already work on this?

It's broadly covered in #3029, but I am not working on it.

src/libs/elektra/key.c
mpranj: In that case you're right, I was not aware of that.

markus2330: Such invariants should also be documented in the struct.

markus2330:
> This is pure speculation and without really good benchmarks

In benchmarks I saw that malloc and elektraUnescapeKeyName have about the same costs.. And it is not pure speculation that calling malloc several times is more expensive than calling it once.  I do not see how this could negatively affect cache behavior, as having everything in a block should not be worse than having it scattered around in the memory. But of course I cannot exclude this possibility.

> I would prefer we decide based on code simplicity and readability.

I agree. As already said, we can optimize this later when it really becomes a bottleneck. Currently malloc usually is only about 5% of the total costs (in a `kdb ls`). So to not optimize it now, is the correct decision.

>  He did yes, my question was more about the "how much granularity do we want".

I think the proposal in #3029 is good. You can say there which other cases you think of.

>  There are definitely some tests that create a key with an empty name and then use keyAddName. There might also be some other places where something like this is used. There may also be some cases where you need a Key, but don't care about its name and want to avoid the malloc. For example if the Key is used for error handling.

When grepping for `'keyNew(0'` or `'keyNew(""' *` I find only very few occurrences. It should not be a big deal to make a `'keyNew("/"'` in these cases instead. Then we could also remove this ugly cases like in kdb.c line 1501 where invalid keys are passed.

> I agree. The easy solution to this would be to allow keyName to return NULL (and keyGetNameSize to return 0). Then the invariant would be keyName returns a valid name or NULL, if the key doesn't have a name. We could also use "" instead of NULL, which would avoid segfaults.

I would prefer if an invalid key is completely impossible.

kodebach:
Okay so to sum up:

1. `Key` should have an invariant that the key name is always valid. Both `keyNew (0, KEY_END)` and `keyNew ("", KEY_END)` should return `NULL`. For error handling `keyNew ("/", KEY_END)` is the recommended one, if the key name isn't important and you don't have any other key to use.
2. Return codes should give some indication on the type of error that occurred.
3. No backwards compatibility will be provided.
4. No additional functions will be made public right now.
5. All the owner stuff will be removed.
6. Test cases that don't agree with the specification in #3102 will be adapted.

Questions still open:

1. Should we replace `ssize_t` with `size_t`, if `return 0` can be for errors?
2. `const void *` vs `const Key *` in `keyCmp`, `keyCompareByName`, et al. ?
3. Is there any use left for the options argument in `elektraKeySetName`?

kodebach:
A few more things came up:

1. The `keyGetX`, `keySetX` functions for X = `UID`, `GID`, `Owner`, `Dir`, `Mode`, `ATime`, `MTime` and `CTime` are all marked deprecated. I assume I can ignore/remove these and comment out the test cases?
2. `keyIsInactive` checks the key for parts starting with `.` everytime it is called. Should we instead do the check during `keySetName` and store the state in a flag?

markus2330:
> The keyGetX, keySetX functions for X = UID, GID, Owner, Dir, Mode, ATime, MTime and CTime are all marked deprecated. I assume I can ignore/remove these and comment out the test cases?

Yes you can remove all of them (including test cases), I wrote about that just before in #3124 

> keyIsInactive checks the key for parts starting with . everytime it is called. Should we instead do the check during keySetName and store the state in a flag?

keyIsInactive can also be removed. If we reintroduce inactive keys, we will use meta data for it. It is not needed for 1.0.0.

kodebach:
@markus2330 I stumbled upon a problem with the `trie`. (I think) it relies on the existence of keys with empty names. Also, the trie operates on escaped key names AFAICT. We should probably change that as well. The `_Trie` struct is not well optimized in general... It uses 4 arrays of pointers/`size_t` each of length 256, that means a single `_Trie` struct occupies 8 KiB of memory on a 64 bit machine (unless I am mistaken), most of which will be empty space (zeros).

kodebach:
Although on second though, I think `Key`s are only used for `trieLookup`... We could easily use strings there too

markus2330:
Yes, the trie internally does not know about keys anyway. I would not optimize there anything unless you find a problem. The 8 KiB should be no problem as they are only needed per mountpoint. Afaik this part was bug free since it was written.

src/libs/elektra/keyhelpers.c
```diff
@@ -197,10 +197,8 @@ int keyNameIsSystem (const char * name)
  */
 int keyInit (Key * key)
 {
-	if (key->meta) ksClear (key->meta);
-	memset (key, 0, sizeof (struct _Key) - sizeof (KeySet *));
-
-	if (!key->meta) key->meta = ksNew (0, KS_END);
+	memset (key, 0, sizeof (struct _Key));
+	key->meta = ksNew (0, KS_END);
```
mpranj: Is it possible to not allocate key->meta eagerly? It was discussed in #3142 to not do this.

kodebach: This code is based on the state before #3142.
If it is possible in #3142, then it should be possible here too. I will wait for #3142 to be merged an then integrate it's changes. (Same goes for any other PRs that make changes to the core)

kodebach:
@markus2330 The current situation is this:
I decided to implement `elektraKeySetName` (and `keyAddName`) in 3 steps:
1. Call `elektraKeyNameValidate`.
   - If it returns 1, the key name is valid (in `keyAddName` we use the existing name as a prefix). Both keySize and keyUSize will be updated to the correct values and we can call `elektraRealloc` and continue.
   - If it returns 0, the key name is NOT valid. Neither keySize nor keyUSize will have been modified and we return immediately to ensure the key still in a valid state when we return.
2. Call `elektraKeyNameCanonicalize`. The function has no return type, since it assumes a valid name and therefore cannot fail.
3. Call `elektraKeyNameUnescape`. The function has no return type, since it assumes a valid canonical name and therefore cannot fail.

Most of this works already, but I have some problems with `..` parts in `elektraKeyNameValidate`. I can't quite figure out, how to best calculate the correct `keySize` and `keyUSize`, if `..` parts are involved. The only solution I could think of that should (not tried yet) work without problems is to do a second pass over the key name starting from the back.
However, I think there is a simpler (and probably faster) solution, if we introduce a limit on the number of parts in a key name. Let's say 1024. (Note: the limit is only so that I don't need to malloc) What do you think about that?

markus2330:
Thank you for the update! Sounds very good!

> elektraKeySetName

Shouldn't it be called keySetName? Or is elektraKeySetName still needed?

kodebach:
> Shouldn't it be called keySetName? Or is elektraKeySetName still needed?

No its not needed anymore, I just didn't update that part yet.

kodebach:
I ended up validating the whole key name (except for the namespace) in reverse, but there are now working versions of `elektraKeyNameValidate`, `elektraKeyNameCanonicalize` and `elektraKeyNameUnescape`. Now it should only be a matter of going through the whole code base and updating all key names ...

markus2330:
Really great work is happening here, thank you so much!

> Now it should only be a matter of going through the whole code base and updating all key names ...

I do not envy you for that. Maybe @Piankero can help you out a bit?

kodebach:
> Maybe @Piankero can help you out a bit?

Don't think that is necessary or as easy as you'd think. It happens almost on every second test I update that I discover some kind of minor bug in my implementation (mostly off-by-one errors).

If everything works, changing the key names can be done quite easily with regex replace, since it is mostly just inserting the `:` after the namespace.

markus2330:
> If everything works, changing the key names can be done quite easily with regex replace, since it is mostly just inserting the : after the namespace.

I am afraid it will not be so easy, as there are e.g. at least 79 places (grep -rH '"user"' * | grep keyNew | wc -l) where it is simply "user" without an indication if this is a key name or something else (especially also in the docu, where not necessarily a "keyNew" is in the same line.

And user, user:, user/ were all valid and now should be replaced with user:/.

But I agree, it is minor to the other work you did here.

kodebach:
Just a sidenote: It probably makes sense to squash the commits in this PR when merging. There are quite a few commits now in here that don't even compile or were more or less completely undone by later commits...

markus2330:
> Just a sidenote: It probably makes sense to squash the commits in this PR when merging. There are quite a few commits now in here that don't even compile or were more or less completely undone by later commits...

This does not matter, a longer history does not hurt.

kodebach:
> @markus2330 I stumbled upon a problem with the trie. (I think) it relies on the existence of keys with empty names.

The problems is actually not just with the trie (which I changed to use normal strings), but the whole backend/mount system relies on keys with empty names. I am not sure yet how to solve this, but I think it probably requires some deeper changes than I first thought.

markus2330:
I am afraid #3115 will see quite a few merge conflicts. Do you want me to inform you about this?

E.g. #3179

kodebach:
> I am afraid #3115 will see quite a few merge conflicts.

That's fine, I don't really look at the build server results at this time. Once I change this PR from a draft PR to a real PR, we should discuss, which PRs are merged before this and which after, since the ones that are merged afterwards, will have to be updated. My plan is to keep this PR as a draft until `make run_all` succeeds on my machine, or at the very least until a simple `kdbOpen, kdbGet, kdbSet, kdbClose` works.

kodebach:
The docu for `struct _Backend` says:

https://github.com/ElektraInitiative/libelektra/blob/9a3413f7dee83e8d7d628056092781b4088f1269/src/include/kdbprivate.h#L337

Is it correct that after #2969, a backend should consist of a single backend plugin that calls the other plugins? Because then the whole `struct _Backend` is probably unnecessary, we could just store `Plugin *` in `struct _Trie`, since we always call same plugin for all positions anyways.

I am asking because as stated above, I discovered that we use a Key with and empty name as the `mountpoint` in a `struct _Backend` for the default backend. Since the empty string is not a valid key name, this won't be possible after #3115, so I am looking to what needs to change so that we don't need the empty names anymore. Removing `struct _Backend` would solve most of the problem, I think.

Another solution would be to replace the `Key *` in `struct _Backend` with a `const char *`, but I haven't discovered yet, whether or not the key value is actually used for anything useful. As far as I've seen so far, we just write things like "default" and "root" into the key value, but I haven't found where the key value is used yet.

kodebach:
> The docu for `struct _Backend` says: [...]

Ignore that, it seems like removing the `struct _Backend` would be a huge task, so even if it is possible and we want to do it, waiting for that is not an option for this PR.

PhilippGackstatter:
Will this PR also fix #3050?

kodebach:
> Will this PR also fix #3050?

Yes.

kodebach:
@mpranj Please take a look at the changes to mmapstorage, I think I missed something... As an example of what I mean run `valgrind --leak-check=full bin/test_storage $PWD/tests/ctest/` in your build directory:

```
==24620== Memcheck, a memory error detector
==24620== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==24620== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==24620== Command: bin/test_storage /home/klemens/data/bsc/libelektra/cmake-build-debug/tests/ctest/
==24620== 
STORAGE     TESTS
==================

Testing plugin mmapstorage_crc
Tmp-file: /tmp/elektra-test.F962RP/elektra-tmp.NLL9zP
==24620== Invalid read of size 8
==24620==    at 0x483EEE8: bcmp (vg_replace_strmem.c:1113)
==24620==    by 0x48703E3: keyCompareByName (keyset.c:516)
==24620==    by 0x48704E7: keyCompareByNameOwner (keyset.c:554)
==24620==    by 0x4870671: ksSearchInternal (keyset.c:727)
==24620==    by 0x48707B6: ksAppendKey (keyset.c:830)
==24620==    by 0x4870B36: ksAppend (keyset.c:947)
==24620==    by 0x48700E9: ksDup (keyset.c:319)
==24620==    by 0x10BC7A: test_ksDupFun (test_storage.c:122)
==24620==    by 0x12538C: main (test_storage.c:950)
==24620==  Address 0x8e4a4a0 is not stack'd, malloc'd or (recently) free'd
==24620== 
==24620== 
==24620== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==24620==  Access not within mapped region at address 0x8E4A4A0
==24620==    at 0x483EEE8: bcmp (vg_replace_strmem.c:1113)
==24620==    by 0x48703E3: keyCompareByName (keyset.c:516)
==24620==    by 0x48704E7: keyCompareByNameOwner (keyset.c:554)
==24620==    by 0x4870671: ksSearchInternal (keyset.c:727)
==24620==    by 0x48707B6: ksAppendKey (keyset.c:830)
==24620==    by 0x4870B36: ksAppend (keyset.c:947)
==24620==    by 0x48700E9: ksDup (keyset.c:319)
==24620==    by 0x10BC7A: test_ksDupFun (test_storage.c:122)
==24620==    by 0x12538C: main (test_storage.c:950)
==24620==  If you believe this happened as a result of a stack
==24620==  overflow in your program's main thread (unlikely but
==24620==  possible), you can try to increase the size of the
==24620==  main thread stack using the --main-stacksize= flag.
==24620==  The main thread stack size used in this run was 8388608.
==24620== 
==24620== HEAP SUMMARY:
==24620==     in use at exit: 7,034 bytes in 41 blocks
==24620==   total heap usage: 357 allocs, 316 frees, 515,401 bytes allocated
==24620== 
==24620== LEAK SUMMARY:
==24620==    definitely lost: 0 bytes in 0 blocks
==24620==    indirectly lost: 0 bytes in 0 blocks
==24620==      possibly lost: 0 bytes in 0 blocks
==24620==    still reachable: 7,034 bytes in 41 blocks
==24620==         suppressed: 0 bytes in 0 blocks
==24620== Reachable blocks (those to which a pointer was found) are not shown.
==24620== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==24620== 
==24620== For lists of detected and suppressed errors, rerun with: -s
==24620== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
fish: “valgrind --leak-check=full bin/…” terminated by signal SIGSEGV (Address boundary error)
```

It would be nice if you could tell me whats wrong (or fix it yourself, you should have permission to push to this branch). I need to modify some of the cache related code in `kdbGet`, since I decided to use `backend->mountpoint = NULL` instead of `backend->mountpoint = keyNew ("", KEY_VALUE, "default", KEY_END)` for default backends, since empty key names are no longer allowed.

mpranj:
> Please take a look at the changes to mmapstorage

@kodebach sure I'll take a look on friday. Sorry but I can't do much earlier.

kodebach:
> Sorry but I can't do much earlier.

No, problem I excluded the cache plugin to fix the other stuff in the meantime.

markus2330:
Is it possible to deactivate this lgtm-com notifications about FIXME?

kodebach:
> Is it possible to deactivate this lgtm-com notifications about FIXME?

AFAICT we can only disable notifications about FIXME in general, but not tell lgtm to ignore this one going forward. It's kind of ridiculous the notification is sent every time I push something...

I will replace the `// FIXME` with `// TODO` in the next push and change it back, before we merge.

markus2330:
>  AFAICT we can only disable notifications about FIXME in general

This would be okay for me.

kodebach:
@markus2330 Currently, the new key names CANNOT contain a part starting with `%`, unless the whole part is `%`. However, the C++ contextual stuff uses placeholders of the form `/%name%/`. Should we change the rules so that such parts are allowed or should we require escaping (i.e. `/\%name%/`) for the C++ contextual stuff?

markus2330:
> Currently, the new key names CANNOT contain a part starting with %, unless the whole part is %. 

Why? We should restrict possible key names (after the namespace) as little as possible. So it should be possible to use % anywhere.

The whole part being % has a special meaning but why do disallow other names with %?

> However, the C++ contextual stuff uses placeholders of the form /%name%/. Should we change the rules so that such parts are allowed

I would allow every name with a valid namespace. (Maybe with the exception of a tangling \). Why imposing restrictions?

> or should we require escaping (i.e. /\%name%/) for the C++ contextual stuff?

The escaping is already needed to tell that it is not a placeholder, so this will not work.

kodebach:
> The whole part being % has a special meaning but why do disallow other names with %?

Ok I'll change it, my interpretation was that "a part starting with % has special meaning", of which only `/%/` is currently allowed.

> I would allow every name with a valid namespace. (Maybe with the exception of a tangling ). Why imposing restrictions?

With the new version of `%` current restrictions are:
1. Keys must start with one of `/`, `meta:/`, `spec:/`, `proc:/`, `dir:/`, `user:/`, `system:/` or `default:/`.
2. After the first `/` follows a series of parts and `/`. (some combinations may not be canonical)
3. For parts we have these restrictions:
   1. Parts may not start with `@`.
   2. Parts starting with `#` must either be a of the form (regex) `#([0-9]|_+[1-9][0-9]+)`, with the additional restriction that the number of underscores must either be 0 or exactly 1 less than the number of digits.
   3. The parts `.` and `..` have special meaning. They cannot occur in canonical names.
   4. The part `%` has special meaning. It can appear in canonical names. 
   5. A `\` must be followed by a `/` or another `\`. At the start of the part, the sequences `\#`, `\.` and `\%` are also allowed (to prevent special behaviours). Backslashes may NOT occur in any other position.

Explanations:
- 1, 2, 3.iii and 3.iv are obvious.
- 3.i is just taken from the old version. These parts were always reserved.
- 3.v makes the unescaping very simple. Because we validate the key name before unescaping, we can just replace each backslash by the character following it no matter what that character is.

https://github.com/ElektraInitiative/libelektra/blob/b0854c77e6185aa7afdff8cf3cc4a25512cd746c/src/libs/elektra/keyname.c#L1005-L1008

- Lastly, 3.ii could be lifted and we could allow things like `#123key`, but I think it is better to require explicit escaping (i.e. `\#123`), when `#` is used for something other than arrays.

---

In #2698 we also discussed using `^` for arrays instead of `#`, because `#` is commonly used for comments. Do we still want to do that?

kodebach:
There are also some more restrictions, especially related to `..`. If you are interested, take a look at [test_keyname.c](https://github.com/ElektraInitiative/libelektra/blob/b0854c77e6185aa7afdff8cf3cc4a25512cd746c/tests/ctest/test_keyname.c).

kodebach:
And another question: Should `keyBaseName` continue to return an unescaped name or should it return the escaped base name? Should we provide both versions?

markus2330:
Can you please also document these rules/restrictions?

Btw. did you also read #38?

> 1. Keys must start with one of [...]

Yes, definitely.

> 2. 

is not a restriction. This is also documented by your grammar.

>      Parts may not start with @.

This was an informal restriction so that we can give @ a special meaning without conflicting to what applications do. (Same as %). Maybe we should reserve even more characters for our use. I opened #3223.

These characters should still be allowed, but it should also possible to escape them. (So that applications can use these characters without their meaning for Elektra).

 >       Parts starting with # must either be a of the form (regex) #([0-9]|_+[1-9][0-9]+), with the additional restriction that the number of underscores must either be 0 or exactly 1 less than the number of digits.

Yes, I like this restriction. But why allowing _ at all? Let us simply go for [0-9]|[1-9][0-9]+ (see #2698)

>        The parts . and .. have special meaning. They cannot occur in canonical names.
>        The part % has special meaning. It can appear in canonical names.

Yes, but \. or \% can exist, so . and % is allowed as (escaped) basename.

>        A \ must be followed by a / or another \. At the start of the part, the sequences \#, \. and \% are also allowed (to prevent special behaviours). Backslashes may NOT occur in any other position.
> 3.v makes the unescaping very simple. Because we validate the key name before unescaping, we can just replace each backslash by the character following it no matter what that character is.

I fully agree that this simplification makes sense. But the % and @ (or even others) should be escapable at every position.

>   Lastly, 3.ii could be lifted and we could allow things like #123key, but I think it is better to require explicit escaping (i.e. \#123), when # is used for something other than arrays.

I agree to be very strict about that.

> In #2698 we also discussed using ^ for arrays instead of #, because # is commonly used for comments. Do we still want to do that?

If we already have natural sorting: why not simply require no character in front? So /1/ is an array, `/\1/` not (as it is escaped).  See #2698

Btw. I am open for any changes if you can implement them correctly in reasonable time and there is some technical improvement (like easier to understand, simplification, more robust, more extensible, faster; see doc/GOALS.md). But slowly slowly this PR should find an end :smile: 

> And another question: Should keyBaseName continue to return an unescaped name or should it return the escaped base name?

It makes sense to me that `keySetBaseName`, `keyAddBaseName`, and `keyBaseName` set/get the same (escaped) name. And that this name can be anything and is guaranteed to have no meaning.

> Should we provide both versions?

If we find a nice name consistent to all the other *Name functions.

kodebach:
> Can you please also document these rules/restrictions?

Once the code is finished, I will add documentation as well, yes. Until then, this PR and the issue #3102 serve as documentation.

> Btw. did you also read #38?

Not sure, what you want to tell me here. The issue has been closed in 2014 and seems to be about problems in the mountpoint resolution. IMO it has nothing to do with key names, at least not directly.

>> 2.
> is not a restriction. This is also documented by your grammar.

Well, it is a restriction on the set of allowed key names. Whether it is covered by the grammar or not is another thing entirely.

> But why allowing _ at all? Let us simply go for `[0-9]|[1-9][0-9]+` (see #2698)

You mean only prefixing the array in the unescaped name? We could probably do that, I would have to update many tests again and there are some problems that need addressing. See https://github.com/ElektraInitiative/libelektra/issues/2698#issuecomment-554636250.

> But the % and @ (or even others) should be escapable at every position.

See https://github.com/ElektraInitiative/libelektra/issues/3223#issuecomment-554641557.

> But slowly slowly this PR should find an end

Right now, everything seems to work with the rules I decided on. The only thing still open is fixing all tests.

markus2330:
> Not sure, what you want to tell me here. The issue has been closed in 2014 and seems to be about problems in the mountpoint resolution. IMO it has nothing to do with key names, at least not directly.

The problem was in the key names. This issue lead to the escaping implementation we have now. The later benchmarks could be of interest.

> Right now, everything seems to work with the rules I decided on.

Good that there is progress!

src/libs/tools/tests/testtool_umount.cpp
```diff
@@ -60,25 +60,27 @@ TEST (Umount, InvolvedRoot)
...
 }
 TEST (Umount, InvolvedCascading)
 {
 	testUmount ("/is///a//./more/complex/../complicated///issue//.");
 }
 
+// TODO (kodebach): what is oldmount ??
+
 /**
  * @brief Test umount with compatibility umount-names
```
kodebach: @markus2330 What are these "old mount" tests below? Do we still need them or can I remove them?

kodebach: Here is the permalink: https://github.com/ElektraInitiative/libelektra/blob/3982b81ed209964ff598db8bedf97fa02c7320c5/src/libs/tools/tests/testtool_umount.cpp#L85-L139

markus2330: If we get rid of the fallback code (see permalink below), then we can also remove the tests. But maybe in a different PR?

https://github.com/ElektraInitiative/libelektra/blob/3982b81ed209964ff598db8bedf97fa02c7320c5/src/libs/elektra/kdb.c#L211-L249

kodebach: > But maybe in a different PR?

Exactly what I didn't want...

`testtool_umount` currently fails because of these "old mount" tests. I haven't found the cause yet, that's why I am asking, if removing them would be okay.

markus2330: Yes, it is okay. doc/decisions/bootstrap.md also needs to be adapted, then.

kodebach: Please check, that I adapted `doc/decisions/bootstrap.md` correctly.

kodebach:
I had to update the `augeas` plugin, to handle comments. I have no idea how Augeas works and I suspect that I just implemented a workaround and did not fix the actual problem. @markus2330 Do you now more about Augeas? Can you check?

doc/decisions/bootstrap.md
```diff
@@ -4,7 +4,7 @@
 
 Currently the default backend (default.ecf) will also be used for bootstrapping. There are two problems with this approach:
 
-1. Thus the default backend first will be read with parentKey `system/elektra` and later with parentKey `system`, it needs to store absolute paths and thus won't work with the current INI plugin
+1. Thus the default backend first will be read with parentKey `system:/elektra` and later with parentKey `system`, it needs to store absolute paths and thus won't work with the current INI plugin
```
markus2330:
```suggestion
1. Thus the default backend first will be read with parentKey `system:/elektra` and later with parentKey `system:/`, it needs to store absolute paths and thus won't work with most of the plugins (except dump).
```

doc/decisions/bootstrap.md
```diff
@@ -19,7 +19,7 @@ Currently the default backend (default.ecf) will also be used for bootstrapping.
 
 ## Considered Alternatives
 
-- Implement a hack so that `system/elektra` is actually read as `system`. (Will not solve problem 2.)
+- Implement a hack so that `system:/elektra` is actually read as `system`. (Will not solve problem 2.)
```
markus2330:
```suggestion
- Implement a hack so that `system:/elektra` is actually read as `system:/`. (Will not solve problem 2.)
```

markus2330:
> I had to update the augeas plugin, to handle comments. I have no idea how Augeas works and I suspect that I just implemented a workaround and did not fix the actual problem. @markus2330 Do you now more about Augeas? Can you check?

@fberlakovich wrote it. I think the only proper solution is to iterate over the whole augeas name and add all parts with keyAddBaseName.

The docu at https://github.com/hercules-team/augeas/wiki/Path-expressions describes the valid expressions but I could not find which paths actually could be returned from the augeas tree.

If it helps you, I can ask on the augeas list which paths might be returned from a aug_match.

kodebach:
> I think the only proper solution is to iterate over the whole augeas name and add all parts with keyAddBaseName.

I had the same idea, but I didn't know whether augeas paths are like UNIX paths and every `/` separates to parts, or like key names and `/` could be escaped.

We also need the reverse, constructing the augeas path from the unescaped key name. For that we need to know, if there are any strings that are not valid parts in an augeas path and if so, whether there is a way to escape/quote these.

> The docu at https://github.com/hercules-team/augeas/wiki/Path-expressions describes the valid expressions

I saw that, but the text is not very precise and the "Formal" grammar is just broken. They refer to "NameTest", which is never defined. They also say they are using EBNF but it seems they are using some kind of regex syntax, which is not defined anywhere and not part of EBNF.

markus2330:
Ok, this seems to be quite a big topic. I think we will simply fix the testcase (or is this already done?) and leave this as issue for post-1.0. For non-trivial files, augeas is not the right backend anyway.

kodebach:
The test case is fixed (I added a workaround for `/#comment`). If we don't fix the plugin it self, we should make it `experimental` or even `unfinished` and definitely remove the `maintained` status.

fberlakovich:
I wish I could answer the questions, but I can't either. Whenever I had such an issue with augeas I resorted to trial and error or studying the source code, which, given the complexity of augeas, is not very pleasant.

markus2330:
@fberlakovich thank you for the quick answer! Yes, I agree that this needs quite some effort. Maybe the augeas people themselves need to find out what the path actually can contain as the EBNF seems to be only a draft.

> The test case is fixed (I added a workaround for /#comment). If we don't fix the plugin it self, we should make it experimental or even unfinished and definitely remove the maintained status.

@fberlakovich what do you think? It would be great if we can keep you as maintainer. @tom-wa is currently using the augeas plugin for writing configuration files via https://www.ansible.com. This is quite promising as the augeas library is not so popular in Ansible (compared to Puppet, where it is commonly used, good integrated and Elektra has a much more difficult position).

fberlakovich:
> @fberlakovich thank you for the quick answer! Yes, I agree that this needs quite some effort. Maybe the augeas people themselves need to find out what the path actually can contain as the EBNF seems to be only a draft.

Asking the augeas people seems to be the right way as it possibly saves hours of debugging.

> @fberlakovich what do you think? It would be great if we can keep you as maintainer. @tom-wa is currently using the augeas plugin for writing configuration files via https://www.ansible.com. This is quite promising as the augeas library is not so popular in Ansible (compared to Puppet, where it is commonly used, good integrated and Elektra has a much more difficult position).

Unfortunately my time is still very limited and I don't think I can consistently maintain it (I already couldn't in the past). However, I am happy to answer questions whenever I can or help potential new maintainers with the handover. I've seen that the plugin was already extended and improved a lot 👍 .

As for the general purpose of the plugin: Great to hear that @tom-wa uses it for ansible. Nonetheless I am afraid that the plugin will always be limited by the level of abstraction Augeas's lenses can provide.

markus2330:
Thank you for your willingness to help. Nevertheless, I think then it is best to remove "maintained" and maybe even add "experimental" and/or "unfinished" depending on the experience of @tom-wa.

kodebach:
On my local machine all tests (for which I have the dependencies installed) now run successfully. I sometimes (can't say when exactly) get this mysterious error:

```
Sorry, module kdb issued the warning C01320:
Interface: Postcondition of backend was violated: drop key system:/elektra/modules/cache/exports/close not belonging to "system:/elektra/modules/cache" with name "modules" but instead to "(null)" with name "(null)" because it is hidden by other mountpoint
```
Note: Currently I use `NULL` as the mountpoint of the default backend. That's where the `"(null)"` string come from.

Once the error occurs, all following shell tests that use the `kdb` fail with the same error. Manually removing the storage file for the default backend fixes the error. 

I will now write documentation and then wait until #2963 is merged before I do anything else in this PR. The reason is that #2963 removes much of the backend code from the core and therefore will fix some of the remaining problems.

doc/proposals/keynames.md
```diff
@@ -0,0 +1,226 @@
...
+  - If `u` is _u-part-valid_ then `pushU(k, u) := push (k, J(u), u)`.
+  - Otherwise `pushU(k, u)` is undefined.
+- `pushL(k, s)`: helper for adding a "literal" part `s` to the key `k`.
+  This can be used to avoid a part having special meaning.
+  Defined as:
+  - If `s[0]` is in `R` then `pushL(k, s) := pushU (k, \:s)`.
+  - Otherwise `pushL(k, s) := push (s, s)`.
+- `putE(k, e)`, `putU(k, u)` and `putL(k, s)` are the same as removing the last part of `k` and then calling `pushE(k, e)`, `pushU(k, u)` and `pushL(k, s)` respectively.
+
+### Storage Plugins
```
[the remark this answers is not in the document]
kodebach: I will update the storage plugin tutorial, once this is actually implemented.

doc/proposals/keynames.md
```diff
@@ -0,0 +1,226 @@
...
+
+Storage plugins for flat formats (without a native nesting hierarchy) should just use the `keyName` to store values (and metadata, if possible).
+
+Storage plugins with support for a native hierarchy should use the operations `store` and `load` to find the names to use in that hierarchy (e.g. field names in JSON).
```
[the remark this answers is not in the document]
kodebach: See the note further down in the file.

No, neither the current `keySetBaseName`/`keyBaseName` nor any proposed form have the correct semantics for this. That is why line 70 reads (emphasis added):

> The following operations would be provided in a **separate library** for storage plugins:

doc/proposals/keynames.md
```diff
@@ -0,0 +1,226 @@
...
+I will use JSON as an example, it requires that a `\` is escaped as `\\`.
+
+##### Store procedure
+
+If there are keys below the key `k`, store the value of `k` in the nested field `\value`:
+
+```json
+"k": {
+    "\\value": "value of k",
+    ...
```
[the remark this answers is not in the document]
kodebach: This was just intended as an example for you. Of course actual documentation will contain full and correct examples.

doc/proposals/keynames.md
```diff
@@ -0,0 +1,226 @@
...
+    "\\array": [ ... ]
+}
+```
+
+The benefit of this solution is, it also works for mixed array and non-array parents.
+However, I wouldn't generally recommend using such structures.
+
+##### Load procedure
+
+If we encounter a field named `\value`, we store its value not in `p/\value` (which is invalid anyway), but instead in the parent key `p`.
```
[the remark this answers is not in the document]
kodebach: `p` just stands for any key name. I should have clarified that.

doc/proposals/keynames.md
```diff
@@ -0,0 +1,226 @@
...
+    "\\array": [
+        "element1",
+        {
+             "element2": 2
+        }
+    ]
+}
+```
+
+These would correspond to the `ni` format files:
```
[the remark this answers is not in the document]
kodebach: I'll improve that in the final documentation, if we decide to implement this.

doc/proposals/keynames.md
```diff
@@ -0,0 +1,226 @@
...
+For example a directory key with value and metadata in JSON:
+
+```json
+{
+"key": {
+    "\\value": "value of key",
+    "\\meta": {
+        "type": "string"
+    },
+    "subkey": 1
```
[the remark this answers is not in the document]
kodebach: > This would implicitly create metadata `type=long`.

Please don't assume anything based on knowledge of existing plugins here. In an actual JSON plugin this should create `type=long` metadata yes. The plugin also shouldn't store the `type` metadata of strings, since it is redundant. But again, I wasn't trying to show, how a JSON plugin should work. I only explained how directory keys, metadata and values can be stored together in a format without native support for that. JSON is used, only because it is easy to read and well known.

doc/proposals/keynames.md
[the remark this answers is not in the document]
markus2330: Yes, but we definitely cannot call it store/load and the API is essential here for this decision (does it integrate nicely to the other API functions).

Do I understand correctly that in essence there would be 3 different forms of key names:

- the "escaped key name" in which storage plugins render key names
- the "unescaped key name" used for comparing keys
- the "key name" which will be used to create keys/GUIs/CLIs/...

Maybe `keyGetEscapedName(Key *, char *, size_t len)?? and keyGetEscapedBaseName? Or which API do you propose?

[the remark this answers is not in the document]
kodebach: > Yes, but we definitely cannot call it store/load and the API is essential here for this decision (does it integrate nicely to the other API functions).

What is the problem with using the placeholders `store` and `load` for now? If you prefer we could use `storageEncode` and `storageDecode`.

> Do I understand correctly that in essence there would be 3 different forms of key names:

We would still only store the (canonical) escaped key name in `key->key` and the unescaped key name in `key->ukey`. They both serve the same purpose as before, but unescaped key names will be interpreted a bit differently.

Before the meaning of an unescaped key name essentially depended on the context. `#0` might be the first array element, if it is accessed or specified as such, but it might also just be a key part that is called `#0`; no other meaning at all.
Now every unescaped key name has a well-defined intended meaning. `#0` is intended to be the first array element. If it isn't used as such, this is could be considered a bug. The important idea here is that `\#0` is intended to be just a key part named `#0`. The same goes for any other unescaped part starting with `\` its intended meaning is just the rest of the string.

Storage plugins could also still use `key->key` for storage (like `quickdump` and `dump` do for example).

If they are more JSON-like and need to deconstruct the key name, simply splitting into unescaped parts also still works. But then there might be confusion around things that can be represented natively in the format.

For example in a format that has native arrays there is no good reason to interpret the name `#0` as the first array element, but if the plugin simply splits into unescaped parts, this would be the case.

> the "key name" which will be used to create keys/GUIs/CLIs/...

For creating the same (canonical or non-canonical) escaped key name format as ever is used.

How keys are displayed in user interfaces entirely depends on the interface.

A CLI or any other interface that represents a key in a single string should use the same format as `keyNew`, which has exactly these requirements.

A GUI can have so many shapes and forms, it is hard to say what should be used.
If the GUI is table-like the escaped name lends itself, as it is a single string that can be put into a single column.
If tree-like hierarchies are used splitting the unescaped name into parts and using these as node names is a good start. How these unescaped parts are shown is a similar question to how are field names in a JSON-like nested object format mapped to key parts. You could just use the exact unescaped parts, or you could create native representations for some or all special parts (array-elements in blue text, parts starting with dot on red background, etc). In which case you can remove leading backslashes just like you would in storage plugins.

> Maybe `keyGetEscapedName(Key *, char *, size_t len)?? and keyGetEscapedBaseName? Or which API do you propose?

Could we please first decide whether or not we even want to do this at all. It is pointless to discuss about function names, if we don't even know whether or not we are going to to do anything.

kodebach:
@markus2330 Do you know which storage plugins support metadata? Is there some way to easily find out, other than looking at the code?

mpranj:
Screenshot 2019-11-28 at 15 40 37

kodebach:
> [Screenshot]

@mpranj What do you want to tell us? Yes, the PR is insanely huge, but it can't be any smaller... Changing a basic fact about key names (colon after namespace) will naturally break everything.

You could use [the VS Code Github Extension](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github), if you are interesting the contents of this PR. It seems it is able to handle the enormous size of the PR.

mpranj:
@kodebach I just thought it's funny that github cannot handle this PR. It's clear that the change implies huge changes in the complete codebase.

kodebach:
@markus2330 Whats the plan for the Python 2 binding? [Are we going to remove it soon?](https://pythonclock.org/) Or do I have to fix it?

mpranj:
> Whats the plan for the Python 2 binding?

It will be removed soon via #3291.

kodebach:
I hit a problem with `testshell_markdown_kdb_global_umount`. It breaks other tests, because `system:/elektra/modules` keys are serialized to disk.

The specific circumstances in which this caused a problem is the `Backup-and-Restore` directive in the markdown shell tests. For some reason it doesn't just backup the keys below the given key, but the whole namespace. So in `kdb-global-unmount.md`, we make a backup of all of `system/` because of the line `Backup-and-Restore: system:/elektra/globalplugins`. The restore then causes the following error:

```
Sorry, module kdb issued the warning C01320:
        Interface: Postcondition of backend was violated: drop key system:/elektra/modules/cache/infos/description not belonging to "system:/elektra/modules/cache" with name "modules" but instead to "(null)" with name "(null)" because it is hidden by other mountpoint
```

Also the file `$KDB_DB_SYSTEM/elektra.ecf` then contains all the `system/elektra/modules` keys.

I don't know why this didn't happen before the key name update. Does anyone have an idea, where these keys should be removed or what else should happen to avoid this problem?

markus2330:
> @markus2330 Do you know which storage plugins support metadata? Is there some way to easily find out, other than looking at the code?

Do you mean being able to store arbitrary metadata? Then it should be:

- ni
- dump
- quickdump
- mmapstorage
- ini

Hopefully I did not forget one.

kodebach:
> Do you mean being able to store arbitrary metadata? Then it should be:

Yes, thanks. I think `xerces` is also one of them.

> Hopefully I did not forget one.

We'll find out if someone notices incompatible files, because of the `meta:/` namespace.

mpranj:
Hey @kodebach, sorry this PR has been sitting around for so long. Since we had a fresh release this week I think it could make sense to rebase & finish this up. The backend plugin will take a bit longer so we could actually merge this PR first.

What do you think about including it in the next release and do you even have time to pick up and finish this work?

kodebach:
I created #3447. It contains the same commits as this PR, but squashed and rebased onto `master` (rebasing without squashing would have been a nightmare).

> What do you think about including it in the next release and do you even have time to pick up and finish this work?

I don't know when the next release is going to be, but I am quite certain that I won't have time to finish this PR before July. However, if somebody else has the time, feel free to help. IIRC the main part of this PR is finished, it is just a matter of updating code that was added since last year and some bug fixing.

PR #3125, opened 2019-10-28
SWIG performance

manuelm:
Hi Markus,

as requested I've looked into this in more detail. The first commit fixes one of the extra copies. Fixing the other was only possible by providing a `char *`-alternative (making the first commit actually unnecessary). As all SWIG wrapped languages operate on char pointers wrapping a function with `std::string` always adds an extra copy.

Can you please assign (+ping if already assigned) issues you want me to fix. Thanks

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -187,6 +187,7 @@ you up to date with the multi-language support provided by Elektra.
 
 - Warnings about cmake policies are avoided. _(Markus Raab)_
 - We removed the Haskell and GI bindings. _(Markus Raab)_
+- Avoid unnecessary copying std::string where possible (setString and setMeta only). _(Manuel Mausz)_
 
 ### Java
 
--- a/src/bindings/cpp/include/key.hpp
+++ b/src/bindings/cpp/include/key.hpp
@@ -175,6 +175,7 @@ public:
 	inline void set (T x);
 
 	inline std::string getString () const;
+	inline void setString (const char * newString);
 	inline void setString (const std::string & newString);
 	inline ssize_t getStringSize () const;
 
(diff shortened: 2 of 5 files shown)


markus2330:
So good to see you active again :sparkling_heart: Thank you very much for this PR!

> Can you please assign (+ping if already assigned) issues you want me to fix. Thanks

Assignments are done. I definitely do not expect the low-priority issues to be done. Simply see if they are easy to do and worth the effort.

markus2330:
Please add release notes as instructed by the build server. In future it is good if you leave the text in pull request intact, so that you see what you need to do.

manuelm:
I deliberately didn't append a release note. Do you want the first commit included? + Are you happy with the 2nd approach? Or should I add the `kdb::Key::setString(char *)` in the bindings only (included coping the body of `original setString`).

markus2330:
> I deliberately didn't append a release note.

Please always add, so that the build server will do its checks.

> Do you want the first commit included? + Are you happy with the 2nd approach?

348cae5f40ec7288622df55b0c56720e24e82e6a is not recommended anymore and should not be needed with C++11. Maybe simply adding the C++11 compiler-option is enough (or even brings other benefits).

Does any commit change the way the API is to be used? Can you give an example how it changes?

Otherwise, if it is only about performance, the faster method is preferred.

manuelm:
> [348cae5](https://github.com/ElektraInitiative/libelektra/commit/348cae5f40ec7288622df55b0c56720e24e82e6a) is not recommended anymore and should not be needed with C++11. Maybe simply adding the C++11 compiler-option is enough (or even brings other benefits).

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

markus2330:
I refer to http://cpp-next.com/archive/2009/08/want-speed-pass-by-value/ (or if it is down, did not load here when I tried it see https://stackoverflow.com/questions/21605579/how-true-is-want-speed-pass-by-value)

But in the end it does not matter: you simply do what your benchmark says is best.

manuelm:
Good to go then. No API changes.

markus2330:

markus2330:

markus2330:

markus2330:
Thank you, great job! I am looking forward to see the benchmarks on this.

manuelm:
Do I have to do is? 😓😭 It's removing one of the alloc spikes

markus2330:
No, you can also simply say that this case was optimized later.

PR #3133, opened 2019-10-30
Python enhancements

### LINKED ISSUE #2207: various python enhancements
Some small nice to have for Python:

## k==None comparison after lookup does not work.

 ```
	ks = kdb.KeySet()
	ks.append (kdb.Key("user/something"))
	k = ks.lookup("user/something")
	if k == None:
		pass
```

expected outcome: k is not None, so "pass" should not be executed but the program should continue afterwards (it works if k != None)

actual outcome: exception ValueError: invalid null reference in method 'Key___eq__', argument 2 of type 'kdb::Key const &'

workaround: use `key is None` which is arguably the correct thing to do anyway

##  KeySet not printed

expected outcome: print key by key, similar to built-in data structures in Python
actual outcome:
```
<kdb.KeySet; proxy of <Swig Object of type 'kdb::KeySet *' at 0x7f05b3eb4db0> >
```

workaround: write loop `for k in ks: print k,`

## Add key objects to hashmap

before c88d0b09414916f80c3eccdb376ad6a87805d867 it was not possible. Now it is possible but it might be buggy because users might change the name after adding the key to a hashmap. To avoid this, the readonly flag should be set. Maybe we even need a FrozenKey?

workaround: use the KeySet, which soon is a hashmap on lookups anyway

## No len(key)

```
>>> import kdb
>>> key = kdb.Key("user/hello/sub")
>>> len(key)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object of type 'Key' has no len()
```

Expected behavior: return 3

workaround:
```
sum (1 for _ in key)
```

### LINKED ISSUE #2584: python: provide more high-level APIs
I have some common helpers which are often needed and should be directly added to our binding (so that they are always available, well tested and well documented)

```python
def make_set (ks, where):
	s = set ()
	for k in ks:
		if k.isDirectBelow (where):
			s.add (k)
	return s;

def unpack_names (ks):
	s = set ()
	for k in ks:
		s.add (k.name)
	return s

def unpack_basenames (ks):
	s = set ()
	for k in ks:
		s.add (k.basename)
	return s

def is_array_entry (k, arrayname):
	if k.basename[0] != '#':
		return False
	x = copy.copy(k)
	x.delBaseName()
	if x.basename == arrayname:
		return True

def append (ks, name, p):
	n = kdb.Key(name)
	n.string = str(p)
	ks.append (n)
	print n

# this is only a workaround, ideally this would be directly in kdb.KeySet:
class GetKeySet(kdb.KeySet):
	def get(self, n):
		k = ks.lookup(n)
		if not k:
			raise KeyError(n+" not found")
		return k.string
	def set(self, n, s):
		k = ks.lookup(n)
		if not k: # TODO: call append here
			raise KeyError(n+" not found")
		k.string = str(s)
```

manuelm:
Fixes #2207 
Fixes #2584


manuelm:
jenkins build error is unrelated

mpranj:
Dbus tests sometimes fail #2310 and you should add a line in the release notes, otherwise most tests are not even run.

manuelm:
What do you mean by "add a line in the release notes"? There already is https://github.com/ElektraInitiative/libelektra/pull/3133/commits/fc9e1c3972b7d64a4bbd057647e20599f7964886 (I just fixed the typos). And why do tests depend on adding release notes?

mpranj:
> What do you mean by "add a line in the release notes"? 

You added a line, but not with the correct formatting. It is non-optional to add the author name at the end of an entry. See the output from the [build server](https://build.libelektra.org/jenkins/blue/organizations/jenkins/libelektra/detail/PR-3133/1/pipeline/). Here's an excerpt:

```
+ scripts/build/run_check_release_notes

The release notes were not extended correctly (with 2 lines changed).

Please make sure you add at least one line describing your contribution followed

by the text _(your name)_.


For example, if Sandra Power fixed a typo, she adds to doc/news/_preparation_next_release.md:

- typo fixed _(Sandra Power)_
```

> And why do tests depend on adding release notes?

PR's are not merged if there are no news in the release notes. @markus2330 wants each contribution to appear in the notes. Some notes are then condensed shorly before the release, if the notes are too verbose/long, but every author should be mentioned. Therefore it makes no sense to let the build servers execute all tests, if the change will never be merged.

manuelm:
Ah forgot about my name. Will be fixed in a second. I'll also add other features to this PR to avoid merge conflicts.

markus2330:
Thank you for the PR and thanks to the others for giving the infos.

> Ah forgot about my name. Will be fixed in a second. I'll also add other features to this PR to avoid merge conflicts.

You can also create several PRs (if these are unrelated things).

manuelm:
@markus2330 please take a look. I'll add `doc/tutorials/python-kdb.md` after you are happy.

src/bindings/swig/python2/tests/testpy2_key.py
```diff
@@ -155,5 +167,15 @@ def test_iterator(self):
 		self.assertEqual(iter(k).value(),     "user")
 		self.assertEqual(reversed(k).value(), "c")
 
+	def test_helpers(self):
+            k = kdb.Key("user/noarray")
+            self.assertEqual(k.array_elements().index,    False)
+            self.assertEqual(k.array_elements().name,     None)
+            self.assertEqual(k.array_elements().basename, None)
+            k = kdb.Key("user/some/array/#2")
+            self.assertEqual(k.array_elements().index,    2)
```
markus2330: Why is array_elements in plural? It is the second array element? Maybe simply `.array()`?

src/bindings/swig/python2/tests/testpy2_key.py
```diff
@@ -155,5 +167,15 @@ def test_iterator(self):
 		self.assertEqual(iter(k).value(),     "user")
 		self.assertEqual(reversed(k).value(), "c")
 
+	def test_helpers(self):
+            k = kdb.Key("user/noarray")
+            self.assertEqual(k.array_elements().index,    False)
```
markus2330: So .index is the way to ask if it is an array? What about providing `array_element().is` or similar for a True/False? Then this here could return None.

src/bindings/swig/python/tests/test_keyset.py
```diff
@@ -113,11 +120,18 @@ def test_python_copy(self):
...
 		self.assertEqual(ks[0].getKey(), self.ks[0].getKey())
+		self.assertTrue(ks == self.ks)
 
 		# deep copy
 		ks = copy.deepcopy(self.ks)
 		self.assertEqual(len(ks), len(self.ks))
 		self.assertNotEqual(ks[0].getKey(), self.ks[0].getKey())
+		self.assertTrue(ks == self.ks)
+
+	def test_helpers(self):
```
markus2330: Duplicate of test cases? (See below)

manuelm: What do you mean? The other file is python2

markus2330:
@ulrikeschaefer do you also want to take a look?

src/bindings/swig/python2/tests/testpy2_key.py
manuelm: Because it returns the elements which makes the key an array entry. But in general I don't care. I don't use it. I'm just more of a descriptive guy who likes self explaining function names. `.array()` sounds like its converting the key to an array.

src/bindings/swig/python/tests/test_keyset.py
markus2330: Ahh ok, I did not expect that you will update python2 as we discussed in the other issue to remove it.

src/bindings/swig/python2/tests/testpy2_key.py
manuelm: I can add it but it feels unnecessary. You can check if it's an array style by checking any member of the namedtuple. I can also set `index` to None in the non-array-style case right away.

src/bindings/swig/python2/tests/testpy2_key.py
markus2330: Then lets leave it as is for now.

src/bindings/swig/python2/tests/testpy2_key.py
markus2330: > You can check if it's an array style by checking any member of the namedtuple.

Such statements are important for the tutorial so that it is clear how to use it.

> I can also set index to None in the non-array-style case right away.

Would be more consistent, wouldn't it?

manuelm:

manuelm:
and stop failing!

src/bindings/swig/python2/tests/testpy2_key.py
manuelm: I could also throw an exception. Not sure if that's more handy though.

manuelm: ValueError it is

manuelm:
@markus2330 please look at the latest commit. I'm sure you don't like it :-)

src/bindings/cpp/include/key.hpp
```diff
@@ -238,6 +238,18 @@ class Key
 	ckdb::Key * key; ///< holds an elektra key
 };
 
+/**
+ * Key class returned from keysets.
+ * In particular the key name has been locked.
+ */
+class FrozenKey
```
markus2330: Which advantage does the C++ user have from this? Isn't it enough if this is a python-only thing?

manuelm: > Isn't it enough if this is a python-only thing?

That's not possible unless I custom wrap *all* keyset getters. It's slow and not maintainable. It also defeats the purpose of SWIG. If you want to make the key hashable that's the only way I'm seeing right now.

> Which advantage does the C++ user have from this?

At least they know that they can't change the key name :-)

markus2330: Maybe we should even disable the setters for the name in the FrozenKey (add them doing nothing as private).

manuelm: Sure. Go ahead.

I just had another idea. I could do it python-only if the user also wants to do some more work. Not sure it will work though. In particular the user has to convert all keys into a `FrozenKey` manually. And the FrozenKey constructor will throw if the Key isn't name locked.

So something like this:
```
keyset = kdb.KeySet(kdb.Key("user/foo"), kdb.Key("user/bar"))
key = keyset[0] # still Key
# hash(key) is None
frozenkey = FrozenKey(key) # will throw if not named lock
# hash(frozenkey) is not None

Comment (markus2330) [reply src/bindings/cpp/include/key.hpp]: > And the FrozenKey constructor will throw if the Key isn't name locked.

Why not lock the key there?

@PhilippGackstatter can you give some input here? (as you recently made something similar in Rust)

Comment (manuelm) [reply src/bindings/cpp/include/key.hpp]: I'll leave this for today. In general I think the current approach is better in terms of usability as in "It just works". However since the *only* usecase is to make the key hashable, manual conversation might be the better approach. *Unless* the FrozenKey approach won't be extended. Because it gets messy really fast. Like `kdb.KeySet(1, FrozenKey(kdb.Key("user/foo")))[0]` is a kdb.Key again.

Comment (markus2330) [reply src/bindings/cpp/include/key.hpp]: Wouldn't it be possible in python to dynamically use the correct class depending on if the lock bit is set?

Comment (manuelm) [reply src/bindings/cpp/include/key.hpp]: Yes, but it means you need to wrap *every* c++ function which returns a key. And that's  slow + huge mainainence effort. Don't like the Idea at all.

Comment (markus2330) [reply src/bindings/cpp/include/key.hpp]: What about making FrozenKey only a typedef in C++? (Could be in an `#ifdef` that SWIG sees something else.)

Comment (manuelm) [reply src/bindings/cpp/include/key.hpp]: Thought about this myself. Doable but also horrible. 

I just found: https://stackoverflow.com/questions/28517290/python-classes-with-hash-support-depending-on-the-instance

What about doing the same: Throwing in `__hash__` and move on. L

Comment (manuelm) [reply src/bindings/cpp/include/key.hpp]: Thought about this myself. Doable but also horrible. 

I just found: https://stackoverflow.com/questions/28517290/python-classes-with-hash-support-depending-on-the-instance

What about doing the same: Throwing in `__hash__` and move on. L

Comment (markus2330) [reply src/bindings/cpp/include/key.hpp]: Yes, sounds like a simple and straightforward solution. (If not locked, throw in `__hash__`). Together with some way to lock the keys (at least at construction) should be everything we want about `__hash__`.

The important part here is anyway that this is properly described how to use it in the tutorial.

Comment (PhilippGackstatter) [reply src/bindings/cpp/include/key.hpp]: > @PhilippGackstatter can you give some input here? (as you recently made something similar in Rust)

In the Rust bindings, calling `set_name` on a key in a KeySet returns an error. 

There's also a `ReadOnly` wrapper that returns a key that only implements `get_`-methods. For example, the Rust equivalent of `keyGetMeta` returns such a `ReadOnly` key. With that it is a compile time error to call `set_name` on that key.

However, this is probably not the right approach for Python.

**Comment by manuelm** [comment]
@markus2330 please take a quick look at the latest commit. Is that sufficient for your needs?

**Comment by manuelm** [comment]
btw locking the key value in the constructor doesn't work. Because in https://github.com/ElektraInitiative/libelektra/blob/master/src/libs/elektra/keyhelpers.c the value gets set *after* the lock flag has been set (Line 303 vs 349).

**Comment by markus2330** [comment]
> @markus2330 please take a quick look at 20212c5. Is that sufficient for your needs?

I do not like the fiddling around in private data structures. Better less functionality and not doing this.

> locking the key value in the constructor doesn't work. Because in https://github.com/ElektraInitiative/libelektra/blob/master/src/libs/elektra/keyhelpers.c the value gets set after the lock flag has been set (Line 303 vs 349).

As it is offtopic here, I opened a new issue #3151

**Comment by markus2330** [comment]

**Comment by markus2330** [comment]

**Comment by markus2330** [comment]

File: src/bindings/swig/python/kdb.i [hunk 5]
```diff
@@ -289,6 +325,28 @@
         key = self._lookup(item)
         return True if key else False
       raise TypeError("Invalid argument type")
+
+    def __str__(self):
+      """print the keyset in array style"""
+      return str(list(map(lambda k: str(k), self)))
+
+    def __repr__(self):
+      items = list(map(lambda k: repr(k), self))
```
markus2330: Some docu would be helpful here and in the helpers.

markus2330:
Looks good to me. @mpranj do you want to add this for the release?

I think the build job is failing because we already increased the version number.

@manuelm it would be great to have an extended tutorial how to use these features, @ulrikeschaefer would appreciate it.

mpranj:
Sure, let's add it to 0.9.1!

markus2330:
Feel free to merge :wink:

mpranj:
Looks nice, thank you for the improvements!

PR #3142, opened 2019-10-31
Improvements for exposed meta keyset

PhilippGackstatter:
Improvements for #2983

## Basics

      contains `_(my name)_`)
- [x] The buildservers are happy.

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/src/libs/elektra/key.c
+++ b/src/libs/elektra/key.c
@@ -73,22 +73,6 @@
  * ordering or different Models of your configuration.
  */
 
-/*
- * @internal
- *
- * Allocates and initializes a key
- * @returns 0 if allocation did not work, the key otherwise
- */
-static Key * elektraKeyMalloc (void)
-{
-	Key * key = (Key *) elektraMalloc (sizeof (Key));
-	if (!key) return 0;
-	key->meta = NULL;
-	keyInit (key);
-
-	return key;
-}
-
 /**
  * A practical way to fully create a Key object in one step.
  *
@@ -187,7 +171,7 @@ Key * keyNew (const char * name, ...)
 
 	if (!name)
 	{
-		k = elektraKeyMalloc ();
+		k = elektraCalloc (sizeof (Key));
 	}
 	else
 	{
@@ -209,7 +193,7 @@ Key * keyNew (const char * name, ...)
  */
 Key * keyVNew (const char * name, va_list va)
 {
-	Key * key = elektraKeyMalloc ();
+	Key * key = elektraCalloc (sizeof (Key));
 
 	keyVInit (key, name, va);
 	return key;
@@ -269,9 +253,6 @@ Key * keyDup (const Key * source)
 
 	dest = keyNew (0, KEY_END);
 	if (!dest) return 0;
-
-	// Will be overwritten by keyCopy
-	ksDel (dest->meta);
 
 	/* Copy the struct data */
 	*dest = *source;
@@ -521,7 +502,6 @@ int keyClear (Key * key)
 
 	ksDel (key->meta);
 
-	key->meta = NULL;
 	keyInit (key);
 
 	if (keyStructInMmap) key->flags |= KEY_FLAG_MMAP_STRUCT;
--- a/src/libs/elektra/key.c
+++ b/src/libs/elektra/key.c
@@ -452,7 +452,7 @@ int keyDel (Key * key)
 
 	rc = keyClear (key);
 
-	if (key->meta) ksDel (key->meta);
+	ksDel (key->meta);
 
 	if (!keyInMmap)
 	{
@@ -508,7 +508,6 @@ int keyClear (Key * key)
 	if (key->data.v && !test_bit (key->flags, KEY_FLAG_MMAP_DATA)) elektraFree (key->data.v);
 
 	ksDel (key->meta);
-	key->meta = NULL;
 
 	keyInit (key);
 
--- a/src/libs/elektra/key.c
+++ b/src/libs/elektra/key.c
@@ -73,6 +73,21 @@
  * ordering or different Models of your configuration.
  */
 
+/*
+ * @internal
+ *
+ * Allocates and initializes a key
+ * @returns 0 if allocation did not work, the key otherwise
+ */
+static Key * elektraKeyMalloc (void)
+{
+	Key * key = (Key *) elektraMalloc (sizeof (Key));
+	if (!key) return 0;
+	key->meta = NULL;
+	keyInit (key);
+
+	return key;
+}
 
 /**
  * A practical way to fully create a Key object in one step.
@@ -172,9 +187,7 @@ Key * keyNew (const char * name, ...)
 
 	if (!name)
 	{
-		k = (Key *) elektraMalloc (sizeof (Key));
-		if (!k) return 0;
-		keyInit (k);
+		k = elektraKeyMalloc ();
 	}
 	else
 	{
@@ -196,8 +209,7 @@ Key * keyNew (const char * name, ...)
  */
 Key * keyVNew (const char * name, va_list va)
 {
-	Key * key = (Key *) elektraMalloc (sizeof (Key));
-	if (!key) return 0;
+	Key * key = elektraKeyMalloc ();
 
 	keyVInit (key, name, va);
 	return key;
@@ -509,6 +521,7 @@ int keyClear (Key * key)
 
 	ksDel (key->meta);
 
+	key->meta = NULL;
 	keyInit (key);
 
 	if (keyStructInMmap) key->flags |= KEY_FLAG_MMAP_STRUCT;
--- a/src/libs/elektra/key.c
+++ b/src/libs/elektra/key.c
@@ -74,23 +74,6 @@
  */
 
 
-/*
- * @internal
- *
- * Allocates and initializes a key
- * @returns 0 if allocation did not work, the key otherwise
- */
-static Key * elektraKeyMalloc (void)
-{
-	Key * key = (Key *) elektraMalloc (sizeof (Key));
-	if (!key) return 0;
-	key->meta = NULL;
-	keyInit (key);
-
-	return key;
-}
-
-
 /**
  * A practical way to fully create a Key object in one step.
  *
@@ -189,7 +172,9 @@ Key * keyNew (const char * name, ...)
 
 	if (!name)
 	{
-		k = elektraKeyMalloc ();
+		k = (Key *) elektraMalloc (sizeof (Key));
+		if (!k) return 0;
+		keyInit (k);
 	}
 	else
 	{
@@ -211,8 +196,9 @@ Key * keyNew (const char * name, ...)
  */
 Key * keyVNew (const char * name, va_list va)
 {
-	Key * key = elektraKeyMalloc ();
+	Key * key = (Key *) elektraMalloc (sizeof (Key));
 	if (!key) return 0;
+
 	keyVInit (key, name, va);
 	return key;
 }
@@ -269,7 +255,7 @@ Key * keyDup (const Key * source)
 
 	if (!source) return 0;
 
-	dest = elektraKeyMalloc ();
+	dest = keyNew (0, KEY_END);
 	if (!dest) return 0;
 
 	// Will be overwritten by keyCopy
--- a/src/libs/elektra/keyhelpers.c
+++ b/src/libs/elektra/keyhelpers.c
@@ -197,10 +197,9 @@ int keyNameIsSystem (const char * name)
  */
 int keyInit (Key * key)
 {
-	if (key->meta) ksClear (key->meta);
-	memset (key, 0, sizeof (struct _Key) - sizeof (KeySet *));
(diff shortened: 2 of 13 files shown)


markus2330:

src/libs/elektra/keyhelpers.c
```diff
@@ -197,10 +197,13 @@ int keyNameIsSystem (const char * name)
...
 int keyInit (Key * key)
 {
-	if (key->meta) ksClear (key->meta);
-	memset (key, 0, sizeof (struct _Key) - sizeof (KeySet *));
+	KeySet * ks = key->meta;
+	if (ks) ksClear (ks);
 
-	if (!key->meta) key->meta = ksNew (0, KS_END);
+	memset (key, 0, sizeof (struct _Key));
+	key->meta = ks;
```
PhilippGackstatter: I had a simpler solution before, that did not require the `NULL` precondition for `key->meta`. It works fine for the abi, but the `mmapstorage` plugin test will not work with that change and I could not find the source of the error in reasonable time, so I rolled some of those changes back.

PhilippGackstatter:
You also asked for tests for the `alloc == 0` change. But since that is not visible on the public interface, how would I test that?

src/libs/elektra/keyhelpers.c
mpranj: If there's a problem with `mmapstorage` please report it, maybe I can help or take a look.

I don't get why `keyInit` requires changes here.

I think that use of an [uninitialized stack variable](https://cwe.mitre.org/data/definitions/457.html) is introduced here.  `ksLookupByName` makes a call to `keyInit` by supplying an uninitialized Key struct on the stack. To my knowledge this means that some garbage from memory will be passed to `keyInit` and lead to calling `ksClear` on an invalid pointer.

PhilippGackstatter: This is a continuation of #3081 where I already made changes to `keyInit` to initialize the `key->meta` on key creation. But the implementation wasn't portable and not ideal so I changed it again in this PR.

In #3081 I changed `ksLookupByName` to initialize the meta variable to null, so that shouldn't be an issue.

If you could take a look that would be great. The breaking change was introduced in 
1130515. Maybe check that out and run `test_mmapstorage`. You can also see it [here on Travis](https://travis-ci.org/ElektraInitiative/libelektra/jobs/605429496?utm_medium=notification&utm_source=github_status).

mpranj: Thank you, I will take a look!

src/libs/elektra/key.c
```diff
@@ -73,6 +73,21 @@
...
 
+/*
+ * @internal
+ *
+ * Allocates and initializes a key
+ * @returns 0 if allocation did not work, the key otherwise
+ */
+static Key * elektraKeyMalloc (void)
+{
+	Key * key = (Key *) elektraMalloc (sizeof (Key));
```
mpranj: This function could be a one-liner: `return elektraCalloc (sizeof (Key));`.
This will initialize all members to 0, making it a bit more flexible for future changes (e.g. adding another pointer member). Default values different from 0 of course have to be initialized correctly, but it's less common.

There's also no need to cast `void *` to other pointer types, I think.

src/libs/elektra/key.c
```diff
@@ -172,9 +187,7 @@ Key * keyNew (const char * name, ...)
 
 	if (!name)
 	{
-		k = (Key *) elektraMalloc (sizeof (Key));
-		if (!k) return 0;
-		keyInit (k);
+		k = elektraKeyMalloc ();
```
mpranj: Nice simplification!

src/libs/elektra/keyhelpers.c
```diff
@@ -197,10 +197,13 @@ int keyNameIsSystem (const char * name)
...
+
+	if (!ks) key->meta = ksNew (0, KS_END);
```
mpranj: Can't we lazily alloc/initialize the meta keyset if needed? I think allocating the data structures for all keys is not needed. This additionally forces mmapstorage/cache to write all the empty keysets to disk, or allocate them while loading the cache. Maybe there are some other opinions?

PhilippGackstatter: This is the discussion we had in #2983. See in particular https://github.com/ElektraInitiative/libelektra/issues/2983#issuecomment-534575198

The KeySet array is no longer allocated if called by `ksNew(0, KS_END)`. That is how we initialize `key->meta`, so that it isn't wasting so much memory.

We decided to always allocate `key->meta` because otherwise `keyMeta` would have been a much more complex function.

mpranj: What about:
```
KeySet * keyMeta (Key * key)
{
	if (!key) return 0;
	if (!key->meta) key->meta = ksNew (0, KS_END);

	return key->meta;
}
```
Is that too complex?

src/libs/elektra/keyhelpers.c
mpranj: I took a look at it and the problem with `mmapstorage` is caused by the allocation of the empty meta keyset. 

I attached a patch with changes that make 11305155ac3efc58357741161d827c60af280b4a work with mmapstorage, lazily create the meta keyset and use elektraCalloc as suggested in my previous comments. (The .patch is renamed to .txt because github doesnt like .patch files)
[keyMeta-lazy.txt](https://github.com/ElektraInitiative/libelektra/files/3800466/keyMeta-lazy.txt)

src/libs/elektra/keyhelpers.c
PhilippGackstatter: This is also what I suggested https://github.com/ElektraInitiative/libelektra/issues/2983#issuecomment-534109621

@markus2330 wrote that with that you cannot check whether metadata exists at the moment. But thinking about it now, with the current solution there simply always exists metadata so you don't need that check and with your solution, too, just lazily. The main concern here was memory, but this solution is better memory-wise than to always create it, so it seems like an improvement to me.

src/libs/elektra/key.c
PhilippGackstatter: This is just rolling back my own changes :smile:

src/libs/elektra/keyhelpers.c
PhilippGackstatter: Thank you, works great!

src/libs/elektra/keyhelpers.c
mpranj: I'd prefer that approach, as it is transparent to the user anyway. That would allow `mmapstorage` to simply not store empty meta keysets. If the approach is eager then `mmapstorage` would need to persist this "small overhead per key" also to disk, which does not seem practical to me.

If there is a different reasoning which I am overlooking here, then of course my suggestion might not apply.

markus2330: Yes, I am also fine that we are lazy at two points in time:

1. key->meta might be empty if keyMeta was never called
2. key->meta->array might be empty if not meta keys were added

The implementation of @mpranj is fine. My concern is that the user should not get 0 if key->meta is null (as this complicates the use of this API, a second function or parameter is needed). But to create it lazily within keyMeta is fine. But I think this optimization (1.) will not have much effect as keyMeta will always be called at some point.

So instead of (1.) it might more sense to detect within mmap code that key->meta->array is not there, thus (in this case) the whole key->meta does not need to be persisted.

@mpranj what do you think about that?

mpranj: We can detect that inside mmapstorage, yes. But then we persist a nullpointer instead of the meta keyset, so we again need to lazily allocate it later. (Or eagerly allocate it inside mmap, which contradicts the mmap approach of trying to do as little as possible)

To me it sounds like a good solution to:
1. Filter empty meta keysets inside mmap.
2. Lazily allocate `key->meta` in the `keyMeta` function, s.t. mmap does not have to do it eagerly.

markus2330: Yes, sounds okay. In any case we can add both lazy allocations.

markus2330:
> You also asked for tests for the alloc == 0 change. But since that is not visible on the public interface, how would I test that?

You can also add tests that have access to the private interface (tests/ctest/test_ks.c)

kodebach:
https://github.com/ElektraInitiative/libelektra/blob/6aea503521a35cbeef3a719f8c73c2fe40830c1d/src/libs/elektra/symbols.map#L89

Please move the `keyMeta` symbol to a new version group called `libelektra_1.0` like this:

```
libelektra_0.8 {
...
};

libelektra_1.0 {
	keyMeta;
};

libelektraprivate_1.0 {
...
};
```

Each symbol should be tagged with the version in which it was introduced or its ABI was last changed.

markus2330:
@kodebach do you think it is better if we declare all symbols to be 1.0 instead of saying in which 0.9 version it was introduced? My plan was to release as many 0.9 versions (as release candidates) until we are happy and declare one of the 0.9 versions without (or nearly without) changes as 1.0). But I am open for discussion to change this.

PhilippGackstatter:
> @kodebach do you think it is better if we declare all symbols to be 1.0 instead of saying in which 0.9 version it was introduced?

@kodebach can you answer that question? Thanks!

kodebach:
Sorry, I missed that... Yes, it is probably better declare everything as 1.0 since we agreed that we accept any and all breaking changes until the release of 1.0.

It might even make sense to merge all existing versions into a single one before the release of 1.0. Since there are major changes (e.g. #3115) users need to change their code before switching to 1.0 anyway. And linking older applications against libelektra 1.0 won't work, not even with symbol versioning.

markus2330:
Yes, I agree.

src/libs/elektra/keyhelpers.c
```diff
@@ -195,13 +195,9 @@ int keyNameIsSystem (const char * name)
  * clears key (all data members are set to zero)
  * Initializes an empty metadata keyset if null or clears it.
  */
-int keyInit (Key * key)
+void keyInit (Key * key)
 {
-	if (key->meta) ksClear (key->meta);
```
mpranj: If it is possible that someone calls `keyInit` with a key including metadata, maybe we should `ksDel(key->meta)`.
Otherwise maybe just update the documentation.

src/libs/elektra/keyhelpers.c
```diff
@@ -195,13 +195,9 @@ int keyNameIsSystem (const char * name)
...
-	memset (key, 0, sizeof (struct _Key) - sizeof (KeySet *));
-
-	if (!key->meta) key->meta = ksNew (0, KS_END);
-	return 0;
+	memset (key, 0, sizeof (struct _Key));
```
mpranj:
```suggestion
	memset (key, 0, sizeof (Key));
```

src/libs/elektra/keyhelpers.c
PhilippGackstatter: As far as I know this is not the case anywhere in `elektra-core`, where `keyInit` is only called on key creation or similar. Not sure about bindings.

If we have to `ksDel(key->meta)`, then we again have to set `key->meta = NULL` before all  `keyInit` calls, otherwise we pass a garbage pointer to `ksDel`.

mpranj: > otherwise we pass a garbage pointer to ksDel.

Not if they are allocated using calloc.

mpranj:
Is there an ETA on this PR? I'm hoping it fixes #3158 and it would be **great** for the benchmarks @raphi011 and I are currently doing. If it will take longer maybe we can cherry pick/merge the meta keyset allocation improvements sooner?

PhilippGackstatter:
> Is there an ETA on this PR? I'm hoping it fixes #3158 and it would be **great** for the benchmarks @raphi011 and I are currently doing. If it will take longer maybe we can cherry pick/merge the meta keyset allocation improvements sooner?

It is done, I'm just trying to make the builds succeeds to make sure I didn't break anything else.

raphi011:
I will create a branch tonight based on your PR and see if it fixes the memory leak. Thanks for pointing me here @mpranj

src/libs/elektra/keyhelpers.c
kodebach: `keyInit` is not listed in the relevant `symbols.map` and therefore is not accessible outside of `elektra-core`. Inside of `elektra-core` it is only called on uninitialised keys. This is explains the modifications to `keyInit` I made in #3115. I'm not sure you can do the same without some of the other changes of #3115, but in any case it doesn't really matter how you solve it, since #3115 will change it anyway.

PhilippGackstatter:

mpranj:

markus2330:

mpranj:

markus2330:

mpranj:

PhilippGackstatter:

markus2330:

markus2330:
I added the label urgent as this might fix a nasty memory leak, see #3158

PhilippGackstatter:
It went through :+1:

markus2330:
Thank you very much!

mpranj:
Thank you!

PR #3154, opened 2019-11-03
Cirrus: Add Job to Check Man Pages And PDF Documentation

sanssecours:


mpranj:
Thank you! To my knowledge there are already outdated man pages, so it's great to have this.

markus2330:
Thank you so much for the PR!

sanssecours:
> To my knowledge there are already outdated man pages, so it's great to have this.

Thank you for the hint. I noticed that `📚 Check` does currently not generate the man pages:

- `kdb-elektrify-getenv.1`,
- `kdb-qt-gui.1`,
- `kdb-run-elektrad.1`,
- `kdb-run-web.1`,
- `kdb-run-website-backend.1`, and
- `kdb-run-website-frontend.1`.

Switching to a distro that uses the GNU C library and installing more software in the Cirrus Dockerfile should hopefully get rid of that problem.

markus2330:
Thank you great job!

Is there a specific reason why you used Cirrus and not Jenkins here?

sanssecours:
> Is there a specific reason why you used Cirrus and not Jenkins here?

There are a lot of reasons. I list some of them, more or less in the order of their importance, below.

```xml
<rant>
```

- Cirrus is more **stable**
   - No **disk quota exceeded**
   - No need to worry about the amount of **free space** on a node 
   - The **uptime** of the Cirrus servers seems higher
- No **unnecessary dependencies** between build jobs. I do not have to care if ten build jobs finish successfully before the build job I am interested even starts.
- No need to **restart all build jobs**, just because there was a tiny problem in one of them.
- Cirrus also **works with my copy of the repository**. As consequence I do not have to open a pull request to check if the tests finish successfully.
- **No need to worry about administrating** build hardware or software. If there is no dedicated person to take care of the Jenkins build hardware and software, the quality of the continuous integration will suffer. Right now the Jenkins build will almost always fail. This was not the case right after Lukas updated the build system. Of course the situation was also bad, when Lukas was working on upgrading the Jenkins build.
- Cirrus uses **“standard” Dockerfiles**. No need to worry about a Jenkins user.
- I **prefer writing Shell** code more or less directly in a YAML file, instead of writing Groovy code that executes shell code.

```xml
</rant>
```

markus2330:
Thank you for the rant!

>  No unnecessary dependencies between build jobs. I do not have to care if ten build jobs finish successfully before the build job I am interested even starts.

This was done more or/less on purpose so that obviously failing PRs do not waste too much resources. But I see it is annoying for senior devs. Any idea how to improve there?

> Cirrus uses “standard” Dockerfiles. No need to worry about a Jenkins user.

Would be interesting how they did it?

With most points I agree, I hope it improves once @Mistreated familiarized himself with the system.

sanssecours:
> This was done more or/less on purpose so that obviously failing PRs do not waste too much resources.

I know, but ironically we now have a situation, where we waste even more resources. Restarting all build jobs again, because of a  little problem just does not make that much sense.

> Any idea how to improve there?

I think it would be a good idea to split the Jenkins build into smaller pieces. Maybe something like one build job for every stage we have now. We could also introduce groups to collect similar build jobs, which might make it more obvious why a build failed. For example one Jenkins build (job) for the Web UI, one for the webpage, one for Debian builds, one for documentation…

markus2330:
> I think it would be a good idea to split the Jenkins build into smaller pieces.

We already had this and unfortunately it was unmaintainable.

sanssecours:
> We already had this and unfortunately it was unmaintainable.

I agree that the old Jenkins setup was kind of bad. I do think that using the approach of a few build jobs seems to work reasonably well for both Travis and Cirrus, though.

markus2330:
There are only few because they do not cover so much. Once the reach the complexity of what Jenkins does, it becomes messy. And splitting is also a lot of effort.

PR #3164, opened 2019-11-05
Whitelist antlr4 website

Chemin1:
The website responds with 404 on this link (at least there is some 404 in my network traffic when clicking it).

### DIFF
--- a/tests/linkchecker.whitelist
+++ b/tests/linkchecker.whitelist
@@ -17,3 +17,4 @@ http://standards.freedesktop.org/basedir-spec/basedir-spec-0.8.html
 http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
 https://freedesktop.org
 https://www.freedesktop.org/wiki/Software/pkg-config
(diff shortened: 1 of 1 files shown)


mpranj:
I'm not sure but it seems that something was restructured and maybe the project just moved to some other link. The purpose of the link checker is exactly to find and fix such problems, I don't think whitelisting is the best approach here.

@sanssecours can you check where this should point now? There exists a https://repology.org/project/antlr/versions project with all versions, but no explicit antlr4 site.

markus2330:
Yes, I agree. It looks like the project changed is repo, so we should update the link and not whitelist it.

Chemin1:
Makes sense.

I thought that the 404 made the link checker fail, but apparently this was just coincidence and re-running the link checker some times gives a successful check.

sanssecours:
I opened a [new pull request](https://github.com/ElektraInitiative/libelektra/pull/3166) that removes the link and adds install instructions for Arch Linux, Debian Linux and macOS instead.

Issue #2794, opened 2019-06-18
avoid strerror


markus2330:
From #2686:
> We should not recommend to call strerror(errno) as it is not thread-safe. strerror_r should be used instead. It would be better if we can embed the strerror in a generic way, so that not every caller needs to worry about that.

@Piankero wrote:
> Always allocating and freeing memory just for an error message seems very cumbersome.

Exactly, this is why our error system should take care of this. Then we only implement it once, and everyone can safely report errno.

> So why now introduce thread safety just for error messages?

Thread safety is always supported with Elektra. The current uses of strerror are simply a bug.

It is okay if your implementation uses strerror for now (if you find the other strerror APIs too difficult). But it is important that we have a single place where we can fix the problem.

markus2330:
Btw. this issue is already listed in #2572, so it is not new. And if you use mustache for the template, it should be very easy to fix strerror at the one place.

This issue is also related to #2697.

kodebach:
> Always allocating and freeing memory just for an error message seems very cumbersome.

I don't think that is needed. The following is safe to use and message longer than 2048 characters (if they even exist anywhere) wouldn't be read by the user anyway, so IMO its fine to truncate them. A smaller limit would probably, be fine as well.

```c
char buf[2048];
strerror_r (errno, buf, sizeof(buf));
```

Piankero:
Not related to visible error messages.
As discussed in the Elektra meeting I may unassign myself.

PR #3176, opened 2019-11-07
Error message pitfall remedies. Discussion PR

Piankero:
This PR is a basis for discussing possible enhancements to the error messages which I have seen throughout my thesis. Feel free to add comments to certain paragraphs.

Any concrete implementation makes no sense as of now until https://github.com/ElektraInitiative/libelektra/pull/2976 is merged since it restructures the error code significantly.

This PR is not intended to be merged so build failures are irrelevant

### DIFF
--- a/doc/dev/error-message-pitfalls-proposal.md
+++ b/doc/dev/error-message-pitfalls-proposal.md
@@ -27,6 +27,7 @@ these messages and afterwards show potential remedies which are open for discuss
    4. Leaking internals
 
       Some messages I have seen state which internal method has failed. E.g., in LCDproc there was a message `xosd: xosd_set_font() failed`. This though should be of no relevance for the user.
+      "Internal" means variable names and method names.
 
 2. Forgetting crucial error information
 
--- a/doc/dev/error-message-pitfalls-proposal.md
+++ b/doc/dev/error-message-pitfalls-proposal.md
@@ -1,0 +1,76 @@
+# Introduction
+
+After the huge effort to harmonize and standardize our error message & code system, there is a great likelihood
+that any newly added message will again violate these standards. The only remedy as of now are good PR reviews
+which again are prone to human errors. I want to present common pitfalls I have seen when correcting
+these messages and afterwards show potential remedies which are open for discussion.
+
+# Problems
+
+1. Inconsistent Error Messages
+
+   1. Too advanced vocabulary
+
+      Many error and warning messages included words which are too advanced for normal users.
+      Examples are stat, normalize, glob, uncommon acronyms, etc.
+
+   2. Forgetting single quotes around %s expressions
+
+      As written in the error message design guideline, string formatters with %s have to be surrounded by single quotes so users
+      know where they start and end as variables with spaces do not make it clear otherwise.
+
+   3. Inconsistent sentence pattern
+
+      The error message design guideline states that sentences have to start with an upper case letter and should not contain a dot at the end.
+      Also exclamation marks should not be used as it is interpreted as shouting to the user.
+
+   4. Leaking internals
+
+      Some messages I have seen state which internal method has failed. E.g., in LCDproc there was a message `xosd: xosd_set_font() failed`. This though should be of no relevance for the user.
+
+2. Forgetting crucial error information
+
+   'errno' usually contains a very important message for the user but might have been forgotten by the developer to add.
+
+3. Forgetting key or key value information in error messages
+
+   In many cases the key name (keyName) was not given by the error message writer. This though is needed for users to locate the invalid configuration setting value
+
+# Potential remedies
+
+## Problem 1
+
+These problems should be checked by the build server and let the build fail if something is not conforming. This would require a test
+that grabs all error messages and analyzes them for certain points. 1.2. and 1.3. are very easy to check by providing a regular
+expression that has to match. 
+
+Another way to address 1.2 and 1.3 beforehand are message templates which could be used by developers. Personally I think this is difficult
+to maintain and new developers will surely do not see and use it (right away). Elektra already has many things which developers should know
+when they want to participate and I think this should be kept as few as possible.
+For out of memory errors though I would suggest to rewrite the method signature to not provide any text as parameter but take a standard error text which contains the errno as this is clearly identical in all cases.
+
+1.1 is more difficult to address as it would require some sort of blacklist which gets expanded as reviews say which words are too advanced. This is though a larger project as every word has to be stemmed in the error message and then compared to a list with also stemmed
+words inside. Otherwise decent error message would trigger a build fail (e.g., "Could not generate statistics" would trigger a fail if the word "stat" was on the blacklist) or illegal words pass the build server (e.g., "Normalizing did not work" if "normalize" was on the blacklist).
+
+1.4 is probably the least important problem as this problem happened quite rarely. This would require a grep for all method names and see
+if it is not contained in the error message.
+
+The one problem I see here is how to get all error messages in the src folder.
+I have forged the following perl regex which could be used to get most of the error messages: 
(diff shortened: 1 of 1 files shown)


doc/dev/error-message-pitfalls-proposal.md
```diff
@@ -0,0 +1,76 @@
...
+these messages and afterwards show potential remedies which are open for discussion.
+
+# Problems
+
+1. Inconsistent Error Messages
+
+   1. Too advanced vocabulary
+
+      Many error and warning messages included words which are too advanced for normal users.
+      Examples are stat, normalize, glob, uncommon acronyms, etc.
```
markus2330: We could have something like the spell-correction for error messages.

doc/dev/error-message-pitfalls-proposal.md
```diff
@@ -0,0 +1,76 @@
...
+      The error message design guideline states that sentences have to start with an upper case letter and should not contain a dot at the end.
+      Also exclamation marks should not be used as it is interpreted as shouting to the user.
+
+   4. Leaking internals
+
+      Some messages I have seen state which internal method has failed. E.g., in LCDproc there was a message `xosd: xosd_set_font() failed`. This though should be of no relevance for the user.
+
+2. Forgetting crucial error information
+
+   'errno' usually contains a very important message for the user but might have been forgotten by the developer to add.
```
markus2330: Can you give a checklist for what is crucial here.

doc/dev/error-message-pitfalls-proposal.md
```diff
@@ -0,0 +1,76 @@
...
+      know where they start and end as variables with spaces do not make it clear otherwise.
+
+   3. Inconsistent sentence pattern
+
+      The error message design guideline states that sentences have to start with an upper case letter and should not contain a dot at the end.
+      Also exclamation marks should not be used as it is interpreted as shouting to the user.
+
+   4. Leaking internals
+
+      Some messages I have seen state which internal method has failed. E.g., in LCDproc there was a message `xosd: xosd_set_font() failed`. This though should be of no relevance for the user.
```
markus2330: Can you give a checklist for what is internal here.

doc/dev/error-message-pitfalls-proposal.md
```diff
@@ -0,0 +1,76 @@
...
+## Problem 1
+
+These problems should be checked by the build server and let the build fail if something is not conforming. This would require a test
+that grabs all error messages and analyzes them for certain points. 1.2. and 1.3. are very easy to check by providing a regular
+expression that has to match. 
+
+Another way to address 1.2 and 1.3 beforehand are message templates which could be used by developers. Personally I think this is difficult
+to maintain and new developers will surely do not see and use it (right away). Elektra already has many things which developers should know
+when they want to participate and I think this should be kept as few as possible.
+For out of memory errors though I would suggest to rewrite the method signature to not provide any text as parameter but take a standard error text which contains the errno as this is clearly identical in all cases.
```
markus2330: I agree.

doc/dev/error-message-pitfalls-proposal.md
Piankero: variable names and method names

doc/dev/error-message-pitfalls-proposal.md
Piankero: errno is crucial if it is set. Other information has to be provided by the developer and cannot be automatically added.

Piankero:
> I think the first part should be also added to our error guidelines

This is already documented in https://github.com/ElektraInitiative/libelektra/blob/b87e3d2b4a53a6d031f34d7429f98e6a8af6d9ed/doc/dev/error-message.md

doc/dev/error-message-pitfalls-proposal.md
markus2330: (please update in the document)

doc/dev/error-message-pitfalls-proposal.md
markus2330: And how can we ensure that errno is added when it is needed?

markus2330:
This guide here is in a style easier to read than what we already have in error-message.md

If we cannot add everything from here, at least more checklists could be added to error-message.md

doc/dev/error-message-pitfalls-proposal.md
Piankero: By checking if errno != 0

Piankero:
@markus2330, can you take a look at this document and check if my ideas are reasonable?

markus2330:
Where exactly would you like which kind of feedback? I already reviewed the PR.

markus2330:
Btw. if you want feedback about tasks you want to do, better to create issues for the tasks you plan.

Piankero:
> Where exactly would you like which kind of feedback?

It feels like you reviewed everything under `## Problems` but nothing about my ideas how to solve these issues.

Some sample questions:
Does it make sense to `grep` through all error messages and integrate a test which enforces certain properties? Should we add a generic message indicating an errno has been set? Does the vocabulary checker make sense to you? Or is this task probably too much effort for too little value? Etc.

markus2330:
> It feels like you reviewed everything under ## Problems but nothing about my ideas how to solve these issues.

PR is not well suited if you want answers to every point. PRs are more to give feedback about what should be changed, not to acknowledge which parts are ok.

> Does it make sense to grep through all error messages and integrate a test which enforces certain properties?

Yes, especially if it will also help yourself to bring all messages to one style.

That said: many things make sense. A better question, however, is what should have the highest priority to be done.

Imho, the templates are better suitable to enforce properties and you can also change the properties at one place. But start making a template for the out-of-memory and let us go from there.

> Should we add a generic message indicating an errno has been set?

I already suggested several times that errno should be handled using the framework, see also #2572. We could only add the number (to not run into  #2794, and render when the error message is rendered) and only show it with `-d`, to keep the default message short.

> Does the vocabulary checker make sense to you?

Yes, it makes sense but it has much less priority.

> Or is this task probably too much effort for too little value? Etc.

Yes, better drop the idea.

doc/dev/error-message-pitfalls-proposal.md
markus2330: Yes, errno could be added to the meta data if it is != 0.

Piankero:
> Imho, the templates are better suitable to enforce properties and you can also change the properties at one place. But start making a template for the out-of-memory and let us go from there.

Since OOM errors are basically done, my POC is done in this regards.
There are also other possibilities such as for semantic validation errors I found:
My idea would be a macro that requires the key (containing keyName [1] + wrong keyValue [2]) and a text indicating the reason [3]. The text would then probably look like 
`Key [1] with value [2] does not fulfill [3]`. I will add this to my thesis as well as my other ideas from this PR to make it deeper.

markus2330:
Why is it specific to semantic validation errors?

Without trying it out it is unclear if this will work.

For `[3]` we most likely need a formatted string.

Piankero:
> Why is it specific to semantic validation errors?

This is not specific to this category but I have already shown that the POC works for out of memory errors and shared my thoughts about validation semantic errors too because I mentioned it in another post.

> Without trying it out it is unclear if this will work.

Well I can theoretically do it for all categories but this is a big amount of work which is out of scope to finish my master thesis. You stated yourself I should just "start" it and see how it works, not finish it to 100%.

markus2330:
>  Well I can theoretically do it for all categories but this is a big amount of work which is out of scope to finish my master thesis. You stated yourself I should just "start" it and see how it works, not finish it to 100%.

But you have barely started it? That it will work when you simply don't put any information into the error message was obvious. You did not even find a way to add "size". But okay, it is not worth the effort for out-of-memory-errors, as there are much more frequent errors, like validation.

PR #3192, opened 2019-11-11
template: Added error template for out of memory errors

Piankero:
## Basics

      contains `_(my name)_`)

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -334,6 +334,7 @@ you up to date with the multi-language support provided by Elektra.
 - The reformatting scripts now run in parallel. _(Markus Raab)_
 - Improved various error messages and synchronized documentations. _(Michael Zronek)_
 - Improved `range` plugin error message. _(Michael Zronek)_
+- Introduced error message template for out of memory errors. _(Michael Zronek)_
 - Improved error codes documentation to clarify the hierarchy for developers. _(Michael Zronek)_
 - Release notes now use git's union merge driver. _(Dominic Jäger)_
 
--- a/src/include/kdbmacros.h
+++ b/src/include/kdbmacros.h
@@ -161,4 +161,11 @@
  */
 #define ELEKTRA_SYMVER_DECLARE(ver, sym, impl) ELEKTRA_SYMVER_COMMAND (ELEKTRA_STRINGIFY (ELEKTRA_SYMVER (sym, impl)), #sym "@" ver)
 
+/** Error message templates
+ *
+ * The following templates should be used whenever possible as they are ensured to be consistent with all other messages
+ * **/
+
+#define OUT_OF_MEMORY_MSG "Memory allocation failed"
+
 #endif
--- a/src/plugins/augeas/augeas.c
+++ b/src/plugins/augeas/augeas.c
@@ -415,7 +415,7 @@ static int saveTree (augeas * augeasHandle, KeySet * ks, const char * lensPath, 
 
 memoryerror:
 	elektraFree (keyArray);
-	ELEKTRA_SET_OUT_OF_MEMORY_ERROR (parentKey, "Unable to allocate memory while saving the augeas tree");
+	ELEKTRA_SET_OUT_OF_MEMORY_ERROR (parentKey, OUT_OF_MEMORY_MSG);
 	return -1;
 }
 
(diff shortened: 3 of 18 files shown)


markus2330:
Why didn't you directly modfiy ELEKTRA_SET_OUT_OF_MEMORY_ERROR? Is there any reason to have this macro?

Piankero:
@markus2330 I added my first error message template here and changed the majority of the existing messages.

Here are though some things I noticed:

- There is also another macro available [here](https://github.com/ElektraInitiative/libelektra/blob/9b48487e0321b8b415722892af5a92ae37eb11e6/src/include/kdbmacros.h#L64) which is only used 3 times. One could also make the error message templates like this.
- I though about completely removing the possibility to add any text in the for that particular error and take a default one instead but some custom messages make sense such as in [filename.c](https://github.com/Piankero/libelektra/blob/79be253491e35d12e3c7daadf23b4d4cec239c06/src/plugins/resolver/filename.c#L512-L513)

markus2330:
> There is also another macro available here which is only used 3 times. One could also make the error message templates like this.

Yes, I found this method to be the best. But actually I forgot about ELEKTRA_MALLOC_ERROR. So it would be important that we do not have  ELEKTRA_SET_OUT_OF_MEMORY_ERROR, so that people will use the correct one (ELEKTRA_MALLOC_ERROR, imho it should be renamed to ELEKTRA_SET_OUT_OF_MEMORY_ERROR for consistency).

Maybe we can find clever grouping of errors, so that for a specific group, only one template (macro) matches, like we achieved with out-of-memory.

> I thought about completely removing the possibility to add any text for that particular error and take a default one instead but some custom messages make sense such as in filename.c

I think it is not a big problem: if it was realloc or malloc does not matter, you immediately see it in the source code. The size is dynamic, so you cannot see it in the source code. So this information can be helpful (for a very large allocation it is obvious that it will fail).

Piankero:
> ELEKTRA_MALLOC_ERROR, imho it should be renamed to ELEKTRA_SET_OUT_OF_MEMORY_ERROR for consistency

I am confused what you actually want. You want the original macro `ELEKTRA_SET_OUT_OF_MEMORY_ERROR` and `ELEKTRA_SET_OUT_OF_MEMORY_ERRORF` to be removed and be substituted by a renamed version of ELEKTRA_MALLOC_ERROR?

> I think it is not a big problem

The text I marked in filename.c does not give more information about malloc or realloc but it gives a concrete message that it takes a different directory which is definitely important for the user. A complete removal of the dynamic text seems too restricting

> Maybe we can find clever grouping of errors, so that for a specific group, only one template (macro) matches, like we achieved with out-of-memory.

I highly doubt that this is possible but I will take a look at each group. OOM errors are the exception because they are nearly identical in every case whereas I could not think of any template to put in all semantic validation errors.

markus2330:
>  I am confused what you actually want. You want the original macro ELEKTRA_SET_OUT_OF_MEMORY_ERROR and ELEKTRA_SET_OUT_OF_MEMORY_ERRORF to be removed and be substituted by a renamed version of ELEKTRA_MALLOC_ERROR?

Yes, exactly.

Open for me is if there is a "size" in all cases.

> The text I marked in filename.c does not give more information about malloc or realloc but it gives a concrete message that it takes a different directory which is definitely important for the user

This is a very exotic case and the code seems to be quite broken. This case does not justify to add a dynamic text.

Furthermore, "defaulting to /" could be in a different warning (resource) as above. (As it is only a warning, we can yield several ones.)

> I highly doubt that this is possible but I will take a look at each group. OOM errors are the exception because they are nearly identical in every case whereas I could not think of any template to put in all semantic validation errors.

Maybe "semantic value validation", i.e. where the value is incorrect according the specification/type.

But let us first overcome the problems related to out-of-memory errors.

Piankero:
> Open for me is if there is a "size" in all cases.

It seems that some sizes are not given: https://github.com/ElektraInitiative/libelektra/blob/8f4da8abb5124d5cf97bc4151d93f32e5e552f70/src/plugins/augeas/augeas.c#L435

markus2330:
In this case it is `sizeof("Unable to initialize augeas: ")+sizeof(aug_error_message (augeasHandle))`

markus2330:
Btw. I do not think that the given size needs to be byte-precise. Actually a rough estimation would be enough. It is only to know if there was a bug in the system (e.g. that it tried to allocate -1 bytes).

Piankero:
Since the other approach is more fitting I will close this PR

Issue #3220, opened 2019-11-15
cmerge: segfault


mpranj:
## Steps to Reproduce the Problem

Configure and make libelektra.

```sh
make run all
```

## Expected Result

All tests succeed.

## Actual Result

```
The following tests FAILED:
	210 - testshell_markdown_cmerge (Failed)
	225 - test_cmerge (SEGFAULT)
```

```
 valgrind ./bin/test_cmerge
==200629== Memcheck, a memory error detector
==200629== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==200629== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==200629== Command: ./bin/test_cmerge
==200629== 
CMERGE       TESTS
==================

Executing testValuesWithGivenLength with size 59
==200629== Jump to the invalid address stated on the next line
==200629==    at 0x0: ???
==200629==    by 0xF: ???
==200629==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==200629== 
==200629== 
==200629== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==200629==  Bad permissions for mapped region at address 0x0
==200629==    at 0x0: ???
==200629==    by 0xF: ???
==200629== 
==200629== HEAP SUMMARY:
==200629==     in use at exit: 4,388 bytes in 64 blocks
==200629==   total heap usage: 169 allocs, 105 frees, 81,447 bytes allocated
==200629== 
==200629== LEAK SUMMARY:
==200629==    definitely lost: 0 bytes in 0 blocks
==200629==    indirectly lost: 0 bytes in 0 blocks
==200629==      possibly lost: 0 bytes in 0 blocks
==200629==    still reachable: 4,388 bytes in 64 blocks
==200629==         suppressed: 0 bytes in 0 blocks
==200629== Rerun with --leak-check=full to see details of leaked memory
==200629== 
==200629== For lists of detected and suppressed errors, rerun with: -s
==200629== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
[1]    200629 segmentation fault (core dumped)  valgrind ./bin/test_cmerge

```

## System Information

- Elektra Version: master
- Fedora 31

## Further Log Files and Output
[cmerge.txt](https://github.com/ElektraInitiative/libelektra/files/3851971/cmerge.txt)

mpranj:
[...]

PR #3209, opened 2019-11-13
Merge

Chemin1:
#3106, #3128 and #3141 merged together. Each of them should work alone, so I hope this here works, too.

## Basics

These points need to be fulfilled for every PR:

- [ ] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [ ] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy.
- [ ] The PR is rebased with current master.

If you have any troubles fulfilling these criteria, please write
about the trouble as comment in the PR. We will help you.
But we cannot accept PRs that do not fulfill the basics.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

Reviewers will usually check the following:

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

If you are already Elektra developer:

- Add the "work in progress" label if you do not want the PR to be reviewed yet.
- Add the "ready to merge" label **if the basics are fulfilled** and you also
  say that everything is ready to be merged.

Issue #3131, opened 2019-10-29
Integrate new 3-way merge into tools


dominicjaeger:
Tools should be patched to use the new merge

- [ ] qt-gui (together with @darddan)
- [ ] elektrad (together with @raphi011)
- [ ] gsettings elektra_settings_backend_sync method (together with @mpranj)
- [ ] kdb import
- [ ] kdb editor
- [ ] kdb shell
- [ ] kdb mount
- [ ] kconfig (https://github.com/ElektraInitiative/kconfig together with @FelixResch)
- [ ] replace "kdb merge", remove "kdb cmerge"

markus2330:
Thank you for reporting this issue! I added a full list of all tools above.

raphi011:
Is there any example or documentation how to do this?

Chemin1:
`src/tools/kdb/cmerge.cpp` is the only thing close to an example for this

markus2330:
I agree with @raphi011 there definitely should be a nice example how to use the API.

markus2330:
I added "kconfig (https://github.com/ElektraInitiative/kconfig together with @FelixResch)"

raphi011:
An example and a pkg-config file would really help to use this.

markus2330:
I fully agree, @raphi011 can you maybe add this pkg-config file as you already know how to do it and it would take @Chemin1 quite some time to learn about it.

Examples you have in #3235 and #3236

Chemin1:
> Examples you have in #3235 and #3236

Actually, #3235 does not have a call to `elektraMerge` yet. I'm not sure how implement the merge here yet but will take a look at it as soon as possible.

#3222 in contrast has one.  

> src/tools/kdb/cmerge.cpp is the only thing close to an example for this

"close to" might have been misleading. `cmerge.cpp` **is** an example, too.

> it would take @Chemin1 quite some time to learn about it.

@raphi011  Thank you for #3241! @markus2330 is right, I have never done this before and no idea how it works.


@darddan, @FelixResch it would be great if you could take a look at those examples, too.

markus2330:
@Chemin1 I think all of these examples are not exactly what kconfig or elektrad need. What they need is something like what happens in the qt-gui.

[...]

PR #3221, opened 2019-11-15
Merge: Always initialize LibGit

Chemin1:
Should fix #3206 

This is required by their documentation. This certainly gives us memleaks but
avoids segfaults.

## Basics

      contains `_(my name)_`)

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/src/libs/merge/kdbmerge.c
+++ b/src/libs/merge/kdbmerge.c
@@ -988,21 +988,13 @@ KeySet * elektraMerge (KeySet * our, Key * ourRoot, KeySet * their, Key * theirR
 	}
 
 #ifdef LIBGITFOUND
-#ifndef CMERGE_ON_LINUX
 	git_libgit2_init ();
-	ELEKTRA_LOG ("Initializing LibGit2");
-#else
-	ELEKTRA_LOG ("Not initializing LibGit2, but using it all the same.");
-#endif
 	ELEKTRA_LOG ("cmerge can use libgit2 to handle arrays");
 	if (handleArrays (ourCropped, theirCropped, baseCropped, result, informationKey, strategy) > 0)
 	{
 		ksDel (result);
 		return NULL;
 	}
-#ifndef CMERGE_ON_LINUX
-	git_libgit2_shutdown ();
-#endif
 #else
 	ELEKTRA_LOG ("cmerge can NOT use libgit2 to handle arrays");
 #endif
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -338,6 +338,7 @@ you up to date with the multi-language support provided by Elektra.
 - Improved `range` plugin error message. _(Michael Zronek)_
 - Improved error codes documentation to clarify the hierarchy for developers. _(Michael Zronek)_
 - Release notes now use git's union merge driver. _(Dominic Jäger)_
+- Please remove me. I'm only here for the build server. _(Dominic Jäger)_
 
 ## Infrastructure
 
--- a/src/libs/merge/CMakeLists.txt
+++ b/src/libs/merge/CMakeLists.txt
@@ -4,9 +4,6 @@ if (LibGit2_FOUND)
 	set (CMERGE_INCLUDE_DIRS ${LibGit2_INCLUDE_DIRS})
 	set (CMERGE_LIBRARY_DIRS ${LibGit2_LIBRARIES})
 	add_definitions (-DLIBGITFOUND)
-	if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-		add_definitions (-DCMERGE_ON_LINUX)
-	endif ()
 endif ()
 
 # This is derived from LibAddPlugin. LibAddLib has no such functionality yet. This is required so that libelektra-full.so has defined
--- a/tests/ctest/CMakeLists.txt
+++ b/tests/ctest/CMakeLists.txt
@@ -40,6 +40,5 @@ target_link_elektra (test_opts elektra-opts)
 
 target_link_elektra (test_cmerge elektra-merge)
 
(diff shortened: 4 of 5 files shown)


src/libs/merge/kdbmerge.c
```diff
@@ -988,21 +988,13 @@ KeySet * elektraMerge (KeySet * our, Key * ourRoot, KeySet * their, Key * theirR
...
-	ELEKTRA_LOG ("Not initializing LibGit2, but using it all the same.");
-#endif
 	ELEKTRA_LOG ("cmerge can use libgit2 to handle arrays");
 	if (handleArrays (ourCropped, theirCropped, baseCropped, result, informationKey, strategy) > 0)
 	{
 		ksDel (result);
 		return NULL;
 	}
-#ifndef CMERGE_ON_LINUX
-	git_libgit2_shutdown ();
```
mpranj: Why do we do the init but no shutdown here?
Btw if it is definitely an upstream problem/leak we can make a valgrind suppression for it and report the issue there.

markus2330: Some libraries are not safe to be initialized again after shutdown. Furthermore we do not really know if anybody else is using libgit2 in our process. So it can make sense to not call shutdown (if it is not working properly).

The whole global init and shutdown is calling for a lot of trouble. Unfortunately, many libraries do not use proper solutions (like handles).

In theory (with mutex protection and ref counting) multiple init/shutdown could be supported but it is rare that this is working properly.

> and report the issue there

I agree.

mpranj: > Some libraries are not safe to be initialized again after shutdown.

Ok, but the shutdown was clearly called before, and now we only call init. I assumed it was removed by mistake. If there is a reason not to call it, I'm fine with that!

Chemin1: > Why do we do the init but no shutdown here?

That was a mistake :/ Nonetheless, it is actually optional

> Usually you don’t need to call the shutdown function as the operating system will take care of reclaiming resources, but if your application uses libgit2 in some areas which are not usually active, you can use git_libgit2_shutdown(); to ask the library to clean up the global state. ([source](https://libgit2.org/docs/guides/101-samples/))

---

> Btw if it is definitely an upstream problem/leak we can make a valgrind suppression for it and report the issue there.

The person who answered my  [stackoverflow question](https://stackoverflow.com/questions/58477412/memory-problems-with-libgit2-initialization) also appears in the LibGit2 contributor list, so I assume they are aware.

markus2330: > I added memleak labels for both tests in this PR.

It is better if we suppress everything from libgit. This way we still check if our code has memleaks.

>   The person who answered my stackoverflow question also appears in the LibGit2 contributor list, so I assume they are aware.

You never know. Better we report it. But they are obviously only interested in the latest versions, maybe they already fixed it?

Chemin1:
> It is better if we suppress everything from libgit. This way we still check if our code has memleaks.

Those memleak labels are already in #3209 btw.

> It is better if we suppress everything from libgit. 

I don't know how (yet).

> But they are obviously only interested in the latest versions, maybe they already fixed it?

This is why I'd prefer to test with their latest version first and report afterwards if necessary.

@markus2330 What is higher priority: Stuff like tool integration for the merge library #3131 or this? I don't mind spending time on researching this further, but there is not too much time left until December.

markus2330:
> I don't know how (yet).

`valgrind --gen-suppressions` and add this to tests/valgrind.suppression

> Stuff like tool integration for the merge library #3131 or this?

It should not segfault and should work reliable. If it does not segfault, of course tool integration is more important than valgrind supressions.

markus2330:
So you are right, maybe having memleak label is a good solution.

Chemin1:
I created a Fedora 31 VM and compiled this branch in it. Strangely enough some tests (e.g. base64, check gen, testscr_check_gen) randomly fail with `make run_all`. However, nothing seems related to merging and `valgrind ./bin/test_cmerge` exits successfully with 0 bytes in use at exit.

Therefore I am relatively sure that this PR really fixes #3220.

markus2330:
Thank you!

> I created a Fedora 31 VM [...] Strangely enough some tests (e.g. base64, testscr_check_formatting, testscr_check_gen IIRC) randomly

If you already created the VM: Can you check if one of them always fails? (And report if it is so.)

> Therefore I am relatively sure that this PR really fixes #3220.

Ok. Just to be sure: when doing the cleanup there are also problems/crashes? Otherwise we should call the cleanup. (First I thought you removed it by purpose and I gave reasons why you may have done it. But later you said you removed it by accident.)

Chemin1:
> Just to be sure: when doing the cleanup there are also problems/crashes?

I haven't tested out the effects of the library function actually... IIRC I always had the shutdown when I initialized it. Until yesterday night, when I accidentally removed the shutdown while keeping the initialization.

> later you said

That was unfortunate because I started to answer before your answer was posted and I did not reload.

Nevertheless, both versions, shutting down and not shutting down should be OK according to the LibGit documentation.

mpranj:
@Chemin1 thank you!

Regarding the other problems: I use Fedora mainly and found no problems unique to the system. The usual suspects (dbusrecv, zeromq,..) can time out but this is not unique to Fedora.

Nevertheless it might be nice to add a Fedora docker image to the tests as the package versions are usually much newer than Debian.

markus2330:
> Nevertheless it might be nice to add a Fedora docker image to the tests as the package versions are usually much newer than Debian.

Yes, I fully agree that it would be amazing to have Fedora docker images for our Jenkins tests.

Chemin1:
All the errors that I got seem to really have been random. I just got three successful `kdb run_all`.

PR #3234, opened 2019-11-17
Base for tool integration

Chemin1:
These are some fixes necessary for integration of the merge library into the tools

Issue #1306, opened 2017-01-30
make mount not fail if mp exists


markus2330:
If someone tries to mount something that is mounted already, we should not fail. (If it would do exactly the same action)

Affected:

- [ ] `mount` (the script `elektra-mount` is basically already a workaround for it)
- [ ] `global-mount` see  #1301 src/plugins/conditionals/README.md 
- [ ] `elektra-mount`can be removed then

Unclear:

- [ ] do we need an option to disable this behavior?

markus2330:
I have set the milestone to 0.8.21 because it should be part of LCDproc release.

We also need this feature for #1374 and #1663

markus2330:
A simple way to implement that feature would be to have a `-f` option which cuts the mountpoint's config before doing the mount.

If this is not a good way, we should rather move this feature to 0.8.22.

markus2330:
With `--strategy preserve` this works now for normal mounts.

markus2330:
@Chemin1 It would be great to have a fully working preserve strategy, so that this finally correctly works for mount and global-mount.

markus2330:
Sorry, actually we need "--strategy unchanged". And this should also be default for mounting.

The strategy is actually very simple: It only succeeds if ours, theirs and base is the same.

Chemin1:
I'll have to take a closer look at how that works, as I did not even know that mounting has strategies until now.

Looking at `our`, `their` and `base`: Is this related to the install-config-file script that we talked about? I mount a file and get the `base` and `our` version. I mount the same file again (this is now `their`) with strategy unchanged and the mount succeeds?

markus2330:
> I'll have to take a closer look at how that works, as I did not even know that mounting has strategies until now.

Basically every tool that writes more than a single key (in this case merging is trivial) different strategies are useful. So also `kdb shell` would profit from having 3-way merging integrated.

markus2330:
Project would be together with #1322.

PR #3235, opened 2019-11-17
Integrate merge into kdb mount

### LINKED ISSUE #3131: Integrate new 3-way merge into tools

Chemin1:
Depends on #3248 which depends on #3234
Partly closes #3131

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -350,6 +350,7 @@ you up to date with the multi-language support provided by Elektra.
 - Please remove me. I'm only here for the build server. _(Dominic Jäger)_
 - I'm only here for the build server. This PR contains only fix ups. _(Dominic Jäger)_
 - Please remove me. I'm only here for the build server. _(Dominic Jäger)_
+- Please remove me. I'm only here for the build server. _(Dominic Jäger)_
 
 ## Infrastructure
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -244,6 +244,7 @@ you up to date with the multi-language support provided by Elektra.
 - `kdb list-commands` and `kdb plugins-list` now sort their output in an alphabetical order _(Anton Hößl)_
 - `kdb plugin-list` does now mention in the helptext that with option `-v` the output is sorted by the plugin status _(Anton Hößl)_
 - `kdb import`, `kdb export` and `kdb editor` now search the plugin database for suitig plugins so it's now possible to run `kdb export /hello json` instead of having to specify the plugin for the desired format directly. _(Anton Hößl)_
+- Make `kdb mount` use merge library. _(Dominic Jäger)_
 - <<TODO>>
 
 ## Scripts
--- a/src/include/kdbmerge.h
+++ b/src/include/kdbmerge.h
@@ -23,6 +23,7 @@ enum
 	MERGE_STRATEGY_INTERACTIVE = 2,
 	MERGE_STRATEGY_OUR = 3,
 	MERGE_STRATEGY_THEIR = 4,
+	MERGE_STRATEGY_EQUAL = 5,
 };
 
 KeySet * elektraMerge (KeySet * our, Key * ourRoot, KeySet * their, Key * theirRoot, KeySet * base, Key * baseRoot, Key * resultKey,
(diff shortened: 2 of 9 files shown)


doc/news/_preparation_next_release.md
```diff
@@ -343,6 +344,7 @@ you up to date with the multi-language support provided by Elektra.
 - Improved error codes documentation to clarify the hierarchy for developers. _(Michael Zronek)_
 - Release notes now use git's union merge driver. _(Dominic Jäger)_
 - Please remove me. I'm only here for the build server. _(Dominic Jäger)_
+- Fix includes for merge. _(Dominic Jäger)_
```
markus2330: This is not needed.

src/include/kdbmerge.h
```diff
@@ -10,11 +10,12 @@
 #define KDBMERGE_H_
 
 #include "kdb.h"
-#include "kdberrors.h"
 #include "kdbprivate.h"
 
 #ifdef __cplusplus
 extern "C" {
+using Key = ckdb::Key;
```
markus2330: to fix your problem, put elektraMerge also under the namespace ckdb.

src/libs/merge/kdbmerge.c
```diff
@@ -952,6 +953,17 @@ KeySet * elektraMerge (KeySet * our, Key * ourRoot, KeySet * their, Key * theirR
 		       int strategy, Key * informationKey)
 {
 	ELEKTRA_LOG ("cmerge starts with strategy %d (see kdbmerge.h)", strategy);
+	if (strategy == MERGE_STRATEGY_EQUAL)
+	{
+		if ((strcmp (keyName (ourRoot), keyName (theirRoot)) == 0) && (strcmp (keyName (theirRoot), keyName (baseRoot)) == 0))
```
markus2330: With this check you do not know if the keysets are the same?

src/include/kdbmerge.h
Chemin1: Thank you for the hint! Fixed in #3234

src/libs/merge/kdbmerge.c
Chemin1: #1306 is the reason why I started this (= strategy unchanged). I thought if we merge user/test/a with user/test/a and user/test/a, how could there be any difference?

I took a look at the mount code and don't really have an idea how to include the merge yet. Consequently, this strategy was a bit guessed, too.

I couldn't find any comparison function for key sets. Did I overlook this or does it really not exist?

doc/news/_preparation_next_release.md
Chemin1: Fixed in #3234

src/libs/merge/kdbmerge.c
markus2330: > I thought if we merge user/test/a with user/test/a and user/test/a, how could there be any difference?

Ok, then different question: why does it depend on the strategy if we anyway merge "the same thing".

> I couldn't find any comparison function for key sets. Did I overlook this or does it really not exist?

It does not exist as there are so many variants of what it could mean that two KeySets are equal. And it is quite trivial to implement. One variant is implemented in compare_keyset tests/cframework/tests.h

I think in your case you want a quite strict variant: i.e. also the metadata needs to be the same.

tests/ctest/test_cmerge.c
```diff
@@ -302,6 +302,173 @@ static void testArrayWithDifferentLengths (void)
 	}
 }
 
+/**
+  * KeySets equality depends only on what is below root. Consequently the three sets
+  * user/test1 and user/test2 and user/test3 must be equal
```
Chemin1: @markus2330 Is this ok?

markus2330: GitHub only shows me the comment and can you ask me a more concrete question so that I know what should be ok?

PR #3236, opened 2019-11-17
Integrate merge into kdb editor

### LINKED ISSUE #3131: Integrate new 3-way merge into tools

Chemin1:
Depends on #3248 which depends on #3234
Partly closes #3131

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -244,6 +244,7 @@ you up to date with the multi-language support provided by Elektra.
 - `kdb list-commands` and `kdb plugins-list` now sort their output in an alphabetical order _(Anton Hößl)_
 - `kdb plugin-list` does now mention in the helptext that with option `-v` the output is sorted by the plugin status _(Anton Hößl)_
 - `kdb import`, `kdb export` and `kdb editor` now search the plugin database for suitig plugins so it's now possible to run `kdb export /hello json` instead of having to specify the plugin for the desired format directly. _(Anton Hößl)_
+- Make `kdb editor` use the new merge library. _(Dominic Jäger)_
 - <<TODO>>
 
 ## Scripts
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -350,6 +350,7 @@ you up to date with the multi-language support provided by Elektra.
 - Please remove me. I'm only here for the build server. _(Dominic Jäger)_
 - I'm only here for the build server. This PR contains only fix ups. _(Dominic Jäger)_
 - Please remove me. I'm only here for the build server. _(Dominic Jäger)_
+- Please remove me. I'm only here for the build server. _(Dominic Jäger)_
 
 ## Infrastructure
 
--- a/src/tools/kdb/editor.cpp
+++ b/src/tools/kdb/editor.cpp
@@ -13,19 +13,16 @@
 
 #include <kdbmacros.h>
 
+#include "kdbmerge.h"
+#include "helper/cmergehelper.hpp"
 #include <export.hpp>
 #include <import.hpp>
 
 #include <iostream>
 #include <string>
 
-#include <mergehelper.hpp>
-#include <merging/metamergestrategy.hpp>
-#include <merging/threewaymerge.hpp>
-
 using namespace kdb;
 using namespace kdb::tools;
-using namespace kdb::tools::merging;
 using namespace std;
 
 EditorCommand::EditorCommand ()
@@ -35,6 +32,7 @@ EditorCommand::EditorCommand ()
 EditorCommand::~EditorCommand ()
 {
 }
+
 
 void EditorCommand::tmpFile ()
 {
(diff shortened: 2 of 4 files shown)


src/tools/kdb/editor.cpp
```diff
@@ -36,6 +32,39 @@ EditorCommand::~EditorCommand ()
 {
 }
 
+
+static kdb::Key prependNamespace (kdb::Key const & root, std::string const & ns)
```
markus2330: I think I already saw this somewhere. Maybe we can move this to libtools?

Chemin1: > Maybe we can move this to libtools?

Something like that would be good I think. libtools = a new file in src/libs/tools/src ?

markus2330: Yes, and the header file could be in ~libtools/include/helper

doc/news/_preparation_next_release.md
```diff
@@ -344,6 +345,8 @@ you up to date with the multi-language support provided by Elektra.
 - Release notes now use git's union merge driver. _(Dominic Jäger)_
 - Please remove me. I'm only here for the build server. _(Dominic Jäger)_
 - I'm only here for the build server. This PR contains only fix ups. _(Dominic Jäger)_
+- Please remove me. I'm only here for the build server. _(Dominic Jäger)_
```
markus2330: Is there any advantage in including this? This definitely increases the chances of merge conflicts.

src/libs/tools/include/helper/cmergehelper.hpp
```diff
@@ -0,0 +1,25 @@
+/**
+ * @file
+ *
+ * @brief Helper for merge library
+ *
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ *
+ */
+
+#ifndef CMERGEHELPER_HPP_
```
markus2330: I really recommend you (again) to avoid using the word "cmerge". It will be a lot of work to rename all of this.

Furthermore, this file actually has nothing to do with merging, I would call it key (or keyset) helpers.

src/libs/tools/src/helper/cmergehelper.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+	KeySet ret;
+	for (auto const & k : resultKeys)
+	{
+		ret.append (prependNamespace (k, ns));
+	}
+	return ret;
+}
+
+void applyMeta (kdb::KeySet & imported, kdb::KeySet const & base)
```
markus2330:
```suggestion
void copyAllMeta (kdb::KeySet & to, kdb::KeySet const & from)
```

src/tools/kdb/cmerge.cpp
```diff
@@ -7,6 +7,7 @@
  */
```
markus2330: in cpp files you can use `using` (if you want)

doc/news/_preparation_next_release.md
Chemin1: I thought each PR needs a change in this file?

src/libs/tools/include/helper/cmergehelper.hpp
Chemin1: There is already a file called keyhelper.hpp and keyhelper.cpp. I now moved the functions into it instead of creating a new file like [initially planned](https://github.com/ElektraInitiative/libelektra/pull/3236#discussion_r347163691).
> It will be a lot of work to rename all of this.

Unfortunately, I don't know how we could completely avoid this. For example, I cannot simply rename `kdb cmerge` to `kdb merge` as it is already in use and it seems like the most appropriate name to me.

Chemin1:
@markus2330 Thank you for looking at it! I pushed an improved version and rebased all the dependent PRs (except those that I labelled ready to merge so that the successful checks remain).

Chemin1:

doc/news/_preparation_next_release.md
markus2330: Yes, but only one. In your PR there is also another line added. (- Make `kdb editor` use the new merge library. _(Dominic Jäger)_)

src/libs/tools/include/helper/cmergehelper.hpp
markus2330: >  Unfortunately, I don't know how we could completely avoid this. For example, I cannot simply rename kdb cmerge to kdb merge as it is already in use and it seems like the most appropriate name to me.

Of course you can, the old `kdb merge` is to be removed.

PR #3263, opened 2019-11-21
Semantic validation message template

Piankero:
## Proof of concept for semantic validation errors

This PR is a proof of concept and not intended to be merged as it is a basis for discussion. It compiles on my system (with minor warnings), tests are not fixed.

## What I have done

I introduced a different error signature for semantic validation errors that force the developer to provide the faulty key. This key then will get printed out along with its value.

An error message for a non-existing error not looks like this:

```sh
Using name user/tests/abc
Sorry, module network issued the error C03200:
Validation Semantic: Key user/tests/abc with value sshaa does not fulfill: Could not find service
```

instead of 

```sh
Using name user/tests/abc
Sorry, module network issued the error C03200:
Validation Semantic: Could not find service with name sshaa on key user/tests/abc
```

## What does it solve?

In the majority of the error messages either the key or the value was missing. 
Any new error message will easily forget one of these two essential parts. This signature enforces a correct error message which helps the user/admin find the erroneous configuration setting.

## Problems with this approach 

1.  https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/fcrypt/fcrypt.c#L677
    Erroneous keyset, not concrete Key.
2. opts.c
    Value usually not needed, just keyName (or keyBaseName)
3. lua.cpp https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/lua/lua.cpp#L78
    Probably no validation semantic errors
4. reference.c https://github.com/Piankero/libelektra/blob/master/src/plugins/reference/reference.c#L410
    Giving parentKey here is ok as other error message should yield concrete key

## Remedies

1. With current approach a Key has to be generated that holds the key + value if for example a keyset is erroneous. Furthermore to remedy the opts.c problem I would suggest that
        if value == null then do not show the value in the error message:
        Eg. Key "user/tests" failed to validate because of ....
        instead of
        Eg. Key "user/tests" with value "wrong" failed to validate because of ....
2. Rewrite signature to accept keyName + value instead of Key
	This solves the problem when a keyset is faulty and not a single key
3. Provide an extra macro that is primarily used for specification errors that expect the Key with a value
        The problematic cases could then be ignored
        The knowledge of this extra macro though might not be obvious to new developers

## Implications

Some methods have to be rewritten to provide the needed information

### DIFF
--- a/src/include/kdberrors.h
+++ b/src/include/kdberrors.h
@@ -237,19 +237,19 @@ using KeySet = ckdb::KeySet;
 						       ELEKTRA_STRINGIFY (ELEKTRA_MODULE_NAME), reason, __VA_ARGS__);                      \
 	} while (0)
 
-#define ELEKTRA_SET_VALIDATION_SEMANTIC_ERROR(parentKey, wrongKey, reason)                                                                                 \
+#define ELEKTRA_SET_VALIDATION_SEMANTIC_ERROR(parentKey, wrongKey, reason)                                                                 \
 	do                                                                                                                                 \
 	{                                                                                                                                  \
 		ELEKTRA_LOG ("Add Error %s: %s", ELEKTRA_ERROR_VALIDATION_SEMANTIC, reason);                                               \
-		elektraSetErrorVALIDATION_SEMANTIC (parentKey, wrongKey, __FILE__, ELEKTRA_STRINGIFY (__LINE__), ELEKTRA_STRINGIFY (ELEKTRA_MODULE_NAME),  \
-						    reason);                                                                               \
+		elektraSetErrorVALIDATION_SEMANTIC (parentKey, wrongKey, __FILE__, ELEKTRA_STRINGIFY (__LINE__),                           \
+						    ELEKTRA_STRINGIFY (ELEKTRA_MODULE_NAME), reason);                                      \
 	} while (0)
-#define ELEKTRA_SET_VALIDATION_SEMANTIC_ERRORF(key, wrongKey, reason, ...)                                                                           \
+#define ELEKTRA_SET_VALIDATION_SEMANTIC_ERRORF(key, wrongKey, reason, ...)                                                                 \
 	do                                                                                                                                 \
 	{                                                                                                                                  \
 		ELEKTRA_LOG ("Add Error %s: " reason, ELEKTRA_ERROR_VALIDATION_SEMANTIC, __VA_ARGS__);                                     \
-		elektraSetErrorVALIDATION_SEMANTIC (key, wrongKey, __FILE__, ELEKTRA_STRINGIFY (__LINE__), ELEKTRA_STRINGIFY (ELEKTRA_MODULE_NAME),  \
-						    reason, __VA_ARGS__);                                                                  \
+		elektraSetErrorVALIDATION_SEMANTIC (key, wrongKey, __FILE__, ELEKTRA_STRINGIFY (__LINE__),                                 \
+						    ELEKTRA_STRINGIFY (ELEKTRA_MODULE_NAME), reason, __VA_ARGS__);                         \
 	} while (0)
 #define ELEKTRA_ADD_VALIDATION_SEMANTIC_WARNING(key, reason)                                                                               \
 	do                                                                                                                                 \
@@ -291,8 +291,10 @@ extern const char * ELEKTRA_ERROR_VALIDATION_SEMANTIC_NAME;
 extern const char * ELEKTRA_WARNING_VALIDATION_SEMANTIC;
 extern const char * ELEKTRA_WARNING_VALIDATION_SEMANTIC_NAME;
 
-void elektraSetErrorVALIDATION_SEMANTIC (Key * parentKey, Key * wrongKey, const char * file, const char * line, const char * module, const char * reason, ...);
-void elektraAddWarningVALIDATION_SEMANTIC (Key * parentKey, const char * file, const char * line, const char * module, const char * reason, ...);
+void elektraSetErrorVALIDATION_SEMANTIC (Key * parentKey, Key * wrongKey, const char * file, const char * line, const char * module,
+					 const char * reason, ...);
+void elektraAddWarningVALIDATION_SEMANTIC (Key * parentKey, const char * file, const char * line, const char * module, const char * reason,
+					   ...);
 
 #undef DECLARE_ERROR_CODE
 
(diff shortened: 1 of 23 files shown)


markus2330:
Please do not write text in the PR description but add such a text in a design decision so that it will be available for people trying to understand the decisions of Elektra.

> This signature enforces a correct error message which helps the user/admin find the erroneous configuration setting.

I very much agree to this.

>   https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/fcrypt/fcrypt.c#L677
>    Erroneous keyset, not concrete Key.

This is not a semantic **validation** error, as there is simply no gpg key for this encrypted file. If we say "gpg keys" are resources, it is a resource error.

>    opts.c   Value usually not needed, just keyName (or keyBaseName)

Which code are you referring at?

>    lua.cpp https://github.com/ElektraInitiative/libelektra/blob/master/src/plugins/lua/lua.cpp#L78
>    Probably no validation semantic errors
 >   reference.c https://github.com/Piankero/libelektra/blob/master/src/plugins/reference/reference.c#L410
 >   Giving parentKey here is ok as other error message should yield concrete key

Yes, and actually we have N keys that build the circle and could be written out here.

So we learned that such templates make sense: with them you find out that you actually wrongly classified keys.

Furthermore (in the cycle) we also learned that syntax and semantic validation is not everything but there is also something like "structural validation" which says that these keys in this combination cannot be valid even if all of these keys individually are ok.

>Rewrite signature to accept keyName + value instead of Key
This solves the problem when a keyset is faulty and not a single key

I think for structural problems it is not possible to give a single keyname+value. You could pass a KeySet of the minimal set of conflicting keys to the error message printer.

Piankero:
> Please do not write text in the PR description but add such a text in a design decision so that it will be available for people trying to understand the decisions of Elektra.

I did not expect that this POC will be a design decision. As written in the description this is only a basis for discussion.

> Which code are you referring at?

Eg here: 

https://github.com/ElektraInitiative/libelektra/blob/0534984b416d4f8b2c42401e2635b726afdbe44d/src/libs/opts/opts.c#L1205

The actual key value is not needed because the long opt does not seem to have a value associated with it. That's why I think that the error message only prints the value if it is =!null.

> This is not a semantic validation error, as there is simply no gpg key for this encrypted file. If we say "gpg keys" are resources, it is a resource error.

Alright, this was not clear to me since fcrypt and gpg errors are difficult to understand for me.

> So we learned that such templates make sense: with them you find out that you actually wrongly classified keys.

I agree that templates can guide the developers to find the correct (or at least best fitting) category.

> Furthermore (in the cycle) we also learned that syntax and semantic validation is not everything but there is also something like "structural validation" which says that these keys in this combination cannot be valid even if all of these keys individually are ok.

I think that a structural error is more like a suberror of semantic errors. A correct structure is semantic requirement in my opinion.

Probably an overloaded error method is more appropriate where a KeySet can be passed instead of a single Key? This could probably also be used for the math plugin where all keys can be given that lead to an error instead of the single one the user set. I though remember that you once told me that overloading in C is very tricky.

markus2330:
> I did not expect that this POC will be a design decision. As written in the description this is only a basis for discussion.

Discussions should lead to decisions at some point.

> The actual key value is not needed because the long opt does not seem to have a value associated with it. That's why I think that the error message only prints the value if it is =!null.

This is really quite interesting. In fact, it **could** have a value (some options have arguments). I do not know where the value can be read from, though. Maybe @kodebach can help out.

> I think that a structural error is more like a suberror of semantic errors. A correct structure is semantic requirement in my opinion.

Yes, I agree.

> Probably an overloaded error method is more appropriate where a KeySet can be passed instead of a single Key? This could probably also be used for the math plugin where all keys can be given that lead to an error instead of the single one the user set. I though remember that you once told me that overloading in C is very tricky.

Overloading does not exist (neither for functions nor macros). Of course we could have a macro with a different name. But imho making a new category would be the right choice here.

But lets see if there are any other exceptions where there is no name/value. Or did you already go through all semantic validation errors?

kodebach:
> > The actual key value is not needed because the long opt does not seem to have a value associated with it. That's why I think that the error message only prints the value if it is =!null.
> 
> This is really quite interesting. In fact, it **could** have a value (some options have arguments). I do not know where the value can be read from, though. Maybe @kodebach can help out.

There is no `keyName` in the error message here, because there is no faulty `Key`. This error happens, if a program is called with an unknown option, e.g. calling `kdb --foo` would result in `"Unknown long option: --foo"` (if `kdb` used the opts library). That the option name comes from the base name of a `Key` is just an implementation detail.

If the error class "Validation Semantic" requires a key, we need to choose a different error class here.

> reference.c https://github.com/Piankero/libelektra/blob/master/src/plugins/reference/reference.c#L410
Giving parentKey here is ok as other error message should yield concrete key

Here, the cycle is already written to the log (although only with `ELEKTRA_LOG_NOTICE`), since that is much easier than creating a single message for the error. You could just add one of the `Key`s in the cycle to the error. They all reference each other so finding the cycle shouldn't be that hard.

Also, I'm pretty sure there are two bugs in the logging code:
https://github.com/ElektraInitiative/libelektra/blob/b83c54f7321342cde2130d4a5cb0889bf44caf52/src/plugins/reference/reference.c#L208-L218

Inside the loop `root` should be `node` and `ksDel (nodes)` is missing after the loop.

> Of course we could have a macro with a different name. But imho making a new category would be the right choice here.

A general macros e.g. `ELEKTRA_ERROR_ADD_KEY` / `ELEKTRA_WARNING_ADD_KEY` that add a key to the metaarray `error/keys/#` / `warning/#NN/keys/#` (`NN` is the last warning). Might be better, then you don't have to construct a throw away `KeySet`, if you have e.g. two keys that are in conflict.

These could also be used for all kinds of errors. "Validation semantic" would require a key in the normal macros and just add it as the first in the `keys/#`metaarray.

markus2330:
> This error happens, if a program is called with an unknown option, e.g. calling kdb --foo would result in "Unknown long option: --foo" (if kdb used the opts library). That the option name comes from the base name of a Key is just an implementation detail.

So it is basically the same as an unknown commands in the dump file. This was categorized as syntactic validation error.

> Here, the cycle is already written to the log (although only with ELEKTRA_LOG_NOTICE), since that is much easier than creating a single message for the error. You could just add one of the Keys in the cycle to the error. They all reference each other so finding the cycle shouldn't be that hard.

Yes, I agree that we should not put too much effort into this and any key to be within the error message suffices.

>  A general macros e.g. ELEKTRA_ERROR_ADD_KEY / ELEKTRA_WARNING_ADD_KEY that add a key to the metaarray error/keys/# / warning/#NN/keys/# (NN is the last warning). Might be better, then you don't have to construct a throw away KeySet, if you have e.g. two keys that are in conflict.

But how should these keys be rendered in the error message?

Most of the validation errors relate to a single key, so it makes sense to me to make this situation the easiest. The idea was that maybe the structure of the error message and the category can be the same in some situations.

PR #3300, opened 2019-11-28
Docker: add alpine image with Elektra installed

### LINKED ISSUE #3307: Release Alpine image: update Googletest
After the 0.9.2 release, update the official elektra docker image and update to Googletest 1.10.0 (or latest).

Cannot be done before 0.9.2 is released, as 0.9.1 tag does not work with Googletest >= 1.10.0.

mpranj:
Closes #3307.

### DIFF
--- a/scripts/docker/alpine/3.10/release.Dockerfile
+++ b/scripts/docker/alpine/3.10/release.Dockerfile
@@ -1,17 +1,23 @@
-FROM alpine:3.10
+FROM alpine:3.10.3
 
 RUN apk update \
     && apk add --no-cache --upgrade\
+        augeas-dev \
+        bash \
         bison \
+        boost \
+        boost-dev \
         build-base \
         cmake \
         curl \
-        libgit2 \
         git \
+        libgit2 \
+        libgit2-dev \
         ninja \
-        tcl \
-        yaml-cpp-dev \
-        sudo
+        sudo \
+        yajl-dev \
+        yaml-cpp-dev
+
 
 # Google Test (TODO: update before 0.9.2 to gtest 1.10.0, but does not work with elektra 0.9.1)
 ENV GTEST_ROOT=/opt/gtest
@@ -56,7 +62,7 @@ RUN mkdir build \
     && rm -Rf ${GTEST_ROOT}
 
 
-FROM alpine:3.10
+FROM alpine:3.10.3
 COPY --from=0 ${ELEKTRA_ROOT} \
               ${ELEKTRA_ROOT}
 ARG USERID=1000
@@ -71,10 +77,11 @@ RUN cd build \
 
 RUN apk del \
         bison \
+        boost-dev\
         build-base \
         cmake \
         git \
-        tcl \
+        libgit2-dev \
         && rm -rf /var/cache/apk/*
 
 RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
--- a/scripts/docker/alpine/3.10/release.Dockerfile
+++ b/scripts/docker/alpine/3.10/release.Dockerfile
@@ -14,7 +14,9 @@ RUN apk update \
         libgit2-dev \
         ninja \
         sudo \
+        yajl \
         yajl-dev \
+        yaml-cpp \
         yaml-cpp-dev
 
 
--- a/scripts/docker/alpine/3.10/release.Dockerfile
+++ b/scripts/docker/alpine/3.10/release.Dockerfile
@@ -43,6 +43,7 @@ RUN mkdir build \
              -DTOOLS="ALL" \
              -DENABLE_DEBUG="OFF" \
              -DENABLE_LOGGER="OFF" \
+             -DCMAKE_BUILD_TYPE="Release" \
              -DKDB_DB_SYSTEM='/home/elektra/.config/kdb/system' \
              -DKDB_DB_SPEC='/home/elektra/.config/kdb/spec' \
              -DKDB_DB_HOME='/home/elektra/.config/kdb/home' \
@@ -51,7 +52,8 @@ RUN mkdir build \
     && ctest -T Test --output-on-failure -j ${PARALLEL} \
     && rm -Rf '/home/elektra/.config' '/home/elektra/.cache' \
     && cmake -DBUILD_TESTING=OFF -UKDB_DB_SYSTEM -UKDB_DB_SPEC -UKDB_DB_HOME . \
-    && make -j ${PARALLEL}
+    && make -j ${PARALLEL} \
+    && rm -Rf ${GTEST_ROOT}
 
 
 FROM alpine:3.10
--- a/scripts/docker/alpine/3.10/release.Dockerfile
+++ b/scripts/docker/alpine/3.10/release.Dockerfile
@@ -6,14 +6,13 @@ RUN apk update \
         build-base \
         cmake \
         curl \
-        diffutils \
-        file \
+        libgit2 \
         git \
         ninja \
-        ronn \
+        tcl \
         yaml-cpp-dev
 
-# Google Test
+# Google Test (TODO: update before 0.9.2 to gtest 1.10.0, but does not work with elektra 0.9.1)
 ENV GTEST_ROOT=/opt/gtest
 ARG GTEST_VER=release-1.8.1
 RUN mkdir -p ${GTEST_ROOT} \
@@ -32,6 +31,12 @@ RUN mkdir -p ${ELEKTRA_ROOT} \
     && tar -zxvf elektra.tar.gz --strip-components=1 -C ${ELEKTRA_ROOT} \
     && rm elektra.tar.gz
 
+ARG USERID=1000
+RUN adduser \
+    -u ${USERID} \
+    -D \
+    elektra
+
 ARG PARALLEL=8
 WORKDIR ${ELEKTRA_ROOT}
 RUN mkdir build \
@@ -40,18 +45,17 @@ RUN mkdir build \
              -DTOOLS="ALL" \
              -DENABLE_DEBUG="OFF" \
              -DENABLE_LOGGER="OFF" \
+             -DKDB_DB_SYSTEM='/home/elektra/.config/kdb/system' \
+             -DKDB_DB_SPEC='/home/elektra/.config/kdb/spec' \
+             -DKDB_DB_HOME='/home/elektra/.config/kdb/home' \
              .. \
     && make -j ${PARALLEL} \
-    && ctest -T Test --output-on-failure -j ${PARALLEL} -LE kdbtests
+    && ctest -T Test --output-on-failure -j ${PARALLEL}
 
 
 FROM alpine:3.10
 COPY --from=0 ${ELEKTRA_ROOT} \
               ${ELEKTRA_ROOT}
-
-RUN apk update \
-    && apk add --no-cache --upgrade\
-        bash
 
 ENV ELEKTRA_ROOT=/opt/elektra
 WORKDIR ${ELEKTRA_ROOT}
@@ -60,16 +64,11 @@ RUN cd build \
     && ldconfig /usr/local/lib/elektra/ \
     && rm -Rf ${ELEKTRA_ROOT}
 
+RUN echo "alias sudo='' # in this image we do not need to be root" >> /etc/profile
+RUN echo "export PS1='\u $ '" >> /etc/profile
 ENV LD_LIBRARY_PATH=/usr/local/lib/elektra/
+ENV ENV="/etc/profile"
 
-# Create User:Group
-# The id is important as jenkins docker agents use the same id that is running
-# on the slaves to execute containers
 ARG USERID=1000
-RUN adduser \
-    -u ${USERID} \
-    -D \
-    -s /bin/bash \
-    elektra
 USER ${USERID}
 WORKDIR /home/elektra
--- a/scripts/docker/alpine/3.10/release.Dockerfile
+++ b/scripts/docker/alpine/3.10/release.Dockerfile
@@ -96,7 +96,7 @@ RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 RUN echo "alias sudo='sudo -i' # in this image we do not need to be root" >> /etc/profile
 RUN echo "export PS1='\u $ '" >> /etc/profile
 RUN echo "export LD_LIBRARY_PATH=/usr/local/lib/elektra/" >> /etc/profile
-RUN echo "export ALLUSERSPROFILE=/" >> /etc/profile
+RUN echo "export ALLUSERSPROFILE=''" >> /etc/profile
 
 USER ${USERID}
 WORKDIR /home/elektra
--- a/scripts/docker/alpine/3.10/release.Dockerfile
+++ b/scripts/docker/alpine/3.10/release.Dockerfile
@@ -9,8 +9,6 @@ RUN apk update \
         build-base \
         cmake \
         curl \
-        fts \
-        fts-dev \
         git \
         libgit2 \
         libgit2-dev \
@@ -81,7 +79,6 @@ RUN apk del \
         bison \
         build-base \
         cmake \
-        fts-dev \
         git \
         libgit2-dev \
         yajl-dev \
--- a/scripts/docker/alpine/3.10/release.Dockerfile
+++ b/scripts/docker/alpine/3.10/release.Dockerfile
@@ -39,6 +39,9 @@ RUN mkdir -p ${ELEKTRA_ROOT} \
     && tar -zxvf elektra.tar.gz --strip-components=1 -C ${ELEKTRA_ROOT} \
     && rm elektra.tar.gz
 
+# PATCH YAML ENDLINE (TODO: remove this)
+RUN sed -i '536s/output << data;/output << data << endl;/' ${ELEKTRA_ROOT}/src/plugins/yamlcpp/write.cpp
+
 ARG USERID=1000
 RUN adduser -u ${USERID} -G wheel -D elektra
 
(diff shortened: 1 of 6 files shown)


scripts/docker/alpine/3.10/release.Dockerfile
```diff
@@ -0,0 +1,75 @@
+FROM alpine:3.10
+
+RUN apk update \
+    && apk add --no-cache --upgrade\
+        bison \
+        build-base \
+        cmake \
+        curl \
+        diffutils \
```
markus2330: Why is this needed? libgit2 would be needed for two plugins.

scripts/docker/alpine/3.10/release.Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+        file \
+        git \
+        ninja \
+        ronn \
```
markus2330: Is this needed? Does someone wants to compile the man pages in this docker image?

scripts/docker/alpine/3.10/release.Dockerfile
```diff
@@ -0,0 +1,75 @@
+FROM alpine:3.10
+
+RUN apk update \
+    && apk add --no-cache --upgrade\
+        bison \
```
markus2330: also flex will be needed for toml

scripts/docker/alpine/3.10/release.Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+        build-base \
+        cmake \
+        curl \
+        diffutils \
+        file \
+        git \
+        ninja \
+        ronn \
+        yaml-cpp-dev
+
```
markus2330: Maybe some additional deps like yajl, xerces, augeas would be nice. (If they are available)

scripts/docker/alpine/3.10/release.Dockerfile
mpranj: I blindly copied the other alpine image and adapted it quickly. :see_no_evil:

scripts/docker/alpine/3.10/release.Dockerfile
mpranj: Same as above.

scripts/docker/alpine/3.10/release.Dockerfile
mpranj: Please do note that this is only for building. Then we copy it to a fresh image with only bash installed and the layers are squashed so the image should have minimal size.

Basically my only goal was to provide a minimal image and drop the user into a shell where kdb is available. I do agree that it would be nice to add more plugins for a rich Elektra experience.

markus2330: Yes, this can also be done in the next release. I hope everything for the tutorial is included in the current image?

mpranj: I did not go through the tutorial. I'll update the image if something is missing. (hopefully alpine has packages for everything needed)

scripts/docker/alpine/3.10/release.Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+RUN apk update \
+    && apk add --no-cache --upgrade\
+        bison \
+        build-base \
+        cmake \
+        curl \
+        libgit2 \
+        git \
+        ninja \
+        tcl \
```
sanssecours: Are you sure we need [Tcl](https://en.wikipedia.org/wiki/Tcl) for Elektra? As far as I know no part of Elektra requires the programming language. If you want to include the [`tcl` plugin](https://www.libelektra.org/plugins/tcl), then I think adding a package for [Boost](https://www.boost.org) would be the right choice.

mpranj: Yes, the idea was to include the tcl plugin as it was part of our asciinema tutorial.

I saw tcl being installed in other docker images so I thought this is needed for the plugin. Thanks for reviewing it, I will add boost.

Further suggestions are absolutely welcome.

mpranj:
I pushed `elektra/elektra:0.9.1-6` to docker hub. Unfortunately boost pulls in python2 and python3 as dependencies, so the image naturally got much larger. (162MB, ~54MB compressed)

It's really a tradeoff, we will have to decide what should be in this image. To me the size is not extreme, but the minimal image had about 7MB.

In return we get a nice list of [plugins](https://github.com/ElektraInitiative/libelektra/files/3908306/plugins.txt).

scripts/docker/alpine/3.10/release.Dockerfile
markus2330: Yes, this is a common error :wink: Please remove the references of the tcl dependency in the docker images. Elektra never used the tcl interpreter.

markus2330:
Impressive list of plugins :100: 

> I pushed elektra/elektra:0.9.1-6 to docker hub. Unfortunately boost pulls in python2 and python3 as dependencies, so the image naturally got much larger. (162MB, ~54MB compressed)

Unfortunately, the python plugin nevertheless is not included. The tcl plugin is very low priority, so better to not add boost if it pulls python without giving us the python plugin.

> It's really a tradeoff, we will have to decide what should be in this image. To me the size is not extreme, but the minimal image had about 7MB.

I think you balanced it very nicely. Do you know why multifile is not included? https://pkgs.alpinelinux.org/package/edge/main/armv7/fts with 24k sounds to be quite cheap :wink:

markus2330:
Btw. do not waste too much time on fts, maybe we should not use it anyway (and rewrite multifile), see https://wiki.musl-libc.org/faq.html "Why is fts.h not included?"

Let us see if @darddan will find multifile useful for KConfig or if we go with individual mountpoints anyway.

mpranj:
I tried to add `fts` and `fts-dev`. It leads to failures because:
```
Installation: Dlopen failed. Could not load module libelektra-multifile.so. Reason: Error relocating /opt/elektra/build/lib/libelektra-multifile.so: fts_open: symbol not found
```

I don't know how what's going on here, the `LD_LIBRARY_PATH` is set correctly and `nm` says the symbol exists in libfts. Only multifile fails out of all tests.

markus2330:
> I don't know what's going on here, the LD_LIBRARY_PATH is set correctly and nm says the symbol exists in libfts. Only multifile fails out of all tests.

Yes, then simply remove libfts again. Maybe the implementations are somehow incompatible as suggested in the FAQ.

scripts/docker/alpine/3.10/release.Dockerfile
mpranj: Let's add that when it's needed.

markus2330:
Maybe it makes sense to wait for @Piankero with his demo before merging this.

markus2330:
And it seems like the little patch will not be needed: #3316

mpranj:
> Maybe it makes sense to wait for @Piankero with his demo before merging this.

Sure, let's wait.

> And it seems like the little patch will not be needed

It is needed if we do not publish a new release or change the image to use master instead of the 0.9.1 release. I think for now it's easier to leave it as is until 0.9.2, unless there are some other major fixes.

markus2330:
> It is needed if we do not publish a new release or change the image to use master instead of the 0.9.1 release. I think for now it's easier to leave it as is until 0.9.2, unless there are some other major fixes.

Yes, sounds ok in this case because people trying out the tutorial will most likely use 0.9.2 (as usually only with releases new people start trying out Elektra) or use the docker image.

See also #3319 for a proposal of versioning in future.

Piankero:
I'll write here what actually did not work.

I can user the hosts plugin but I can not add anything to the hosts file:
```sh
sudo kdb mount --with-recommends /etc/hosts system/hosts hosts
sudo kdb set system/hosts/ipv4/my 88.198.134.177
Sorry, 1 warning was issued ;(
	Sorry, module kdb issued the warning C01310:
	Internal: Error during commit. This means backend is broken: system/hosts
Sorry, module resolver_fm_hpu_b issued the error C01100:
Resource: Could not rename file '/etc/hosts.11:1575400385.54897.tmp'. Reason: Resource busy
Create a new key system/hosts/ipv4/my with string "88.198.134.177"
```

this seems to be docker related because I can do it locally

markus2330:
Yes, "Resource busy" sounds like /etc is some overlay-fs or similar. @mpranj any ideas?

mpranj:
Wow, we've managed to hit a very specific error.

In docker you cannot replace bind-mounted files, but you can edit them. Elektra does exactly that, it tries to rename the temp file to the final file name to have a clean commit.

Unfotunately it seems that Docker auto-mounts about 3 files, and `/etc/hosts` is one of them 😢. I'll see if something can be done.

markus2330:
The easiest workaround is to tell Elektra to use a different resolver that does not move files, e.g. wresolver. So `kdb mount -R wresolver ...`.

mpranj:
> The easiest workaround is to tell Elektra to use a different resolver that does not move files, e.g. wresolver. So kdb mount -R wresolver ....

Thank you for this hint. I think it's best to do this, as I did not find any way to tell docker not to bind-mount this file. (putting /etc in a volume also does not help).

markus2330:
@Piankero actually you can use this to show error messages (first mount without sudo, then with wrong resolver, ...) and in between echo to say how you interpret the error message and find the solution.

Piankero:
> actually you can use this to show error messages

this is *such* as specific error which users will not run into (who really mounts the hosts file inside docker which has its own DNS infront?)

> first mount without sudo

I would like to avoid this because the non-sudo error message is very verbose

> then with wrong resolver

The concept of a resolver is already hard to grasp for new users. The fact that there are even multiple resolvers is even more confusing

To me it looks like you urgently want to show the hosts plugin which should be well shown if you mount it, get your localhost IP and try to set a wrong IP. If tutorial users run into this problem here themselves they will also google that this is a non-elektra bug in docker.

markus2330:
> this is such as specific error which users will not run into (who really mounts the hosts file inside docker which has its own DNS infront?).

I agree that it is not a common error.

> Its confusing to be put in a tutorial. Also what if docker ever fixes this issue? Then the tutorial is again outdated

The tutorial needs to be recorded from time to time anyway, also because Elektra changes.

> I would like to avoid this because the non-sudo error message is very verbose

This is strange as there was a PR to specifically fix this error. Why didn't you fix it in a way that you want to show it?

> The concept of a resolver is already hard to grasp for new users. The fact that there are even multiple resolvers is even more confusing

As we see in this case the concept is necessary.

> To me it looks like you urgently want to show the hosts plugin which should be well shown if you mount it, get your localhost IP and try to set a wrong IP.

No, my idea was that the tutorial should also guide through the common/usual errors beginners make. And of course to show many different errors, so that you maybe improve them a bit more.

> If tutorial users run into this problem here themselves they will also google that this is a non-elektra bug in docker.

It is not so easy to understand what is really going on there. Partly because the error messages from the filesystem are really crappy (due to limitations of errno EBUSY is reused in rename for different things).

Anyway, it is okay if you type in something like `kdb mount ... -R wresolver # wresolver doesn't do atomic updates, which is needed because /etc/hosts is a mountpoint in docker`.

This is not only an explanation of why `wresolver` is needed but also that Elektra usually does atomic updates.

Piankero:
> This is strange as there was a PR to specifically fix this error. Why didn't you fix it in a way that you want to show it?

The best case would be to only show that the permission was denied but there are so many warnings gathered via the resolvers which might have other useful messages actually inside them for other non-permission related errors. Thats why I let it as it is and think that users will most likely read the yellow bold part first. For a tutorial though this is way too verbose

> As we see in this case the concept is necessary.

Imo not in an asciinema tutorial but I will simply record what you want ...

> No, my idea was that the tutorial should also guide through the common/usual errors beginners make. And of course to show many different errors, so that you maybe improve them a bit more.

Imo this is done via specification mounting and the 2 errors I make.

> Anyway, it is okay if you type in something like

As you wish

Piankero:
Another problem with the new resolver:

```
elektra $ sudo kdb mount --with-recommends /etc/hosts system/hosts hosts -R wres
olver # wresolver doesn't do atomic updates, which is needed because /etc/hosts 
is a mountpoint in docker
elektra $ sudo kdb get system/hosts/ipv4/localhost
127.0.0.1
 Sorry, 4 warnings were issued ;(
	Sorry, module wresolver issued the warning C01200:
	Installation: Could not retrieve from passwd using getpwuid_r. Could not get ALLUSERSPROFILE for spec, using /
	Sorry, module wresolver issued the warning C01200:
	Installation: Could not get ALLUSERSPROFILE environment variable, using /
	Sorry, module wresolver issued the warning C01200:
	Installation: Could not retrieve from passwd using getpwuid_r. Could not get ALLUSERSPROFILE for spec, using /
	Sorry, module wresolver issued the warning C01200:
	Installation: Could not get ALLUSERSPROFILE environment variable, using /
```

markus2330:
> The best case would be to only show that the permission was denied but there are so many warnings gathered via the resolvers which might have other useful messages actually inside them for other non-permission related errors. Thats why I let it as it is and think that users will most likely read the yellow bold part first. For a tutorial though this is way too verbose

If it works, it also should work in the tutorial. We could also consider to not show warnings if `-v` is not given.

> Imo not in an asciinema tutorial but I will simply record what you want ...

It is only necessary because of the Docker bug. Not showing /etc/hosts validation would be wrong though, as this is one of the killer features of validation.

> Imo this is done via specification mounting and the 2 errors I make.

This is difficult to see from the script. In general a little bit repetition does not harm. Maybe you can somehow divide the whole tutorial in 3 parts to make a better structure.

> Another problem with the new resolver:

Please report a bug, this warning seems to be totally unnecessary as you do not even mount spec. Also for absolute paths this warning makes no sense.

The workaround would be to set the environment variable ALLUSERSPROFILE
@Piankero can you check if this works?

If it works, @mpranj can you do this for the docker container?

Piankero:
> If it works, it also should work in the tutorial. We could also consider to not show warnings if -v is not given.

Yes probably this would be an idea but this would somebody else need to do.

> It is only necessary because of the Docker bug. Not showing /etc/hosts validation would be wrong though, as this is one of the killer features of validation.

as you wish ... but these warnings are not very good in the tutorial ...

> The workaround would be to set the environment variable ALLUSERSPROFILE
@Piankero can you check if this works?

I did `export ALLUSERSPROFILE="elektra"`
before the command and it does not work, still 4 warnings

markus2330:
> Yes probably this would be an idea but this would somebody else need to do.

:cry: 

> as you wish ... but these warnings are not very good in the tutorial ...

Yes, they are not so good in general but if they are there...

> I did export ALLUSERSPROFILE="elektra"
before the command and it does not work, still 4 warnings

Something was done wrongly, maybe the shell does not support `export`? Please do not forget to create the issue, an additional problem is that in line 129 the error message seems to be wrong.

mpranj:
`sudo` intentionally does not preserve the environment. There are options to preserve ENV, but of course it seems like this is not available in sudo of alpine linux.

You can work arount it by:
```
elektra $ sudo ALLUSERSPROFILE="elektra" kdb get system/hosts/ipv4/localhost
Did not find key 'system/hosts/ipv4/localhost'
```
or storing it in `/etc/profile`. It already contains some workarounds that I had to implement, so we can add more.

It seems to me that it still does not help. There is no error but localhost is also not found, even though it is there.

Piankero:
> It seems to me that it still does not help. There is no error but localhost is also not found, even though it is there.

So what is the plan? Should I do the workaround?

markus2330:
> or storing it in /etc/profile. It already contains some workarounds that I had to implement, so we can add more.

Yes, sounds good.

mpranj:
Just to repeat: it seems to me that the workaround does not help. It just leads to different errors such as `Did not find key 'system/hosts/ipv4/localhost'` even though it exists in the config.

markus2330:
`ALLUSERSPROFILE="elektra"` obviously is wrong (I do not know how @Piankero came up with this), for me `ALLUSERSPROFILE=/ kdb get system/hosts/ipv4/localhost` works as expected (without warning).

mpranj:
Ok, that works. I pushed it with `elektra/elektra:0.9.1-8`.

```
elektra $ sudo kdb mount --with-recommends /etc/hosts system/hosts hosts -R wresolver
elektra $ sudo kdb get system/hosts/ipv4/localhost
127.0.0.1
```

markus2330:
Thank you so much!

@Piankero https://www.borgbackup.org/demo.html made quite nice tutorials with more text. (It is maybe already too much text and you need to be quite fast to read all of it.)

scripts/docker/alpine/3.10/release.Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+        yajl-dev \
+        yaml-cpp-dev \
+        && rm -rf /var/cache/apk/*
+
+RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
+RUN echo "alias sudo='sudo -i' # in this image we do not need to be root" >> /etc/profile
+RUN echo "export PS1='\u $ '" >> /etc/profile
+RUN echo "export LD_LIBRARY_PATH=/usr/local/lib/elektra/" >> /etc/profile
+RUN echo "export ALLUSERSPROFILE=/" >> /etc/profile
```
markus2330:
```suggestion
RUN echo "export ALLUSERSPROFILE=''" >> /etc/profile
```

an empty string is actually better for the error message:

```
Sorry, 1 warning was issued ;(
        Sorry, module sync issued the warning C01100:
        Resource: Could not open config file //etc/hosts. Reason: Permission denied
Sorry, module hosts issued the error C01100:
Resource: Insufficient permissions to open configuration file //etc/hosts for writing. You might want to retry as root. Reason: Permission denied
```

With my suggestion it is (hopefully, if the syntax is ok) /etc/hosts instead of //etc/hosts.

mpranj: Thank you for the improvement! I pushed it to `elektra/elektra:0.9.1-9`.

mpranj:
@markus2330 do you maybe want to build on top of this and update our asciicast? Does not have to be for this release. I can update the docker image if you have further ideas what to add.

You could either make an issue containing a high level "script" of what should be done or directly record the asciicast. I think you have the best feeling for what we should add into our marketing material.

markus2330:
Thank you, yes, we should record an asciicast based on this. We already had a script which was very good, I think we only need to record.

mpranj:
I updated this PR to the current elektra release and pushed it to docker hub (elektra/elektra:0.9.2-1).

I also took the liberty to upgrade our alpine image versions. Let's first see if this does not break our jenkins CI.

markus2330:
Great! Please mention this also in the release notes! :rocket:

Issue #3308, opened 2019-11-30
Write plugin in rust


darddan:
I've looked at the rust bindings and trying to write a storage plugin. If I'm not overlooking something there's two things stopping me from building it. There's no types for the `Plugin` class and the `from_ptr` function in KeySet (`elektra`) is  private.

I've copied `elektra-sys` to `elektra-plugin-sys`, changed the `wrapper.h` to contain only `#include "kdbplugin.h"` and changed lines 22 and 23 in `build.rs` to:
```
        .whitelist_function("(elektraPlugin).*")
        .whitelist_var("(ELEKTRA_PLUGIN|Plugin).*")
```

With those I was able to create a macro which would generate the needed functions for a plugin (handling the unsafe `elektra-plugin-sys` code). By using that macro, creating a storage plugin in rust can look like this:

```
generate_storage_plugin!(
    elektraPluginSymbol: "exampleplugin",
    getFunction: example_plugin_get, // name of function declared below
    setFunction: example_plugin_set, // name of function declared below
    author: "Author Name <author@email.com>",
    license: "MIT",
    needs: "anything", // Optional field, delete if not needed
    recommends: "anything", // Optional field, delete if not needed
    status: "experimental unfinished",
    metadata: "anything", // Optional field, delete if not needed
    version: "1",
    description: r###"
    Reads and writes the example plugin format
    You can add the content of README.md here
    "###,
);

// Type signature needs to be same as here.
// KeySet and StringKey are from the safe bindings `elektra` crate
fn example_plugin_get(_: &mut KeySet, _parent: &StringKey) -> ElektraReturnCode {
    // Parse code here
    ElektraReturnCode::Success
}

// Type signature needs to be same as here.
fn example_plugin_set(_: &KeySet, _parent: &StringKey) -> ElektraReturnCode {
    // Serialize code here
    ElektraReturnCode::NoUpdate
}
```

Would this be of use?

markus2330:
Great job, looks nice! But this might not be enough, you also need some ways to set error/warnings, access the handle, ... Maybe @PhilippGackstatter knows more about that.

I think we should focus on making the PR for kconfig now.

darddan:
I'm working on the `kconfig` task, but it takes so long with the CI and the docker images so I've tried this in the meantime

markus2330:
Very good idea to do this in the meantime :100:

markus2330:
[...]

PR #3319, opened 2019-12-03
Versioning

markus2330:
## Basics

      contains `_(my name)_`)

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


mpranj:
Thank you so much for working on this. I think this represents pretty well what was informally agreed upon in the Elektra meetings.

markus2330:
@raphi011 @BernhardDenner @PhilippGackstatter Dear maintainers of bindings: I added a small section of how bindings should update their version numbers. Can you please review the changes?

@mpranj I am looking forward to any further feedback :+1:

markus2330:
@kodebach Could you please also take a look, especially that part with symbol versioning. I think we did not discuss this yet.

doc/VERSION.md
```diff
@@ -2,47 +2,84 @@
...
+[Semantic Versioning 2.0.0](https://semver.org/),
+with the extensions explained here.
+
+In `1.0.*` the API and ABI must be always forward-compatible
+and backwards-compatible.
+That means that Elektra's libraries may be upgraded and downgraded
+without any effect on applications, only bug or docu fixes
+are allowed.
+
+In `1.*` the API and ABI must be always forward-compatible,
```
kodebach:
```suggestion
In minor or patch version updates the API and ABI must be always forward-compatible,
```

doc/VERSION.md
```diff
@@ -2,47 +2,84 @@
...
 but not backwards-compatible.
-That means that a program written and compiled against 0.8.0
-compiles and links against 0.8.1. But because it is
+That means that a program written and compiled against 1.0.0
+compiles and links against 1.1.0. But because it is
 not necessarily backwards-compatible a program written
-for 0.8.1 may not link or compile against elektra 0.8.0
-(but it may do when you use the compatible subset, maybe
-with #ifdefs).
+for 1.1.0 may not link or compile against Elektra 1.0.0
```
kodebach: The ABI can be incompatible, even if compiling and linking works (e.g. changes to key names). We should note that. It would be better to specify that the program must also work as originally intended (apart from bug-fixes within Elektra).

doc/VERSION.md
```diff
@@ -59,24 +96,34 @@ a complex topic, so better don't underestimate it, but
 generally said the methods should behave on the same data
 the same way.
 
+As we use [symbol versioning](dev/symbol-versioning.md), the SO_VERSION
+of Elektra always remains the same, even if the major version changes.
+That means, if we rename or remove a function in the `2.0` release,
+applications that linked against Elektra `1.0` can still link against
+Elektra `2.0` but they do not necessarily compile with Elektra `2.0`
```
kodebach: This makes it seem as if we intend to maintain any function we add to the API indefinitely. We should define a limit, for how many versions removed symbols will be supported. e.g. If we remove `x` in 2.0, it is available for dynamic linking until 3.0 is released, but then it is removed entirely.

Of course there can be exceptions, where it makes sense to keep things around longer, but I think we should limit our guarantees.

Also AFAIK "linking against" in this context only includes the runtime part of dynamic linking. Taking an already compiled `.o` file and only calling `ld` won't work AFAIK. I may be wrong, but we should clarify, what exactly "linking against" means.

doc/VERSION.md
```diff
@@ -59,24 +96,34 @@ a complex topic, so better don't underestimate it, but
...
-
-Note: All 3 versioning infos are handled separately!
-
-http://www.gnu.org/software/libtool/manual/libtool.html#Versioning
-http://semver.org/
-
-> TODO write about SO_VERSION and rest version
+## Bindings
+
+Bindings are in general tied to Elektra's version number except
```
kodebach: Semantic versioning allows to add a build metadata suffix. We could use that for the binding version, e.g. `1.2.3+rust1` is the first version of the Rust binding for `1.2.3`, `1.2.3+rust2` is a second version that fixes some bugs only in the binding, etc.

Otherwise, I would clarify that the patch version is essentially language specific. That this is the case is easy to see:
1. `1.0.0` of the C API is released
2. `1.0.0` of the Rust API is released at the same time
3. `1.0.1` of the Rust API is released with a bugfix
4. `1.0.2` of the Rust API is released with another bugfix
5. `1.0.1` of the C API is released
6. `1.0.3` of the Rust API is released to incorporate the `1.0.1` C API
7. `1.0.2` of the C API is released
8. `1.0.4` of the Rust API is released to incorporate the `1.0.2` C API

From the version number alone, there is no way to know which C bugfixes are included in which Rust version. The intuitive rule, that all smaller patch versions are included doesn't hold because of 6 and 7.

doc/VERSION.md
mpranj: This sounds like very strong guarantee.

Strictly speaking if nothing else is specified, the implementation is the specification (as suggested in the Behavior chapter of this PR). Thus you cannot patch anything without changing the "original intention".

I think we would only be able to deliver a weak guarantee here that "we will try not to break" things intentionally. The perfect guarantee would be something we can automatically test.

doc/VERSION.md
```diff
@@ -2,47 +2,84 @@
...
+
+The following order defines the behavior of Elektra:
+
+1. [ABI](/tests/abi) test cases
+2. Non-ABI test cases (including Shell Recorder)
+3. The [API documentation](https://doc.libelektra.org/api/latest/html/)
+4. The [man pages](/doc/help)
+5. Tutorials (excluding Shell Recorder)
+6. Examples
+7. Implementation
```
kodebach: A few more comments on this list:

As mentioned below, IMO the implementation should not be part of the list. AFAIK this list defines the *intended behaviour* of Elektra. Putting the implementation on this list makes all bugs part of the intended behaviour. The implementation is Elektra's *actual behaviour*, which should of course match the *intended behaviour*, but not be part of its definition.

In other words, any behaviour that is undocumented (docu or examples) and untested (tests), outside of its implementation should be considered a private implementation detail for which we make no guarantees.

The points regarding test cases should be restricted to the parts of the tests that test public API function. For example there are test cases, which test the internal key unescape and key escape functions. Since these are not part of the public API, we should not make any guarantees for their behaviour (as long as public functions remain unaffected).

doc/VERSION.md
kodebach: I missed the the implementation is on the list that specifies behaviour. I think that is a big mistake.  The implementation should never be part of the what defines the intended behaviour. This even makes bug-fixes impossible, since the bug is part of the implementation.

> "we will try not to break" things intentionally

Is very weak... Guaranteeing that compiling and linking works is fine, since it can be tested. But we should at least mention that this is not sufficient for a version to be ABI compatible.

doc/VERSION.md
mpranj: As a customer of Elektra (a developer of an elektrified program) I would be most interested in the API docu. I would usually not read through the ABI test cases to understand the intention of the API.

Internally this list makes sense, as the implementation obviously defines the actual behavior. Maybe I just misinterpreted the intention of the list.

doc/VERSION.md
markus2330: Thank you very much for your excellent feedback. It was incorporated in 03b6eb64bc24d172e1deb4c612ff0f83908634c9

> Semantic versioning allows to add a build metadata suffix. We could use that for the binding version, e.g. 1.2.3+rust1 is the first version of the Rust binding for 1.2.3, 1.2.3+rust2 is a second version that fixes some bugs only in the binding, etc.

We discussed this in #3308. Unfortunately the build metadata is unsuitable for most of the changes (more than rebuilding or docu) and there is no alternative in semver. It seems like semver simply did not consider downstream versions.

PhilippGackstatter: I think that making the patch level binding-specific is the best solution. It is not ideal while we're on `0.9.x` but once `1.0` is out this will not matter as much.
I would also state what Elektra version the binding is currently compatible with in its Readme to make it explicit for users.

doc/VERSION.md
markus2330: > The implementation is Elektra's actual behaviour, which should of course match the intended behaviour, but not be part of its definition.

You are completely right. I fixed it in d13860247e5dd85aba9cdc107b228987fc5cf609. I had a different starting sentence before (about where to fix the bugs, and then implementation is on the list).

> The points regarding test cases should be restricted to the parts of the tests that test public API function. For example there are test cases, which test the internal key unescape and key escape functions. Since these are not part of the public API, we should not make any guarantees for their behaviour (as long as public functions remain unaffected).

Difficult to say without any actual bug at hand. But what is clear that these test cases are subject to be changed if the documentation says otherwise. If the documentation said nothing before, we might decide to fix the documentation. And then, according to this proposed list, the docu would win against these test cases anyway.

doc/VERSION.md
markus2330: I changed at both places now, it is best if you look again if it suits now.

I think it is good and needed to provide strong guarantees. Obviously we are human and might fail from time to time.

That the implementation is on the list was an error on my part. Even alternative implementations of Elektra should not look at the implementation but instead open an issue if we did not test and document something properly.

doc/VERSION.md
markus2330: (GitHub did not show the comment of @mpranj thats why I answer now)

> As a customer of Elektra (a developer of an elektrified program) I would be most interested in the API docu. I would usually not read through the ABI test cases to understand the intention of the API.

Yes, I agree.

> Internally this list makes sense, as the implementation obviously defines the actual behavior. Maybe I just misinterpreted the intention of the list.

No, you interpreted correctly and the order in the list is up to discussion. We can swap these points. My rationale was that it is very easy to verify if the implementation behaves according the test cases and very difficult (impossible) to verify if it behaves according to sentences in natural languages. Thus I think it is more practical to adjust the sentences and not the test cases.

Of course, this is not so nice for application developers that implemented applications according to wrong documentation. We can improve the situation by including more test cases in the documentation.

Another reason for as-is is that applications with special requirements can, if the developers are serious about it, contribute test cases. Then they have the guarantee that we will only adapt docu and not their test cases (which would make their application fail).

markus2330:
@Mistreated do you know why the Jenkins build does not show to be completed (although it was running the whole night) and the link does not go to some job but to the general page https://build.libelektra.org/blue/organizations/jenkins/libelektra/activity?

mpranj:

mpranj:
This was my fault, you pushed while I was changing the config of our docker hub so I disabled all jenkins executors. Apparently, that is what happens when no executor is available.

markus2330:
@mpranj I added ready to merge :wink:

PR #3324, opened 2019-12-04
Cache fix synopsis

### LINKED ISSUE #3301: cache: fix tool synopsis
Synopsis was not updated, `kdb cache default` is missing here but available in the man page.

https://github.com/ElektraInitiative/libelektra/blob/b04c7f85a902fd247b628d88d752370d33dd034a/src/tools/kdb/cache.hpp#L26-L33

mpranj:
Closes #3301.

### DIFF
--- a/src/tools/kdb/cache.hpp
+++ b/src/tools/kdb/cache.hpp
@@ -36,9 +36,9 @@ public:
 	virtual std::string getLongHelpText () override
 	{
 		return "This command is used to enable or disable the cache and to revert\n"
-			"to the default settings. The default settings will let the system\n"
-			"decide whether to use the cache or not. The clear command will\n"
-			"remove the generated cache files in a safe way.\n";
+		       "to the default settings. The default settings will let the system\n"
+		       "decide whether to use the cache or not. The clear command will\n"
+		       "remove the generated cache files in a safe way.\n";
 	}
 
 	virtual int execute (Cmdline const & cmdline) override;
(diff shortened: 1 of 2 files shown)


src/tools/kdb/cache.hpp
```diff
@@ -25,17 +25,20 @@ class CacheCommand : public Command
...
 	virtual std::string getShortHelpText () override
 	{
-		return "Enable, disable or clear the cache.";
+		return "Enable, disable, clear the cache or revert to default.";
 	}
 
 	virtual std::string getLongHelpText () override
 	{
-		return "";
+		return "This command is used to enable or disable the cache and to revert\n"
```
markus2330: Afaik the getLongHelpText is no longer used. So it is better to leave it empty (or even better: remove the whole function in all the classes).

markus2330:
Thank you for the PR!

src/tools/kdb/cache.hpp
mpranj: My `kdb` tool displays the long help text. I can remove it in all classes, but likely not today.

Many tools do have some information in the long help text. Do we just throw that out, or is the idea to merge everything into one single help text, be it long or short?

markus2330: Yes, you are right. The text is shown when typing in wrong arguments or similar.

So then leave it as is, as it would be a lot of work to unify all these texts and there would be only little gain.

PR #3328, opened 2019-12-09
Plugins: Use Explicit Typing

sanssecours:


src/plugins/augeas/augeas.c
```diff
@@ -323,9 +323,9 @@ static char * loadFile (FILE * fh)
 
 	if (fileSize > 0)
 	{
-		content = elektraMalloc (fileSize * sizeof (char) + 1);
+		content = elektraMalloc ((size_t) fileSize * sizeof (char) + 1);
```
markus2330: Didn't us the compiler actually tell us about a potential bug here? If the fileSize is larger than SIZE_MAX, the conversation would do something wrongly.

Not really important (platforms with small SIZE_MAX are exotic but the standard actually allows it to be as small as 65535) but maybe it is better to use the second branch in this case (then we would at least not write into non-allocated memory).

If it is not easy to fix, maybe you can add a static assert so that someone who will port to such platforms finds all places more easy.

sanssecours: Thank you for the hint. The code now checks that the maximum allocation size is not exceeded (, assuming that the size of `char` is one byte on the target platform). If you are not happy with the proposed solution, then please just add another comment.

mpranj: C99 dictates it is `1`. I think we have many other places in Elektra where it is assumed that `sizeof (char) == 1` holds. It would not hurt the runtime to put `sizeof (char)` instead of `1`, but don't spend time on fixing this. There are more important issues.

sanssecours: Since the code:

```c
if (fileSize > 0 && (unsigned long long) fileSize < SIZE_MAX)
```

introduces a [warning from LGTM](https://lgtm.com/projects/g/ElektraInitiative/libelektra/rev/pr-b7054743afcf920ab3f6bbdce9afd95f9290c464):

> Comparison is always true because fileSize <= 9223372036854776000.

, I wanted to solve the problem using `static_assert`:

```c
static_assert (SIZE_MAX >= LONG_MAX,
	       "the augeas plugin requires that `size_t` is able to represent all positive values of the datatype `long`");
```

. This does not work though, since [`static_assert` requires C11](https://en.cppreference.com/w/c/error/static_assert). For now I removed the updates to handle large file sizes. This means the code does not handle situations in which the return value of `ftell` is larger than the maximum size of `size_t` incorrectly, just like before.

mpranj: I also ran into a scenario, where `static_assert ()` would have been great to statically guarantee some assumptions for `mmapstorage`. Since the checks were important to avoid crashes/inconsistencies I implemented them as runtime checks in the `kdbOpen()` function of the plugin.

You can decide if such a check is needed.

sanssecours: > You can decide if such a check is needed.

I filled [an issue](https://github.com/ElektraInitiative/libelektra/issues/3332) instead. Let us hope some other person is more interested in fixing the underlying problem 😊.

src/plugins/augeas/CMakeLists.txt
```diff
@@ -28,6 +28,11 @@ if (DEPENDENCY_PHASE)
 	set (AUGEAS_PLUGIN_LIBRARIES ${LIBAUGEAS_LIBRARIES} ${LIBXML2_LIBRARIES})
 endif ()
 
+# ——————————————————————————————————————————————————————————
+# TODO: Remove this after we enabled `-Wconversion` globally
+set_source_files_properties (aug.h augeas.c PROPERTIES COMPILE_FLAGS "-Wconversion")
```
mpranj: Can this also be set for the other plugins or maybe even globally already?

mpranj:
Thank you for the improvements!

src/plugins/augeas/CMakeLists.txt
sanssecours: I already tried that, but in my opinion it was just too much work to do it at once. That is why I decided to introduce the compile flag in multiple smaller pull requests.

PR #3329, opened 2019-12-09
Various Minor Enhancements

sanssecours:


src/bindings/swig/python/CMakeLists.txt
```diff
@@ -1,3 +1,16 @@
+# ~~~
+# Disable warnings about unset CMake policy.
+# TODO: Remove the calls to `cmake_policy` after we upgraded to new behavior.
```
mpranj: Can you please create an issue as well such that it's harder to forget about this?

sanssecours: I opened issue #3330 to keep track of the problem. I do not think we will forget about this to-do item though, since we will not be able to build the code of the Python binding after the CMake developers remove the option to use the old behavior for policy `CMP0078`.

mpranj:
Thank you so much for the improvements!

markus2330:
Thank you, looks great!

PR #3333, opened 2019-12-13
CMake: Require CMake 3.4

sanssecours:
# Description

While I think requiring CMake `3.7` (current [CMake version from Debian oldstable](https://packages.debian.org/stretch/cmake)) should be fine, I found no code that specifically targets CMake versions between CMake `3.4` and CMake `3.7`.


sanssecours:

doc/news/_preparation_next_release.md
```diff
@@ -151,12 +151,16 @@ you up to date with the multi-language support provided by Elektra.
 
   plugin any more. After this update, the code compiles without any warnings, even though we now use the compiler switch `-Wconversion`. _(René Schwaiger)_
 
+### Support
+
+- Debian 9 “stretch” (oldstable) is now the oldest supported platform. _(René Schwaiger)_
```
mpranj: Debian 8 Jessie is still supported until [June 2020](https://wiki.debian.org/LTS) and it uses cmake 3.0. Are there any benefits from removing it right now?

sanssecours: According to the link above the distro is not supported by the Debian security team (and hence Debian) anymore:

> Debian LTS is **not handled by the Debian security team**, but by a separate group of volunteers and companies interested in making it a success.

While I think the benefit of not supporting CMake `3.3` and below is minimal (a little bit less CMake code), I also do not think it makes sense to officially support software that is over five years old.

mpranj: You're right, let's keep the PR as it is!

mpranj:
Thank you so much for cleaning this up!

markus2330:
Also a big thanks from my side :heart:

PR #3336, opened 2019-12-14
Check Formatting: Undo Formatting Changes at End

sanssecours:


tests/shell/check_formatting.sh
```diff
@@ -87,4 +87,9 @@ if [ -n "$git_diff_output" ]; then
 
 fi
 
+# We remove working directory and staged changes. This operation should only remove formatting modifications, since we already checked for a
+# clean state at the beginning of this script. We reset all file changes, since otherwise the test will always succeed on the second
+# invocation. See also: http://issues.libelektra.org/3281
+git reset --hard
```
mpranj: Would it make sense to stash the changes such that they could be applied if that is wanted?

I don't know whether it would create too much garbage in git.

sanssecours: I also thought that `git stash` might create too much useless data. That is why I used `git reset`. Stashing the changes it the much safer option though, so it definitely makes sense to do that instead. I modified the pull request accordingly.

mpranj:
Thank you! :rocket:

markus2330:
Thank you for fixing and merging this :heart:

PR #3337, opened 2019-12-14
Style: Reformat Code With cmake-format 0.6.3

sanssecours:


mpranj:
Thank you! Do you want this in immediately or shortly before the next release?

sanssecours:
> Do you want this in immediately or shortly before the next release?

I would prefer we merge this as soon as possible. Since the formatting changes between version `0.6` and version `0.6.3` are relatively minor – only six CMake files were changed – I do not think that should be that big of a problem. Of course, If there are bigger pull requests that are “ready to merge” soon, I have no problem whatsoever if you merge those first.

mpranj:
Thank you so much for maintaining it!

I agree that it should not be a huge problem. I just asked because that's how it was previously done.

markus2330:
We only updated tools shortly before releases as the constant updating of the tools (or the docker image to reformat) is a quite big barrier for Elektra developers (at least it was for me, as I work on several computers). Furthermore, requiring people to rebase PRs just because of reformatting is not so nice. After releases, everything needs to be rebased anyway, so there is no issue then.

You do not need to follow this advice. As I currently do not change code, it doesn't affect me.

PR #3342, opened 2020-01-27
Cirrus: Fix macOS Build Jobs

sanssecours:

PR #3344, opened 2020-01-28
Yan LR: Support ANTLR 4.8

sanssecours:
This pull request is based on PR #3342, since otherwise some build jobs would fail.


mpranj:
Thank you so much!

PR #3346, opened 2020-01-29
YAMBi: Remove Plugin

sanssecours:
This pull request contains an updated version of PR #2939. Since I do not intend to develop the YAMBi plugin any further, I think it makes sense to remove it. For more information about why I would prefer to improve Yan LR instead, please take a look at my [Master Thesis](https://github.com/sanssecours/Configuration-File-Parsing/releases).

PR #3347, opened 2020-01-29
Various Minor Enhancements

sanssecours:
This pull request also contains a news item about my finished thesis as suggested in PR #3346.


mpranj:
Thank you so much!

PR #3349, opened 2020-02-05
Cirrus: Update D-Bus Startup Command on macOS

sanssecours:
This pull request should fix the build problems of the Cirrus macOS build jobs.


mpranj:
Thank you for fixing this so quickly!

PR #3350, opened 2020-02-23
cirrus: remove python2 installation

mpranj:
.

PR #3351, opened 2020-02-24
CI: Fix macOS Build Jobs

sanssecours:
This pull request is partially based on PR #3350. As far as I know, it should fix the current build problems on both Cirrus and Travis. It is a little bit less intrusive than PR #3350, since afterwards `ronn` should still build the man pages in the Cirrus macOS build jobs. This has the advantage that the uninstall check:

https://github.com/ElektraInitiative/libelektra/blob/5f742352a5bc1a6a25d8e153cf5c30203fbcc89a/.cirrus.yml#L197-L205

makes sure that uninstalling also removes all man pages. Please feel free to close this PR in favor of pull request #3350 if you like.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -153,7 +153,6 @@ mac_task:
       brew update # Work around for [Homebrew Services issue 206](https://github.com/Homebrew/homebrew-services/issues/206)
       brew services start dbus
     - | # Install Python
-      brew install python@2; brew link --overwrite python@2
       brew install python || brew upgrade python
     - | # Install Python packages
       pip install cmake-format[yaml]==0.6.3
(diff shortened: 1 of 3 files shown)


mpranj:
Thank you for working on this @sanssecours . I would merge this as it is less invasive as you said and #3350 has some regressions. Can we reopen and merge or was there a problem with the PR?

sanssecours:
>  Can we reopen and merge or was there a problem with the PR?

Of course we can 😊. I just thought that it would make sense to close it, since PR #3350 seemed to already contain all of the changes of this pull request and some additional improvements.

mpranj:
Thank you for the fixes. This rather stable PR makes more sense right now compared to my chnages.

PR #3354, opened 2020-02-25
libopts extensions

### LINKED ISSUE #3232: opts: args=remaining with non-array
Currently there is the limitation that args=remaining can only be for arrays (see src/libs/opts/opts.c 317).

For many applications, only one argument is needed, e.g. `kdb ls`.

Is it easily possible to lift this restriction? Of course, the command-line parser then should fail with an error if more than one argument is passed.

This would simplify writing the specifications for all the `kdb` commands for @ulrikeschaefer.

E.g. `kdb ls` could be like:

```ini
[help]
opt=H
opt/long=help

[...]

[path]
args = remaining
completion/shell = kdb complete
```

`kdb complete` would be called to complete the one argument of `kdb ls`.

@kodebach what do you think? Or is it better if we allow to refer to specific args, like `args = 1` to assign the first argument to `path`? This might be a bigger change, though?

### LINKED ISSUE #3341: [WIP] opts: Specification for sub-commands
**This proposal is still a work in progress. The specification is not final, but comments are welcome**

Currently `libopts` can be used to implement an application with multiple sub-commands (e.g. `kdb`), but it is a bit tedious and has to be done manually. This is a proposal for a specification that would allow `elektraGetOpts` to parse sub-commands.

The main idea is the following:

_If a key is marked with the `opt/command` (name not final) metakey, `elektraGetOpts` will **not** immediately process keys below it._

The keys are only processed, when the `elektraGetOpts` determines that the command was actually invoked. The example below illustrates this main rule.

When `elektraGetOpts` processes such a command-key `ck` one of three things can happen:
- A sub-command of the `ck` was invoked. The value of `ck` will be `/[subcmd]`, where `subcmd` is the basename of the sub-command's command-key.
- `ck` was invoked but none of its sub-commands (this would be `get` in `kdb get`). We call it the terminal sub-command. Here `ck`'s value will simply be `/`.
- `ck` was not invoked at all (e.g. the `set` command-key when `kdb get` is called). `ck` will be set to an empty string.

Here is an example specification:

```ini
[kdb]
opt/command = kdb

[kdb/printversion]
type = boolean
opt = v
opt/long = version
opt/arg = none

[kdb/get]
opt/command = get

[kdb/get/verbose]
type = boolean
opt = v
opt/long = verbose
opt/arg = none

[kdb/get/keyname]
type = string
args = single

[kdb/set]
opt/command = get

[kdb/set/verbose]
type = boolean
opt = v
opt/long = verbose
opt/arg = none
```

To illustrate the behaviour we will look at a few sample invocations:

- `kdb -v`: Keys below `kdb/get` and `kdb/set` are not touched. The result is:
  - `kdb = /`
  - `kdb/printversion = 1`
  - `kdb/get = ""`
  - `kdb/set = ""`
- `kdb get -v name`: Keys below `kdb/set` are not touched. The result is:
  - `kdb = /get`
  - `kdb/get = /`
  - `kdb/get/verbose = 1`
  - `kdb/get/keyname = name`
  - `kdb/set = ""`
- `kdb -v set -v`: Keys below `kdb/get` are not touched. The result is:
  - `kdb = /set`
  - `kdb/printversion = 1`
  - `kdb/set = /`
  - `kdb/set/verbose = 1`
  - `kdb/get = ""`

---

Additional rules:

- Sub-command processing only happens when the parent key has the `opt/command` metakey.
- Sub-command processing implies the `posixly` version of `elektraGetOpts`, i.e. **all** options must be specified before non-options.
- If sub-command processing is not active, **no key**  below the parent (no matter how deep) may have the `opt/command` metakey.
- Sub-command may be arbitrarily deep below there parent command. They don't have to be directly below. e.g. `kdb/commands/get` can be a sub-command of `kdb` (if `kdb/commands` is not a command itself).
- Below any given key `k` (no matter how deep) there may only be either keys with `args` or with `opt/command` metadata never both. Because of the main rule above, this doesn't apply for keys below sub-commands. e.g. in the example above `kdb/get/keyname` is allowed, because `kdb/get` has no sub-commands, but a similar `kdb/keyname` would not be allowed.

kodebach:
Closes #3232
Closes #3341

## Basics

      contains `_(my name)_`)
- [x] The buildservers are happy.

## Checklist

      (not in the PR description)

## Review

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -92,11 +92,9 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
 - <<TODO>>
 - <<TODO>>
 
-### <<Library1>>
+### Opts
 
-- <<TODO>>
-- <<TODO>>
-- <<TODO>>
+- Add Sub-commands _(Klemens Böswirth)_ # FIXME
 
 ### <<Library2>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -94,7 +94,10 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
 
 ### Opts
 
-- Add Sub-commands _(Klemens Böswirth)_ # FIXME
+- The library function `elektraGetOpts` now supports sub-commands. For example `add` is a sub-command in `git add`, and
+  interprets `-p` differently from `git`: `git -p add` is `git --paginate add`, but `git add -p` is `git add --patch`.
+  For more information take a look at the [tutorial for command-line-options](../tutorials/command-line-options.md).
+  By extension this functionality is also available via the `gopts` plugin. _(Klemens Böswirth)_
 
 ### <<Library2>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -77,6 +77,8 @@ The following section lists news about the [modules](https://www.libelektra.org/
 - The plugin now supports an offset into `argv` given by the `/offset` config key. When `/offset` is set, `gopts` will
   ignore a number of arguments at the start of `argv`. This can be used in e.g. python scripts to ignore the interpreter
   arguments. _(Klemens Böswirth)_
+- `gopts` now also writes help message into the key `proc/elektra/gopts/help/message` in addition to setting
+  `proc/elektra/gopts/help = 1`. This is also useful in non-C/C++ environments. _(Klemens Böswirth)_
 - `gopts` is also affected by the changes and improvements to the `opts` library outlined below.
 
 ## Libraries
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -98,6 +98,10 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
   interprets `-p` differently from `git`: `git -p add` is `git --paginate add`, but `git add -p` is `git add --patch`.
   For more information take a look at the [tutorial for command-line-options](../tutorials/command-line-options.md).
   By extension this functionality is also available via the `gopts` plugin. _(Klemens Böswirth)_
+- The generated help message was improved. It now also gives details about parameter arguments, sub-commands and
+  environment variables in addition to the existing support for option arguments. This also means that it is no longer
+  possible to have multiple keys with the `args=remaining` metadata (because their `opt/help` may not bep the same).
+  _(Klemens Böswirth)_
 
 ### <<Library2>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -100,7 +100,7 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
   By extension this functionality is also available via the `gopts` plugin. _(Klemens Böswirth)_
 - The generated help message was improved. It now also gives details about parameter arguments, sub-commands and
   environment variables in addition to the existing support for option arguments. This also means that it is no longer
-  possible to have multiple keys with the `args=remaining` metadata (because their `opt/help` may not bep the same).
+  possible to have multiple keys with the `args=remaining` metadata (because their `opt/help` may not be the same).
   _(Klemens Böswirth)_
 
 ### <<Library2>>
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -72,6 +72,13 @@ The following section lists news about the [modules](https://www.libelektra.org/
 - The CMake code of the plugin does not print error messages produced by the tool `ldd` any more. _(René Schwaiger)_
 - The plugin now also supports ANTLR 4.8. _(René Schwaiger)_
 
+### GOpts
+
+- The plugin now supports an offset into `argv` given by the `/offset` config key. When `/offset` is set, `gopts` will
+  ignore a number of arguments at the start of `argv`. This can be used in e.g. python scripts to ignore the interpreter
+  arguments. _(Klemens Böswirth)_
+- `gopts` is also affected by the changes and improvements to the `opts` library outlined below.
+
 ## Libraries
 
 The text below summarizes updates to the [C (and C++)-based libraries](https://www.libelektra.org/libraries/readme) of Elektra.
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -94,8 +94,11 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
 
 ### Opts
 
-- The library function `elektraGetOpts` now supports sub-commands. For example `add` is a sub-command in `git add`, and
-  interprets `-p` differently from `git`: `git -p add` is `git --paginate add`, but `git add -p` is `git add --patch`.
+- The library function `elektraGetOpts` now supports sub-commands.
+  Sub-commands are best explained by looking at an application that uses them, like `git`.
+  For example `add` is a sub-command in `git add`, and interprets `-p` differently from `git`:
+  `git -p add` is `git --paginate add`, but `git add -p` is `git add --patch`.
+  `elektraGetOpts` now implements this notion of sub-commands.
   For more information take a look at the [tutorial for command-line-options](../tutorials/command-line-options.md).
   By extension this functionality is also available via the `gopts` plugin. _(Klemens Böswirth)_
 - The generated help message was improved. It now also gives details about parameter arguments, sub-commands and
--- a/doc/tutorials/command-line-options.md
+++ b/doc/tutorials/command-line-options.md
@@ -208,6 +208,7 @@ then repeatedly look at the current key's value, append that to the current key 
 was the empty string `""`. Each of the examined keys corresponds to one of the sub-commands in the invocation and the keys
 directly below those, contain the relevant options (and for the last sub-command also parameters).
 
+The C code for this example is located in examples/optsCommands.c 
 ## Help Message
 
 When the help option `--help` is encountered in `argv`, `elektraGetOpts` only reads the specification, but does not
--- a/doc/tutorials/command-line-options.md
+++ b/doc/tutorials/command-line-options.md
@@ -208,7 +208,8 @@ then repeatedly look at the current key's value, append that to the current key 
 was the empty string `""`. Each of the examined keys corresponds to one of the sub-commands in the invocation and the keys
 directly below those, contain the relevant options (and for the last sub-command also parameters).
 
-The C code for this example is located in examples/optsCommands.c 
+The C code for this example is located in `examples/optsCommands.c`.
+
 ## Help Message
 
 When the help option `--help` is encountered in `argv`, `elektraGetOpts` only reads the specification, but does not
--- a/doc/tutorials/command-line-options.md
+++ b/doc/tutorials/command-line-options.md
@@ -110,7 +110,7 @@ assigned as follows:
 This is because the options that `git` understands are separate from the options that its sub-command `add` knows. However,
 the option `-p` is understood by both. In `git` it is short for `--paginate` and in `add` it is short for `--patch`.
 
-The first thing you need to know about sub-commands is that they imply `posixly = 1` mode. This means **all** options must
+The first thing you need to know about sub-commands is that they turn on `posixly = 1` mode. This means **all** options must
 be given before any non-option arguments (such as parameters or sub-commands). Otherwise, we couldn't distinguish between
 `git -p add` and `git add -p`.
 
--- a/doc/tutorials/command-line-options.md
+++ b/doc/tutorials/command-line-options.md
@@ -79,7 +79,9 @@ then there must also be keys for all integers `0 <= X < N` with `args=indexed` a
 use `args/index=0` and `args/index=2` without `args/index=1`.
 
 Combining `args=indexed` and `args=remaining` in the same specification (on different keys) is also possible. The key with
-`args=remaining` will only contain those elements not used via `args=indexed`.
+`args=remaining` will only contain those elements not used via `args=indexed`. E.g. if there are keys with `args/index=0`
+and `args/index=1` then the `args=remaining` array will start with the third (index 2) parameter argument. Note however,
+the `args=remaining` array **always** starts with index `#0`, even if it doesn't contain the first parameter argument.
 
 ### Example
 
@@ -106,13 +108,18 @@ assigned as follows:
 
 ## Sub-Commands
 
-`elektraGetOpts` also supports sub-commands. For example calling `git -p add` and `git add -p` result in different behavior.
-This is because the options that `git` understands are separate from the options that its sub-command `add` knows. However,
-the option `-p` is understood by both. In `git` it is short for `--paginate` and in `add` it is short for `--patch`.
+`elektraGetOpts` also supports sub-commands. Explaining sub-commands is easiest through the help of an example: `add` and
+`commit` are both sub-commands of `git`, since we can call `git add` and `git commit` and they do entirely different things.
+The most important impact of using sub-commands is their effect on option arguments. For example calling `git -p add` and
+`git add -p` result in different behavior, since the `-p` option is interpreted differently. The options that `git`
+understands are separate from the options that its sub-command `add` knows. However, the option `-p` is understood by both.
+In `git` it is short for `--paginate` and in `add` it is short for `--patch`.
 
-The first thing you need to know about sub-commands is that they turn on `posixly = 1` mode. This means **all** options must
-be given before any non-option arguments (such as parameters or sub-commands). Otherwise, we couldn't distinguish between
-`git -p add` and `git add -p`.
+An important thing to know about sub-commands in `elektraGetOpts` is that they turn on `posixly = 1` mode. This means **all**
+options for a specific sub-command must be given before any non-option arguments (such as parameters or sub-commands).
+Otherwise, we couldn't distinguish between `git -p add` and `git add -p`. In other words an option argument is always assigned
+to the first sub-command to its left. Any element of `argv` that is not the argument for an option argument, either switches to
+a new sub-command or is the start of the parameter arguments.
 
 A sub-command is created by specifying `command` on a key. To enable sub-command processing the parent key of the whole
 specification must have `command` set to an empty string. All keys marked with `command` **directly** below another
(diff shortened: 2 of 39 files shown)


mpranj:
Thank you for working on this proposal!

markus2330:
Thank you so much for working on this! This will help @ulrikeschaefer!

It would be amazing if you can also give a very short example of how to use your command-line parsing from Python. Then @ulrikeschaefer could check if the command-line completion and parsing do the same thing.

kodebach:
> give a very short example of how to use your command-line parsing from Python

If someone adds `elektraGetOpts` to the python binding (not sure how to do that, AFAIK we'd need a C++ binding first), it should be trivial to call `elektraGetOpts`.

markus2330:
> If someone adds elektraGetOpts to the python binding (not sure how to do that, AFAIK we'd need a C++ binding first), it should be trivial to call elektraGetOpts.

I was thinking about a solution with kdbEnsure/kdbOpen. (Or also manually mounting the opts plugin should be okay for @ulrikeschaefer)

kodebach:
> I was thinking about a solution with kdbEnsure/kdbOpen.

`kdbEnsure` is also not part of the Python binding (it is however part of the C++ binding). Once `kdbEnsure` is accessible (in whatever language you are using), it is just a matter of calling `kdbEnsure` with a contract that contains the key `system/elektra/ensure/plugins/global/gopts = mounted` (between `kdbOpen` and `kdbGet`). Then `gopts` will call `elektraGetOpts` during `kdbGet`.

After that it is up to you what you want to do. As stated in the documentation, the recommended way is:

0. Start with the parent key as `K`
1. Look at the value of `K`. If it is an empty string, go to 4; else go to 2.
2. Run what ever code is relevant for the command associated with `K`, keeping in mind that a sub-command of `K` was invoked (e.g. do the `git -p` part of `git -p add -p`)
3. Add the value of `K` as a basename to `K` and go to 1
4. The command associated with `K` the actual sub-command that was invoked. Run the relevant code. (e.g. do the `add -p` part of `git -p add -p`)

If you think it is really necessary, I can add a python example, but I think description above and the C samples I'll add should be sufficient.

mpranj:
@ulrikeschaefer regarding the python bindings for this: seems that kdbEnsure was automatically wrapped by swig so you can likely already use it with the python bindings. See #3365 for a simple test case and usage example.

kodebach:

doc/tutorials/command-line-options.md
```diff
@@ -63,12 +63,151 @@ into the array.
...
+If an unknown sub-command is encountered without an `args` key, an error is returned.
+
+All of this is best understood with an example:
+
+```ini
+[kdb]
+command = ""
+
+[kdb/printversion]
+type = boolean
```
markus2330: Some description texts here might be helpful. Ideally, this would be identical to the KeySet used in the example below.

examples/optsCommands.c
```diff
@@ -0,0 +1,161 @@
...
+	{
+		fprintf (stderr, "ERROR: %s\n", keyString (keyGetMeta (errorKey, "error/reason")));
+		keyDel (errorKey);
+		ksDel (ks);
+		return EXIT_FAILURE;
+	}
+
+	if (result == 1)
+	{
+		char * help = elektraGetOptsHelpMessage (errorKey, NULL, NULL);
```
markus2330: As examples can have more verbose comments, I would write here that --help will make this code execute.

examples/optsCommands.c
```diff
@@ -0,0 +1,161 @@
...
+		return EXIT_SUCCESS;
+	}
+
+	lookup = ksLookupByName (ks, BASE_KEY "", 0);
+	if (lookup != NULL && elektraStrCmp (keyString (lookup), "getter") == 0)
+	{
+		lookup = ksLookupByName (ks, BASE_KEY "/getter/keyname", 0);
+		if (lookup == NULL || strlen (keyString (lookup)) == 0)
+		{
+			printf ("report the error 'missing parameter: keyname'\n");
```
markus2330: how can I come here? When I execute `bin/optsCommands get`, already the ERROR: path above will be executed (ERROR: Expected at least 1 non-option arguments, but only got 0).

examples/optsCommands.c
```diff
@@ -0,0 +1,161 @@
...
+			keyDel (errorKey);
+			ksDel (ks);
+
+			return EXIT_SUCCESS;
+		}
+		printf ("read the key '%s'\n", keyString (lookup));
```
markus2330:
```suggestion
		printf ("get the key '%s'\n", keyString (lookup));
```

examples/optsCommands.c
```diff
@@ -0,0 +1,161 @@
...
+		lookup = ksLookupByName (ks, BASE_KEY "/setter/value", 0);
+		if (lookup == NULL)
+		{
+			printf ("report the error 'missing parameter: value'\n");
+			keyDel (errorKey);
+			ksDel (ks);
+
+			return EXIT_SUCCESS;
+		}
+		printf ("write the value '%s' to the key '%s'\n", keyString (lookup), keyname);
```
markus2330:
```suggestion
		printf ("set the key '%s' with the value '%s'\n", keyname, keyString (lookup));
```

examples/optsCommands.c
```diff
@@ -0,0 +1,161 @@
...
+
+		lookup = ksLookupByName (ks, BASE_KEY "/setter/verbose", 0);
+		if (lookup != NULL && elektraStrCmp (keyString (lookup), "1") == 0)
+		{
+			printf ("print where the key value is stored now\n");
```
markus2330: `bin/optsCommands set abc -v` is not coming here, what I am doing wrong?

And `bin/optsCommands set -v abc` fails with `ERROR: Expected at least 2 non-option arguments, but only got 1` which is confusing for me.

Ideally, the shell recorder would check if the example program accepts arguments as expected.

examples/optsCommands.c
```diff
@@ -0,0 +1,161 @@
...
+		KeySet * dynamicCommand = elektraArrayGet (arrayParent, ks);
+
+		if (ksGetSize (dynamicCommand) > 0)
+		{
+			printf ("dynamically invoke the command '");
+			ksRewind (dynamicCommand);
+			Key * cur = NULL;
+			while ((cur = ksNext (dynamicCommand)) != NULL)
+			{
+				printf ("%s ", keyString (cur));
```
markus2330: I get here the output `dynamically invoke the command 'dynamic abc '` but would not expect `dynamic` to be executed`(but only abc)

doc/news/_preparation_next_release.md
```diff
@@ -92,11 +92,16 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
...
 
-### <<Library1>>
-
-- <<TODO>>
-- <<TODO>>
-- <<TODO>>
+### Opts
+
+- The library function `elektraGetOpts` now supports sub-commands. For example `add` is a sub-command in `git add`, and
+  interprets `-p` differently from `git`: `git -p add` is `git --paginate add`, but `git add -p` is `git add --patch`.
```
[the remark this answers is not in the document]
kodebach: I found it easiest to explain by using an example. I would have used `kdb`, but since it currently uses global options there is no real-world example.

doc/tutorials/command-line-options.md
```diff
@@ -63,12 +63,151 @@ into the array.
...
+a non-option argument is encountered. This is similar to the POSIX version of getopt(3). NOTE: `posixly = 1` must be set on
+the parent key passed to `elektraGetOpts`, not in the specification itself.
+
+Additionally, there is `args=indexed`. If it is specified on a key, the key must also have the metakey `args/index=N` set to
+an integer `N`. Such a key will be set to the unused element at index `N`. If a key has `args=indexed` and `args/index=N`,
+then there must also be keys for all integers `0 <= X < N` with `args=indexed` and `args/index=N` set. For example, you cannot
+use `args/index=0` and `args/index=2` without `args/index=1`.
+
+Combining `args=indexed` and `args=remaining` in the same specification (on different keys) is also possible. The key with
+`args=remaining` will only contain those elements not used via `args=indexed`.
```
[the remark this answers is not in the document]
kodebach: What do you mean with "which index the remaining args get"?

doc/tutorials/command-line-options.md
```diff
@@ -63,12 +63,151 @@ into the array.
...
+assigned as follows:
+
+- `from = apple`
+- `to = banana`
+- `more/#0 = cherry`
+- `more/#1 = date`
+
+## Sub-Commands
+
+`elektraGetOpts` also supports sub-commands. For example calling `git -p add` and `git add -p` result in different behavior.
```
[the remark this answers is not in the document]
kodebach: It's not a difference, `git`  has the exact same behaviour  as `elektraGetOpts` with respect to sub-commands.

I will add a general explanation of the term "sub-command" before using the example to explain the effect on options-processing. (Explaining it without the example will only get more confusing.)

doc/tutorials/command-line-options.md
```diff
@@ -63,12 +63,151 @@ into the array.
...
+This is because the options that `git` understands are separate from the options that its sub-command `add` knows. However,
+the option `-p` is understood by both. In `git` it is short for `--paginate` and in `add` it is short for `--patch`.
+
+The first thing you need to know about sub-commands is that they imply `posixly = 1` mode. This means **all** options must
```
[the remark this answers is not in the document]
kodebach: If sub-commands are used (i.e. the spec parent has `command = `), yes. Otherwise, no.

doc/tutorials/command-line-options.md
```diff
@@ -63,12 +63,151 @@ into the array.
...
+`git -p add` and `git add -p`.
+
+A sub-command is created by specifying `command` on a key. To enable sub-command processing the parent key of the whole
+specification must have `command` set to an empty string. All keys marked with `command` **directly** below another
+key `K` marked with `command` (e.g. the parent key) are sub-commands of `K`. It is an error, if the immediate parent of
+a key `X` marked with `command` is not marked with `command` and `X` is not the parent of the whole specification.
+
+To inform the application about the invoked sub-commands, `elektraGetOpts` sets each `command` key to one of two values:
+
+- All invoked sub-commands `S` will be set to `XYZ`, where `XYZ` is the base name of the key whose sub-command of `S` was invoked.
```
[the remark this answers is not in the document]
kodebach: I will rewrite the sentence. The idea is that in the example below, e.g. `kdb`  is set to `setter` when `kdb set` is invoked.

examples/optsCommands.c
```diff
@@ -0,0 +1,161 @@
...
+	{
+		char * help = elektraGetOptsHelpMessage (errorKey, NULL, NULL);
+		fprintf (stderr, "%s\n", help);
+		elektraFree (help);
+		keyDel (errorKey);
+		ksDel (ks);
+		return EXIT_SUCCESS;
+	}
+
+	printf ("A real implementation would now \n");
```
kodebach:
```suggestion
	printf ("A real implementation would now\n");
```

examples/optsCommands.c
kodebach: the `lookup == NULL` should never be true AFAIK know, but `strlen (keyString (lookup)) == 0` is true, if you call `bin/optsCommands get ""`

examples/optsCommands.c
kodebach: Because of posixly evaluation (otherwise commands don't work), `bin/optsCommands set abc -v` means "set key `abc` to value `-v`".

In `bin/optsCommands set -v abc` you are missing the value for the `set` command. `bin/optsCommands set -v abc x` should work.

examples/optsCommands.c
kodebach: What was the shell command you invoked?

src/libs/opts/opts.c
```diff
@@ -719,17 +1010,177 @@ bool processEnvVars (KeySet * usedEnvVars, Key * specKey, Key ** keyWithOpt, Key
...
+		{
+			ELEKTRA_SET_VALIDATION_SEMANTIC_ERRORF (
+				errorKey, "'args/index=" ELEKTRA_LONG_LONG_F "' is already used by '%s'. Offending Key: %s", indexValue,
+				existing, keyName (specKey));
+			return false;
+		}
+
+		keySetMeta (indexKey, indexMetaName, keyName (specKey));
+
+		char * argHelp = elektraFormat ("  %-28s%s", keyBaseName (specKey), help);
```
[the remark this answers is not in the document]
kodebach: There is 2 spaces in front for indentation then `%-28s` prints the basename left-aligned padded to 28 characters. Therefore the `%s` prints the `help` string starting at column 30.

Technically this breaks for basenames longer than 28 characters, but I assumed that wouldn't happen very often.

src/libs/opts/opts.c
```diff
@@ -1297,6 +1807,124 @@ void setOption (Key * option, const char * value, bool repeated)
...
+	KeySet * envValues = parseEnvp (envp);
+
+	Key * argsParent = keyNew (keyName (command), KEY_END);
+	keyAddBaseName (argsParent, "args");
+	KeySet * args = elektraArrayGet (argsParent, options);
+	keyDel (argsParent);
+
+	Key * keyWithOpt;
+	ksRewind (spec->keys);
+	while ((keyWithOpt = ksNext (spec->keys)) != NULL)
```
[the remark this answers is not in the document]
kodebach: I'm not even sure that would work, but even if it works, it would be very confusing, since there are two different `result` variables used in different context.

I will instead add comments to the `continue`s and `return`s to clarify what is happening.

doc/news/_preparation_next_release.md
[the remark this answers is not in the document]
markus2330: My confusion here is that I thought you are describing how `git` handles it. Only much later I found out, that your implementation and `git` handles it the same way.

doc/tutorials/command-line-options.md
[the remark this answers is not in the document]
markus2330: If the returned KeySet of args=remaining is identical whether or not args=indexed are present.

doc/tutorials/command-line-options.md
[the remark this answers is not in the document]
markus2330: > It's not a difference, git has the exact same behaviour as elektraGetOpts with respect to sub-commands.

This was unclear to me, adding this information would be very valuable.

> I will add a general explanation of the term "sub-command" before using the example to explain the effect on options-processing. (Explaining it without the example will only get more confusing.)

Good!

doc/tutorials/command-line-options.md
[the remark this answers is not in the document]
markus2330:
```suggestion
The first thing you need to know about sub-commands is that they turn on `posixly = 1` mode. This means **all** options must
```

examples/optsCommands.c
markus2330:
```suggestion
			printf ("report the error 'empty parameter: keyname'\n");
```

examples/optsCommands.c
markus2330: > in/optsCommands set abc -v means "set key abc to value -v".

Ok, this should be written where you write that `posixly = 1` is implied. Because we are not doing exactly what POSIXLY_CORRECT does but actually something else: the original meaning is "If it is set, parsing stops as soon as a parameter is found that is not an option or an option argument." but for us it is: "If it is set, parsing stops as soon as a parameter is found that is not **a sub-command,** an option or an option argument."

examples/optsCommands.c
markus2330:
```sh
bin/optsCommands dynamic abc
#> A real implementation would now 
#> dynamically invoke the command 'dynamic abc '

File: doc/tutorials/command-line-options.md (hunk 9 above)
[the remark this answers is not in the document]
Comment (kodebach) [reply doc/tutorials/command-line-options.md]: > The key with `args=remaining` will only contain those elements not used via `args=indexed`.

i.e. if there is a `args/index=0` then `args=remaining` will not contain the first parameter.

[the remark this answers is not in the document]
Comment (markus2330) [reply doc/tutorials/command-line-options.md]: Yes, and to also clarify the returned (assigned) keys. E.g. `more/#0 = cherry` and not `more/#2 = cherry` even though it is the third parameter.

File: src/libs/opts/opts.c (hunk 15 above)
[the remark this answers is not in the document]
Comment (markus2330) [reply src/libs/opts/opts.c]: Thank you!

File: src/libs/opts/opts.c (hunk 14 above)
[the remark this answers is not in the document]
Comment (markus2330) [reply src/libs/opts/opts.c]: Ok, so maybe nothing to do then. So currently we will get a bug report when someone wants different formatting of the help text.

File: examples/optsCommands.c (hunk 7 above)
Comment (kodebach) [reply examples/optsCommands.c]: Ah ok, I think your confusion stems from the use of `dynamically`  and `dynamic`.
It might become clearer, when you look at `bin/optsCommands test abc`:

```
bin/optsCommands test abc
#> A real implementation would now 
#> dynamically invoke the command 'test abc '
```

File: examples/optsCommands.c (hunk 6 above)
Comment (kodebach) [reply examples/optsCommands.c]: In a sense, we are doing what `POSIXLY_CORRECT` does, namely once we encounter a string in `argv` that doesn't start with `-`  we stop looking for options **for the current command**.

In other words, the behaviour is as if a POSXILY_CORRECT `getopt_long` was invoked separately for each sub-command. And everytime `getopt_long` returns, we either start another run, or collect into `args=indexed`/`args=remaining` depending on the next element of `argv`.

I will clarify this.

File: examples/optsCommands.c (hunk 7 above)
Comment (markus2330) [reply examples/optsCommands.c]: Ahh, you are right. So you basically already reimplemented what `kdb` does :smile:  The output `dynamically invoke the command 'test' with the arguments 'abc '` would be imho more pretty, though.

Comment (kodebach) [reply examples/optsCommands.c]: > The output dynamically invoke the command 'test' with the arguments 'abc ' would be imho more pretty, though.

I will change the output.

File: src/bindings/swig/python/examples/example_gopts.py [hunk 16]
```diff
@@ -0,0 +1,160 @@
...
+	elif rc == -1:
+		print("ERROR: Contract malformed/NULL pointer!\n", error_key.getMeta("error/reason").value, file=sys.stderr)
+		removeSpec()
+		exit(1)
+	
+	ks = kdb.KeySet(0)
+	db.get(ks, spec_base_key)
+	
+	if "proc/elektra/gopts/help" in ks and ks["proc/elektra/gopts/help"].value == "1":
+		# TODO: call elektraGetOptsHelpMessage (not exposed to python right now) and print result
```
markus2330: Did you consider to add the help text directly as value of "proc/elektra/gopts/help"?

kodebach: I used `proc/elektra/gopts/help/message` to maintain backwards compatibility.

markus2330: Very nice :fireworks:

kodebach:
Anybody know, why every Github link causes the `🔗 Check` build job to fail?

mpranj:
Currently unclear, you can ignore github link issues for this PR. I will take a look at the link problems later.

mpranj:
Thank you so much!

PR #3358, opened 2020-03-01
YAML CPP: Set and Require Boolean Metadata

sanssecours:
This pull request is based on PR #3357. If you want a clean history of the additional changes, then please merge #3357 first.


mpranj:
Thank you for improving YAML CPP and always keeping the docu consistent. :heart:

PR #3359, opened 2020-03-01
Docker: Fix Install of `apiaryio` Gem

sanssecours:


mpranj:
Thank you for working on this!

Issue #3373, opened 2020-03-19
Cirrus: Link Checking Fails For GitHub URLs


sanssecours:
# Problem Description

It seems that the connection attempts of the [link checker](https://github.com/ElektraInitiative/libelektra/blob/master/scripts/link-checker) in the Cirrus build job `🔗 Check` fail for all GitHub links. To debug this problem I removed the `--quiet` option of the `wget` invocations inside the link checker:

https://github.com/ElektraInitiative/libelektra/blob/7098ab6f12a2f8624107731f4d64e6f3b66b881d/scripts/link-checker#L63-L65

Afterwards the output of the command shows that GitHub returns `429` error codes:

```
Connecting to github.com (github.com)|140.82.113.4|:443... Loaded CA certificate '/etc/ssl/certs/ca-certificates.crt'
Resolving github.com (github.com)... 140.82.113.4
Connecting to github.com (github.com)|140.82.113.4|:443... 140.82.114.3
Connecting to github.com (github.com)|140.82.114.3|:443... connected.
connected.
connected.
connected.
connected.
connected.
connected.
connected.
connected.
HTTP request sent, awaiting response... HTTP request sent, awaiting response... HTTP request sent, awaiting response... HTTP request sent, awaiting response... HTTP request sent, awaiting response... HTTP request sent, awaiting response... HTTP request sent, awaiting response... HTTP request sent, awaiting response... HTTP request sent, awaiting response... 429 too many requests
Remote file does not exist -- broken link!!!
```

It think we need to handle these rate limiting problems. I would recommend we just [whitelist](https://github.com/ElektraInitiative/libelektra/blob/master/tests/linkchecker.whitelist) all links that point to GitHub, until we find a better solution.

mpranj:
Thank you for reporting. I agree that we whitelist the links in the meantime. In the long term I think that a regex-whitelist would be a nicer solution (as suggested in #3374).

markus2330:
The regex-whitelist is quite trivial to do, removing the `-F` from the grep arguments in the link-checker should be enough (assuming that none of our links use chars that have special meaning as regex)..

mpranj:
Closed via #3376.

PR #3376, opened 2020-03-19
link checker: enable regex

mpranj:
#3373


scripts/link-checker
```diff
@@ -51,7 +51,7 @@ check() {
 	link=$(echo "$1" | grep -oE "(https|http|ftp):[^|]*")
 	files=$(echo "$1" | grep -oE "\|.*" | sed 's/|/ /g')
 
-	if grep -Fxq "$link" "$WHITELIST"; then
+	if grep -xq "$link" "$WHITELIST"; then
```
sanssecours: This will (probably) not work. I know, because I already tried it 😊. One option to support regular expressions is to iterate through all patterns of the whitelist. My current solution for this approach looks like this:

```sh
while IFS= read -r pattern; do
	if printf '%s' "$link" | grep -q "$pattern"; then
		printf "whitelisted: “%s” because of pattern “%s”\n" "$link" "$pattern"
		printf "%i/%i\r" "$COUNTLINKS" "$NUMLINKS"
		return
	fi
done < "$WHITELIST"
```

mpranj: Thank you for pointing it out. I think I found a way around it, at least it worked locally.

mpranj:
@sanssecours thank you again for the input. I think it's done, could you take a look?

The script now uses the whitelist as a regex input file for grep. Since empty lines match everything we strip all empty lines with sed.  Broken links are apparently still detected ✅.

scripts/link-checker
```diff
@@ -42,6 +42,7 @@ END {
 
 WHITELIST=$(mktemp)
 grep -v '^#' "$SCRIPT_DIR/../tests/linkchecker.whitelist" > "$WHITELIST"
+sed -i '/^$/d' "$WHITELIST"
```
sanssecours: Unfortunately [using `-i` without any additional argument will not work, if you use the BSD version of `sed`](https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux). I think we can work around this problem using `grep` to drop both the empty lines and comments.

```suggestion
grep -Ev '^(#|[[:space:]]*$)' "$SCRIPT_DIR/../tests/linkchecker.whitelist" > "$WHITELIST"
```

scripts/link-checker
```diff
@@ -51,7 +52,7 @@ check() {
 	link=$(echo "$1" | grep -oE "(https|http|ftp):[^|]*")
 	files=$(echo "$1" | grep -oE "\|.*" | sed 's/|/ /g')
 
-	if grep -Fxq "$link" "$WHITELIST"; then
+	if echo "$link" | grep -Eqxf "$WHITELIST"; then
```
sanssecours: I think it makes sense to remove the option `x`:

>     -x, --line-regexp
>             Only input lines selected against an entire fixed string or regu-
>             lar expression are considered to be matching lines.

Dropping it should allow us to use simpler patterns like

```
github.com
```

in the whitelist instead of

```
https://github.com(\S)*
```

mpranj: While I would love a simpler syntax, wouldn't this also cause any other links like `https://example.com/fancy/github.com/stuff` to match?

sanssecours: Of course, but I would assume we do not use such strange URLs in our documentation 😊. Anyway, the decision is up to you. If you do not want to allow partial matching, then you obviously do not have to implement it. It is also quite easy to add this feature afterwards, if we decide that the current system makes writing whitelist patterns to complicated

mpranj: Rethinking it, I changed it according to your suggestion but left a note in the whitelist docu. It's better if the syntax is simple. Thank you for your valuable review!

Issue #3339, opened 2020-01-19
Build error: Compilation stops at ruby plugin?


stefanhusmann:
## Steps to Reproduce the Problem

I try to build elektra using this build function:
``` sh
build() {
  cd lib$pkgname-$pkgver
  [[ -d build ]] || mkdir build
  cd build
  cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr \
        -DPLUGINS:STRING="ALL" \
        -DTOOLS:STRING="ALL" \
        -DBUILD_STATIC:STRING=OFF \
        -DBINDINGS:STRING="ALL" \
        -DSWIG_EXECUTABLE:STRING="/usr/bin/swig" \
        -DLUA_INCLUDE_DIR:PATH=/usr/include \
        -DLUA_LIBRARY:STRING=/usr/lib/liblua.so ..
  make clean 
  make
}
```
With ruby 2.7.0 installed this fails. 
[buildlog.txt](https://github.com/ElektraInitiative/libelektra/files/4082969/buildlog.txt)
This is true for both  tarball version 0.9.1 and fresh sources from master.

Downgrading to ruby 2.6.5 or deinstalling ruby at all solves the issue. But as I want to provide packages for AUR (Arch Linux), thes solutions are second grade solutions.

Is there a cmake directive to disable ruby support?

mpranj:
Thank you for reporting the problem! As a workaround you can disable the ruby bindings and plugin via CMake directives like this:

```
...
        -DPLUGINS:STRING="ALL;-ruby" \
        -DBINDINGS:STRING="ALL;-ruby" \
...
```

We will leave the issue open until the underlying problem is resolved. @BernhardDenner would you have time to look at this?

markus2330:
I am not sure if this is something that can be fixed on our side. For me it looks like the used swig is incompatible with the used ruby? Which distribution is this?

```
In file included from /home/haawda/paketierung/meine_Pakete/elektra/src/libelektra-0.9.1/src/plugins/ruby/ruby.cpp:22:
/home/haawda/paketierung/meine_Pakete/elektra/src/libelektra-0.9.1/build/src/plugins/ruby/runtime.h: In function 'void SWIG_RubyInitializeTrackings()':
[...]

PR #3377, opened 2020-04-02
docker: fix debian sid packages

mpranj:
Debian sid bumped some package versions so some packages we used are no longer available. Notably python 2 was dropped and ruby was bumped to 2.7.


markus2330:
Thank you for working on this!

Why we need ruby 2.5? Would be good to have this documented.

mpranj:
> Why we need ruby 2.5? Would be good to have this documented.

SWIG does not support it ([see issue](https://github.com/ElektraInitiative/libelektra/issues/3339)).

It also seems that we cant compile with the latest python version on debian unstable. I haven't had time to dig deeper, maybe someone knows what the issue is?

PR #3383, opened 2020-04-08
Fix macOS builds

mpranj:
Various build fixes.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -168,8 +168,11 @@ mac_task:
       npm install dredd --global
 
   script:
-    - > # Use latest version of LLVM to translate Elektra
+    - | # Use latest version of LLVM to translate Elektra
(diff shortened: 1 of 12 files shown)


mpranj:
@sanssecours within this PR we now get some leaks and undefined behaviour reported by 🍎 Clang ASAN. Note that I changed some linker and compile flags according to the homebrew install instructions. Do you think these are upstream issues or should I roll back those changes? I think the notable difference is that it now uses llvm's bundled c++ library.

Log:
[tests.log.txt](https://github.com/ElektraInitiative/libelektra/files/4456301/tests.log.txt)

mpranj:
Thank you so much @sanssecours!

sanssecours:
> Do you think these are upstream issues or should I roll back those changes?

I assume the ASan problems are caused by upstream (either ANTLR/yaml-cpp or libc++). I am not super sure though. Anyway, for now I just added another Sanitizer blacklist entry.

PR #3389, opened 2020-04-09
CI: improve fedora image

### LINKED ISSUE #3227: Jenkins CI: add Fedora docker image
Add a current Fedora docker image to the Jenkins CI pipeline to test compatibility with newer packages than on Debian.

If I take on this issue I will probably replace one of the `mmap` builds with fedora, so that we don't **just** increase build times.

mpranj:
Jenkins: main build stage with debian and fedora stable. Move oldstable to second build stage.

Improve cirrus fedora image and add fedora docker image for jenkins to cover many tests.

Closes #3227.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -215,7 +215,7 @@ mac_task:
 linux_task:
   env:
     INSTALL_DIR: install
-    LUA_CPATH: ${BUILD_DIR}/install/lib/lua/5.2/?.so;
+    LUA_CPATH: ${CIRRUS_WORKING_DIR}/${BUILD_DIR}/${INSTALL_DIR}/lib/lua/5.2/?.so;
 
   matrix:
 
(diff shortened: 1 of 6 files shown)


markus2330:
Amazing to finally see Fedora Images for Jenkins!

mpranj:
What do you think about moving debian oldstable to the second build stage and keeping current debian and fedora stable as the main builds? Is there any specific reason to keep strech in the main builds?

markus2330:
Both Stretch and Buster was some time together in the main builds because Stretch did not build/test anything. But I think this is already fixed since quite some time so there is actually no reason to keep Stretch in the main builds.

Having one Fedora and and one Debian build in the main builds sounds very reasonable to me.

mpranj:
@sanssecours only if you want/have time to glance over it. Feel free to merge if you do look it over.

Issue #3378, opened 2020-04-06
SWIG <=4.0.1: upstream issues with ruby 2.7


mpranj:
When compiling with the SWIG bindings we currently do not support ruby >= 2.7. This was already reported in #3339. The [upstream issue](https://github.com/swig/swig/issues/1689) is resolved but the fix is not yet released. It should be resolved with SWIG 4.0.2. The issue was encountered again when Debian unstable bumped ruby package versions (#3377).

Check before releases if we can bump the ruby package versions.

mpranj:
Closed via #3411.

PR #3411, opened 2020-04-26
swig: exclude incompatible swig, ruby, python combinations

### LINKED ISSUE #3379: SWIG <=4.0.1: upstream issues with python 3.8
When compiling with the SWIG bindings we currently do not support python >= 3.8. The issue was resolved [upstream](https://github.com/swig/swig/commit/eaa16914d1686002d33cb3f61875832977146805), but the fix is not yet released. It should be resolved with SWIG 4.0.2.

Check before releases if we can enable the python binding on `debian-unstable-full-clang` again.

mpranj:
Closes #3379 and #3378.
Pin prettier version to 1.19.1 on cirrus since 2.0 introduces many changes.
Disable crypto plugin on `fedora-31-full`and `debian-unstable-full-clang`.
Prefer ruby 2.6 when installed (to avoid problems with 2.7).
Prefer swig 4.0 over old versions.


mpranj:
Seems that it's not worth the hassle to  downgrade ruby on cirrus macOS jobs. Ruby 2.7 is preinstalled (required by other packages?) and using rvm to manage versions implies compiling ruby 2.5 on every build because binaries are not available.

We now exclude incompatible versions by CMake.

markus2330:
Yes, this is a much better approach.

PR #3419, opened 2020-04-29
cirrus, jenkins: upgrade/add fedora 32 images

### LINKED ISSUE #3423: crypto: memleak in gpg2
Memleaks in `testmod_crypto` are still occuring, this time on `debian-unstable-full-clang`. See [build log](https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3419/7/pipeline).

Viewing the log locally is less scrambled than in the web browser via jenkins: [DynamicAnalysis.txt](https://github.com/ElektraInitiative/libelektra/files/4558674/DynamicAnalysis.txt)

By temporarily enabling more valgrind output I think I was able to narrow the leak down.

mpranj:
Adds Fedora 32 images to cirrus and jenkins.
Closes #3423.

### DIFF
--- a/scripts/cmake/ElektraCache.cmake
+++ b/scripts/cmake/ElektraCache.cmake
@@ -353,7 +353,7 @@ set (
 	""
 	CACHE STRING "Optional suffix to use on lib folders (e.g. 64 for lib64)")
 
-set (MEMORYCHECK_COMMAND_OPTIONS "--trace-children=yes --leak-check=full --gen-suppressions=all --error-exitcode=1")
+set (MEMORYCHECK_COMMAND_OPTIONS "--trace-children=yes --leak-check=full --show-leak-kinds=all --track-origins=yes --gen-suppressions=all --error-exitcode=1")
 set (
 	MEMORYCHECK_SUPPRESSIONS_FILE
 	"${CMAKE_SOURCE_DIR}/tests/valgrind.suppression"
--- a/scripts/cmake/ElektraCache.cmake
+++ b/scripts/cmake/ElektraCache.cmake
@@ -353,8 +353,7 @@ set (
 	""
 	CACHE STRING "Optional suffix to use on lib folders (e.g. 64 for lib64)")
 
-set (MEMORYCHECK_COMMAND_OPTIONS
-     "--trace-children=yes --leak-check=full --show-leak-kinds=all --track-origins=yes --gen-suppressions=all --error-exitcode=1")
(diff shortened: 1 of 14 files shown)


tests/valgrind.suppression
```diff
@@ -207,6 +207,23 @@
...
+   fun:_dl_map_object
+   fun:dl_open_worker
+   fun:_dl_catch_exception
+   fun:_dl_open
+   fun:dlopen_doit
+   fun:_dl_catch_exception
+   fun:_dl_catch_error
+   fun:_dlerror_run
+   fun:dlopen@@GLIBC_2.2.5
+}
```
mpranj, line 226: Is this suppression too general? I took the output directly from valgrind. My concern is that the suppression is not limited to the gpg2 binary, but I don't know if this can be specified such that it matches only for that binary.

markus2330: Looks very specific to me. You did not include any catch-all, so it should not be possible that such a leak is actually triggered by our code?

scripts/cmake/ElektraCache.cmake
```diff
@@ -353,7 +353,8 @@ set (
 	""
 	CACHE STRING "Optional suffix to use on lib folders (e.g. 64 for lib64)")
 
-set (MEMORYCHECK_COMMAND_OPTIONS "--trace-children=yes --leak-check=full --gen-suppressions=all --error-exitcode=1")
+set (MEMORYCHECK_COMMAND_OPTIONS
+     "--trace-children=yes --leak-check=full --show-leak-kinds=all --track-origins=yes --gen-suppressions=all --error-exitcode=1")
```
mpranj, line 357: @petermax2 maybe you can use this to show which leaks are detected in crypto?

scripts/jenkins/Jenkinsfile
```diff
@@ -936,6 +945,7 @@ def buildAndTest(testName, image, extraCmakeFlags = [:],
 
         if(tests) {
           artifacts.add("\"${buildDir}\"/Testing/*/*.xml")
+          artifacts.add("\"${buildDir}\"/Testing/Temporary/MemoryChecker.*.log")
```
mpranj, line 948: @petermax2 I use this to add the valgrind output to the artifacts. Can not be done on master (uses too much space), but can be enabled in a PR at any time to have non-scrambled valgrind output.

scripts/cmake/ElektraCache.cmake
petermax2: We would probably need the debug symbols of gpg nevertheless.

mpranj: I am working on adding debug symbols for some of the used libraries. 👍

Issue #3436, opened 2020-05-20
rust bindings: compile error


mpranj:
Builds on [master are broken](https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/master/182/pipeline) due to a compilation error. It seems to be caused by the dependencies.

## Error log

```sh
[ 28%] Built target elektra-highlevel

error[E0658]: use of unstable library feature 'alloc': this library is unlikely to be stabilized in its current form or name (see issue #27783)

  --> /home/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/object-0.19.0/src/lib.rs:15:1

   |

15 | extern crate alloc;

   | ^^^^^^^^^^^^^^^^^^^



Scanning dependencies of target elektra-merge

[ 28%] Built target gtest

Scanning dependencies of target cpp_example_io

[ 28%] Built target elektra-merge

Scanning dependencies of target cpp_example_hierarchy

error[E0658]: use of unstable library feature 'alloc': this library is unlikely to be stabilized in its current form or name (see issue #27783)

  --> /home/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/gimli-0.21.0/src/lib.rs:41:1

   |

41 | extern crate alloc;

   | ^^^^^^^^^^^^^^^^^^^



error[E0658]: use of unstable library feature 'maybe_uninit' (see issue #53491)

 --> /home/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/gimli-0.21.0/src/read/cfi.rs:5:23

  |

5 | use core::mem::{self, MaybeUninit};

  |                       ^^^^^^^^^^^



error[E0658]: use of unstable library feature 'maybe_uninit' (see issue #53491)

    --> /home/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/gimli-0.21.0/src/read/cfi.rs:2312:20

     |

2312 |     rules_storage: MaybeUninit<[(Register, RegisterRule<R>); MAX_RULES]>,

     |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^



error[E0658]: use of unstable library feature 'maybe_uninit' (see issue #53491)

    --> /home/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/gimli-0.21.0/src/read/cfi.rs:2329:28

     |

[...]

PR #3437, opened 2020-05-20
disable rust bindings

mpranj:
Disable rust bindings due to #3436.


markus2330:
I think it is better if we simply take the plugin out of our CI but not from CMake.

@PhilippGackstatter is a quick fix of  #3436 possible?

mpranj:
The goal here was to create one commit which would be immediately reverted in the PR that fixes the problems.

I'm open to better approaches, but I currently don't have time to implement anything more elaborate.

markus2330:
Yes, I understand, the first time it is quite time consuming to find the right places to add `-rust` to BINDINGS everywhere. But luckily @PhilippGackstatter was super-quick in fixing the problem.

PR #3478, opened 2020-09-03
TOML as default storage plugin

bauhaus93:
PR for finding errors when the toml plugin is set as the default storage plugin.

## Basics

      contains `_(my name)_`)

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -317,7 +317,7 @@ docu_task:
     - >
       set --
       -DBUILD_PDF=ON
-      -DPLUGINS='dump;list;resolver_fm_hpu_b;spec;sync;yajl'
+      -DPLUGINS='toml;base64;list;resolver_fm_hpu_b;spec;sync;yajl'
       -DTOOLS='ALL;web;website'
     - *generate
     - *build
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -65,6 +65,10 @@ The following section lists news about the [plugins](https://www.libelektra.org/
 
 - Fixed a bug ([#3896](https://issues.libelektra.org/3896)) that caused the `toml` plugin to swallow the first letter of
   all keys (after the namespace), if the parent key was a root key (e.g. `user:/`). _(Klemens Böswirth)_
+- Added the TOML plugin, which can read and write TOML files using flex and bison. _(Jakob Fischer)_
+- Removed the `null` plugin dependency of the plugin. _(Jakob Fischer)_
+- The `type` metakey is now set for numbers on reading. _(Jakob Fischer)_
+- Set TOML as default storage plugin for testing purposes _(Jakob Fischer)_
 
 ### Python
 
--- a/scripts/cmake/ElektraCache.cmake
+++ b/scripts/cmake/ElektraCache.cmake
@@ -134,7 +134,7 @@ set (
 
 set (
 	KDB_DEFAULT_STORAGE
-	"dump"
+	"toml"
 	CACHE STRING "This storage plugin will be used initially (as default and for bootstrapping).")
 
 if (KDB_DEFAULT_STORAGE STREQUAL "storage")
--- a/scripts/docker/cirrus/arch/Dockerfile
+++ b/scripts/docker/cirrus/arch/Dockerfile
@@ -1,10 +1,14 @@
 FROM archlinux:base-devel
 
 RUN pacman -Sy --noconfirm \
+    base-devel \
+    bison \
+    boost \
     cmake \
     diffutils \
     discount \
     doxygen \
+    flex \
     git \
     go \
     graphviz \
--- a/src/plugins/toml/README.md
+++ b/src/plugins/toml/README.md
@@ -5,7 +5,7 @@
 - infos/needs = base64
 - infos/recommends = type
 - infos/placements = getstorage setstorage
-- infos/status = experimental unfinished
+- infos/status = maintained
 - infos/metadata = order comment/# comment/#/start comment/#/space type tomltype origvalue
 - infos/description = This storage plugin reads and writes TOML files using Flex and Bison.
 
--- a/src/plugins/toml/lexer.l
+++ b/src/plugins/toml/lexer.l
@@ -285,8 +285,7 @@ static char * readString (char terminatorChar, bool isMultiline, Driver * driver
 				bufferAddChar (buffer, c, driver);
 			}
 			terminatorCount++;
-		}
-		else if (!isMultiline && c == '\n') {
+		} else if (!isMultiline && c == '\n') {
 			driverError (driver, ERROR_SYNTACTIC, yylineno,
 				     "Malformed Input: Found newline in non-multiline string, but is only allowed in multiline strings");
 			return bufferConsume(buffer);
@@ -310,6 +309,9 @@ static char * readString (char terminatorChar, bool isMultiline, Driver * driver
 		{
 			unput (c);
 			readNonAsciiChar (buffer, driver);
+		}
+		if (terminatorCount > 0 && c != terminatorChar) {
+			terminatorCount = 0;
 		}
 	}
 	return bufferConsume (buffer);
--- a/src/plugins/toml/utility.c
+++ b/src/plugins/toml/utility.c
@@ -437,3 +437,4 @@ bool isValidEscapeSequence (const char * seq)
 	{
 		return false;
 	}
+}
--- a/src/plugins/toml/utility.c
+++ b/src/plugins/toml/utility.c
@@ -410,25 +410,30 @@ bool isNullString (const char * str)
 	return elektraStrCmp (str, nullIndicator) == 0;
 }
 
-bool isValidEscapeSequence(const char * seq) {
-	if (seq[0] == '\\') {
-		switch(seq[1]) {
-			case 'b':
-			case 't':
-			case 'n':
-			case 'f':
-			case 'r':
-			case '"':
-			case '\\':
-				return true;
-			case 'u':
-				return validUtf8FromUnicode(&seq[2], 4);
-			case 'U':
-				return validUtf8FromUnicode(&seq[2], 8);
-			default:
-				// TODO: Correctly identify handle line ending backslashes
-				return false;
+bool isValidEscapeSequence (const char * seq)
+{
+	if (seq[0] == '\\')
+	{
+		switch (seq[1])
+		{
+		case 'b':
+		case 't':
+		case 'n':
+		case 'f':
+		case 'r':
+		case '"':
+		case '\\':
+			return true;
+		case 'u':
+			return validUtf8FromUnicode (&seq[2], 4);
+		case 'U':
+			return validUtf8FromUnicode (&seq[2], 8);
+		default:
+			// TODO: Correctly identify handle line ending backslashes
+			return false;
 		}
-	} else {
+	}
+	else
+	{
 		return false;
 	}
--- a/src/plugins/toml/utility.c
+++ b/src/plugins/toml/utility.c
@@ -14,6 +14,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "codepoint.h"
+
 void dumpKS (KeySet * keys)
 {
 	printf ("DUMPING KS, size = %lu\n", ksGetSize (keys));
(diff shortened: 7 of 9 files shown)


markus2330:
Thank you for creating the PR! Without updating the release notes the CI does not do much.

mpranj:
Also the `experimental unfinished` status excludes the plugin from being built by default, so nothing is really tested yet.

kodebach:
@markus2330 Did we ever solve the issue of storing arbitrary metakeys in TOML files? If not, this PR will never work.

markus2330:
Maybe #3500 already does enough. Otherwise we can think about having different default storages for spec: and other namespaces?

kodebach:
> Maybe #3500 already does enough. 

AFAICT there is some logic to convert metakeys into comments starting with `!!ELEKTRA_META`. I would call that a workaround, but not a solution. There are many issues with such a solution and it was one of the explicit downsides of the old INI plugin.

Some of this could be improved, but the current code has two big issues:
1. It breaks if the name of the metakey contains a space
2. If the TOML file is edited externally, comments must not be deleted and must stay attached to the following key-value line. This essentially goes against the definition of a comment. _Comments are for humans not for machines_.

> Otherwise we can think about having different default storages for spec: and other namespaces?

This is not enough. A default storage plugin must be able to handle `kdb set-meta user:/mykey mymeta myvalue`.

markus2330:
> AFAICT there is some logic to convert metakeys into comments starting with !!ELEKTRA_META. I would call that a workaround, but not a solution. There are many issues with such a solution and it was one of the explicit downsides of the old INI plugin.

The alternative would be to have a "meta" mode, similar to ni.

> Some of this could be improved, but the current code has two big issues:
>
>    It breaks if the name of the metakey contains a space

doc/METADATA.ini doesn't contain such a metakey (except in the explanation of the syntax), so it would be easy for us to prohibit it.

 >   If the TOML file is edited externally, comments must not be deleted and must stay attached to the following key-value line. This essentially goes against the definition of a comment. Comments are for humans not for machines.

Does TOML define a proper extension possibility?

>>    Otherwise we can think about having different default storages for spec: and other namespaces?
>
> This is not enough. A default storage plugin must be able to handle kdb set-meta user:/mykey mymeta myvalue.

Failing in that case should be fine, too. It is recommended to store properties only in `spec:`. Or does the e.g. high-level API require this possibility?

kodebach:
> doc/METADATA.ini doesn't contain such a metakey (except in the explanation of the syntax), so it would be easy for us to prohibit it.

We had endless discussions about restricting key names and you always were very much convinced that it would be a bad idea. Now you suddenly would accept restrictions?

> Does TOML define a proper extension possibility?

Not AFAIK.

> Failing in that case should be fine, too.

IMO if the `kdb` tool has a command, it should work with the default configuration (i.e. default storage plugin, etc.). It would just be confusing to people, if we say `kdb` has this command, but it doesn't work by default.

> Or does the e.g. high-level API require this possibility?

Something like `kdb set-meta user:/myarray array "#4"` is needed to tell the `spec` plugin, how big an array should be. Otherwise it won't know, how many default keys it needs to create.

kodebach:
> The alternative would be to have a "meta" mode, similar to ni.

Well there is one thing we could do...

This `ni` file
```ini
mykey/foo/bar = 123
mykey/foo/baz/bar = abc
mykey/foo/baz/boo = true

[mykey/foo/bar]
type = long_long
check/min = 100
check/max = 150

[mykey/foo/baz/bar]
type = string
check/minlength = 2

[mykey/foo/baz/boo]
type = boolean
```

could be written as this `toml` file:

```toml
[mykey.foo]
bar = 123
baz.bar = "abc"
baz.boo = true
"baz.boo/meta.default" = false

[mykey.foo.bar/meta.check]
min = 100
max = 150

[mykey.foo.baz.bar/meta]
check.minlength = 2
```

The key here is that `.` is TOML's separator for nested levels. So `mykey.foo` means `foo` within `mykey`. In Elektra we normally use `/` for that. In TOML `/` has no special meaning, so other parsers will just treat it as any old character. In the `toml` plugin however, we could give it special meaning.

The idea is that we use `/meta.` within a TOML key to separate the Elektra keyname from the metaname. So `baz.boo/meta.default` stands for the `default` metakey on the key `baz/boo`. To a standard TOML parser it would mean the `default` field within the `boo/meta` table within the `baz` table.

The `[mykey.foo.baz.bar/meta]` example shows that we actually wouldn't use `/meta.` as a separator, but instead would give a special meaning to all TOML keys/fields that have `/meta` at the end.

I haven't tested this idea yet, but I think this may be a solution. We would of course need a way to escape the `/` in cases where the Elektra key actually contains a `/`, but that shouldn't be too hard. TOML has a Unicode escape mechanism that would allow writing `/` as `\u002F`, but other tools might translate this back into a `/`, which would cause confusion. A simple way to fix this would be to use one of Unicode's Private-Use code points e.g. `\uEABC` to represent an actual `/`.

If it actually works, this would be a good solution IMO, because normally you wouldn't use a `/` within a key name part anyway (because of the need for escaping within Elektra), so this would minimize the collisions and therefore the need for escaping within TOML files.

markus2330:
> We had endless discussions about restricting key names and you always were very much convinced that it would be a bad idea. Now you suddenly would accept restrictions?

I am still against APIs which impose such restrictions.
But if the restriction is only caused by the implementation anyone can go ahead and fix it.
I don't think this is release critical, we simply need to document it.

> Something like kdb set-meta user:/myarray array "#4" is needed to tell the spec plugin, how big an array should be. Otherwise it won't know, how many default keys it needs to create.

Yes, but this does not demand arbitrary meta-data, only support for one metadata: array.

> If it actually works, this would be a good solution IMO

I agree. Unfortunately Harald already needs a solution now. When would you be able to implement it? (Assuming @bauhaus93 does not have time now.)

kodebach:
> I am still against APIs which impose such restrictions.
> But if the restriction is only caused by the implementation anyone can go ahead and fix it.

Restricting the default implementation is almost the same as restricting the API. First, the parts of the API that are not implemented by default would be mostly untested and therefore likely cause bugs in various places. And second I doubt any application developer would actually want to rely on parts of the API that are not supported by default.

> Yes, but this does not demand arbitrary meta-data, only support for one metadata: array.

There could be other similar cases, where metadata is needed for some value to be generated. I don't know any of the top of my head, but again IMO the default plugin should support all possible cases.

> Unfortunately Harald already needs a solution now.

Is full support actually needed, or is there a limited use case that has to be supported?

> When would you be able to implement it?

Not sure, with testing and fixing other issues it could take a few weeks for sure until TOML is fully functional as a default plugin with full metadata support.

markus2330:
> Restricting the default implementation is almost the same as restricting the API. 

The difference is: Nobody is restricting anything on purpose here. In this case it would simply be an unresolved bug=limitation=restriction until a better and compatible syntax without this limitation would be found and implemented.

> Is full support actually needed, or is there a limited use case that has to be supported?

Of course it will be somehow limited but what will not be needed is very difficult to say at project start.

> Not sure, with testing and fixing other issues it could take a few weeks for sure until TOML is fully functional as a default plugin with full metadata support.

One possibility would be to skip the full metadata support for any namespace and instead reduce to support the best practices in Elektra: i.e. storing specs only in spec:.

kodebach:
> The difference is:

Let's not discuss the restrictions for now, since I may have found a way to support everything in TOML...

> Of course it will be somehow limited but what will not be needed is very difficult to say at project start.

Oh, the OpenWRT stuff is a new project? I thought we are talking about an update to an existing package (in which case there should be some known scope)

> One possibility would be to skip the full metadata support for any namespace and instead reduce to support the best practices in Elektra: i.e. storing specs only in spec:.

Since there is currently no way to store metadata in **any** namespace, restricting to `spec:` doesn't really help. In fact it might make testing a lot more complicated. The basic tests need to be duplicated for `spec:` and non-`spec:` keys to ensure everything works, and then we need extra tests for the special features of `spec:`.

Before we hastily decide to do something, we should ask Harald what exactly he needs. If the fix in #3901 is already enough for his use-case, then we can take our time and create a well-designed solution.

In the meantime I will try to find some time and add a quick&dirty test implementation of the idea above (https://github.com/ElektraInitiative/libelektra/pull/3478#issuecomment-871404740).

markus2330:
> Oh, the OpenWRT stuff is a new project?

Yes, it is a new project, also using OpenWRT.

> Before we hastily decide to do something, we should ask Harald what exactly he needs. If the fix in #3901 is already enough for his use-case, then we can take our time and create a well-designed solution.

I summarized the requirements for you:  #3909

> then we can take our time and create a well-designed solution.

The idea is not that we design something new but "only" to (quickly) fix bugs hindering the success of his project. TOML as replacement for INI was loudly announced...

> In the meantime I will try to find some time and add a quick&dirty test implementation of the idea above (#3478 (comment)).

:rocket: :guitar:

kodebach:
> TOML as replacement for INI was loudly announced...

Seeing as there is still and open PR for TOML and some known problems, maybe we jumped the gun on this one...

---

Also, I think we should close this PR and use #3500 for further TOML developments and discussions. To test TOML as a default storage plugin, I will add a build job that just configures TOML as the default (like `debian-buster-full-mmap-asan` does for mmap). I'll use Github Actions or Cirrus, so that we can ignore the failing job until TOML as the default actually works.

markus2330:
As long as we do not merge the changes in the Jenkinsfile, other builds are also not affected. I would prefer if you can add a job like debian-buster-full-mmap-asan in Jenkins.

The important thing now is to get rid of other nasty essential bugs. Spec (or metadata) is not a must-have, see #3909.

kodebach:
> The important thing now is to get rid of other nasty essential bugs. Spec (or metadata) is not a must-have, see #3909.

I honestly don't really care what the requirements are for some fancy _unfinished_ **external** library. Elektra is open-source, if somebody wants to use it, but it doesn't meet their requirements, they can always make a PR themselves.

For me the TOML thing is only an acceptable priority, because we want to make TOML our default as well and we already recommended it as a replacement for INI.

PS. I'll close this PR now, further discussion about TOML in #3500. Discussion related to libautomation in #3909.

markus2330:
> they can always make a PR themselves

This is not how it works. If you don't care as maintainer if your software works, nobody else will. We are only fixing what Elektra promises to do.

> For me the TOML thing is only an acceptable priority, because we want to make TOML our default as well and we already recommended it as a replacement for INI.

Exactly. Did you already test TOML as default locally a bit?

> I'll close this PR now. further discussion about TOML in #3500.

The PR was not for discussion, it has some content (=commits). Did you migrate all the relevant commits to #3500? At least the "maintained" in status seem to be missing. Please do not close PRs if we do not have PRs taking over the relevant commits.

kodebach:
> This is not how it works. If you don't care as maintainer if your software works, nobody else will. We are only fixing what Elektra promises to do.

I didn't say I don't care whether it works or not. I have no problem with fixing bugs that some found by using Elektra in an external package (e.g. #3896). I do however have a problem with adjusting our internal priorities for **new/unfinished** features (e.g. arbitrary metadata in TOML, or TOML as a default plugin) based on external needs. A quick test is fine (and I will do that for TOML), but I really don't want to further delay more important things (e.g. #3693).

> Did you already test TOML as default locally a bit?

I did not have time for that yet. I will do a quick test, but again, IMO using `toml` as the default plugin is not a high priority issue. We never stated that this would work and if you use Elektra "properly" (i.e. custom mountpoints) then the default plugin shouldn't really matter.

> The PR was not for discussion, it has some content (=commits).

Yes, I know... My point was that this PR is even less appropriate for this discussion. (Which is why I won't answer here anymore)

> Did you migrate all the relevant commits to #3500? At least the "maintained" in status seem to be missing.

I did not move anything to #3500, because #3500 is now in a working state, while the code in this PR is either not working or untested AFAICT. I will move the relevant pieces of code into a new PR once #3500 is merged.

> Please do not close PRs if we do not have PRs taking over the relevant commits.

This PR will never be merged and there won't be any more commits to this PR, so the correct thing was to close it. The code isn't lost, just because the PR is close. Even if the branch is deleted (could happen with the PR open too), I do have a local copy of the branch, just in case.

markus2330:
> I didn't say I don't care whether it works or not.

I also didn't say that I ever had the impression that you don't care, it is the opposite :wink:.

With opening/closing PRs we are also communicating something, and in my understanding it means that we either already have it in another PR/branch or we don't want the commits anymore.

> I do however have a problem with adjusting our internal priorities for new/unfinished features (e.g. arbitrary metadata in TOML, or TOML as a default plugin) based on external needs.

"Needs" is what should drive any library. It also can be your need, and then you also would be happy if we care about your use case.

> A quick test is fine (and I will do that for TOML),

Thank you very much! It is very good that TOML has more than one maintainer.

> but I really don't want to further delay more important things (e.g. #3693).

I agree that finishing the open PRs is also very important.

> The code isn't lost, just because the PR is close.

Even if it is not lost physically, it might be lost in the sense that we forget to merge it main. Open PRs are a reminder, like issues.

> Even if the branch is deleted (could happen with the PR open too), I do have a local copy of the branch, just in case.

This works in your case (as you are reliable and know how to find commits in your repos) but it does not work in general.

kodebach:
> It is very good that TOML has more than one maintainer.

I would definitely not consider myself a maintainer... I'm fixing things right now, because I've got time and that's it. I don't guarantee anything for the future

> This works in your case

Since, I'm the one currently working on this, the rest doesn't matter... If I don't have time to finish this, I will open an issue or another draft PR, etc.

PR #3491, opened 2020-09-16
Remove plugins

### LINKED ISSUE #3481: tcl plugin test fails

Compile and run `make run_all`. (Specifically `testshell_markdown_tcl`)

## Expected Result

Compilation and tests should be sucessful.

## Actual Result

```
Sorry, I crashed by the signal SIGSEGV
This should not have happened!

~/libelektra/cmake-build-debug/tests/shell/shell_recorder/shell_recorder.sh: line 267: 2999324 Aborted                 (core dumped) "$KDB" export spec dump > "$SPECEXPORT"
fatal: Could not export spec config
```

Upon closer inspection, it turns out that the error is caused by the `€`-symbol (and any other special characters) in the testdata.
Specifically somewhere in the Boost Qi library (`boost::spirit::qi`) an assert fails, because the special character is decoded as a negative integer which is not in the range of an `unsigned char`.

## System Information

- Elektra Version: master
- Operating System: Manjaro Stable Version, up-to-date as of 2020-09-05
- Boost Version: 1.72.0 (specifically 1.72.0-2 from the Manjaro extra repo)

### LINKED ISSUE #3483: remove null plugin
In order to simplify Elektra we want to get rid of trivial plugins, which have unintended side effects (like modifying unmodified keys). One candidate is the "null" plugin:

- its functionality is trivial
- it modifies keys (which is the main source of trouble)
- there are not many plugins using the plugins, it is only:
 - [ ] line (I can fix this)
 - [ ] simpleini (plugin can be removed)
 - [ ] tcl (plugin can be removed)
 - [ ] toml

@bauhaus93 can you remove the dependency of the null plugin and implement its functionality directly in the escaping/unescaping code? Doing so also has the advantage, that null values/empty strings will also work when toml is the default plugin. (binary values are probably no problem anyway as we do not write binary data to /elektra). If you can fix this in a separate PR (or together with other escaping fixes but not together with making toml the default storage plugin) this week, we could remove the null plugin already in the upcoming release.

markus2330:
Closes #3481 (`tcl` removed). 
Closes #3483 (`null` removed).

## Basics

- [X] Short descriptions of your changes are in the release notes
      contains `_(my name)_`)
- [X] Details of what you changed are in commit messages

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


markus2330:
@bauhaus93 When switching in the base64 test cases from INI to TOML I got some difference in behavior:

```sh
45: kdb mount test_config.toml user/tests/base64 toml base64 binary/meta=
45: kdb set user/tests/base64/encoded dmFsdWUA
45: kdb file user/tests/base64/encoded | xargs cat | grep encoded
45: kdb meta-set user/tests/base64/encoded type binary
45: kdb get user/tests/base64/encoded
45: 
45: ERROR - STDOUT:
45: “dmFsdWUA”
45: does not match
45: “\x76\x61\x6c\x75\x65\x0”
```

Do you have some idea why I suddenly get the encoded value and not the decoded one? Except of ini -> toml I did not change much: 75653e3

Btw. please push to the branch if you know the fix.

bauhaus93:
I looked further into this issue and I think it would be better, if the base64 meta mode should only be supported after the TOML plugin is able to write/preserve metakeys. Then it would be possible to handle the metamode base64 strings without any explicit checking of the base64 configuration.
I can look into implementing the metakey preservation in the next few days.

markus2330:
Thank you for the hint! I was not aware that it could be a problem of the meta mode.

> How would I check for the configuration of a mountpoint, to see if meta mode is enabled?

I am afraid that is not possible as the `binary/meta` is a private configuration to the base64 plugin. It is also not a good approach that plugins try to rely on implementation-details of other plugins. That the strings start with `@BASE64` (or anything else) is actually an implementation-detail. Why does TOML need to know this? If there is a reason, we should properly mark these keys with additional meta-data (e.g. `binary/encoded`).

bauhaus93:
> Why does TOML need to know this? If there is a reason, we should properly mark these keys with additional meta-data (e.g. binary/encoded

At least on reading it is necessary to check for strings with `@BASE64` prefix, so that the plugin doesn't add a `type` metakey of `string` to it (the type plugin would throw an error, because it's not a string after decoding).
The dependency on that prefix could probably be removed, when appropriate metakeys are set (and the TOML plugin can preserve them).

On writing, I check for `@BASE64` prefixes without any encoded data (`NULL` values). which I translate to `@NULL` strings. I thought that that would be better readable in a file.
I can revert that and make the plugin less reliant on that implementation detail.

markus2330:
Yes, I do not think that rewriting to `@NULL` is a good idea, it only complicates things and its more likely that someone knows what `@BASE64` than what `@NULL` is. (At least you can somehow guess that the docu of the base64 plugin will enlighten you. The null plugin is gone soon #3491.).

What about not giving strings a type at all? (@bauhaus93 Then the whole `@BASE64` detection problem would disappear. wouldn't it?) Afaik other plugins also do not do this. "doc/METADATA.ini" is very vague about this, it should be clarified there. (@kodebach do you have an opinion about that?)

kodebach:
> What about not giving strings a type at all?

AFAIK `type = string` means we expect a non-empty string value. As far as I'm concerned, `type = string` should not be set by `toml`. It should mostly just come from specifications. Sadly this is a hard case again, because a storage plugin should not add `type = string` during `kdbGet`, but it should also not remove it during a `kdbSet` -> `kdbGet` cycle. So if a user explicitly sets the `type = string` metadata that should be preserved, IMO.

> Yes, I do not think that rewriting to `@NULL` is a good idea, it only complicates things and its more likely that someone knows what `@BASE64` than what `@NULL` is. 

I'd actually find 

```toml
mykey = "@BASE64"
```

a lot more confusing than

```toml
mykey = "@NULL"
```

If `mykey` is supposed to be a key with value `NULL`.

---

As a suggestion for binary data in TOML specifically. `toml` already has the `\uXXXX` escape sequence in strings. So it should be possible to encode any non-printable byte or non-ASCII byte as the appropriate value in `\u0000`-`\u00FF`. The [TOML spec](https://toml.io/en/v1.0.0-rc.2#string) only says that the `XXXX` in `\uXXXX` must be a valid [Unicode scalar value](https://unicode.org/glossary/#unicode_scalar_value) (link from the spec). So that encoding should even be complaint with the TOML spec.

Of course that wouldn't change anything about the problem with `base64` plugin's test case.

@markus2330 You could just use `dump` for the test case. It doesn't make a lot of sense as an example, and might require a bit more modifications (because of the zero-bytes in the `dump` format), but it would avoid the problem that `toml` doesn't store metadata right now.

bauhaus93:
>What about not giving strings a type at all?

The only issue I see when not assigning the `string` type anymore would be that strings read from a TOML file, that also match a number/date value would be written back unquoted to the file, losing their string typing inside TOML.
However, this may be prevented by assigning such ambiguous values an `origvalue` metakey which also contains the enclosing quotation characters.

> Sadly this is a hard case again, because a storage plugin should not add type = string during kdbGet, but it should also not remove it during a kdbSet -> kdbGet cycle. So if a user explicitly sets the type = string metadata that should be preserved, IMO.

But wouldn't come that capability with the general preservation of metakeys?

> As a suggestion for binary data in TOML specifically. toml already has the \uXXXX escape sequence in strings. So it should be possible to encode any non-printable byte or non-ASCII byte as the appropriate value in \u0000-\u00FF. The TOML spec only says that the XXXX in \uXXXX must be a valid Unicode scalar value (link from the spec). So that encoding should even be complaint with the TOML spec.

Hm, yes this sounds like good way to store binary data inside the file, thanks. With this, `@NULL` could also be kept for NULL values.

kodebach:
> The only issue I see when not assigning the `string` type anymore would be that strings read from a TOML file, [...]

I should clarify, IMO you `toml` should not set `type = string`, because it has a more specific meaning than `toml` can infer from the TOML file. However, adding a `toml/type = string` to keep track of whether this was a string or not in the TOML file is totally fine.

> But wouldn't come that capability with the general preservation of metakeys?

Kind of, yes. You still need to keep track of what metadata comes from the file and what was generated by plugin. The file
```toml
mykey = 5
```
should produce some kind of `type` metadata (probably `type = long` to be on the safe side), but the next `kdbSet` shouldn't automatically write that `type` metadata back to the file, IMO.

> With this, `@NULL` could also be kept for NULL values.

If you can store metadata, adding something like `binary=null` to the stored metadata and storing an empty string might be better. Then you could avoid the escaping problems with `@NULL`.

bauhaus93:
> I should clarify, IMO you toml should not set type = string, because it has a more specific meaning than toml can infer from the TOML file. However, adding a toml/type = string to keep track of whether this was a string or not in the TOML file is totally fine.

Ah, ok, so I'd only set  a `type` of `string` for ambiguous  values.

> You still need to keep track of what metadata comes from the file and what was generated by plugin [...]

So I'd have to check, what type the plugin would derive on a read and what type is actually present on the key. And only if they're different, the present type should be explicitly written as metadata.

kodebach:
> Ah, ok, so I'd only set a type of string for ambiguous values.

The problem here is that `type` metadata is an instruction to the `type` plugin. Unless you actually want to give that instruction, you should not use it.

So, `toml` should never generate `type = string`, because that would imply that the value is required to be a non-empty string. But you only want to keep track of whether something was a string in the file or not. You probably need something like `tomltype = basicstring`, `tomltype = literalstring`, etc. similar to what you already do for tables, so you can use that. If you don't want to differentiate like that just use `tomltype = string`.

Sidenote: `tomltype` should probably be replaced by `internal/toml/type`, since it is only used by the `toml` plugin to keep track fo the original file structure. At least that's how I understood it...

> So I'd have to check, what type the plugin would derive on a read and what type is actually present on the key. And only if they're different, the present type should be explicitly written as metadata.

I am not sure anymore what the solution is... Everything I come up with has some kind of problem. There are so many factors at play here (the `spec` plugin could also be involved) that it is probably best you just implement whatever works and leave it at that until there is a real world problem...

markus2330:
> If mykey is supposed to be a key with value NULL.The only issue I see when not assigning the string type anymore would be that strings read from a TOML file, that also match a number/date value would be written back unquoted to the file, losing their string typing inside TOML.

I do not think it is a good idea trying to do any automatic inferring. If a user does `kdb set /blah 1979-05-27T07:32:00Z` it is simply a string. Only when the type is added it is something else. Reading from a file it is always clear anyway as quotes are needed for strings.

> However, this may be prevented by assigning such ambiguous values an origvalue metakey which also contains the enclosing quotation characters.

Which ambiguous values do you mean?

> Sidenote: tomltype should probably be replaced by internal/toml/type, since it is only used by the toml plugin to keep track fo the original file structure.

I fully agree!

> [arbitrary metadata] I am not sure anymore what the solution is...

The current solution is that plugins writing arbitrary meta-data are only mounted to `spec`. For the TOML plugin a solution could be that the feature is only activated if mounted to spec or the user explicitly activated it explicitly via a plugin config.

markus2330:
@PhilippGackstatter Can you help out here please and fix some tests so that "ini" finally can be removed? Replace with toml, or if not possible (e.g. the base64 test), with mini.

PhilippGackstatter:
```
The following tests FAILED:
	 38 - testshell_markdown_base64 (Failed)
	 39 - testmod_blockresolver (Failed)
	 44 - testshell_markdown_ccode (Failed)
	115 - testshell_markdown_shell (Failed)
```

I cannot reprocude `ccode` and `shell` tests locally :monocle_face:. I'll ignore them for now, until the others have been fixed.

@bauhaus93  Is the `base64` test supposed to be resolved when #3500 lands?

PhilippGackstatter:
> @bauhaus93 Is the `base64` test supposed to be resolved when #3500 lands?

Nevermind, replacing `toml` with the `ni` plugin resolved the issue :+1:

PhilippGackstatter:
The `ccode` test fails because `tcl` was removed.
Is the plugin still needed? Some plugins already do their own escaping, e.g. `yajl`, `ni`. But I suppose for more complex escaping `ccode` is still useful.

I tried various plugins, but somehow using them (e.g. `mini`) with `ccode` doesn't work as I thought it would.

```
kdb mount config.tcl user/tests/ccode mini ccode
kdb set user/tests/ccode/multiline "`printf 'one\ntwo\nthree'`"
grep 'multiline' `kdb file user/tests/ccode` | sed 's/[[:space:]]*//'

ERROR - STDOUT:
“multiline=one”
does not match
“multiline = one\ntwo\nthree”
```
@markus2330 can you give me a hint which plugin would be a good replacement for `tcl` here?

markus2330:
> [ccode] Is the plugin still needed?

Currently mini and c depend on it and simpleini can use it. But having the escaping in a storage library and directly use the storage library for escaping would be the better approach (as this also works for the names and not only the values). The code of the ccode plugin is probably useless as it is written in C++, so the escaping would need to be written from scratch or maybe we can borrow the code from @bauhaus93's TOML plugin. Is this something you would like to do?

> can you give me a hint which plugin would be a good replacement for tcl here?

Not really, I am afraid this test needs to be rewritten or removed.

Btw. we decided to not merge this for the upcoming release, so there is no hurry.

PhilippGackstatter:
> Is this something you would like to do?

So what you mean is extract the escaping from the `toml` plugin into a library that all storage plugins can integrate into their code?

If this is the most important issue that I can work on right now, I can do that.

If we do this, then `ccode` will eventually become obsolete. Then rewriting doesn't make sense and I can probably remove the test, since the plugin didn't change and `testmod_ccode` still passes.

markus2330:
Yes, @PhilippGackstatter please remove the test now from ccode to finalize this PR.

@bauhaus93 can you extract your escaping code from toml to some lib (you know where it is and maybe you can improve the API docu a bit so that others can use it more easily)?

bauhaus93:
Yes, I can do that. However, the toml plugin only has code for unescaping, so the escaping code I would have to write from scratch.

markus2330:
Maybe we take some other code then... I moved the discussion to #3526

So for now we simply remove the test case and consider the code plugins as deprecated.

PhilippGackstatter:
I cannot reproduce the memory leak in the `line` plugin using the `build-elektra-bionic` image. Since the only change to the plugin here was to change `infos/depends` from `null` to `binary`, I guess that's where the error originates from.
Is it possible to clone the exact docker images that were used in the travis or cirrus-ci builds? A quick search didn't bring anything working up for travis, and running the cirrus tool locally also failed.

@markus2330 @mpranj 
What is the usual way to reproduce the build system errors locally, to make testing easier?

mpranj:
> What is the usual way to reproduce the build system errors locally, to make testing easier?

Yes the best shot would be to pull the used docker image directly from our private hub. The second option is to build the image locally yourself. 

The problem is that even with the best option there can be differences depending on the host system (yes, unfortunately).

In this case, however, I guess the memleak is in the CI logs [here](https://travis-ci.org/github/ElektraInitiative/libelektra/jobs/739386084). I think the `jni` plugin does not properly clean up when the `pluginOpen` call fails. I hope 002d3855736ada4f554eae8ad151549c36510a8c fixes it (or one of the) problem(s).

mpranj:
Sorry, it seems my fix was wrong. I undid my changes for now.

markus2330:
@PhilippGackstatter we have a tutorial about how to build a docker image (pulling is currently not possible as it is a private hub) exactly as it is used on the build server and run the tests there: doc/tutorials/run_all_tests_with_docker.md

mpranj:
Actually, you can pull from the private hub, but you can not push. The following will work for example:
```
docker pull hub-public.libelektra.org/build-elektra-debian-buster:202010-d4cbd4f11138a4e3ead489506154579603c58292267506be3f6b675ee4e84b79
```
You find the used docker image in the Jenkins build logs. You just need to replace "hub" with "hub-public". This is documented in `scripts/docker/README.md` and `doc/BUILDSERVER.md`.

The problem is that the errors appeared on macOS afair, so there are no docker images for that.

markus2330:
I just saw that this is actually also described in the linked tutorial but it is not recommended as it takes longer and creates traffic on our server.

mpranj:
Looking at this again I think there are multiple problems (and _maybe_ my first fix was ok for one problem).

1) There seems to be a new problem with the python bindings:
```
    Start 34: test_kdb.py
1/9 Test #34: test_kdb.py ......................***Exception: SegFault  0.03 sec
````
I think I saw this appear somewhere else too, so it might not be this PR. Maybe problems with a new python update on homebrew?

2) There might be multiple memleaks uncovered, this is unclear to me.

```
==77943==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 1480 byte(s) in 1 object(s) allocated from:
    #0 0x10627bdbd in wrap_malloc+0x9d (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x46dbd)
    #1 0x7fff6cdaa674 in _duplocale+0x16 (libsystem_c.dylib:x86_64+0xb674)
    #2 0x7fff6cdaa927 in newlocale+0x9d (libsystem_c.dylib:x86_64+0xb927)
````

Something similar was mentioned a while ago https://github.com/ElektraInitiative/libelektra/pull/2201#discussion_r212228256. I don't fully understand what the outcome was back then, but the `jni` plugin is also mentioned there.

**Comment by PhilippGackstatter** [comment]
> The problem is that the errors appeared on macOS afair, so there are no docker images for that.

Exactly. I'm already using the build images to test Elektra locally. But I cannot reproduce the errors there.
The remaining errors in CI are in Travis-CI and Cirrus-CI, which, as far as I can tell, use their own images.

It's weird, the python issues are new today :thinking:, even though you didn't make any actual changes.

**Comment by mpranj** [comment]
@PhilippGackstatter I'll look at it too, but I moved my stuff to #3537 so I don't interfere with your work here.

**Comment by mpranj** [comment]
The python issues are on master too. The homebrew python packages have been updated recently.

**Comment by markus2330** [comment]
@mpranj can you create an issue about the python problems? Can you reproduce it? This might be relevant for 0.9.3?

**Comment by PhilippGackstatter** [comment]

**Comment by PhilippGackstatter** [comment]

**Comment by PhilippGackstatter** [comment]
Alright, the build servers are finally happy, I think this can be merged :slightly_smiling_face:.

File: doc/news/_preparation_next_release.md [hunk 1]
```diff
@@ -25,15 +25,20 @@ You can also read the news [on our website](https://www.libelektra.org/news/0.9.
...
-- <<HIGHLIGHT3>>
+- TOML plugin
+- <<HIGHLIGHT>>
+- Cleanup
 
-### <<HIGHLIGHT1>>
+### TOML
 
-### <<HIGHLIGHT2>>
+- Added the TOML plugin, which can read and write TOML files using flex and bison. _(Jakob Fischer)_
```
mpranj, line 34: Are all the TOML entries in the news doc a rebase/merge error?

mpranj (changes requested):
Huge thanks to @markus2330 and @PhilippGackstatter for cleaning up so much! It's looking great!

I only think the news have to be updated: 
- Please remove entries which are unrelated to this PR (and were already published in the 0.9.3 release).
- Some other changes were done which are not mentioned in the news: tcl removed, multifile does not do recursive updates, ...

doc/news/_preparation_next_release.md
PhilippGackstatter: Mh, yeah I may have re-introduced that during rebase. I'll clean that up.

markus2330: Thank you very much :sparkle:

mpranj (approved):
Thank you! I'll just wait for one specific build on travis to be green.

@PhilippGackstatter one question: how did you solve the memleaks in the end, it seems that you just added base64 when mounting? 👀

PhilippGackstatter:
> @PhilippGackstatter one question: how did you solve the memleaks in the end, it seems that you just added base64 when mounting?

I think that `kdb mount` may have a memory leak in the error case. For instance, in its shell recorder test, `kdb mount line` did not succeed (return code 1), because to work correctly, it needs another plugin mounted that is a `binary` provider. And that seemed to have caused the memory leak. Once I added `base64`, `kdb mount` succeeded and the memory leak was gone.

mpranj:
> I think that kdb mount may have a memory leak

Thank you for the answer! I don't really get how it searches for providers of e.g. `binary`. What I observed before your fix were seemingly random memleaks in all kinds of plugins. Does kdb mount simply open all plugins until it finds a binary provider, or how does this work @markus2330?

markus2330:
Thank you both for your great work!

Yes, exactly, src/libs/tools/src/plugindatabase.cpp iterates over all plugins and searches for the best providers (based on the rating system with `infos/status`). But this happens also in the successful case, so the memleak is something else. I do not think we need to investigate this further, we should rather give tips how to succeed with an PR: one one important tip is to first make all test cases pass before trying to fix memleaks. What do you think of the change in 0c92788c8ff5ead18172dfa3124907eafd6a56d6?

mpranj:
Thank you for explaining! I agree with the change you provided in 0c92788. In this instance, however, the unit test was passing without ASAN. Furthermore it only complained about the return value not being 0, when run with ASAN. (afaik ASAN modifies the return value if problems are found...)

Issue #3490, opened 2020-09-16
TOML spacing


markus2330:
I played around with the new TOML plugin and must say it works beyond expectations, it is definitely superior to ini. The only things I noticed so far are some texts in error messages:

- [x] we should not expect people to know what rhs or lhs is

and spacing:

- [ ] a practical problem is that some INI parsers (in particular everything related to shells) do not accept spaces around `=`, see also #2065. Ideally, the parser would remember what spaces are around `=` and restore it as it was.
- [ ] a cosmetic problem is that spaces before the key name get removed. Probably nearly nobody would do this, if the example on https://toml.io/en/ would not promote this usage. But as it is, people quickly find out, e.g. #3472. Also here the ideal solution would be the restore it as it was.

String quoting seems to be a bit buggy (You will probably catch these things with `kdb test`):

- [x]  `path = 'C:\Users\nodejs\templates'`in a TOML file gets after roundtrip rewritten to `path = "C:\Users\nodejs\templates"` which leads to a parse error.

- [x] Some rare features like `1e06`numbers seem to be not supported. (The parser currently thinks it is a bare string.) Imho rather low priority.

Something else:

- [x] Keys with empty names seem to be currently not supported. In Elektra you can represent them by giving them an empty base name. In TOML you can represent them with
```toml
"" = "blank"     # VALID but discouraged
'' = 'blank'     # VALID but discouraged
```
(Btw. I think this example within the TOML spec is broken as another example says that multiple keys with the same name is not allowed.)

I do not think that any of these problems is urgent enough that it must be fixed for the next release. If I could choose something, I think the error messages are the most important thing. It might be good, however, that the current gaps to the TOML standard are documented in the README.md (and which version it should be compliant to.)

So in general I am very positive about this plugin, it greatly improves Elektra :sparkling_heart:  I'll remove the ini plugin, as I am very positive that soon the toml plugin will cover essentially every use case the ini plugin fulfilled but unlike the ini plugin, it will be properly maintained.

bauhaus93:
> I played around with the new TOML plugin and must say it works beyond expectations, it is definitely superior to ini. The only things I noticed so far are some texts in error messages

Thanks, I'm glad it works so well! I will look through all error messages and make them more clear.

 > a practical problem is that some INI parsers (in particular everything related to shells) do not accept spaces around =, see also #2065. Ideally, the parser would remember what spaces are around = and restore it as it was.
[...]

PR #3500, opened 2020-09-23
Toml fixes

bauhaus93:
Fixing TOML issues mentioned in #3490.

- [x] Make error messages less technical
- [x] Fix floats in exponential form (didn't allow leading zero after exponent, but is now allowed)
- [x] Allow empty keynames, which are valid but discouraged
- [ ] Write info in README about which TOML standard is supported / what the current gaps are
- [ ] Preserve whitespace in front of key names / around assignments
- [ ] Proper string quoting on writing, don't assume just `basic`. First tries of that in #3478.

Additional fixes:
- [x] Fix multiline string reading. There was a problem, when a string contained non-terminating quotation characters. Also fixed function to not loop endlessly when encountering EOF during reading.
- [x] Check, if non-decimal conversion functions are handling overflows properly (They didn't, so added checks if a base 2/8/10/16 integer over/underflows).
- [x] More strict handling of truth values on writing.
- [ ] Fix parser so that some invalid TOML files don't get parsed as valid. (Happens on invalid simple table/tablearray combinations or redefintions).
- [ ] Add metakey preservation capabilties.

Metakey preservation problems:
- Metakeys assigned to a table array root key. The table array root key (which holds the `array` and `tomltype` metakey of an table array) doesn't have any anchors in a TOML file.
- Metakeys assigned to inline table elements (since no comments are allowed in direct inline table elements by the TOML standard). This is a general comment problem. Either we change an existing `inlinetable` to a `simpletable` or we may throw an error in that case.

## Basics

      contains `_(my name)_`)

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


markus2330 (approved):
Looks very good, thank you for the improvements of the error messages. The unit tests are very nicely written, too.

bauhaus93:
@markus2330 
I've added a for now very basic implementation of metakey preservation in this PR (Metakey names assumed to have no spaces - I  have to find a proper way/format to differentiate between metakey name and value in the comment string; Also, special characters/escape sequences are not yet handled). 

But at least the failing example from  #3491 with the base64 metamode should work now as intended (without any need of knowing that metamode is on).

markus2330:
Thank you so much, great work :fireworks:

mpranj:
@bauhaus93 thank you for implementing so many improvements to the TOML plugin. This is even more important than #3098. 😉

kodebach:
@bauhaus93 @markus2330 Currently `toml` uses the `tomltype` metadata. I think it would be a good idea to generalize this into `type/[plugin]`, where `[plugin]` is a storage plugin. Very few storage formats have a type system that exactly matches Elektra's, so storing the storage format's type separately makes sense. For example a JSON plugin would use one of these:
```
type/json = string
type/json = boolean
type/json = number
type/json = null
```
(`array` and `object` aren't needed, since the are inferred from `KeySet` structure and JSON doesn't have things like TOML's inline tables)
The corresponding `type` values would be:
```
type/json = string -> type = any
type/json = boolean -> type = boolean
type/json = number -> type = long_long [OR] type = double
type/json = null -> type = any
```
(`string` corresponds to `any` because in Elektra `string` means non-empty)

markus2330:
Please avoid making new proposals in PRs, this could be easily misunderstood. Status-quo is:

- In the cases the types match (which is not soo rare) `type` should be used.
- In cases where types cannot be mapped, `internal/<plugin>/*` (in this case `internal/toml/type` should be used), see:

https://github.com/ElektraInitiative/libelektra/blob/f370cc54741f18aea44a29c07da4cd49eecc25b6/doc/METADATA.ini#L1072-L1074

and here also the meta-spec for type:

https://github.com/ElektraInitiative/libelektra/blob/f370cc54741f18aea44a29c07da4cd49eecc25b6/doc/METADATA.ini#L223-L244

If unsupported types are encountered during `kdbSet`, please give the user an error.

kodebach:
Yes, `internal/toml/type` would also be a possibility. But even in this case, we should probably define in `METADATA.ini` that `internal/<plugin>/type` should be used exclusively for this purpose to avoid possible confusion.

markus2330:
Thank you, I clarified in d198a7e0d7f8af0684f3969cd59142f9b99e55ef

src/plugins/toml/meta.c
```diff
@@ -0,0 +1,115 @@
...
+}
+
+int assignMetakeyFromComment (Key * key, const char * comment)
+{
+	ELEKTRA_ASSERT (isMetakeyComment (comment), "Comment should already have been checked for metakey eligibility");
+
+	const char * name = comment + sizeof (METAKEY_COMMENT_PREFIX);
+	const char * nameEnd = strchr (name, ' ');
+	const char * value = nameEnd + 1;
+	if (value != NULL)
```
kodebach, line 94:
```suggestion
	if (nameEnd != NULL)
```

`value` can't be `NULL`

kodebach:
@markus2330 I think this PR is now in a working state. My suggestion would be to merge this and then I'll use new PR(s) to fix the remaining TOML issues (see #3910). If we do a release soon, we should wait until after the release with the merge, or remove the semi-functional comment-based metadata handling from the PR.

markus2330:
Yes, the idea is to release asap so that Harald can continue with his work, see #3908. I think its best if we remove the "semi-functional comment-based metadata handling from the PR" as the specs in TOML is not urgent and having a otherwise working TOML is important&urgent.

kodebach:
I think this is now ready to be merged (assuming the build servers are happy).

I removed the metadata handling for now (new handling will be added in a new PR). I did not yet copy the string quoting stuff from #3478, because I found some issues with the existing code. I will add a new version in another PR.

mpranj:
Thank you so much @bauhaus93 and @kodebach! ❤️

PR #3447, opened 2020-05-30
Keyname overhaul

kodebach:
Continuation of #3115 (commit squashed and rebased)

Implements #3102 

Closes #3050
Closes #3084
Closes #2698
Closes #3082

## Basics

These points need to be fulfilled for every PR:

- [ ] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [ ] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy.
- [ ] The PR is rebased with current master.

## Checklist

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

PR #3507, opened 2020-10-05
New version of dump

### LINKED ISSUE #3506: New version dump plugin
Merging of #3447 will actually break **every** file stored with `dump`, since it (incorrectly) stores absolute key names. The fix would be easy for most files (just add the new `:`). But it's almost impossible for the fix to cover all cases.

This might be the ideal moment to create a new version of `dump`. My goals for this new version would be:
- Store relative key names, so it behaves correctly for exports. (simple fix)
- Get rid of some of the useless parts of the file structure. (remove the `ksNew` and `ksEnd` lines)
- Ensure the files are printable, as long as all key values are printable. (very easy fix)

This would also make `dump` a very good candidate for most of the `testshell_markdown_` tests. It is a very simple format, but still stores all the information a `KeySet` contains. The only problem with the current version is that it store the null-terminator bytes of key names and key values, which makes the markdown tests hard.

kodebach:
As discussed in #3506, this implements a new version of the `dump` plugin.

It should be merged _before_ #3447 to improve compatibility.

Closes #3506

## Basics

      contains `_(my name)_`)

## Checklist

      (not in the PR description)

## Review


kodebach:
I had to add some workarounds to make the `pluginprocess` stuff work...

Also if there is some way of forcing an update of dump files in the KDB, that would be nice (could we set the sync flags for keys in the old format?). It would make the transition to the new key names much easier.

kodebach:

src/libs/elektra/keyset.c
```diff
@@ -1560,7 +1560,7 @@ Key * ksAtCursor (KeySet * ks, elektraCursor pos)
 {
 	if (!ks) return 0;
 	if (pos < 0) return 0;
-	if (ks->size < (size_t) pos) return 0;
+	if (ks->size <= (size_t) pos) return 0;
```
markus2330, line 1563: This looks like a major bug, did you also add a test case?

src/plugins/dump/README.md
```diff
@@ -17,62 +17,114 @@ featured backend. No other plugins are needed.
...
-out the way that the KeySet needs to be constructed. The dump plugin
-interprets such a file. The file also looks similar to C code that
-would create the KeySet. Keys can contain any binary values and arbitrary
-metadata and are still stored and parsed correctly. The dump plugin can
-even reconstruct pointers to metadata to save memory. When a pointer
-to the same region of memory is found, a special command `keyCopyMeta`
-is written out that is able to reconstruct the data structure the way
-it was before. The commands were designed to make parsing of the file
-an easy task.
+The file starts with the magic word `kdbOpen` followed by a version number (currently `2`) and a newline.
```
markus2330, line 20: Why not let it start with `$ks` or similar? It fits better to the new syntax.

src/plugins/dump/README.md
```diff
@@ -17,62 +17,114 @@ featured backend. No other plugins are needed.
...
+
+`<type>` is either `string` or `binary` and indicates what kind of value the key contains.
+`<nsize>` and `<vsize>` are the name size and value size respectively.
+For the name and for string values, the size does not include the null-terminator present in C strings.
+For binary keys all bytes (including any null-terminators) are counted.
+`<name>` and `<value>` are the keyname and key value.
+Because we know their length, they can contain arbitrary characters.
+Even newlines are allowed.
+The newlines between `<name>` and `<value>`, and after `<value>` are just to make the file more readable.
+The must be present, but they do not determine where `<name>` or `<value>` end.
```
markus2330, line 49:
```suggestion
They must be present, but they do not determine where `<name>` or `<value>` end.
```

src/plugins/dump/dump.cpp
```diff
@@ -78,125 +111,347 @@ int serialise (std::ostream & os, ckdb::Key *, ckdb::KeySet * ks)
...
-				os.write (static_cast<const char *> (ckdb::keyValue (ret)), ckdb::keyGetValueSize (ret));
-				os << std::endl;
+				os << "$copymeta " << keyString (ret);
 			}
 		}
-		os << "keyEnd" << std::endl;
-	}
-	os << "ksEnd" << std::endl;
 
+		os.flush ();
```
markus2330, line 118: Is this really needed? (os seems to be closed shortly afterwards anyway, which will do the flush)

src/plugins/dump/dump.cpp
```diff
@@ -78,125 +111,347 @@ int serialise (std::ostream & os, ckdb::Key *, ckdb::KeySet * ks)
...
+	}
 	ksDel (metacopies);
 
+	if (markEnd)
```
markus2330, line 122: I don't like that it is sometimes written and sometimes not. I would prefer if it is always written. Then streaming always works.

src/plugins/dump/dump.cpp
```diff
@@ -78,125 +111,347 @@ int serialise (std::ostream & os, ckdb::Key *, ckdb::KeySet * ks)
...
+
+int unserialiseVersion2 (std::istream & is, ckdb::Key * parentKey, ckdb::KeySet * ks, bool useFullNames)
+{
+	ckdb::Key * cur = nullptr;
+	std::string line;
+
+	std::string rootName (keyName (parentKey));
+	rootName += "/";
+
+	if (useFullNames)
```
markus2330, line 269: I do not like this option, why not always use relative names? (And iterate over all namespaces where needed, like with pluginprocess.)

markus2330 (changes requested):
Thank you, amazing job! Did you try to run `kdb test`? (The test suite by default does not run the longest tests.)

markus2330:
> Also if there is some way of forcing an update of dump files in the KDB, that would be nice (could we set the sync flags for keys in the old format?). It would make the transition to the new key names much easier.

`kdb plugin-check` (without plugin args, with `-f`) does something like this. But this functionality should be moved from `kdb plugin-check` to a separate tool (it ended up there because originally the tool was called `kdb check`). And yes, it should set the sync flags on all keys, the workaround with the `Key b` does not work in general. I also think it will be useful for transition.

src/libs/elektra/keyset.c
kodebach: It doesn't actually change anything, because AFAIK the way `KeySet`s are allocated means that, if `ks->size == 0` then `ks->array[pos] == 0`, so the return value doesn't change.

src/plugins/dump/README.md
kodebach: Because `kdbOpen` was already used as a magic string for the existing version number check. This means that we get a nice error message about the version number from the old plugin. `$ks` would cause an error message about an unknown command, that also suggests the file might not be `dump` at all.

src/plugins/dump/dump.cpp
kodebach: It probably isn't needed. I added it when trying to work out, why `pluginprocess` just hangs (which was because of the missing end marker).

src/plugins/dump/dump.cpp
kodebach: You are right, it is probably best, if we always write it. But I'll keep the read part, such that it is optional and only stops processing of file early.

src/plugins/dump/dump.cpp
kodebach: The problem is that `pluginprocess` wants to send a `KeySet` with `Key`s that are **not** below the `parentKey`. This is very weird and I am totally in favour of changing it, but I didn't want to touch the convoluted `pluginprocess` code more than necessary.

I intentionally left this option undocumented, so that it could be removed easily once `pluginprocess` is implemented properly. Ideally `pluginprocess` should switch to the much faster `quickdump` anyway, but that requires changes to `quickdump` to make streaming possible.

> And iterate over all namespaces where needed, like with pluginprocess.

Not sure what you mean...

src/libs/elektra/keyset.c
mpranj: The edge case here is when `ks->size == pos`, which would be the last element of the keyset array. The last element is always `0`, regardless of `ks->size`, which is why the bug never took effect.

I haven't tested it but the change looks correct to me as well.

src/plugins/dump/testmod_dump.c
```diff
@@ -6,118 +6,589 @@
...
+	}
+
+	elektraFree (infile);
+	elektraFree (outfile);
+	ksDel (ks);
+}
 
 int main (int argc, char ** argv)
 {
-	printf ("MOUNT       TESTS\n");
```
mpranj, line 110: 🙈

mpranj (approved):
Wow, thank you for implementing so many improvements and also thanks for the review.

It's nice that dump finally has some test cases as well. I especially like the neat format version upgrade.

Great work!

mpranj:
Seems like the macOS tests are failing due to dependency installation problems.

kodebach:
> Seems like the macOS tests are failing due to dependency installation problems.

Yes, I saw that too. Not sure however, what's going on with Jenkins...

mpranj:
> Not sure however, what's going on with Jenkins...

It should have applied plugin updates while there are no builds, but it seems that Jenkins did not care that builds are ongoing.

I triggered a rebuild manually now.

src/plugins/dump/README.md
markus2330: For me a nice and consistent syntax is more important than some error message in a legacy system. I also somehow doubt that the new-style dump files will hit old versions of Elektra frequently.

src/plugins/dump/dump.cpp
markus2330: This was an implementation hint how the pluginprocess communication needs to be rewritten. So instead of sending a single KeySet, we would need a protocol outside of KeySets which tells from which namespace the next KeySet will be. Such a protocol would have been needed for all other plugins that are relative anyway. And now all plugins are relative #51. (Actually this PR really closes #51.)

But yes, it can (probably should) be done in a different PR. Can you open an issue about it?

kodebach: I'll open a new issue. `pluginprocess` probably should be rethought entirely. It probably shouldn't rely on a storage plugin to write the `KeySet` it wants to send. It should just write/read the keys directly. Maybe we could even use `mmapstorage` with a shared memory approach and avoid the pipes entirely. The pipes seem to cause problems on macOS anyway (there is another workaround for that in `dump`).

markus2330: I like the idea with mmapstorage! @mpranj what do you think?

markus2330: Discussion moved to #3509

mpranj:
Rebasing from master should help resolve the CI problems.

kodebach:
@mpranj I just noticed I forgot to check something, I will tell you whether this PR can be merged or needs another change tomorrow.

kodebach:
@mpranj you can merge this PR now.

I did find I way to force `kdbSet` to call `elektraDumpSet` for all `dump` mountpoints. So unless you know of a way to do that, config files will only be update, when the config actually changes. Otherwise there shouldn't be any issues.

mpranj:
> So unless you know of a way to do that, config files will only be update, when the config actually changes.

This should be doable but I don't know if it is worth the effort.

Thank you so much for the improvements @kodebach! This was much more effort than I anticipated. Great work!

PR #3513, opened 2020-10-11
Fix website dependency injection issue created by minification

### LINKED ISSUE #3508: website: pages do not render properly (SHOWONGITHUB)
E.g. https://www.libelektra.org/plugins/readme shows on a otherwise empty page (with Firefox ESR+Chromium):

```
{{ 'APP.WEBSITE.LISTFILES.BUTTON.SHOWONGITHUB' | translate }}
```

Did we introduce a syntax error somewhere on this page?

@Namoshek any ideas?

Namoshek:
This PR fixes #3508. Most likely the issue has been minification of the JavaScript sources, which can lead to changed variable names and therefore broken dependency injection. Compare
```js
module.exports = function ($http) { ... }
```
where `$http` can be renamed by the minifier (e.g. to a shorter name like `a`) with
```js
module.exports = ["$http", function ($http) { ... }];
```
where this may also happen, but only to the function argument and not the string.

The code now basically uses a special syntax which lets angular know that all items in the array except the last one should be resolved from the dependency container and injected as arguments to the last entry, which is the module function. The `ng-strict-di` directive on the root element in `index.html` additionally forces that only this type of dependency injection is allowed and will shout immediately when the old (and dangerous) syntax is used.

---

Additionally, I generated a `package-lock.json` with a set of package versions which produced a successful build. Providing such a lock file ensures that builds (for individual commits) are repeatable. If no lock file is provided, npm will resolve the most recent matching package versions, which can lead to broken builds if package maintainers do not respect semver. The versions in the lock file should be updated once in a while though, to pick up latest security patches. Currently only one security issue is reported, although it shouldn't affect us the way we use the library.

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -205,7 +205,8 @@ The website is generated from the repository, so all information about
 plugins, bindings and tools are always up to date. Furthermore, we changed:
 
 - Renamed [website](/src/tools/website) and removed its backend. _(Markus Raab)_
-- <<TODO>>
+- Use strict dependency injection for website modules. _(Marvin Mall)_
+- Added `package-lock.json` to ensure repeatable builds. _(Marvin Mall)_
 - <<TODO>>
 - <<TODO>>
 
--- a/src/tools/website/package-lock.json
+++ b/src/tools/website/package-lock.json
@@ -1,0 +1,5468 @@
+{
+    "name": "website",
+    "version": "1.0.0",
+    "lockfileVersion": 1,
+    "requires": true,
+    "dependencies": {
+        "@iamadamjowett/angular-logger-max": {
+            "version": "1.2.3",
+            "resolved": "https://registry.npmjs.org/@iamadamjowett/angular-logger-max/-/angular-logger-max-1.2.3.tgz",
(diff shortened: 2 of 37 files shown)


mpranj:
@Namoshek thank you so much for fixing this so quickly. Great work!

I added 06e85a5ee38c9882d5e1de3ebdf2b35d9fe9c75f, because our Jenkins aborts early if there are no release notes. Feel free to adapt them as you like.

markus2330 (approved):
All relevant jobs passed (Travis push passed only PR is pending). I think we can merge this to get our Website fixed. Thank you so much, great job :heart:

mpranj:
Thank you so much for fixing this!

Namoshek:
@mpranj Thank you for adding a changelog description, I wasn't aware it being missing will break the build. The additions of you are just fine.

To be honest, I was a bit in a hurry in the morning and wanted to get this out, so the website can be fixed. The PR template looked a bit too sophisticated for this simple fix. 😄

markus2330:
Everything works again, thank you so much! :fireworks: 

> To be honest, I was a bit in a hurry in the morning and wanted to get this out, so the website can be fixed. The PR template looked a bit too sophisticated for this simple fix.

The next time you can write something like "Saved the day by fixing the website" 😄

PR #3537, opened 2020-10-29
Remove plugins mpranj

mpranj:
Clone of #3491 for other testing.


mpranj:
The ASAN leaks with `testshell_markdown_line` are weird. When I run CMake with `ENABLE_DEBUG=ON` the leaks disappear.

Without `ENABLE_DEBUG` I only get unusable output:
```
80: —— Protocol ————————————————————————————————————————————————————
80: CMD: kdb mount line /tests/line line
80: RET: 1
80: === FAILED return value does not match expected pattern 0
80: STDERR: 
80: =================================================================
80: ==173238==ERROR: LeakSanitizer: detected memory leaks
80: 
80: Direct leak of 328 byte(s) in 1 object(s) allocated from:
80:     #0 0x7f783a4c03cf in __interceptor_malloc (/lib64/libasan.so.6+0xab3cf)
80:     #1 0x7f7833225149  (<unknown module>)
80: 
80: Direct leak of 144 byte(s) in 6 object(s) allocated from:
80:     #0 0x7f783a4c03cf in __interceptor_malloc (/lib64/libasan.so.6+0xab3cf)
80:     #1 0x7f7833c40560  (/home/mpranj/workspace/libelektra/build/lib/libelektra-kdb.so.0.9.3+0xdf560)
80:     #2 0xc871399780f505ff  (<unknown module>)

```
(Output was truncated)

Any ideas here? Do you think this is also just related to `RTLD_NODELETE`?

markus2330:
> Any ideas here? Do you think this is also just related to RTLD_NODELETE?

Probably we should simply recommend (or even enforce in the build system) that ENABLE_DEBUG and ENABLE_ASAN are not set together?

Is there a use case to have ENABLE_DEBUG and ENABLE_ASAN at the same time?

mpranj:
> Probably we should simply recommend (or even enforce in the build system) that ENABLE_DEBUG and ENABLE_ASAN are not set together?

Did you misread or did I misunderstand something? When ENABLE_DEBUG and ENABLE_ASAN are both on, the errors are gone.

markus2330:
I misread, then the other way round: enforce ENABLE_DEBUG when ENABLE_ASAN is on. Might make sense anyway: https://github.com/google/sanitizers/issues/89

But isn't the current setup in master that all asan tests run without debug? What did suddenly reverse the situation?

mpranj:
> What did suddenly reverse the situation?

Nothing changed regarding ENABLE_DEBUG and ENABLE_ASAN. Some plugin changes here uncovered new "memleaks" or false positives. That's what we are investigating.

If we can enforce ENABLE_DEBUG when ENABLE_ASAN is on (or at least have RTLD_NODELETE), the errors seem gone anyway. So I guess they are false positives?

markus2330:
Yes, looks for me also like this.

So adding RTLD_NODELETE seems to be reasonable. If you realize it by adding enabling ENABLE_DEBUG or only by adding RTLD_NODELETE is up to you but I think we should be transparent about what we do and write that we enable RTLD_NODELETE and avoid vague statements like "optimize for developing with Elektra" in ENABLE_DEBUG, better is to change this text to "enable RTLD_NODELETE".

mpranj:
> I think we should be transparent about what we do and write that we enable RTLD_NODELETE and avoid vague statements like "optimize for developing with Elektra" in ENABLE_DEBUG

You're right. I think the idea was to add more stuff to ENABLE_DEBUG, thus it was so generic.

I would prefer to just add/force RTLD_NODELETE with ENABLE_ASAN "directly" (if it makes sense anyway). It does not have to be combined with ENABLE_DEBUG.
I will also update the message of ENABLE_DEBUG to be more clear.

markus2330:
> I would prefer to just add/force RTLD_NODELETE with ENABLE_ASAN "directly"

Yes, is maybe better. It can be very confusing if the cmake options turn on/off others.

> I will also update the message of ENABLE_DEBUG to be more clear.

Thank you :sparkling_heart:

src/plugins/jni/jni.c
```diff
@@ -321,8 +321,14 @@ int elektraJniClose (Plugin * handle, Key * errorKey)
...
 
-	if (!data || data->module == 1)
+	if (!data)
+	{
+		return 0;
+	}
+
+	if (data->module == 1)
 	{
+		elektraFree (data);
```
mpranj, line 331: @markus2330 this is one instance where I think it was not cleaned up properly. The leak was triggered when Elektra searched for the best provider for `binary`, which should be `base64`.

In any case: do you agree with the changes?

markus2330 (approved):
Thank you for the improvements! :+1:

PR #3549, opened 2020-11-03
Decisions

markus2330:
Finally, the outcomes from the last meeting. Not perfect yet but we can continue our discussions.

## Basics

- [X] Short descriptions of your changes are in the release notes
      contains `_(my name)_`)

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


doc/decisions/arbitrary_metadata.md
```diff
@@ -6,6 +6,8 @@ To make storage-plugins suitable for `spec` they need to be able to store
 all the metadata as specified in [METADATA.ini](/doc/METADATA.ini).
 Most file formats do not have support for that.
 
+If metadata is merged from different sources, metadata from spec might end up somewhere else.
```
kodebach, line 9: This sentence is a bit unclear to me.

doc/decisions/arbitrary_metadata.md
```diff
@@ -19,6 +21,8 @@ Most file formats do not have support for that.
 
 ## Decision
 
+Do not store metadata in any namespace but spec.
```
kodebach, line 24: Are we sure that this works? I'm still not convinced it works (losslessly) for arrays.

Also what happens for e.g. `kdb set-meta user:/myarray mymeta 5`?

doc/decisions/array.md
```diff
@@ -24,6 +24,10 @@ an array or not.
 ## Decision
 
 Store length in metadata `array` of key, or keep metadata `array` empty if empty array.
+Only children that have `#` syntax are allowed.
```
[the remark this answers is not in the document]
markus2330: I clarified in 04f470537001b885adc8507a263b333b74caea3c

> We should also state that mixed array and non-array keys could be considered "Undefined Behaviour"

Why saying it? If we do not define behavior, it is obviously undefined. And we do not leave it open on purpose, we simply do not yet know a use-case of this. Once we know, we can define something appropriate.

doc/decisions/array.md
```diff
@@ -24,6 +24,10 @@ an array or not.
...
+The index start with `#0`.
+Both `keyAddName("#12")` or `keyAddBaseName("#_12")` is allowed to add the 13th index.
```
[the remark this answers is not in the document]
markus2330: Yes, exactly!

doc/decisions/boolean.md
```diff
@@ -6,46 +6,47 @@ Inconsistent use of bool in various parts of Elektra.
...
 
-- 0 is false, everything else is true (default is true)
-- 1 is true, everything else is false (default is false)
+In the absence of the key, the default can be either:
 
-Example:
+- default is true:
+  `0` is false, everything else is true, or
+- default is false:
+  `1` is true, everything else is false.
```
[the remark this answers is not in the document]
markus2330: Yes, accepting only 0 and 1 is totally legitimate.

Only applications being too lax is problematic, as it leads to undetected misconfiguration (e.g. the admin thinks it will be true but the application actually interpreted it as false). I made it even more strict: 9b40823114cbcc91ba73ad9efef103427c229ce3

doc/decisions/ensure.md
```diff
@@ -0,0 +1,39 @@
...
+Keep kdbEnsure but:
+
+- reduce for only global plugins, as the partial other functionality is confusing
+  and not needed
+- find solution that list plugin is not needed
+- only as implementation detail below libraries (e.g. like done for notification)
+
+## Rationale
+
+- allows interfaces that let the user do `kdbOpen`
```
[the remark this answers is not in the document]
markus2330: I clarified in f8d940d0da47d60c4d5f56d8f2740a3e5fc6d188

Yes, I also would have liked it in kdbOpen more but it seems unrealistic to me that this gets implemented as it needs heavy changes in the notification API. And there are also advantages to `kdbEnsure`, it is more flexible :wink:

doc/decisions/semantics_name.md
```diff
@@ -19,15 +19,29 @@ The alternative would be to have semantics in key names, with following advantag
...
 The `spec` plugin guards these rules.
 
 ## Rationale
 
+- for consistency, whenever possible, meta-data should be preferred
+- no escaping of key base names necessary
+- it is very unlikely that the 9-byte UTF-8 sequence `®elektra` collides with a real key base name
+  a user wanted to have
+- `®elektra` makes very clear that there is a special reserved meaning
+
```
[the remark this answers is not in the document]
markus2330: I added this information in c8f6a0a30c6183e52cbcd2bce626052de7e33a48

doc/decisions/arbitrary_metadata.md
markus2330: I hope I could clarify in 3c1c5bf2649d3a5e0d8ae3a61283846f30375492

doc/decisions/arbitrary_metadata.md
markus2330: I think it should work, I clarified the array case in 349343cb07a219752bddfe7091c23c3a68d2fd49

Did you have something else in mind where it would break?

markus2330:
@mpranj @kodebach which are the decisions where you cannot agree in the current form (as in this PR) or where you think it would be a mistake to decide it without further considerations/discussions?

doc/decisions/array.md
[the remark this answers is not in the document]
kodebach: I should be fine in the decision document, but we should state somewhere in the user-facing documentation should contain a note. Something like:

Mixing array and non-array keys is not supported and might cause problems. In many cases the trivial solution is to move the array part into a separate child-key.

doc/decisions/ensure.md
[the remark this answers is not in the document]
kodebach: I really don't like the way `kdbEnsure` works right now. We set up the default case in `kdbOpen` only to potentially redo everything in `kdbEnsure`. Maybe the solution is an API like this:

```c
KDB * kdbOpen (Key * parent);
int kdbConfigure (KDB * handle, KeySet * contract, Key * parentKey);

KDB * kdbOpenDefault (Key * parent);
```

The new `kdbOpen` only does the absolute minimum work, in particular it doesn't set up any global plugins. If you use `kdbOpen` you _must_ call `kdbConfigure` otherwise `kdbGet` will fail. `kdbConfigure` configures global plugins (basically just a renamed `kdbEnsure`). Lastly, `kdbOpenDefault` does more or less what the old `kdbOpen` does. It sets up the default case and you can call `kdbGet` immediately. _But_ you _cannot_ call `kdbConfigure` after `kdbOpenDefault`.

(The names are just suggestions)

---

The high-level API is a bad example for why we should keep `kdbEnsure`. It takes the contract in `elektraOpen`, so it really doesn't matter.

---

The notifications API only works because of the `list` plugin. Without the `list` plugin we need considerable changes. In particular, `zeromqrecv` and `dbusrecv` must be mounted _before_ `elektraIoSetBinding` is called. That means you need to do:

```c
KDB * kdb = kdbOpen (key);
kdbEnsure (kdb, contract, key); // contract must contain `zeromqrecv` or `dbusrecv`
elektraIoSetBinding (kdb, binding);
elektraNotificationOpen (kdb); // calls kdbEnsure again to mount `internalnotification`
```

It would be significantly simpler and much less error-prone, if `kdbOpen` took the contract. Then we could do:

```c
KDB * kdb = kdbOpen (key, contract); // contract contains notification config
elektraIoSetBinding (kdb, binding);
elektraNotificationOpen (kdb); // fails if kdb is incorrectly configured
```

`kdbOpen` would mount `internalnotification` and `zeromqrecv` or `dbusrecv`, depending on the notification config in the `contract`.

The alternative would be to move the ["deferred call"](https://github.com/ElektraInitiative/libelektra/blob/f3ab8d68be71a6bbd46b6d478dd8cc6683dbfecc/doc/decisions/deferred_plugin_calls.md) functionality from `list` to the core and thereby allowing `elektraIoSetBinding` to be called before `zeromqrecv` or `dbusrecv` are mounted. I don't think this would be a good idea, it is a lot more complex than merging `kdbOpen` and `kdbEnsure`.

---

The `gopts` "wrapper" also can't be seen as an argument for a separate `kdbEnsure`. Because what we want actually is **not** a wrapper. We want something like `elektraSetIoBinding` or `elektraNotificationOpen` that is called after `gopts` is mounted and before `kdbGet`. What we want is actually this:

```c
KDB * kdb;
/* set up kdb such that `gopts` is mounted */
elektraGOptsSetArgs (kdb, argc, argv);

kdbGet (kdb, ks, parent);
```

`elektraGOptsSetArgs` finds the `Plugin *` for `gopts` in `kdb` and stores `argc` and `argv` in there, so that they can be used in `kdbGet`.

---

>  there are also advantages to kdbEnsure, it is more flexible

More flexible also means more error prone. (at least in this case)

doc/decisions/arbitrary_metadata.md
kodebach: I don't have a concrete case, I'm just not entirely convinced we didn't miss an edge case.

Your clarification actually confused me even more...

> In other namespaces only store metadata that is handled by the `spec` plugin.

What? `spec` only handles metadata from the `spec:/` namespace. 

> for `array`, metadata might be stored in any namespace

The way I understand your proposal, `array` should **never** be stored anywhere but `spec:/`. And that's because in `spec:/` it has a different meaning.

In `spec:/` the metadata `array=` (empty value) means  "this is an array". If you give it a value e.g. `array=#4` it means "this is an array with default size X" (e.g. `#4` = size 5).

In any other namespace (at runtime) `array=` means "this is an empty array" and e.g. `array=#4` means "this is an array with max index `#4`".

The `spec:/` meaning of `array=` must obviously be stored. The other meaning should _never_ be stored and instead should be inferred from the stored data. Either by the storage plugin (e.g. JSON array) or by the `spec` plugin. At least that is what I got from the sentence: "[if] the storage plugin does not support arrays, then the metadata will be discarded".

I can live with this approach, but just in case you didn't notice, this means we loose some (possibly important) functionality compared to the current approach. Take this spec:

```ini
[myarray]
array=#0

[myarray/#/data]
default=42
```

An application can (now and with your proposal) read `/myarray/#0/data` and will get `42`. The difference is, currently you can do 
```
kdb set-meta user:/myarray array #1
```
and then the `spec` plugin will generate 
```
user:/myarray/#0/data=42
user:/myarray/#1/data=42
```
The second key is the important one, because of it `/myarray/#1/data` will result in `42`. With your proposal this doesn't work, because `user:/` won't store the array size. Again I'm not quite sure what your intention was, maybe in such a case a plugin with array support like JSON should store e.g. `{ "myarray": [ { "data": 42 }, { "data": 42 } ] }`, i.e. store the defaults explicitly. But what about a plugin without array support?

The thing is arrays are a special case, so there will be some solution here. But that is actually why I asked you about the general case:

> Also what happens for e.g. kdb set-meta user:/myarray mymeta 5?

You didn't answer the question, so I assume you don't really know either... The way I understand your proposal this wouldn't do anything, unless `mymeta` can somehow be generated by `spec`. But if `spec` didn't generate `mymeta=5` before the `kdb set-meta` and it does afterwards, then **something** has to be stored in `user:/`, which means metadata **is** stored in `user:/`, just not explicitly.

doc/decisions/boolean.md
```diff
@@ -6,46 +6,39 @@ Inconsistent use of bool in various parts of Elektra.
...
-- 1 is true, everything else is false (default is false)
+The spec/docu should mention that a `boolean` is used and may specify the default.
+The absence of a non-required key without a specified default should be interpreted as false.
 
-Example:
+Example for an implementation in C in an application:
 
 ```c
-if ( strcmp(keyString(k), "0")) {/*true*/} else {/*false*/}
 if (!strcmp(keyString(k), "1")) {/*true*/} else {/*false*/}
```
kodebach, line 31:
```suggestion
if (k != NULL && strcmp(keyString(k), "1") == 0) {/*true*/} else {/*false*/}
```

I know the other code works as well, but we should use the most explicit and least confusing code variant in the documentation.

doc/decisions/boolean.md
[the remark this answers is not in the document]
kodebach: > The spec/docu should mention that a `boolean` is used and may specify the default.
> The absence of a non-required key without a specified default should be interpreted as false.

I think this (plus your C code) is a rather convoluted way of saying:

Only keys defined as `boolean` by the spec or by documentation should be treated as boolean. If no default value is specified the key should be interpreted as false. Users should assume that `1` is the only value that is interpreted as true.

We don't need to bring the notion of required keys into the mix here, since those are handled elsewhere and are totally independent of the type system.

doc/decisions/semantics_name.md
```diff
@@ -19,15 +19,33 @@ The alternative would be to have semantics in key names, with following advantag
 Do not encode any semantics into the key names.
 All semantic must be in metadata.
 
+Nevertheless, there are guidelines (without any checks in `keySetBaseName`):
+
+- `#` is used to indicate that array numbers follow.
+- `®` is used to indicate that some information was encoded in the key name.
+  This is usually only needed internally in storage plugins.
+- the UTF-8 sequence `®elektra` (i.e. the 9-byte sequence `C2 AE 65 6C 65 6B 74 72 61`) is reserved,
+  key names should not start with that sequence.
```
kodebach, line 28:
```suggestion
  key name parts should not start with that sequence.
```

kodebach:
It should be clear from my comments, which decision still need work IMO. But just in case:

- `arbitrary_metadata` needs some clarifications regarding things like `kdb set-meta user:/key mymeta 5`.
- `boolean` is basically done, but could be worded a bit more clearly.
- `ensure` I disagree with the entire decision, except that it should only work for global plugins
- `escaped_name` I agree with the basis, but I don't think it can be decided as is. What can be decided is that the escaped name will no longer be stored and should used only when necessary. The exact API (what will be renamed and, how `keyNew` will work etc.) is still very much up in the air IMO.

doc/decisions/arbitrary_metadata.md
markus2330: Thank you, here is the update: 459e489ac

> I'm just not entirely convinced we didn't miss an edge case.

I added the assumption that such a problem would be fixed by the definition in the respective metadata.

> because user:/ won't store the array size

Only not explicitly, of course it infers the correct size and sets the metadata `array` correctly. I added the constraint.

> `kdb set-meta user:/myarray mymeta 5` You didn't answer the question

For me this question was answered with the very first sentence "Do not store arbitrary metadata in any namespace except spec.", so this call cannot do anything useful. I clarified now it should fail and not discard the metadata.

doc/decisions/spec_expressiveness.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Assumptions
+
+## Considered Alternatives
+
+## Decision
+
+- no defaults for `sw/_/key` specifications
+  (default will not work for `ksLookup(/sw/sthg/key)`)
+- plugins are not allowed to create keys
```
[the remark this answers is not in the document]
markus2330: I don't know how this statements should influence the text in the decision.

doc/decisions/boolean.md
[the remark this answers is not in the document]
markus2330: Thank you, I updated the text: e5394e5d5

doc/decisions/array.md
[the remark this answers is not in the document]
markus2330: :+1: 34016adfb3bd8cd9ee7fd0a7f7cbcbbf353b55fd

doc/decisions/ensure.md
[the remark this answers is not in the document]
markus2330: Ok, convinced, I turned around the whole decision: 3e79ed7a1747765ca6a427b5fa506878aa6da995

doc/decisions/boolean.md
```diff
@@ -2,62 +2,53 @@
...
 
 - most easy to implement
-- allows presence to be true
+- allows non-presence to be false
 - plugins allow us to convert to any other behavior
 
 ## Implications
 
-- change code with different behavior
+- Storage plugins are only allowed to emit `0` or `1`
```
kodebach, line 48:
```suggestion
- Storage plugins are only allowed to emit `0` or `1` as key values
```

("emit" could also mean into storage files)

doc/decisions/ensure.md
```diff
@@ -0,0 +1,68 @@
...
+
+Similar contract getters need to be introduced for other functionality, like gopts:
+
+```c
+KeySet * contract = ksNew (0, KS_END);
+elektraGOptsGetContract (contract, argc, argv, environ);
+KDB * kdb = kdbOpen (key, contract);
+```
+
+The high-level API can make this more pretty, though.
```
kodebach, line 62: Not 100% sure this is the best possible API, but yes something like this is what I thought of.

The proposal is a bit inconsistent. AFAICT `elektraIoSetBinding` is a separate function to avoid having a `Key` whose value is a `ElektraIoInterface *`. But then there is `elektraGOptsGetContract` which needs to store `char **`, which has the same problems.

So either, both cases use a separate setter function called after `kdbOpen` that directly manipulates the relevant `Plugin` struct, or they both just use a `*GetContract` function that creates binary keys.

Personally, I think a separate getter makes more sense. The whole point of having a `contract` KeySet is that the API/ABI doesn't have to change if additional config options are added. A `elektraNotificationGetContract` function would have to change, if e.g. notifications get some new config option.

doc/decisions/spec_expressiveness.md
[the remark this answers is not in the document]
kodebach:
```suggestion
- plugins are not allowed to create keys (may change in future; depends on plugin positions)
```

Just a small note like this should be enough. What is and isn't possible heavily depends on the how the different plugin positions will be defined and how they interact with each other. For 1.0 it is probably good enough to say "using a non-storage plugin that creates keys may result in unexpected behaviour".

kodebach:
I like were the decision on metadata is going, but I think we might need another more in depth discussion and should maybe move it to a separate PR.

What I got from your proposal are these key points ("store" means "write into storage file"):
- metadata that is used as specification (e.g. `check/ipaddr`, or `execute/set`) are stored only in `spec:/` files
- other namespaces do not store any metadata, but storage plugins can infer some metadata from files (e.g. `type`, `comment`, `array`)
- the `spec` plugin does **a lot** of heavy lifting

AFAICT it we are basically going in the direction of separating the two uses of metadata: specification, and inferred meta-information.

IMO the best solution would be, if plugins like `network` and `shell` would take their specifications (e.g. `check/ipaddr` and `execute/set`) directly from the `spec:/` keys, without the `spec` plugin having to copy everything to all namespaces. This would however be a massive change, so it is probably out of scope. It is also unclear, how the `spec:/` keys would even be passed to non-global plugins.

The next best thing would be, if the `spec` plugin copied the metadata into meta-keys with namespace `metaspec:/`. So a `Key` can have `meta:/` and `metaspec:/` metadata.  The `meta:/` keys are the one inferred from the storage file or from the specification, and the `metaspec:/` keys are the actual specification. This would still require updating many plugins, but it would be a quite simple change (just adding the `metaspec:/` prefix shouldn't be too hard). This also avoids the dual meaning of e.g. `array`; `metaspec:/array=#0` is "default size 1" and `meta:/array=#0` is "actual size is 1". It would also make the `spec` plugin a bit simpler. For example `spec` doesn't need to handle merging `type` from `spec:/` keys and other namespaces. [1] Instead the `type` plugin could handle that. For example:

```json
{ "mykey": 4 }
```

produces a key `mykey=4` with metadata `type=long_long`. If there is also a specification where `mykey` has `type=int`, this wouldn't be a problem. The `type` plugin would just do the normal validation between `metaspec:/type=int` and the value `4` and then do an additional check that a whether `meta:/type` can be assigned to `metaspec:/type` (e.g. `int` can be assigned to `long_long`).

Assuming that applications don't need direct access to the `metaspec:/` metadata, the `metaspec:/` metadata could also be discarded at the end of `kdbGet` and regenerated at the start of `kdbSet` to avoid any inconsistencies between the specification in the `spec:/` namespace and the `metaspec:/` metadata. 

---

[1] `spec` would not only copy to `metaspec:/` it also still needs to create some `meta:/` metadata (e.g. `type` and `array`) in some cases.

doc/decisions/ensure.md
markus2330: > Not 100% sure this is the best possible API,

Also not yet sure, here is an improvement: 581c6bf3018bea540343de30f342e948c781d772

doc/decisions/ensure.md
```diff
@@ -0,0 +1,80 @@
...
+
+```c
+KeySet * contract = ksNew (0, KS_END);
+elektraNotificationContract (contract, iobinding);
+```
+
+The same for gopts:
+
+```c
+elektraGOptsContract (contract, argc, argv, environ));
```
kodebach, line 55:
```suggestion
elektraGOptsContract (contract, argc, argv, environ);
```

doc/decisions/ensure.md
kodebach: Yes, the new API looks a lot better. However, I would prefer, if either `kdbOpen` took ownership of the `KeySet * contract` or at least copied what is needed, so that `ksDel (contract)` can be called right after `kdbOpen`. Otherwise there is too much potential for memory leaks IMO.

markus2330:
> ensure I disagree with the entire decision, except that it should only work for global plugins

Ok, as you saw I reworked it completely. 581c6bf3018bea540343de30f342e948c781d772

> escaped_name I agree with the basis, but I don't think it can be decided as is. What can be decided is that the escaped name will no longer be stored and should used only when necessary. The exact API (what will be renamed and, how keyNew will work etc.) is still very much up in the air IMO.

Yes, true, I renamed the sections so that it is clear that there is still a lot to be decided about the API. 788d273166c1d7515cc707f97c598cad8f2ccc32

doc/decisions/ensure.md
markus2330: Yes, of course it will copy, so ksDel afterwards is safe: a6429dc9a89ba25340f3e72b2883daff85b00333

Taking ownership can easily lead to misunderstandings, especially because KeySet is different to Key (which is ref-counted, so the ownership can be shared even if "taken" like in ksNew). In elektraPluginOpen we have such an API and I actually think we should change it there (if we decide to make this API public).

mpranj:
Thank you for updating the decisions from the meeting!

markus2330:
Thank you for all your great and valuable input at the meeting :sparkling_heart:

doc/decisions/ensure.md
```diff
@@ -0,0 +1,86 @@
...
+ksDel (contract);
+```
+
+The cleanup of the global plugins happens within:
+
+```c
+kdbClose (kdb, errorKey);
+```
+
+It is save to use the contract `KeySet` also for `kdbGet` and `kdbSet`
```
mpranj, line 79:
```suggestion
It is safe to use the contract `KeySet` also for `kdbGet` and `kdbSet`
```

mpranj (approved):
Looks great, I only found some typos.

mpranj:
Thank you so much!

PR #3551, opened 2020-11-05
SWIG/Lua: remove ipairs support, add our own iterator

### LINKED ISSUE #3480: Lua binding: ipairs skips first key

Compile and run `make run_all`. (Specifically `test_keyset.lua`)

## Expected Result

Compilation and tests should be sucessful.

## Actual Result

`test_keyset.lua` fails with a nondescript error. On closer inspection, it turns out that `ipairs` (Line 66) always skips the first Key in the KeySet.

## System Information

- Elektra Version: master
- Operating System: Manjaro Stable Version, up-to-date as of 2020-09-05
- Lua Version: 5.4.0 (specifically 5.4.0-2 from the Manjaro extra repo)

## Further Information

AFAIK Lua uses indexes starting from 1. Maybe this isn't adjusted to a zero-based index somewhere.

manuelm:
Closes #3480.

My approach to fixing #3480

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -100,7 +100,9 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
 Bindings allow you to utilize Elektra using [various programming languages](https://www.libelektra.org/bindings/readme). This section keeps
 you up to date with the multi-language support provided by Elektra.
 
-### <<Binding1>>
+### Lua
+
+- Remove ipairs support and add our own iterator to add support for Lua 5.4, since `__ipairs` was deprecated. _(Manuel Mausz)_
 
 ### <<Binding2>>
 
--- a/src/bindings/swig/lua/kdb.i
+++ b/src/bindings/swig/lua/kdb.i
@@ -361,6 +361,16 @@
 %rename("_%s") kdb::KeySet::lookup;
 %rename("_lookup") kdb::KeySet::at;
 
+%fragment("LuaSTLIterator_T");
+%{
+  inline static void key_iterator_iter(lua_State *L,
+    myswig::LuaSTLIterator_T<kdb::KeySet::iterator>::reference val)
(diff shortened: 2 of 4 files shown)


mpranj:
Thank you so much! 🚀 

A few questions:

1. Is this now presumably compatible with all Lua 5.x versions?
2. Is another way of doing this to add the `__index` metamethod and then have different implementations for different Lua versions? (Yes, this sounds ugly)
3. Is this done, do you want to add a few words to `doc/news/_preparation_next_release.md`? If you don't want to, ping me so that I add some release notes and make Jenkins happy.

manuelm:
1. yes, should be compatible

2. we already have an `__index` method. And we already support numeric access. However we start at 0, Lua starts at 1. So its an incompatible change anyway. I'm not really sure which approach is better. `pairs`/`ipairs` is for key/value-based tables. I don't think our KeySet is really key/value-based, although we support a few common fast-access patterns. So if you think changing `__index` to always do a `- 1` (for positives numbers), I'm fine with that as well. We could also add the magic `__pairs` and return pairs of `tuple(keyname,key)`.

3. yes, please go ahead

mpranj (approved):
Thank you so much for fixing this so quickly @manuelm!

This seems even simpler than `ipairs` to me, with the benefit of supporting more Lua versions. Since the Elektra 0.9 releases are incompatible anyway, I think there is not much downside.

> I don't think our KeySet is key/value-based (that's why it's called a set, right?)

I want to go a bit off-topic here. 😄 Why is it not key/value based?
I don't think we support all set-theoretical operations (e.g. intersection of two KeySets, which could actually be implemented), but I also don't see that as a problem. Since the Key is unique in a KeySet, we can also view it as key/value-based.

I'm just curious because afaik we even mention key-value in our docu.

markus2330:
Thank you both very much for this fix :1st_place_medal:

PR #3560, opened 2020-11-15
Ruby 2.7.2 test

mpranj:


markus2330:
Thank you for looking into this :sparkling_heart:

markus2330:
Is it maybe possible to upgrade swig? Probably it is an incompatibility from the used swig version to Ruby 2.7.2?

PR #3566, opened 2020-11-24
YAML CPP: Fix Use After Free Bug

sanssecours:


markus2330 (approved):
Thank you so much! I love the fix, especially it also makes the exception overhead go away.

mpranj (approved):
Thank you so much, I also think we can merge this!

Is it still WIP for a reason? I think the build failures are unrelated, and master is not green anyway.

sanssecours:
> Is it still WIP for a reason?

Yep, I am still waiting for either me or some other contributor to fix the remaining CI problems.

mpranj:
It seems that the last remaining error is the debian branch. The `hidden` plugin was removed from master but not from the debian packages.

I removed it now and triggered a rebuild. 🤞

markus2330:
Thank you @mpranj :sparkling_heart: 

These kind of problems will hopefully soon disappear due to @robaerd' cpack work.

scripts/docker/debian/sid/doc.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM debian:sid
```
mpranj, line 1: Would it work on Debian buster (stable) as well?

mpranj (approved):
Thank you for fixing everything that stood in the way of green builds, as well as the use-after-free bug!

scripts/docker/debian/sid/doc.Dockerfile
sanssecours: I tried, but unfortunately Debian buster ships with the same (horribly outdated) version of Doxygen (version `1.8.13`). The [bug in the generator for the PDF manual](https://github.com/doxygen/doxygen/issues/6456) should be fixed with Doxygen `1.8.15` (released about two years ago).

mpranj: Good to know. If we encounter further problems, we can move the doc task to Fedora. I think they use much more recent packages in general.

PR #3589, opened 2020-12-11
qtgui: update for keyname overhaul

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

### DIFF
--- a/src/tools/qt-gui/src/confignode.cpp
+++ b/src/tools/qt-gui/src/confignode.cpp
@@ -11,9 +11,9 @@
 
 using namespace kdb;
 
-ConfigNode::ConfigNode (QString name, QString path, const Key & key, TreeViewModel * parentModel)
+ConfigNode::ConfigNode (QString name, QString path, const Key & key, TreeViewModel * parentModel, bool root)
 : m_name (std::move (name)), m_path (std::move (path)), m_key (key), m_children (new TreeViewModel), m_metaData (nullptr),
-  m_parentModel (parentModel), m_isExpanded (false), m_isDirty (false)
+  m_parentModel (parentModel), m_isExpanded (false), m_isDirty (false), m_root (root)
 {
 	setValue ();
 
@@ -247,6 +247,11 @@ void ConfigNode::updateNode (Key key)
 void ConfigNode::setIsExpanded (bool value)
 {
 	m_isExpanded = value;
+}
+
+bool ConfigNode::isRoot () const
+{
+	return m_root;
 }
 
 void ConfigNode::populateMetaModel ()
--- a/src/tools/qt-gui/src/confignode.cpp
+++ b/src/tools/qt-gui/src/confignode.cpp
@@ -11,9 +11,9 @@
 
 using namespace kdb;
 
-ConfigNode::ConfigNode (QString name, QString path, const Key & key, TreeViewModel * parentModel, bool root)
+ConfigNode::ConfigNode (QString name, QString path, const Key & key, TreeViewModel * parentModel)
 : m_name (std::move (name)), m_path (std::move (path)), m_key (key), m_children (new TreeViewModel), m_metaData (nullptr),
-  m_parentModel (parentModel), m_isExpanded (false), m_isDirty (false), m_root (root)
+  m_parentModel (parentModel), m_isExpanded (false), m_isDirty (false)
 {
 	setValue ();
 
@@ -249,9 +249,9 @@ void ConfigNode::setIsExpanded (bool value)
 	m_isExpanded = value;
 }
 
-bool ConfigNode::isRoot () const
+bool ConfigNode::isNamespaceRoot () const
 {
-	return m_root;
+	return m_path.indexOf ('/') < 0;
 }
 
 void ConfigNode::populateMetaModel ()
--- a/src/tools/qt-gui/src/confignode.hpp
+++ b/src/tools/qt-gui/src/confignode.hpp
@@ -37,7 +37,7 @@ public:
 	 * @param key
 	 * @param parentModel
 	 */
-	explicit ConfigNode (QString name, QString path, const kdb::Key & key, TreeViewModel * parentModel, bool root = false);
+	explicit ConfigNode (QString name, QString path, const kdb::Key & key, TreeViewModel * parentModel);
 	/// Needed by Qt. This copy constructor is supposed to create a DEEP COPY.
 	ConfigNode (const ConfigNode & other);
 	/// Needed by Qt/QSharedPtr
@@ -241,9 +241,9 @@ public:
 	void updateNode (kdb::Key key);
 
 	/**
-	 * @return true if this is ConfigNode is a root node
+	 * @return true if this is ConfigNode is the root of a namespace
 	 */
-	bool isRoot () const;
+	bool isNamespaceRoot () const;
 
 private:
 	QString m_name;
@@ -257,7 +257,6 @@ private:
 
 	bool m_isExpanded;
 	bool m_isDirty;
-	bool m_root;
 
 	/**
 	 * @brief Populates the TreeViewModel which holds the metakeys of this ConfigNode.
--- a/src/tools/qt-gui/src/confignode.hpp
+++ b/src/tools/qt-gui/src/confignode.hpp
@@ -37,7 +37,7 @@ public:
 	 * @param key
 	 * @param parentModel
 	 */
-	explicit ConfigNode (QString name, QString path, const kdb::Key & key, TreeViewModel * parentModel);
+	explicit ConfigNode (QString name, QString path, const kdb::Key & key, TreeViewModel * parentModel, bool root = false);
 	/// Needed by Qt. This copy constructor is supposed to create a DEEP COPY.
 	ConfigNode (const ConfigNode & other);
 	/// Needed by Qt/QSharedPtr
@@ -240,6 +240,11 @@ public:
 	void setIsDirty (bool dirty);
 	void updateNode (kdb::Key key);
 
+	/**
+	 * @return true if this is ConfigNode is a root node
+	 */
+	bool isRoot () const;
+
 private:
 	QString m_name;
 	QString m_path;
@@ -252,6 +257,7 @@ private:
 
 	bool m_isExpanded;
 	bool m_isDirty;
+	bool m_root;
 
 	/**
 	 * @brief Populates the TreeViewModel which holds the metakeys of this ConfigNode.
--- a/src/tools/qt-gui/src/treeviewmodel.cpp
+++ b/src/tools/qt-gui/src/treeviewmodel.cpp
@@ -81,11 +81,12 @@ QVariant TreeViewModel::data (const QModelIndex & idx, int role) const
 
 	// TODO: document fallthrough if it was desired
 	case NameRole:
-		if (node->isRoot ())
+		if (node->isNamespaceRoot ())
 		{
 			kdb::Key k;
 			k.setNamespace (static_cast<ElektraNamespace> (node->getName ().at (0).unicode ()));
-			return QVariant::fromValue (QString::fromStdString (k.getName ()));
+			std::string name = k.getName ();
+			return QVariant::fromValue (QString::fromStdString (name.substr (0, name.length () - 1)));
 		}
 		else
 		{
@@ -128,6 +129,9 @@ QVariant TreeViewModel::data (const QModelIndex & idx, int role) const
 
 	case IsExpandedRole:
 		return QVariant::fromValue (node->isExpanded ());
+
+	case IsNamespaceRootRole:
+		return QVariant::fromValue (node->isNamespaceRoot ());
 
 	default:
 		emit showMessage (tr ("Error"), tr ("Unknown role: %1").arg (role), "TreeViewModel::data");
@@ -498,26 +502,23 @@ void TreeViewModel::populateModel ()
 	populateModel (config);
 }
 
-void TreeViewModel::addRootNode (ElektraNamespace ns)
-{
-	kdb::Key k;
-	k.setNamespace (ns);
-
-	auto name = k.getName ();
-	name = name.substr (0, name.length () - 1);
-
-	m_model << ConfigNodePtr (new ConfigNode (QString::fromStdString (std::string{ static_cast<char> (ns) }),
-						  QString::fromStdString (name), nullptr, this, true));
-}
-
 void TreeViewModel::populateModel (KeySet const & keySet)
 {
 	m_model.clear ();
 
-	addRootNode (ElektraNamespace::SPEC);
-	addRootNode (ElektraNamespace::DIR);
-	addRootNode (ElektraNamespace::USER);
-	addRootNode (ElektraNamespace::SYSTEM);
+	auto nsToShow = { ElektraNamespace::SPEC, ElektraNamespace::DIR, ElektraNamespace::USER, ElektraNamespace::SYSTEM };
+
+	for (auto & ns : nsToShow)
+	{
+		kdb::Key k;
+		k.setNamespace (ns);
+
+		auto name = k.getName ();
+		name = name.substr (0, name.length () - 1);
+
+		m_model << ConfigNodePtr (new ConfigNode (QString::fromStdString (std::string{ static_cast<char> (ns) }),
+							  QString::fromStdString (name), nullptr, this));
+	}
 
 	createNewNodes (keySet);
 }
@@ -865,6 +866,7 @@ QHash<int, QByteArray> TreeViewModel::roleNames () const
 	roles[IndexRole] = "index";
 	roles[IsNullRole] = "isNull";
 	roles[IsExpandedRole] = "isExpanded";
+	roles[IsNamespaceRootRole] = "isNamespaceRoot";
 
 	return roles;
 }
--- a/src/tools/qt-gui/src/treeviewmodel.cpp
+++ b/src/tools/qt-gui/src/treeviewmodel.cpp
@@ -81,7 +81,16 @@ QVariant TreeViewModel::data (const QModelIndex & idx, int role) const
 
 	// TODO: document fallthrough if it was desired
 	case NameRole:
-		return QVariant::fromValue (node->getName ());
+		if (node->isRoot ())
+		{
+			kdb::Key k;
+			k.setNamespace (static_cast<ElektraNamespace> (node->getName ().at (0).unicode ()));
+			return QVariant::fromValue (QString::fromStdString (k.getName ()));
+		}
+		else
+		{
+			return QVariant::fromValue (node->getName ());
+		}
 
 	case PathRole:
 		return QVariant::fromValue (node->getPath ());
@@ -489,43 +498,26 @@ void TreeViewModel::populateModel ()
 	populateModel (config);
 }
 
+void TreeViewModel::addRootNode (ElektraNamespace ns)
+{
+	kdb::Key k;
+	k.setNamespace (ns);
+
+	auto name = k.getName ();
+	name = name.substr (0, name.length () - 1);
+
+	m_model << ConfigNodePtr (new ConfigNode (QString::fromStdString (std::string{ static_cast<char> (ns) }),
+						  QString::fromStdString (name), nullptr, this, true));
+}
+
 void TreeViewModel::populateModel (KeySet const & keySet)
 {
 	m_model.clear ();
 
-	using namespace ckdb; // for namespaces
-	for (int i = KEY_NS_FIRST; i <= KEY_NS_LAST; ++i)
-	{
-		elektraNamespace ns = static_cast<elektraNamespace> (i);
-		ConfigNodePtr toAdd;
-		switch (ns)
-		{
-		case KEY_NS_SPEC:
-			toAdd = ConfigNodePtr (new ConfigNode ("spec", "spec", nullptr, this));
-			break;
-		case KEY_NS_PROC:
-			// TODO: add generic commandline parsing
-			break;
-		case KEY_NS_DIR:
-			toAdd = ConfigNodePtr (new ConfigNode ("dir", "dir", nullptr, this));
-			break;
-		case KEY_NS_USER:
-			toAdd = ConfigNodePtr (new ConfigNode ("user", "user", nullptr, this));
-			break;
-		case KEY_NS_SYSTEM:
-			toAdd = ConfigNodePtr (new ConfigNode ("system", "system", nullptr, this));
-			break;
-		case KEY_NS_NONE:
-			break;
-		case KEY_NS_META:
-			break;
-		case KEY_NS_CASCADING:
-			break;
-		case KEY_NS_DEFAULT:
-			break;
-		}
-		if (toAdd) m_model << toAdd;
-	}
+	addRootNode (ElektraNamespace::SPEC);
+	addRootNode (ElektraNamespace::DIR);
+	addRootNode (ElektraNamespace::USER);
+	addRootNode (ElektraNamespace::SYSTEM);
 
 	createNewNodes (keySet);
 }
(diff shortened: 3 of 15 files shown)


kodebach:
@markus2330 Please check that the QtGUI now works as expected. The solution is a bit of a hack, but I'm no Qt expert so a better solution has to be done by somebody else...

0003088:
I will build the project and check it out. Based on a short superficial view I'm not sure I get the purpose of the root node though. Is there some documentation what the 'new key name structure' actually is? Why is 'a better solution' necessary? What are the shortcomings of the current state?

kodebach:
> what the 'new key name structure' actually is?

The "new key name structure" is #3555. For the Qt GUI specifically only the colon after the namespace was relevant AFAIK.
IMHO we can remove the release notes line later, it doesn't actually serve any purpose other than pleasing the build server. In an ideal world, this fix would have been part of #3555.

> Why is 'a better solution' necessary?

I just wanted to say: I have no experience with Qt and don't really know what is and is not a good solution. With this PR we can see keys in the Qt GUI again. But if somebody with more Qt experience says my code is not ideal, that person should probably fix it rather than me.

> What are the shortcomings of the current state?

I called this PR a hack, because you have to explicitly declare that a `ConfigNode` is a root node. IMO a better solution would be, if this would be detected somehow. E.g. some parent is NULL or something like that.

src/tools/qt-gui/src/treeviewmodel.cpp
```diff
@@ -489,43 +498,26 @@ void TreeViewModel::populateModel ()
...
-			toAdd = ConfigNodePtr (new ConfigNode ("spec", "spec", nullptr, this));
-			break;
-		case KEY_NS_PROC:
-			// TODO: add generic commandline parsing
-			break;
-		case KEY_NS_DIR:
-			toAdd = ConfigNodePtr (new ConfigNode ("dir", "dir", nullptr, this));
-			break;
-		case KEY_NS_USER:
-			toAdd = ConfigNodePtr (new ConfigNode ("user", "user", nullptr, this));
```
markus2330, line 513: Simply adding `:` here did not help?

markus2330:
> Is there some documentation what the 'new key name structure' actually is?

Relevant for the qt-gui is, afaik, only that a colon `:` was added after the namespaces. 

> IMHO we can remove the release notes line later

Do not worry, this will be done anyway as needed. But it is good to mention which parts are touched and currently nearly nothing is written about the keyname overhaul. It is much easier to shorten down release notes than to look up in the diffs+logs what happened.

> I called this PR a hack, because you have to explicitly declare that a ConfigNode is a root node. IMO a better solution would be, if this would be detected somehow.

I also do not understand why the keyname overhaul made this necessary. See review comment above.

src/tools/qt-gui/src/treeviewmodel.cpp
kodebach: No, because we use the C++ key name iterator. That iterates over unescaped key names, so we need `std::string{ static_cast<char> (ns) }` to match correctly. In other words the `ConfigNode::getName` should return an unescaped key name part.

What I don't know is, if the `name` of the root `ConfigNode`s is ever checked outside of for loop in `createNewNodes`. If not there may have been easier solutions.

markus2330:
For me qt-gui does not start at all with these changes.

I get:

```
terminate called after throwing an instance of 'kdb::tools::PluginNoContract'
  what():  No contract found for that plugin!
Make sure you export kdbGet correctly!
```

stacktrace:

```
#0  0x00007ffff6135afd in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x000055555556395a in kdb::Key::setBaseName (baseName="user:/MyApp/mykey", this=0x555555797210) at /usr/include/c++/8/ext/new_allocator.h:86
#2  ConfigNode::setName (this=0x5555557971e0, name=...) at /home/markus/Projekte/Elektra/current/src/tools/qt-gui/src/confignode.cpp:103
#3  0x000055555556c123 in ConfigNode::populateMetaModel (this=0x5555557b8610) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:569
#4  0x000055555556c740 in ConfigNode::ConfigNode (this=0x5555557b8610, name=..., path=..., key=..., parentModel=0x55555573e5e0, root=<optimized out>) at /home/markus/Projekte/Elektra/current/src/tools/qt-gui/src/confignode.cpp:23
#5  0x00005555555959f0 in TreeViewModel::sink (this=0x7fffffffdd90, node=..., keys=..., key=...) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qsharedpointer_impl.h:312
#6  0x00005555555955c5 in TreeViewModel::sink (this=0x7fffffffdd90, node=..., keys=..., key=...) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qsharedpointer_impl.h:330
#7  0x0000555555595e56 in TreeViewModel::createNewNodes (this=0x7fffffffdd90, keySet=...) at /usr/include/c++/8/bits/atomic_base.h:295
#8  0x0000555555595fc4 in TreeViewModel::populateModel (this=0x7fffffffdd90, keySet=...) at /home/markus/Projekte/Elektra/current/src/bindings/cpp/include/keyset.hpp:653
#9  0x000055555559601a in TreeViewModel::populateModel (this=this@entry=0x7fffffffdd90) at /home/markus/Projekte/Elektra/current/src/tools/qt-gui/src/treeviewmodel.cpp:498
#10 0x0000555555569193 in main (argc=<optimized out>, argv=<optimized out>) at /home/markus/Projekte/Elektra/current/src/tools/qt-gui/src/main.cpp:69
```

`kdb ls /` works without problems.

src/tools/qt-gui/src/treeviewmodel.cpp
0003088: The root node is only there to be the parent of all other nodes, AFAIR it is never touched again.

kodebach:
That stack trace doesn't make sense for the error message... `kdb::Key::setBaseName` cannot throw a `kdb::tools::PluginNoContract`.

If the stack trace is correct, then `keySetBaseName` failed. This can only happen in 3 cases:
1. the Key is `NULL`
2. the Key is marked as `KEY_FLAG_RO_NAME`
3. the Key has a broken name. "Broken" because `key->key` has to be `NULL` or not contain a slash, both of which cannot happen under normal circumstances.

> `kdb ls /` works without problems.

AFAIK that doesn't use the `Plugin::parse ()` (the only location where `PluginNoContract` is thrown). Have you tried `kdb mount` to mount a new plugin?

markus2330:
Sorry, I see now that there are many exceptions (dozens), I didn't give you the stacktrace of the last exception. I'll try to find out the stacktrace of the last one some other day.

> AFAIK that doesn't use the Plugin::parse () (the only location where PluginNoContract is thrown). Have you tried kdb mount to mount a new plugin?

Mounting works without problems. And the startup of the qt-gui should do something similar to `kdb ls /` and not something like mounting.

markus2330:
So the problem I had was only because of a tidy build dir (see long description below. Maybe we should apply this patch, though). The only relevant comment for this PR: I do not like that we write `user:/` but all other keys do not have `/` appended. Either we add `/` everywhere, where are subkeys are, or nowhere.

---

With:

```
set pagination off
catch throw
commands
backtrace
continue
end
run
```

I got the backtrace of the last exception which is:

```
#0  0x00007ffff6135afd in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff7507e94 in kdb::tools::Plugin::parse (this=0x5555563e7a20) at /home/markus/Projekte/Elektra/current/src/bindings/cpp/include/key.hpp:842
#2  0x00007ffff7531511 in kdb::tools::Modules::load (this=0x7fffffffcea0, spec=...) at /usr/include/c++/8/bits/unique_ptr.h:342
#3  0x00007ffff75315d1 in kdb::tools::Modules::load (this=0x7fffffffcea0, pluginName=..., config=...) at /usr/include/c++/8/bits/char_traits.h:287
#4  0x00007ffff7531738 in kdb::tools::Modules::load (this=this@entry=0x7fffffffcea0, pluginName="null") at /home/markus/Projekte/Elektra/current/src/libs/tools/src/modules.cpp:39
#5  0x000055555557e317 in GUIBackend::availablePlugins (this=0x7fffffffde30, includeStorage=true, includeResolver=true) at /home/markus/Projekte/Elektra/current/src/tools/qt-gui/src/guibackend.cpp:262
#6  0x000055555559b92e in GUIBackend::qt_static_metacall (_o=<optimized out>, _c=<optimized out>, _id=<optimized out>, _a=0x7fffffffd110) at /home/markus/Projekte/Elektra/build/src/tools/qt-gui/qt-gui_autogen/UVLADIE3JM/moc_guibackend.cpp:145
#7  0x000055555559d723 in GUIBackend::qt_metacall (this=0x7fffffffde30, _c=QMetaObject::InvokeMetaMethod, _id=7, _a=0x7fffffffd110) at /home/markus/Projekte/Elektra/build/src/tools/qt-gui/qt-gui_autogen/UVLADIE3JM/moc_guibackend.cpp:211
#8  0x00007ffff7908b8d in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#9  0x00007ffff7870421 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#10 0x00007ffff7871855 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#11 0x00007ffff78725a2 in QV4::QObjectMethod::callInternal(QV4::Value const*, QV4::Value const*, int) const () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#12 0x00007ffff788ca69 in QV4::Runtime::method_callProperty(QV4::ExecutionEngine*, QV4::Value*, int, QV4::Value*, int) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#13 0x00007ffff78801f2 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#14 0x00007ffff79364ff in QQmlJavaScriptExpression::evaluate(QV4::CallData*, bool*) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#15 0x00007ffff793b0c7 in QQmlBinding::evaluate(bool*) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#16 0x00007ffff793fbc2 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#17 0x00007ffff793c353 in QQmlBinding::update(QFlags<QQmlPropertyData::WriteFlag>) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#18 0x00007ffff7949054 in QQmlObjectCreator::finalize(QQmlInstantiationInterrupt&) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#19 0x00007ffff78c10b1 in QQmlComponentPrivate::complete(QQmlEnginePrivate*, QQmlComponentPrivate::ConstructionState*) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#20 0x00007ffff78c11d7 in QQmlComponentPrivate::completeCreate() () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#21 0x00007ffff78c0fb0 in QQmlComponent::create(QQmlContext*) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#22 0x00007ffff79412bb in QQmlApplicationEnginePrivate::finishLoad(QQmlComponent*) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#23 0x00007ffff794157e in QQmlApplicationEnginePrivate::startLoad(QUrl const&, QByteArray const&, bool) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#24 0x00007ffff79415bd in QQmlApplicationEngine::load(QUrl const&) () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#25 0x00005555555691c1 in main (argc=<optimized out>, argv=<optimized out>) at /home/markus/Projekte/Elektra/current/src/tools/qt-gui/src/main.cpp:81
terminate called after throwing an instance of 'kdb::tools::PluginNoContract'
  what():  No contract found for that plugin!
Make sure you export kdbGet correctly!
```

I applied:
```
diff --git a/src/tools/qt-gui/src/guibackend.cpp b/src/tools/qt-gui/src/guibackend.cpp
index c62f166b3..a979d370d 100644
--- a/src/tools/qt-gui/src/guibackend.cpp
+++ b/src/tools/qt-gui/src/guibackend.cpp
@@ -261,8 +261,14 @@ QStringList GUIBackend::availablePlugins (bool includeStorage, bool includeResol
                {
                        ptr = modules.load (s);
                }
+               catch (PluginNoContract const & ex)
+               {
+                       cout << "no contract: " << s << endl;
+                       continue;
+               }
                catch (NoPlugin const & ex)
                {
+                       cerr << "no plugin: " << s << endl;
                        continue;
                }
```

and got the output:

```
no plugin: hidden
no plugin: ini
no contract: null
no contract: tcl
no plugin: hidden
no plugin: ini
no contract: null
no contract: tcl
```

So it looks like qt-gui somehow internally cached an out-dated list of plugins and for some non-existent plugins it did not get the expected NoPlugin but the PluginNoContract exception.

So probably we should add the second catch statement, it does not look like it would harm but helps in cases like I had.

kodebach:
> I do not like that we write `user:/` but all other keys do not have `/` appended

I did intentionally use a different appearance for the root keys. But I also don't like the slash... I just didn't know what the best solution would be. Using `user` as root, is just wrong as it suggests `user/abc` should work. The easiest solution would be to show `user:`, but I hate that. It suggests `user:` is a normal part, when in fact it is a namespace.

For example the key `user:/user:/abc` is valid and would be shown in the GUI as `user:/` -> `user:` -> `abc`. This shows that the first and second `user:` are different. That is what I wanted to achieve. I don't really know enough about Qt to use more subtle stylings to show this difference, e.g. color namespace, make them bold, etc.

> So probably we should add the second catch statement, it does not look like it would harm but helps in cases like I had.

Agreed, I'll add the catch

0003088:
I'm not sure I get it 100%, but it seems there are some kinds of metanodes introduced which indicate a namespace and are not actually nodes? In this case wouldn't it be best to hide these nodes in the gui and only show the namespace when editing ( == opening) the key? Or is the ns information crucial to the user? 

Another possibility could be to use curved brackets like in {user} -> user: -> abc?

But maybe colorcoding would be best

kodebach:
> Or is the ns information crucial to the user?

Yes it is. Basically, if you open the GUI you will see these root nodes that are always there. One for every namespace (that the GUI supports). Previously, they were implemented like normal nodes, since the namespace was just the first part of the key name. e.g. `user/abc/mykey` is in namespace `user`. With #3555 we made these namespaces more explicit. The key name now looks like this `user:/abc/mykey` (where again `user` is the namespace). In addition to that, in an unescaped key name the namespace is represented by a single byte (and parts are separated by `\0` bytes).

> Another possibility could be to use curved brackets like in {user} -> user: -> abc?

The problem is that `{user}` could again be a normal part of the key name. But `{user}` would certainly be better than `user:` and `user`.

0003088:
Then I would say the best way to go is to colorcode the namespace nodes and to make it impossible to edit them.

src/tools/qt-gui/src/confignode.hpp
```diff
@@ -252,6 +257,7 @@ class ConfigNode : public QObject
 
 	bool m_isExpanded;
 	bool m_isDirty;
+	bool m_root;
```
0003088, line 260: I wouldn't call these 'm_root' but 'm_namespace' instead (because there is only one single root node).

0003088: The name change includes the methods as well of course

markus2330:
Yes, I also like colorcoding. It is possible to edit them, as they also represent the root key (/) of each namespace.

@0003088 can you help out with the implementation?

0003088:
Do you mean 'help with' or 'do' the implementation? I can help right now with providing some steps with what I think should be done:

1. add 'isNamespace' to the TreeViewModelRoles enum in treeviewmodel.hpp
2. add it also to TreeViewModel::data() and TreeViewModel::roleNames ()
3. in TreeView.qml

```
property Component delegate: Row {
		spacing: defaultSpacing

		Label {
			id: label

			text: rowLoaderModel === null || rowLoaderModel === undefined ? "" : rowLoaderModel.name
			color: rowLoaderModel === null || rowLoaderModel === undefined ? "transparent" : (rowLoaderModel.isNull ? guiSettings.nodeWithoutKeyColor : guiSettings.nodeWithKeyColor)
          ====>         font.weight: rowLoaderModel.isNamespace ? Font.Bold : Font.Normal // add this for example 
			onColorChanged: indicator.updateIndicator()
		}
```

For documentation on font weight look here https://doc.qt.io/qt-5/qml-qtquick-text.html#font.weight-prop

I you have problems don't hesitate to ask.

src/tools/qt-gui/src/confignode.cpp
```diff
@@ -249,6 +249,11 @@ void ConfigNode::setIsExpanded (bool value)
 	m_isExpanded = value;
 }
 
+bool ConfigNode::isRoot () const
+{
+	return m_root;
```
0003088, line 254: Isn't it somehow possible to get the property if it is a namespace key from the key itself?

src/tools/qt-gui/src/confignode.cpp
```diff
@@ -11,9 +11,9 @@
 
 using namespace kdb;
 
-ConfigNode::ConfigNode (QString name, QString path, const Key & key, TreeViewModel * parentModel)
+ConfigNode::ConfigNode (QString name, QString path, const Key & key, TreeViewModel * parentModel, bool root)
 : m_name (std::move (name)), m_path (std::move (path)), m_key (key), m_children (new TreeViewModel), m_metaData (nullptr),
-  m_parentModel (parentModel), m_isExpanded (false), m_isDirty (false)
+  m_parentModel (parentModel), m_isExpanded (false), m_isDirty (false), m_root (root)
 {
```
0003088, line 17: If it stays this way you would have to add the 'm_root(root)' part to the copy constructor too?

0003088: m_isRoot (other.m_isRoot)

src/tools/qt-gui/src/treeviewmodel.cpp
```diff
@@ -489,43 +498,26 @@ void TreeViewModel::populateModel ()
...
-			break;
-		case KEY_NS_DEFAULT:
-			break;
-		}
-		if (toAdd) m_model << toAdd;
-	}
+	addRootNode (ElektraNamespace::SPEC);
+	addRootNode (ElektraNamespace::DIR);
+	addRootNode (ElektraNamespace::USER);
+	addRootNode (ElektraNamespace::SYSTEM);
```
0003088, line 520: Why not iterate here over 'static const ElektraNamespace ELEKTRA_NAMESPACES[]'?

kodebach: Because we do not show all namespaces. I find it more readable to list all the namespaces that are shown instead of excluding those that aren't.

src/tools/qt-gui/src/confignode.cpp
kodebach: It would be pretty easy, if every `ConfigNode` stored a `ckdb::Key *` or `kdb::Key`. But only those nodes that actually represent keys of the KDB (not the intermediate ones) do that.

It is technically possible to check the `path` (= escaped name), but this is a lot easier. We could check the `path` just once in the constructor and in `setPath()` (assuming `path` isn't modified anywhere else).

kodebach:
> It is possible to edit them, as they also represent the root key (/) of each namespace.

Only editing the value should be allowed. Changing the name of root keys doesn't make sense.

> some steps with what I think should be done

Thank you. I will implement these steps (and some of your other suggestions).

I know the solution with `m_root` is less than ideal. I'll see if I can find a better (but still small) solution. However, I think the Qt GUI is due for a more general update. AFAICT it was written before the introduction of unescaped names and doesn't really support escaped slashes. In some places it uses the iterator from the C++ binding, which does use unescaped names. But in other placed, key name parts are just split and joined with `/`.

src/tools/qt-gui/src/confignode.cpp
0003088: But for our purpose we don't need every `ConfigNode` to have a key just the 'namespace' nodes? Why not add a `kdb::Key` to these?

src/tools/qt-gui/src/treeviewmodel.cpp
0003088: Can't the namespaces that are not shown be simply removed from the array and the enum then?

0003088:
As far as I understood somebody is working on migrating the gui to the qml treeview? These issues need to be addressed during the modification I suppose.

src/tools/qt-gui/src/treeviewmodel.cpp
kodebach: The general `ELEKTRA_NAMESPACES` array is part of the C++ binding. It has to contain **all** namespaces, not just those that are used in some arbitrary part (the Qt GUI) of Elektra. We could create a separate local array, but I don't see why that would be better.

markus2330:
> As far as I understood somebody is working on migrating the gui to the qml treeview?

Yes, but this is paused at the moment due to illness.

kodebach:
I think this new solution should be better:

- It doesn't use the explicit `m_root` anymore. Instead we check, if `m_path` contains a slash. This works because `m_path` for a root key doesn't have a slash e.g. `user:/` has `m_path = "user:"`
- `isRoot` is now `isNamespaceRoot`
- Namespace root keys are shown in bold and without the slash, e.g. **system:**
- Editing namespace root keys is allowed, but their name is locked. It should be possible to change the value and edit metadata.
- The `addRootNode` helper function was inlined into a loop over a local `std::initializer_list`.

markus2330:
It looks now amazing, I like the style very much.

There are some qml errors though:

```
qrc:/qml/EditKeyWindow.qml:59: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
qrc:/qml/TreeView.qml:44: TypeError: Cannot read property 'isNamespaceRoot' of null
...
```

And also thanks for adding the iterator tests! :sparkling_heart:

kodebach:
> There are some qml errors though:

fixed

markus2330:
The output was the only thing I noticed, so I am happy now.

@0003088 are you also happy with the PR?

0003088:
Looks absolutely fine to me, cannot say anything against it ...

markus2330:
Feel free to say anything in favor of it :wink:

0003088:
Well what can I say? I like this solution way more than the one with the root boolean. Also the naming scheme is better. It's perfectly fine. 

Btw I'm looking forward to the new version, whenever it will be done or by whom.

kodebach:

mpranj:
Thank you!

Issue #2856, opened 2019-08-04
gopts, quickdump, specload: tests fail


markus2330:
## Steps to Reproduce the Problem

Compile and install Elektra, remove (or rename) the source/build directory. Then run `kdb run_all`

## Expected Result

All test cases should run successfully.

## Actual Result

```
Running testmod_gopts

GOPTS     TESTS
==================

test empty
GOPTS     TESTS
==================

test empty
/home/jenkins/workspace/libelektra_master-Q2SIBK3KE2NBEMJ4WVGJXAXCSCB77DUBUULVLZDKHQEV3WNDXBMA/libelektra/src/plugins/gopts/testmod_gopts.c:78: error in run_test: child process test failed
test singleopt
/home/jenkins/workspace/libelektra_master-Q2SIBK3KE2NBEMJ4WVGJXAXCSCB77DUBUULVLZDKHQEV3WNDXBMA/libelektra/tests/cframework/tests.c:523: error in clean_temp_home: Could not delete TMPHOME via nftw
GOPTS     TESTS
==================
```

```
Running testmod_quickdump
QUICKDUMP     TESTS
==================

test varint
test basics
/home/jenkins/workspace/libelektra_master-Q2SIBK3KE2NBEMJ4WVGJXAXCSCB77DUBUULVLZDKHQEV3WNDXBMA/libelektra/src/plugins/quickdump/testmod_quickdump.c:111: error in test_basics: call to kdbSet was not successful

Program received signal SIGSEGV, Segmentation fault.
_IO_getc (fp=0x0) at getc.c:37
37      getc.c: Datei oder Verzeichnis nicht gefunden.
(gdb) bt
#0  _IO_getc (fp=0x0) at getc.c:37
#1  0x0000555555556bb7 in compare_binary_files (filename1=<optimized out>, filename2=<optimized out>) at ./src/plugins/quickdump/testmod_quickdump.c:31
#2  0x0000555555556f9a in test_basics () at ./src/plugins/quickdump/testmod_quickdump.c:113
#3  0x0000555555556807 in main (argc=1, argv=0x7fffffffe278) at ./src/plugins/quickdump/testmod_quickdump.c:332
```

```
Running testmod_specload

SPECLOAD     TESTS
==================

test basics
[...]

Issue #3522, opened 2020-10-23
Ubuntu Focal (20.04) Python Bindings


markus2330:
It looks like Ubuntu Focal (20.04) has incompatible SWIG/Python versions according to:

https://github.com/ElektraInitiative/libelektra/blob/cbb0909f6ed189d1a04872818ae48556f3d44d8d/src/bindings/swig/python/CMakeLists.txt#L28-L30

which gets exactly triggered by:

https://packages.ubuntu.com/focal/swig 4.0.1-5build1
https://packages.ubuntu.com/focal/python3 3.8.2-0ubuntu2

@manuelm do you remember which incompatibility is addressed there?

The first question is, if our check is correct. Maybe 4.0.1-5build1 already contains a patch that fixes this incompatibility.

If they really are incompatible, then we could try to upgrade swig, which hopefully still will lead to functioning Ubuntu Focal (20.04) packages, as swig should not show up as run-time dependency.

@robaerd Can you already build Ubuntu Focal (20.04) packages for the upcoming release? (At the moment we also do not even have a Docker image for 20.04, so it is probably quite a bit of work.) There were already several requests to have it, e.g. #3412

@mpranj It would be great if this issue can be fixed for 0.9.3. (If the swig upgrade is necessary and works, nothing is to be changed in the source of Elektra.)

@dev2718 Thank you for reporting this problem!

mpranj:
The original issue regarding this incompatibility is #3379. It was resolved by adding the version check.

If ubuntu backported the fix, which I do not think, there is no guarantee that other distris did the same.

manuelm:
> If ubuntu backported the fix, which I do not think, [...]

They didn't backport the commit.

See https://bugs.launchpad.net/ubuntu/+source/swig/+bug/1876764

manuelm:
While further looking: `tp_print` is marked deprecated. So the compilation should still succeed and the generated bindings should work fine. Do you guys treat `__attribute__((__deprecated__))` as a failure? If so this should be fixed and #3379 reverted.

markus2330:
Thank you so much for looking into this!

If `-Werror` is passed we treat most warnings as errors in master, in #3447 f90ec817965b86402743583c0f7af9cd31338707 @kodebach changed it, as again plenty of new warnings came up. We should probably get this commit to master for less troubles in general.

> While further looking: tp_print is marked deprecated. So the compilation should still succeed and the generated bindings should work fine.

[...]

PR #3601, opened 2020-12-20
CPack RPM packaging

robaerd:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


robaerd:
Limitations that couldn't be solved:
* CPack RPM raises a few warnings that CPACK_RPM_PACKAGE_RELOCATABLE will be deactivated because CPACK_SET_DESTDIR is set. CPACK_RPM_PACKAGE_RELOCATABLE is explicitly unset to deactivate this, but apparently CPack still handles it, as it would be set. I tried setting it to "OFF", "FALSE", unseting, caching it etc, but none of this worked. I think this could be a bug with CPack RPM itself. But since this only yields warnings and doesn't affect any functinality this shouldn't be a problem.

* The default path of the source of debuginfo packages is `/usr/src/debug/`. To be able to place the source in this directory, the Source dir must be on a path that is longer than the path where the debuginfo source is installed. E.g.: debuginfo src of package libelektra4 is installed at `/usr/src/debug/Elektra-0.9.3-Fedora33-amd64/libelektra4/src_0/` so the source dir must have at least 6 parent dirs e.g. `/folder/folder/folder/folder/folder/folder/libelektra`. But this shouldn't be a problem since we can simply copy the libelektra dir to another folder in the CI.

markus2330:
> I think this could be a bug with CPack RPM itself. But since this only yields warnings and doesn't affect any functinality this shouldn't be a problem.

Can you do a bug report about these warnings we receive (also the ones in Debian). Maybe the CMake devs know some easy way to fix it or at least they'll know about the problem and can fix it in the future.

> But this shouldn't be a problem since we can simply copy the libelektra dir to another folder in the CI.

I do not really understand the problem. Can someone who wants to build the RPM packages on his machine trigger this problem? If yes, we should write about it in doc/INSTALL.md. In general we should update doc/INSTALL.md as from now on, creating packages via cpack will be the favorable option, and `make install` should be only used if you are on a system where cpack support is not (yet) available.

robaerd:
> Can you do a bug report about these warnings we receive (also the ones in Debian).

Yes I will do this.

> Can someone who wants to build the RPM packages on his machine trigger this problem?

Yes, unless someone sets CPACK_RPM_BUILD_SOURCE_DIRS_PREFIX to a smaller path, they will need to place the libelektra source to a longer path (with 6 parent directories) and build the package from there. I will update doc/INSTALL.md and document these things.

robaerd:
Since `make package` will become the standard way of installing Elektra we will need to explicitly disable the generation of individual packages depending on what plugins/bindings are included. If someone decides to build a package with less plugins/bindings than expected by each package, the packaging will fail because of the debuginfo stripping of the CPack generators. 

If will add this right to this PR if the approach is OK for you @markus2330.

markus2330:
Can you give more details about the approach? How do you want to disable packages?

What is important for CMake usability is to have as little CACHE variables as possible, as with these variables every maintainer gets confronted. So I would not like a solution where we would need to introduce a CACHE variable for every package. Two ways that sound fine to me would be:

1. In the case of `make package` we simply require more dependencies to be installed so that we can produce useful modular packages. If someone wants to build Elektra without these dependencies, she needs to use `make install`. This is more or less a usability&documentation task: we somehow need to avoid that people get confused with the output of the "debuginfo stripping" problem and tell them they need to install the deps instead.
2. We can somehow fix the "debuginfo stripping" problem and produce empty packages or automatically exclude packages if the whole content of a package was not built.

robaerd:
I just testet the case for RPM and DEB generators. The build does not fail for RPM packages because the RPM generator detects that there is no source (except for /usr/share/) present and disables the debuginfo generation.
The DEB generator does really weird things, it doesn't throw any errors, but instead shuffles around the dependencies of the dbgsym packges. For example: if qt-gui is disabled, the qt-gui-dbgsym package is not built, but the elektra-bin-dbgsym packages gets the control file of the qt-gui-dbgsym package (with the dependencies to qt-gui and the package name qt-gui-dbgsym). 

My idea was to check the ADDED_PLUGINS, ADDED_BINDINGS and ADDED_TOOLS variables to determine which of the plugins, bindings and tools are not included. Especially for the components/packages where there is a one-to-one relationship to plugins etc.
My initial approach to avoid wrapping everything in conditionals was to modify CPACK_COMPONENTS_ALL after the `include (CPack)` command, but it seems that this must happen before. So now, we would have to check for each of the packages (at least for those with a one-to-one relationship to a plugin/binding/tool) if the plugin etc. is present and if it is not present disable the debuginfo package generation.

This way no CACHE variables would need to be set.

markus2330:
> does not fail for RPM packages [...] The DEB generator does really weird things

Maybe this is a bug in CMake and what we do is actually correct and should already work?

> determine which of the plugins, bindings and tools are not included; [...] This way no CACHE variables would need to be set.

Yes, this proposal is perfect from the user's point of view!

> ADDED_PLUGINS, ADDED_BINDINGS and ADDED_TOOLS

Maybe it is better to track which components are added (e.g. within `add_plugin`) and then simply exclude the component which were never added? But if it is less code the other way round, feel free to ignore this suggestion.

It would be good to print which components were added or excluded.

robaerd:
> Maybe this is a bug in CMake and what we do is actually correct and should already work?

I'm not 100% sure about this, since they state in their [docu](https://cmake.org/cmake/help/v3.16/cpack_gen/deb.html#packaging-of-debug-information) that binaries must contain debug symbols before packaging. But I will include this in the bug report, because the RPM generator actually handles this case.

> Maybe it is better to track which components are added (e.g. within add_plugin) and then simply exclude the component which were never added? 

This way we would not have the option to yield a message that a part of the expected plugins of a package is missing. For example libelektra4-zeromq depends on the plugins zeromqrecv and zeromqsend.  But if this isn't a problem for you, then I can also rewrite it like you suggested.

markus2330:
No need to rewrite if there are no problems, let us see how the current implementation works.

doc/INSTALL.md
```diff
@@ -111,7 +109,49 @@ Please refer to the section OS Independent below.
...
+
+Then use:
+
+```sh
+make package
+```
+
+which will create a package for distributions where a Generator is implemented.
+
+You can find the generated packages in the `packages` directory of the build directory.
```
markus2330, line 129: Maybe we can rename this folder to "package" as, at least zsh, always suggests to correct package to packages.

markus2330 (commented):
Looks good! Can you also update doc/INSTALL.md how to install the Ubuntu packages produced by our build server. At the moment only buster is explained (with a wrong heading, where it says "For Stretch.") and the outdated Jessie (does it even work? I think we can remove it now).

robaerd:
> Can you also update doc/INSTALL.md how to install the Ubuntu packages produced by our build server. 

Yes of course.


Since the update of the release-pipeline depends on the changes of this PR (and also fixes a few things), should I add all these changes (including the update of the release-pipeline) to this PR or create a separate one after this one is merged?

markus2330:
Thank you! We should also start offering both released packages and packages build from master. Do we have all of them (Debian+Ubuntu) already in repositories?

It is up to you how you prefer to split up PRs. Smaller is usually better. The docu should only reflect what is, not what it might be with upcoming PRs.

robaerd:
None of the released packages are currently offered in our repositories. I will make myself familiar with the `reprepro` tool and add this functionality to the release pipeline.

markus2330:
Thank you! Do we already have Ubuntu Focal debs build from master? The docu does not mention it at all even though #3522 is resolved.

robaerd:
Oh I think I misunderstood #3522 and thought that Ubuntu Focal packages should only be built for releases. I will add this to the upcoming PR with the updated release pipeline.

markus2330:
You are right, in #3522 nothing was mentioned about master builds. I think, however, it would be the easiest for the user if we build the same set of packages for both master and releases.

Specifically, for Ubuntu Focal we need packages once #3608 is merged. It would be possible to do a release afterwards. So if it is a lot of effort to do master packages, we can consider to provide **only** release packages and make some pre-releases so that we have the ability to create packages when needed. Or you find a way to share code between master and release packaging (or even creating packages for branches), this would be best from my point of view. What do you think?

robaerd:
I don't think it is a lot of effort to also provide them in the main pipeline, since I also have to create the repositories for the released packages.

Should we then also include the stage where the built packages are installed and tested in a docker container? This is currently done in the updated release-pipeline, but IMO it wouldn't make much sense to include it in the main pipeline until #2856 is not fixed. 

Since we also create Fedora packages and the `reprepro` tool doesn't support RPM, should these packages also be created in the main pipeline, or should we wait until we also have a repository for RPM packages?

>  Or you find a way to share code between master and release packaging

I planned to refactor the shared code of the main and release Jenkinsfile into a shared groovy library, since I already introduced a lot of duplicate code. Depending on when the release will happen, I'm not sure if I manage to get this done for the current one. 

> ... (or even creating packages for branches)

So for Pull Requests?

markus2330:
> Should we then also include the stage where the built packages are installed and tested in a docker container?

As we want the possibility of problems during release to be kept at a minimum, such tests ideally are in the main pipeline.

> This is currently done in the updated release-pipeline, but IMO it wouldn't make much sense to include it in the main pipeline until #2856 is not fixed.

Then let us fix #2856 by excluding these tests from the installation.

> Since we also create Fedora packages and the reprepro tool doesn't support RPM, should these packages also be created in the main pipeline, or should we wait until we also have a repository for RPM packages?

Is there maybe some tool that supports several distributions and package formats? reprepro is some legacy tool I used to use many years ago. There is no special reason to keep it.

> I planned to refactor the shared code of the main and release Jenkinsfile into a shared groovy library, since I already introduced a lot of duplicate code.

That would be amazing.

> Depending on when the release will happen, I'm not sure if I manage to get this done for the current one.

In any case it is good that you always create PRs whenever you have something that is working.

>  So for Pull Requests?

To have this would be the best. But consider it as strictly nice-to-have. Afaik I am the only one sometimes building packages from branches. With cpack it is not a big deal to build packages for yourself anyway.

robaerd:
> Is there maybe some tool that supports several distributions and package formats?

I only could find [Pulp](https://pulpproject.org) as a tool that supports both, but it seems to be really hard to setup and to integrate it into our current setup. [aptly](https://www.aptly.info) is another well used tool for debian package management which appears to be much easier to use than `reprepro` and they have RPM repositories on their roadmap. The go-to tool for RPM repositories is `createrepo` and it seems very easy to use. Maybe switching to `aptly` would be a good idea but for now I think using `reprepro` and `createrepo` would be best.

robaerd:
Can I set the `ready to merge` label, or is there still something open? All the packaging related things were done in #3623.

mpranj:
When you finish your work you can always set the `ready to merge` label. This signals that the PR is ready for review and can be merged if everything is fine. If the reviewer requests changes the label is removed and you re-add the label when you've implemented the changes.

The label just avoids premature reviews and also gets PRs merged faster (visibility).

robaerd:
Oh then I misunderstood it, thanks!

mpranj (approved):
Thank you so much! I'm looking forward to testing the Fedora packages.

doc/INSTALL.md
```diff
@@ -111,7 +116,55 @@ Please refer to the section OS Independent below.
 
 First follow the steps in [COMPILE](COMPILE.md).
 
-After you completed building Elektra on your own, there are multiple options how to install it. For example, with make or cPack tools.
+After you completed building Elektra on your own, there are multiple options how to install it. For example, with make or CPack tools.
+We recommend that you generate your own packages with CPack so ensure compatibility with future releases.
```
markus2330, line 120:
```suggestion
We recommend to use the packages from our build server or that you generate your own packages with CPack.
```

doc/INSTALL.md
```diff
@@ -111,7 +116,55 @@ Please refer to the section OS Independent below.
...
+Simply `cd` into the build directory and run following command:
+
+```sh
+LD_LIBRARY_PATH=$(pwd)/lib:${LD_LIBRARY_PATH} make package
+```
+
+To install the packages run this in the `package` directory:
+
+```sh
+dpkg -i *
```
markus2330, line 150: With `apt` you can install directly with also installing dependencies.

doc/INSTALL.md
```diff
@@ -134,20 +187,6 @@ or in the build directory (will not honor `DESTDIR`!):
 xargs rm < install_manifest.txt
 ```
 
-### CPack
-
-First follow the steps in [COMPILE](COMPILE.md).
```
markus2330, line 139: Maybe we should keep this info somewhere?

scripts/cmake/ElektraCache.cmake
```diff
@@ -354,6 +354,20 @@ if (CMAKE_SKIP_BUILD_RPATH AND BUILD_TESTING)
 	message (WARNING "You specified to remove RPATH, but did not disable tests. Please read doc/COMPILE.md#RPATH")
 endif ()
 
+#
+# Packaging options
+#
+
+set (
+	CPACK_DEBIAN_PACKAGE_RELEASE
```
markus2330, line 362: Why not only one variable CPACK_PACKAGE_RELEASE?

markus2330 (commented):
Great work! (Partly reviewed)

scripts/cmake/ElektraCache.cmake
robaerd: CPACK_DEBIAN_PACKAGE_RELEASE and CPACK_RPM_PACKAGE_RELEASE are both predefined variables of CPack, but I agree, caching only CPACK_PACKAGE_RELEASE and setting CPACK_DEBIAN_PACKAGE_RELEASE and CPACK_RPM_PACKAGE_RELEASE to CPACK_PACKAGE_RELEASE would be better.

doc/INSTALL.md
robaerd: It is already present in the parent heading `## OS Independent` (not visible in the git diff). Is it really necessary to write this multiple times?

scripts/cmake/ElektraCache.cmake
markus2330: If they are already predefined we do not need to set them as cache variable anyway?

doc/INSTALL.md
markus2330: Yes, you are right, I did not see it.

scripts/packaging/package.sh
```diff
@@ -4,41 +4,70 @@ ELEKTRA_PLUGINS='ALL;mozprefs;multifile;-gitresolver;jni;-ruby;-haskell;yamlcpp;
...
-  -DCMAKE_MODULE_LINKER_FLAGS=$LDFLAGS \
-  -DCMAKE_SHARED_LINKER_FLAGS=$LDFLAGS"
+OS_NAME=$(grep "^NAME=" /etc/os-release | awk -F= {' print $2'} | sed 's/\"//g')
+
+echo "DIST: $OS_NAME"
+
+if [ $OS_NAME = "Fedora" ]; then
+
+	CMAKE_ARGS="-DTARGET_PLUGIN_FOLDER='elektra4' \
+    -DBUILD_STATIC=OFF \
```
markus2330, line 14: Some args, like -DBUILD_STATIC=OFF, seem to be the same in both branches.

scripts/cmake/Modules/PackagingFedora.cmake
```diff
@@ -0,0 +1,199 @@
...
+set (CPACK_RPM_ELEKTRA-MISC_PACKAGE_DESCRIPTION "${PACKAGE_DESCRIPTION}")
+
+set (CPACK_RPM_ELEKTRA-DOC_PACKAGE_NAME "${CPACK_COMPONENT_ELEKTRA-DOC_DISPLAY_NAME}")
+set (CPACK_RPM_ELEKTRA-DOC_PACKAGE_SUMMARY "${CPACK_COMPONENT_ELEKTRA-DOC_DESCRIPTION}")
+set (CPACK_RPM_ELEKTRA-DOC_PACKAGE_DESCRIPTION "${PACKAGE_DESCRIPTION}")
+set (CPACK_RPM_ELEKTRA-DOC_PACKAGE_ARCHITECTURE "noarch")
+
+# install copyright file
+configure_file ("${CMAKE_SOURCE_DIR}/LICENSE.md" "${CMAKE_BINARY_DIR}/doc/LICENSE" COPYONLY)
+configure_file ("${CMAKE_SOURCE_DIR}/doc/THIRD-PARTY-LICENSES" "${CMAKE_BINARY_DIR}/doc/THIRD-PARTY-LICENSES" COPYONLY)
```
markus2330, line 193: Do we really need distribution-specific installation of files? Maybe put it together with execute_process, here it is a bit surprising.

scripts/cmake/Modules/PackagingFedora.cmake
```diff
@@ -0,0 +1,199 @@
+set (CPACK_RPM_PACKAGE_VERSION "${PROJECT_VERSION}")
```
markus2330, line 1: Give a short info what this file is about (that it sets CPACK_RPM_ vars and also installs some rpm specific files). Maybe also split this into two files (with separate headers)

scripts/cmake/ElektraCache.cmake
robaerd: I agree, but I would at least define CPACK_PACKAGE_RELEASE as cache variable with a default value. This is especially useful for our release pipeline where the package revision of each package can be changed with the parameterized build.

scripts/cmake/ElektraPackaging.cmake
```diff
@@ -1,19 +1,88 @@
...
+#
+# dependency_name  name of the dependency the component depends on
+# component_name   name of the component
+# PLUGIN           whether the dependency is a plugin
+# BINDING          whether the dependency is a binding
+# TOOL             whether the dependency is a tool
+# ADDITIONAL_DEPENDENCIES additional dependencies if the component
+# 				          has multiple dependencies
+# ~~~
+macro (check_component_dependencies dependency_name component_name)
```
markus2330, line 16: Maybe you can put this to some other file, the files already get lengthy.

scripts/cmake/ElektraPackaging.cmake
```diff
@@ -1,19 +1,88 @@
-set (PACKAGE_NAME "elektra")
+# ~~~
+# check_component_dependencies
+#
+# check and excludes a component from packaging if
```
markus2330, line 4: nice docu :+1:

markus2330 (approved):
Great job! Can imho be merged and the small things fixed later.

scripts/cmake/ElektraCache.cmake
markus2330: Cache variables are afaik only needed if you want it to be visible in the cmake GUIs. I do not think this is needed in this situation.

With something like:
```
if (CPACK_RPM_PACKAGE_RELEASE) set (CPACK_RPM_PACKAGE_RELEASE ...)
```
you should be able to set a default if the variable is not set from outside.

scripts/cmake/Modules/PackagingFedora.cmake
robaerd: Unfortunately yes. `/usr/share/licenses/` is the default directory for licenses on Fedora based distributions and is not present on Debian based distributions. I will move it up like you suggested.

scripts/packaging/package.sh
robaerd: This is already fixed in #3623.

mpranj:
Thank you. 🚀

PR #3631, opened 2021-01-24
ksReference related changes

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/src/libs/elektra/kdb.c
+++ b/src/libs/elektra/kdb.c
@@ -1536,13 +1536,13 @@ int kdbSet (KDB * handle, KeySet * ks, Key * parentKey)
 		return -1;
 	}
 
-	if (test_bit (parentKey->flags, KEY_FLAG_RO_META))
+	if (keyIsLocked (parentKey, KEY_LOCK_META) == KEY_LOCK_META)
 	{
 		ELEKTRA_LOG ("parentKey KEY_FLAG_RO_META");
 		return -1;
 	}
 
-	if (test_bit (parentKey->flags, KEY_FLAG_RO_NAME))
+	if (keyIsLocked (parentKey, KEY_LOCK_NAME) == KEY_LOCK_NAME)
 	{
 		clearError (parentKey);
 		ELEKTRA_SET_INTERFACE_ERROR (parentKey, "parentKey with read-only name passed");
@@ -1550,7 +1550,7 @@ int kdbSet (KDB * handle, KeySet * ks, Key * parentKey)
 		return -1;
 	}
 
-	if (test_bit (parentKey->flags, KEY_FLAG_RO_VALUE))
+	if (keyIsLocked (parentKey, KEY_LOCK_VALUE) == KEY_LOCK_VALUE)
 	{
 		clearError (parentKey);
 		ELEKTRA_SET_INTERFACE_ERROR (parentKey, "parentKey with read-only value passed");
--- a/src/libs/elektra/key.c
+++ b/src/libs/elektra/key.c
@@ -371,8 +371,7 @@ int keyCopy (Key * dest, const Key * source)
 {
 	if (!dest) return -1;
 
-	if (test_bit (dest->flags, KEY_FLAG_RO_NAME) || test_bit (dest->flags, KEY_FLAG_RO_VALUE) ||
-	    test_bit (dest->flags, KEY_FLAG_RO_META))
+	if (keyIsLocked (dest, KEY_LOCK_NAME | KEY_LOCK_VALUE | KEY_LOCK_META) > 0)
 	{
 		return -1;
 	}
@@ -730,5 +729,5 @@ int keyIsLocked (const Key * key, elektraLockFlags what)
 	if (!key) return -1;
 	what &= (KEY_LOCK_NAME | KEY_LOCK_VALUE | KEY_LOCK_META);
 	what >>= 16; // to KEY_FLAG_RO_xyz
-	return (test_bit (key->flags, what) << 16);
+	return (test_bit (key->flags, what) << 16) | (key->ksReference > 0 ? KEY_LOCK_NAME : 0);
 }
--- a/src/libs/elektra/key.c
+++ b/src/libs/elektra/key.c
@@ -540,21 +540,17 @@ int f (Key *k)
  * @endcode
  *
  * @retval returns 0 on success
- * @retval -1 on null pointer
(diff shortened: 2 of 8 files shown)


kodebach:
There is a very big problem with checking `ksReference` in e.g. `keySetName`:

https://github.com/ElektraInitiative/libelektra/blob/5a35769f63740225e441f88ed508f675c2d0b1b4/src/bindings/cpp/include/key.hpp#L553-L556

All `Key` constructors of the C++ binding immediately increase `ksReference`. Therefore any key created via the C++ binding or any binding based on the C++ binding, cannot be modified. Which in turn means lots and lots of code breaks.

There is also an issue with checking `ksReference` in `keyClear`. Although, AFAICT the issue only happens in a single test case (`test_keyClear` in `test_storage`)

https://github.com/ElektraInitiative/libelektra/blob/d9a8478478b0009971a070d2990af0b310abfd7e/tests/ctest/test_storage.c#L565-L570

Here `keyClear` is deliberately used to modify a key that is part of a keyset. I assume this test case is just broken and should be removed?

markus2330:
Yes, I am also afraid that the locking is unavoidable. Having separated counters for the bindings and for data structures sounds more complicated than locking. (As explanation: several (C++) objects can refer to the same C object, thus the ref counting is used in bindings. I am not sure if any other binding uses this technique, though.)

> Here keyClear is deliberately used to modify a key that is part of a keyset. I assume this test case is just broken and should be removed?

Yes, we should fix that keyClear will not allow to clear locked keys (e.g. those which are in a KeySet). Or we remove keyClear altogether.

kodebach:
> Yes, I am also afraid that the locking is unavoidable. 

That is not what I wanted to say; I do not agree. The big problem with the locking approach is that there is no way of knowing, when the key is removed from all keysets and can be unlocked again. Since `key->ksReference` may be incremented by non-keyset sources, not even `keyClear` would work.

The only way to change the name of key that was part of a keyset **at any point** is to `keyDup` it and use a new key. To me this just seems like a bad API.

> Having separated counters for the bindings and for data structures sounds more complicated than locking.

I think this would actually be simpler in the end. We use one general reference counter `key->references` that is incremented by _every_ reference (i.e. by bindings and by keysets). This counter is checked in `keyClear` and `keyDel` (i.e. when you want to reset the key), but nowhere else. The existing counter `key->ksReference` is now only incremented by `ksAppendKey`. Whenever this counter is > 0, the key name is implicitly locked.

The other solution would be to change the C++ binding. IMO there is a problem with the way it currently works anyway. Basically, a `Key *` is similar to a `std::shared_ptr` in C++. It is a reference counted object with shared ownership. When you initially add a key into a keyset, it takes full ownership and when you add the key into further keysets, each of them gets partial ownership. So far so good. The problem is, that the C++ class `kdb::Key` behaves like a keyset and also unconditionally takes (partial) ownership.

```cpp
ckdb::Key * cKey = ckdb::keyNew ("/", KEY_END);
{
kdb::Key cppKey(cKey);
}
foo (cKey);
```

That means in the above snippet `cppKey` takes ownership of `cKey` and therefore `keyDel (cKey)` is called when `cppKey` is destructed. So `foo (cKey)` will access already `free`d memory. The only way to prevent this is to call `keyIncRef` and `keyDecRef` manually. I'm not sure how to solve this, but it seems the Rust binding should have some solution for these ownership bindings. Maybe we can take some idea from there?

AFAICT the Rust binding doesn't automatically increment or decrement `ksReference`, but it does call `keyDel` when the Rust struct is dropped. That would be the Rust binding has an all-or-nothing approach to ownership. It takes full ownership if possible and otherwise doesn't take any ownership.

Maybe a solution is to use a `std::shared_ptr<Key *>` in `kdb::Key` and adjust the copy constructors accordingly. Additionally, the C++ wouldn't increment `key->ksReference` in the constructor anymore. Then multiple C++ objects could share a `Key *` and it wouldn't be deleted until all of them are destructed. As long as the key is also in one or more keysets the C++ objects would never delete the `Key *`.

The problem above still happens (it also happens in Rust AFAICT) and another problem appears to:

```cpp
void bar (ckdb::Key * k) {
    kdb::Key key(k);
}

void foo() {
    kdb::Key cppKey("/", KEY_END);
    bar (cppKey.getKey());
    std::cout << cppKey.getName() << std::endl;
}
```

The problem here is that `key` in `bar()` doesn't know anything about `cppKey` in `foo()`. `key` creates a new `std::shared_ptr` and at the end of `bar()` it will call `keyDel()`, because the `std::shared_ptr` has no other references. However, the `Key *` actually came from `cppKey`, which after `bar()` now contains an invalid pointer and the `std::cout` line will fail.

This could be solved by a second reference counter. But a simpler solution would be a flag similar to the `KEY_FLAG_MMAP_*` ones (maybe we could even rename and reuse `KEY_FLAG_MMAP_STRUCT`). The flag would indicate that the memory is managed by some external system (in this case the C++ runtime) and would prevent `keyDel()` from `free`ing anymore.

If `kdb::Key` calls `keyNew()`, it sets the flag and remembers that it did so, e.g via a `bool owner` field. When ever the `std::shared_ptr` constructed by this object is copied, the `owner` field is also copied. When `kdb::Key` is called with an existing `Key *`, it does not set the flag. That way there is one `std::shared_ptr` that owns the memory and when its reference count hits zero, it will correctly call `keyDel()`.
Because there may be legitimate cases where `kdb::Key` should to take ownership of an existing `Key *` there should be a constructor parameter to do that.

**What about other bindings?**
As I said, Rust already works kind of similar to this. Except for the fact that it doesn't set the flag and always tries to take full ownership of a `Key *` when possible.

The GLib binding does some weird stuff with swapping keys around and lots of `keyDel`. AFAIK the `GObject` stuff has its own reference counting, so something similar to the `std::shared_ptr` stuff should be possible. Currently the binding does increment/decrement `ks->Reference` just like the C++ one.

Java: Currently, works like C++. No idea what the best solution would be here.

AFAIK the remaining bindings all use Swig and therefore are based on top of C++. So any solution from C++ would automatically be applied there as well.

kodebach:
Since the original idea behind this PR didn't work, I'll close it.

For future reference: I now thin the best approach would be separate reference counters for reference from KeySet to a Key and from elsewhere to a Key. If there are references from KeySets to a Key that Key should be treated, as if its name was locked. The other reference counter is used for determining when `keyDel` can free the key.

PR #3590, opened 2020-12-12
simplify infos/status

markus2330:
fixes #666
fixes #3504

## Basics

These points need to be fulfilled for every PR:

- [X] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [X] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [X] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [X] The PR is rebased with current master.

If you have any troubles fulfilling these criteria, please write
about the trouble as comment in the PR. We will help you.
But we cannot accept PRs that do not fulfill the basics.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

Reviewers will usually check the following:

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

If you are already Elektra developer:

- Add the "work in progress" label if you do not want the PR to be reviewed yet.
- Add the "ready to merge" label **if the basics are fulfilled** and you also
  say that everything is ready to be merged.

PR #3640, opened 2021-01-30
Release prep 0.9.4

mpranj:


markus2330:
Thank you for this! :sparkling_heart:

src/plugins/cache/README.md
```diff
@@ -5,7 +5,7 @@
 - infos/provides =
 - infos/recommends =
 - infos/placements = pregetcache postgetcache
-- infos/status = maintained unittest shelltest specific global
+- infos/status = maintained unittest shelltest specific nodep global
```
mpranj, line 8: @markus2330 do we really want `./scripts/dev/update-infos-status --auto`?

Seems like it updates quite a few things. Here for instance it added nodep because no `find_package` or `LINK_LIBRARIES` was found. Afaik nodep means only a C99 compiler suffices, which is not the case. Cache needs a posix compliant system.

At other places it removed memleak status etc., it seems quite suspicious to me.

markus2330: I think in the long term we want `--auto` but I see it as task in #3590 to fix the wrong heuristics. For the current release I think we should not use it. Adding "nodep" for "cache" is the only change that sounds okay to me without further investigation. But I also agree with not having "nodep", as it needs POSIX, so it is borderline.

doc/news/2021-02-01_0.9.4.md
```diff
@@ -0,0 +1,375 @@
...
+## Get the Release!
+
+You can download the release from [here](https://www.libelektra.org/ftp/elektra/releases/elektra-0.9.<<VERSION>>.tar.gz)
+or [GitHub](https://github.com/ElektraInitiative/ftp/blob/master/releases/elektra-0.9.<<VERSION>>.tar.gz?raw=true)
+
+The [hashsums are:](https://github.com/ElektraInitiative/ftp/blob/master/releases/elektra-0.9.<<VERSION>>.tar.gz.hashsum?raw=true)
+
+<<`scripts/generate-hashsums elektra-0.9.<<VERSION>>.tar.gz`>>
+
+The release tarball is also available signed by Mihael Pranjić using GnuPG from
```
robaerd, line 352: The release tarball actually already gets signed inside the release pipeline and the signed version then also gets published, or do you actually want to sign it manually with your GPG key?

markus2330: It would be great to use that automatically signed one. The release notes were not updated about which GPG key it is, I only fixed that it is not me :wink:

mpranj: No I misunderstood that. Let's use the bot! Do you have the key fingerprint at hand to update it?

Btw:
One of the steps before triggering the release pipeline is to move the news file to the final place. The link checker will inevitably fail as the tarball etc is not available. This again will block the release pipeline from building the tarball (I am guessing).

Should we simply postpone moving the news file until after the pipeline is done?

robaerd: > Do you have the key fingerprint at hand to update it?

Unfortunatly no, to access it we will probably need to print it inside a Jenkins job, since accessing credentials from Jenkins is apparently not possible. But I have a gpg signed file from the release pipeline from which we can get the key that was used to sign the file, if this is enough?

```sh
$ gpg --verify /Users/robertsowula/Downloads/0.9.3\ 21/elektra-0.9.3.tar.gz.gpg
gpg: Signatur vom Sa 16 Jan 16:39:50 2021 CET
gpg:                mittels RSA-Schlüssel 12CC44541E1B8AD9B66AFAD55262E7353324914A
gpg: Signatur kann nicht geprüft werden: No public key
```

> Should we simply postpone moving the news file until after the pipeline is done?

Thank you for pointing this out, I didn't think of this. Yes I think we have no way around this, we also did it that way in the last release. Is it actually possible to exclude a file from link-checker with a comment (or similar)? Maybe this would be more elegant since we would then only need to remove the comment from this file after the pipeline is done.

mpranj: > key that was used to sign the file, if this is enough?

Yep, thank you!

> Is it actually possible to exclude a file from link-checker

The simplest way is to add a link to `tests/linkchecker.whitelist` in general. The `doc/news/_preparation_next_release.md` is already a special case where links containing `<<VERSION>>` are excluded from the check. It seems like the simplest to just keep this name until after the release, which avoids all those problems.

PR #3642, opened 2021-02-02
Update Alpine Linux images to 3.13.1

mpranj:


markus2330:
@robaerd can we also automate the uploading of the Alpine Linux images?

robaerd:
Yes, should be possible. I will include this (including a few other fixes) in a follow-up PR in the next days.

Issue #3599, opened 2020-12-20
Scope of implementation work done within FUSE Thesis


dev2718:
## Added by Markus

- [x] `cd ~/mount/47/proc: && cd ../54` yields "Input/output error" instead of "No such file or directory"
- [ ] no example given in Docker container where proc: is actually used and cascading case is interesting
- [x] maybe use `sleep infinity` instead of `sleep 9999999`
- [X] `rm dir:/dirkey` does not work (still present in FUSE afterwards, after several invocations it works though)
_Reply_: Observe that there is masking going on in the example. Before the first delete, the contents of `dirkey` are "Key from /root/dirkeys/a", afterwards its "Key from /root/dirkeys", after yet another rm `dirkey` is gone.

- [X] `--dir_file_special_name ®elektra/value --xattr_kdb_file ®elektra/file` broken (leads to input/output errors when accessing). I still do not think that it is a good idea to offer such options. I am fine with `@elektra.value` if `@elektra/value` does not work (and let us forget about ® if it leads to problems in practice): let us decide on something and update doc/KEYNAMES.md accordingly.
_Reply_: In `KEYNAMES.md` we already have "Any key name containing the key name part `®elektra`:" (is reserved). I removed the command line option; `®elektra.value` and `®elektra.file` are now used.

- [x] Bug: `cd mount/1; ls; ls` hangs endless
- [x] Implement `proc:` using #3651 (pass env+argc of process to )
- [x] Write semantics for cascading writes or disable cascading writes
- [x] ```cd /home/testuser; kdb set dir:/info "test user, I am here"; cd ~/mount/47``` ls fails, dir: is broken, ~/mount/47/cascading does not contain `...I am here`.

_Reply_: The reason for this is a permission problem: In the example, `/home/testuser/.dir/default.ecf'` is created by the root user with restrictive permissions, so testuser cannot see nor change the value of `dir:/info`. Therefore the fix is to deal with the permission exception, but in this precise scenario testuser will not have access to the key.

- [x] `kdb set user:/info/@elektra.value` creates two files called `@elektra.value`. You could use `®elektra` which is reserved (see KEYNAMES.md).
- [x] get file name of underlying file?
- [ ] put the whole README.md into shell recorder format to automatically check the behavior? (I am not sure if docker in docker works on our build server, though)
- [x] Access related errorcodes not yet implemented (e.g. normal user tries to write to system:/) (defaulting to read only fs - error)
[...]

PR #3657, opened 2021-02-15
Implement fuse-prototype including context-mocking and README.md

dev2718:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
+If a key (`A`) exists, and has some child key (`B`), `A` is treated as a directory containing `B` and a special file `@elektra.value`. This file is used to access the value of `A`.
+
+Some side effects arise:
+E.g. recursively deleting directories with `rm -r` containing `@elektra.value` files raises (task uncritical) errors when attempting to delete a parent directory which was already deleted with the removal of an `@elektra.value` file directly below.
+
+### "Leaf"-Keys that act as as directories
+
+Keys at the bottom of the hierarchy are normally treaded as files. However in certain cases, like explicitly creating directories using `mkdir`, this would result in files beeing displayed, making it impossible to expand the hierarchy using standard tools.
+Therefore, when a directory is created, the special meta-key `fuse-directory` is set to override this behaviour.
```
markus2330, line 138: use / as separator within (meta)keys.

```suggestion
Therefore, when a directory is created, the special meta-key `fuse/directory` is set to override this behaviour.
```

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+Another is that copies created with `cp -r` will not have these vacuous intermediaries but instead empty keys.
+
+### Keys that act both as file and directory
+
+In Elektra, there are only keys and no distinction between directories and files are made. To enable a bidirectional mapping into a classical file system, the following mechanism is employed:
+
+If a key (`A`) exists, and has some child key (`B`), `A` is treated as a directory containing `B` and a special file `@elektra.value`. This file is used to access the value of `A`.
+
+Some side effects arise:
+E.g. recursively deleting directories with `rm -r` containing `@elektra.value` files raises (task uncritical) errors when attempting to delete a parent directory which was already deleted with the removal of an `@elektra.value` file directly below.
```
markus2330, line 133:
```suggestion
E.g. recursively deleting directories with `rm -r` containing `®elektra.value` files raises (task uncritical) errors when attempting to delete a parent directory which was already deleted with the removal of an `®elektra.value` file directly below.
```

see #3599

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
+### "Leaf"-Keys that act as as directories
+
+Keys at the bottom of the hierarchy are normally treaded as files. However in certain cases, like explicitly creating directories using `mkdir`, this would result in files beeing displayed, making it impossible to expand the hierarchy using standard tools.
+Therefore, when a directory is created, the special meta-key `fuse-directory` is set to override this behaviour.
+
+### Filesystem attributes
+
+The filesystem attributes are sourced from the file obtained by `kdb file`.
```
markus2330, line 142: Is it also possible to receive the file name itself? see #3599

What is the meaning of "number of links or directories", it seems to differ for the real file and the FUSE file?

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+For some keys, no such file is returned, or the returned file does not exist in actuality. In these cases, fake attributes are used to preserve usability.
+
+To increase compatibility, currently the timestamps from these files are not used, as this prompts software like `vim` to produce warnings/errors.
```
markus2330, line 145: Which time stamps? The modification date seems to be present.

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
+## Meta-Keys
+
+Meta-keys are mapped bidirectionally to extended file attributes (xattrs).
+To enable interoperability with other tools, the `meta:/` prefix is optional when reading/writing accessing meta keys. The listing of attributes however will show them prefix-less.
+
+Furthermore, no binary values for these attributes are allowed. (As libelektra does not accept those).
+
+See `xattr(1)` on how to read/write/list these attributes.
```
markus2330, line 154: The man page is not present in Docker and on my system (7) opens by default which is full of descriptions of file-system-specific behavior. A small example how to get+set will not harm, e.g.:

```
xattr -w abc def person/name
kdb meta-get system:/person/name abc
```

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
+## Binary-Keys
+
+If a value is to be written that cannot be decoded using the systems default encoding, it is treated as binary, and the appropriate `meta:/binary` meta-key is set.
+
+## Known issues
+
+- Moving an ``@elektra.value``-file corrupts the filesytsem/key database (See this [issue](https://github.com/ElektraInitiative/libelektra/issues/3648)). As editors like `vim` may move a file during editing, such editors cannot reliebly be used on these pseudo-files.
```
markus2330, line 162:
```suggestion
- Moving an ``@elektra.value``-file corrupts the filesystem/key database (See this [issue](https://issues.libelektra.org/3648)). As editors like `vim` may move a file during editing, such editors cannot reliably be used on these pseudo-files.
```

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+- Only functional on UNIX-like systems
```
markus2330, line 163: Isn't the limitation simply that it will only work on systems that support FUSE?

```suggestion
- Only functional on Unix-like systems
```

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+- chmod/chown is currently not implemented (and how it should be is not obvious). A call does not signal a "not supported error" to enable compatibility with common tools like `cp`.
```
markus2330, line 164:
```suggestion
- chmod/chown is not implemented and does not signal a "not supported error" to enable compatibility with common tools like `cp`.
```

scripts/fuse/docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# run this container with any of the run scripts (in order to mount the source code folder, ensure privileged mode to be able to mount fuse, setup logging to nohup.out/stdout)
```
markus2330, line 1: Any reason why this is not in scripts/docker/?

scripts/fuse/docker/create_keys.sh
```diff
@@ -0,0 +1,48 @@
...
+sudo -u root kdb set user:/info "This is information originating from user:/info (root)"
+
+# ... and with a fallback system: key
+kdb set system:/person/name "This is a system:/ fallback"
+
+#test dirs that are files at the same time
+kdb set user:/dir_and_file_at_once "Contents of a non-leaf node"
+kdb set user:/dir_and_file_at_once/leaf "leaf node"
+
+#mount json file (stopped since libelektra v5, TODO: investigate)
```
markus2330, line 23: The ini plugin was removed. You can try with the yajl (json) plugin.

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
```diff
@@ -0,0 +1,218 @@
...
+#project imports
+from elektra_util import *
+import mock_context
+
+startup_time = time.time()
+
+#TODO: give appropriate error codes (instead of readonly fs) when non-authenticated user tries to do illegal things
+
+def getattr(path, fh=None):
+
```
markus2330, line 17: Can you write a (short) docu about what the functions are doing?

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
```diff
@@ -0,0 +1,218 @@
...
+        mode = stat.S_IFREG
+        #resolve the liml (/.../@elektra.value) key to the real key (the parent)
+        path = str(Path(path).parent)
+    elif not is_dir and not is_file:
+        mode = errno.ENOENT
+    elif is_dir and is_file:
+        mode = stat.S_IFDIR
+    elif is_dir:
+        mode = stat.S_IFDIR
+    elif is_file and has_meta(path, "meta:/fuse-directory"):
```
markus2330, line 32:
```suggestion
    elif is_file and has_meta(path, "meta:/fuse/directory"):
```

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
```diff
@@ -0,0 +1,218 @@
...
+            mode = errno.ENOENT
+
+
+    try:
+        kdb_file_stat = _stat_kdb_file(path)
+        st_mode_only_permission_bits = kdb_file_stat.st_mode & 0o777
+    except FileNotFoundError:
+        # some keys, e.g. "system:/elektra" resolve to a file like "/etc/kdb/elektra.ecf", which does not exist
+        # for the filesystem to remain useful, in this case dummy file attributes are used
+        st_mode_only_permission_bits = 0o770
```
markus2330, line 50:
```suggestion
        st_mode_only_permission_bits = 0o000
```

Maybe this makes clear that the file actually does not exist.

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
```diff
@@ -0,0 +1,218 @@
...
+            st_atime = startup_time
+        )
+    
+    
+
+    #common attributes for files and directories
+    key_stat = dict(
+        st_mode = (mode | st_mode_only_permission_bits),
+
+        #TODO: using the real timestamps results in vim complaining on write: "The file has been changed since reading it!!!"
```
markus2330, line 63: The problem is probably only atime which also gets incremented by FUSE access.

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
```diff
@@ -0,0 +1,218 @@
...
+        st_ctime = startup_time,
+        st_mtime = startup_time,
+        st_atime = startup_time
+    )
+
+    if mode == stat.S_IFDIR:
+        key_stat["st_nlink"] = 2
+        return key_stat
+    elif mode == stat.S_IFREG:
+        key_stat["st_nlink"] = 1 #TODO: maybe consider key.getReferenceCounter?
```
markus2330, line 77: Is probably useless as the refcounter is only some implementation-specific value (of the FUSE implementation).

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
```diff
@@ -0,0 +1,218 @@
...
+        return key_stat
+    else:
+        raise OSError(mode)
+
+#throws OSError when:
+# -) `kdb file` does not return a path
+# -) the returned path does not actually exist
+def _stat_kdb_file(os_path):
+    elektra_path = os_path_to_elektra_path(os_path)
+    resolved_file_path = subprocess.check_output(["kdb", "file", elektra_path]).decode().strip()
```
markus2330, line 88: This could be done much more efficiently to directly check the value of the parent key after a `KDB.get` from within python (This is also exactly what `kdb file` does.).

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
```diff
@@ -0,0 +1,218 @@
...
+    if returncode != 0:
+        raise OSError(errno.EROFS) #TODO: differentiate between different errors
+
+# does not raise OSError(errno.EOPNOTSUPP), as this blocks tools like 'cp -r'
+def chmod(path, mode):
+    #TODO: maybe this can be handled better?
+    return 0
+
+def chown(path, uid, gid):
+    return 0
```
markus2330, line 218: no newline

scripts/fuse/elektra_fuse/elektra_util.py
```diff
@@ -0,0 +1,196 @@
+from pathlib import Path
+import os, re, errno
+
+import kdb
+
+
+#TODO: Add 'default:' and 'proc:'. These namespaces do not work out of the box like "user:", "sysem:", ... but need extra treatment.
```
markus2330, line 7: See #3651

scripts/fuse/elektra_fuse/elektra_util.py
```diff
@@ -0,0 +1,196 @@
...
+elektra_namespaces = ["user:", "system:", "dir:", "spec:", "cascading:"]
+
+dir_file_special_name = "@elektra.value"
```
markus2330, line 10:
```suggestion
dir_file_special_name = "®elektra.value"
```

scripts/fuse/elektra_fuse/mock_context.py
```diff
@@ -0,0 +1,72 @@
...
+
+def _mock_process_context_and_run(process_context, func, args, kwargs):
+
+    #libelekra depends both on $HOME and the real user id (not the effective user id!) to resolve the user: namespace
+    #therefore, temporarely changing the effective user id does not work
+
+    #mock cwd
+    os.chdir(process_context.cwd)
+
+    #mock environment (minus $HOME)
```
markus2330, line 40: Should not be necessary, see #3651

scripts/fuse/elektra_fuse/mock_context.py
```diff
@@ -0,0 +1,72 @@
...
+
+    #in case euid == ruid, not sure how to set euid without potential PermissionError
+
+    # os.setegid(process_context.egid)
+    # os.seteuid(process_context.euid)
+    
+    os.setgid(process_context.rgid)
+    os.setuid(process_context.ruid)
+
+    #mock argv
```
markus2330, line 57: Also see #3651

scripts/fuse/elektra_fuse/rootlevel_resolver.py
```diff
@@ -0,0 +1,286 @@
...
+    def __init__(self, mountpoint):
+        self.mountpoint = mountpoint
+        self.fd = 0
+
+    #e.g '/' and '/12' are considered rootlevel
+    @staticmethod
+    def is_rootlevel_path(path):
+        return len(Path(path).parts) <= 2
+
+    # resolves a path "/pid/suffix" to "/suffix", where pid need to be a valid process id
```
markus2330, line 45:
```suggestion
    # resolves a path "/<pid>/suffix" to "/suffix", where pid need to be a valid process id
```

scripts/fuse/elektra_fuse/rootlevel_resolver.py
```diff
@@ -0,0 +1,286 @@
...
+        potential_ancestor_path = Path(potential_ancestor_path).resolve()
+        return potential_ancestor_path in [*potential_child_path.parents, potential_child_path]
+
+    @with_translated_exceptions
+    def readdir(self, path, fh):
+
+        if path == "/":
+            all_pids = self._get_all_pids()
+
+            #filter out processes with working dir under mountpoint to prevent endless recursions (e.g. an interactive shell navigatung below the mountpoint)
```
markus2330, line 108:
```suggestion
            #filter out processes with working directory under mount point to prevent endless recursions (e.g. an interactive shell navigating below the mount point)
```

markus2330 (commented):
Amazing work, a lot of improvements! Please also add your implementation as hightlight in the release notes. Hopefully we can already release them with 0.9.5 (maybe as "beta"?)

scripts/fuse/elektra_fuse/elektra_util.py
```diff
@@ -0,0 +1,203 @@
...
+        try:
+            new_value_as_string = new_value.decode(encoding="utf-8", errors="strict")
+            key.value = new_value_as_string
+        except UnicodeDecodeError:
+            key.value = new_value
+
+        db.set(ks, path) #using key instead of path here deleted the key
+
+
+#unexpected behavior, this deletes child keys too (why?)
```
kodebach, line 67: > this deletes child keys too (why?)

Because you used `ks.cut (key)` which is defined as "remove everything below (and including) `key` from `ks`, put it into a new KeySet and return that".

If you just want to delete a single key by name, you should use `keyDel(ksLookupByName (ks, name, KDB_O_POP))` (not sure how exactly that is exposed in Python).

kodebach (changes requested):
I've only had time for a look at the documentation and a very brief look at the code.

In general, the FUSE binding seems okay to me (assuming it works, haven't had time to test it either). However, I'm not a fan of the `<mount-point>/<pid>/<namespace>/...` structure. I assume other structures that cause less duplication (e.g. `system:/` should always be the same) would be to complicated. But maybe you could at least add a mechanism similar to `/proc/self` so a process doesn't need to know it's own PID.

And maybe you can add a `<mount-point>/user` section that doesn't map the `proc:/` namespace and resolves to the current user. I suspect most people would find that very useful. Most people will probably use the FUSE binding as a replacement for the `kdb` tool (i.e. to configure the KDB) and not for reading your applications config inside your code (the existing bindings are much better at doing that). To do this, you don't need access to the `proc:/` namespace (which is read-only anyway) and normally don't need to modify the `user:/` namespace of a different user (which would still be possible via `su` or `sudo -u` anyway).

kodebach:
Also you should take a look a the [output on lgtm.com](https://lgtm.com/projects/g/ElektraInitiative/libelektra/rev/pr-f7d3c1c8c2d797b35159b0888e865d586dc6dfa9). If something is detected incorrectly, just ignore it, but otherwise please fix.

scripts/fuse/README.MD
markus2330: >Some (older) file systems may have trouble with the ® character (or any other non-ASCII characters).

As we implement a file system here, this should not be a problem. Terminals or language settings not supporting the ® character, however, can be a problem. In specific, the fuse Docker image we provide doesn't have locales installed, so ® does not work (at least when I tried the last time).

In any case, we should follow our own standard, so either, we

1. use ®elektra
2. also reserve ®elektra.value (or even ®elektra.*)

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+## Binary-Keys
+
+If a value is to be written that cannot be decoded using the systems default encoding, it is treated as binary, and the appropriate `meta:/binary` meta-key is set.
+
+## Known issues
+
+- Moving an ``@elektra.value``-file corrupts the filesystem/key database (See this [issue](https://issues.libelektra.org/3648)). As editors like `vim` may move a file during editing, such editors cannot reliably be used on these pseudo-files.
+- Only functional on UNIX-like systems
+- chmod/chown is not implemented and does not signal a "not supported error" to enable compatibility with common tools like `cp`.
+- Access related errorcodes not yet implemented (e.g. normal user tries to write to system:/) (defaulting to read only fs - error)
```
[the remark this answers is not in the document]
markus2330: I see reporting Elektra's errors to the user within the scope of the thesis.

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+- without the need to build with libelektra
+- with additional features: e.g. the ability to effortlessly write binary keys
+
+## Quickstart
+
+### Docker
+
+Ensure docker is installed and run:
+```sh
+./run_new_container.sh
```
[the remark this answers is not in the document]
markus2330: @kodebach were you able to reproduce?

markus2330 (commented):
Responses to @kodebach.

@kodebach: Thank you for the review!

scripts/fuse/README.MD
kodebach: I would suggest we use the `®elektra` folder if possible. From a general point of view it makes more sense for Elektra to only reserve `®elektra` as a key name part. Since the KDB is hierarchical anyway there is no need to reserve all key name parts that start with X.

scripts/fuse/README.MD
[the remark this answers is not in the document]
kodebach: I do get this message, but afterwards the container still builds and starts up. The script just doesn't account for the fact that there might not be an existing container. Some `if`s could fix that, but a simple `> /dev/null` might work too.

markus2330:
>But maybe you could at least add a mechanism similar to /proc/self so a process doesn't need to know it's own PID.

Yes, I like this idea.

> Most people will probably use the FUSE binding as a replacement for the kdb tool (i.e. to configure the KDB) and not for reading your applications config inside your code (the existing bindings are much better at doing that).

The idea is that admins can see the config exactly as the process sees it. The `kdb` tool currently cannot do this.

scripts/fuse/README.MD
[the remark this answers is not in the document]
dev2718: Resolved in 2d8bd32adaa0eb746f3787ca3db02d80bdf5c054

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+### Process Context Mocking
+
+All Elektra operations are performed "from the perspective" of the respective process.
+This is done by only calling libelektra/kdb from special "mock"-processes that mimic all of the relevant attributes Elektra uses to resolve its keys.
+
+The mocked attributes are:
+- the user and group (for the `user:` namespace)
+- working directory (for the `dir:` namespace)
+- environment variables (for the `proc:` namespace)
+- in the future: the process arguments
```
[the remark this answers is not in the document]
dev2718: Could you further elaborate? (What exactly is already in the proc: namespace?)

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+This (re)creates the container if necessary and starts an interactive session with keys for testing already created (see `docker/create_keys.sh`).
+The filesystem is mounted below `~/mount`.
+For debugging purposes, a logfile is written to `~/nohup.out` inside the container. 
+
+### Native
+
+Required (debian) packages:
+- libfuse-dev
+- fuse
+- libelektra5-all (not in official repositories, follow [these](https://www.libelektra.org/installation/installation) instructions)
```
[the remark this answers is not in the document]
dev2718: Dependencies are listed, nothing needed for building as nothing is built. Minimum Elektra version mentioned as of f0cd885d86f02425bcf208ddad30bafc02ce3aa2

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
+Below any such pid-directory, the elektra key database is mounted bidirectionally using directories that correspond to the different namespaces:
+
+- `user:`
+- `system:`
+- `dir:`
+- `spec:`
+
+Exceptions (at the moment, these namespaces behave differently)
+- `default:`
```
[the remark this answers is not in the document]
dev2718: Readme updated (f0cd885d86f02425bcf208ddad30bafc02ce3aa2)

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+Below any such pid-directory, the elektra key database is mounted bidirectionally using directories that correspond to the different [namespaces](https://www.libelektra.org/tutorials/namespaces):
+
+- `user:`
+- `system:`
+- `dir:`
+- `spec:`
+
+Exceptions (at the moment, these namespaces behave differently)
+- `default:`
+- `proc:`
```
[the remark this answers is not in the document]
dev2718: Easily implementable in future version, mentioned in f0cd885d86f02425bcf208ddad30bafc02ce3aa2

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
+Furthermore, there is `cascading:`, which correspons to the `/` Elektra-namespace.
+This is useful for inspection purposes, but beware:
```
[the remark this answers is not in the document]
dev2718: Elaborated in next thesis draft

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+Elektra cannot map perfectly to the structure of a classical filesystem as described [here](https://www.libelektra.org/docgettingstarted/big-picture).
+Whenever deviations are necessary, the goal is to maximize usability and adhere to the expectations of common Unix tools.
+
+### Intermediate directories
+
+In Elektra, there is no need to have a `path` of continous keys between ancestors.
+E.g. the existence of `user:/a` and `user:/a/b/c` does not imply the existence of `user:/a/b`.
+Such intermediate keys are mapped to "virtual" directories, s.t. navigation in an interactive manner is possible and the directory tree is continous.
+
+A side effect is that creating/deleting a single key can create/delete multiple "virtual" directories at once.
```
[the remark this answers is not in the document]
dev2718: `touch` and `mkdir` do behave differently in the implementation (mkdir creates the meta-key "fuse/directory")

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+Another is that copies created with `cp -r` will not have these vacuous intermediaries but instead empty keys.
```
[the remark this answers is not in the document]
dev2718: See cd892f3ce0a60971894e8b959af1a736eccdfd8c

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
```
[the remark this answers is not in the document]
dev2718: See cd892f3ce0a60971894e8b959af1a736eccdfd8c

scripts/fuse/README.MD
dev2718: Since 66acb118f758035459212e3e730bc4e001d430b9 the functionality of `kdb file` is implemented natively.
Or do you propose the ability to query the file name for example as an xattr?

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+The filesystem attributes are sourced from the file obtained by `kdb file`.
+For some keys, no such file is returned, or the returned file does not exist in actuality. In these cases, fake attributes are used to preserve usability.
+
+To increase compatibility, currently the timestamps from these files are not used, as this prompts software like `vim` to produce warnings/errors.
+
+## Meta-Keys
+
+Meta-keys are mapped bidirectionally to extended file attributes (xattrs).
+To enable interoperability with other tools, the `meta:/` prefix is optional when reading/writing accessing meta keys. The listing of attributes however will show them prefix-less.
```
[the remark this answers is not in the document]
dev2718: Resolved in a4cc4e0cf020690b0601f486f4394e94d84b3c58

scripts/fuse/README.MD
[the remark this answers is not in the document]
dev2718: As far as I know FUSE does not support custom error messages, we only have the standard UNIX-error codes available.

scripts/fuse/docker/Dockerfile
dev2718: Imho because it serves the singular purpose of testing the fuse implementation interactively, therefore it should be where it is needed.
Of course they could be moved and long relative file paths to them introduced, but the benefit is not clear to me.

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
dev2718: Resolved in 530e3d46acc3efed9b01eaacfbb93ff1f43aa8a9

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
dev2718: Regrettably, in my tests I could not pinpoint the problem to atime. The cause remains unknown.

scripts/fuse/elektra_fuse/elektra_util.py
dev2718: Could you further elaborate on how #3651 pertains to this?

scripts/fuse/elektra_fuse/elektra_util.py
dev2718: Removed the offending method in 378c276ca07183e62180a72391eaaafa93b12c97. (kdb is used directly as a remedy)

scripts/fuse/README.MD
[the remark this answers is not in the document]
markus2330: Most file systems simply log what they wanted to say if they cannot via errno, e.g. they report file corruption that way. KDB reporting errors is very similar as it is also "the layer below". You could simply use [syslog](https://docs.python.org/3/library/syslog.html).

Special files would be another option to report errors (like lost+found), but that would conflict with your stateless approach.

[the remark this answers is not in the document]
dev2718: Just to avoid any misunderstandings:

I am supposed to catch (unexpected, e.g. binary writes to meta-keys) Elektra exceptions and log them via syslog in my implementation?

scripts/fuse/elektra_fuse/mock_context.py
markus2330: #3651 allows you to pass your own argv, envp. So you basically:

1. open /proc/pid/cmdline and store content to argv
3. open /proc/pid/environ and store content to envp
4. when doing kdb.goptsContract pass this argv&envp (instead of sys.argv, os.environ)

see also https://master.libelektra.org/src/bindings/swig/python/examples/example_gopts.py

scripts/fuse/README.MD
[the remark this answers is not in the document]
markus2330: No, things like binary writes to meta-keys can be done via normal errno codes maybe with ENOTSUP.

The logger would be used when kdbOpen/kdbGet/kdbSet return errors.

scripts/fuse/elektra_fuse/mock_context.py
dev2718: Thanks, but how to call the interface (or where to get the required arguments) was not of concern, but rather whether a call has any effects on future queries to the namespaces `user:` and `dir:`.

markus2330: I thought that the implementation is completely stateless, so that any query to FUSE can never influence another query? So I am still not sure what the question is.

These 3 statements should only influence the proc: and the cascading / namespace, and should have no effect on user: or dir: (nor system: nor proc:) namespaces. Does this answer your question?

scripts/fuse/README.MD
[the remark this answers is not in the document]
markus2330: command-line arguments and environment variables, see also https://github.com/ElektraInitiative/libelektra/pull/3657#discussion_r606660739

scripts/fuse/README.MD
```diff
@@ -0,0 +1,166 @@
...
+
+Exceptions (at the moment, these namespaces behave differently)
+- `default:`
+- `proc:`
+
+Furthermore, there is `cascading:`, which correspons to the `/` Elektra-namespace.
+This is useful for inspection purposes, but beware:
+Write operations in this namespace produce unexpected effects (and create operations might fail), as for the same "address" a write operation can be resolved to a different real key then the corresponding read operation.
+
+These first two layers, (e.g `mountpoint/12/system:`) are read only. Deeper layers support writing: any change is directly updated in the backing Elektra-Key-Database.
```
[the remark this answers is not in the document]
markus2330: Btw. it is good if you also push a commit fixing the problem when you resolve conversations.

scripts/fuse/elektra_fuse/mock_context.py
kodebach: Take a look at `src/bindings/swig/python/examples/example_gopts.py`. Specifically this bit:

https://github.com/ElektraInitiative/libelektra/blob/5702ce0496823fdd12c4e1e6ff9d6d1ca1b71a40/src/bindings/swig/python/examples/example_gopts.py#L72-L77

You can pass any Python lists to `kdb.goptsContract`, it doesn't have to be `sys.argv` and `os.environ`. So in your case it would be something in `process_context`.

You should then pass the `contract` every time you call `kdb.KDB()`, e.g. here
https://github.com/ElektraInitiative/libelektra/blob/378c276ca07183e62180a72391eaaafa93b12c97/scripts/fuse/elektra_fuse/elektra_util.py#L37-L42

Strictly speaking, you should only need the `contract` when you want to access a `proc:/` or cascading key, but if you are unsure just use it everywhere.

---


> whether a call has any effects on future queries to the namespaces `user:` and `dir:`.

The `gopts` plugin only adds keys into the `proc:/` namespace. So using `goptsContract` does not have any influence on `user:/` or `dir:/` keys. It may, however, affect which keys are returned by a cascading lookup.

The `goptsContract` also only applies to the one instance that receives it during `kdbOpen()` (or `kdb.KDB()` in the Python Binding). Other instances that existed before or are created later on without the contract are not affected.

markus2330:
> On the other hand, I don't think that using ®elektra/value instead of ®elektra.value really reduces usability. It's just one character difference. So unless the implementation is really dramatically more complicated I don't see the problem here.

Yes, let us simply depend on UTF-8 locales and install locales in the Docker container

I agree with @kodebach that people want the FUSE bindings mainly in its installed form, so it would be great if the FUSE binding would install like the rest of Elektra. src/tools/pythongen/ is an example of how python programs can install together with Elektra's CMake system (unfortunately python2, probably a lot has changed since then on the python side but the CMake integration should be more or less the same).

Yes, `®elektra/value` is best (directory with one file "value" below), then we also have possibilities for extensions. The compromises sound horrible to me.

scripts/fuse/README.MD
markus2330: >    reduces usability (one file versus one directory with always one file in it)

Yes, it is pressing 2x tab instead of one. The question is if there is a possibility that more files get added in the future, e.g. for exporting functionality?

> Furthermore, again because of usability considerations, I'm not happy with the use of ® instead of @ due to the mentioned encoding problems. Imho, any such naming clash is extremely unlikely in practical use cases, whereas the pain of obfuscated encodings is real.

I see your point. Is there also a possibility of obfuscated encodings except if the locales package is not installed at all? We chose ® exactly because it also has somewhat reasonable encoding if the wrong locale is chosen (even in ASCII).

@kodebach there is still a "TODO: link to detailed ®elektra document" in doc/KEYNAMES.md. Would be better to have this document then long discussions here in this sub-thread never to be found again. We can continue the discussion in the PR.

I am open to `@elektra` but then `@elektra` and not `®elektra` should be the reserved key name.

Btw. I didn't see the initial response https://github.com/ElektraInitiative/libelektra/pull/3657#discussion_r604835066 of @dev2718 when I wrote https://github.com/ElektraInitiative/libelektra/pull/3657#issuecomment-814239192

scripts/fuse/README.MD
markus2330: Is this maybe also a candidate for a file below `®elektra`?

scripts/fuse/elektra_fuse/elektra_util.py
markus2330: Please create issues if there are problems with the Python binding. I created an issue now: #3749

@dev2718 wrote:
> kdb is used directly as a remedy

What do you mean? Is there some fallback where the `kdb` CLI tool is used?

scripts/fuse/elektra_fuse/mock_context.py
markus2330: Btw. and working directory is a completely different topic.

It works with `os.chdir(path)` (before opening `kdb`). It changes keys in `dir:` and cascading namespace.

scripts/fuse/elektra_fuse/mock_context.py
markus2330: Btw. and working directory is a completely different topic.

It works with os.chdir(path) (before opening kdb). It changes keys in dir: and cascading namespace.

scripts/fuse/elektra_fuse/mock_context.py
dev2718: Thank you for the detailed elaboration! It turns out, my misunderstanding stemmed from me assuming all lines marked in green in the initial comment were of concern, whereas in reality only the last one was relevant.

The functionality is implemented as of a4f59b349a6eefd327facb1322b802da23fa2289.

scripts/fuse/README.MD
dev2718: As of f13604c3f71c210cbc9dbc115069d96ec50ba9ad `®elektra` is standard, but configurable via command line option.

scripts/fuse/README.MD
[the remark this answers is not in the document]
dev2718: As of 83d67af36959a3181e9641ccd1cf299ce10961f5 there is configurable logging (via command line options:(stdout/syslog/none)).  Exceptions thrown by Elektra are also logged.

scripts/fuse/docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+FROM ubuntu:focal 
+
+RUN apt-get -y update
+RUN apt-get -y update
```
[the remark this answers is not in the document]
dev2718: As of c2e0bda9600d495ee18242852c23cc843464921b the file backing a key can be queried as an xattr. (By default `®elektra.file`)

scripts/fuse/README.MD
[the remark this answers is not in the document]
dev2718: Cascading writes in the FUSE implementation are disabled since e42cfd8a1eec378550556d4242a0a5ee2ac1a095.

See also #3794.

scripts/fuse/README.MD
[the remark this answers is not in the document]
dev2718: See https://github.com/ElektraInitiative/libelektra/pull/3657#discussion_r615415441.

scripts/fuse/README.MD
dev2718: See in https://github.com/ElektraInitiative/libelektra/pull/3657#discussion_r615417258.

scripts/fuse/docker/Dockerfile
dev2718: It is only intended for developers, normal users only need the FUSE python dependency as described in README.md

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
dev2718: You are right, I removed the comment already.

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
dev2718: See 66acb118f758035459212e3e730bc4e001d430b9.

scripts/fuse/elektra_fuse/elektra_fuse_interface.py
dev2718: Removed already.

scripts/fuse/elektra_fuse/elektra_util.py
dev2718: `proc:` is now available as a read-only namespace.

scripts/fuse/README.MD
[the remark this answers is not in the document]
dev2718: I added an example as you proposed in 63e59cfd461611eca3f219c832beb5ae97c5b9c9.

scripts/fuse/docker/Dockerfile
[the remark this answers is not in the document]
dev2718: The double update will be removed in a future release.

markus2330:
Can you give a status update here? Do you need help with the build server?

dev2718:
> Can you give a status update here? Do you need help with the build server?

All things from here https://github.com/ElektraInitiative/libelektra/issues/3599 (except a few mission uncritical todos regarding possible improvements to status/error codes and the proposed remormulation of README.md into shell recorder format) are done.
I do not know how the FUSE module relates to the buildserver. Nothing is built/complied/packaged currently.
Do you mean the linting done by lgtm-com? (Reporting mostly things that are quick to fix, e.g unneeded imports).

mpranj:
> I do not know how the FUSE module relates to the buildserver.

He just asked if you need any hints how to make the buildservers happy.

In other words, before we merge the FUSE module
1. All CI builds must be green, thus not break master builds.
2. Tests should be added and executed for the new module, while still not breaking CI.
3. Ideally no new LGTM issues are introduced.

dev2718:
> > I do not know how the FUSE module relates to the buildserver.
> 
> He just asked if you need any hints how to make the buildservers happy.
> 
> In other words, before we merge the FUSE module
> 
> 1. All CI builds must be green, thus not break master builds.
> 2. Tests should be added and executed for the new module, while still not breaking CI.
> 3. Ideally no new LGTM issues are introduced.

I indeed have problems with the build process, as some issues raised are clearly outside my code, and others appear irrelevant/errounious as e.g. syntax validation for bash on a jpeg file meant for testing, the requirement for a shebang line on files not meant to be executed (e.g xml files) do not seem appropriate. Furthermore, the system complains about naming conventions outside my scope, eg. `KeySet()`

mpranj:
> syntax validation for bash on a jpeg file meant for testing

Most issues appeared because the `script` directory was originally meant for shell scripts (with some exclusions). I'll update the exclusions (add jpeg, python files) and ping you when it's done.

Many other things have been fixed if you rebase from master. To run the full test-suite on Jenkins, you will need to add some release notes though.

The simplest is that you rebase when I fix the exclusions.

scripts/fuse/README.MD
```diff
@@ -0,0 +1,183 @@
...
+Below any such pid-directory, the elektra key database is mounted bidirectionally (i.e. read-write capable) using directories that correspond to the different [namespaces](https://www.libelektra.org/tutorials/namespaces):
+
+- `user:`
+- `system:`
+- `dir:`
+- `spec:`
+
+Exceptions:
+- `default:` (only exists for cascading lookups)
+- `proc:` (to be implemented in future version (read-only))
```
markus2330, line 62: Docu not updated?

scripts/fuse/README.MD
```diff
@@ -0,0 +1,183 @@
...
+
+Furthermore, there is `cascading:`, which correspons to the `/` Elektra-namespace.
```
markus2330, line 64: Why is it not in the list above?

scripts/fuse/README.MD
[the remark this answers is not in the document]
markus2330: This line in the README.md still needs to be updated, though.

markus2330 (changes requested):
Yet another review of README.md

mpranj:
> Conclusion: In its current form the FUSE module makes more sense in "src/tools", then everything can stay together as is.

Makes sense to me.

> I assume the exclusions would also work in "src/tools"?

Then the exclusion would not be needed. The `check_bashisms` test only looks for files in the scripts directory afaik.

src/tools/fuse/README.md
```diff
@@ -0,0 +1,189 @@
...
+- `fuse` (for example via the debian package `fuse`) 
+- `python3 >= 3.8`, `pip3`
+
+Required python3 packages (will be installed by `pip3` automatically if not present):
+- `fusepy>=3.0.1`
+- `psutil>=5.8.0`
+
+Build the `elektra_fuse` python3-wheel and install it using `pip3` with:
+```
+./build.sh && sudo ./install.sh
```
markus2330, line 42: This should be integrated in a `make install`.

src/tools/fuse/README.md
```diff
@@ -0,0 +1,189 @@
...
+```
+This, if needed, will also install the python-requirements `fusepy` and `psutil`.
+`install.sh` is run as root in order to make the new shell command `elektra_fuse` globally available (if run as a normal user, `$PATH` needs to be adapted as specified in the command's output).
```
markus2330, line 45: Better to install elektra_fuse to Elektra's tools (like the other tools), so that `kdb mount-fuse` (or similar) can be written.

src/tools/fuse/rebuild_and_install.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+sudo ./uninstall.sh
+sudo ./build.sh
```
markus2330, line 3:
```suggestion
./build.sh
```

Does building really need sudo?

src/tools/fuse/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[metadata]
+license_files = LICENSE.md
```
markus2330, line 2: Can you specify the top-level LICENSE.md here?

markus2330 (commented):
I could not test due to `Not Found errors` while installing packages (probably temporary problems). Please make the build server happy so that we can merge this and have follow-up changes in new PRs.

src/tools/fuse/README.md
dev2718: Would the following pose a proper way to integrate it?
- Hook into Elektras installation procedure and install the elektra_fuse python package to the global scope via pip3
- Expose it in kdb by having a new subcommand "mount-fuse" that essentially performs a subprocess call along the lines of "python3 -m elektra_fuse"

src/tools/fuse/setup.cfg
dev2718: Yes, I changed it to use the top-level license in cc580244ec501d2e8c18b4e4f45e216bb7fd9155.

src/tools/fuse/README.md
dev2718: Migrated to a makefile-workflow as of e4d3715c8485d00822d54e1dc5767b59826f39a9.

src/tools/fuse/rebuild_and_install.sh
dev2718: No - Thanks for pointing that out.

src/tools/fuse/LICENSE.md
```diff
@@ -0,0 +1,28 @@
+# BSD 3-Clause License
```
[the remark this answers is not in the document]
dev2718: Only the root-license is used as of cc580244ec501d2e8c18b4e4f45e216bb7fd9155.

src/tools/fuse/README.md
markus2330: >  Hook into Elektras installation procedure and install the elektra_fuse python package to the global scope via pip3

It is not simply anywhere globally but the CMake installation paths should also be respected. In particular, the installation should be below ${CMAKE_INSTALL_PREFIX}. As said, src/tools/pythongen/CMakeLists.txt already does something like that.

> Expose it in kdb by having a new subcommand "mount-fuse" that essentially performs a subprocess call along the lines of "python3 -m elektra_fuse"

Exactly, it is basically simply installing the script to ${CMAKE_INSTALL_PREFIX}/${TARGET_TOOL_EXEC_FOLDER}. Please also add metadata so that scripts/kdb/find-tools will work (see also doc/help/kdb-find-tools.md).

I would very much appreciate if you can document these steps in src/tools/README.md (as all tools should respect kdb-find-tools and install in CMake paths). And don't forget to also add the FUSE tool there.

markus2330:
Also here something went wrong in the rebase.

> I do not know how the FUSE module relates to the buildserver. Nothing is built/complied/packaged currently.

It should be built, or even better: packaged. Nevertheless, even if it is not the build server can fail due to many different reasons, e.g. wrong formatting.

markus2330:
If we lose the whole commit history you can also create a new PR. All comments are tangling here.

dev2718:
> If we lose the whole commit history you can also create a new PR. All comments are tangling here.

Sorry for the hassle, the rebase should be sane now.

src/tools/fuse/README.md
dev2718: a32e9e9b554c43cf6f420280fcbebf6ef02603ac adds CMake integration satisfying these requirements, src/tools/README.md I will update next.

dev2718: I updated src/tools/README.md with information on the fuse tool and the required steps for new tools in 0df6a502ef3de4bad250ef74873060f522a297ff.

dev2718:
@markus2330 There seems to be a problem with the CI: 
It fails during `Scanning dependencies of target elektra-ni` with `Caught the following exception: null`. Also, the precise point of failure seems to be nondeterministic: A previous run using the same code resulted in the same error in a different step (`Scanning dependencies of target elektra-mmapstorage_crc`), this time with the additional output `Sending interrupt signal to process`. Maybe this is a timeout issue?

markus2330:
@robaerd can you have a look at the CI problem?

robaerd:
> @robaerd can you have a look at the CI problem?

Jenkins CI is failing because the `formatting-check` stage failed. If one stage fails, the other running stages are cancelled. This is also the reason why @dev2718 saw `Scanning dependencies of target elektra-ni` as last log output, because this stage was aborted at this position, due to the failing formatting-stage.

@dev2718 simply merging the restyled PR (https://github.com/ElektraInitiative/libelektra/pull/3924) should solve this issue. Otherwise just apply the suggestions of the `formatting-check` stage (https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3657/49/pipeline/401).

dev2718:
> > @robaerd can you have a look at the CI problem?
> 
> Jenkins CI is failing because the `formatting-check` stage failed. If one stage fails, the other running stages are cancelled. This is also the reason why @dev2718 saw `Scanning dependencies of target elektra-ni` as last log output, because this stage was aborted at this position, due to the failing formatting-stage.
> 
> @dev2718 simply merging the restyled PR (#3924) should solve this issue. Otherwise just apply the suggestions of the `formatting-check` stage (https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3657/49/pipeline/401).

Thank you for helping me figure this out! Merging the restyled PR alone did not suffice - it did not include all affected files, but applying the patch from from the logs did the job.

Now it reaches the testing stage, but is then interrupted by this:
I unfortunately don't know what to do with this message; do you know what the problem is? 
```
INFO: Processing CTest-Version 3.x (default)
[CTest-Version 3.x (default)] - No test report file(s) were found with the pattern 'build directory/Testing/**/*.xml' relative to '/home/jenkins/workspace/libelektra_PR-3657' for the testing framework 'CTest-Version 3.x (default)'.
Did you enter a pattern relative to (and within) the workspace directory?
Did you generate the result report(s) for 'CTest-Version 3.x (default)'?"
```

mpranj:
The spell-check failed and then again, the other test was aborted:
https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3657/51/pipeline/397

When you see `Sending interrupt signal to process`, this usually means that another test failed and then all tests were aborted.

mpranj:
@markus2330 @kodebach I think this is good to go into the release. Are there any blockers from your perspective, as you followed this PR more closely?

kodebach:
I don't know of any problems.

markus2330 (approved):
Lets merge this for the release. The release notes need to be improved, though:

1. say that it is very experimental
2. link to the tool

doc/news/_preparation_next_release.md
```diff
@@ -33,6 +33,8 @@ docker run -it elektra/elektra
 
 ## Highlights
 
+- add FUSE tool _(Alexander Firbas)_
```
mpranj, line 36: Something like this @markus2330 ?

```suggestion
- Add **experimental** preview of Filesystem in User Space [FUSE](../../src/tools/fuse/README.md) tool. This tool enables the inspection and modification of the KDB, in the form of a classical filesystem. _(Alexander Firbas)_
```

mpranj:
@dev2718 please fix the new alerts from lgtm in the next PR. (at least the unused imports)

mpranj:
Thank you for the new FUSE tool @dev2718, amazing work! :rocket:

src/tools/fuse/README.md
markus2330: What is the current state? Did you try to package the fuse binding?

PR #3702, opened 2021-03-10
[CM H0] updated wording + improved markdown code example

aaronabebe:
Minor rewording for H0 of CM2021.


markus2330:
Thank you very much!

Issue #3269, opened 2019-11-23
maven build failure


markus2330:
https://travis-ci.org/ElektraInitiative/libelektra/jobs/615962665

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test (default-test) on project libelektra4j: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test failed: Plugin org.apache.maven.plugins:maven-surefire-plugin:2.12.4 or one of its dependencies could not be resolved: Could not transfer artifact org.apache.commons:commons-lang3:jar:3.1 from/to google-maven-central (https://maven-central.storage-download.googleapis.com/maven2/): GET request of: org/apache/commons/commons-lang3/3.1/commons-lang3-3.1.jar from google-maven-central failed: Connection reset -> [Help 1]

[ERROR] 

[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.

[ERROR] Re-run Maven using the -X switch to enable full debug logging.

[ERROR] 

[ERROR] For more information about the errors and possible solutions, please read the following articles:

[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginResolutionException
```

Piankero:
Did this occur in other places too?

markus2330:
I do not know.

Is this maybe only some network problem?

What about adding `-e` to the mvn calls?

Piankero:
> Is this maybe only some network problem?

Thats why I was asking. To me it looks like a temporary network issue.

> What about adding -e to the mvn calls?

The main message is there so I do not think that we need an extra stacktrace

Let's wait some time and see if this happens again.

markus2330:
> The main message is there so I do not think that we need an extra stacktrace

Yes, but in general `-e` might be better? If it is successful, we will not see anything anyway?

Piankero:
> Yes, but in general -e might be better? If it is successful, we will not see anything anyway?

In your posted message the relevant part is already given (`Connection reset`). Adding `-e` to it would just give you an additional huge stacktrace that is unnecessary.

markus2330:
Lets better turn to more important things.

markus2330:
Happened again: https://github.com/ElektraInitiative/libelektra/pull/3549/checks?check_run_id=1492810253

mpranj:
[...]

Issue #3171, opened 2019-11-07
[CM P0] implement external iterators


markus2330:
As written in  #2991 we want to switch to external iterators. Which bindings do still make use of the internal iterators (ksNext and friends?). They should be removed/deprecated and an external iterator provided instead.

- [x] #4281
- [x] Java (remove internal)
- [X] Rust
- [X] Go
- [X] GLIB
- [ ] #4280
- [x] #4279

---

The functions that will be removed are:

```c
int ksRewind (KeySet *ks);
Key *ksNext (KeySet *ks);
Key *ksCurrent (const KeySet *ks);

elektraCursor ksGetCursor (const KeySet *ks);
int ksSetCursor (KeySet *ks, elektraCursor cursor);

int keyRewindMeta (Key *key);
const Key *keyNextMeta (Key *key);
const Key *keyCurrentMeta (const Key *key);
```

manuelm:
Both python and lua are using the CPP iterators. So if you remove them you are also removing them from python+lua.

Piankero:
The java implementation already uses an [external iterator](https://github.com/ElektraInitiative/libelektra/blob/38fc1a4046449d36b6007f70ede8395b4fac36c8/src/bindings/jna/libelektra4j/src/main/java/org/libelektra/KeySetIterator.java#L9)

Once ksNext() is removed, [this](https://github.com/ElektraInitiative/libelektra/blob/c4c31f047428a77c60d68cbcff52dbd2d7fa1c68/src/bindings/jna/libelektra4j/src/main/java/org/libelektra/KeySet.java#L256-L258) line should also be removed.

manuelm:
@Piankero I assume `ksRewind`, `ksNext`, `ksCurrent`

@markus2330 You should add Ruby. It's actually wrapping all three and also using them in the ruby iterator implementation

markus2330:
Thank you for the hint, I added Ruby (@BernhardDenner).

@Piankero can you fix Java, it should be quite easy for you to write an iterator?

Piankero:
> @Piankero can you fix Java, it should be quite easy for you to write an iterator?

There is already an external iterator written as I have written in the upper comment.

Piankero:
[...]

PR #3714, opened 2021-03-18
3713 java binding maven to gradle migration

tucek:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


markus2330:
Something went wrong in the rebase. Probably cherry-picking you own commits is the easiest way to resolve the problem here.

src/bindings/jna/gradlew
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env sh
+
+#
+# Copyright 2015 the original author or authors.
```
markus2330, line 4: Please also add this to doc/THIRD-PARTY-LICENSES

scripts/docker/debian/buster/Dockerfile
```diff
@@ -117,18 +116,19 @@ RUN useradd \
     --shell "/bin/bash" \
     jenkins
 
-# install gradle via sdkman (part 1/2) - update shell as root
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
+# download and install gradle
+RUN cd /tmp && wget https://services.gradle.org/distributions/gradle-6.8.3-bin.zip && unzip gradle-6.8.3-bin.zip && rm gradle-6.8.3-bin.zip && mv gradle-6.8.3 /opt/gradle
```
markus2330, line 120: Is gradle 4.4.1-6 (which is in buster) not working?

src/bindings/jna/gradlew
tucek: will do

scripts/docker/debian/buster/Dockerfile
tucek: IMHO no. Gradle 4.4 is ages old and there have been multiple potentially breaking changes since then.
Since the main purpose of my contribution is to bring the Java bindings up to scratch, i would rather refrain from using an outdated tool chain.

src/bindings/jna/libelektra/src/main/java/org/libelektra/plugin/NativePlugin.java
```diff
@@ -25,6 +25,7 @@
 	 * Constructor for loading an Elektra plugin
 	 * @param pluginName The plugin name
 	 * @param errorKey The errorKey
+	 * @param modules #3713 add parameter description and update other
```
mpranj, line 28: Description seems wrong.

src/bindings/jna/libelektra/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+				}
+
+				//developers {
+				//	developer {
+				//		id = 'johnd'
+				//		name = 'John Doe'
+				//		email = 'john.doe@example.com'
+				//	}
+				//}
+
```
mpranj, line 39: Remove if this is not needed?

```suggestion
```

src/bindings/jna/libelektra/src/main/java/org/libelektra/plugin/NativePlugin.java
tucek: Java doc is WIP... - I'm currently reworking the java doc

scripts/docker/fedora/33/Dockerfile
```diff
@@ -95,8 +97,20 @@ RUN useradd \
...
     --shell "/bin/bash" \
     jenkins
+
+# download and install gradle
+RUN cd /tmp && wget https://services.gradle.org/distributions/gradle-6.8.3-bin.zip && unzip gradle-6.8.3-bin.zip && rm gradle-6.8.3-bin.zip && mv gradle-6.8.3 /opt/gradle
+ENV PATH "${PATH}:/opt/gradle/bin"
+RUN echo "export JAVA_HOME=$(dirname $(dirname $(readlink $(readlink $(which javac)))))" >> /etc/profile.d/java.sh
+RUN echo "export PATH=$PATH:$JAVA_HOME/bin" >> /etc/profile.d/java.sh
+RUN echo "CLASSPATH=.:$JAVA_HOME/jre/lib:$JAVA_HOME/lib:$JAVA_HOME/lib/tools.jar" >> /etc/profile.d/java.sh
+RUN source /etc/profile.d/java.sh
```
mpranj, line 107: How much of this is truly needed? At least the `PATH` is set correctly by default afaik.

mpranj (commented):
I found some minor things in an initial review.

src/bindings/jna/libelektra/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+				}
+
+				// TODO #3731 - add developers responsible for code
+				developers {
+					developer {
+						id = 'libelektra'
+						name = 'Elektra Initiative'
+						email = 'elektra@libelektra.org'
+					}
+				}
```
robaerd, line 39: This part is required for the transition from the staging to release repository.
The [documentation](https://maven.apache.org/pom.html#developers) states that only the developers directly responsible for this code should be  mentioned.
It is also rather uncommon to list an organization as developer.
@markus2330 which people should we add here?

markus2330: Simply list @tucek as developer and @kodebach, me and you as contributor. (Or also as devs, however they see themselves.)

Generating the XML out of doc/AUTHORS.md seems to be over-the-top.

kodebach: > Generating the XML out of doc/AUTHORS.md seems to be over-the-top.

That is actually pretty much the opposite of what the docs suggest. To quote: "A good rule of thumb is, if the person should not be contacted about the project, they do not need to be listed here."

> It is also rather uncommon to list an organization as developer.

I don't think it is that uncommon. Especially, if a package is developed by a company.

Examples:

Android Libraries from Google just list "The Android Open Source Project" as developer. https://mvnrepository.com/artifact/androidx.activity/activity/1.2.2 

Official Kotlin Libraries also list various organizations:
https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib/1.5.0-M2
https://mvnrepository.com/artifact/io.ktor/ktor-client-core/1.5.3
https://mvnrepository.com/artifact/org.jetbrains.kotlinx/kotlinx-coroutines-core/1.4.3

Log4J doesn't list any developers and instead just defines two mailing lists:
https://mvnrepository.com/artifact/log4j/log4j/1.2.17

I think the most important part is that people know how to get in touch, if there is a problem. To that end listing `elektra@libelektra.org` is probably a good idea. Especially, since the people involved in Elektra could change.

src/bindings/jna/gradlew
tucek: @markus2330 done. do i need to list newly used third party software somewhere? (gradle, a few gradle plugins)

scripts/docker/fedora/33/Dockerfile
tucek: @mpranj the initial problem was JAVA_HOME not being set. i will check whether i can omit updating the PATH

mpranj: Ok, the cleanup can also be done in a follow-up PR.

src/bindings/jna/libelektra/build.gradle
markus2330: > That is actually pretty much the opposite of what the docs suggest. To quote: "A good rule of thumb is, if the person should not be contacted about the project, they do not need to be listed here."

I do not know what you mean, doc/AUTHORS.md is exactly the list of people that are to be contacted (for their respective modules). We do not distinguish devs and contributors, though.

> If it's also okay for @markus2330 then I will just keep the organization as the only developer.

Yes, if they allow it, I would prefer it. It is hard enough to maintain doc/AUTHORS.md. E.g. @tucek is not listed there and @robaerd only states "debian packaging" but "cpack packaging" or similar (next to other things like Docker images) would be more appropriate. Some people without any code in the current repo are also to be removed. I added the task to  #3746

kodebach: > I do not know what you mean, doc/AUTHORS.md is exactly the list of people that are to be contacted (for their respective modules).

AFAIK `doc/AUTHORS.md` is supposed to list all people that contributed to Elektra (that's not the case, but that was the idea I think). I wouldn't really want to be listed as a point of contact, just because I wrote a few lines of code somewhere. Listing the general contact info makes a lot more sense. Often somebody other than the person who wrote the code might be able to help as well.

Also, by generating the Maven authors list from `AUTHORS.md` we may run into out-of-sync issues. What if we publish the Maven package today and tomorrow `AUTHORS.md` changes?

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
```diff
@@ -0,0 +1,325 @@
...
+
+	/**
+	 * Singleton instance of the native library proxy.
+	 */
+	Elektra INSTANCE = Native.loadLibrary("elektra-kdb", Elektra.class);
+
+	// KDB methods
+
+	/**
+	 * Opens the session with the Key database.<br >
```
markus2330, line 24: Why do you add such extensive documentation here. These functions are not supposed to be used directly, the interface is only a necessity to make JNA bindings work. Such a disclaimer should probably written next to "The JNA interface to libelektra.".

src/bindings/jna/libelektra/build.gradle
markus2330: > AFAIK doc/AUTHORS.md is supposed to list all people that contributed to Elektra (that's not the case, but that was the idea I think). I wouldn't really want to be listed as a point of contact, just because I wrote a few lines of code somewhere.

These people also don't get added.

> Listing the general contact info makes a lot more sense.

As most contact happens via the issue tracker, doc/AUTHORS.md is rarely used anyway.

> Also, by generating the Maven authors list from AUTHORS.md we may run into out-of-sync issues.

Shouldn't have mentioned the idea to generate the authors list :wink:

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
tucek: I've also thought about that. I would argue that for JNA binding developers, these java docs are relevant.
The other side of the story is maintaining doc sync with the C API might be too much of a burden. The alternative would be to minimize the documentation here and link to the C API docs.

scripts/docker/fedora/33/Dockerfile
tucek: It seems you are right and we don't need it after all... I'm not sure how i got the impression that we do, but it was late, that evening ;)
I've removed it from the fedora Dockerfiles for now.

tucek:
The following issue may be closed when this PR is merged: #3269

tucek:
https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3714/43/pipeline failed with non-zero exit value 134 on `fedora-32-full´. Seems like a resource problem. Any inputs?

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
markus2330: I would prefer linking to it, as it is supposed to be a 1:1 representation of the C API.

markus2330:

markus2330:
I cannot read from the output that it is about too little RAM*, SIGSEGV can be anything. As it happens in ksRewind, maybe we did something wrong in the JNA binding?

Btw. just because the tests run locally does not mean that there isn't a memory bug. Did you try to run with valgrind?

*Actually the machines have plenty of RAM (even i7 has 8GB).

kodebach:
`ksRewind` doesn't allocate any memory, so it can't be an out-of-memory problem. Looks more like an invalid pointer was passed to `ksRewind`.

src/bindings/jna/libelektra/src/main/java/org/libelektra/plugin/PropertiesStorage.java
```diff
@@ -1,6 +1,12 @@
 package org.libelektra.plugin;
 
-import java.io.*;
+import static org.libelektra.Elektra.KeyNewArgumentFlags.KEY_END;
+
+import java.io.BufferedInputStream;
```
markus2330, line 5: Are they used?

src/bindings/swig/python/kdb.i
```diff
@@ -296,8 +303,18 @@
       return ret
 
     def extend(self, list):
+      """Extend the keyset by appending all the items from the iterable"""
       return self.append(*list)
 
+    def remove(self, name):
```
markus2330, line 309: Problems with rebase?

markus2330 (approved):
Great job! So good to finally see this arriving.

src/bindings/swig/python/kdb.i
mpranj: I think you use `merge` instead of `rebase`, and some weird issues appear on github.

I keep my fork's master identical to upstream, and always rebase from master:
```
git checkout master
git fetch upstream master # assuming you have upstream set already
# do not do this if you ever work on your master branch
git reset --hard upstream/master # this is fine, I never work on master and I only want it to mirror upstream

git checkout mybranch
git rebase master
```
During this process you might have to fix some conflicts manually.
Maybe there are improvements to this process, but I think we **should** probably put this somewhere in the docs.

markus2330:
The ksRewind is reproducible. This needs to be fixed before merging this.

markus2330:
@kodebach help is probably very much appreciated, as this is probably hard to solve without C knowledge.

src/bindings/swig/python/kdb.i
kodebach: I would also recommend, that the `master` branch of your fork is kept identical to that of the main repo.

I have set up my remotes like this (output of `git remote -vv`):
```
origin	git@github.com:kodebach/libelektra.git (fetch)
origin	git@github.com:kodebach/libelektra.git (push)
upstream	git@github.com:ElektraInitiative/libelektra (fetch)
upstream	git@github.com:ElektraInitiative/libelektra (push)
```

To update the `master` branch I do:

```sh
git fetch upstream master:master
git push origin master:master
```

This works without switching to the `master` branch first. After that you can do `git rebase master` to rebase the current branch.

kodebach:
> @kodebach help is probably very much appreciated, as this is probably hard to solve without C knowledge.

I don't really have time right now to debug this (maybe I can make time next week). However, I also don't really know how to debug JNI/JNA code. You can probably use `gdb` somehow to debug the C code, but I have no idea how to set `gdb` up so that it runs the `testjna_gradle` tests, since there is no `testjna_gradle` executable. Probably you'll have to use a very very long command for the `java` executable. Maybe there is also a way to set this up via `gradle`, or just attach via a process ID.

For the last JNA bug I debugged, I just manually checked all possible locations for errors. In this case I would start by finding a way to run the tests through `valgrind`. Maybe that detects a double free somewhere. But again, no idea how to set up `valgrind` for this. Possibly using an ASAN build would be easier.

Note: Both for `valgrind` and ASAN builds I would expect quite a few false positives. So read the output carefully...

src/bindings/swig/python/kdb.i
markus2330: @kodebach can you copy what you wrote here to doc/GIT.md?

src/bindings/jna/libelektra/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+		libelektra(MavenPublication) {
+			groupId rootProject.group
+			artifactId 'libelektra'
+			version project.version
+
+			from components.java
+
+			pom {
+				name = "$project.name - $parent.name"
+				description = 'Low-level Elektra java binding to elekrify Java applications and write Elektra plugins using Java.'
```
mpranj, line 22:
```suggestion
				description = 'Low-level Elektra Java binding to elektrify Java applications and write Elektra plugins using Java.'
```

src/bindings/jna/libelektra/src/main/java/org/libelektra/plugin/NativePlugin.java
tucek: Will be addressed via #3754

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
tucek: I will resort to minimal duplication still resulting in a helpful javadoc. will be addressed via #3754.

src/bindings/swig/python/kdb.i
kodebach: Sure, but do we really need a git tutorial? Doesn't Elektra have enough barely maintained docs already?

src/bindings/jna/libelektra/src/main/java/org/libelektra/plugin/PropertiesStorage.java
tucek: What exactly do you mean by they?

tucek:
regarding the ksRewind problem:

1. I cannot reproduce the problem locally
2. the only places where ksRewind is (still) called is NativePlugin.java set(), get() and error() - all other functions related to the internal cursor of a key set have been removed since #3171
3. all references to any function using ksRewind are found in the follwing three test:
    1. ExceptionMapperIT::passErrorToPluginAndExecute
    2. ExceptionMapperIT::kdbSetWithWarning_shouldNotTriggerException
    3. ExceptionMapperIT::kdbSetWithWarningAndError_shouldHaveWarnings
4. Having a look at these 3 functions, I do not see any reason, why the pointer passed to ksRewind should not be valid
5. @kodebach Do we really have to rewind a key set passed to an elektra plugin kdbGet/kdbSet/kdbError?

markus2330:
> I cannot reproduce the problem locally

Did you also try the Docker containers?

> all references to any function using ksRewind are found in the follwing three test: 

Did you try to run with valgrind? It usually gives very detailed information about such problems (even if they do not occur locally).

As second step you could try to reduce the surface of the problem, especially if you are forced to do remote debugging. You can temporarily comment out test (suites) or even systems under tests in scripts/jenkins/Jenkinsfile. Then you should probably also pass the additional flags to Java, so that you get a more verbose crash dump.

> Do we really have to rewind a key set passed to an elektra plugin kdbGet/kdbSet/kdbError?

Yes, as this is normally done by `KDB` itself.

tucek:
I'm now able to reproduce the error using dockerimage buildelektra-debian-buster with `-DBUILD_DOCUMENTATION=OFF -DPLUGINS=resolver_mf_xp_x;dump;sync;base64;spec;error;list;timeofday;profile;mathcheck;tracer;hosts;network;glob -DKDB_DEFAULT_RESOLVER=resolver_mf_xp_x`

mpranj:
@tucek I noticed in the Fedora 32 images that there is a mismatch between java and javac, one is using java 1.8.0 and javac is 11. I will try to fix this in a separate PR and report back if there is any success. (This is unrelated to the segfault, and related to tools.jar not found etc.)

mpranj:
@tucek it seems my approach fixes the problems on Fedora. The changes are in https://github.com/ElektraInitiative/libelektra/pull/3767/commits/b5cff9653466816b15e2f024fda82bfd2d2c67f5

May I push this change to your branch (here), or do you object to the change?

tucek:
> @tucek it seems my approach fixes the problems on Fedora. The changes are in [b5cff96](https://github.com/ElektraInitiative/libelektra/commit/b5cff9653466816b15e2f024fda82bfd2d2c67f5)
> 
> May I push this change to your branch (here), or do you object to the change?

That sounds like great news! I will be able to test your changes in about one out two hours.

tucek:
> @tucek it seems my approach fixes the problems on Fedora. The changes are in [b5cff96](https://github.com/ElektraInitiative/libelektra/commit/b5cff9653466816b15e2f024fda82bfd2d2c67f5)
> 
> May I push this change to your branch (here), or do you object to the change?

@mpranj Thanks, please push your fix to my branch!

tucek:
Regarding the segfault, i will have to continue my investigation tomorrow morning...

tucek:
When i am isolating the org.libelektra.GOptsTest test (using the reproduction env above) i get the following error instead of the segfault (if i run all test again, then i get the segfault)

```
Testing org.libelektra.GOptsTest
Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
org.libelektra.GOptsTest.test_gopts finished with FAILURE (65ms)
java.lang.IllegalStateException: Couldn't set up spec, keys exist!
        at org.libelektra.GOptsTest.setupSpec(GOptsTest.java:50)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
        at org.junit.internal.runners.statements.RunBefores.invokeMethod(RunBefores.java:33)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
        at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:110)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:38)
        at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:62)
        at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
        at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
        at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
        at com.sun.proxy.$Proxy2.processTestClass(Unknown Source)
        at org.gradle.api.internal.tasks.testing.worker.TestWorker.processTestClass(TestWorker.java:119)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
        at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
        at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:182)
        at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:164)
        at org.gradle.internal.remote.internal.hub.MessageHub$Handler.run(MessageHub.java:414)
        at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
        at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:48)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:56)
        at java.base/java.lang.Thread.run(Thread.java:834)

org.libelektra.GOptsTest > test_gopts FAILED
    java.lang.IllegalStateException at GOptsTest.java:50

1 test completed, 1 failed
```

tucek:
@mpranj There also seems another problem being poped up: this night gradle released 7.0 and the macOS task fails with:

```
==> Downloading https://services.gradle.org/distributions/gradle-7.0-all.zip
==> Downloading from https://downloads.gradle-dn.com/distributions/gradle-7.0-all.zip
Error: gradle 6.8.3 is already installed
To upgrade to 7.0, run:
  brew upgrade gradle
Error: Process completed with exit code 1.
```
no relevant change from my side!? do you have any ideas?

tucek:
The segfault only seem to occur if -DPLUGINS=resolver_mf_xp_x;dump;sync;base64;spec;error;list;timeofday;profile;mathcheck;tracer;hosts;network;glob -DKDB_DEFAULT_RESOLVER=resolver_mf_xp_x
is used for building.

I seems to be triggered somewhere in src/bindings/jna/libelektra/src/test/java/org/libelektra/GOptsTest.java setupSpec()...

@kodebach @mpranj I have to admit, i am a little bit out of my depth here... I could neither come up with a relevant change i did to the dockerfile, nor the JNA binding for that matter...

tucek:
I can also reproduce the segfault (only using the above build config with resolver `resolver_mf_xp_x`) in dockerimage buildelektra-ubuntu-focal

```
Testing org.libelektra.GOptsTest
Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007ffa06fcc0ac, pid=7986, tid=8012
#
# JRE version: OpenJDK Runtime Environment (11.0.10+9) (build 11.0.10+9-Ubuntu-0ubuntu1.20.04)
# Java VM: OpenJDK 64-Bit Server VM (11.0.10+9-Ubuntu-0ubuntu1.20.04, mixed mode, sharing, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [libelektra-kdb.so+0xc0ac]  splitDel+0x1c
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /root/libelektra/src/bindings/jna/libelektra/hs_err_pid7986.log
#
# If you would like to submit a bug report, please visit:
#   https://bugs.launchpad.net/ubuntu/+source/openjdk-lts
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
org.libelektra.GOptsTest.test_gopts finished with SKIPPED (37ms)
```

tucek:
@mpranj Tomorrow i will revisit the other open comments and will be available at noon for discussing how to tackle the situation....

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
```diff
@@ -0,0 +1,435 @@
+package org.libelektra;
+
+import com.sun.jna.Library;
+import com.sun.jna.Native;
+import com.sun.jna.Pointer;
+import javax.annotation.Nullable;
+import org.libelektra.plugin.NativePlugin;
+
+/**
+ * The JNA interface to libelektra.
```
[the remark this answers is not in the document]
tucek: done

src/bindings/jna/README.md
```diff
@@ -9,33 +9,53 @@
...
-use the jni plugin to develop plugins for Elektra itself, whereas the jna
+use the jni plugin to develop plugins for Elektra itself, whereas the JNA
 bindings allow to use Elektra to access configuration in Java projects. The jni
-plugin is _not_ required for the jna bindings to work. But, to develop
-jni plugins, jna can be used. [Here](libelektra5j/plugin) are example
+plugin is _not_ required for the JNA bindings to work. But, to develop
+jni plugins, JNA can be used. [Here](libelektra/plugin) are example
 plugins, which need jni at runtime.
 
+## Installing the JNA binding to the local Maven repository
```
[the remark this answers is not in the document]
tucek: done

src/bindings/jna/README.md
```diff
@@ -50,33 +70,44 @@ Using the bindings via the command line works the same way as in Linux. The
...
-Then you can simply add libelektra as dependency using:
+Simply add libelektra as dependency using:
 
 ```xml
     <dependency>
       <groupId>org.libelektra</groupId>
-      <artifactId>libelektra5j</artifactId>
-      <version>0.9.4</version>
+      <artifactId>libelektra</artifactId>
+      <version>YOUR_DESIRED_VERSION_HERE</version>
```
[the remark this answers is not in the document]
tucek: done

kodebach:
> There also seems another problem being poped up: this night gradle released 7.0 and the macOS task fails with:

Aren't you using Gradle Wrapper? There should be no need to install gradle through homebrew. AFAIK the point of Gradle Wrapper is that it downloads a fixed version and uses that without the need to have a system-wide gradle install.

> When i am isolating the org.libelektra.GOptsTest test (using the reproduction env above) i get the following error instead of the segfault (if i run all test again, then i get the segfault)

This is not an error. The `GOptsTest` adds some keys to the KDB at the start and remove them again at the end. If you get a segfault in between or the process is killed for some other reason, the keys are not removed. Because we cannot know, if the keys come from `GOptsTest` or were added by the user, the test simply fails, if keys are present below `spec:/tests/java/gopts`.

The fact that the errors alternate, seems to suggest that `GOptsTest` fails because of the existing keys, but also removes them. This would be a bug. If this Exception is triggered
https://github.com/ElektraInitiative/libelektra/blob/97d945dff25a3df6cb78e290649b64b6af54be0d/src/bindings/jna/libelektra5j/src/test/java/org/libelektra/GOptsTest.java#L51

No further code should be executed by `GOptsTest`.

> The segfault only seems to occur if -DPLUGINS=resolver_mf_xp_x;dump;sync;base64;spec;error;list;timeofday;profile;mathcheck;tracer;hosts;network;glob -DKDB_DEFAULT_RESOLVER=resolver_mf_xp_x
> is used for building.
> 
> I seems to be triggered somewhere at the end of src/bindings/jna/libelektra/src/test/java/org/libelektra/GOptsTest.java setupSpec()...

This suggests that it happens during `kdbSet()` and only if the XDG resolver is used. If this indeed where the error occurs, I doubt it has anything to do with the Java binding. The Java binding could only be at fault, if `ks` is already broken (e.g. `free`'d) when `kdb.set` is called and I don't see how that could happen.

There is one more possibility. I noticed that `KeySet` has a `finalize` method that calls `ksDel`. If the garbage collector calls that during the `kdb.set` call this would cause problems. You should be able to detect that by adding a breakpoint to the `finalize` method to see if/when it is called (maybe add temporary `System.out.print` after `kdb.set` to know when it is finished).

tucek:
@kodebach thank you for insights!

> > There also seems another problem being poped up: this night gradle released 7.0 and the macOS task fails with:
> 
> Aren't you using Gradle Wrapper? There should be no need to install gradle through homebrew. AFAIK the point of Gradle Wrapper is that it downloads a fixed version and uses that without the need to have a system-wide gradle install.

Gradle wrapper would download gradle each time a fresh container is used for building, therefore we decided to include the gradle binaries with the images.

> > When i am isolating the org.libelektra.GOptsTest test (using the reproduction env above) i get the following error instead of the segfault (if i run all test again, then i get the segfault)
> 
> This is not an error. The `GOptsTest` adds some keys to the KDB at the start and remove them again at the end. If you get a segfault in between or the process is killed for some other reason, the keys are not removed. Because we cannot know, if the keys come from `GOptsTest` or were added by the user, the test simply fails, if keys are present below `spec:/tests/java/gopts`.

That makes sense!

> The fact that the errors alternate, seems to suggest that `GOptsTest` fails because of the existing keys, but also removes them. This would be a bug. If this Exception is triggered
> https://github.com/ElektraInitiative/libelektra/blob/97d945dff25a3df6cb78e290649b64b6af54be0d/src/bindings/jna/libelektra5j/src/test/java/org/libelektra/GOptsTest.java#L51
> 
> No further code should be executed by `GOptsTest`.
> 
> > The segfault only seems to occur if -DPLUGINS=resolver_mf_xp_x;dump;sync;base64;spec;error;list;timeofday;profile;mathcheck;tracer;hosts;network;glob -DKDB_DEFAULT_RESOLVER=resolver_mf_xp_x
> > is used for building.
> > I seems to be triggered somewhere at the end of src/bindings/jna/libelektra/src/test/java/org/libelektra/GOptsTest.java setupSpec()...
> 
> This suggests that it happens during `kdbSet()` and only if the XDG resolver is used. If this indeed where the error occurs, I doubt it has anything to do with the Java binding. The Java binding could only be at fault, if `ks` is already broken (e.g. `free`'d) when `kdb.set` is called and I don't see how that could happen.

I concur with this conclusion, but the thing that bugs me is that the problem does not seem to occur on `master`.
And since i have not changed any C side stuff, the only difference i could imaging would be, that the GOptsTests is not executed in the `xdg` environment on `master`, but is now executed because i made a mistake in translating the test exceptions in the jna CMakeLists.txt. - The problem is, I do not see any exclusion for the GOptTests that apply for the failing environment...
 
> There is one more possibility. I noticed that `KeySet` has a `finalize` method that calls `ksDel`. If the garbage collector calls that during the `kdb.set` call this would cause problems. You should be able to detect that by adding a breakpoint to the `finalize` method to see if/when it is called (maybe add temporary `System.out.print` after `kdb.set` to know when it is finished).

I am trying to narrow it down a little further right now.

@mpranj @kodebach If i continue to fail getting to the root of the problem, we would have to think what we should do about today's release. Disable GOptsTest and list it as known issue?

kodebach:
> Gradle wrapper would download gradle each time a fresh container is used for building, therefore we decided to include the gradle binaries with the images.

Then you should probably make sure that the system-wide install of Gradle is the same version as the one configured via Gradle Wrapper. That means using e.g. `brew install gradle@6.8.3` instead of `brew install gradle`. Otherwise you might get inconsistencies between local tests and build server tests.

>  but the thing that bugs me is that the problem does not seem to occur on `master`.

The memory bug might be present on `master` as well. But if it only occurs combination with the Java Binding, it won't be detected because the Java binding tests are not run with `valgrind` or ASAN, because you would always get false positive from the JVM.

Just because there is a memory bug does not necessarily mean there will be a segfault. That is danger of using a language with manual memory management.

> If i continue to fail getting to the root of the problem, we would have to think what we should do about today's release. Disable GOptsTest and list it as known issue?

I can't make a decision about the release, but as long as we don't actually know the cause of the segfault, we probably shouldn't make a release.

tucek:
I now can reproduce it locally with the defualt resolver set as mentioned above. Interestingly now at `Problematic frame: C  [libc.so.6+0x9d870]  cfree+0x20`...

I will now debug the JNA GOptsTests to narrow down what exact call to the native library triggers the problem down the line....

tucek:
> > Gradle wrapper would download gradle each time a fresh container is used for building, therefore we decided to include the gradle binaries with the images.
> 
> Then you should probably make sure that the system-wide install of Gradle is the same version as the one configured via Gradle Wrapper. That means using e.g. `brew install gradle@6.8.3` instead of `brew install gradle`. Otherwise you might get inconsistencies between local tests and build server tests.

It seems just removing the `brew gradle` does the trick. Gradle 6.8.3 seems to be installed prior to brew. Cmake includes JNA plugin, therefore gradle has been found. What image are these macOS builds based?

kodebach:
> It seems just removing the brew gradle does the trick. Gradle 6.8.3 seems to be installed prior to brew. 

I think you should still add it, in case the pre-installed Gradle version changes. Cirrus CI uses Anka for it's macOS VMs and their configs can be found here: https://github.com/cirruslabs/osx-images

See also: https://cirrus-ci.org/guide/macOS/

tucek:
> > It seems just removing the brew gradle does the trick. Gradle 6.8.3 seems to be installed prior to brew.
> 
> I think you should still add it, in case the pre-installed Gradle version changes.

this results in `Error: No available formula or cask with the name "gradle@6.8.3"`
I also think normally only enforcing a minimum gradle version is really necessary.

kodebach:
> this results in `Error: No available formula or cask with the name "gradle@6.8.3"`

I don't use Homebrew, possibly my quick Google search didn't give the correct way to install a fixed version.

> I also think normally only enforcing a minimum gradle version is really necessary.

A minimum version isn't enough I think. Fixing a major version might be enough, but there are definitely Gradle releases with breaking changes (i.e. not backwards compatible) from time to time.

src/bindings/swig/python/kdb.i
kodebach: > @kodebach can you copy what you wrote here to doc/GIT.md?

See #3769

tucek:
Final analysis from my side follows

Steps to reproduce
- cmake with `DPLUGINS=resolver_mf_xp_x;dump;sync;base64;spec;error;list;timeofday;profile;mathcheck;tracer;hosts;network;glob` and `DKDB_DEFAULT_RESOLVER=resolver_mf_xp_x`
- `make install && ldconfig && cd src/bindings/jna`
- run GOptsTest (or run all tests via `./gradlew test`)
- remove left over test specs after crash: `rm /usr/local/share/elektra/specification/default.ecf`

The last call to the native library is triggered by `try (final KDB kdb = KDB.open (contract, parentKey))`
(https://github.com/tucek/libelektra/blob/405e2b19c70725fb19e4a8a15638cbbe17a7cc90/src/bindings/jna/libelektra/src/test/java/org/libelektra/GOptsTest.java#L69) only if the `contract` parameter is specified.
-> https://github.com/tucek/libelektra/blob/405e2b19c70725fb19e4a8a15638cbbe17a7cc90/src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java#L56

I've included some JVM crash dumps [jvm_crash_dumps.zip](https://github.com/ElektraInitiative/libelektra/files/6290315/jvm_crash_dumps.zip)

I am afraid, at the moment, i cannot be of further help regarding this issue.

tucek:
> > this results in `Error: No available formula or cask with the name "gradle@6.8.3"`
> 
> I don't use Homebrew, possibly my quick Google search didn't give the correct way to install a fixed version.
> 
> > I also think normally only enforcing a minimum gradle version is really necessary.
> 
> A minimum version isn't enough I think. Fixing a major version might be enough, but there are definitely Gradle releases with breaking changes (i.e. not backwards compatible) from time to time.

I have opened ticket #3770 to track this issue.

mpranj:
**Thank you all** for working hard on this! ❤️ 
I'll take a look at this today before doing the release, but I can't give a guarantee that I can fix it in this time frame.

Please ping me if anything changes in the mean time. (Fyi, I'll copy this PR to a different branch so I can test independently.)

src/bindings/jna/CMakeLists.txt
```diff
@@ -23,136 +30,166 @@ if (Java_JAVAC_EXECUTABLE)
...
+					java_exclude_test (GOptsTest)
+					java_exclude_test (KDBExceptionTest)
+					java_exclude_test (KDBTest)
+					java_exclude_test (KeyNameIteratorTest)
+					java_exclude_test (KeySetTest)
+					java_exclude_test (KeyTest)
+					java_exclude_test (PluginLoaderIT)
+					java_exclude_test (ReturnTest)
+					java_exclude_test (WarningEntryTest)
+				endif ()
```
mpranj, line 71: Test exclusion did not work at all until now, so I suspect this code had no effect. If so, I'd like to get rid of that after the release.

mpranj:
I was able to get rid of the segfault by fixing the test exclusion mechanism.

The root cause was that the GOpts Tests were executed on a system where the `gopts` plugin was not compiled. `dlopen()` failed and caused a double `free()` as well in a "rare" execution path.

src/bindings/jna/CMakeLists.txt
kodebach: Possibly, this is no longer required with the new setup. But with the old Maven setup, tests that need the KDB definitely did not work under macOS.

tucek: What exactly did not work with the test exclusions?

mpranj: It is fixed now so no need to do anything, but the problem was this line:
https://github.com/tucek/libelektra/blob/dc6914eca9ed1a9929f93ef7d650288a58622eb2/src/bindings/jna/CMakeLists.txt#L14

Basically the static string `"test"` is passed instead of the variable `${test}`. I took a look at the generated `test-exclusion.gradle` and confirmed that.

src/bindings/jna/README.md
```diff
@@ -50,33 +52,63 @@ Using the bindings via the command line works the same way as in Linux. The
...
+}
+```
+
+[Here](../../../examples/external/java/read-keys-example/build.gradle) is a full example using Gradle, which should work out of the box by executing `gradle run` in the example directory `../../../examples/external/java/read-keys-example/`.
+
+### Installing the JNA binding to the local Maven repository
+
+Release versions of the Elektra JNA bindings are published to Maven Central with group id `org.libelektra`, artefact id `libelektra` and the same version as Elektra.
+
+If you want to depend on a modified binding or just want to install it to your local maven repository, change your current working directory to the JNA binding folder `/src/bindings/jna`. Either use the bundled Gradle wrapper (`./gradlew` for Unix style OS or `./gradlew.bat` for windows) or ensure a recent Gradle version is available on your system and execute:
```
mpranj, line 99: Is it recommended that we keep both the wrapper and also installed versions of gradle, or should we get rid of one of the two now?

mpranj (commented):
Great work, nice to see such a refresh to the Java bindings.

src/bindings/jna/README.md
kodebach: Developers might have different Gradle versions installed. The wrapper is the recommended way to get consistent Gradle versions AFAIK.

Within CI and Docker it might be problematic (i.e. a waste of time) to let Gradle Wrapper download the Gradle binaries every time. If this can be avoided (e.g. via caching or multi-stage Docker builds), I think it would be preferable to use Gradle Wrapper everywhere. If this is not possible, then we should probably list all the places where Gradle Version is defined in a central location.

kodebach: PS. for in the README I would only recommend using Gradle Wrapper, since that gives a consistent environment

src/bindings/jna/libelektra/src/main/java/org/libelektra/plugin/PropertiesStorage.java
markus2330: The import statements (there are also several in the following lines).

tucek: yes, all stated imports are used.

PR #3752, opened 2021-04-07
Yan LR & YAML Smith: Remove Plugins

sanssecours:


markus2330:
Thank you so much! CI seems to be happy now! Would be nice to have this in 0.9.5 (then people would not be confused which YAML plugin to use) :rocket:

mpranj:
Thank you so much! :heart:

PR #3588, opened 2020-12-10
elektrad: attempt to fix for keyname overhaul

kodebach:
## Basics

These points need to be fulfilled for every PR:

- [x] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
- [ ] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [ ] The PR is rebased with current master.

If you have any troubles fulfilling these criteria, please write
about the trouble as comment in the PR. We will help you.
But we cannot accept PRs that do not fulfill the basics.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

Reviewers will usually check the following:

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

PR #3789, opened 2021-04-16
elektrad: update

mpranj:
Based on #3588 (huge thanks to @kodebach). Updates go-elektra and some modules and fixes keyDup.
Enable building of `elektrad` again.

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -148,6 +148,7 @@ you up to date with the multi-language support provided by Elektra.
 - Update specmount error message #3666 _(@a-kraschitzer)_
 - Update specmount error message _(@a-kraschitzer)_
 - Update `elektraMemDup` to `void *` and update the documentation. _(Mihael Pranjić)_
+- There have been a few bugfixes for elektrad. _(Klemens Böswirth)_
 - <<TODO>>
 - <<TODO>>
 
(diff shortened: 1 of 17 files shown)


kodebach:
I don't have time to test this right now. But if you say it works for you (and the tests seem to agree), then I would also say we merge this and if there issues arise we fix them later. The state with this PR merged is definitely better than we the current master version.

IIRC the segfaults in my PR rarely occurred during tests. If think I could reproduce them a bit more consistently (although still not all the time), if I started the server and then executed many requests in quick succession. AFAIK `elektrad` uses a pool of `KDB` handles and somewhere in the management of this pool something went wrong.

mpranj:
> IIRC the segfaults in my PR rarely occurred during tests.

Thanks for taking a look and describing how the segfaults were happening. I'll check again and try to stress the server.

mpranj:
The current state is:
* I can not reproduce a segfault, not even with benchmarks.
* Some memory is not being freed and I can not investigate now. (Definitely happens for `Debug` and `RelWithDebInfo` builds.) Requesting the same key via REST-API fills up memory.
* Benchmarks run partially. The `delete/meta` benchmarks fail, not sure why.

The state is definitely better than master, where afaik nothing works due to the core API changes. Therefore I will merge as soon as the CI is green.

markus2330:
Amazing :sparkle:

PR #3775, opened 2021-04-13
[CM H2]  Fixing small issues

a-kraschitzer:
Closes #3685.
Closes #3665.
Closes #1995.
Closes #3681.

## Basics

These points need to be fulfilled for every PR:

- [X] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [X] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [X] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [X] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [X] The PR is rebased with current master.

If you have any troubles fulfilling these criteria, please write
about the trouble as comment in the PR. We will help you.
But we cannot accept PRs that do not fulfill the basics.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

Reviewers will usually check the following:

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

If you are already Elektra developer:

- Add the "work in progress" label if you do not want the PR to be reviewed yet.
- Add the "ready to merge" label **if the basics are fulfilled** and you also
  say that everything is ready to be merged.

PR #3790, opened 2021-04-17
Fixing small issues

mpranj:
Contains #3775 thanks to @a-kraschitzer and adds some improvements to `doc/todo/RELEASE.md`.


src/tools/kdb/specmount.cpp
```diff
@@ -3,7 +3,7 @@
  *
  * @brief source file of spec mount command
  *
- * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ * @Copyright BSD License (see LICENSE.md or https://www.libelektra.org)
```
a-kraschitzer, line 6:
```suggestion
 * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
```
Needs to be changed back to lower case. (upper casing slipped through my checks)

src/tools/kdb/gen/templates/collect.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-# Copyright (C) 2019 Elektra Initiative.
+# Copyright BSD License (see LICENSE.md or https://www.libelektra.org)
```
kodebach, line 3: @a-kraschitzer 
All the files in `src/tools/kdb/gen/templates` are templates for client side code generation. This means the copyright line must make sense outside of the Elektra repo as well. 

The copyright should be changed in all these files (and the corresponding test files) to something like:

```sh
Copyright (c) 2019-2021 Elektra Initiative (https://www.libelektra.org)
```

The year could be omitted, but I think the generated code should reflect from which version of Elektra it originates. More importantly the BSD Zero Clause License used for the template files is different from the BSD 3-Clause License used for the rest of the project (i.e. the one in LICENSE.md).

---

PS. this file should also contain the full license and the title "BSD Zero Clause License" like the `*.mustache` files.

markus2330: Can you please create a new issue or even better a PR fixing the problem? I think nobody saw it here.

> The year could be omitted

Yes, please omit the year. One place to update every year is enough.

kodebach: > One place to update every year is enough.

We could add a basic GitHub Action that only checks whether a set of predefined locations (e.g. line 3 of `LICENSE.md`) contains the current year. For example via:
```sh
sed '3!d' LICENSE.md | grep -Fq $(date +%Y)
```
It would fail for the first PR every year, so we can't forget to update the year.

markus2330: > We could add a basic

There are dozens of problems that happen frequently and have severe impact but could be easily detected by CI. So we should not waste time doing things in CI that happen rarely and have basically no impact and could be avoided in the first place by simply not introducing the problem (not adding the year).

> GitHub Action

Exactly the same could be done in Jenkins, which is fully under our control.

kodebach: > simply not introducing the problem (not adding the year).

Should've been more clear. I'm fine with leaving the year out of the template files. But it has to be there in LICENSE.md and it is quite easy to forget about it.

> Exactly the same could be done in Jenkins,

Yes, but IMO creating a GitHub Action is easier than a Jenkins job.

> which is fully under our control.

True, but I don't think this job so important that we need to be fully in control of it ;)

markus2330: This is what you mean :wink:

For this particular year we already have a reminder in doc/todo/RELEASE.md

We could add the check what you suggested in the release pipeline but imho it is not worth failing a release for a wrong date.

Issue #3772, opened 2021-04-13
KeySet::lookup(String) returns Key instead of `null` on unsuccessful search [BUG JNA binding]


tucek:
## Steps to Reproduce the Problem

```java
KeySet set = KeySet.create();
if (set.lookup("user:/unsetkey") != null) {
  throw new AssertionError();
}
```

## Expected Result
`null` should be returned.

## Actual Result

`AssertionError` is thrown

## System Information

- Elektra Version: 0.9.5
- Operating System: Ubuntu focal on wsl2

## Proposed solution
`KeySet::lookup(...)` should return `Optional<Key>` and would return `Optional.empty()` if `Elektra.INSTANCE.ksLookup(ByName)?` returns null.

## Workaround for 0.9.5
```java
KeySet set = KeySet.create();
// check whether Key pointer is actually null
if (set.lookup("user:/unsetkey").get() != null) {
  throw new AssertionError();
}
```

mpranj:
Thank you for reporting and working on this.

I am not a user of the Java bindings, so take this with a grain of salt.
From what I see this is not a bug. Since the Key class wraps JNA pointers you can not compare the Key object to `null`. You have to call `get()` (which is your workaround) or even simply use the existing `isNull()` function.

```Java
if (!set.lookup("user:/unsetkey").isNull()) {
  throw new AssertionError();
}
```

That being said, improvements to the Java bindings/API are welcome and a consistent use of optionals could make sense. :+1:

tucek:
At least the current behavior contradicts the documentation of the function: `@return Key if search successful, null otherwise` - Therefore i would consider this a bug.

Apart from that i would consider returning a `Key` object which has to be checked for referencing a `null` pointer (and therefore being completely useless), a rather crude approach for a Java API.

I will update the Java API to use `Optional`s accordingly.

mpranj:
> At least the current behavior contradicts the documentation of the function: @return Key if search successful, null otherwise - Therefore i would consider this a bug.

You're right, I overlooked the docu. According to the docu a fix would be to just return `null` instead of `new Key(null)`. That is irrelevant though, if you want to change the API.

Issue #3710, opened 2021-03-17
Invalid key name log message when executing some JNA tests


tucek:
## Steps to Reproduce the Problem

I've built Elektra with `ENABLE_LOGGER` = `ON`. When executing the JNA binding tests, multiple `Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)` messages are being logged by Elektra.

```sh
cd src/bindings/jna/libelektra5j/
cp pom.xml.in pom.xml
mvn test
```

## Expected Result

No log messages issued by Elektra regarding invalid names.

## Actual Result

```
#> ...
#> -------------------------------------------------------
#>  T E S T S
#> -------------------------------------------------------
#> Running org.libelektra.exception.model.WarningEntryTest
#> Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.494 sec
#> Running org.libelektra.ExceptionMapperTest
#> Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.111 sec
#> Running org.libelektra.GOptsTest
#> Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
#> Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
#> Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
#> Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.091 sec
#> Running org.libelektra.KDBExceptionTest
#> Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 sec
#> Running org.libelektra.KDBTest
#> Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
#> Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
#> Invalid name:  (in keyVNew at src/libs/elektra/key.c:238)
#> Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.002 sec
#> Running org.libelektra.KeySetTest
#> Tests run: 17, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 sec
#> Running org.libelektra.KeyTest
#> Tests run: 30, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 sec
#> 
#> Results :
#> 
#> Tests run: 54, Failures: 0, Errors: 0, Skipped: 0
[...]

PR #3825, opened 2021-05-04
JNI Test Fix

tucek:
## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


markus2330:
@tucek can you please update the README.md of jni with latest introduction of how you installed everything? Most parts there are very outdated. The Debian 9+8 instructions can be safely removed.

When I try to compile this branch I get 
```
INFOJNI test activated.
```

Probably some wrongly-used cmake debugging message you forgot to remove?
It was confusing for me as we usually only output something if tests were deactivated.

On Debian 10 with:

- openjdk-11-jre and openjdk-11-jdk
- gradle-6.8.3 manually installed

I can also reproduce the segfault. Running with valgrind, the segfault does not happen: the test can be executed successfully.

@tucek can you please try different Java versions (e.g. the one of Oracle) on different OSs to see which systems are affected and for which it works (if any)?

If we have a problem in our code it may be related to how we execute JNI_CreateJavaVM¹ or in line 120 or 309 (NewObject), as valgrind is pointing there.
¹ https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/invocation.html they don't use the options on the stack.

@kodebach can you take a look about that, please? It would be a great help. Ideally even fixing the TODO in line 201.

Would be great if we can give the students a smooth experience when using the JNI plugin!


Note for myself, I started valgrind with: `LD_LIBRARY_PATH=/home/markus/Projekte/Elektra/build/lib DB_TEST_BIN_DIR=/home/markus/Projekte/Elektra/build/bin valgrind --tool=memcheck --error-limit=no --track-origins=yes -v --leak-check=full --suppressions=/home/markus/Projekte/Elektra/current/tests/valgrind.suppression --show-reachable=yes /home/markus/Projekte/Elektra/build/bin/testmod_jni "/home/markus/Projekte/Elektra/current/src/plugins/jni"`

kodebach:
I do not have time to fully diagnose this problem, but here are my findings from a quick investigation:

1. Just executing `testmod_jni` with a debugger shows that the double free happens in `keyDel()`.
2. GDB doesn't show a backtrace between https://github.com/ElektraInitiative/libelektra/blob/d9a8478478b0009971a070d2990af0b310abfd7e/src/plugins/jni/testmod_jni.c#L31 and `keyDel()` here https://github.com/ElektraInitiative/libelektra/blob/729a6725d25ce0649500bb5f45f2cad6eec36171/src/libs/elektra/key.c#L526 
This is because `keyDel()` is called inside the `PLUGIN_OPEN` macro.
3. Since we have a double free, there has to be an earlier call to `keyDel()`. This lead me to believe that `keyDel()` is called once via JNI from Java and then again by the `PLUGIN_OPEN` macro.
4. To confirm this, I created a "Remote JVM Debug" configuration in IntelliJ (default settings are fine). This gives you a set of arguments with which the Remote JVM needs to be started. In my case this was
    ```
    -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
    ```
    To use these arguments you need to modify `testmod_jni`. 
    ```diff
    diff --git a/src/plugins/jni/jni.c b/src/plugins/jni/jni.c
    index 1577242b3..1c20933aa 100644
    --- a/src/plugins/jni/jni.c
    +++ b/src/plugins/jni/jni.c
    @@ -206,11 +206,12 @@ int elektraJniOpen (Plugin * handle, Key * errorKey)
            */
    
            JavaVMInitArgs vmArgs;
    -       JavaVMOption options[2];
    +       JavaVMOption options[3];
            options[0].optionString = classpath;
            options[1].optionString = option;
    +       options[2].optionString = "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005";
            vmArgs.version = JNI_VERSION_1_8;
    -       vmArgs.nOptions = 2;
    +       vmArgs.nOptions = 3;
            vmArgs.options = options;
            vmArgs.ignoreUnrecognized = ign;
    ````
    When you start `testmod_jni` now, it will wait for a Java Debugger to attach during the call to `JNI_CreateJavaVM`.
5. I started `testmod_jni` again with a debugger. In IntelliJ I added a breakpoint in `Key.release()` and attached the Java Debugger `testmod_jni` was waiting. Sure enough, the breakpoint triggered and the Java Code called `keyDel()`. I continued and then I got the double free exception in GDB. [*]

**Summary**
The `errorKey` passed to `elektraJniOpen` is improperly freed by the Java code (via `Key.release()`). One solution might be to call `keyIncRef()`/`keyDecRef()` before/after passing the key to the JVM. This should probably happen somewhere in `call2Arg`. A similar issue will likely exist for `KeySet`s. 

@markus2330 For the future, we should probably reconsider #3631 (separate ref counters for KeySet -> Key and Other -> Key, references) and maybe think about even more changes to the reference counting. For example, maybe `kdbOpen`, `kdbGet` etc. should increment the ref count before passing to plugins to ensure the plugin doesn't accidentally delete the parent key.

---

[*] When I run `testmod_jni` with GDB I also get a segfault from somewhere within `libjvm.so` during the call to `JNI_CreateJavaVM` but _before_ the JVM waits for the Java Debugger. There was no usable backtrace so I can't tell why it happens, but you can just continue and everything works.

kodebach:
PS. @tucek If you need to debug this further and would prefer a graphical interface for the debugger over GDB, you can use this [VSCode config](https://code.visualstudio.com/docs/editor/debugging#_add-a-new-configuration) (assumes your build folder is called `build`):

```json
{
  "name": "testmod_jni",
  "type": "cppdbg",
  "request": "launch",
  "program": "${workspaceFolder}/build/bin/testmod_jni",
  "args": [
    "${workspaceFolder}/src/plugins/jni"
  ],
  "stopAtEntry": false,
  "cwd": "${workspaceFolder}/build/bin",
  "environment": [
    {
      "name": "LD_LIBRARY_PATH",
      "value": "${workspaceFolder}/build/lib"
    }
  ],
  "externalConsole": false,
  "MIMode": "gdb",
  "setupCommands": [
    {
      "description": "Enable pretty-printing for gdb",
      "text": "-enable-pretty-printing",
      "ignoreFailures": true
    }
  ]
}
```

tucek:
> When I try to compile this branch I get
> 
> ```
> INFOJNI test activated.
> ```
> 
> Probably some wrongly-used cmake debugging message you forgot to remove?
> It was confusing for me as we usually only output something if tests were deactivated.

I intentionally had not yet removed this debug message. (Since this is still a draft PR...)

> On Debian 10 with:
> 
> * openjdk-11-jre and openjdk-11-jdk
> * gradle-6.8.3 manually installed
> 
> I can also reproduce the segfault. Running with valgrind, the segfault does not happen: the test can be executed successfully.

I've tested Ubuntu 20.04 and Debian 10 with openjdk-11-jdk and get identical results. The only difference I found was, that i had to set JAVA_HOME on Debian 10 for the FindJNI cmake module to actually find the JNI header file. Are you sure plugin jni was successfully included by cmake and `ctest -V -R testmod_jni` did finish successfully?

> @tucek can you please try different Java versions (e.g. the one of Oracle) on different OSs to see which systems are affected and for which it works (if any)?

Since i did not get different results between the two configurations above (in contrary to your report), I do not really see the indication for doing a (quite time consuming) extensive test with different java distribution / os combinations. Nonetheless i've tested the following additional combinations:

- Debian 10 / AdoptOpenJDK 11.0.11.j9
- Debian 10 / Microsoft 11.0.11.9.1
- Debian 10 / Oracle GraalVM 21.1.0.r16 (jdk 16.0.1)

All yielded the same result.

kodebach:
A temporary workaround would be to remove the `finalize()` methods from the JNA binding. This would create memory leaks (Keys created in Java would not be deleted), but it should avoid the double-free problem and the segfaults.

markus2330:
@tucek 
>  When did it last work?

Last CM it was working. A git-bisect would be needed to find the last working commit. But maybe the code never worked with recent Javas.

>  What should participants of CM2021s, committed to writing a JNI plugin, should hand in tomorrow?

We will discuss this tomorrow.

@kodebach 
> A temporary workaround would be to remove the finalize() methods from the JNA binding.

As you already debugged and found a plausible reason, please let us try this first. It is more important to have a tested&reliable solution than to have something very quick.

tucek:
@kodebach Thank you for the thorough analysis! I was also wondering why release method is being called directly. but when i am understanding correctly:

1. jni plugin instantiated the Key within the JVM via Constructor
2. Constructor explicitly increasing the reference count from 0 to 1

Before removing release in jni plugin, while finalize never being called:
3. jni plugin calls Key::release (decreasing the reference count from1 to 0 and calling keyDel unallocating the key)
4. Key is double freed by https://github.com/ElektraInitiative/libelektra/blob/d9a8478478b0009971a070d2990af0b310abfd7e/tests/cframework/tests_plugin.h#L21

After removing release in jni plugin, while finalize never being called:
3. Is the reference counter is not being decreased before https://github.com/ElektraInitiative/libelektra/blob/d9a8478478b0009971a070d2990af0b310abfd7e/tests/cframework/tests_plugin.h#L21
it is not freed at all

While KeySet is being freed correctly by https://github.com/ElektraInitiative/libelektra/blob/d9a8478478b0009971a070d2990af0b310abfd7e/src/libs/elektra/plugin.c#L313
 
Now, if we fix the Key finalization, we would get the same problem again.

Ad key finalization:

- Introducing Closable for Key and KeySet would also mean having to check whether the underlying pointer has already been "closed" for each method of the Key / KeySet Java. (e.g. KeyClosedException, KeySetClosedException)
- A more transparent method would be declaring the internal pointer reference of Key / KeySet as volatile and usinf a Cleaner. This would still introduce the need for checking the internal pointer for being null on each class's method and would raise the minimum required Java version to 9.

kodebach:
Yes, I missed that removing the `release()` calls would leave the reference count at 1. But the solution here is simple. Instead of calling the Java `Key.release` method to decrease the reference count, we can just call `keyDecRef (errorKey)` directly (at exactly the same  point in code). Then the subsequent `keyDel()` call would correctly free the `Key`'s memory.

`KeySet` doesn't have a reference counter, so just removing the `KeySet.release` call should solve everything. Neither the Java Code nor the JNI plugin would call `ksDel()` and the `KeySet *` would survive and be freed correctly by the caller of the plugin (probably `libelektra-kdb`, but not actually the line you posted -- that one is just in an error case).

tucek:
> Yes, I missed that removing the `release()` calls would leave the reference count at 1. But the solution here is simple. Instead of calling the Java `Key.release` method to decrease the reference count, we can just call `keyDecRef (errorKey)` directly (at exactly the same point in code). Then the subsequent `keyDel()` call would correctly free the `Key`'s memory.
> 
> `KeySet` doesn't have a reference counter, so just removing the `KeySet.release` call should solve everything. Neither the Java Code nor the JNI plugin would call `ksDel()` and the `KeySet *` would survive and be freed correctly by the caller of the plugin (probably `libelektra-kdb`, but not actually the line you posted -- that one is just in an error case).

Very good! The last peace missing would be to provide an alternative constructor signature for suppressing the finalization of the Key / KeySet. Then we could leave the finalization code as is for now and i would create a ticket to deal with it in the future.

@markus2330 If that solution is fine with you, i would update my branch.

markus2330:
> If this solution is fine with you

First and foremost the people using JNI and the CI needs to be happy with the solution. So yes, please create a PR and let us see if the CI is happy. Obviously it would be better to not create further issues and have a solution that fixes the problem without leftovers. Maybe you or @kodebach find a clean&nice way? But if this is not possible now, we'll have to live with one more open issue.

tucek:
> > If this solution is fine with you
> 
> First and foremost the people using JNI and the CI needs to be happy with the solution. So yes, please create a PR and let us see if the CI is happy. Obviously it would be better to not create further issues and have a solution that fixes the problem without leftovers. Maybe you or @kodebach find a clean&nice way? But if this is not possible now, we'll have to live with one more open issue.

I think the proposed solution fixed the problem in a relatively clean way. The additional issue i would want to create is not actually directly related to this issue, but to the fact that the finalize mechanism has been deprecated several Java version ago and we might want to be prepared when it finally gets removed.

markus2330:
> The additional issue i would want to create is not actually directly related to this issue, but to the fact that the finalize mechanism has been deprecated several Java versions ago and we might want to be prepared before it finally gets removed.

I agree this is an additional issue but still it would be nice to have this fixed within the release, too. But you are right: let us first fix the important&urgent problems like segfaults, and then let us reevaluate.

tucek:
> > The additional issue i would want to create is not actually directly related to this issue, but to the fact that the finalize mechanism has been deprecated several Java versions ago and we might want to be prepared before it finally gets removed.
> 
> I agree this is an additional issue but still it would be nice to have this fixed within the release, too. But you are right: let us first fix the important&urgent problems like segfaults, and then let us reevaluate.

@markus2330 means we never really supported Java 8 and therefore we can require Java 9+ and update the clean up procedures right away with this PR.

tucek:
PR will also fix #3772

tucek:
I have one question (also related to #3710) i cannot answer myself:
There are several locations in the JNA binding where a key with an empty (non-null) name is being created. This seems to be exclusively done in order to provide a key reference for Elektra function calls to put error information into. But:
Creating a key with empty or null name results in no key object being allocated. How is this supposed to work?

Null names seem to result in no key being allocated here: https://github.com/ElektraInitiative/libelektra/blob/4da0d2a9e5ebb5e167265f7141d7a0ac89ed46db/src/libs/elektra/key.c#L152
while the documentation states NULL names allocate a key object here: https://github.com/ElektraInitiative/libelektra/blob/4da0d2a9e5ebb5e167265f7141d7a0ac89ed46db/src/libs/elektra/key.c#L140

Empty names seem result in a key being allocated and freed here: https://github.com/ElektraInitiative/libelektra/blob/4da0d2a9e5ebb5e167265f7141d7a0ac89ed46db/src/libs/elektra/key.c#L239 (this also explains the log messages form #3710) because of https://github.com/ElektraInitiative/libelektra/blob/4da0d2a9e5ebb5e167265f7141d7a0ac89ed46db/src/libs/elektra/keyname.c#L497

@markus2330 / @mpranj / @kodebach can anyone of you clarify this for me please?

kodebach:
>  How is this supposed to work?

This may be an artifact from before #3555. If you just need a key for e.g. errors but don't care about the name, you should now use `keyNew ("/", KEY_END)` (or the equivalent of whatever binding you are using), because all instances of `Key` must now contain a valid key name. I probably didn't catch these instances of keys with empty/null names, because the tests weren't executed.

tucek:
PR will also fix #3710

.github/workflows/macOS_gcc10.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           swig \
           yajl \
           zeromq
-        brew install --cask adoptopenjdk9
+        brew install openjdk@11
```
mpranj, line 57: To save you some time: I tried **many** combinations and only adoptopenjdk worked due to some upstream linking issues.

tucek: Thank you! The problem is, this implies we are stuck on jdk 1.8 when using this image... Can't we upgrade to a more recent macOS version?

mpranj: Last time I checked this should be using AdoptOpenJDK packaged Java 16 on master, which is much more recent than 1.8. Where do you get this number from?

Also the macOS version is the latest supported image for GitHub Actions.

kodebach:
```sh
brew install --cask adoptopenjdk9
```

Will use the Java 9 Version of AdoptOpenJDK, but `brew install --cask adoptopenjdk` will use the latest and `brew install --cask adoptopenjdk16` is fixed to Java 16. See [here](https://github.com/AdoptOpenJDK/homebrew-openjdk)

The Github Actions image also comes with AdoptOpenJDK versions for Java 8, 11, 12, 13 and 14 pre-installed. See [here](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#java)

If we build with Java 11 (currently latest LTS) then we don't even need to install via homebrew. Setting `JAVA_HOME="$JAVA_HOME_11_X64"` should be enough.

mpranj: > If we build with Java 11 (currently latest LTS) then we don't even need to install via homebrew. Setting JAVA_HOME="$JAVA_HOME_11_X64" should be enough.

It does not work.

I tried to save you all some time as I tested this already: the other shipped macOS JDKs are linked improperly (upstream issue) and the only currently working one is the adoptopenjdk brew cask.

tucek:
My changes should be complete, except for removing debug output once the following problem is solved:
(works on my machine - WSL2 Ubuntu 20.04 LTS with openjdk-11-jdk)

- I cannot get macOS_gcc10 to install a higher jdk version than 1.8, because the appropriate brew packages are not found (https://github.com/AdoptOpenJDK/homebrew-openjdk)
  - any ideas anyone?
  - maybe the used macos image is just too old?
  - see comment above: https://github.com/ElektraInitiative/libelektra/pull/3825#discussion_r632538340
  - another option might be to not build JNA/JNI on macOS_gcc10 (which currently is the case in this branch because jdk 1.9 requirements are not met and therefor cmake excludes JNI/JNA)

tucek:

.github/workflows/macOS_gcc10.yml
tucek: > I tried to save you all some time as I tested this already: the other shipped macOS JDKs are linked improperly (upstream issue) and the only currently working one is the adoptopenjdk brew cask.

which seems to install jdk 1.8... at tleast that's the version used and detected by the cmake scripts

mpranj:
> I cannot get macOS_gcc10 to install a higher jdk version than 1.8, because the appropriate brew packages are not found 

I will take a look if you don't mind. It worked with Java 16 before so I'm really curious.

mpranj:
@tucek can I push changes to your branch if my changes look promising?

tucek:
@mpranj Thank you!

CI still seems to find some mem leaks... are these false positives because of the jvm? (@kodebach mentioned something above: https://github.com/ElektraInitiative/libelektra/pull/3825#issuecomment-838423095)

.github/workflows/macOS_gcc10.yml
```diff
@@ -64,6 +64,9 @@ jobs:
         brew postinstall dbus
         brew services restart dbus
         cmake -E make_directory ${{runner.workspace}}/build
+        export JAVA_HOME=$(/usr/libexec/java_home -v 16)
```
markus2330, line 67: Please check if this is really required and add remarks to jni/README.md

doc/news/_preparation_next_release.md
```diff
@@ -49,7 +49,15 @@ The following section lists news about the [plugins](https://www.libelektra.org/
 - <<TODO>>
 - <<TODO>>
 
-### <<Plugin2>>
+### JNI
```
markus2330, line 52: Please put this as highlight, give more text about what changed, put links to jni/jna docu/tutorials, explain what was tested and what works now.

doc/news/_preparation_next_release.md
```diff
@@ -139,7 +147,24 @@ you up to date with the multi-language support provided by Elektra.
...
+- Migrated native resource clean-up from `finalize()` to `Cleaner`
+- Extracted exceptions from Key class introducing the following changes
+  - Removed unused `KeyTypeConversionException`
+  - Introduced KeyCreateFailedException, KeyReleasedException
+  - Renamed `KeyInvalidNameException` to `KeySetNameFailedException`
+  - Renamed `KeyTypeMismatchException` to `KeyBinaryTypeNotSupportedException`
+- Introduced `KeySetReleasedException`
+- `KeySet::lookup*` now returns `Optional<Key>`
+- `Key::getMeta` now returns `Optional<Key>`
+- Removed `Key::isNull`
```
markus2330, line 164: Write what needs to be done instead now, give code examples and so on.

scripts/cmake/ElektraClasspath.cmake
```diff
@@ -1,8 +0,0 @@
-set (SOURCE_FILE "${SOURCE_PATH}/testmod_jni.h.in")
```
markus2330, line 1: Thank you for cleaning up!

doc/news/_preparation_next_release.md
```diff
@@ -139,7 +147,24 @@ you up to date with the multi-language support provided by Elektra.
 
 - Remove `-Wno-shift-overflow warnings` option from SWIG bindings compile flags. _(Mihael Pranjić)_
 
-### <<Binding2>>
+### JNA
+
+- Gradle wrapper upgraded to 7.0
```
markus2330, line 152: 6.8.3 is still supported?

markus2330 (commented):
Start review.

doc/news/_preparation_next_release.md
tucek: yes

.github/workflows/macOS_gcc10.yml
kodebach: AFAIK this should normally be done during the installation of the JDK. How exactly it happens depends on your OS and distro. I think, Homebrew normally doesn't set such environment variables automatically and instead prints a message for the user that this might be needed. But when using e.g. Debian the `JAVA_HOME` variable would probably be set by `apt` during the installation process.

If we add anything to the README it should only be along the lines of "You need a working JDK installation for Java version X or higher" (not sure what we require).

tucek: > If we add anything to the README it should only be along the lines of "You need a working JDK installation for Java version X or higher" (not sure what we require).

I agree. (minimum required Java version is 9)

tucek:
@markus2330 ad review comments: I will try to get it done over the weekend, but am currently needing to focus on something else....

.github/workflows/macOS_gcc10.yml
markus2330: The problem is that we do not only require "a working JDK installation" but we require more. Compiling JNI plugins is far away from what typically is done with a JDK.

> But when using e.g. Debian the JAVA_HOME variable would probably be set by apt during the installation process.

No, it is not: in Debian `java(c)` is in PATH but this is not enough for our needs. Usually JAVA_HOME is not needed in Debian but in our case it might be. If it is needed, we should mention it.

markus2330 (commented):
A few more comments.

.github/workflows/macOS_gcc10.yml
kodebach: `JAVA_HOME` normally shouldn't be needed. Gradle and all the standard Java tools will detect a Java installation, if it is located in one of the standard places. However, if you have multiple JDKs installed or used a non-standard installer (e.g. Homebrew) setting `JAVA_HOME` will tell the tools where to look for the JDK. This is entirely a standard Java development thing.

The Github Actions macOS image contains multiple JDKs and we installed via Homebrew (which results in a non-standard path) so setting `JAVA_HOME` is required to ensure the correct JDK is used.

If we want to be on the really safe side, we could state in the README: "A working JDK installation for Java 9 or higher is required. You may need to set `JAVA_HOME` to point to the correct JDK install".

mpranj: > The Github Actions macOS image contains multiple JDKs

Yes I can confirm that this was the reason. Otherwise a different/wrong JDK is used. This is specific to Github Actions and thus is only in this job's config and not in CMake.

For a normal installation (on my machine) where only one JDK is present, this is not needed at all.

doc/news/_preparation_next_release.md
tucek: updated

doc/news/_preparation_next_release.md
tucek: release notes updated

I do not think examples regarding usage or Java optionals are appropriate. Are there any specific examples you would want to see?

src/plugins/jni/jni.c
```diff
@@ -271,31 +267,24 @@ int elektraJniOpen (Plugin * handle, Key * errorKey)
...
 		return -1;
 	}
 
-	data->midKeySetRelease = (*data->env)->GetMethodID (data->env, data->clsKeySet, "release", "()V");
-	if (data->midKeySetRelease == 0)
+	data->midKeyDecRef = (*data->env)->GetMethodID (data->env, data->clsKey, "decRef", "()V");
+	if (data->midKeyDecRef == 0)
 	{
-		ELEKTRA_SET_RESOURCE_ERROR (errorKey, "Cannot find release of KeySet");
+		ELEKTRA_SET_RESOURCE_ERROR (errorKey, "Cannot find method of Key::decRef()");
```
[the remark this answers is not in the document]
tucek: updated error messages

src/plugins/jni/README.md
```diff
@@ -36,47 +36,34 @@ The Java plugin itself needs to have the following methods:
...
 
-### Java Prerequisites on macOS
+### Java prerequisites on macOS
 
 macOS includes an old apple specific version of java, based on 1.6.
-However, for the jni plugin version 1.8 of Java is required, so either the openjdk or the oracle jdk has to be installed.
+However, for the jni plugin version 1.9 of Java is required, so either the openjdk or the oracle jdk has to be installed.
 
 Please install oracle's jdk8 via their provided installer.
 After that, you have to set the JAVA_HOME environment variable to the folder where the jdk is installed, usually like
```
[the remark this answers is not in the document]
tucek: updated

src/plugins/jni/README.md
```diff
@@ -36,47 +36,34 @@ The Java plugin itself needs to have the following methods:
...
 
-When using non-standard paths, you have to set JAVA_HOME before invoking cmake.
-(For example when you unpack Oracle Java to `/usr/local` or `/opt`.)
-For example:
+If you get an error like the following ensure `JAVA_HOME` is set correctly:
 
-```sh
-JAVA_HOME=/usr/local/jdk-9.0.1
 ```
+CMake Error at src/bindings/jna/CMakeLists.txt:47 (if):
```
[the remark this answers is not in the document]
tucek: I've updated docs and the message output by cmake

tucek:
When this PR is merged CM H3 java plugin writers should be able to commence...

doc/news/_preparation_next_release.md
markus2330: Not "I want to see" but think of users of your binding. If someone already uses `lookup` or `getMeta` he/she wants to know how to adapt the existing code. The easiest way to see that is with an example.

tucek: I see your point, even though the API change is trivial. Further, there are many different ways, how to process the resulting Optional, depending on the callers structure and the context, the method result is to be used...
I've added an example to the changelog.

markus2330: It is essential that it is always clear that every change somehow improves the user experience. This should be clear everywhere: in the code, in the release notes and also in the bachelor thesis :heart:

tucek:
@markus2330
are we still providing gentoo packages?
``````

**Comment by tucek** [comment]
Can someone please help me interpret https://build.libelektra.org/job/libelektra/job/PR-3825/22/display/redirect ?

**Comment by kodebach** [comment]
I'm not 100% certain, but maybe `call2Arg` doesn't correctly cleanup the JVM data. Possible we should be doing this at the end:

```c
(*data->env)->DeleteLocalRef (data->env, jks);
checkException (data, method, errorKey);
(*data->env)->DeleteLocalRef (data->env, jkey);
checkException (data, method, errorKey);
```

I don't have any experience with JNI and most Google results assume you are calling native code from Java not the other way around.

If you cannot fix these issues, just disable tests if ASAN is used and create a new issue to remind us that there might be legitimate memory leaks here. I think it is more important right now to have something that at least somewhat works than to have a perfect solution.

kodebach:
I just compared the build logs from before and after the addition of `DeleteLocalRef`. It seems the change only resulted in more memory leaks being detected. I think it is safe to say that at least some of the memory leaks detected are false positives, so we should just disable the tests if ASAN is active (see other comment) and not waste more time on this right now.

src/plugins/jni/CMakeLists.txt
```diff
@@ -40,29 +40,17 @@ add_plugin (
...
 	check_binding_was_added ("jna" BINDING_WAS_ADDED)
-	find_program (MAVEN_EXECUTABLE mvn)
 
-	if (ENABLE_BROKEN_TESTS)
-		if (BUILD_TESTING
-		    AND BINDING_WAS_ADDED
-		    AND MAVEN_EXECUTABLE)
-			add_plugintest (jni MEMLEAK)
-			include_directories (${CMAKE_CURRENT_BINARY_DIR})
+	if (BUILD_TESTING AND BINDING_WAS_ADDED)
```
kodebach, line 44:
```suggestion
	if (BUILD_TESTING AND NOT ENABLE_ASAN AND BINDING_WAS_ADDED)
```

tucek:
There still seems to be the occasional segfault... https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3825/26/pipeline

kodebach:
@robaerd The `buildPackage` stages all start with a clean Debian/Fedora Docker image and don't install any packages via `apt-get`/`dnf`. Is it possible that we are missing some dependencies in the `deb`/`rpm` file?

I don't think this is likely, since both the binding and the plugin shouldn't need any dependencies beyond a JVM. I don't really have another explanation for what is happening...
I also tried a local `make package` (although just the `.tar.bz2` not a `deb` or `rpm`, because I'm on Manjaro) and couldn't reproduce the issue either.

PS. The Fedora 34 package build now fails, because it cannot find `diff`:
https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3825/27/pipeline/2120

tucek:
> PS. The Fedora 34 package build now fails, because it cannot find `diff`:
> https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3825/27/pipeline/2120

Why is this happening... nothing change there?!

kodebach:
> > PS. The Fedora 34 package build now fails, because it cannot find `diff`:
> > https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3825/27/pipeline/2120
> 
> Why is this happening... nothing change there?!

The base docker image may have changed upstream.

tucek:
And there it is again... a sporadic (double?) free segfault for testjna_gradle (!), not testmod_jni, in fedora...

kodebach:
> And there it is again... a sporadic (double?) free segfault for testjna_gradle (!), not testmod_jni, in fedora: https://github.com/ElektraInitiative/libelektra/pull/3825/checks?check_run_id=2638629600

Okay that is definitely a problem... A quick fix would be to disable the `Cleaner` stuff entirely and just never cleanup anything from Java. This will of course result in lots of memory leaks, but shouldn't cause any segfaults.

I'm kind of running out of ideas here, but the [JDK docs](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/ref/Cleaner.html) state this:
> Cleaning actions should be prepared to be invoked concurrently with other cleaning actions. Typically the cleaning actions should be very quick to execute and not block. If the cleaning action blocks, it may delay processing other cleaning actions registered to the same cleaner. All cleaning actions registered to a cleaner should be mutually compatible.

Possibly, the GC calls `keyDel` on a `Key` that is part of a `KeySet` and simultaneously calls `ksDel` on said `KeySet`. Since `ksDel` calls `keyDecRef` and then `keyDel` on all the `Key`s in the `KeySet`, there is technically a very small chance that all these calls happen in such a way that there are two simultaneous `keyDel` calls to the same `Key`  with zero references. Both of these would try to free the `Key`.

However, this is so incredibly unlikely that I don't think this is the case.

I don't really know what else to do here... Maybe we can find another OSS project that wraps native memory in a Java object and see how they deal with the memory management. For now I think we just have to give up and accepts that Java code leaks all the memory it allocates, i.e. remove `Cleaner` and just not ever call `keyDel`/`ksDel`.

Long-term we should really redesign the whole reference counting mechanism. It clearly wasn't designed for the way we are using it. Even the name `ksReference` indicates that it was only designed to count references from `KeySet`s and **not** as a complete reference counting memory management solution like e.g. C++'s `shared_ptr`.

tucek:
I will disable the Cleaner, reenable the test for the packages...
If that fixes the current problem, i will update the java docs to manually release Keys and KeySets whenever feasable...

tucek:
> > And there it is again... a sporadic (double?) free segfault for testjna_gradle (!), not testmod_jni, in fedora: https://github.com/ElektraInitiative/libelektra/pull/3825/checks?check_run_id=2638629600
> 
> Okay that is definitely a problem... A quick fix would be to disable the `Cleaner` stuff entirely and just never cleanup anything from Java. This will of course result in lots of memory leaks, but shouldn't cause any segfaults.

Manual release should still be ok for java Key object created via Key::create.

> I'm kind of running out of ideas here, but the [JDK docs](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/ref/Cleaner.html) state this:
> 
> > Cleaning actions should be prepared to be invoked concurrently with other cleaning actions. Typically the cleaning actions should be very quick to execute and not block. If the cleaning action blocks, it may delay processing other cleaning actions registered to the same cleaner. All cleaning actions registered to a cleaner should be mutually compatible.
> 
> Possibly, the GC calls `keyDel` on a `Key` that is part of a `KeySet` and simultaneously calls `ksDel` on said `KeySet`. Since `ksDel` calls `keyDecRef` and then `keyDel` on all the `Key`s in the `KeySet`, there is technically a very small chance that all these calls happen in such a way that there are two simultaneous `keyDel` calls to the same `Key` with zero references. Both of these would try to free the `Key`.
> 
> However, this is so incredibly unlikely that I don't think this is the case.

Very unlikely, i agree. especially if it seem 100% reproducible.

> I don't really know what else to do here... Maybe we can find another OSS project that wraps native memory in a Java object and see how they deal with the memory management.

That's a very good idea!

> For now I think we just have to give up and accepts that Java code leaks all the memory it allocates, i.e. remove `Cleaner` and just not ever call `keyDel`/`ksDel`.

I'll try this next as stated in https://github.com/ElektraInitiative/libelektra/pull/3825#issuecomment-845910953 above...

tucek:
interestingly the fedora missing diff problem seems to be gone: https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3825/29/pipeline/2120

tucek:
https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-3825/29/pipeline/2120 failed as expected. i will now commit the disabled clean-up

tucek:
most interestingly, removing the gc time clean up did prevent the seg fault... ;(

tucek:
the ci system seems to be a little bit erratic: now some dbus test fails on mac os... and the dreebsd 12 check has been terminated...

tucek:
debian-bullseye-full failed in testruby_tools_plugindatabase.rb...

tucek:
@mpranj @markus2330 ist there currently some problem with the ci environment or do you have another explanation for this more or less random errors?

kodebach:
> the ci system seems to be a little bit erratic: now some dbus test fails on mac os... and the freebsd 12 check has been terminated...

The dbus tests are somewhat timing sensitive and do sometimes fail. This is a known issue and rerunning the failed job should fix it.

CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ endif (CMAKE_VERSION VERSION_LESS 3.9)
 # fix macOS RPATH issues
 set (CMAKE_MACOSX_RPATH 1)
 set (CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+set (CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```
robaerd, line 33: Should we really add the rpath to **all** installed binaries? I think rpath is generally discouraged for packaging IIRC.
If the rpath however is only needed for jni/jna, I think it would be best to only add it for the jni/jna target, like in https://github.com/ElektraInitiative/libelektra/pull/3755.
@kodebach maybe this was also the reason why the packaged kdb tests failed? (only a guess)

tucek: @kodebach i will wait for your response regarding this issue...
If we are still out of ideas, i will reinstate the DeletLocalRef in jni.c and comment out kdb all in Jenkinsfile, so we can proceed like discussed after the Elektra meeting...
Since the sporadic testjna_gradle segfault most likely is associated with the Cleaner...

kodebach: Yes, this could be the reason, although I really don't know so just try and see what works...

tucek: @robaerd i've tried, but i honestly don't really know what i'am doing ;)
could you maybe give me some insight or just try it yourself / propose a change?

robaerd:
The testmod_jni error does also occur if the build directory is deleted after installing.
E.g.: 
- `make install`
- `kdb run_all` -> all tests succeed
- delete `build/`
- `kdb run_all` -> testmod_jni fails 

[error-report.log](https://github.com/ElektraInitiative/libelektra/files/6526320/error-report.log)

tucek:
> The testmod_jni error does also occur if the build directory is deleted after installing.
> E.g.:
> 
> * `make install`
> * `kdb run_all` -> all tests succeed
> * delete `build/`
> * `kdb run_all` -> testmod_jni fails
> 
> [error-report.log](https://github.com/ElektraInitiative/libelektra/files/6526320/error-report.log)

I'm trying to find the dependency to the build dir using strace...

tucek:
testmod_jni wants to access build/src/bindings/jna/libelektra/build/libs/libelektra-0.9.5-all.jar because it is injected into the code at build time in https://github.com/tucek/libelektra/blob/315a909579f5036c39544c65e6caa51cd5bd61fb/src/plugins/jni/testmod_jni.h.in#L12

kodebach:
> We should find another way to set the classpath for the test, so it works with installed jna bindings and the build local version...

I don't think there is currently an easy way to pass such a parameter to the test. There is a mechanism for plugins that need extra executables to work (e.g. `gopts`), but this is not directly usable here, since that would just create an second copy of the JAR file so that a relative path works.

I think it is totally fine to just exclude the test in the installed version. Either via some CMake setup, or just by checking the `main()` of the test if the file `CLASSPATH` exists (`if (access(CLASSPATH, F_OK ) == 0)`).

tucek:
@kodebach what about first checking for the build version and then falling back to the install version and failing verbosely when neither has been found?

kodebach:
> @kodebach what about first checking for the build version and then falling back to the install version and failing verbosely when neither has been found?

This is also just a work around and not a permanent solution. Take this scenario for example:
- I download the source code, build and install.
- I now change some code, rebuild and get a new JAR in my build folder
- Crucially, I _do not_ install again.
- If I know run the installed version, `testmod_jni` will happily use the new JAR in my build folder. That could results in a test failure, because the installed plugin doesn't match the build JAR.

A similar scenario is possible, if we first check for the installed version.

The only permanent solution is, if the installed version always uses the installed JAR and the build folder version always uses the build folder JAR. If you really want to invest the time, look into how `srcdir_file` and `bindir_file` are implemented for tests:

https://github.com/ElektraInitiative/libelektra/blob/3a3b7af421995d0ad17c09b3fb4d4ea4b178b43f/tests/cframework/tests.c#L281-L289

However, I have to stress again, I **do not** think it is worth your time right now and disabling the tests in installed versions (probably easiest by just not installing the `testmod_jni` file) is the better solution.

tucek:
The jni / package step is hereby solved.

tucek:
The last thing i would like to try is enabling the Cleaner again, while adding a locking so that KeySets and Keys are not being cleaned up in parallel.

tucek:
CI fails again because of some links being temporary unavailable... IMHO checks for broken links should only be executed manually or maybe automatically by the release pipeline...

mpranj:
> CI fails again because of some links being temporary unavailable... IMHO checks for broken links should only be executed manually or maybe automatically by the release pipeline...

But then we do not know when a PR introduces a broken link until the release day, which is suboptimal. The way it is now we immediately see broken links and can fix or temporarily whitelist them, even if it is not our own fault.

tucek:
> > CI fails again because of some links being temporary unavailable... IMHO checks for broken links should only be executed manually or maybe automatically by the release pipeline...
> 
> But then we do not know when a PR introduces a broken link until the release day, which is suboptimal. The way it is now we immediately see broken links and can fix or temporarily whitelist them, even if it is not our own fault.
That's correct, but IMHO the CI build should not depend on all referenced websites being up to succeed. But it's not my place to prioritize such issues for this project.

kodebach:
> But then we do not know when a PR introduces a broken link until the release day, which is suboptimal.

A good compromise would probably to set the link checker, such that it only checks links in files that have been modified by the PR. That way we can catch broken links before they are added, but don't annoy people with websites that are temporarily offline.

But such a setup is more complicated than the current one and since the Link Check is a separate build job, we can always just ignore a failure and merge anyway.

mpranj:
> the CI build should not depend on all referenced websites being up to succeed.

Agreed.

> But it's not my place to prioritize

Good suggestions are always welcome. :heart: 

> But such a setup is more complicated than the current one and since the Link Check is a separate build job, we can always just ignore a failure and merge anyway.

I also think the simplest is just to ignore it, but be aware of what is going on. Better approaches are definitely welcome but currently probably not worth our time.

tucek:
From my perspective, this PR is rdy to be merged.

markus2330 (approved):
Amazing job! :sparkling_heart: Thanks to everyone participating! :revolving_hearts: 

Now lets do further testing, especially also with a more real-world plugin (H3) and `kdb mount`.

mpranj (approved):
Great work, thank you so much! :rocket:

doc/news/_preparation_next_release.md
```diff
@@ -139,7 +151,42 @@ you up to date with the multi-language support provided by Elektra.
...
+- Updated documentation
+- Increased minimum required JDK version to 9
+- Migrated native resource clean-up from `finalize()` to `Cleaner`
+  - Please revisit the documentation for `Key::release` and `KeySet::release` for recommended resource release handling
+- Extracted exceptions from Key class introducing the following changes
+  - Removed unused `KeyTypeConversionException`
+  - Introduced `KeyCreateFailedException`, `KeyReleasedException`
+  - Renamed `KeyInvalidNameException` to `KeySetNameFailedException`
+  - Renamed `KeyTypeMismatchException` to `KeyBinaryTypeNotSupportedException`
+- Introduced `KeySetReleasedException` being thrown when a release `KeySet` is being accessed
```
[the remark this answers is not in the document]
tucek: i will fix that in my next pr, thx

Issue #3750, opened 2021-04-07
elektraMalloc et al.


kodebach:
Continuation of `elektraMalloc` et al. discussion from #3731.

The question is: Should `elektraMalloc` et al. be public API? Should we even keep them at all?

---

The following is a response to https://github.com/ElektraInitiative/libelektra/issues/3731#issuecomment-814615596

> I agree that they provide little benefit. But removing them also provides little benefit and a lot of effort.

Removing the functions that have the same API as their libc counterpart should just be a regex-replace.

> These are probably bugs, as our API design says that for every (public) function allocating something there needs to be a function freeing it.

Technically the function allocating is `elektraMalloc`/`elektraCalloc` and the function freeing is `elektraFree`. So no bug here ;)

More seriously, there aren't any functions in `libelektra-core` or `libelektra-kdb` AFAIK that return something the caller has to free with `elektraFree`. But there are some functions in e.g. `libelektra-ease` that just return a newly allocated string. Do you really want to add something like `elektraFreeString`, which then just wraps `elektraFree`?

> I don't know of any if you want to change the allocator **only** for Elektra. There are many libraries providing similar mechanisms, e.g. [libltdl](https://www.gnu.org/software/libtool/manual/html_node/Using-libltdl.html) or glyphy (IIRC, project seems to be gone now) or even implement their own allocator, e.g. [harfbuzz](https://github.com/harfbuzz/harfbuzz/blob/c39ab82c90479341dcf28eaa8174af6f08c0d7ae/src/hb-pool.hh)

`libltdl`'s approach is totally different. They use `lt_dlmalloc`, but that is not a function. It is global function _pointer_ that by default is set to `malloc`. So calling code can just set the pointer to a different value, which is much simpler than actually replacing a function like `elektraFree`. However, calling `malloc` through a function pointer will (probably) cause a small performance hit.

`harfbuzz` implements it's own allocator. That can make sense, if there is specific knowledge that can be used to improve performance. But that means the allocator would be specific to Elektra. Then you could just recompile Elektra and use compiler/linker options to replace `malloc` et al.

[...]

Issue #2202, opened 2018-08-21
ksAppendKey locks name, ksPop does not unlock name


mpranj:
It makes sense that ksAppendKey() locks the key name, since it determines the position of the key in the keyset. However, I did not find an API function that clears the `KEY_FLAG_RO_NAME` flag.

It would seem intuitive to me, that ksPop (or lookup with `KDB_O_POP`) would unlock the name again. Is a key not meant to be re-used? Am I missing some important part of the API design, or would my suggestion be of interest.

The question comes from the context of writing the storage plugin tests/scenarios. I moved many tests out of my mmapstorage plugin to ctest. Currently, it tests dump and mmapstorage only.

e1528532:
I've encountered the same issue in the pluginprocess library and worked around by copying the relevant stuff manually after i finally understood what caused the issue :) To me this behavior was also unintuitive.

markus2330:
Yes, you are right. If the ref counter is back to its original value, unlocking a key's name should be safe. Are there unit tests checking for this?

mpranj:
> Are there unit tests checking for this?
I'm not sure what you mean. I encountered it when writing tests. There is however, no unit test, checking whether a key name is unlocked when the ref counter is 0 again.

markus2330:
> There is however, no unit test, checking whether a key name is unlocked when the ref counter is 0 again.

I meant if there is a unit test checking the other way round (if the key name stays locked, even after removing it from every KeySet). This would break the ABI tests then. If the test suite says nothing after changing this behavior, we are fine about that.

Another problem might be with applications manually tempting with the ref counter. One could add a key to a keyset, then decrease the ref counter, then change the name, breaking the KeySet's invariant (which now got even more important because of the OPMHP from @KurtMi). To solve this problem we would need a second ref counter which only remembers the number of KeySet's a key is in.

mpranj:
For me the problem is "_solved_". I work around this by clearing the flag by hand (since in the unit tests I know I hold the only reference to it). I wanted to be able to pop a key from a keyset, change the name, and then let the storage plugin store it again. In the case of mmapstorage, it's a very interesting test case since it changes a lot in the data and reveals memory handling problems.

The code change seems trivial. If it really has large implications, we can put it to future ideas or close it.

markus2330:
Yes, it is no problem if you clear the flag in the unit test (it is, however, not an ABI test then).

markus2330:
[...]

PR #3919, opened 2021-07-04
doc: add decision docs for decision meeting 15.07.2021

### LINKED ISSUE #3839: open decisions
Decide about:

- [ ] Error Handling (e.g. adding thread-local var) #3797  #3737  #3029
- [ ] Memory Management (elektraMalloc etc.) #3750 #3731

Analysis needed first:

- [ ] aliases #3740 #3736 #3730
- [ ] iterators #3171 #2991 #3189
- [ ] refcounting #3869

lawli3t:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels

Closes #3839

### DIFF
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -50,17 +50,17 @@ section here.
 - [Ensure](ensure.md) (@kodebach)
 - [Capabilities](capabilities.md) (@markus2330)
 - [Error Semantics](error_semantics.md) (API)
+- [Error Handling](error_handling.md)
+- [Remove elektraMalloc et al.](remove_elektra_malloc.md)
+- [Functions copying into buffer](functions_with_buffers.md)
 
 ## Decided
 
 - [Array for Warnings](warning_array.md)
 - [Array](array.md)
-- [Error Handling](error_handling.md)
 - [Iterators](iterators.md)
 - [keyString() return value](key_string_return_value.md)
 - [Reference Counting](reference_counting.md)
-- [Remove elektraMalloc et al.](remove_elektra_malloc.md)
-- [Functions copying into buffer](functions_with_buffers.md)
 
 ## In Discussion
 
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -55,6 +55,12 @@ section here.
 
 - [Array for Warnings](warning_array.md)
 - [Array](array.md)
+- [Error Handling](error_handling.md)
+- [Iterators](iterators.md)
+- [keyString() return value](key_string_return_value.md)
+- [Reference Counting](reference_counting.md)
+- [Remove elektraMalloc et al.](remove_elektra_malloc.md)
+- [Functions copying into buffer](functions_with_buffers.md)
 
 ## In Discussion
 
@@ -62,11 +68,6 @@ section here.
 - [Store the escaped and/or unescaped key name](store_name.md)
 - [Spec Expressiveness](spec_expressiveness.md)
 - [Metadata in Spec Namespace](spec_metadata.md)
-- [Error Handling](error_handling.md)
-- [Iterators](iterators.md)
-- [keyString() return value](key_string_return_value.md)
-- [Reference Counting](reference_counting.md)
-- [Remove elektraMalloc et al.](remove_elektra_malloc.md)
 
 ## Delayed
 
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -62,6 +62,11 @@ section here.
 - [Store the escaped and/or unescaped key name](store_name.md)
 - [Spec Expressiveness](spec_expressiveness.md)
 - [Metadata in Spec Namespace](spec_metadata.md)
+- [Error Handling](error_handling.md)
+- [Iterators](iterators.md)
+- [keyString() return value](key_string_return_value.md)
+- [Reference Counting](reference_counting.md)
+- [Remove elektraMalloc et al.](remove_elektra_malloc.md)
 
 ## Delayed
 
--- a/doc/decisions/error_handling.md
+++ b/doc/decisions/error_handling.md
@@ -1,0 +1,28 @@
+# Error Handling
+
+## Problem
+
+Currently, a lot of functions return the error value, regardless of error. These poses some problems, for instance authors of bindings, where it is hard to tell which error exactly occured. A big improvement to this would be returning different error codes for different types of errors, in order to be able to provide better information in cases of errors. Some functions also make use of an `errorKey`. While this seems like a decent approach, to provide additional informations in cases of error, it is currently very badly documented. It is not really transparent for the users of the function, what kind of information is included in the `errorKey` and even which messages could possibly be returned.
+
+## Constraints
+
+We also have to consider that many parts of the API very constrained in what values they can use for error codes. If a function returns a pointer the only possible error value by definition is NULL.
+
+## Assumptions
+
+## Considered Alternatives
+
+- Change return types for many functions across the API
+
+## Decision
+
+- Add documentation for errorKey to Doxygen
+- Add thread-local error handling for getting information about errors
+
+## Rationale
+
+## Implications
+
+## Related Decisions
+
+## Notes
--- a/doc/decisions/error_handling.md
+++ b/doc/decisions/error_handling.md
@@ -2,7 +2,7 @@
 
 ## Problem
 
-Currently, a lot of functions return the error value, regardless of error. These poses some problems, for instance authors of bindings, where it is hard to tell which error exactly occured. A big improvement to this would be returning different error codes for different types of errors, in order to be able to provide better information in cases of errors. Some functions also make use of an `errorKey`.
+Currently, a lot of functions return the error value, regardless of error. These poses some problems, for instance authors of bindings, where it is hard to tell which error exactly occurred. A big improvement to this would be returning different error codes for different types of errors, in order to be able to provide better information in cases of errors. Some functions also make use of an `errorKey`.
 
 While this seems like a decent approach, to provide additional informations in cases of error, it is currently very badly documented. It is not really transparent for the users of the function, what kind of information is included in the `errorKey` and even which messages could possibly be returned.
 
--- a/doc/decisions/error_handling.md
+++ b/doc/decisions/error_handling.md
@@ -18,8 +18,8 @@ We also have to consider that many parts of the API very constrained in what val
 
 ## Decision
 
-- Update documentation in `doc/dev/error-*` and link to them in the documentation 
-for the module `kdb`
+- Update documentation in `doc/dev/error-*` and link to them in the documentation
+  for the module `kdb`
 - Add second channel for getting information about errors
 - Return error codes directly from functions where failures are expected, e.g. `kdbGet`, `keySetName`
 - Harmonize return values from all functions and move error reporting to second channel
--- a/doc/decisions/error_handling.md
+++ b/doc/decisions/error_handling.md
@@ -20,7 +20,7 @@ We also have to consider that many parts of the API very constrained in what val
 
 - Add documentation for errorKey to Doxygen
 - Add second channel for getting information about errors
-- Return additional information about errors from functions where errors can be expected
+- Return error codes directly from functions where failures are expected, e.g. `kdbGet`, `keySetName`
 - Harmonize return values from all functions and move error reporting to second channel
 
 ## Rationale
--- a/doc/decisions/error_handling.md
+++ b/doc/decisions/error_handling.md
@@ -2,7 +2,9 @@
 
 ## Problem
 
-Currently, a lot of functions return the error value, regardless of error. These poses some problems, for instance authors of bindings, where it is hard to tell which error exactly occured. A big improvement to this would be returning different error codes for different types of errors, in order to be able to provide better information in cases of errors. Some functions also make use of an `errorKey`. While this seems like a decent approach, to provide additional informations in cases of error, it is currently very badly documented. It is not really transparent for the users of the function, what kind of information is included in the `errorKey` and even which messages could possibly be returned.
+Currently, a lot of functions return the error value, regardless of error. These poses some problems, for instance authors of bindings, where it is hard to tell which error exactly occured. A big improvement to this would be returning different error codes for different types of errors, in order to be able to provide better information in cases of errors. Some functions also make use of an `errorKey`.
+
+While this seems like a decent approach, to provide additional informations in cases of error, it is currently very badly documented. It is not really transparent for the users of the function, what kind of information is included in the `errorKey` and even which messages could possibly be returned.
 
 ## Constraints
 
@@ -17,9 +19,17 @@ We also have to consider that many parts of the API very constrained in what val
 ## Decision
 
 - Add documentation for errorKey to Doxygen
-- Add thread-local error handling for getting information about errors
+- Add second channel for getting information about errors
+- Return additional information about errors from functions where errors can be expected
+- Harmonize return values from all functions and move error reporting to second channel
 
 ## Rationale
+
+This simplifies and unifies error reporting across all functions. It also introduces
+a default way for the developers to handle error reporting. Developers of
+bindings for Elektra can rely on the existing error reporting feature without
+having to introduce custom error messages, that can vary wildly between different
+bindings.
 
 ## Implications
 
--- a/doc/decisions/error_handling.md
+++ b/doc/decisions/error_handling.md
@@ -18,7 +18,8 @@ We also have to consider that many parts of the API very constrained in what val
 
 ## Decision
 
-- Add documentation for errorKey to Doxygen
+- Update documentation in `doc/dev/error-*` and link to them in the documentation 
+for the module `kdb`
 - Add second channel for getting information about errors
 - Return error codes directly from functions where failures are expected, e.g. `kdbGet`, `keySetName`
 - Harmonize return values from all functions and move error reporting to second channel
--- a/doc/decisions/functions_with_buffers.md
+++ b/doc/decisions/functions_with_buffers.md
@@ -1,0 +1,40 @@
+# Functions copying into buffers
+
+## Problem
+Currently the way functions like keyGetName() work is by passing a buffer with 
+a maxSize and if the buffer is large enough, the value gets copied into the 
+buffer. This leads to the user having to write a lot of surrounding boilerplate 
+code, checking for the size of every value / name they want to copy into a buffer.
+
+## Constraints
+
+## Assumptions
+
+## Considered Alternatives
+
+## Decision
+
+- Remove Functions:
+  - keyGetName()
+  - keyGetUnescapedName()
+  - keyGetBaseName()
+  - keyGetString()
+  - keyGetBinary()
+- add documentation that you should use strncpy() / memcpy() instead:
+```c
+// str values
+strncpy(..., keyName (k), ...)
+// binary values
+memcpy(..., keyValue (k), ...)
+```
+
+## Rationale
+The functions clutter the API and try to replace existing builtin functionality 
+for little to no gain. This makes the API leaner while also retaining its 
+functionality.
+
+## Implications
+
+## Related Decisions
+
+## Notes
--- a/doc/decisions/functions_with_buffers.md
+++ b/doc/decisions/functions_with_buffers.md
@@ -1,9 +1,10 @@
 # Functions copying into buffers
 
 ## Problem
-Currently the way functions like keyGetName() work is by passing a buffer with 
-a maxSize and if the buffer is large enough, the value gets copied into the 
-buffer. This leads to the user having to write a lot of surrounding boilerplate 
+
+Currently the way functions like keyGetName() work is by passing a buffer with
+a maxSize and if the buffer is large enough, the value gets copied into the
+buffer. This leads to the user having to write a lot of surrounding boilerplate
 code, checking for the size of every value / name they want to copy into a buffer.
 
 ## Constraints
@@ -21,6 +22,7 @@ code, checking for the size of every value / name they want to copy into a buffe
   - keyGetString()
   - keyGetBinary()
 - add documentation that you should use strncpy() / memcpy() instead:
+
 ```c
 // str values
 strncpy(..., keyName (k), ...)
@@ -29,8 +31,9 @@ memcpy(..., keyValue (k), ...)
 ```
 
 ## Rationale
-The functions clutter the API and try to replace existing builtin functionality 
-for little to no gain. This makes the API leaner while also retaining its 
+
+The functions clutter the API and try to replace existing builtin functionality
+for little to no gain. This makes the API leaner while also retaining its
 functionality.
 
 ## Implications
--- a/doc/decisions/functions_with_buffers.md
+++ b/doc/decisions/functions_with_buffers.md
@@ -21,7 +21,7 @@ code, checking for the size of every value / name they want to copy into a buffe
   - keyGetBaseName()
   - keyGetString()
   - keyGetBinary()
-- add documentation that you should use strncpy() / memcpy() instead:
+- add documentation in API docu about life-time and add in release notes that you should use strncpy() / memcpy() instead:
 
 ```c
 // str values
--- a/doc/decisions/iterators.md
+++ b/doc/decisions/iterators.md
@@ -27,6 +27,9 @@ iterators for 1.0 and only the external instead.
   - ksSetCursor
   - ksHead
   - ksTail
+  - keyRewindMeta
+  - keyNextMeta
+  - keyCurrentMeta
 - change `ksAtCursor` to `ksAt`
 - add implementation / documentation / tests for the external iterator
 - start using external iterators in new code
--- a/doc/decisions/iterators.md
+++ b/doc/decisions/iterators.md
@@ -6,13 +6,13 @@ The internal iterator inside KeySets seems to cause more problems than it solves
 
 ## Constraints
 
-To ensure backwards compatibility, we could embed one such iterator into 
+To ensure backwards compatibility, we could embed one such iterator into
 KeySets, but recommend that people use an external instance (int) in new code.
 
 ## Assumptions
 
-@raphi011 made benchmarks showing that external and internal iterators have 
-about the same speed. So it should be a safe choice to not provide the internal 
+@raphi011 made benchmarks showing that external and internal iterators have
+about the same speed. So it should be a safe choice to not provide the internal
 iterators for 1.0 and only the external instead.
 
 ## Considered Alternatives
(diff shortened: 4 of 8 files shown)


lawli3t:
@kodebach in this comment https://github.com/ElektraInitiative/libelektra/issues/3750#issuecomment-816099318 you mentioned something about

> changes to the backend and global plugin stuff

Would you like to add this to the agenda of this meeting?

kodebach:
> > changes to the backend and global plugin stuff
> 
> Would you like to add this to the agenda of this meeting?

I "discussed" this with @markus2330. We both think that this would probably too much and is somewhat unrelated to the rest of the agenda (which is more concerned with the C API). Despite that I would still like to briefly discuss the changes of #3693 at the end of the meeting. I would just give a brief overview of what the idea is to get some quick feedback on the general concepts. For a more in-depth discussion, we will definitely need a separate meeting and people will probably have to read the new documentation first.

doc/decisions/key_string_return_value.md
```diff
@@ -0,0 +1,33 @@
...
+On one hand always returning a string allows things like `strlen(keyString(k)) == 0` to check for an empty string. But at the same time people might expect that `strlen(keyString(k)) < MAX_LEN` also works, but this might silently break code down the line, if `k` is binary and `strlen("(binary)") < MAX_LEN`.
+
+## Considered Alternatives
+
+An alternative would be to always store a zero byte after all key values, even if they are binary (might be done already). Then we can safely return `k->data.c == NULL ? "" : k->data.c`. It may contain incomplete data and the `MAX_LEN` problem from above still applies, but there are no segfaults and you don't get return values that have nothing to do with the actual data.
+
+## Decision
+
+- Obviously just return the string value, if the key is a non-empty string key.
+- Return NULL, if the key is binary (no matter what the value is)
```
markus2330, line 22: What do we need a keyString then at all? The idea of keyString was that it can safely be passed to printf or similar.

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,45 @@
...
+
+`keyIncRef` and `keyDecRef` would be internal/private/static (whatever fits). The same `ks*` functions should exist for `KeySet`.
+
+## Rationale
+
+The end effect is basically the same. Users should
+
+- use `keyNew` and `keyCopy`/`keyDup` to create entirely new `Key *` that don't have any links to other `Key *`.
+- use `keyBorrow` to create a new `Key *` that is a reference an existing `Key *`.
+- use `keyDel` to dispose of a `Key *` (whether the `Key *` comes from `keyNew`, `keyCopy`/`keyDup` or `keyBorrow`).
```
markus2330, line 39: At the moment keyDel does not decrement the counter, so how to use keyDel to get rid of a key from keyBorrow?

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Implications
+
```
markus2330, line 42: Implications should state which code needs to be changed.

doc/decisions/remove_elektra_malloc.md
```diff
@@ -0,0 +1,40 @@
+# Remove elektraMalloc() et al.
```
markus2330, line 1: A related question is the documentation of the life-time of our objects. Here also improvements of the docu is needed.

doc/decisions/remove_elektra_malloc.md
```diff
@@ -0,0 +1,40 @@
...
+- `elektraStrDup`
+- `elektraStrLen`
+
+## Constraints
+
+Proprietary apps are sometimes delivered together with a libc (either static or a startup script makes sure their own libs are found). If they would use Elektra, and the person wants the app to use the global Elektra it is a valid use case for the dedicated functions.
+
+## Assumptions
+
+According to Klemens' opinion, the only safe way to keep elektraMalloc et al. is to define that they will always call their libc counterpart (malloc et al.) and their purpose is simply to add assertions. Then we can also make them private, because you can just call free to free any pointer returned by Elektra. Ideally, I'd remove the functions, but it seems unlikely this will be accepted.
```
markus2330, line 21: What does "I" refer to? Of course removing the functions is an option, otherwise we would not need the meeting.

doc/decisions/remove_elektra_malloc.md
```diff
@@ -0,0 +1,40 @@
...
+- Completely replace elektraMalloc / elektraCalloc with simple calls to malloc
+- Leave as it is
+
+## Decision
+
+## Rationale
+
+My main goal with removing elektraMalloc et al. is to improve developer friendliness, especially for newcomers. It is still annoying that I using e.g. strndup may be problematic, because it doesn't use elektraMalloc for allocations. But when I was new to Elektra (and I suspect many other felt the same), it was downright frustrating that you have to remember to use Elektra's special sauce instead of the standard stuff.
+
+The functions `elektraRealloc` and `elektraMemDup` (renamed from elektraStrNDup) can stay, because they are an improvement upon the libc.
```
markus2330, line 34: This would be an decision, not a rationale.

markus2330 (approved):
Thank you so much for the preparation of the meeting!

markus2330:
Aliases are not mentioned?

lawli3t:
I have added and edited the documents with the decision from our meeting. Feel free to comment and add to those documents in case I have forgotten something.

@markus2330 @kodebach @mpranj @tucek

doc/decisions/error_handling.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Assumptions
+
+## Considered Alternatives
+
+- Change return types for many functions across the API
+
+## Decision
+
+- Add documentation for errorKey to Doxygen
```
kodebach, line 21: Just to be sure, this means add documentation in one central location and link to that?

doc/decisions/error_handling.md
```diff
@@ -0,0 +1,38 @@
...
+- Add second channel for getting information about errors
+- Return additional information about errors from functions where errors can be expected
```
kodebach, line 23:
```suggestion
- Return error codes directly from functions where failures are expected, e.g. `kdbGet`, `keySetName`
```

To clarify further, IMO this should be the distinction between expected and unexpected failures:
- Expected failures are hard or impossible to detect before calling a function. For example an invalid key name is hard to detect (in general) without calling `keySetName` or a similar function. IO errors during `kdbGet` are impossible to detect beforehand.
- Unexpected failures are errors that can easily prevented by the caller. For example it is very easy to check for `NULL` pointers or locked keys before calling a function.

To be even more concrete, if the check to prevent an error case is almost as expensive as the function call itself (e.g. checking for invalid key names before `keySetName`), then we should return a specific error directly. Depending on the kind of function "return directly" could mean via return value or `errorKey`.

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,39 @@
...
+## Assumptions
+
+## Considered Alternatives
+
+- make ref counting thread safe (probably useful for JNI)
+- start with 1 for reference counting and let keyDecRef do keyDel
+
+## Decision
+
+- add second reference counter to Key (size 1 byte)
```
kodebach, line 21:
```suggestion
- add second reference counter to Key
```

I think the size is still TBD. Also it shouldn't really be something that is fixed in the decision.

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,39 @@
...
+- one is for KeySet, the other is for the rest (for external keyset references)
```
kodebach, line 22:
```suggestion
- One counter is for locking references, the other one for general references. A locking reference automatically locks/unlocks the keyname.
```

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,39 @@
...
+- introduce reference counter for KeySets (for external keyset references)
+- keyIncRef Fehlercode returnen on MAX_SIZE
```
kodebach, line 24:
```suggestion
- `keyIncRef` (or equivalent) return error in case of reference counter overflow
```

doc/decisions/remove_elektra_malloc.md
```diff
@@ -0,0 +1,58 @@
...
+- Completely replace `elektraMalloc` / `elektraCalloc` with simple calls to malloc
+
+## Decision
+
+- keep current state with the custom functions
+- make it optional for plugin developers to use c builtins or the custom Elektra
+  functions
+- Fix places where non-Elektra functions get used
+- Remove `elektraMalloc` / `elektraCalloc` / `eletraRealloc` from public API
+
```
kodebach, line 45:
```suggestion
- Remove all functions that don't actually involve any memory allocations (e.g. `elektraStrLen`, `elektraStrCmp`)
- Add `elektraStrNDup` and other `stdlib` equivalents that do memory allocation.

```

Not sure, if we agreed on removing all of those functions or just the utterly useless `elektraStrLen`.

We recently renamed `elektraStrNDup` to `elektraMemDup`, because it didn't actually do a `strndup`. But if we want to keep the possibility of swapping the allocator, we need an equivalent for `strndup`. We should also check the calls of `elektraMemDup` again and ensure `memdup` is wanted and not `strndup`.

doc/decisions/remove_elektra_malloc.md
```diff
@@ -0,0 +1,58 @@
...
+- Completely replace `elektraMalloc` / `elektraCalloc` with simple calls to malloc
+
+## Decision
+
+- keep current state with the custom functions
+- make it optional for plugin developers to use c builtins or the custom Elektra
+  functions
+- Fix places where non-Elektra functions get used
+- Remove `elektraMalloc` / `elektraCalloc` / `eletraRealloc` from public API
```
kodebach, line 44:
```suggestion
- Remove everything except `elektraFree` from public API
```

Not sure how many of the functions are actually public API, but if we remove `elektraMalloc`, we also need to remove things like `elektraStrDup`.

doc/decisions/error_handling.md
lawli3t: We haven't really discussed that in the meeting, i forgot to mention it sadly. From the discussion on the issue it seems that this was the preferred solution.

doc/decisions/functions_with_buffers.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Decision
+
+- Remove Functions:
+  - keyGetName()
+  - keyGetUnescapedName()
+  - keyGetBaseName()
+  - keyGetString()
+  - keyGetBinary()
+- add documentation that you should use strncpy() / memcpy() instead:
```
markus2330, line 24: strndup seems to be the easiest way for users who simply want a duplicate string

```suggestion
- add documentation that you should use strndup() instead:
```

the second argument should be a call to keyGet*Size

doc/decisions/iterators.md
```diff
@@ -0,0 +1,52 @@
+# Iterators
+
+## Problem
+
+The internal iterator inside KeySets seems to cause more problems than it solves.
```
markus2330, line 5: Maybe write a bit about the background why it was introduced and that this is not used anymore (like discussed in the meeting: to remember errorKey in kdbSet).

doc/decisions/iterators.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Constraints
+
+To ensure backwards compatibility, we could embed one such iterator into
```
markus2330, line 9: Remove (no backward compatibility needed)

doc/decisions/key_string_return_value.md
```diff
@@ -0,0 +1,38 @@
+# keyString() return value
+
+## Problem
+
+When using keyString() on empty / binary values the return values are the literal strings (null) / (binary). This seems very awkward and unintuitive from a user's perspective.
+
+## Constraints
+
+Because the return value of the function is const char\*, there are not so many possibilities for a change to this behaviour, as every possible return value - except for a NULL pointer - could just be the value of the string of the Key. Possible changes to this behaviour would be:
```
markus2330, line 9: Actually we could also change the return value, so this is not really a constraint.

doc/decisions/key_string_return_value.md
```diff
@@ -0,0 +1,38 @@
...
+- `key == <binary>` return 0, error code via second channel
+- everything else as is
+
+## Rationale
+
+0 seems like the most intuitive value to return in case of an error, although
+this introduces the possibility of segfaults for users of the library. With
+the introduction of a second channel for reporting errors, users can check the
+error messages in case of segfaults - which alleviates this issue. The first
+thing in case of an error should be checking the error message.
```
markus2330, line 32: I think the main argumentation is that printf does not segfault on null pointers anymore.

doc/decisions/error_handling.md
markus2330: We already have `doc/dev/error-*`, so updating and link them into the API docu of `kdb*` should be enough.

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,39 @@
...
+## Assumptions
+
+## Considered Alternatives
+
+- make ref counting thread safe (probably useful for JNI)
+- start with 1 for reference counting and let keyDecRef do keyDel
+
+## Decision
+
+- add second reference counter to Key
```
markus2330, line 21: See #2202: probably it is simply a documentation issue and no second ref counter is needed

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,39 @@
...
+- One counter is for locking references, the other one for general references. A locking reference automatically locks/unlocks the keyname.
+- introduce reference counter for KeySets (for external keyset references)
+- `keyIncRef` (or equivalent) return error in case of reference counter overflow
```
markus2330, line 24: Let us settle for keyBorrow (see #3949) unless someone comes up with a better name (maybe keyShare). The rename should not be too costly, keyIncRef is not heavily used.

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,39 @@
...
+- use fixed sized types for reference counters
+- increment/decrement references before/after passing instances to plugins
+
+## Rationale
+
+- Adding a second reference counter to Key and reducing the size of both significantly
+  actually saves memory compared to the previous solution.
```
markus2330, line 31: not really, in #3949 is now 16bit proposed, so at best we stay at the same size.

doc/decisions/reference_counting.md
```diff
@@ -0,0 +1,39 @@
...
+- The added complexity of maintaining two reference counters is worth the
```
markus2330, line 32: This is not an argument, what is the gained functionality?

doc/decisions/reference_counting.md
markus2330: Still missing.

markus2330 (approved):
Thank you, great work!

doc/decisions/functions_with_buffers.md
kodebach: > strndup seems to be the easiest way for users who simply want a duplicate string

True, but `strndup(keyName(k), size)` is not a replacement for what `keyGetName(k, buffer, size)` does, because it cannot copy into a pre-existing buffer.

doc/decisions/key_string_return_value.md
kodebach: I think we can say "we don't want to change the return value", because that would make the function very inconvenient to use.

doc/decisions/reference_counting.md
kodebach: Honestly, what was the decision meeting for, if we immediately reverse one of the main decisions?

See https://github.com/ElektraInitiative/libelektra/pull/3949#issuecomment-882484784 for an idea on how documentation/API naming can make the second ref counter much easier to understand.

doc/decisions/reference_counting.md
kodebach: The current ref counter is `size_t`, so on most modern (i.e. 64 bit) machines we safe 4 Bytes, assuming both counters will be 16bit. #3949 actually only proposes a size of `KeySet` not for `Key`.

doc/decisions/reference_counting.md
kodebach: I understood it as: Adding the second ref counter can be done in a way that introduces very little complexity. At the same time (automatically) unlocking keys is a big benefit, because you don't need to make unnecessary copies of keys.

doc/decisions/functions_with_buffers.md
markus2330: Good point, we need to be clear who we want to address with our API docu. I was thinking about users who want to use a name or value beyond the life-time of the key object (and give them the easiest way to do so).

For copying into other buffers there are endless ways, that is why I thought it is better we do not describe this. @kodebach which kind of user (or use case) were you thinking of? Did you think about the migration guide for users of our previous API? Shouldn't that be better in the release notes?

doc/decisions/key_string_return_value.md
markus2330: "return value" is an error from the original text. Probably changing the API is meant?

doc/decisions/reference_counting.md
markus2330: > Honestly, what was the decision meeting for, if we immediately reverse one of the main decisions?

In some cases we also went into timeouts and agreed there would be a follow up.

> See #3949 (comment) for an idea on how documentation/API naming can make the second ref counter much easier to understand.

I like the idea but this immediately raises the question why locking the value or meta-data does not get ref-counted.

doc/decisions/functions_with_buffers.md
kodebach: > Shouldn't that be better in the release notes?

I was thinking about release notes, or similar documentation for how e.g. `keyGetName` should be replaced. I assumed this is what the line is about because it says "instead", which implies "to replace".

> use a name or value beyond the life-time of the key object

there is also endless ways to do that ;)

I don't think we need to mention how to do this in the Doxygen of e.g. `keyName()`. Saying `keyName(k)` returns an internal pointer that will only be valid until you call `keyDel (k)` or any function that modifies the name of `k`, should be enough. Elektra's docs are not the right place to explain "How to copy a string C".

markus2330: I agree!

```suggestion
- add documentation in API docu about life-time and add in release notes that you should use strncpy() / memcpy() instead:
```

doc/decisions/reference_counting.md
kodebach: > raises the question why locking the value or meta-data does not get ref-counted.

To me the question is: Why is locking value/meta-data a thing at all? Is there a use-case for it?

AFAICT we only directly use `KEY_FLAG_RO_VALUE` and `KEY_FLAG_RO_META` inside `keySetMeta`. `KEY_LOCK_VALUE` and `KEY_LOCK_META` are not used at all.

For the metadata case we should actually check the key name in `keySetRawValue` and `keyMeta`. Otherwise it won't work with `ksAppendKey (keyMeta (k), ...)` or if you create a meta key via `keySetName (k, "meta:/foo")`.

markus2330: > To me the question is: Why is locking value/meta-data a thing at all? Is there a use-case for it?

It is an invariant of a KeySet that its keys are ordered. The locking ensures this invariant.

> KEY_LOCK_META

The idea was to use this for meta keys. But probably this is now useless, as meta keys can be recognized by the namespace.

kodebach: > It is an invariant of a KeySet that its keys are ordered. The locking ensures this invariant.

ordered **by name**

Locking **value** or **metadata** has nothing to do with this.

markus2330: Sorry, misread. I was thinking about if we can get rid of locking at all. And probably we can, we could simply disallow name changes with ref counter != initial value.

To answer your previous question: Locking of values was added for completeness. Other than read-only keys there is probably no usecase.

kodebach: > we could simply disallow name changes with ref counter != initial value.

Then you can't ever change the name from Java or another language with automatic garbage collection, where incrementing the ref counter is required.

tucek: As far as i understand:
* keysets store keys in sorted by key names
* a key's order within a key set is determined on insertion
* the purpose of locking a key's name is to prevent (potentially) compromising the order of keys for keysets the key is contained in.

To better understand the issue:
* what are the use cases for changing a key's name? why not making it immutable?
* why is order within keysets so important? would it be sufficient to ensure order explicitly for use-cases requiring it?

mpranj: > what are the use cases for changing a key's name? why not making it immutable?

In short, to avoid doing some memory allocations. In my case I had a loop where a key was removed and a new one was added with a new basename. The same can be done with immutable keys, but it requires allocating new memory for N keys instead of reusing the old keys.

I was just confused that unlocking is currently not possible at all. When a key is added to a keyset the name is locked, but when it is removed from the keyset the name is not unlocked. To be consistent we should either have immutable keys or provide a way to unlock the keys again.

> why is order within keysets so important? 

It is required because we use binary search to find a key in the keyset. It is also a nice property and I don't see much reason to rework this, unless there are some huge benefits?

kodebach: I think the Hashmap requires the order as well, since it only stores indices into the array. Also `ksCut` relies on the order of the `KeySet` and in general many things are easier or more efficient to implement, if there is a fixed order.

> I was just confused that unlocking is currently not possible at all.

I had the same experience.

doc/decisions/iterators.md
```diff
@@ -0,0 +1,52 @@
...
+@raphi011 made benchmarks showing that external and internal iterators have
+about the same speed. So it should be a safe choice to not provide the internal
+iterators for 1.0 and only the external instead.
+
+## Considered Alternatives
+
+## Decision
+
+- remove all functions related to the internal iterator:
+  - ksRewind
```
lawli3t, line 23: We should also add `keyRewindMeta` to the list of functions to remove

lawli3t: what about `keyNextMeta()` / `keyCurrentMeta`?

kodebach: Yes, `keyRewindMeta()`, `keyNextMeta()` and `keyCurrentMeta()` should be removed. Both because of the iterator changes and because of the `keyMeta()` changes.

markus2330:
How can we help you to get this merged?

lawli3t:
Good question, I have resolved everything I could immediately yesterday. The following things are still open:

Stuff I can & will add soon:

- [ ] Reworking Problem part of error_handling
- [ ] Improvements to remove_elektra_malloc (restructuring Rationale, )
- [ ] Improve problem part of Iterators


Stuff where I feel I don't have enough knowledge/overview to properly improve the existing documents:

- [ ] Resolving the big discussion on the reference counting Decision part
- [ ] Adding implications for reference counting

kodebach:
I think reference counting was basically resolved with https://github.com/ElektraInitiative/libelektra/pull/3949#issuecomment-918107986. Additionally there will be second counter for the key name lock. However, I wouldn't call that reference counting. I would simply say the lock is shared via a counter.

With the decision to keep `keyIncRef`/`keyDecRef`, the only implications of the reference counting stuff are (AFAIK):
- KeySets can be reference counted now. This is useful for bindings, and some (e.g. C++)  may need updates.
- Reference counts are now limited to `UINT16_MAX - 1` instead of `SIZE_MAX`. This probably won't have any real impact, apart from a change in return type.

The implication of the counter for the name lock is simply that key names will be automatically unlocked again, when a key is removed from all keysets.

markus2330:
> I think reference counting was basically resolved with https://github.com/ElektraInitiative/libelektra/pull/3949#issuecomment-918107986

This still needs to be updated here in the decisions. 1-3 I agree, 4 see below, 5 we need to investigate if it actually solves the JNI problem.

> Additionally there will be second counter for the key name lock. However, I wouldn't call that reference counting. I would simply say the lock is shared via a counter.

This is a separate issue #2202 where the current consensus is contrary: to not have a second ref counter just for being able to rename keys. Let us continue the discussion there.

> This is useful for bindings, and some (e.g. C++) may need updates.

Yes, it needs an update. The C++ binding was very awkward: Key and KeySets had very different behavior when being copied or copy-constructed. It might be a quite big change though as most bindings wrap the C++ KeySet.

> Reference counts are now limited to UINT16_MAX - 1 instead of SIZE_MAX. This probably won't have any real impact, apart from a change in return type.

Actually it is even an improvement as we get rid of platform-dependent behavior.

> The implication of the counter for the name lock is simply that key names will be automatically unlocked again, when a key is removed from all keysets.

This is not an implication of anything said above. Please let us separate topics which are unrelated whenever possible. -> #2202

kodebach:
> Please let us separate topics which are unrelated whenever possible. -> #2202

I just mentioned it, because we talked about it in the meeting for this issue.

doc/decisions/error_handling.md
```diff
@@ -0,0 +1,39 @@
+# Error Handling
+
+## Problem
+
+Currently, a lot of functions return the error value, regardless of error. These poses some problems, for instance authors of bindings, where it is hard to tell which error exactly occurred. A big improvement to this would be returning different error codes for different types of errors, in order to be able to provide better information in cases of errors. Some functions also make use of an `errorKey`.
```
markus2330, line 5: Be more concrete. Afaik the only problems are:

- name modifications which can be either invalid name or locking the key name
- getting values of (non-)binary keys, see also #4194

doc/decisions/error_handling.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Constraints
+
+We also have to consider that many parts of the API very constrained in what values they can use for error codes. If a function returns a pointer the only possible error value by definition is NULL.
+
+## Assumptions
+
+## Considered Alternatives
+
+- Change return types for many functions across the API
```
markus2330, line 17: We could also fix the concrete instances of where multiple errors exist.

doc/decisions/error_handling.md
```diff
@@ -0,0 +1,39 @@
...
+
+- Update documentation in `doc/dev/error-*` and link to them in the documentation
+  for the module `kdb`
+- Add second channel for getting information about errors
+- Return error codes directly from functions where failures are expected, e.g. `kdbGet`, `keySetName`
+- Harmonize return values from all functions and move error reporting to second channel
+
+## Rationale
+
+This simplifies and unifies error reporting across all functions. It also introduces
```
markus2330, line 29: On the contrary: it would make it more complicated and if we are not very careful less unified.

doc/decisions/functions_with_buffers.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Constraints
+
+## Assumptions
+
+## Considered Alternatives
+
+## Decision
+
+- Remove Functions:
```
markus2330, line 18: :+1:

doc/decisions/key_string_return_value.md
```diff
@@ -0,0 +1,39 @@
+# keyString() return value
+
+## Problem
+
+When using keyString() on empty / binary values the return values are the literal strings (null) / (binary). This seems very awkward and unintuitive from a user's perspective.
+
+## Constraints
+
+Because the return value of the function is const char\*, there are not so many possibilities for a change to this behavior, as every possible return value - except for a NULL pointer - could just be the value of the string of the Key. Possible changes to this behavior would be:
+
```
markus2330, line 10: sentence stops in the middle?

doc/decisions/README.md
```diff
@@ -55,6 +55,12 @@ section here.
 
 - [Array for Warnings](warning_array.md)
 - [Array](array.md)
+- [Error Handling](error_handling.md)
+- [Iterators](iterators.md)
+- [keyString() return value](key_string_return_value.md)
```
markus2330, line 60: As this depends on error_handling.md, probably we should also move it to "In Discussion".

doc/decisions/key_string_return_value.md
```diff
@@ -0,0 +1,39 @@
...
+  this introduces the possibility of segfaults for users of the library. printf
+  doesn't cause segfaults anymore, which improves this problem a lot.
+- With the introduction of a second channel for reporting errors, users can check the
+  error messages in case of segfaults - which alleviates this issue. The first
+  thing in case of an error should be checking the error message.
+
+## Implications
+
+## Related Decisions
+
```
markus2330, line 38: error_handling.md should be here.

doc/decisions/reference_counting.md
markus2330: Still missing.

markus2330 (changes requested):
In order to finish this, I think we should move still unclear decisions to "In Discussion". Then we are hopefully able to merge this soon.

markus2330:
Thank you so much! :sparkling_heart:

PR #3889, opened 2021-06-09
CM H3 Java JNI whitelist plugin

tucek:
- Added whitelist Java JNI checker plugin
- Removed plugin loaded java plugin loading

## Basics

These points need to be fulfilled for every PR:

- [ ] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [ ] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [ ] The PR is rebased with current master.

If you have any troubles fulfilling these criteria, please write
about the trouble as comment in the PR. We will help you.
But we cannot accept PRs that do not fulfill the basics.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

Reviewers will usually check the following:

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

If you are already Elektra developer:

- Add the "work in progress" label if you do not want the PR to be reviewed yet.
- Add the "ready to merge" label **if the basics are fulfilled** and you also
  say that everything is ready to be merged.

PR #3971, opened 2021-07-31
#3790 hid errorKey from KDB and refactored Key::createNameless

tucek:
## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -108,13 +108,16 @@ you up to date with the multi-language support provided by Elektra.
 ### Java binding
 
 - Renamed zero argument static factory method `Key::createNameless` to `Key::create`. To migrate to this change, just update calling code to use the new method name.
-- Removed signatures publishing the error key based error handling approach to the Java binding consumer. Such arguments are now created implicitly by the binding. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Affected signatures:
-  - Removed `KDB::open(Key)`, please update usages by calling `KDB::open()` instead.
-  - Removed `KDB::open(KeySet, Key)`, please update usages by calling `KDB::open(KeySet)` instead.
-  - Removed `KDB::close(Key)`, please update usages by calling `KDB::close()` instead.
-  - Updated javadoc for `KDB::set(KeySet, Key)`, better explaining the relevance of the second argument. Please review API usage to consider the more elaborated explanation of how Elektra uses this argument's value.
+- Updated method documentation previously publishing the error key based error handling approach to the Java binding consumer. Such arguments are now explicitly only used for returning warning information in case no error occurred. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Please review API usage to consider the more elaborated explanation of how Elektra uses this argument's value. Affected signatures:
+  - Updated javadoc for `KDB::open(Key)`
+  - Updated javadoc for `KDB::open(KeySet, Key)`
+  - Updated javadoc for `KDB::close(Key)`
+  - Updated javadoc for `KDB::get(Key)`
+  - Updated javadoc for `KDB::get(KeySet, Key)`
+  - Updated javadoc for `KDB::set(KeySet, Key)`, better explaining the relevance of the second argument `parentKey`.
 - Introduced `KeySet::remove(Key)` and `KeySet::remove(String)`
 - Removed `KeySet::lookup(Key, int)` and `KeySet::lookup(String, int)` as well as accompanying flag definitions `KeySet::KDB_O_NONE`, `KeySet::KDB_O_DEL` and `KeySet::KDB_O_POP`. Please use `KeySet::lookup(Key)` and `KeySet::lookup(String)` instead. Instead of `KeySet::KDB_O_DEL`, please consider using `Key::release`. The proper replacement for `KeySet::KDB_O_POP` is `KeySet::remove(Key)` or `KeySet::remove(String)`.
+- Native library proxy interface `Elektra` is now package private (previously was public).
 
 _(Michael Tucek)_
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -105,7 +105,17 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
 Bindings allow you to utilize Elektra using [various programming languages](https://www.libelektra.org/bindings/readme). This section keeps
 you up to date with the multi-language support provided by Elektra.
 
-### <<Binding1>>
+### Java binding
+
+- Renamed zero argument static factory method `Key::createNameless` to `Key::create` and narrowed accessibility to protected, effectively removing it from the public API.
+- Removed signatures publishing the error key based error handling approach to the Java binding consumer. Such arguments are now created implicitly by the binding. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Affected signatures:
+  - removed `KDB::open(Key)`
+  - removed `KDB::open(KeySet, Key)`
+  - removed `KDB::close(Key)`
+  - updated javadoc for `KDB::set(KeySet, Key)`
+  - introduced `KDB::set(KeySet)`
+
+_(Michael Tucek)_
 
 ### <<Binding2>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -113,8 +113,8 @@ you up to date with the multi-language support provided by Elektra.
   - Removed `KDB::open(KeySet, Key)`, please update usages by calling `KDB::open(KeySet)` instead.
   - Removed `KDB::close(Key)`, please update usages by calling `KDB::close()` instead.
   - Updated javadoc for `KDB::set(KeySet, Key)`, better explaining the relevance of the second argument. Please review API usage to consider the more elaborated explanation of how Elektra uses this argument's value.
-- Introduced `KeySet::remove(Key)`
-- Removed `KeySet::lookup(Key, int)` and `KeySet::lookup(String, int)` as well as accompanying flag definitions `KeySet::KDB_O_NONE`, `KeySet::KDB_O_DEL` and `KeySet::KDB_O_POP`. Please use `KeySet::lookup(Key)` and `KeySet::lookup(String)` instead. Instead of `KeySet::KDB_O_DEL`, please consider using `Key::release`. The proper replacement for `KeySet::KDB_O_POP` is `KeySet::remove(Key)`.
+- Introduced `KeySet::remove(Key)` and `KeySet::remove(String)`
+- Removed `KeySet::lookup(Key, int)` and `KeySet::lookup(String, int)` as well as accompanying flag definitions `KeySet::KDB_O_NONE`, `KeySet::KDB_O_DEL` and `KeySet::KDB_O_POP`. Please use `KeySet::lookup(Key)` and `KeySet::lookup(String)` instead. Instead of `KeySet::KDB_O_DEL`, please consider using `Key::release`. The proper replacement for `KeySet::KDB_O_POP` is `KeySet::remove(Key)` or `KeySet::remove(String)`.
 
 _(Michael Tucek)_
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -107,13 +107,12 @@ you up to date with the multi-language support provided by Elektra.
 
 ### Java binding
 
-- Renamed zero argument static factory method `Key::createNameless` to `Key::create` and narrowed accessibility to protected, effectively removing it from the public API.
+- Renamed zero argument static factory method `Key::createNameless` to `Key::create`.
 - Removed signatures publishing the error key based error handling approach to the Java binding consumer. Such arguments are now created implicitly by the binding. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Affected signatures:
   - removed `KDB::open(Key)`
   - removed `KDB::open(KeySet, Key)`
   - removed `KDB::close(Key)`
   - updated javadoc for `KDB::set(KeySet, Key)`
-  - introduced `KDB::set(KeySet)`
 
 _(Michael Tucek)_
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -107,12 +107,12 @@ you up to date with the multi-language support provided by Elektra.
 
 ### Java binding
 
-- Renamed zero argument static factory method `Key::createNameless` to `Key::create`.
+- Renamed zero argument static factory method `Key::createNameless` to `Key::create`. To migrate to this change, just update calling code to use the new method name.
 - Removed signatures publishing the error key based error handling approach to the Java binding consumer. Such arguments are now created implicitly by the binding. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Affected signatures:
-  - removed `KDB::open(Key)`
-  - removed `KDB::open(KeySet, Key)`
-  - removed `KDB::close(Key)`
-  - updated javadoc for `KDB::set(KeySet, Key)`
+  - Removed `KDB::open(Key)`, please update usages by calling `KDB::open()` instead.
+  - removed `KDB::open(KeySet, Key)`, please update usages by calling `KDB::open(KeySet)` instead.
+  - removed `KDB::close(Key)`, please update usages by calling `KDB::close()` instead.
+  - Updated javadoc for `KDB::set(KeySet, Key)`, better explaining the relevance of the second argument. Please review API usage to consider the more elaborated explanation of how Elektra uses this argument's value.
 
 _(Michael Tucek)_
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -110,9 +110,11 @@ you up to date with the multi-language support provided by Elektra.
 - Renamed zero argument static factory method `Key::createNameless` to `Key::create`. To migrate to this change, just update calling code to use the new method name.
 - Removed signatures publishing the error key based error handling approach to the Java binding consumer. Such arguments are now created implicitly by the binding. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Affected signatures:
   - Removed `KDB::open(Key)`, please update usages by calling `KDB::open()` instead.
-  - removed `KDB::open(KeySet, Key)`, please update usages by calling `KDB::open(KeySet)` instead.
-  - removed `KDB::close(Key)`, please update usages by calling `KDB::close()` instead.
+  - Removed `KDB::open(KeySet, Key)`, please update usages by calling `KDB::open(KeySet)` instead.
+  - Removed `KDB::close(Key)`, please update usages by calling `KDB::close()` instead.
   - Updated javadoc for `KDB::set(KeySet, Key)`, better explaining the relevance of the second argument. Please review API usage to consider the more elaborated explanation of how Elektra uses this argument's value.
+- Introduced `KeySet::remove(Key)`
+- Removed `KeySet::lookup(Key, int)` and `KeySet::lookup(String, int)` as well as accompanying flag definitions `KeySet::KDB_O_NONE`, `KeySet::KDB_O_DEL` and `KeySet::KDB_O_POP`. Please use `KeySet::lookup(Key)` and `KeySet::lookup(String)` instead. Instead of `KeySet::KDB_O_DEL`, please consider using `Key::release`. The proper replacement for `KeySet::KDB_O_POP` is `KeySet::remove(Key)`.
 
 _(Michael Tucek)_
 
--- a/doc/tutorials/java-kdb.md
+++ b/doc/tutorials/java-kdb.md
@@ -15,7 +15,6 @@ try (KDB kdb = KDB.open()) {
     // code to manipulate keys
 } catch (KDB.KDBException e) {
     e.printStackTrace();
-    e.releaseErrorKey(); // optional clean-up
 }
 ```
 
@@ -27,7 +26,8 @@ If an error occurs, detailed information can be obtained from the thrown `KDBExc
 
 There are 3 kinds of native resources having to be cleaned up properly to prevent memory leaks: `KeySet`, `Key` and `KDB` or rather their backing native key set, key or KDB session.
 
-Fortunately the only resource you are strictly required to release is `KDB` via either `KDB::close` or `try-with-resouces`. For `KeySet` and `Key` the garbage collector cleans them up automatically using a [`Cleaner`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html). Nonetheless, it might be a good idea to clean up these native resources as soon as they are no longer needed, since we do not have any control of when or if the garbage collector cleans up after us.
+Fortunately the only resource you are strictly required to release is `KDB` via either `KDB::close` or `try-with-resouces`. For `KeySet` and `Key` the garbage collector cleans them up automatically using a [`Cleaner`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html)
+In situations where memory is very constrained, it might be a good idea to clean up these native resources as soon as they are no longer needed, since we do not have any control of when or if the garbage collector cleans up after us. For this use-case, the follwing functions are provided: `KeySet::release`, `Key::release` and `KDBException::releaseErrorKey``.
 
 ## Fetching keys
 
@@ -58,7 +58,7 @@ kdb set user:/my/presaved/key it_works!
 
 `value` would equals `it_works!`.
 
-Afterward, we may clean-up no longer needed native resources:
+Afterwards, we may clean-up no longer needed native resources:
 
 ```
 parentKey.release();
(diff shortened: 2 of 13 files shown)


tucek:
Please can somebody explain the relevance of `parentKey` when calling `kdbSet`?

The documentation says:
https://github.com/ElektraInitiative/libelektra/blob/e3a7b8da66d4c883e01aef617e2c1fbc359cf751/src/libs/elektra/kdb.c#L70-L75

Changes outside of the specified parent key get saved. I can not make any sense of this parameter from a callers perspective... ?!

tucek:
failing check is for broken links not related to this PR and therefore can be ignored

kodebach:
The documentation for `kdbSet` is pretty bad. Basically what it wants to say is:
- You must call `kdbGet` before `kdbSet`
- Because `kdbGet` may return keys that are not below `parentKey`, you must ensure these keys are still present for `kdbSet`, even though they are above `parentKey`.

It sounds a bit like `parentKey` is completely useless, but it's not. The issue is just that we can only read/write a complete mountpoint (*). So the best we can do, is use `parentKey` to find the minimal set of mountpoints we need to read/write.

For example, if you just have a mountpoint at `system:/foo`, but use `system:/foo/bar` as `parentKey`, we will still read/write all of `system:/foo`. Moreover, if you didn't create any custom mountpoints, all of `system:/` will be read/written.

With #3693 things will be much more clearly defined and hopefully the documentation (currently still WIP) will be a lot more clear in the end.

---

(*) In theory a partial read would be possible. A partial write would be a lot harder for file-based backends, but e.g. a database backend should in theory be able to support partial writes quite easily. However, these are all just ideas for the far future. Far beyond 1.0

tucek:
> The documentation for `kdbSet` is pretty bad. Basically what it wants to say is:
> 
> * You must call `kdbGet` before `kdbSet`
> * Because `kdbGet` may return keys that are not below `parentKey`, you must ensure these keys are still present for `kdbSet`, even though they are above `parentKey`.
> 
> It sounds a bit like `parentKey` is completely useless, but it's not. The issue is just that we can only read/write a complete mountpoint (*). So the best we can do, is use `parentKey` to find the minimal set of mountpoints we need to read/write.
> 
> For example, if you just have a mountpoint at `system:/foo`, but use `system:/foo/bar` as `parentKey`, we will still read/write all of `system:/foo`. Moreover, if you didn't create any custom mountpoints, all of `system:/` will be read/written.
> 
> With #3693 things will be much more clearly defined and hopefully the documentation (currently still WIP) will be a lot more clear in the end.
> 
> (*) In theory a partial read would be possible. A partial write would be a lot harder for file-based backends, but e.g. a database backend should in theory be able to support partial writes quite easily. However, these are all just ideas for the far future. Far beyond 1.0

So, if i'm understanding correctly, the parent key for `kdbSet` is a kind of optimization in case a key set, containing keys for multiple mountpoints, is provided, to only write the to the affected mountpoints?

Example:
`system:/foo` (mountpoint)
`system:/bar` (mountpoint)

Now someone calls `kdbGet` with parent key `system:/` and adds the key `system:/foo/a/b`, `system:/foo/c` and `system:/bar/d` to the returned key set.
* Calling `kdbSet` with parent key `system:/` would update both mountpoints (`system:/foo` and `system:/bar`) according to the specified key set an therefore adds `system:/foo/a/b`, `system:/foo/c` and `system:/bar/d` to the key database.
* Calling `kdbSet` with parent key `system:/foo` would write to mountpoint `system:/foo` only, ignoring `system:/bar/d` and adding adds `system:/foo/a/b` and `system:/foo/c` to the key database
* Calling `kdbSet` with parent key `system:/foo/a` would write to mountpoint `system:/foo` only, ignoring `system:/bar/d` and adding not only `system:/foo/a/b` but also `system:/foo/c` to the key database

Am i understanding this correctly? And if I am, i would argue that this optimization ist rather special since the caller should neither have to know, not to care about mountpoints at all. One thing is that you may get more keys by `kdbGet`, than you've asked for and that you have to pass the whole key set back to `kdbSet`. But to what's the chance an API caller specifies `system:/foo/a` as parent key for `kdbSet` the key set was fetched with `system:/`?

kodebach:
You understood the mechanism correctly. I guess technically you could call it an optimization. But at least for `kdbGet` I wouldn't say it is an additional feature that we can do partial reads of the KDB (i.e. not all mountpoints). If the take the `parentKey` in `kdbGet`, we need it in `kdbSet` too, so that we can do the "set"-operation corresponding to the previous "get".

In general it is highly recommended, that you use the same `parentKey` for `kdbGet` and `kdbSet`. It is also recommended to use the most specific `parentKey` possible in `kdbGet`, so using `system:/` is rarely correct.

tucek:
> You understood the mechanism correctly. I guess technically you could call it an optimization. But at least for `kdbGet` I wouldn't say it is an additional feature that we can do partial reads of the KDB (i.e. not all mountpoints). If the take the `parentKey` in `kdbGet`, we need it in `kdbSet` too, so that we can do the "set"-operation corresponding to the previous "get".
> 
> In general it is highly recommended, that you use the same `parentKey` for `kdbGet` and `kdbSet`. It is also recommended to use the most specific `parentKey` possible in `kdbGet`, so using `system:/` is rarely correct.

Ok, thank you. I update the java doc affected by this PR accordingly.

tucek:

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
```diff
@@ -67,52 +41,37 @@
...
 	 */
-	@Nonnull public static KDB open (KeySet contract, Key errorKey) throws KDBException
-	{
-		argNotNull (contract, "KeySet 'contract'");
-		argNotNull (errorKey, "Key 'errorKey'");
-		return checkKDBPointer (Elektra.INSTANCE.kdbOpen (contract.getPointer (), errorKey.getPointer ()), errorKey);
-	}
-
-	@Nonnull private static KDB checkKDBPointer (@Nullable Pointer pointer, Key errorKey) throws KDBException
+	@Nonnull private static KDB openInternal (@Nullable KeySet contract) throws KDBException
```
kodebach, line 63: Why don't we just allow `KDB.open(null)`?

kodebach (commented):
Most of the code LGTM, but I have a few questions

doc/news/_preparation_next_release.md
```diff
@@ -105,7 +105,16 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
 Bindings allow you to utilize Elektra using [various programming languages](https://www.libelektra.org/bindings/readme). This section keeps
 you up to date with the multi-language support provided by Elektra.
 
-### <<Binding1>>
+### Java binding
+
+- Renamed zero argument static factory method `Key::createNameless` to `Key::create`.
+- Removed signatures publishing the error key based error handling approach to the Java binding consumer. Such arguments are now created implicitly by the binding. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Affected signatures:
```
markus2330, line 111: Please write here what API users of these functions would do now.

doc/tutorials/java-kdb.md
```diff
@@ -11,30 +11,17 @@ In order to use `kdb` you need to include the dependency in your project. [Here]
...
 ```java
-var key = Key.createNameless();
-try (KDB kdb = KDB.open(key)) {
+try (KDB kdb = KDB.open()) {
     // code to manipulate keys
 } catch (KDB.KDBException e) {
     e.printStackTrace();
-} finally {
-    key.release(); // optional clean-up
+    e.releaseErrorKey(); // optional clean-up
```
markus2330, line 18: Why would this be necessary?

markus2330 (commented):

markus2330:
> Ok, thank you. I'll update the java doc affected by this PR accordingly.

Please also update the docu of `kdbSet` itself.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
```diff
@@ -67,52 +41,37 @@
 	 *                                  specialization of {@link KDBException}
 	 * @throws KeySetReleasedException  if {@code contract} has already been
 	 *                                  released
-	 * @throws IllegalArgumentException if {@code contract} or {@code errorKey} is
-	 *                                  {@code null}
+	 * @throws IllegalArgumentException if {@code contract} is {@code null}
+	 * @see #open()
 	 */
 	@Nonnull public static KDB open (KeySet contract) throws KDBException
```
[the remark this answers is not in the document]
tucek: All parameters for all functions are deemed not to be allowed to be null because of the package level annotation `@javax.annotation.ParametersAreNonnullByDefault` (see `package-info.java`).
Therefore @Nonnull annotation is only necessary for return values.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
tucek: Not permitting NULL valued arguments in a (public) Java APIs is considered good practice in my professional experience. If you want to provide a `contract`, use the signature providing the opportunity to do so. Otherwise use the signature without the `contract` argument. Since Java does not support optional method arguments, it is common to provide different signatures for different combinations of valid arguments. For more than two optional parameters, Builders are commonly expected to be provided.
Together with static code analysis in conjunction with the relevant annotations, this should lead to a better developer experience, less errors and better code readability.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
```diff
@@ -271,6 +219,7 @@ private KDB (Pointer pointer)
 	 *                                  released
 	 * @throws IllegalArgumentException if {@code keySet} or {@code parentKey} is
 	 *                                  {@code null}
+	 * @see #get(Key)
 	 */
 	public KDB get (KeySet keySet, Key parentKey) throws KDBException
```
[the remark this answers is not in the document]
tucek: The intend for providing both signatures (`KeySet get(Key)` and `KDB get(KeySet, Key)`) is to preserve the ability for re-using an existing key set while also relieving the caller of the need to allocate a target key set beforehand calling `get`. In Java it is common practice to provide a fluent interface whenever possible.
In this case the fluent interface enables chaining multiple `KDB get(KeySet, Key)` and `KDB set(KeySet, Key)` calls. I would argue, that it is better to support this use case than not supporting it, because i don't see any drawbacks resulting from supporting it.
Ultimately it is a design decision: Do we want to provide a fluent API where applicable?.

doc/tutorials/java-kdb.md
tucek: When the cleaner is enabled and working, it is not. It's optional to manually release the native resource, if one wants to take control over when these resources are being released. The example code includes this for completeness.
Do we want to decide to leave out references to native resource release for all code examples?

doc/news/_preparation_next_release.md
tucek: done

tucek:
> > Ok, thank you. I'll update the java doc affected by this PR accordingly.
> 
> Please also update the docu of `kdbSet` itself.

updated doc for `kdbGet` and `kdbSet`

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
kodebach: I get that the `KDB.open()` overload without arguments is useful, but why have the additional copy of `KDB.open(KeySet)`?

This just makes it more difficult, if you want to wrap `KDB.open` in another function and keep the contract optional. You either need to also create two overloads, or you have to add an useless `if` to call the correct overload when both overloads will call the same implementation anyway.

Take this hypothetical wrapper for example:
```java
// we need to do this
KDB openWrapper(KeySet ​contract) {
    KDB kdb;
    if (contract == null) {
        kdb = KDB.open();
    } else {
        kdb = KDB.open(contract);
    }
    // do something
    return kdb;
​}

// when this would also be an option
KDB openWrapper(KeySet contract) {
   ​KDB kdb = KDB.open(contract);
   ​// do something
   ​return kdb;
}
```

Note: Using `Optional<KeySet>` can make this a bit more readable, but we still have the useless branching and then executing the same thing anyway.

doc/tutorials/java-kdb.md
kodebach: I think the tutorial should show the simplest way of doing things. You could add a hint that a manually releasing is an option. But unless you are very memory constrained, it's probably better to leave the cleanup to the garbage collector.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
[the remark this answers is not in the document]
kodebach: I get why `get(KeySet, Key)` returns `KDB` and not `void`. But I don't really get why I don't think there is much use for a fluent API here.

Chaining like this `kdb.set (ks, parent).get(ks, parent)` could make sense, but I don't see why I would want to call `get` and then immediately `set`. Unless there is something broken somewhere the `set` call would never do anything, because `ks` wasn't modified since `get`.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
tucek: Correct. This is always the problem to consider, in cases when having too few optional parameters to sensibly opt for a builder approach. Regarding the decision for whether to introduce nullable parameters and therefore weaken means of static  analysis of potential null pointer problems, i would opt for going with the separate signatures.
I'm not sure how important (= frequent) the open wrapper use case example is, but one can always write it in a readable way as a one liner using a trinary:
```
@Nonull KDB openWrapper(@Nullable KeySet contract) {
   try (​KDB kdb = contract == null ? KDB.open() : KDB.open(contract)) {
     ​// do something
     ​return kdb;
  }
}
```

However we decide we should do so by establishing a rule, when to allow null valued parameters in the public Java binding API.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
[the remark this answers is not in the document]
tucek: A use case coming to mind would be loading more than one key set for different parent keys into the same or different keysets: `kdb.get(ks, pk1).get(ks, pk2)`

Even if that use case may not come up that frequent, i don't see it hurting the API in any way, like enabling or promoting to misuse it.

Again however we decide, it would definitely make sense to establish a rule when to enable a fluent API and when not to. I would propose to provide a fluent API if here is at least one valid use-case.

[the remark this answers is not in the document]
kodebach: I think the double `get` example could be pretty confusing, but I guess returning `KDB` doesn't hurt and you can still ignore it and use the function as if it returned `void`. So let's just keep the `KDB` return type.

doc/tutorials/java-kdb.md
tucek: I agree and have updated the doc accordingly.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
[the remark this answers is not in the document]
tucek: if there are no further objections (@markus2330 you gave a thumbs-up to the initial comment), i would mark this conversation as resolved...

tucek:
@kodebach @markus2330 I've added the implementation for #3982 to this PR with commit https://github.com/ElektraInitiative/libelektra/pull/3971/commits/bc44a02ab2b24e0fda97ed3bedc9537a6b956c83

Please have a quick look. I would like to merge this PR as soon as possible. Thank you!

src/bindings/jna/libelektra/src/main/java/org/libelektra/KeySet.java
```diff
@@ -263,6 +259,22 @@ public int size ()
...
+	 * Removes the the specified key from key set
+	 *
+	 * @param key Key to remove
+	 * @return True, if the key was found and removed, false otherwise
+	 * @throws KeySetReleasedException  if this {@link KeySet} has already been
+	 *                                  released
+	 * @throws KeyReleasedException     if {@code key} has already been released
+	 * @throws IllegalArgumentException if {@code key} is {@code null}
+	 */
+	@Nonnull public boolean remove (Key key)
```
kodebach, line 272: LGTM, except for two things:

1. We should return `Key` here. The `Key` returned by lookup is (normally) not the same as the `Key` you pass in. Only the name will be the same. This would also match the signature of the standard `Map.remove` method.
2.  I would add a `public Key remove (String name)` version that uses `ksLookupByName` for convenience.

tucek: > 1. We should return `Key` here. The `Key` returned by lookup is (normally) not the same as the `Key` you pass in. Only the name will be the same. This would also match the signature of the standard `Map.remove` method.
The signature of `Map.remove` is slightly different, in taking `Object` instead of `Key` as well as returning null instead of throwing an exception. As long as we do not implement the Map interface, we should not care that much about mimicking its contracts.
I will change the return type and introduce throwing a NoSuchElementException

kodebach: > introduce throwing a NoSuchElementException

I wouldn't do that. I would just return `null`, if nothing was removed.

tucek: done

src/bindings/jna/libelektra/src/main/java/org/libelektra/KDB.java
tucek: As a general rule i would propose not allowing null valued arguments in the public Java binding except for significant use cases. This should then be pointed out (emphasized) in the corresponding method java doc.

src/bindings/jna/libelektra/src/main/java/org/libelektra/KeySet.java
tucek: > > introduce throwing a NoSuchElementException
> 
> I wouldn't do that. I would just return `null`, if nothing was removed. Then a caller can get back the same `boolean` by just doing the `!= null` outside the function.

I don't like returning null values... but it comes down to the following question:
What's the use-case for calling remove when i do not know whether the key is contained within the key set?

tucek: btw. `KeySet::remove(int)` currently also follows a "removes specified key..." semantic. if we go for the "tries..." approach, we may have to update this method...

kodebach: > I don't like returning null values...

Yes, `null` values are annoying to deal with, **but** using exceptions as return values (for control flow) is _much much_ worse IMO.

The semantics of `remove(Key)` are well defined, if the element doesn't exist. The goal here is simply that the given `Key` is not present in the `KeySet` after calling `remove`. Therefore we can just do nothing if the key is not present, throwing a `NoSuchElementException` is not a good solution IMO.

For `remove(int)` things are different. Here we want to explicitly remove the element at index X. The goal is _not_ that there is no key at index X. For that we would have to remove every key from X to the end of the keyset. So for an out-of-bounds index the behaviour is not well-defined.

> i would consider it better and more consistent to return an Optional.

Yes, returning an `Optional<Key>` is also an option. I agree that this would also be more consistent with the rest of the API.

tucek: @kodebach I agree, thank you for taking the time for such thorough feedback! done.

tucek:
I'll wait for the final approval by @markus2330 and will then merge this PR. The following PR #3889 is already rebased to this PR's branch.

doc/news/_preparation_next_release.md
```diff
@@ -105,7 +105,18 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
...
 you up to date with the multi-language support provided by Elektra.
 
-### <<Binding1>>
+### Java binding
+
+- Renamed zero argument static factory method `Key::createNameless` to `Key::create`. To migrate to this change, just update calling code to use the new method name.
+- Removed signatures publishing the error key based error handling approach to the Java binding consumer. Such arguments are now created implicitly by the binding. In case of an exceptional state, appropriate exceptions are thrown. Such exceptions provide access to the underlying key containing warning and error information as meta data. Affected signatures:
+  - Removed `KDB::open(Key)`, please update usages by calling `KDB::open()` instead.
+  - Removed `KDB::open(KeySet, Key)`, please update usages by calling `KDB::open(KeySet)` instead.
+  - Removed `KDB::close(Key)`, please update usages by calling `KDB::close()` instead.
```
markus2330, line 114: How can I get warnings then?

doc/news/_preparation_next_release.md
```diff
@@ -105,7 +105,18 @@ The text below summarizes updates to the [C (and C++)-based libraries](https://w
...
+  - Updated javadoc for `KDB::set(KeySet, Key)`, better explaining the relevance of the second argument. Please review API usage to consider the more elaborated explanation of how Elektra uses this argument's value.
+- Introduced `KeySet::remove(Key)` and `KeySet::remove(String)`
+- Removed `KeySet::lookup(Key, int)` and `KeySet::lookup(String, int)` as well as accompanying flag definitions `KeySet::KDB_O_NONE`, `KeySet::KDB_O_DEL` and `KeySet::KDB_O_POP`. Please use `KeySet::lookup(Key)` and `KeySet::lookup(String)` instead. Instead of `KeySet::KDB_O_DEL`, please consider using `Key::release`. The proper replacement for `KeySet::KDB_O_POP` is `KeySet::remove(Key)` or `KeySet::remove(String)`.
```
markus2330, line 117: :+1:

doc/tutorials/java-kdb.md
```diff
@@ -11,36 +11,23 @@ In order to use `kdb` you need to include the dependency in your project. [Here]
...
 
-Fortunately the only resource you are strictly required to release is `KDB` via either `KDB::close` or `try-with-resouces`. For `KeySet` and `Key` the garbage collector cleans them up automatically using a [`Cleaner`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html). Nonetheless, it might be a good idea to clean up these native resources as soon as they are no longer needed, since we do not have any control of when or if the garbage collector cleans up after us.
+Fortunately the only resource you are strictly required to release is `KDB` via either `KDB::close` or `try-with-resouces`. For `KeySet` and `Key` the garbage collector cleans them up automatically using a [`Cleaner`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html)
+In situations where memory is very constrained, it might be a good idea to clean up these native resources as soon as they are no longer needed, since we do not have any control of when or if the garbage collector cleans up after us. For this use-case, the follwing functions are provided: `KeySet::release`, `Key::release` and `KDBException::releaseErrorKey``.
```
markus2330, line 30:
```suggestion
In situations where memory is very constrained, it might be a good idea to clean up these native resources as soon as they are no longer needed, since we do not have any control of when or if the garbage collector cleans up after us. For this use-case, the following functions are provided: `KeySet::release`, `Key::release` and `KDBException::releaseErrorKey``.
```

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
```diff
@@ -237,7 +237,8 @@ int elektraGOptsContractFromStrings (Pointer contractKeySet, long argsSize, Stri
 	 *
 	 * @see #keyDel(Pointer) keyDel()
 	 *
-	 * @param name A valid name to the key, or {@link #KEY_NAME_SIMPLE_INITALIZED_BUT_EMPTY_OBJECT} to get a
+	 * @param name A valid name to the key, or
+	 *             {@link #KEY_NAME_SIMPLE_INITALIZED_BUT_EMPTY_OBJECT} to get a
```
markus2330, line 241: This sentence is confusing, it somehow implies that KEY_NAME_SIMPLE_INITALIZED_BUT_EMPTY_OBJECT is not valid. Do we still need this long alias for the cascading root key?

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
```diff
@@ -354,6 +360,25 @@ int elektraGOptsContractFromStrings (Pointer contractKeySet, long argsSize, Stri
...
+	 * Flag for use with {@link #ksLookup(Pointer, Pointer, int)} and
+	 * {@link #ksLookupByName(Pointer, String, int)}. Represents no option set.
+	 */
+	static final int KDB_O_NONE = 0;
+
+	/**
+	 * Flag for use with {@link #ksLookup(Pointer, Pointer, int)}. Will cause the
+	 * release of parameter {@code key} using {@link #keyDel(Pointer)}.
+	 */
+	static final int KDB_O_DEL = 1;
```
markus2330, line 373: From the release notes I understood that this was removed, not added?

markus2330 (commented):
Many great improvements. I think we can merge and adjust the small things in a soon follow-up PR. @tucek is this okay for you?

doc/news/_preparation_next_release.md
tucek: A very interesting question, i have not considered ;) With these changes, you only get warnings, if there is also an error. (via the KDBException)
To support returning warnings i will introduce a key parameter like `warningsKey`.

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
tucek: It has been removed from the Java binding's public API. But since it is still part of Elektra's `ksLookup` API, it has been moved to the Elektra native library proxy interface (which i will make package private with the next commit ;))

src/bindings/jna/libelektra/src/main/java/org/libelektra/Elektra.java
tucek: Fixed documentation and renamed to CASCADING_ROOT_KEY_NAME

doc/news/_preparation_next_release.md
tucek: > I think we can merge and adjust the small things in a soon follow-up PR. @tucek is this okay for you?

@markus2330 i think all things, even the small ones, should be taken care of now. (see https://github.com/ElektraInitiative/libelektra/pull/3790)
If not, i think we should do it within this PR. (I can always rebase my next PR without any effort)

markus2330 (approved):

Issue #3871, opened 2021-06-05
Introduce ReadOnlyKey [JNA bindings]


tucek:
The following functions return keys which are supposed to be neither modified nor deleted, because the backing key is part of the meta data key set:

- `Key::nextMeta()` backed by `keyNextMeta(Pointer)`
- `Key::currentMeta()` backed by `keyCurrentMeta(Pointer)`
- `Key::getMeta(String)` backed by `keyGetMeta(Pointer, String)`

This issue proposes to enforce this semantic by using Java's type system and introducing a ReadOnlyKey which is extended by Key to prevent the API from allowing modifications to such keys.

Read-only means no methods for modification of name or value should be available. Further the key should not be deleted. This means omitting incrementing the reference counter when constructing the Java `Key` object representing the native key resource as well as decrementing it on release and also not calling `keyDel`. This implies potentially having the native key resource being freed by someone else, but incrementing and decrementing the reference counter without deleting the key may lead to memory leaks. Therefore the documentation will state to only use returned `ReadOnlyKey` very locally and do not retain references outside of the scope of immediate processing of the return value of the method returning it.

Also see https://github.com/ElektraInitiative/libelektra/issues/3824#issuecomment-852912886

To enable iterator based handling of a key's meta data key set, support of `keyMeta(Pointer)` is introduced. The returned key set should prohibit deleting or adding keys from or to the key set. Another requirement is that the name for contained keys are not editable. This is already covered by a key's name being locked when being contained by a key set. The return type for the newly introduced method `Key::getMeta()` will be `FixedKeySet`.

tucek:
@kodebach if you have time, could you maybe do a quick sanity check of the updated description?

kodebach:
Both `keyNextMeta()` and `keyCurrentMeta` will be removed with #3976.

> This means omitting incrementing the reference counter [...]

Internally, a `ReadOnlyKey` should still increment/decrement the reference counter and call `keyDel`. Otherwise you just run into problems. However, the interface should not expose any of these function to the user (e.g. no `release()` method). Calling `keyDel` is totally safe (with the current reference counting), since the metadata keyset will have incremented the reference count. Therefore calling `keyDel` won't do anything until the metadata keyset has decremented the reference count.

The note about not deleting the key in the C documentation is there, because in C you could do this:

```c
Key * m = (Key *) keyGetMeta (k, "meta:/foo"); // cast to get rid of const
keyDecRef (m);
keyDel (m);
[...]

PR #3884, opened 2021-06-07
jni mounting

tucek:
## Basics

These points need to be fulfilled for every PR:

- [ ] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [ ] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [ ] The PR is rebased with current master.

If you have any troubles fulfilling these criteria, please write
about the trouble as comment in the PR. We will help you.
But we cannot accept PRs that do not fulfill the basics.

## Checklist

Check relevant points but **please do not remove entries**.
For docu fixes, spell checking, and similar none of these points below
need to be checked.

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned every code not directly written by me in [THIRD-PARTY-LICENSES](doc/THIRD-PARTY-LICENSES)

## Review

Reviewers will usually check the following:

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

If you are already Elektra developer:

- Add the "work in progress" label if you do not want the PR to be reviewed yet.
- Add the "ready to merge" label **if the basics are fulfilled** and you also
  say that everything is ready to be merged.

PR #3991, opened 2021-08-12
#3871 read only keys

tucek:
## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


tucek:
#3871 implemented with commit https://github.com/ElektraInitiative/libelektra/pull/3991/commits/7f646456af15b3654ed496f5d9452674847bea0d

tucek:
This PR's branch is based on PR #3889 (!)

tucek:
> Which part of this PR is actually new? It seems there is a lot of overlap with other Java Binding PRs...

Yes, this is because this PR's branch is based on the branch for PR #3889, which is not yet merged and github comparing it to the upstream master by default. I have combined all changes into one commit to make it more clear: https://github.com/ElektraInitiative/libelektra/pull/3991/commits/c0c5dcdc86d3327ae0d321abb4ef6aacc2b3e476

doc/news/_preparation_next_release.md
```diff
@@ -114,10 +114,30 @@ you up to date with the multi-language support provided by Elektra.
...
+  - Introduced `Key::getBinary()` and `Key::setBinary(byte[])`
+- Fixed example project in `examples/external/java/read-keys-example`
+- Introduced abstraction `ReadOnlyKey` for better reflecting the limitations meta data keys via a type hierarchy, leading to meta data keys are now returned as `ReadOnlyKey`s:
+  - `Key` extends `ReadOnlyKey`
+  - Changed `Key Key::nextMeta()` to `Optional<ReadOnlyKey> Key::nextMeta()`, no longer throwing NoSuchElementException for non-exceptional behavior
+  - Changed `Key Key::currentMeta()` to `ReadOnlyKey Key::currentMeta()`
+  - Changed `Optional<Key> Key::getMeta(String)` to `Optional<ReadOnlyKey> Key::getMeta(String)`
+  - Meta data keys can no longer be manually released
+- `Key` class is now final
+- Removed `Key::incRef`, `Key::decRef` and `Key::getRef`
```
markus2330, line 135:
```suggestion
- Removed `Key::getRef`
```

src/bindings/jna/libelektra/src/main/java/org/libelektra/ReadOnlyKey.java
```diff
@@ -0,0 +1,481 @@
...
+import org.libelektra.exception.KeyException;
+import org.libelektra.exception.KeyReleasedException;
+import org.libelektra.exception.KeyStringValueException;
+
+/**
+ * Read only key representing a native Elektra key providing access to its name
+ * and value
+ *
+ * @apiNote This abstraction is primarily used to represent meta keys being read
+ *          only by definition and cannot contian binary data
```
markus2330, line 21:
```suggestion
 *          only by definition and cannot contain binary data
```

src/bindings/jna/libelektra/src/main/java/org/libelektra/exception/KeyBinaryValueException.java
```diff
@@ -0,0 +1,13 @@
+package org.libelektra.exception;
+
+import org.libelektra.Key;
+
+/**
+ * Indicates a {@link Key key's} underlying native key value is not of type
```
markus2330, line 6:
```suggestion
 * Indicates a {@link Key key's} underlying native key value is not
```

src/bindings/jna/libelektra/src/main/java/org/libelektra/exception/KeySetReleasedException.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Indicates that an already released {@link KeySet} has been accessed
  */
-public class KeySetReleasedException extends RuntimeException
+public class KeySetReleasedException extends IllegalStateException
```
markus2330, line 8: Do we really need our own exception for this? Why would an API user want to catch this?

markus2330 (commented):
Thank you for the PR!

kodebach (commented):
The changes look mostly fine to me, but I have a few concerns.

1. We need to decide what the "read-only" part of `ReadOnlyKey` means. Is it the equivalent of `const Key *` in the C API or is it a different notion of "read-only"?

    IMO a different notion of "read-only" would be better. For example, in C you cannot access metadata of a `const Key * k` in C, because `keyMeta (k)` may create the metadata keyset and then needs to store it in `k`.

   In Java we don't really have these concerns. Here we could use a more sensible approach and say: `ReadOnlyKey` allows _reading_ any part of the key, including name, value and metadata, but _not writing_ anything, no matter what the C API does underneath.

2. I think we should also create interfaces `ReadableKey` and `WriteableKey` and let `ReadOnlyKey` implement `ReadableKey`, while `Key` would also implement `WriteableKey`. Then the API becomes a bit clearer.
    
    Currently e.g. `public Optional<Key> remove (ReadOnlyKey key)` is a bit confusing. You may ask "Why must the key be read-only to remove it?", until you remember that `Key` extends `ReadOnlyKey` (also a bit confusing at first).
    
    With the interfaces, we would have `public Optional<Key> remove (ReadableKey key)`. It makes sense that the key must be readable to delete it, and it is nice that it doesn't have to be writable.

src/bindings/jna/libelektra/src/main/java/org/libelektra/exception/KeySetReleasedException.java
tucek: I think you are right, we could do without KeySetReleasedException and KeyReleasedException, throwind an IllegalStateException instead.

src/bindings/jna/libelektra/src/main/java/org/libelektra/ReadOnlyKey.java
tucek: fixed

doc/news/_preparation_next_release.md
tucek: @markus2330 I do not understand your suggestions to remove mentioning `Key::incRef` and `Key::decRef`. All three functions previously were protected and `Key` was not final. Effectively making it part of the public Java binding API.

src/bindings/jna/libelektra/src/main/java/org/libelektra/exception/KeyBinaryValueException.java
tucek: I would rather leave "of type binary"... as opposed to "of type string" in `KeyStringValueException`... IMHO the value of a key being binary means only being able to assume 2 distinctive states.

tucek:
> The changes look mostly fine to me, but I have a few concerns.
> 
> 1. We need to decide what the "read-only" part of `ReadOnlyKey` means. Is it the equivalent of `const Key *` in the C API or is it a different notion of "read-only"?
>    IMO a different notion of "read-only" would be better. For example, in C you cannot access metadata of a `const Key * k` in C, because `keyMeta (k)` may create the metadata keyset and then needs to store it in `k`.
>    In Java we don't really have these concerns. Here we could use a more sensible approach and say: `ReadOnlyKey` allows _reading_ any part of the key, including name, value and metadata, but _not writing_ anything, no matter what the C API does underneath.

i agree.

> 2. I think we should also create interfaces `ReadableKey` and `WriteableKey` and let `ReadOnlyKey` implement `ReadableKey`, while `Key` would also implement `WriteableKey`. Then the API becomes a bit clearer.
>    Currently e.g. `public Optional<Key> remove (ReadOnlyKey key)` is a bit confusing. You may ask "Why must the key be read-only to remove it?", until you remember that `Key` extends `ReadOnlyKey` (also a bit confusing at first).
>    With the interfaces, we would have `public Optional<Key> remove (ReadableKey key)`. It makes sense that the key must be readable to delete it, and it is nice that it doesn't have to be writable.

Currently `ReadOnlyKey` (as stated in java doc) primarily intended to be used for meta keys. Therefore `ReadOnlyKey` does not provide access to meta data. I currently do not see any use for other keys being of a read only type. (Are there any C API function, apart from meta data, always returning a read only key? (= locked value?))

I agree, that splitting into interfaces sounds nice at first, but i don't think it is really useful: We would need at least `ReadableKey` (without functions for returning meta data), `WritableKey`, some internal interface (e.g.`NativePointerHolder`) declaring `getPointer`, define all key parameters as `<T extends NativePointerHolder & ReadableKey>` or `<T extends NativePointerHolder & ReadableKey>`. I also don't think we need to allow different implementations of the these interfaces. `ReadOnlyKey` and `Key` are just vehicles to represent the native key resources. Therefore i made `Key` final and as of Java 17, i would also make `ReadOnlyKey` sealed. (permitting `Key` of course).

IMHO, it might be more sensible to think about a better naming for `ReadOnlyKey` ;)

tucek:
Feedback implemented in https://github.com/ElektraInitiative/libelektra/pull/3991/commits/7def2a54929881a14cbee4e270d5ca4dc5b9efb7

kodebach:
> Are there any C API function, apart from meta data, always returning a read only key? (= locked value?)

I don't think the C API is important here. You could just use the `ReadableKey` interface as an indicator in some Java API (possibly just inside your application, not even as public API).

> We would need at least `ReadableKey` (without functions for returning meta data), `WritableKey`, some internal interface (e.g.`NativePointerHolder`) declaring `getPointer`

That is true, it's as easy as I thought...

> IMHO, it might be more sensible to think about a better naming for `ReadOnlyKey` ;)

Definitely. Maybe just calling the class `ReadableKey` is the solution. What bugs me most about `ReadOnlyKey` is that it implies that writing is explicitly forbidden, but `Key` (which is writable) is a subclass.

tucek:
> > IMHO, it might be more sensible to think about a better naming for `ReadOnlyKey` ;)
> 
> Definitely. Maybe just calling the class `ReadableKey` is the solution. What bugs me most about `ReadOnlyKey` is that it implies that writing is explicitly forbidden, but `Key` (which is writable) is a subclass.

I like `ReadableKey` and will update the name and improve the docs.
Regarding `Key` being a descendant of `ReadableKey` (after the rename from `ReadOnlyKey`): I don't see a problem there, since all `Key` inherit all properties of `ReadableKey`, which reflects reality. The rename resolves the semantic conflict.

tucek:
rename done in https://github.com/ElektraInitiative/libelektra/pull/3991/commits/bf79065b18c99fc64b88bdc5ee101466b80e3a37

As soon as PR #3889 has been merged, i would like to merge this PR as well, provided there are no further objections from your side. @markus2330 please add final comments on the two unresolved discussions above.

src/tools/kdb/CMakeLists.txt
```diff
@@ -21,7 +21,15 @@ if (BUILD_SHARED)
 	add_executable (kdb $<TARGET_OBJECTS:kdb-objects>)
 	add_dependencies (kdb kdb_gen_templates_generated)
 
-	target_link_libraries (kdb elektra-core elektra-kdb elektratools elektra-opts elektra-merge)
+	find_package (Threads)
+	target_link_libraries (
+		kdb
+		${CMAKE_THREAD_LIBS_INIT}
```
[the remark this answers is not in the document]
mpranj: Why does the kdb tool need to be linked with pthreads? I thought it would be enough to only link `jni` with pthreads.

Maybe @markus2330 knows as it was his suggestion?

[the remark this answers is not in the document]
tucek: See #3884

src/bindings/jna/libelektra/src/main/java/org/libelektra/ReadableKey.java
```diff
@@ -0,0 +1,480 @@
...
+	 * Checks whether this key is direct sub-key of the {@code other} key
+	 *
+	 * @param other Key that is used in check as parent key
+	 * @return Boolean if this key is direct sub-key of other key ("child")
+	 * @throws IllegalStateException    if this or the {@code other}
+	 *                                  {@link ReadableKey} has already been
+	 *                                  released
+	 * @throws IllegalArgumentException if {@code other} is {@code null}
+	 */
+	public boolean isDirectBelow (ReadableKey other)
```
mpranj, line 389: Typo or is the C API naming not suitable?
```suggestion
	public boolean isDirectlyBelow (ReadableKey other)
```

src/tools/kdb/CMakeLists.txt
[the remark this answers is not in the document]
mpranj: It is not really explained there either.

mpranj (approved):
Thank you, great work! :rocket: 

Let's merge it and add any further changes in a new PR.

src/tools/kdb/CMakeLists.txt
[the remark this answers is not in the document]
tucek: That's true, only mentioned it to give some reference, to where it comes from.

src/bindings/jna/libelektra/src/main/java/org/libelektra/ReadableKey.java
tucek: Thank you, i will include this change in PR #4002

tucek:
@mpranj Thank you for merging this PR!

src/tools/kdb/CMakeLists.txt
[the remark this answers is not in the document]
markus2330: Technically it is needed if any library requires it. As the libraries are depending on which plugins are in use we are in a gray area. I think I suggested it because of some segfault we had in JNI which looked like that the missing pthread could be the cause. Iirc it was actually not the case. But linking against it also does not harm.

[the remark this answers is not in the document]
tucek: When I'm remembering correctly, I think it was part of the problem.

PR #4010, opened 2021-08-27
Debian Bullseye

mpranj:


markus2330:
Thank you so much for working on this!

src/plugins/crypto/gcrypt_operations.c
```diff
@@ -370,7 +370,7 @@ int elektraCryptoGcryDecrypt (elektraCryptoHandle * handle, Key * k, Key * error
 	}
 
 	// prepare buffer for plain text output and crypto operations
-	kdb_octet_t * output = elektraCalloc (payloadLen);
+	kdb_octet_t * output = elektraMalloc (payloadLen);
```
markus2330: Is this on purpose? It contradicts what was said in the commit message?

mpranj: Yes it reverts a previous change which was not necessary. See the diff between master an the PR instead of just the one commit.

PR #4047, opened 2021-09-15
highlevel: detect if spec properly mounted and unmodified

### LINKED ISSUE #3998: Help message not printed when high level API with code generation is used
@kodebach I'd like to fix this myself as part of my thesis, but beforehand I'd appreciate your assessment:
Can you take a look at this? Have I configured something wrong or is this a bug/regression?

0. Execute the steps below which follow the tutorial [High-level API (with code-generation)](https://www.libelektra.org/tutorials/high-level-api). 
Note: The steps below use a different signature for `loadConfiguration()` than the tutorial, because the tutorial is outdated (see https://github.com/ElektraInitiative/libelektra/pull/3997/files).

1. Create file test.ini with content:
```
[]
mountpoint = test.ecf

[temp/day]
type = unsigned_short
description = The color temperature the screen should have during daytime.
default = 6500
example = 6500
opt/long = temp-day
opt/arg = required
```

2. Execute `kdb gen -F ni=test.ini highlevel "/sw/example/myapp/#0/current" conf`

3. Create file myapp.c with content:
```c
#include <stdlib.h> 
#include <stdio.h> 
#include <elektra.h> 
#include "conf.h" 
 
int main(int argc, const char * const * argv, const char * const * envp) { 
        exitForSpecload (argc, argv); 
        ElektraError * error = NULL; 
        Elektra * elektra = NULL; 
        int rc = loadConfiguration (&elektra, argc, argv, envp, &error); 
 
        if (rc == -1) 
        { 
            fprintf (stderr, "An error occurred while opening Elektra: %s", elektraErrorDescription (error)); 
            elektraErrorReset (&error); 
            exit (EXIT_FAILURE); 
        } 
 
        if (rc == 1) 
        { 
            // help mode - application was called with '-h' or '--help' 
            // for more information see "Command line options" below 
            printHelpMessage (elektra, NULL, NULL); 
            elektraClose (elektra); 
            exit (EXIT_SUCCESS); 
        } 
} 
```

4. Link and compile

`` cc myapp.c conf.c `pkg-config --cflags --libs elektra-codegen` -I. -o myapp -Wl,-rpath `pkg-config --variable=libdir elektra-codegen` ``

5. Run executable with `--help`

`./myapp --help`

6. Run executable with `-h`

`./myapp -h`

## Expected Result
The help message generated by code generator should be printed including information about the supported command line argument `--temp-day`.

## Actual Result
Nothing is printed.

## System Information

- Elektra Version: 0.9.7
- Operating System: Ubuntu 20.04

qwepoizt:
Fixes #3998 

## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


qwepoizt:
Mentioning @markus2330 as discussed.

src/libs/ease/hash.c
```diff
@@ -0,0 +1,118 @@
...
+	// Duplicate ks, then cut out parentKey and all keys below. These are the ones we take into account for token calculation.
+	KeySet * dupKs = ksDup (ks);
+	KeySet * cutKs = ksCut (dupKs, parentKey);
+
+	/**
+	 * Loop through all keys relevant for token calculation.
+ 	*/
+	Key * currentKey;
+	ksRewind(cutKs);
+	while ((currentKey = ksNext (cutKs)) != NULL) {
```
markus2330, line 63: This uses the old kind of iteration, we should deprecate ksRewind/ksNext more clearly! @lawli3t can you please do that?

Iteration now works via `ksAtCursor`.

We should also have an example how to iterate the whole KeySet.

src/libs/ease/hash.c
```diff
@@ -0,0 +1,118 @@
...
+			ELEKTRA_SET_INTERNAL_ERROR (parentKey, "keyDup() failed!");
+			return false;
+		}
+		/**
+		 * Key namespace is ignored for token calculation (via setting it to cascading).
+		 * Reason: Different callers of this function pass the keys for token calculation using different namespaces,
+		 * but the specification they pass is actually the same.
+		 * (e.g. tools/kdb/gen.cpp passes keys in cascading namespace while src/libs/highlevel/elektra.c passes keys in spec namespace).
+		 */
+		int result = keySetNamespace (cascadingKey, KEY_NS_CASCADING);
```
markus2330, line 75: Is this already the new algorithm?

src/libs/ease/hash.c
```diff
@@ -0,0 +1,118 @@
...
+		// Feed key name into sha_256_write() without NULL terminator (hence -1).
+		// This makes it easier to compare expected results with other sha256 tools.
+		sha_256_write (&sha_256, keyName(cascadingKey), keyGetNameSize(cascadingKey) - 1);
+		// Note: The value of the key itself is not relevant / part of specification. Only the key's name + its metadata!
+
+		KeySet * currentMetaKeys = keyMeta(currentKey);
+		Key * currentMetaKey;
+		ksRewind(currentMetaKeys);
+		// Feed name + values from meta keys into sha_256_write().
+		while ((currentMetaKey = ksNext (currentMetaKeys)) != NULL) {
```
markus2330, line 89: Also here please use new style of iteration.

It is good that you already use `keyMeta`, how did you find it? (keyNextMeta is the old style @lawli3t can you also mark this as deprecated?)

markus2330 (commented):
A looked a bit (even though it is draft, I hope it is okay) and found use of deprecated functions we unfortunately did not mark as deprecated yet.

qwepoizt:
> A looked a bit (even though it is draft, I hope it is okay) and found use of deprecated functions we unfortunately did not mark as deprecated yet.

Yes, thanks, I'm always open for feedback!

src/libs/ease/hash.c
qwepoizt: Yes, this is the algorithm we discussed yesterday!

src/libs/ease/hash.c
qwepoizt: Thanks for pointing that out! I wasn't aware of a new vs. old style of iteration. I picked up ksNext, ksRewind and keyMeta simply by looking at existing code within libelektra.

I'll switch to the new style.

src/libs/ease/hash.c
qwepoizt: Please see https://github.com/ElektraInitiative/libelektra/pull/4047/files/91851070110fad6ab8ba633f6e9ca91b20957d12#r711209833

src/libs/ease/hash.c
qwepoizt: I switched to the new style!

qwepoizt:
Updatee: This is currently blocked by an issue with arrays (see https://github.com/ElektraInitiative/libelektra/issues/3998#issuecomment-922057230).

After fixing the array problem, this should be ready for review. I'll comment again once it is.

qwepoizt:
This was indeed a mistake in my algorithm relating to namespaces!

It is fixed now, but another test is failing, which I'm currently working on.

qwepoizt:
This effectively means that the spec changes when an initially empty array is later non-empty.

The token will then not match anymore.

qwepoizt:
@kodebach Thanks for your helpful input!

@markus2330 I tried to fix this all day, but I can't come up with a working solution.
However, I have a working workaround: Ignore array **parent** items (i.e. `.../format/#`) during specification token calculation. 
I believe this is sensible as:

1. redshift does not use arrays 
2. this bug is hard to fix and the time is probably better spent on other issues, especially #3693.

What do you think?

qwepoizt:
@markus2330 @kodebach Please let's move the discussion for the array issue to the new #4061!

qwepoizt:
@markus2330 @kodebach Please let's continue discussion for this bug in new #4061

qwepoizt:
@markus2330: I'm very happy to report, that - from my point of view - this is now ready to merge!

I've fixed the problems with arrays and some others reported by the build servers.

The 4 unsuccessful checks are caused by:

* 3 broken links, which will soon be updated (https://github.com/ElektraInitiative/libelektra/pull/4058#issuecomment-924129108)
* An unrelated problem introduced in upstream glib (#4062).

src/libs/highlevel/elektra.c
```diff
@@ -303,6 +306,26 @@ void elektraFatalError (Elektra * elektra, ElektraError * fatalError)
...
+ * 
+ * @note:  The returned char array needs to be freed with elektraFree() after usage.
+ * 
+ * @param application The application's name.
+ * @return Pointer to a char. Needs to be freed using elektraFree() after usage.
+ */
+static char * generateSpecProblemErrorMessage( const char * application) {
+	return elektraFormat(
+		"There was a problem with the application's specification. \n\nTo fix this, execute:\n"
+		"\t\"$ sudo kdb rm -r %s\"\n"
```
markus2330, line 321: Why do we recommend to remove everything?

markus2330 (commented):
Thank you for the PR!

src/tools/kdb/gen/highlevel/highlevel.cpp
```diff
@@ -624,6 +626,18 @@ kainjow::mustache::data HighlevelGenTemplate::getTemplateData (const std::string
...
+	// calculate specification token
+	char token[65];
+	kdb::Key parentKeyMaybeWithErrors = kdb::Key(parentKey, KEY_END);
+	kdb_boolean_t success = ckdb::calculateSpecificationToken (token, ks.getKeySet(), parentKeyMaybeWithErrors.getKey());
+	if(!success) {
+		kdb::printWarnings(std::cerr, parentKeyMaybeWithErrors, false, false);
+		kdb::printError(std::cerr, parentKeyMaybeWithErrors, false, false);
+		throw CommandAbortException("Error during calculation of specification token.");
+	}
+	contract.append(kdb::Key("system:/elektra/contract/highlevel/check/spectoken/token", KEY_VALUE, token, KEY_END));
```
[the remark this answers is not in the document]
qwepoizt: Thanks, I agree that this looks much better, the key name was a bit clumsy before.

tests/shell/gen/highlevel/commands.expected.c
```diff
@@ -104,7 +104,9 @@ int loadConfiguration (Elektra ** elektra,
 	KeySet * defaults = embeddedSpec ();
 	
 
-	KeySet * contract = ksNew (2,
+	KeySet * contract = ksNew (4,
+	keyNew ("system:/elektra/contract/highlevel/check/specproperlymounted", KEY_VALUE, "1", KEY_END),
```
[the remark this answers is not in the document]
qwepoizt: You are right, this makes more sense!

src/libs/highlevel/elektra.c
qwepoizt: Actually, you are right: the user should only `sudo kdb rm -r spec:/sw/jonls/...` !

src/libs/ease/hash.c
qwepoizt: Yes, that is correct!

* We only look at spec: keys in the algorithm
* But, when we feed the keyname into the sha-256 API, we want to ignore the namespace prefix. The reason is: When `kdb gen` is used with a specification file as input, these keys are all cascading (even though they are spec: keys). If we don't remove the prefix everytime, the token will not match for `kdb gen` (cascading keys) vs. application startup (spec: keys).

src/libs/highlevel/elektra.c
qwepoizt: Fixed!

tests/shell/gen/highlevel/commands.expected.c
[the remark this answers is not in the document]
qwepoizt: Fixed!

src/tools/kdb/gen/highlevel/highlevel.cpp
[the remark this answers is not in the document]
qwepoizt: Fixed!

tests/ctest/test_hash.c
```diff
@@ -0,0 +1,81 @@
...
+	calculateSpecificationToken (hash_ksWithKeysFromTwoApps, ksWithKeysFromTwoApps, parentKeyForMyApp);
+
+	succeed_if (strcmp (hash_ksWithKeysFroMyApp, hash_ksWithKeysFromTwoApps) == 0,
+		    "Token calculation did not properly cut out irrelevant keys! The hash for two different KeySets that have the same set "
+		    "of Keys below the parentKey should be the same, but it was different!");
+
+	keyDel (parentKeyForMyApp);
+	ksDel (ksOnlyWithKeysFromMyApp);
+	ksDel (ksWithKeysFromTwoApps);
+}
```
markus2330, line 66: Can you please check if key = value gets a different shasum than keyvalue = (and the same for metadata)

src/libs/ease/hash.c
markus2330: I don't understand why the namespace gets removed. I think we should pass everything as unmodified as possible to the sha calculator.

markus2330 (changes requested):
Thank you! 2 (hopefully small) points are yet open for clarification.

tests/ctest/test_hash.c
qwepoizt: I remember we discussed this and I think I am already doing that!

Does https://github.com/ElektraInitiative/libelektra/pull/4047/files/103474af753374944f5d1027cb22c348c1767966#diff-bd783ada118bb1ebb0b45cfca574a755c4dab1299fc78371da4c09ad43f746e0R60 cover your requirement?

src/libs/ease/hash.c
qwepoizt: I think, if my rationale can't be understood from the code + inline comments + further explanations here, I'll have to change the code :)

I'll try to find another solution for the namespace problem!

markus2330: I don't think another solution is needed! I only wonder which other caller has a different namespace? I thought that we only compare specs? Is the idea that this function should be able to be used more general than we currently need? Such information (like that KeySets from any namespace can be compared)) should be in the comments for the function, not somewhere hidden inline.

kodebach: I'm also not sure, you need to ignore the namespace. Even if you want to ignore the namespace, you don't need to `keyDup` the key for that. First you can use the unescaped name instead of the escaped one. Then you could simply do:

```c
// ignore the first byte, which is always the namespace
sha_256_write (&sha_256, keyUnescapedName (currentKey) + 1, keyGetUnescapedNameSize (currentKey) - 2);
```

Using the unescaped name makes it a bit harder to compare against externally calculated hashes. But you could do something similar with `strchr` for the escaped name, by looking for the first `/`. This makes the length calculation a bit more difficult, but it avoid duplicating every key.

kodebach:
@qwepoizt In `kdb gen` we generate a keyset with cascading names and _without_ parentKey prefix. The C++ code calls it simply `spec`. This keyset is used to generate the `embeddedSpec()` function. Here is an example result from the tests:

https://github.com/ElektraInitiative/libelektra/blob/28bbf3e90c1ef13b186acc58d84c1db25155ad13/tests/shell/gen/highlevel/commands.expected.c#L42-L61

So for your solution, you need to remove the parentKey prefix as well as the namespace.

However, I think there is much easier solution. You can simply duplicate the KeySet during `kdb gen` and add the parentKey as a prefix (including the `spec` namespace). Then you should have the exact same keys in `kdb gen` and at runtime. Since `kdb gen` only runs once and at compile-time, the cost of duplicating everything doesn't matter here (if it is even noticeable at all).

qwepoizt:
@markus2330 @kodebach Thank you for your hints!

I've fixed this and am waiting for the checks to run through.

src/libs/ease/hash.c
qwepoizt: Thank you both, I fixed this and the namespace is now not ignored anymore.

src/libs/ease/hash.c
```diff
@@ -0,0 +1,124 @@
...
+		 * See https://github.com/ElektraInitiative/libelektra/issues/4061
+		 */
+		size_t baseNameSize = keyGetBaseNameSize (currentKey);
+		char * kBaseName = elektraMalloc (baseNameSize * sizeof (char));
+		keyGetBaseName (currentKey, kBaseName, baseNameSize);
+		if (strcmp (kBaseName, "#") == 0)
+		{
+			continue;
+		}
+		elektraFree (kBaseName);
```
kodebach, line 85: Shouldn't this much simpler snippet work here too:

```c
if (strcmp (keyBaseName (currentKey), "#") == 0)
{
    continue;
}
```

src/tools/kdb/gen.cpp
```diff
@@ -65,11 +66,30 @@ int GenCommand::execute (Cmdline const & cl)
...
+	 *
+	 * 1. Keys loaded via plugin from file: Here, all keys within the ks will be in the cascading namespace.
+	 * Also, the ks will only contain keys from that file.
+	 * Thus, the parentKeyForTokenCalculation must be in cascading namespace as well.
+	 *
+	 * 2. Keys loaded from KDB: Here, the ks will contain keys from all namespaces (user:, spec: , ...)
+	 * For token calculation, only keys from the spec namespace are relevant.
+	 * Thus, the parentKeyForTokenCalculation must be in spec namespace.
+	 */
+	Key parentKeyForTokenCalculation (parentKeyName, KEY_END);
```
kodebach, line 85: I think you can avoid all of this `parentKeyForTokenCalculation` stuff, by simply requiring that `parentKeyName` starts with `spec:/`. We already check in `highlevel.cpp` that it starts with `spec:/` or `/` (and then generate the other variant). Theoretically other templates might not have this requirement, but for now `kdb gen` only makes sense for a `spec:/` parent, so we might as well enforce this properly.

kodebach (commented):
I've found two more minor things, otherwise I'd say this looks very good.

src/libs/ease/hash.c
qwepoizt: Well spotted, thanks! I wasn't aware this function existed. I've replaced it accordingly!

src/tools/kdb/gen.cpp
qwepoizt: Thanks for your suggestion! Unfortunately, I don't have time to implement and test this change. 

I propose to merge this PR and make any further changes in a new one!

kodebach: Yeah, merging as is is totally fine for me.

mpranj (approved):
Thank you so much, great work! 👍 

Thank you for the review as well.

tests/ctest/test_hash.c
markus2330: This test (test_streaming_aware_of_order) does not cover if an actual KeySet that contains such key and values will get different sha codes or at least if:

```c
hash_to_string ("key", hash);
hash_to_string ("value", hash);
```

```c
hash_to_string ("keyvalue", hash);
```

produce different hashes.

If you cannot easily fix it, please open an issue to track this problem.

qwepoizt: Thanks for the clarification, now I understand! Good idea, I will add such tests in a new PR!

PR #4108, opened 2021-10-19
Test reordering structs

mpranj:
**Do not merge this.**


mpranj:
> 🚀 Did you learn something from this?

In this case we did not make the structs smaller, but we gained 2 bytes of reserved space "for free" instead of having unusable padding.

To really learn something we could benchmark this against Elektra before #3949 was merged. (If that is what you meant?)

markus2330:
> To really learn something we could benchmark this against Elektra before #3949 was merged. (If that is what you meant?)

I was thinking that you created the PR to learn something about reordering.

PR #4208, opened 2021-12-15
[New Backend] Discussion: Mount tooling

kodebach:
**DO NOT MERGE**

This PR contains my suggestions for the new `kdb mount` tooling. (ignore the changes to `doc/dev/mountpoints.md`; just fixes for stuff I missed).

The user-side interface of `kdb mount` is described in the Markdown file `doc/dev/kdb-mount.md`. Apart from maybe the set of `[OPTIONS...]` I'm pretty sure this is the best solution here.

The files `kdbmount.h` and `kdbplacement.h` contain my suggestion for the tooling API that would be used by backend plugins to support `kdb mount`.

The functions in `kdbmount.h` will be needed by every backend plugin that calls other plugins (*). A basic backend like `version` of course doesn't need them.  In most cases `elektraMountResolveAllPlugins` should do the job, but sometimes more granular access could be required, so I would just expose all functions.

(*) In general they could be used by any plugins that calls another plugin. However, I think we should cover this use-case differently, so that `elektraPluginOpen` only happens 

`kdbplacement.h` is the implementation for the `placements` contract of plugins. The idea is that there are different variants for placements, but there is a standardized `basic` variant (what we have now in `infos/placements`). This standarized variant is implemented in `kdbplacement.h`. In future we could add other standard placements to `kdbplacement.h`, however, a backend plugin could also use it's own contract if needed.

Regarding embedded systems: The validation function mentioned in the "Import Mode" docs, is essentially just what already happens in `kdbOpen`, so that shouldn't really impact the binary size. The code needed within `backend` for "Direct Mode" should be too much, but we will need to link against `libelektra-mount` and `libelektra-placement` for it to work. There is probably a way to strip out all this when building for embedded systems, or we could supply dummy libs that have the symbols but only return `NULL` or something like that. However, I will look into putting the mount-logic into a separate `.so` file (e.g. `libelektra-backend-meta.so` for the plugin `libelektra-backend.so`).

@markus2330 Please take a look and let me know, what you think. We can also discuss this on Monday.

### DIFF
--- a/doc/dev/kdb-mount.md
+++ b/doc/dev/kdb-mount.md
@@ -1,0 +1,204 @@
+<!-- TODO: not an appropriate place for this type of docs -->
+
+# `kdb mount` Command
+
+## Basics: Direct Mode
+
+The `kdb mount` command is used to mount a new backend at some mountpoint.
+The basic usage of the command is:
+
+```
+kdb mount [OPTIONS...] <MOUNTPOINT> <BACKEND_PLUGIN> [...]
+```
+
+The arguments are as follows:
+
+- `<MOUNTPOINT>`: the mountpoint (i.e. a key name) at which the new backend shall be mounted
+- `<BACKEND_PLUGIN>`: the backend plugin used for the backend.
+  This is the main plugin responsible for the mountpoint.
+- `[...]`: any arguments after `<BACKEND_PLUGIN>` are passed as is to the backend plugin for further processing.
+  How these options are interpreted is entirely up to the backend plugin and may differ drastically between backend plugins.
+- `[OPTIONS...]`: general options processed by `kdb mount`
+
+This mode of operation is called Direct Mode, because the command-line arguments are processed directly by the backend plugin.
+The `kdb mount` command also supports different modes called Mountpoint Info and Import Mode.
+These are described further below.
+
+### Supported Options
+
+The options supported by `kdb mount` are:
+
+- `-I --import=<IMPORT_PLUGIN>:<IMPORT_FILE>`:
+  Use Import Mode for mounting (see below).
+- `-v --verbose`:
+  Print more information about what is happening.
+- `-d --debug`:
+  Print even more information than `--verbose`.
+  This option automatically implies `--verbose`.
+- `-W --with-recommends`:
+  Instruct the backend plugin to use recommended plugins in addition to required ones.
+  This option will simply be ignored, if the backend plugin doesn't support it.
+  This option is not supported in Import Mode.
+- `-q --quiet`:
+  Print nothing, if the operation was successful.
+  Error information will still be printed.
+- `-C --confirm`:
+  Print what will be changed in the KDB and ask the user for confirmation, before applying the change.
+- `-m --merge`:
+  This option is only supported in Import Mode.
+  Instead of failing when `<MOUNTPOINT>` is already in use, a merge will be attempted.
+- `-s --strategy=<STRATEGY>`:
+  This option is only supported in Import Mode when using `--merge`.
+  Specifies the merge strategy to use to resolve conflicts.
+- `-f --force`:
+  **This option is dangerous.**
+  If `<MOUNTPOINT>` is already in use, the existing backend will be removed and replaced by the newly created one.
+  The operation will still fail, if the backend plugin reports an error, or if the validation in Import Mode fails.
+  It is not possible to create an invalid backend configuration with this option, but replacing existing backends may break applications that relied on it.
+- `-c --config=<KEY>=<VALUE>`:
+  A repeatable option to define config keys that are stored in the `/config` part of the backend configuration.
+  `<KEY>` must be a cascading key and cannot contain the `=` character.
+  This option is not supported in Import Mode.
+- `--config-file=<CONFIG_PLUGIN>:<CONFIG_FILE>`:
+  Defines file `<CONFIG_FILE>` that will be read with the (storage) plugin `<CONFIG_PLUGIN>`.
+  The resulting `KeySet` will be stored in the `/config` part of the backend configuration.
+  This is an alternative to `--config`, which can be used if binary values or keys with `=` characters are needed.
+  This option is not supported in Import Mode.
+
+## Mountpoint Info
+
+The `kdb mount` command can also be used to display information about existing backends.
+For this mode invoke the command without at most one argument:
+
+```
+kdb mount [MOUNTPOINT]
+```
+
+Here `[MOUNTPOINT]` is an optional argument.
+If it is present information about the backend mounted at `[MOUNTPOINT]` will be printed.
+Without the any arguments `kdb mount` prints basic information about all existing backends.
+
+## Import Mode
+
+Some backend plugins may require complex configuration.
+It is not always (easily) possible to express the full scope of such configurations in a series of command-line arguments.
+To solve this problem, `kdb mount` also has an import mode:
+
+```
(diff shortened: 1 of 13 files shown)


doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+
+# `kdb mount` Command
+
+## Basics: Direct Mode
+
+The `kdb mount` command is used to mount a new backend at some mountpoint.
+The basic usage of the command is:
+
+```
+kdb mount [OPTIONS...] <MOUNTPOINT> <BACKEND_PLUGIN> [...]
```
markus2330, line 11: I don't like the "anything might get passed to the backend plugin". We should either do something like `mount (2)` to have key=value options and/or have `kdb mount-backend` scripts which do something backend specific.

It must be properly documented for users which options are taken by which backend, similar to `mount (2)`.

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+
+This mode of operation is called Direct Mode, because the command-line arguments are processed directly by the backend plugin.
+The `kdb mount` command also supports different modes called Mountpoint Info and Import Mode.
+These are described further below.
+
+### Supported Options
+
+The options supported by `kdb mount` are:
+
+- `-I --import=<IMPORT_PLUGIN>:<IMPORT_FILE>`:
```
markus2330, line 31: If we get rid of `-i` interactive, we could use this for import functionality.

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+  Use Import Mode for mounting (see below).
+- `-v --verbose`:
+  Print more information about what is happening.
+- `-d --debug`:
+  Print even more information than `--verbose`.
+  This option automatically implies `--verbose`.
```
markus2330, line 37: If we do something like this, we should do it globally.

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+- `-W --with-recommends`:
+  Instruct the backend plugin to use recommended plugins in addition to required ones.
+  This option will simply be ignored, if the backend plugin doesn't support it.
+  This option is not supported in Import Mode.
+- `-q --quiet`:
+  Print nothing, if the operation was successful.
+  Error information will still be printed.
+- `-C --confirm`:
+  Print what will be changed in the KDB and ask the user for confirmation, before applying the change.
+- `-m --merge`:
```
markus2330, line 47: Isn't `-s` enough?

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+  This option is only supported in Import Mode.
+  Instead of failing when `<MOUNTPOINT>` is already in use, a merge will be attempted.
+- `-s --strategy=<STRATEGY>`:
+  This option is only supported in Import Mode when using `--merge`.
+  Specifies the merge strategy to use to resolve conflicts.
+- `-f --force`:
```
markus2330, line 53: Isn't this also simply one strategy?

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+  **This option is dangerous.**
```
markus2330, line 54: Not much more dangerous than `kdb set` commands, or is it?

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+- If an argument starts with a `/`, it is a config key for the preceding plugin.
+- Otherwise, it specifies a plugin.
+
+Arguments `<plugin>` that specify a plugin result in a new section `plugins/<plugin>` with at least the key `plugins/<plugin>/name = <plugin>` in the backend configuration.
+
+The arguments that start with a `/` must be of the form `<key>=<value>`, where `<key>` must be a cascading key name (`<value>` can be empty).
+The config key `<key>` with the value `<value>` will be assigned to the preceding plugin.
+In other words, this results in the key `plugins/<plugin>/config/<key> = <value>` being added to the backend configuration, where `<plugin>` is the last `<plugin>` that was specified.
+
+An example `kdb mount` invocation could look like this:
```
markus2330, line 133: Can you give a more simple example first? And you first need to introduce the file-based backend so that we know what we talking about.

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+```
+kdb mount \
+  -c general/config/a=1 \
+  --config=general/config/b=2 \
+  --config=general/config/c=3 \
+  /hosts \
+  backend \
+  myhosts \
+  glob \
+  /set/#0= \
```
markus2330, line 144: Why would you write this? What does it mean?

doc/dev/kdb-mount.md
```diff
@@ -0,0 +1,204 @@
...
+
+The arguments that start with a `/` must be of the form `<key>=<value>`, where `<key>` must be a cascading key name (`<value>` can be empty).
+The config key `<key>` with the value `<value>` will be assigned to the preceding plugin.
+In other words, this results in the key `plugins/<plugin>/config/<key> = <value>` being added to the backend configuration, where `<plugin>` is the last `<plugin>` that was specified.
+
+An example `kdb mount` invocation could look like this:
+
+```
+kdb mount \
+  -c general/config/a=1 \
```
markus2330, line 137: Easier to understand if you stay with the same style.

```suggestion
  --config=general/config/a=1 \
```

markus2330 (commented):
I only had a quick look at doc/dev/kdb-mount.md. Thank you for writing user-facing docu first, this is a very good approach.

doc/dev/kdb-mount.md
kodebach: I took this example from `doc/dev/mountpoints.md` and for that I took it from somewhere else (not actually sure anymore where it came from). Yes, it doesn't actually make sense as a config for `glob`. But can't define a config for `glob` like this anyway, because `glob` needs metadata on it's config keys.

In the end this is just some random command so I have something to translate into a section of the KDB below.

kodebach (commented):
I agree that currently the interface between `kdb mount` and a backend plugin is not very well defined. However, I think we should keep this interface as flexible as possible. At the same time there should be a kind of default subset. People who want to create third party backend plugins are not restricted by us, but users of our first-party backend plugins get a better more uniform experience.

IMO the "simply pass part of `argv`" approach is a good idea for the flexible base interface (like the `const void * data` in `mount (2)`). What exactly the default subset would look like, I'm not sure yet. The simple `key=value` approach from `mount (2)` won't work, but I'm sure we can come up with something similarly simple but a bit more powerful.

---

I also have an idea, how to solve the problem of decoupling the tooling support logic from the runtime logic in a backend plugin. I think that, at least in part, this is the reason for your concerns about the "anything might get passed to the backend plugin" idea.

The idea is still a WIP, but it boils down to two core ideas:
1. `libelektra-kdb` (specifically `plugin.c`/`kdbplugin.h`) provides dummy functions. These don't contain code. Instead they are placeholders that implicitly refer to the real functions within `libelektra-mount`/`libelektra-placement` This allows the plugin to delegate to code within `libelektra-mount`/`libelektra-placement` without linking against it.
2. Using a dispatch function that has access to something like `elektraModulesLoad`, plugins that don't use the default functions can also externalize their tooling logic. (This part is more of a WIP, I still need to think about it)

markus2330:
I leave this unanswered for now as nobody has time to work on it.

Issue #3529, opened 2020-10-27
keyNew overhaul


markus2330:
If we decide to keep vaargs for keyNew, we should consider adding something like `KEY_ADD_BASE_NAME` for constructions of special names without the need to know about the escaping rules, e.g.:

`keyNew("system:/", KEY_ADD_BASE_NAME, "elektra", KEY_ADD_BASE_NAME, "mountpoints", KEY_END)` for `system:/elektra/mountpoints`.

kodebach:
Note: I edited the top command, since `user:/%` is not a valid escaped name.

I also thought about and experimented with some alternative APIs:

1. Only allow setting key name in `keyNew`
    ```c
    Key * keyNew (elektraNamespace ns, char * parts, ...);

    // system:/
    keyNew (KEY_NS_SYSTEM, KEY_END);

    // system:/elektra/mountpoints
    keyNew (KEY_NS_SYSTEM, "elektra", "mountpoints", KEY_END);

    // system:/elektra/mountpoints/%
    keyNew (KEY_NS_SYSTEM, "elektra", "mountpoints", "", KEY_END);
    ```
    Pro:
      - Easy to use
      - Hard to misuse

    Con:
      - Hard to transition, `KEY_VALUE`, `KEY_META`, etc. have to translated into explicit function calls.
      - Makes it almost impossible to do `ksNew(10, keyNew (...), keyNew (...), ..., KS_END)` directly.

2. Use unescaped name and size directly:

    ```c
    Key * keyNew (elektraNamespace ns, char * parts, size_t partsSize);

    #define keyNewLiteral (ns, parts) keyNew ((ns), (parts), (sizeof(parts)));

    // system:/
    keyNew (KEY_NS_SYSTEM, NULL, 0);
    keyNew (KEY_NS_SYSTEM, "", 1);
    keyNewLiteral (KEY_NS_SYSTEM, "");

    // system:/elektra/mountpoints
    keyNew (KEY_NS_SYSTEM, "elektra\0mountpoints", 20);
    keyNewLiteral (KEY_NS_SYSTEM, "elektra\0mountpoints");

    // system:/elektra/mountpoints/%
    keyNew (KEY_NS_SYSTEM, "elektra\0mountpoints\0", 21);
    keyNewLiteral (KEY_NS_SYSTEM, "elektra\0mountpoints\0");
    ```
    Pro:
      - Could be extended to still allow `KEY_VALUE`, `KEY_META`, etc.
[...]

PR #4223, opened 2022-01-10
[decisions] extra suggestions for #4201

kodebach:
Suggested changes for #4201


doc/decisions/library_directory_structure.md
```diff
@@ -2,21 +2,67 @@
...
+  libs: one folder per lib, contains headers and source, written in C++
+  plugins: one folder per plugin, only plugins written in C++
+  tools: one folder per tool, only tools written in C++
+  tests: tests written in C++
+ _: {
+         folders for other languages,
+         "binding" is always the folder for the binding to the C API,
+         rest depends on the language
+    }
+```
```
kodebach, line 47: @lawli3t @tucek Some suggestions for the Rust and Java parts of the directory tree would be appreciated.

doc/decisions/elektra_prefix.md
```diff
@@ -29,6 +32,7 @@ Rename all functions to start with elektra.
 
 - Changes in basically every application and tool, but this is automated
   with a refactoring tool @kodebach writes.
+- It is not guaranteed that the code can be compiled with every C99 compiler, because we will not always be below the 31 character limit (e.g. ELEKTRA_ERROR_VALIDATION_SYNTACTIC).
```
markus2330, line 35: Where did you read about this problem? As I understood it, the compilers might only fail if the first 31 characters are identical.

doc/decisions/header_include.md
```diff
@@ -0,0 +1,116 @@
...
+  #include "../otherlib/header.h"
+  ```
+- Including Elektra library headers from a plugin, is done with:
+  ```c
+  #include <elektra/thelib/header.h>
+  ```
+  In other words, plugins are treated as external and must use the same rules an application would also use.
+- Only including system headers and headers from external third-party libraries is done with:
+  ```c
+  #include <external.h>
```
markus2330, line 99: As external users are the most important case to consider we should be also clear that everything they include will be in a sub-folder "elektra". I think we do not have a decision about that yet?

```suggestion
  #include <elektra/external.h>
```

doc/decisions/header_include.md
```diff
@@ -0,0 +1,116 @@
...
+  ```
+
+## Rationale
+
+See also considered alternatives.
```
markus2330, line 104: To put everything under "elektra" has a clear rationale:

```suggestion
We avoid conflicts when header files with the same file name exist.
```

It could be clearer what the whole purpose of this decision is. Why do we want relative includes?

markus2330 (commented):
Thank you for creating the PR! I only added a few quick comments.

doc/decisions/elektra_prefix.md
kodebach: Yes, the standard allows a compiler to ignore any character after the 31st. Therefore, if two identifiers have the same 31 character prefix, they may be treated as identical.

I will rewrite the explanation in the decision, but essentially the problem is that we would have to ensure that all our symbols have a difference in the first 31 characters. For example, we couldn't use e.g. `ELEKTRA_WARNING_VALIDATION_SYTACTIC_LEXER` and `ELEKTRA_WARNING_VALIDATION_SYTACTIC_PARSER`.

doc/decisions/header_include.md
kodebach: I will clarify but the main reasons are:
1. With the new directory structure using relative imports is much easier and makes it clearer which headers are part of the same library.
2. I remember having many problems with mismatched include paths when working on the High-level API. If you use `<>` either the CMake setup gets complicated, or the user must pass the correct directories to their compiler. With `""` it works as long as relative paths are preserved during install.

doc/decisions/header_include.md
kodebach: The `elektra` subfolder is implicitly defined in `header_file_structure.md`. I will add a more explicit note too.

doc/decisions/creating_keysets.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Rust
+
+For example, `libelektra-rust` (targeted at Rust) could provide:
+
+```rust
+trait KeySet {
+    fn new<I: ExactSizeIterator<Item = Key>>(keys: I) -> Self;
+}
+```
```
kodebach, line 54: @lawli3t please check that this Rust API makes sense (at least in theory, it doesn't have to be implemented exactly like this)

kodebach:
@markus2330 I added two more decisions for `keyNew`/`ksNew`, please have a look. Also, if everything is in order, we should merge this PR ASAP and go back to the main #4201.

doc/decisions/private_api.md
```diff
@@ -0,0 +1,52 @@
...
+  in the public API.
+- Plugins can define compatible ranges of core API version via the contract. This version is checked during `elektraPluginOpen` by `libelektra-kdb`.
+
+## Rationale
+
+- allows various users (plugins, applications) to link to (more or less) exactly what they need
+- allows separate versioning for every library
+- allows alternative implementation of parts of Elektra, e.g. a libcore written in Rust
+- facilitates code reuse between plugins
+- The `extern const int` API version can be used for compatibility checks.
```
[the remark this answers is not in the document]
kodebach: But `system:/elektra/version/constants/KDB_VERSION` does not and cannot come from `libelektra-core`. The core library has no knowledge of the KDB and therefore cannot add a Key to it.

Also the name `KDB_VERSION` clearly suggests that this is the version of `libelektra-kdb` and not `libelektra-core`.

doc/decisions/simplify_api.md
```diff
@@ -58,12 +57,44 @@ Remove:
...
-- keyGetNameSize -> keyNameSize
+- keyName -> keyEscapedName ([Escaped Name](escaped_name.md))
+- keyGetNameSize -> keyEscapedNameSize
+- keyGetUnescapedNameSize -> keyNameSize
+- elektraKsPopAtCursor -> ksRemove (also need to make public)
+- keyCmp -> keyCompareName
+
+Merge:
+
+- keyIsBelow: merge with keyIsBelowOrSame, keyIsDirectlyBelow
```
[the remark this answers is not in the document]
kodebach: Yes, I first tried to add more features, but 1) they'll never be used 2) the current features can be done with a single `memcmp` and a few extra checks.

doc/decisions/simplify_api.md
```diff
@@ -74,23 +105,50 @@ Make private:
...
+- elektraFormat;
+
+Move/Add to `libelektra-operations` (actual name TBD):
+
 - ksDeepDup
 - ksGetAlloc
 - ksInit
 - keyGetRef
+- keyNeedSync
+- keyGetNextPart: new API
```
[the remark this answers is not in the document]
kodebach: > I think exposing the "int" cursor and let the user do the arithmetic is simply the best

I'm not sure what you mean by "exposing the int cursor"... Even if we add the function you can still use `keyName` and iterate over the key manually.

The function just makes it easier for new users.

> It is also easier for bindings.

It won't be part of the core API, so there doesn't have to be a binding for this.

> it should be a separate decision

I will move it

doc/decisions/creating_keys.md
```diff
@@ -0,0 +1,145 @@
...
+
+`libelektra-core` only provides the absolute minimal API to create a `Key *`:
+
+```c
+/**
+ * Allocates a new key with the name "/" and returns it.
+ *
+ * The key will have no value and no metadata.
+ */
+Key * keyCreate (void);
```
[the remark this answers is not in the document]
markus2330: > this is what users should use. 

You mean C users? If everyone should use the other form, it should be in the core.

Anyway, I don't see a reason for having a keyNew without arguments because we decided that every key must have a name as invariant. So the minimal form is keyNew (const char* key);

doc/decisions/creating_keys.md
```diff
@@ -0,0 +1,145 @@
...
+## Constraints
+
+- There should be a way to create a `Key *` with name, value and metadata in a single expression.
+
+  This is needed to allow creating a full `KeySet *` with a single expression, which in turn is very useful e.g. when initializing default configurations.
+
+- Using the escaped name in the constructor function should be avoided, since the unescaping process can be expensive.
+- Ideally, creating a `Key *` without metadata should use at most 3 `malloc`s and 2 `memcpy` to:
+
+  1. `malloc` to allocate the `struct _Key`.
```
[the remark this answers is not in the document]
markus2330: > There may of course be optimizations for smaller values.

Which is a common case.

> so that even creating multiple keys only calls malloc once.

I don't think this is relevant. We have mmap for that.

doc/decisions/private_api.md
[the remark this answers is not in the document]
markus2330: I think we should have clear restrictions which libs can be used with which versions of our other libs. `kdb`, `mmap` and `core` always must have the same version, so no separate public identifiers are needed. (Maybe some other identifiers so that we can check ourselves if libs are consistent)

doc/decisions/creating_keys.md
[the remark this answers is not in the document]
kodebach: > > There may of course be optimizations for smaller values.
> 
> Which is a common case.

I agree that optimizations are useful and that small key names/values are a common case. However, this is entirely irrelevant here. The constraint is supposed to describe the general case that holds **for all** cases. In other words it is an upper bound. (*)



> > so that even creating multiple keys only calls malloc once.
> 
> I don't think this is relevant. We have mmap for that.

Yes, `mmapstorage` does something similar, but arena allocation is more general. `mmapstorage` only works when you load from a config file. But arena allocation could be used for keys created at runtime too. We would of course need benchmarks to prove this, but I in my experience there is a lot of code that calls `keyNew`/`ksDel` and then calls `keyDel`/`ksDel` shortly afterwards. All of these currently are `malloc`/`free` calls. With arena allocation we could minimize that. It is of course a very complicated topic and not worth doing without intensive benchmarks, but it was just hypothetical.

---

(*) It's not a hard bound, however. The bound may be violated, _if_ there is a good technical reason (e.g. adding another malloc/memcpy makes the code a lot simpler and benchmarks show there is no significant impact). That is what I meant with "ideally". I'll reformulate this part to make the intentions more clear.

doc/decisions/private_api.md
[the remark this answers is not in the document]
kodebach: 1. I'd would use the name `ELEKTRA_VERSION`, if it is not just for `libelektra-kdb`.
2. The fact that the `system:/elektra` key can only be used in `libelektra-kdb`, global plugins and applications (i.e. the parts that have access to the full KDB), still stands. Only a `extern const int` in `libelektra-core` can be accessed everywhere and is guaranteed to come from the library code (`.so`) and from some header that may be edited/replaced.

If your concern is just about the use of 3 separate symbols, we could combine the whole thing into a single symbol:

```c
const uint32_t ELEKTRA_VERSION = 1002003; // -> 1.2.3
```

This allows more than 1000 major versions each with 1000 minor versions, which in turn could have 1000 patch versions each. That should be enough and would only be one single symbol.

For convenience we could define the macros:

```c
#define ELEKTRA_MAJOR_VERSION = (uint32_t)((ELEKTRA_VERSION) / 1000000);
#define ELEKTRA_MINOR_VERSION = (uint32_t)(((ELEKTRA_VERSION) / 1000) % 1000);
#define ELEKTRA_PATCH_VERSION = (uint32_t)((ELEKTRA_VERSION) % 1000);
```

doc/decisions/private_api.md
```diff
@@ -0,0 +1,52 @@
+# Library Split
```
markus2330, line 1:
```suggestion
# Private API
```

doc/decisions/private_api.md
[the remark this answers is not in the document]
markus2330: No, my concern is that we are getting lost in irrelevant stuff. Nobody asked for something like this, so why bother? In particular I do not see how it supports this decision or solves the problem written in the beginning of the decision.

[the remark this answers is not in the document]
markus2330: If your proposal is actually to rename KDB_VERSION (in kdb.h) to ELEKTRA_VERSION and use const uint32_t instead of macros, I am all in favor but this is a very different decision. The same for KDB_PATH_SEPARATOR, KDB_PATH_ESCAPE, KEY_NULL etc.

About the number of the symbols: please do a test how much these symbols cost in terms of library size. Guessing that it could be expensive and then do optimizations based on guesses doesn't sound very scientific ;)

doc/decisions/creating_keys.md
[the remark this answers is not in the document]
markus2330: > this is entirely irrelevant here

Based on which criteria is this irrelevant? I thought we are discussing to remove the arguments of keyNew? The reason of the current arguments of keyNew is that keyNew could allocate name+value in one malloc (I unfortunately did not write decisions about this but luckily we write them now).

> there is a lot of code that calls keyNew/ksNew [fixed typo] and then calls keyDel/ksDel

Yes, I agree this is a problem, especially because at the moment the whole unescaping needs to be done for every keyNew. I looked at a few places and one problems seems to be that only ksLookup provides an API where a key name can be directly given. This could be solved by a KeyName struct (which I do not like) but it could also be solved by making keyNew cheaper. In previous benchmarks basically only (un)escaping was expensive not the malloc.

Ideally, only one keyNew for the parentKey is needed and everything else is mmaped anyway.

doc/decisions/private_api.md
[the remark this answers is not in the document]
kodebach: > No, my concern is that we are getting lost in irrelevant stuff. Nobody asked for something like this, so why bother? In particular I do not see how it supports this decision or solves the problem written in the beginning of the decision.

The problem this constant solves is that currently there is no way of knowing what version of the API is actually implemented by a `libelektra-core.so` file. Like I also explained [above](https://github.com/ElektraInitiative/libelektra/pull/4223#discussion_r783356201), the `#define`s we have are not good enough.

> If your proposal is actually to rename KDB_VERSION (in kdb.h) to ELEKTRA_VERSION and use const uint32_t instead of macros, I am all in favor but this is a very different decision.

Purely renaming `KDB_*` to `ELEKTRA_*` is a different decision, yes. The type change might be as well. However, the deciding that some kind of version constant should exist within the `libelektra-core.so` definitely belongs here. This decision is concerned with how and why (some) internal structures/APIs may be accessed from outside of `libelektra-core.so`. This can only be done safely, if we know what version of `libelektra-core.so` is used. This in turn requires some sort of version information within `libelektra-core.so`.

> About the number of the symbols: please do a test how much these symbols cost in terms of library size. Guessing that it could be expensive and then do optimizations based on guesses doesn't sound very scientific ;)

I was never really concerned about the number of symbols. You were the one that [brought it up](https://github.com/ElektraInitiative/libelektra/pull/4223#discussion_r783101589).


---

To sum this up: Can we agree on the following?

1. We replace 

    ```c
    #define KDB_VERSION "@KDB_VERSION@"
    #define KDB_VERSION_MAJOR @KDB_VERSION_MAJOR@
    #define KDB_VERSION_MINOR @KDB_VERSION_MINOR@
    #define KDB_VERSION_PATCH @KDB_VERSION_PATCH@
    ```

    with

    ```c
    extern const uint32_t ELEKTRA_VERSION; // defined in libelektra-core.so
    #define ELEKTRA_MAJOR_VERSION = (uint32_t)((ELEKTRA_VERSION) / 1000000);
    #define ELEKTRA_MINOR_VERSION = (uint32_t)(((ELEKTRA_VERSION) / 1000) % 1000);
    #define ELEKTRA_PATCH_VERSION = (uint32_t)((ELEKTRA_VERSION) % 1000);
    ```
2. I won't make this a separate decision file, because writing a whole document for this very small change seems very silly.

doc/decisions/library_directory_structure.md
markus2330: It is a clean separation but I think it is not so helpful for beginners who have no idea in which language we implemented which part. So what about the other way round (put the programming language in the deepest level of nesting)?

```yaml
src:
  libs: one folder per lib
    core:  for libelektra-core, other libs are separated
       include: only the C headers, as the other headers would be in "bindings"
       src:
          c: src in C
          rust: source in Rust
       tests: tests written in any language
          c: tests in C
          rust: tests in Rust
```

if there is only one implementation language, no c/rust/... subfolders are needed.

doc/decisions/creating_keys.md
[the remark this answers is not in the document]
kodebach: > Based on which criteria is this irrelevant?

The fact that _some_ cases of "creating a key with name and value" can be optimized to use less than 3 `malloc` or less than 2 `memcpy` is irrelevant here, because the decision is about the general case. The general case (arbitrary name, arbitrary value) cannot (without extra precautions) use less than what this decision already states.

> keyNew could allocate name+value in one malloc

In theory, you could even allocate struct+name+value in one malloc. However, this may require changes elsewhere, because `realloc`/`free` will no longer work, if name/value points into the middle of a malloc-buffer. In particular, you need to use some flags (maybe the mmap ones) to indicate which pointers are from `malloc` and need to be `free`d.

> his could be solved by a KeyName struct (which I do not like) but it could also be solved by making keyNew cheaper. In previous benchmarks basically only (un)escaping was expensive not the malloc.

I agree, making `keyNew` cheaper is the main goal.

---

To sum up:

I will rewrite this part of the decision to make it more clear that

1. `keyNew` should be as cheap as possible, because it may be called very often.
2. The `keyNew` API should be designed such that `malloc` and `memcpy` can be used as little as possible.

doc/decisions/library_directory_structure.md
kodebach: I don't like the separate `include` folder, it just complicates things. Having the header next to the code file makes browsing the repo much easier and `#include "foo.h"` in `foo.c` is simpler than `#include <foo.h>` and hope include paths are set correctly.

Putting the language at the bottom also makes it much harder to have per-language CMake code. In my example, `src/rust/CMakeLists.txt` would do all the checks for Rust (i.e. is `cargo` installed, do we have the required Rust version, etc.). Only if those checks succeed, will the subfolders be included.
With your system you have to (at the very least) call some kind of check function in every `.../rust/CMakeLists.txt`.

Probably more importantly most languages won't actually use CMake directly as the build tool. Java uses Gradle, Rust uses Cargo, etc. If there isn't a single root directory for all Java/Rust code, you need to setup multiple separate projects for these build tools instead of just a single one with sub-projects/modules.

> it is not so helpful for beginners who have no idea in which language we implemented which part

This could simply be solved by adding a `src/README.md` file similar to what we have now in `src/plugins/README.md`.

I also think my structure has a different advantage for beginners. Your version is better for "I want to know about the code of `libelektra-core.so`". Mine is easier for "I know Rust, and what to contribute to Elektra". In your version a beginner would have to check every library, plugin, tool, etc. for a `rust` folder. In my structure, they just open `src/rust` and see what is there.

If you really want the language at the bottom, I would suggest:

```yaml
src:
 libs:
  core-c: C implementation of libelektra-core.so
  core-rust: Rust implementation of libelektra-core.so
  opts: libelektra-opts.so has only a single implementation, so no suffix
  # ...
 # plugins and tools use the same suffix idea as above
 plugins:
 tools:
 bindings:
  rust: Rust bindings
  cpp: C++ bindings
  java: Java bindings
```

However, I still don't like this structure because of the build system (CMake, Gradle, Cargo, ...) disadvantages.

doc/decisions/creating_keys.md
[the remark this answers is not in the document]
kodebach: > You mean C users? If everyone should use the other form, it should be in the core.

Yes, `libelektra-operations.so` (and therefore `keyNew`/`ksNew`) is just for C users. Other implementations and bindings should provide other APIs for suitable for their language on top of `libelektra-core.so` (i.e. on top of `keyCreate`/`ksCreate`).

> Anyway, I don't see a reason for having a keyNew without arguments because we decided that every key must have a name as invariant.

It still has the name `"/"` (cascading root key) which kind of works like a "null" value for key names. It is the simplest possible name and could be optimized to not use `malloc`.

> So the minimal form is keyNew (const char* key);

In C we need the size (see also https://github.com/ElektraInitiative/libelektra/issues/3529#issuecomment-1023616182). Adding a size argument that can result in errors and is just annoying for other languages doesn't seem like a good idea for the core API. Therefore, I decided the `keyCreate (void)` API is the best solution.

kodebach:
@markus2330 This PR was auto-closed, because the target branch `ElektraInitiative:decisions` was deleted when #4201 was merged. Let's finish the discussion here, and then I will open a new PR targeting `master`.

doc/decisions/library_directory_structure.md
lawli3t: I would also very much prefer having everything related to one language in one folder, it would probably mess with cargo as well if everything was scattered among multiple different folders. Going by the recommended [Rust project structure](https://doc.rust-lang.org/cargo/guide/project-layout.html), most of the stuff should be contained in one parent folder as cargo kind of expects you to adhere to the structure (it doesn't necessitate it but it certainly makes things a lot easier, especially in CI)

I think we should be just fine with the binding and libs folder, tests should be contained in the respective folders anyway. The projects contained in those folder should try to adhere to the structure mentioned above.

It also makes sense from a developers point of view because mostly I will be working on one single project and then have all the stuff related to that in one folder (which i can declare as root folder in the IDE). Other setups would be cumbersome to manage and would require specific configuration for every Editor/IDE....

doc/decisions/creating_keysets.md
lawli3t: Not sure about ExactSizeIterator. Theoretically you could lazily generate KeySets where you don't necessarily know the exact size.

Maybe we would need to separate this? Firstly we should implement (FromIterator)[https://doc.rust-lang.org/std/iter/trait.FromIterator.html] which provides lots of syntactic sugar for devs and is the common trait for Classes that can be instantiated from iterators. Additionally I would add

```rust
trait KeySet {
    fn new (keys: &[Key]) -> Self;
}
```

as this is way more liberal in what it accepts.

Although a certain case could be made that we split this up even further into something like this, if we follow [1] strictly

```rust
trait KeySet {
    // default constructor, only default settings (Default trait)
    fn new () -> Self;
    // for specifying the size
    fn with_capacity(size: usize) -> Self;
    // for creating from a collection / list &[Key] because you can then supply a multitude of things (slices, vectors,...)
    fn from_slice(keys: &[Key]) -> Self;
}

impl FromIterator<Key> for KeySet {
    fn from_iter<I: IntoIterator<Item=Key>>(iter: I) -> Self {
        // ....
    }
}
```

We should also implement `IntoIterator` then, as this would give us a free copy constructor and looping and other stuff....

of course that would make it a bit tricky to create a keySet with a specific size and import Keys in one statement, but from what i've seen from other rust libraries this is the general idiom and no problem at all. You just create the KeySet first and then add all the Keys. It also shouldn't matter performancewise.

[1] https://deterministic.space/elegant-apis-in-rust.html#techniques

kodebach: Okay, then I'll change the example to simply `new (keys: &[Key])` and you just implement whatever you think is best.

doc/decisions/private_api.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Considered Alternatives
+
+- keep current situation
+
+## Decision
+
+- Also allow other libraries (e.g. a new `libelektra-operations`) to access to non-public API.
+- Put struct definitions of Key/KeySet in a separate header file, which gets included by parts that need it (see also [Header File Structure](header_file_structure.md)).
+- Define the current API version of the core API as
```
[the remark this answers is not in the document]
markus2330: I don't think that checking consistency of libraries is a problem we need to solve. Package managers do that.

doc/decisions/private_api.md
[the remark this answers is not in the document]
markus2330: > I was never really concerned about the number of symbols. You were the one that https://github.com/ElektraInitiative/libelektra/pull/4223#discussion_r783101589.

Yes, because once I greatly reduced the size of libelektra-core by reducing symbols. Please **first** benchmark what exactly we need to do to not bloat it again, and then let us design a proper API. If the `extern const uint32_t` are small, several one of them is a better API than macros for calculation.

markus2330:
It was probably too optimistic that the discussion would be finished. Please create new PRs after we merged #4241 which we will hopefully merge today, as it will only document what we discuss in the meeting.

doc/decisions/private_api.md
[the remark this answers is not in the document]
kodebach: I guess that is one way to solve this. It doesn't really help, if no package management software is involved. But in those cases it is probably save to assume that the user is advanced enough to check versions themselves.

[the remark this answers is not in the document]
markus2330: I wrote a bit about it in 740ae92665d680bdaaf33aeb2f2672a7854fde95

doc/decisions/private_api.md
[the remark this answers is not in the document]
kodebach: I didn't even know, if such a symbol would be of interest at all. Therefore, starting benchmarks for something that maybe isn't wanted at all, would be a waste of time.

I will do some benchmarks, but in this case, I don't think a benchmark is gonna tell us a lot. A `uint32_t` symbol should always be exactly 4 bytes. With function symbols this is of course different. Whether or not a function is exported affects what optimizations the compiler can do, so the code-size of the function may change drastically. But AFAIK (other than possible alignment reasons) whether or not a `uint32_t` is exported can only ever make a difference of 4 bytes.

doc/decisions/library_directory_structure.md
kodebach: Gradle supports, creating subprojects with an arbitrary directory layout (see [this SO post](https://stackoverflow.com/questions/20807490/defining-subprojects-in-a-deeply-nested-directory-tree)). Adding dependencies to individual gradle project in CMake might be complicated, but I'm not sure this will ever be needed. Our Java code will probably not be called from a non-JVM language. Therefore, we can simply use Gradle to build everything that depends on the Java code.

@lawli3t For Rust, [this](https://github.com/corrosion-rs/corrosion) looks very helpful.

In general for a multi-language project like Elektra, [Bazel](https://bazel.build/) probably would be a better choice than CMake. But I have no experience with Bazel and I certainly don't think anybody is going to rewrite all our CMake code for Bazel.

kodebach: See https://github.com/ElektraInitiative/libelektra/pull/4243/files#r807268499 for further discussion.

markus2330: I like the idea of having this mixed layout but using the whole src folder for Java (Gradle) and Rust (Cargo) while having tools/plugins/libs in their proper place. Still open is if PLUGINS can then actually be used for Java/Rust plugins.

Btw. in theory CMake (like Bazel) can also build other languages (we actually used the add_jar already), practically Java programmers want Gradle and Rust programmers want Cargo. Bazel won't help.

kodebach: > Still open is if PLUGINS can then actually be used for Java/Rust plugins.

Do you mean the CMake variable `PLUGINS`?

markus2330: Exactly.

markus2330: I don't see why passing PLUGINS to Gradle/Cargo could be a big problem. E.g. you can easily extend Cargo.toml from within CMake and pass anything as you want.

> How to pick between alternative implementations?

I am afraid we need a new CMake variable for that.

> How to express dependencies? e.g. a library needs core but doesn't care which implementation

Then you obviously do not need to specify anything as it should be a full replacement.

> or a Rust library needs a library that only exists in C (so cargo doesn't know how to build it)

I don't think this discussion is helpful for @lawli3t, let us discuss problems he actually has.

kodebach: > > How to express dependencies? e.g. a library needs core but doesn't care which implementation
> 
> Then you obviously do not need to specify anything as it should be a full replacement.

Lets's say `libelektra-foo` depends on `core`, but doesn't care about the implementation. What CMake dependencies should `libelektra-foo` now define?

I think we need a dummy target `libelektra-core` that depending on other configuration either depends on `libelektra-core-c` or on `libelektra-core-rust`.

> you can easily extend Cargo.toml from within CMake and pass anything as you want.

I'm not sure what you mean. What we should absolutely avoid (and also remove where it exists already) is something like having a `Cargo.toml.in` that needs to be processed by CMake's `configure_file` first. This makes it very annoying or even impossible to use IDEs that don't know about CMake.

> I don't see why passing PLUGINS to Gradle/Cargo could be a big problem.

The problem is who determines, if all plugin dependencies of a plugin are present. If plugin `foo` needs plugin `bar`, but one is written in C and the other in Java this is not so easy. It's doable, but probably very annoying to implement. CMake needs to ask Gradle/Cargo for all the plugins (and their dependencies) that can be built with these tools. Alternatively, we need to reimplement the dependency checks in Gradle/Cargo and pass the list of enabled plugins. But this still need two-way communication, because Gradle/Cargo could decide to disable a plugin because of a missing dependency, which in turn could lead to CMake having to disable another plugin, etc.

Whatever we decided to do, we need some interoperation between CMake and Gradle/Cargo.

> I don't think this discussion is helpful for lawli3t, let us discuss problems he actually has.

Neither this PR nor this thread are about specific problems somebody had. This is about a general decision. A decision that has very far reaching consequences, so we should carefully consider all possible problems. Once we decide on a directory structure, I don't think anybody will want to invest the time change it again in the near future.

markus2330: > Once we decide on a directory structure, I don't think anybody will want to invest the time change it again in the near future.

I totally agree but I am afraid that we can discuss a lot and it will never work out that way unless shown with a "prototype". Let us see how #4341 works out (and help in the problems coming up there), and then let us continue finding a solution (if there are still structural problems).

kodebach: AFAICT the most current discussion on the actual library structure is in https://github.com/ElektraInitiative/libelektra/pull/4243/files#r851730391. So if you want to add something, please continue the discussion there.

Regarding "it will never work out that way": The misunderstanding is probably because it has been months since the original discussion, but my original comment was exactly in the spirit of "it never works out that way". I was anticipating a potential problem, and asked about it (https://github.com/ElektraInitiative/libelektra/pull/4223#discussion_r809951795, "How to express dependencies?"). Your response was "That's not helpful right now, please discuss the problems that currently exist" (paraphrased). To which I said, "Potential future problems that come from this change are relevant right now, because changing this again is a lot of work" (paraphrased). So I'm not really sure, what you wanted to say here.

Anyway, since #4341 doesn't actually need to change the structure of the whole repo, I'm fine with postponing this discussion until it is merged.

PR #4239, opened 2022-02-13
Java plugin process

### LINKED ISSUE #3891: put JNI plugins in their own processes
As the attaching to existing JVMs has problems on its own (teardown maybe leads to crashes and we could not modify the classpath or other JVM arguments) @kodebach suggests we will instead put each JNI plugin in its own process, as it is currently done in the python plugin. I agree.

As @kodebach wrote in https://github.com/ElektraInitiative/libelektra/pull/3884#issuecomment-858146418
>  If we send the class name as well as the operation (open, get, etc.) to the Java process, you could also use a single process that dispatches to multiple plugin classes. Of course that requires a more advanced IPC method than a simple pipe, because the Java process would essentially act as a server and there would be multiple client instance on the native side.

I don't think this is a good idea, as it has similar drawbacks to attaching to JVMs: plugins would not be able to have their own JVM classpath+arguments. Furthermore, it is not a good idea to optimize without knowing what the bottleneck will be. Furthermore, I would prefer if we optimize libprocess, as then also the Python plugin could benefit from it. As we have many other important issues to attend to, I prefer not to do any optimizations at this point.

kodebach:
## Changes

Adds a new plugin `stdioproc` that can be used to put plugins into a separate process (an alternative to `process`). Also adds first implementation in the JNA binding.

Closes #3891

## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review


kodebach:
@markus2330 I had to create a separate plugin, because `process` and the `pluginprocess` library make various assumptions that cannot easily be met from the Java side. The new `stdioproc` only needs an executable that observes a certain protocol on stdin/stdout. The unit tests show that this could even be achieved via a shell script (although I didn't implement proper parsing of `dump` in shell code).

I also re-implemented the `dump` format in Java. This may not be entirely required (in theory we could call the native code), but this variant is much easier to debug.

markus2330:
> I had to create a separate plugin, because process and the pluginprocess library make various assumptions that cannot easily be met from the Java side.

Can this new plugin be used to replace the old? Afaik the old one wasn't used anyway.

> The new stdioproc

Library or plugin? Maybe we find a better name.

kodebach:
> Can this new plugin be used to replace the old? Afaik the old one wasn't used anyway.

The new plugin is more general, than the old one. It can work with any executable. For example, the `testmod` tests use a simple shell script. If this new approach works well (so far it looks good), we can probably delete the old `jni` plugin.

For the old `process` plugin, things are bit different. It just calls and existing plugin (via the C API) in another process. I have no idea why that could be useful, but if it is needed, the new `stdioproc` cannot be used to replace it directly. To replace `process` with `stdioproc` we would need a simple executable that calls a plugin via the C API.

To replace the `python` plugin (also uses the `pluginprocess` library) with `stdioproc`, we could just write a small Python script similar to the Java app in this PR.

> Library or plugin? Maybe we find a better name.

Plugin. Suggestions for a better name are welcome. The current name comes from "Standard I/O" (or `stdin`/`stdout`) and (external) process.

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,175 @@
...
+## Introduction
+
+This plugin spawns a new process with a user-defined executable and delegates all operations to the new process.
+
+## Usage
+
+Set the config key `app` and the arrays `args/#`, `env/#` to the path of an executable, the arguments that shall be passed and the environment variables to be set.
+
+```
+kdb mount test.dump /tests/stdioproc specload 'app=/usr/bin/pluginproc' 'args=#1' 'args/#0=--load-plugin' 'args/#1=myplugin'
```
markus2330, line 21: Is this ready to review? Is this really specload?

markus2330:
Please put the API changes to another branch, we do not have the time to review and discuss all of this. (@tucek also needs to be involved.) Please focus on allowing JNI plugins to run as robust as possible. We want to release this in February and there should be as little experimental code as possible.

src/plugins/stdioproc/README.md
kodebach: > Is this ready to review?

There aren't any tests which check that running a Java Plugin via `stdioproc` actually works. I've checked it manually and I'm currently working on automatic tests. Also some parts of the docs still need updates. Otherwise this PR should be done.

> Is this really specload?
It is somewhat similar, but
1. serves an entirely different purpose
2. `specload` uses only one-way communication, while `stdioproc` is two-way
3. `specload` only sends a single keyset, `stdioproc` has a more complex protocol

kodebach:
> Please put the API changes to another branch

I can reduce the API changes, but it won't work with zero changes to the `Plugin` interface. In fact, I'm not sure how the `Plugin` interface was ever supposed to work. The `Plugin.getConfig()` method is just wrong. It makes no sense for Plugin to implement it's own `getConfig` method. The `config` keyset is passed to `Plugin.open()` and if it is needed outside this method it should be stored. However, a `getConfig()` method that could return an arbitrary `KeySet` (completely different from the user defined config) makes no sense.

> Please focus on allowing JNI plugins to run as robust as possible.

It should be pretty robust already. I tried to keep the whole protocol as simple as possible. That's also why I used the `dump` format. As long as no binary keys are involved, the whole protocol is human readable and therefore easy to debug.

> We want to release this in February and there should be as little experimental code as possible.

I assume you are talking about the `StdioProcPlugin` class. I will move changes related to this class to a separate PR, but I really think we should merge this before CM as well. The new API is much more Java-esque. In particular, it hides the fact that we use `Key`s to store errors, which is a very weird concept in Java. At the very least we need something like the `ErrorWriter` class, so that errors are added correctly. The old `WhitelistPlugin` only added `number` and `reason` for errors and warnings.

I also want to emphasize that I didn't actually change any existing APIs. I only added new APIs. The old JNI interface is still there.

kodebach:
@markus2330 I have now rewritten this PR to use the existing Java `Plugin` interface with as few changes as possible. The one remaining change is that I removed the broken `getConfig()`. I also didn't add `ErrorWriter` after all. The `Key` class already had `setError`/`addWarning` methods that were almost correct. I updated those instead.

Apart from tests/docs the PR is now fully read for review.

src/bindings/jna/stdioproc/src/main/java/org/libelektra/stdioproc/StdIoProc.java
```diff
@@ -0,0 +1,194 @@
...
+      k.setName(
+          key.getName()
+              .replace(Plugin.JNI_MODULE_CONTRACT_ROOT, "system:/elektra/modules/stdioproc"));
+      contract.append(k);
+    }
+
+    contract.append(Key.create("system:/elektra/modules/stdioproc/exports/open", "1"));
+    contract.append(Key.create("system:/elektra/modules/stdioproc/exports/get", "1"));
+    contract.append(Key.create("system:/elektra/modules/stdioproc/exports/set", "1"));
+    contract.append(Key.create("system:/elektra/modules/stdioproc/exports/close", "1"));
```
kodebach, line 60: I really don't like this bit. Adding at least a `getContract()` would be much better (even better would be also adding `getSupportedOperations()`).

src/plugins/stdioproc/README.md
kodebach: > Is this really specload?

I missed that the line literally says "specload". That's a copy paste error. It's fixed in the next commit.

markus2330:
> The new plugin is more general, than the old one.

Please remove the old plugin process and rename your plugin to be called "process".

> Apart from tests/docs the PR is now fully read for review.

Please add tests+docs! Please also add code how to use Java plugins. :sparkling_heart:

src/plugins/dump/dump.cpp
```diff
@@ -528,6 +587,7 @@ int elektraDumpGet (ckdb::Plugin * handle, ckdb::KeySet * returned, ckdb::Key *
 	keyDel (root);
 	int errnosave = errno;
 
+	// TODO: document
```
markus2330, line 590: TODO, is this needed for the new version of plugin process, too? (remove if not)

src/plugins/stdioproc/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+	LINK_ELEKTRA elektra-invoke
+	ADD_TEST COMPONENT libelektra${SO_VERSION}-experimental)
+
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/testapp.sh" "${CMAKE_BINARY_DIR}/bin/stdioproc-testapp.sh" COPYONLY)
+
+install (FILES "${CMAKE_BINARY_DIR}/bin/stdioproc-testapp.sh" DESTINATION "${TARGET_TOOL_EXEC_FOLDER}" COMPONENT elektra-tests)
+
+if (ADDTESTING_PHASE AND ENABLE_KDB_TESTING AND "jna" IN_LIST ADDED_BINDINGS)
+	add_msr_test_plugin ("stdioproc" ENVIRONMENT "BUILD_DIR=${CMAKE_BINARY_DIR}")
+endif ()
```
markus2330, line 15:
```suggestion
endif ()

```

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+- infos/metadata =
+- infos/description = one-line description of stdioproc
+
+## Introduction
+
+This plugin spawns a new process with a user-defined executable and delegates all operations to the new process.
+
+## Usage
+
+Set the config key `app` and the arrays `args/#`, `env/#` to the path of an executable, the arguments that shall be passed and the environment variables to be set.
```
markus2330, line 18: app? I would say `executable`

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+```
+kdb mount test.dump /tests/stdioproc stdioproc 'app=/usr/bin/pluginproc' 'args=#1' 'args/#0=--load-plugin' 'args/#1=myplugin'
```
markus2330, line 21: Do we provide this pluginproc as example?

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+During `elektraStdprocioOpen` the plugin will collect the `args/#` and `env/#` values into two arrays `argv` and `envp`.
+It then `fork`s a new process and the child calls `execve` with the path from `app` as well as `argv` and` envp`.
+The child process is expected to listen to `stdin` and write to `stdout` according to the protocol described below.
+
+If communication can be established, the child-process will be kept running until `elektraStdprocioClose`.
```
markus2330, line 28:
```suggestion
If communication can be established, the child process will be kept running until `elektraStdprocioClose`.
```

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+## Protocol
+
+The entire protocol is text-based (apart from binary key values) and request-response-based, and happens over `stdin`/`stdout`.
+The parent process sends request to the child process.
+The child then processes the request and sends a response back.
+
+To encode keysets and keys we use the format of the `dump` plugin.
+It is important to note that the `dump` plugin is instructed to write the full keynames (normally it removes the parent prefix).
+There is, however, an exception during initialization, which is described in the appropriate section.
+The child-process may also delegate directly to the `dump` plugin, or re-implement the encoding.
```
markus2330, line 39:
```suggestion
The child process may also delegate directly to the `dump` plugin, or reimplement the encoding.
```

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+- Initialization: Initial handshake after first starting the child-process.
+- Operation: The main phase, which implements all the operations of an Elektra plugin (open, get, set, close, ...).
+- Termination: The plugin is being closed and the parent-process tells the child to shut down.
+
+> **Note:** In the sections below we use the following format to describe messages:
+>
+> ```
+> Parent > Child
+>
+> HELLO WORLD
```
markus2330, line 52: protocol version here?

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+ELEKTRA_STDIOPROC ACK v1
+(name)
+[contract]
+```
+
+Here `(name)` is the module name the child uses.
+This will be used to replace `stdioproc` in the contract of the plugin.
+The `[contract]` keyset must contain the `infos` keys, which will be used replace the ones at the top of this README and the ones in `stdioproc.c`.
+It must also contain `exports/_` keys for every operation that is implemented by the child.
+All keys in `[contract]` should be below `system:/elektra/modules/stdioproc`.
```
markus2330, line 104: Maybe allow any `system:/elektra/modules/<modulename>`?

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+The parent will rename these keys appropriately.
+
+A child that implements all operations should include these keys (in addition to the relevant `infos` keys):
+
+```
+system:/elektra/modules/stdioproc/exports/open = 1
```
markus2330, line 110: Its not a good idea to reuse the same name for something different. What about

```suggestion
system:/elektra/modules/stdioproc/exports/has/open = 1
```

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+```
+Child > Parent
+
+(result)
+[parent]
+[data]
+```
+
+Here `(result)` is one of `success`, `noupdate` and `error`, which correspond to `ELEKTRA_PLUGIN_STATUS_SUCCESS`, `ELEKTRA_PLUGIN_STATUS_NO_UPDATE` and `ELEKTRA_PLUGIN_STATUS_ERROR` respectively.
```
markus2330, line 149: Why not send ELEKTRA_PLUGIN_STATUS_SUCCESS and so on?

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+This is needed, because the child process cannot request the config keyset otherwise.
+
+The child should then perform the requested operation and respond with
+
+```
+Child > Parent
+
+(result)
+[parent]
+[data]
```
markus2330, line 146:
```suggestion
[returned]
```

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+> **Note**: Under normal circumstances this handshake will always be followed by an `open` request immediately.
+
+### Operation
+
+When the parent needs the child to process an operation (open, get, ...), it will send a request like this:
+
+```
+Parent > Child
+
+(opname)
```
markus2330, line 127: Please be a bit close to normal protocol descriptions, e.g.:

```suggestion
(open|get|set|close)
```

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+```
+Parent > Child
+
+ELEKTRA_STDIOPROC TERMINATE
+```
+
+The child process should now exit (and thereby close its ends of the `stdin`/`stdout` pipes).
+
+> **Note:** Under normal circumstances this only happens, when plugin is being closed.
```
markus2330, line 164:
```suggestion
> **Note:** Under normal circumstances this only happens, after the plugin was being closed.
```

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Errors
+
+If an unexpected error occurs on either side of the protocol, the connection should be terminated and the child process shall exit.
+
+## Examples
+
+```sh
+# mount the Whitelist Java Plugin via stdioproc
+sudo kdb mount config.file user:/tests/stdioproc dump stdioproc 'app=/usr/bin/java' 'args=#3' 'args/#0=-cp' "args/#1=$BUILD_DIR/src/bindings/jna/plugins/whitelist/build/libs/whitelist-$(kdb --version | sed -nE 's/KDB_VERSION: (.+)/\1/gp')-all.jar" 'args/#2=org.libelektra.stdioproc.StdIoProcApp' 'args/#3=org.libelektra.plugin.WhitelistPlugin'
```
markus2330, line 174: Can you please write a `mount-java` wrapper that takes the ugliness out of this?

src/plugins/stdioproc/stdioproc.c
```diff
@@ -0,0 +1,576 @@
...
+	struct
+	{
+		bool open;
+		bool get;
+		bool set;
+		bool close;
+	} hasOp;
+} IoData;
+
+#define MSG_HANDSHAKE_HEADER_V1 "ELEKTRA_STDIOPROC INIT v1\n"
```
markus2330, line 40: const char *?

src/plugins/stdioproc/stdioproc.c
```diff
@@ -0,0 +1,576 @@
...
+int elektraStdioprocOpen (Plugin * handle, Key * errorKey)
+{
+	KeySet * config = elektraPluginGetConfig (handle);
+
+	if (ksLookupByName (config, "system:/module", 0) != NULL)
+	{
+		return ELEKTRA_PLUGIN_STATUS_SUCCESS;
+	}
+
+	Key * appKey = ksLookupByName (config, "/app", 0);
```
markus2330, line 60: See above, `/executable`

src/plugins/stdioproc/README.md
```diff
@@ -0,0 +1,201 @@
...
+kdb set user:/tests/stdioproc/key not_allowed
+# RET: 5
+# STDERR:.*Validation Semantic: .*'not_allowed' does not adhere to whitelist.*
+
+# cleanup
+sudo kdb umount user:/tests/stdioproc
+```
+
+## Limitations
+
```
markus2330, line 198:
```suggestion

- The executable must have an absolute path.
```

src/plugins/stdioproc/stdioproc.c
```diff
@@ -0,0 +1,576 @@
...
+	KeySet * appConfig = ksCut (config, appConfigRoot);
+	keyDel (appConfigRoot);
+	ksAppend (config, appConfig);
+
+	pid_t pid;
+	int parentToChild[2], childToParent[2];
+
+	if (pipe (parentToChild) != 0)
+	{
+		ELEKTRA_SET_RESOURCE_ERRORF (errorKey, "Could not execute app. Reason: %s", strerror (errno));
```
markus2330, line 89: Add more information:

- what the executable was
- that pipe failed

etc.

kodebach:
> Please add tests+docs! Please also add code how to use Java plugins. 💖

I have now added tests/docs. The README for `stdioproc` includes a Java plugin example.

> Please remove the old plugin process and rename your plugin to be called "process".

Will do before the PR is merged. I'll wait for your review, however, because renaming stuff mid-review breaks GitHub's comments.

src/plugins/stdioproc/testmod_stdioproc.c
```diff
@@ -0,0 +1,180 @@
...
+ */
+
+#include <stdlib.h>
+#include <string.h>
+
+#include <kdbconfig.h>
+
+#include <tests_plugin.h>
+
+// TODO: check contract
```
markus2330, line 17: TODO

markus2330 (commented):
Great work! :rocket:

src/bindings/jna/README.md
```diff
@@ -49,7 +49,7 @@ export CLASSPATH=".:/usr/share/java/libelektra.jar:/usr/share/java/jna.jar"
 Then you can compile and run [HelloElektra](HelloElektra.java):
 
 ```sh
-javac HelloElektra.java && java HelloElektra
+javac org.libelektra.stdioproc.StdIoProcApp.java && java HelloElektra
```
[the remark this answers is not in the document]
kodebach: This was not an intentional change.

src/bindings/jna/stdioproc/src/main/java/org/libelektra/stdioproc/StdIoProc.java
kodebach: It is/was planned with the new backend. Implementing the change separately would be a lot of unnecessary overhead.

src/plugins/dump/dump.cpp
```diff
@@ -476,6 +476,63 @@ int unserialise (std::istream & is, ckdb::Key * parentKey, ckdb::KeySet * ks, bo
 	return ELEKTRA_PLUGIN_STATUS_SUCCESS;
 }
 
+class FileStreamBuf : public std::streambuf
```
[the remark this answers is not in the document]
kodebach: I'll add comments, but basically it's the only way I found of reading from a `FILE *` via C++ streams.

src/plugins/stdioproc/README.md
kodebach: No, this is a purely hypothetical example. A real world example (Java) can be found below. I didn't want to use the `testapp.sh` as an example, since it doesn't really do what the child process is "supposed to". It only does things that make it convenient for testing.

src/plugins/stdioproc/README.md
kodebach: IMO this is only an option, if `<modulename>` must match the previously sent `(name)`. This would also create a new source of errors.

src/plugins/stdioproc/README.md
kodebach: This is just an example of how the protocol is described within the README. It is not part of the actual protocol.

src/plugins/stdioproc/README.md
kodebach: Why is sending a much longer string any better? We already have a handshake, so the `ELEKTRA_PLUGIN_STATUS_` prefix doesn't serve any purpose here.

If you mean sending the integer values of the constants, then this would just make the protocol less human-readable. It wouldn't actually have any benefit, because you'd still need to decode the string e.g. `1` into the integer value.

src/plugins/stdioproc/README.md
kodebach: No it actually happens **while** the plugin is being closed, i.e. the `close` command is sent in the middle of a `elektraPluginClose` on an instance of `stdioproc`.

src/plugins/stdioproc/README.md
kodebach: In theory we could do something like

```sh
sudo kdb mount-java config.file user:/tests/stdioproc kdb:dump java:org.libelektra.plugin.WhitelistPlugin 
```

The `java:` and `kdb:` prefixes are to differentiate between `Java` plugins and others. This would also only work for Java plugins that are part of this repo. If you want to use plugins from other JARs, you'd have to define the `CLASSPATH` environment variable accordingly.

However, I don't think this example should be changed at all (adding `mount-java` is still a good idea). This example is not supposed to show how Java plugins are mounted. It is supposed to be an example of how to use `stdioproc`. Therefore, it should not hide the actual `stdioproc` config in some helper script.

src/plugins/stdioproc/stdioproc.c
kodebach: Then we can't use `sizeof`. In most other cases `strlen` would work and the compiler would optimize it away. But in the snippet below, it would change the meaning of the code.

https://github.com/ElektraInitiative/libelektra/blob/38d1d1789759aa8968575ad857e38f90d9c8c9ae/src/plugins/stdioproc/stdioproc.c#L148

src/plugins/stdioproc/README.md
markus2330: @kodebach wrote:
> In theory we could do something like
> 
> sudo kdb mount-java config.file user:/tests/stdioproc kdb:dump java:org.libelektra.plugin.WhitelistPlugin

I think we can assume `kdb:` if no `java:` is given?

> The |java:| and |kdb:| prefixes are to differentiate between |Java| 
> plugins and others. This would also only work for Java plugins that are 
> part of this repo. If you want to use plugins from other JARs, you'd 
> have to define the |CLASSPATH| environment variable accordingly.
> 
> However, I don't think this example should be changed at all (adding 
> |mount-java| is still a good idea). This example is not supposed to show 
> how Java plugins are mounted. It is supposed to be an example of how to 
> use |stdioproc|. Therefore, it should not hide the actual |stdioproc| 
> config in some helper script.

I agree, in the process plugin there should be the full example without any helper scripts. Could you please add/extend a Java-Plugin Tutorial where you describe `mount-java` and also how to set CLASSPATH for out-of-source-tree Java Plugins?

kodebach: > I think we can assume kdb: if no java: is given?

I don't think there is technical reason why `java:com.example.Foo` couldn't be the name of a plugin written in C. The `.so` file certainly could contain `.` and `:`. But we could make `kdb:` optional. Then we'd have:
- `kdb:java:com.example.Foo` means C plugin named `java:com.example.Foo`
- `java:com.example.Foo` means Java plugin implemented in the class `com.example.Foo`
- `foo` and `kdb:foo` both mean C plugin named `foo`


>  Could you please add/extend a Java-Plugin Tutorial where you describe mount-java and also how to set CLASSPATH for out-of-source-tree Java Plugins?

Will do

src/plugins/dump/dump.cpp
kodebach: I will remove this with the removal of the old `process` and renaming of `stdioproc` (after PR reviews are done).

kodebach:
@markus2330 I have now update the docs and added your other suggestions. Apart from the removal of the old `process` and renaming of `stdioproc` to `process`, this PR should be done IMO.

markus2330 (commented):
I reviewed already, please do last changes (removal+renaming), making buildserver happy, etc.

kodebach:
Everything is now renamed. I didn't remove the `pluginprocess` library, because it is also used by the `python` plugin. We should probably add a separate PR that removes the `python` plugin and creates helper scripts in Python similar to what we added here for Java.

markus2330:
There are still several problems with the build server. Can you please fix/rerun?

Btw. did you check that everything gets installed? I'll test when merged with the packages.

kodebach:
> Can you please fix/rerun?

I'm not really sure, what's wrong. It works on my system, so I'll have to investigate a bit. From the error message it seems like Java can't find the `.so` files, which is weird because `LD_LIBRARY_PATH` should be set correctly.

> Btw. did you check that everything gets installed?

At least on my system everything gets installed, yes. I checked, because that's the only way to test `kdb mount-java` (at least the part that automatically sets the classpath).

kodebach:
I found the bug (`LD_LIBRARY_PATH` wasn't propagated to the child process), hopefully everything runs correctly now

src/plugins/dump/dump.cpp
```diff
@@ -476,6 +476,68 @@ int unserialise (std::istream & is, ckdb::Key * parentKey, ckdb::KeySet * ks, bo
...
+		return this->gptr () == this->egptr () ? std::char_traits<char>::eof () :
+							       std::char_traits<char>::to_int_type (*this->gptr ());
+	}
+
+private:
+	FILE * file_;
+	char buf_;
+};
+
+int freadks (KeySet * ks, FILE * file, Key * errorKey)
```
markus2330, line 527: rename to fserialize, funserialize (also rename serialise and unserialise)

markus2330 (commented):
Looks good to me to start testing, added some suggestions for better names.

doc/tutorials/java-plugins.md
```diff
@@ -11,42 +11,65 @@ If you need a tutorial for using Key and KeySet in Java, please check out the [H
...
+Otherwise, there are a few differences between implementing a plugin in C and in Java:
 
-The `Plugin.java` interface contains methods for the five basic functions of plugins:
+1. In C it is possible to implement only `elektraPluginGet` and leave the other functions unimplemented.
+   Because of interface inheritance in Java, it is required to implement all 5 method `open`, `get`, `set`, `error` and `close`.
+   Whether a method is actually supported, must be communicated via the contract.
+   Methods that are not supported, should simply return `Plugin.STATUS_SUCCESS`.
+2. In C the parent key of the contract depends on the plugins name.
+   For example, the contract for `dump` can be found under `system:/elektra/modules/dump` and the `dump` plugin returns it as such.
+   However, in Java the parent key for the contract is always `system:/elektra/modules/jni` (you may use the constant `Plugin.JNI_MODULE_CONTRACT_ROOT`).
```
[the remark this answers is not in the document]
kodebach: I just kept the old key name to minimize the number of changes.

doc/tutorials/java-plugins.md
```diff
@@ -11,42 +11,65 @@ If you need a tutorial for using Key and KeySet in Java, please check out the [H
...
+In the C API this would be taken from the contract, but the `process` protocol requires this separately, so we need a separate method for it.
 
-In order to write a new Java Plugin, the new class has to be created under the `plugin` folder. It has also to extend the `Plugin.java` interface, which contains all required methods to communicate with Elektra’s database.
+Otherwise, there are a few differences between implementing a plugin in C and in Java:
 
-The `Plugin.java` interface contains methods for the five basic functions of plugins:
+1. In C it is possible to implement only `elektraPluginGet` and leave the other functions unimplemented.
+   Because of interface inheritance in Java, it is required to implement all 5 method `open`, `get`, `set`, `error` and `close`.
+   Whether a method is actually supported, must be communicated via the contract.
+   Methods that are not supported, should simply return `Plugin.STATUS_SUCCESS`.
```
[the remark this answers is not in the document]
kodebach: A method that is not marked in the `exports/has` part of the contract shouldn't be called. However, returning `Plugin.STATUS_SUCCESS` is still safer, because if the method is somehow called, simply nothing happens.

Also, if you want to throw something, the appropriate type would be [`UnsupportedOperationException`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/UnsupportedOperationException.html). A `NoSuchMethodError` is a subtype of `Error` and therefore should not be caught and simply lead to a program crash.

doc/tutorials/mount-java.md
```diff
@@ -0,0 +1,33 @@
...
+
+The `mount-java` works as is for all plugins that come bundled with Elektra.
+If you have external plugins, you must also specify the classpath for these plugins like this:
+
+```sh
+CLASSPATH=/path/to/foo.jar:/path/to/bar.jar sudo kdb mount-java config.file user:/tests/foobar java:org.example.Foo java:org.example.Bar java:org.libelektra.plugin.WhitelistPlugin
+```
+
+Here we added `/path/to/foo.jar` and `/path/to/bar.jar` to the classpath, so that the classes `org.example.Foo` and `org.example.Bar` can be found.
+We can still use `org.libelektra.plugin.WhitelistPlugin`, because Elektra's JARs are always added to the classpath.
```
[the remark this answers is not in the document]
kodebach: Compiling the plugin isn't any different from compiling any other Java library into a JAR file. This is not supposed to be a Java programming tutorial, so I only detailed the parts that are special about Elektra plugins.

src/bindings/jna/process/src/main/java/org/libelektra/process/ProcessApp.java
```diff
@@ -0,0 +1,55 @@
...
+        | NoSuchMethodException e) {
+      System.exit(1);
+      return;
+    }
+
+    var protocol = new ProcessProtocol(plugin, System.in, System.out);
+
+    try {
+      if (!protocol.handshake()) {
+        System.exit(1);
```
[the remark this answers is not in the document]
kodebach: Since the handshake failed, the only real option here is to print something to `stderr`. However, that could be really confusing, because (under normal circumstances) we are not in the process that was called by the user.

Even if we print error messages to `stderr` (which could be really out of place depending on what the parent process is doing), in this specific case, I don't think we need one. If the handshake failed, there isn't much information we could provide. If the app was called by the `process` plugin and something went wrong, `process` will issue a "handshake failed" error. If the app was called in some other way, just exiting because the app is being misused seems fine.

src/bindings/jna/process/src/main/java/org/libelektra/process/ProcessApp.java
```diff
@@ -0,0 +1,55 @@
...
+        return;
+      }
+
+      if (!protocol.run()) {
+        System.exit(1);
+      }
+    } catch (IOException e) {
+      System.exit(1);
```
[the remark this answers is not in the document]
kodebach: Where would this error message go? The `IOException` means that something went wrong while communicating with the parent process. Sending an error message via the protocol is clearly not an option here.

For more about errors see above.

src/plugins/dump/dump.cpp
kodebach: > also rename serialise and unserialise

Why?

markus2330: Because it is still British English and we decided to go for American English.

src/bindings/jna/process/src/main/java/org/libelektra/process/ProcessApp.java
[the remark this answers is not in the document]
markus2330: We could exit with 3 different error status? Then the plugin can wait for the child and knows the exit status for an error message.

doc/tutorials/mount-java.md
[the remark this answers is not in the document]
markus2330: As said: then simply put the information to the man page.

src/plugins/dump/dump.cpp
kodebach: Oh okay, I thought you didn't like the term serialize/unserialise for some reason

src/bindings/jna/process/src/main/java/org/libelektra/process/ProcessApp.java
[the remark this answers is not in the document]
kodebach: I'm not sure that would have much benefit. The three cases would be:
1. "Handshake failed": Without any extra information the separate exit code is useless. The plugin already knows that the handshake failed, because it initiated the handshake which the child process aborted.
2. An IOException: In this case, maybe a different error code could convey some information. The parent process probably also got some error from `read`/`write`ing to its end of the pipes.
3. Another error during the protocol execution: This mostly boils down to an encoding/decoding error of the dump format. This shouldn't happen and if it does it's a bug.

Additionally, in all 3 cases the user can't really do anything to solve the error. Not even the developer of the Java plugin could do anything here. All these errors are essentially bugs that need to be reported and fixed by us (*).

---

(*) I didn't use `ELEKTRA_SET_INTERNAL_ERROR` in the `process` plugin, because the errors are only internal errors, if the child app is also provided us (like with Java). If it's an external app, the developer of that app needs to fix the problem.

doc/tutorials/java-plugins.md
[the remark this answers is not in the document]
markus2330: > because if the method is somehow called, simply nothing happens.

Why would it be "somehow" called? If export/has = 0, we could guarantee that we do not call it? And then it would not matter how the implementation is. A typical implementation in Java would be to throw `NoSuchMethodError`, as calling such a method is basically an assertion error (which also extends `Error`).

The sentences as they are written now sound like we might call it, thus the user must implement it as `Plugin.STATUS_SUCCESS`. This is problematic from maintenance point of view, as someone might think the implementation is as it should be while in fact it is not implemented.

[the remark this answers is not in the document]
kodebach: > A typical implementation in Java would be to throw NoSuchMethodError, as calling such a method is basically an assertion error (which also extends Error).

Every implementation I've seen would use `UnsupportedOperationException` in this case. AFAIK `NoSuchMethodError` is basically only used by the JVM itself when bytecode for classes is modified at runtime or when using reflection.

---

Thinking about it, the best solution might actually be to use `default` implementations in the interface:

```java
public interface Plugin {
  default int open(KeySet config, Key errorKey) {
    throw new UnsupportedOperationException();
  }
  
  // no default because it must be implemented for the contract
  int get(KeySet keySet, Key parentKey) throws KDBException;

  default int set(KeySet keySet, Key parentKey) {
    throw new UnsupportedOperationException();
  }

  default int error(KeySet keySet, Key parentKey) {
    throw new UnsupportedOperationException();
  }

  default int close(Key parentKey) {
    throw new UnsupportedOperationException();
  }
}
```

src/bindings/jna/process/src/main/java/org/libelektra/process/ProcessApp.java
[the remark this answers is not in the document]
markus2330: > "Handshake failed": Without any extra information the separate exit code is useless. The plugin already knows that the handshake failed, because it initiated the handshake which the child process aborted.

It does not know if the child agrees about that the handshake failed. Even if this is very unlikely/impossible there is also not a good reason to use the same exit code?

> An IOException: In this case, maybe a different error code could convey some information. The parent process probably also got some error from read/writeing to its end of the pipes.

Good idea!

> Additionally, in all 3 cases the user can't really do anything to solve the error. Not even the developer of the Java plugin could do anything here. All these errors are essentially bugs that need to be reported and fixed by us (*).

Yes, so the error should be designed that it is helpful for us.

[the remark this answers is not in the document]
markus2330: > What do you mean "the child agrees"?

The child might have written the ACK to stdout but the parent process does not receive it (e.g. wrongly connected stdout).

> On a more technical note, it also requires quite extensive changes to the process plugin's code.

If it makes the code much more complicated then leave it. We will see if errors in the protocol occurs and if we will be able to debug it easily :wink:

PR #4244, opened 2022-02-15
[decisions] constructor functions

kodebach:


doc/decisions/constructor_functions_examples.md
```diff
@@ -0,0 +1,105 @@
+### Constructor Functions: Language examples
```
kodebach, line 1: Please note: This file is just for illustration, I will probably remove it before we merge the decision.

doc/decisions/constructor_functions.md
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Allocates a new keyset with space reserved for at least @p keyCount keys.
+ * If `keyCount == 0`, the keyset will be empty (with no space for keys reserved) and @p keys is ignored.
+ * If `keys == NULL`, the keyset will be empty with space reserved for at least @p keyCount keys.
+ * Otherwise the keyset will contain all `keys[i]` for any `0 <= i < keyCount` where `keys[i] != NULL`.
+ * In other words, `NULL`s within @p keys will be ignored and if @p keys contains more than @p keyCount keys
+ * the additional ones will also be ignored.
+ */
+Key * elektraKeySetNew (Key ** keys, size_t keyCount);
+```
```
kodebach, line 59: I decided to use a slightly more useful API than before. I still think that this API (especially `elektraKeyNew`) should almost never be used directly. However, the previous API was essentially just an allocator and had the problem that you always create a key with the wrong name first.

The fact that the size must be passed separately is a bit annoying, but since it should be hidden by another API in practice it doesn't matter much. The API should also still be simple enough that it could be implemented in any language.

doc/decisions/constructor_functions.md
```diff
@@ -0,0 +1,85 @@
+# Constructor Functions
+
+## Problem
+
+The constructor functions that `libelektra-core` defines for creating Keys and KeySets must be possible to implement in other languages without complicated workarounds.
```
markus2330, line 5: Afaik this problem does not exist, it was possible for every language we support?

```suggestion
The constructor functions for creating Keys and KeySets must be easy to be invoked from other languages without complicated workarounds.
```

doc/decisions/constructor_functions.md
```diff
@@ -0,0 +1,85 @@
...
+- Constructor functions should use as little resources as feasible.
+  In particular, the number of `malloc`s and `memcpy` should be kept in check.
+  Temporary allocations during key creation should be avoided.
+- Using the escaped name in the constructor function should be avoided, since the unescaping process can be expensive.
+
+## Assumptions
+
+- The constructor functions from `libelektra-core` will not or only very rarely be used directly by users.
+- For all languages supported by Elektra (including C), there is secondary library on top of `libelektra-core` that provides more idiomatic APIs.
+- Creating a key with a name, no value and no metadata and then calling `keySetValue`/`keySetMeta` does not create significant overhead compared to creating the key with value and metadata directly.
```
markus2330, line 19: This is the key of the whole decision. I don't think we should base this on assumptions but actually benchmark keyNew if avoiding calling `keySetValue`/`keySetMeta` is a relevant topic or not.

Simply leaving it as is is also totally possible. In any case it is good that we now have this decision.

doc/decisions/constructor_functions.md
markus2330: Please write this in the decision (considered alternatives). It is unclear to me which text refers to Key and which to KeySet. In case of the Key** the size is actually not needed and it could be null-terminated.

> The API should also still be simple enough that it could be implemented in any language.

As said above, it was also possible to call keyNew from any language.

doc/decisions/constructor_functions.md
```diff
@@ -0,0 +1,85 @@
...
+ */
+Key * elektraKeySetNew (Key ** keys, size_t keyCount);
+```
+
+Other libraries will provide APIs on top of these functions.
+These functions should be more idiomatic to their target language.
+
+### Language Examples
+
+There are some examples of what the language specific APIs could look like in [another file](constructor_functions_examples.md).
```
markus2330, line 66: I think you can also have these examples here (or in Notes)

doc/decisions/constructor_functions.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Rationale
+
+It is very hard to find a C API that works well in every language, but is still useful for creating keys.
+It is much easier, to provide good APIs for a single language and build those on top of a very basic generic API.
+
+This decision also guides Elektra towards the goal of a minimal core API.
+
```
markus2330, line 74: Argumentation for the actual API you chose is missing. As said: benchmarks would be highly appreciated.

doc/decisions/constructor_functions_examples.md
```diff
@@ -0,0 +1,105 @@
...
+
+This is not a decision.
+This file contains additional information for [Constructor Functions](constructor_functions.md).
+
+## Rust
+
+For example, `libelektra-rust` (targeted at Rust) could provide:
```
markus2330, line 8: I don't see the connection to this decision, this API can be provided in any case?

doc/decisions/constructor_functions_examples.md
```diff
@@ -0,0 +1,105 @@
...
+elektraKeyCreate (KEY_NS_USER, "foo", "bar", KEY_NAME_END,
+        KEY_VALUE_STRING, "value",
+        KEY_META, elektraKeyCreate (KEY_NS_META, "type", KEY_NAME_END, KEY_VALUE_STRING, "string", KEY_END),
+        KEY_META, elektraKeyCreate (KEY_NS_META, "other", KEY_NAME_END, KEY_VALUE_BINARY, &mystruct, sizeof(mystruct), KEY_END),
+        KEY_END);
+```
+
+However, this would require multiple `memcpy` to initialize the key name.
+
+Alternatively, macros could be used extensively to hide the size arguments from (most) users:
```
markus2330, line 45: I don't think we need this.

markus2330 (commented):
Thank you for writing up the decision! :sparkling_heart:

doc/decisions/constructor_functions.md
kodebach: To clarify, the assumption says:

There is no significant overhead when switching from
```c
Key * k = keyNew ("/foo", KEY_VALUE, "value", KEY_META, "meta:/type", "string", KEY_META, "meta:/bar", "baz", KEY_END);
```

to

```c
Key * k =keyNew ("/foo", KEY_END);
keySetString (k, "value");
keySetMeta (k, "meta:/type", "string");
keySetMeta (k, "meta:/bar", "baz");
```

This is currently true, because `keyNew` internally just call `keySetString`/`keySetMeta`. (*) The assumption is just that, if `keyNew` were to directly set the value and directly construct the correct `struct _KeySet`, there still wouldn't be a significant difference. Since there are no optimizations that `keyNew` could do, that couldn't be done by `keySetValue`/`keySetMeta`, I think this is a fair assumption.

Additionally, the assumption is actually only semi-relevant to the decision. In fact, the actual decision, makes the assumption less relevant. Since `keyCreate` could still use the private API and directly construct the `struct _Key`, any optimization could simply be implemented there.

---

(*) In theory, there could be (significant) difference between 1 call and 4 calls to dynamically linked symbols. However, if that change is significant, the whole idea of a minimal API becomes question able. If the API is minimal, you always need more calls to do complicated stuff, compared to an API that has functions for everything.

doc/decisions/constructor_functions.md
kodebach: They are pretty long and don't mostly serve as an illustration during the discussion. I'm not sure it is worth keeping them at all, once we finalize the decision. It's only gonna lead to problems, if we later on decide on different APIs (very possible since the examples are not part of the actual decision here).

doc/decisions/constructor_functions_examples.md
```diff
@@ -0,0 +1,105 @@
...
+However, here we can also add a secondary function `elektraKeyBuildFromParts`, which would build the name from parts with multiple `memcpy` (something that cannot be avoided with non-literal strings):
+
+```c
+Key * elektraKeyBuildFromParts(elektraNamespace ns, const char ** nameParts, const void * value, size_t vSize, const KeySet * meta);
+```
+
+This could be use with non-literal strings like so:
+
+```c
+elektraKeyBuildFromParts(NS_USER, ((const char*[]) { "foo", bar, "baz" }), "value", sizeof("value"), NULL);
```
kodebach, line 92: Just noticed, this doesn't work. We need a `NULL` terminator (or a `partCount`):

```suggestion
elektraKeyBuildFromParts(NS_USER, ((const char*[]) { "foo", bar, "baz", NULL }), "value", sizeof("value"), NULL);
```

doc/decisions/constructor_functions_examples.md
kodebach: I'm not a fan of the varargs API there are a few downsides:
- multiple memcpy for the name (could still be avoided with some macros and/or extra arguments)
- it's very easy to pass wrong arguments and get weird errors. For example, if one of the name parts is a variable which is `NULL` for some reason.
- There are basically no compiler checks for a vararg API. With the many macros below, errors may be hard to read, but the compiler will still check the result after expanding the macros, so fewer errors are possible.

doc/decisions/constructor_functions_examples.md
kodebach: The connection is this line in the actual decision:

> Other libraries will provide APIs on top of these functions.

You _can_ of course always provide additional APIs. However, the point here is that all languages (including C) **will** provide additional APIs to make the very minimal core API more usable. This basically allows us to create a more restricted core API.

----

I should probably also change the Rust example to something like this:

```rust
trait KeyBuilder {
    fn new(ns: Namespace, name: &[u8]) -> Self;
    fn set_value(&mut self, value: &[u8]) -> Self;
    fn set_meta(&mut self, name: &[u8], value: &[u8]) -> Self;
    fn build(self) -> Key;
}
```

With the additional explanation: If benchmarks show that this is more efficient, none of the functions above call any core APIs and `build` directly constructs a `struct _Key` via private APIs.

doc/decisions/constructor_functions.md
kodebach: I will add the Considered Alternative (`keyNew(void)` and `ksNew(size_t)`) and a short explanation why other APIs would be worse (in particular for other languages).

> benchmarks would be highly appreciated.

What exactly would a benchmark achieve here? I didn't propose any specific optimizations, so I can't really benchmark anything.

doc/decisions/constructor_functions_examples.md
kodebach: I think, I have found a compromise. The idea revolves around a public `struct` and a `static inline` function.

```c
// holds a string (with potential zero bytes) of `size` length
// if `ns == ELEKTRA_NS_NONE`, the `name` and `size` are directly useable in struct _Key
// otherwise:
//  - `size` is the length of `name` including the terminating zero byte, but without a namespace
//  - `name` does not contain a namespace
//  - the full name is constructed from `ns` and `name`
struct ElektraKeyName {
    size_t size;
    const char * name;
    ElektraNamespace ns;
};

static inline struct ElektraKeyName elektraKeyName(ElektraNamespace ns, const char * parts[])
{
    struct ElektraKeyName k;
    k.ns = ELEKTRA_NS_NONE;

    k.size = 2;
    for (size_t i = 0; parts[i] != NULL; ++i)
    {
        k.size += strlen(parts[i]) + 1;
    }
          
    char * name = malloc(k.size);
    name[0] = ns;
    name[1] = 0;
    k.size = 2;
    for (size_t i = 0; parts[i] != NULL; ++i)
    {
        size_t s = strlen(parts[i]);
        memcpy(name + k.size, parts[i], s);
        name[k.size + s] = 0;
        k.size += s + 1;
    }
    k.name = name;
    return k;
}
```

This would be part of the header for `libelektra-operations` (or a similar library). This allows us to define a `elektraKeyCreate` function like this:

```c
Key * elektraKeyCreate(struct ElektraKeyName name, void * value, size_t valueSize, KeySet * meta)
```

All of this may seem odd at first, but combining the arguments for namespace, name and size of name into a struct, allows us to create them via macros (which makes the syntax nicer).

```c
#define ELEKTRA_BUILD_NAME(ns, ...) (elektraKeyName(ns, (const char *[]){__VA_ARGS__,NULL}))
#define ELEKTRA_KEY_NAME(_ns, _name, _size) ((struct ElektraKeyName){.size = _size, .name = _name, .ns = _ns})
#define ELEKTRA_LITERAL_NAME(_ns, _name) (ELEKTRA_KEY_NAME(_ns, _name, sizeof(""_name)))
```

All in all the effect is that `elektraKeyCreate` can be called in many different ways:

```c
    struct ElektraKeyName y = ELEKTRA_BUILD_NAME(ELEKTRA_NS_CASCADING, "abc", "def", "xyz", "123", "456", "ghi", "foo", "lll", "kyt");
    Key * k = elektraKeyCreate(y, NULL, 0, NULL);
    
    Key * k1 = elektraKeyCreate(ELEKTRA_BUILD_NAME(ELEKTRA_NS_CASCADING, "abc", "def"), NULL, 0, NULL);
    
    Key * k2 = elektraKeyCreate(ELEKTRA_LITERAL_NAME(ELEKTRA_NS_CASCADING, "foo\0bar"), NULL, 0, NULL);

    const char *x = "foo\0bar";
    Key * k3 = elektraKeyCreate(ELEKTRA_KEY_NAME(ELEKTRA_NS_CASCADING, x, 8), NULL, 0, NULL);

    const char *a = "foo";
    const char *b = "bar";
    Key * k4 = elektraKeyCreate(ELEKTRA_BUILD_NAME(ELEKTRA_NS_CASCADING, "abc", a, b, "def"), NULL, 0, NULL);
```

Crucially when the `static inline elektraKeyName` function is used, we directly use the `elektraMalloc`ed buffer. This in turn makes it possible for the compiler (with `-O3`) to do lots of constant folding and (in theory) fully calculate the key name at compile-time in many instances. After inlining and constant folding, only the `malloc` with a precomputed size and a (vectorized) `memcpy` from static memory to the heap remain.

Sadly, however, it seems GCCs constant folding algorithm isn't powerful enough. But Clang can handle even the more complex cases as this Godbolt link shows: https://godbolt.org/z/Tf1rnxfMW (The important part is that in the assembly, only the last `elektraKeyCreate` is preceeded by a `memcpy` from the inlined function)

Note also, that only the creation of the key name is inlined. The rest is still hidden behind a layer of abstraction, so this does not leak the layout of `struct _Key` to the API consumer. Inlining the allocation of the key name is not a problem, because the structure of the key name is already public API..

doc/decisions/constructor_functions.md
```diff
@@ -0,0 +1,112 @@
+# Constructor Functions
+
+## Problem
+
+The constructor functions for creating Keys and KeySets must be easy to be invoked from other languages without complicated workarounds.
+In particular, another implementation should not need to resort to adding a small amount of C to "bridge the gap".
```
markus2330, line 6: In which binding do we have C code only for constructors? In those we do not have any C code (like JNA), keyNew and ksNew works without any bridge code. In those we already have C code, fixing only the constructors seems pointless to me.

@lawli3t What is the situation in Rust?

kodebach: The Go binding definitely has some wrapper C code already. 

https://github.com/ElektraInitiative/go-elektra/blob/0462a716b69709257ad2e7b0b881eef645a0160f/kdb/key.go#L3-L13

The existing Rust binding doesn't have wrapper code, but it tries it's best to ignore the vararg part of the C API:

https://github.com/ElektraInitiative/libelektra/blob/eac1c6a74cb75bb84ebfae2247b3cc531259ded8/src/bindings/rust/elektra/src/writeable.rs#L55-L75

The Java binding does use the varargs, but needs some pretty horrible translation code:

https://github.com/ElektraInitiative/libelektra/blob/eac1c6a74cb75bb84ebfae2247b3cc531259ded8/src/bindings/jna/libelektra/src/main/java/org/libelektra/Key.java#L116-L137

which at the end just calls another layer of translation code: 

https://github.com/ElektraInitiative/libelektra/blob/eac1c6a74cb75bb84ebfae2247b3cc531259ded8/src/bindings/jna/libelektra/src/main/java/org/libelektra/Key.java#L92-L102

The SWIG bindings are based on C++ and AFAICT the SWIG bindings and C++ are the only ones where the varargs API from C can actually be used.

In summary, maybe the phrasing about "C to bridge the gap" was not quite accurate. It applies in general IMO that there shouldn't be a need for bridge code, but it is not the current problem. Nonetheless, I'd say creating a varargs binding in the core API which then can't always be used directly is a bad idea. That is why I would avoid varargs in the core API and either completely avoid them or at the very least put them into a separate lowlevel C library. It simply allows more flexibility and allows the bindings to better fit their languages.

AFAIK the main reason for the whole varargs API is that it allows creating a whole `KeySet` in a single expression. In C this can only be achieved, if a `KeySet` is created by single function that receives all the `Key`s _and_ a `Key` can also be constructed with a single function. But in other languages there are different, more idomatic possibilities. In Java, Rust and many other languages, a builder pattern makes more sense. I don't see the point in burying an impractical varargs API underneath those builders. Especially not if the varargs API still just calls `keySetName`, `keySetMeta`, etc.

> fixing only the constructors seems pointless to me

To me it seems pointless to have varargs functions. I don't actually know of a single argument in favour of varargs other than single expression thing (which I've now shown multiple times can be achieved differently too). IMO varargs just remove more of the already very lax type checks from the C compiler.

---

At the end there is one thing I am 100% certain of. There is absolutely no place for varargs APIs in the `libelektra-core` API. We can argue about whether or not we want to use them in `libelektra-lowlevel-c` (which targets only C), but for the core API I don't see any reason to have a varargs API.

markus2330: > At the end there is one thing I am 100% certain of. There is absolutely no place for varargs APIs in the libelektra-core API. We can argue about whether or not we want to use them in libelektra-lowlevel-c (which targets only C), but for the core API I don't see any reason to have a varargs API.

If there is no massive performance benefit (with a proper keyNew allocating everything at once), I agree. If there is a massive performance benefit, the status quo makes sense: bindings wrap their keyNew in a way it is suitable for them . E.g. if they have a constructor that always creates key+value, they would call e.g. `keyNew(key, KEY_VALUE, value, KEY_END)`.

So we need a benchmark (that shows that allocations are not the bottleneck of keyNew) to make a proper decision.

Btw. I am very thankful that you are so persisting in these things as this is the only way to look deeply enough to make really good decisions.

markus2330 (approved):
Lets merge this as draft and let us later continue. I think we should fully focus on core API changes.

markus2330:
There are CI problems, though.

kodebach:
I'll merge the decision as drafts in this PR. But at least `constructor_functions.md` relates to the core API, so it should be decided soon.

The other decision also shouldn't be delayed to long, since it will the be replacement for most `keyNew`/`ksNew` calls currently in use. Few places in the repo have the requirement to use only `libelektra-core`, so in plugins etc. I would definitely use a more convenient API when available.

kodebach:

Issue #2987, opened 2019-09-20
Cleanup


markus2330:
A meta-issue to collect all cleanup tasks.

- [ ] old python2 code generator (pythongen)
- [ ] libmeta
- [x] remove website-backend, rename website-frontend
- [ ] remove specload, see #4032 
- [ ] plugins
  - [ ] iterate

PR #4245, opened 2022-02-15
[decisions] extra libraries

kodebach:


doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,61 @@
...
+If a function is generally useful, it should be added to [`libelektra-operations`](operations_library.md).
+However, some functions are only useful as a C API and shouldn't be part of a language binding.
+
+## Constraints
+
+- `libelektra-core` should have a minimal (but still useful) API.
+- `libelektra-operations` should contain generally useful, but not minimal API, that can be used via language bindings
+
+## Assumptions
+
```
markus2330, line 17: We assume it won't grow to big?

doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,61 @@
+# Operations Library
+
+## Problem
+
+There are many functions that can be useful, but are not required to form the minimal API.
+Since we want to keep the API of `libelektra-core` minimal, we cannot add these functions.
+
+If a function is generally useful, it should be added to [`libelektra-operations`](operations_library.md).
+However, some functions are only useful as a C API and shouldn't be part of a language binding.
```
markus2330, line 9: Sentence unclear: I am missing the info that `libelektra-operations` should be part of a language binding?

doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,61 @@
...
+Other examples include `elektraKeyCreate` and `elektraKsCreate`.
+Their API allows the creation of an entire KeySet within a single expression.
+However, the API was specially designed for use in C and is very hard or impossible to use from other languages.
+This makes them the ideal candidates for `libelektra-lowlevel`.
+
+## Rationale
+
+The separation between `libelektra-operations` and `libelektra-lowlevel` has a few advantages:
+
+- It is clear what the target audience for the API is (everyone vs. C-only)
```
markus2330, line 47: I fully agree that (at least) two different libraries based on this distinction must exist.

Open questions are:

- is it only 2?
- what is the best name? ("lowlevel" actually suggests that a binding for it might be useful, as "highlevel" also is open to bindings)

doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,61 @@
...
+- Binding authors now that they should (in most cases) creates bindings for everything in `libelektra-operations`.
+  If we just had one library, binding authors would have to check every function.
+- Because `libelektra-lowlevel` targets C exclusively, it can make heavy use of macros and other special features, if needed.
```
markus2330, line 50: :+1:

doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Implications
+
+`libelektra-lowlevel` depends on `libelektra-operations` and `libelektra-core`, so `libelektra-core` and `libelektra-operations` cannot use functions from `libelektra-lowlevel`.
```
markus2330, line 54: :+1:

doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,61 @@
...
+Anything that is needed in `libelektra-core`/`libelektra-operations` and should be public must be implemented there.
+
+## Related Decisions
+
+- [Iterating Keyname Parts](iterating_name_parts.md.md)
```
markus2330, line 59:
```suggestion
- [Iterating Keyname Parts](iterating_name_parts.md)
```

doc/decisions/key_below.md
```diff
@@ -0,0 +1,62 @@
...
+## Rationale
+
+The function is required for a minimal API, but it needs access to private API to be efficient.
+Therefore, it is part of the operations library.
+
+## Implications
+
+## Related Decisions
+
+- [Operations Library](operations_library.md)
```
markus2330, line 60:
```suggestion
- [Operations Library](operations_library.md)
- [Lowlevel Library](lowlevel_library.md)
```

doc/decisions/operations_library.md
```diff
@@ -0,0 +1,50 @@
+# Operations Library
+
+## Problem
+
+There are many functions that can be useful, but are not required to form the minimal API.
+Since we want to keep the API of `libelektra-core` minimal, we cannot add these functions.
+
+## Constraints
+
+- `libelektra-core` should have a minimal (but still useful) API.
```
markus2330, line 10:
```suggestion
- `libelektra-core` should have a minimal API.
```

doc/decisions/operations_library.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Assumptions
+
+## Considered Alternatives
+
+## Decision
+
+Create a new `libelektra-operations` library.
+This library provides generally useful, but not minimal, functions.
```
markus2330, line 19:
```suggestion
This library provides generally useful and performant, but not minimal, functions.
```

doc/decisions/operations_library.md
```diff
@@ -0,0 +1,50 @@
...
+The implementations should still be as efficient as possible and will therefore access private API directly.
+
+`libelektra-operations` only depends on `libelektra-core`.
+
+Move basic, but not minimal, functions from `libelektra-core` to `libelektra-operations`:
+
+- `keyNeedSync`
+- `keyGetRef`
+- `ksDeepDup`
```
markus2330, line 28: This creates another implication: libelektra-kdb will depend on `libelektra-operations`

markus2330 (changes requested):
I generally agree with having such libs but I think we need to have a second thought how exactly the libs should be named and what their scope is. Maybe also drawing an UML helps here a bit (so that we get an overview).

doc/decisions/iterating_name_parts.md
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * This function defins the "next part" after @p currentPart as:
+ * - the first part of @p key (after namespace), if `currentPart == NULL`
+ * - `NULL`, if @p currentPart is the last part of @p key, i.e. `currentPart == keyBaseName (key)`
+ * - the next part within @p key after the location @p currentPart points to
+ *
+ * To iterate over the parts of a key you can use:
+ * @code{.c}
+ * Key * k;
+ * for (const char * part = keyGetNextPart (k, NULL); part != NULL; part = keyGetNextPart (k, part))
```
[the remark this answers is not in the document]
kodebach: > `keyGetNextPart` for escaped key names would be more interesting, wouldn't it?

You may want the unescaped part for various reasons and implementing it this is a lot easier. The user can always call the escaping function on the returned string. Since the escaped name won't be stored anymore, if we returned the escaped name we would basically do the same anyway.

I actually think, the unescaped name is a lot more useful. It is the actual value of the name part, not just a representation.

> Shouldn't we put all the key name escaping stuff in a separate library? (operations already looks oversized)

I'm not sure what you are referring to here, but all the (un)escaping code has to part of `libelektra-core`. Unless of course the core now only supports unescaped names. That would be pretty big change, but then moving the (un)escaping to a separate library would make sense.

doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,61 @@
+# Operations Library
+
+## Problem
+
+There are many functions that can be useful, but are not required to form the minimal API.
+Since we want to keep the API of `libelektra-core` minimal, we cannot add these functions.
+
+If a function is generally useful, it should be added to [`libelektra-operations`](operations_library.md).
```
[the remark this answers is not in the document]
kodebach: "generally useful" was intended to mean: the functionality is useful in itself, i.e. would be useful to have in any language. I'll clarify this. I also hope this explains the next sentence a bit better.

doc/decisions/lowlevel_library.md
kodebach: > is it only 2?

For now there is not enough proposed functionality to justify more. We can always break apart e.g. `libelektra-operations`  in future. AFAIK this could be backwards compatible, as long as we continue to ship a `libelektra-operations` that simply links to the libraries that actually implement the functions.

However, if the library isn't big enough to justify a split before 1.0, I don't think it will grow so fast that we need to split it, before we have some other reason to go to 2.0. I say this, because there would still be `libelektra-ease` (possibly with a different) for more rarely using things (e.g. topological sort) and others like `libelektra-opts` for specialized things.

> as "highlevel" also is open to bindings

The current `libelektra-highlevel` is really not all that open to bindings. It would make a lot more sense to implement high-level abstractions on top of low-level bindings of other languages. The concepts behind the library (shorter key names, check on open, no fail in get, ...) and maybe even the code-generation could be applied to other languages. But the actual API of `libelektra-highlevel` is really only suitable for C or a similarly basic language. For example, in an object oriented language it would make much more sense to utilise classes. Some cases could even produce an API (with code generation) that utilises the type system, e.g. if it compiles you won't get "key not found" errors.

doc/decisions/operations_library.md
kodebach: In general, I don't see this as a problem.  `libelektra-kdb` is a pretty complex library, it may be quite hard to implement all of its features just based on the minimal `libelektra-core` API. However, if `ksDeepDup` would be the only reason for this dependency, we should probably add `ksDeepDup` to the core.

doc/decisions/iterating_name_parts.md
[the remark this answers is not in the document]
markus2330: > I'm not sure what you are referring to here, but all the (un)escaping code has to part of libelektra-core. Unless of course the core now only supports unescaped names. That would be pretty big change, but then moving the (un)escaping to a separate library would make sense.

Exactly, I am referring to to-be-clarified decisions about if and how (un)escaped of key names should be in core or in an extra library. As these parts also need bindings, it could be libelektra-operations (in the proposed terminology).

The big question is the overall dependencies: at least for some parts of Elektra the core alone must suffice, otherwise the core lib would be useless.

doc/decisions/lowlevel_library.md
markus2330: > For now there is not enough proposed functionality to justify more. We can always break apart e.g. libelektra-operations in future. AFAIK this could be backwards compatible, as long as we continue to ship a libelektra-operations that simply links to the libraries that actually implement the functions.

It is much better to get it right in the first place. Which symbols are in which library is part of the API.

> However, if the library isn't big enough to justify a split before 1.0, I don't think it will grow so fast that we need to split it, before we have some other reason to go to 2.0.

We cannot even do such a big change in 2.0 as it basically breaks every application using these libs. It is also very annoying (speak: deal breaker) for developers to constantly relearn in which header/libs they find which kind of functionality.

> The current libelektra-highlevel is really not all that open to bindings.

Ok, then call it "lowlevel" makes sense. We need to be careful to only have functionality there that is really only useful for C. Maybe we should clarify the name even more by calling them libelektra-lowlevel-c and libelektra-highlevel-c?

kodebach: > Which symbols are in which library is part of the API.

Let's say 1.0 has `libelektra-operations.so` which contains the implementations for `foo` and `bar` and comes with the header `elektra/operations.h`

```c
int foo (int arg);
char * bar(const char * x, int y);
```

Now let's say in 2.0 we decide to split this library. We split `libelektra-operations.so` into `libelektra-foo.so` and `libelektra-bar.so`. `libelektra-foo.so` contains the implementation for `foo` and `libelektra-bar.so` the implementation for `bar`. We also add new headers `elektra/foo.h`

```c
int foo (int arg);
```

and `elektra/bar.h`

```c
char * bar(const char * x, int y);
```

If we only do this, it would be major breaking change and everything that depends on these functions would have to be rebuilt. IMHO this would be fine, since that is basically what breaking change means. However, we could continue to ship  `libelektra-operations.so` and `elektra/operations.h` to avoid the breaking change.

`libelektra-operations.so` would simply have dependencies on `libelektra-foo.so` and `libelektra-bar.so`. The system linker would then load these at startup. Similarly, `elektra/operations.h` would now look like this

```c
#include "foo.h"
#include "bar.h"
```

to avoid source-breaking changes.

> Maybe we should clarify the name even more by calling them libelektra-lowlevel-c and libelektra-highlevel-c?

Yes, that is probably a good idea.

kodebach:
@markus2330 Is the only open question here, how many extra libraries we need? If so, I would propose leaving this open for now and creating an issue to decide before 1.0 That way we can start with the implementation of the new operations library.

markus2330:
We could start the new operations library but nevertheless we should continue with the work to architect how the libraries should look like. Important to consider is, however, that we should focus on cleanup first, and then create new things. Otherwise the new libs will get as messy as the old ones.

kodebach:
We could simply start by moving & merging `keyIsBelow` as proposed in #4346. That would be a sort of cleanup and would get the library started. From there we can see what else we find. I think it is pretty clear now that creating an `operations` library is a goal, so we have to start somewhere.

markus2330:
AFAICS several core functions (e.g. ksCut) depends on keyIsBelow*, so it is probably not a good starting point. We definitely do not want wrong library dependencies from the start and circular dependencies is a strong architectural smell.

doc/decisions/lowlevel_library.md
```diff
@@ -0,0 +1,66 @@
+# Operations Library
```
kodebach, line 1:
```suggestion
# Lowlevel Library
```

kodebach:
> AFAICS several core functions (e.g. ksCut) depends on keyIsBelow*, so it is probably not a good starting point. We definitely do not want wrong library dependencies from the start and circular dependencies is a strong architectural smell.

Yes, `keyIsBelow*` is used in some parts of `libelektra-core` and `libelektra-kdb`. IMO having a dependency from `libelektra-kdb` to the new `libelektra-operations` would be fine. As for `libelektra-core`, I only found references from these functions: `ksCut`, `ksFindHierarchy`, `ksRename`, `keyReplacePrefix`.
None of these, are used anywhere inside `libelektra-core` (except among themselves) and I don't really see why these would be core minimal API. So I would actually move them to `libelektra-operations` as well.

That said, `ksIsBelow` could actually be considered core minimal API and should probably stay in `libelektra-core`. It's is kinda the only thing that (publically) defines the hierarchy of a `KeySet`.

So my new suggestion would be:

1. Start `libelektra-operations`  by moving `ksCut`, `ksFindHierarchy`, `ksRename`, `keyReplacePrefix` there.
2. Rewrite `ksIsBelow` into a single function, but keep it in `libelektra-core`.

markus2330:
> IMO having a dependency from libelektra-kdb to the new libelektra-operations would be fine

Yes, I agree.

> None of these, are used anywhere inside libelektra-core (except among themselves) and I don't really see why these would be core minimal API. So I would actually move them to libelektra-operations as well.

I fully agree!

> So my new suggestion would be

Perfect, please update the decision (if not yet done).

Please always ask for review/write status reports in the decision PRs, so that I know what to review next.

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+
+  **Prefix:** `elektraPlugin*`
+  (Note: previously `libelektra-plugin`)
+
+- `libelektra-plugin`:
+  The API for _interacting with_ plugins.
+  This includes loading plugins and calling exported functions.
+
+  **Prefix:**: `elektraPlugin*`
+  (Note: previously `libelektra-invoke` & `loader`)
```
markus2330, line 56: loader was part of kdb. Do you want to use the invoke lib within kdb?

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+
+  **Names:** `elektra<TYPE>ToString` and `elektraKeyTo<TYPE>`
+  (Note: extracted from `libelektra-ease`)
+
+- `libelektra-ease`:
+  A collection of various other APIs that help when interacting with `ElektraKey` and `ElektraKeyset`.
+  Contains APIs that are not large, specific, or complex enough to justify a separate library and not minimal enough for `libelektra-core`, but can be implemented on top of the public API.
+  The APIs should not specifically target C and should be usable via bindings (if appropriate for the other language).
+
+  **Prefix:** `elektra*`
```
markus2330, line 74: I would prefer if we can be consistent and always have (several) specific prefixes for all libs.

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+  (Note: rest of old `libelektra-ease`, merged with `libelektra-meta` & `libelektra-utility`; cleanup needed)
+
+- `libelektra-operations`:
+  Like `libelektra-ease`, this library contains various additional `ElektraKey` and `ElektraKeyset` APIs beyond the minimal API.
+  However, these APIs specifically need access to private APIs of `libelektra-core` to be implemented (efficiently).
+  The APIs should not specifically target C and should be usable via bindings (if appropriate for the other language).
+  All APIs must abstract over unstable details of private APIs and provide a stable API, or **very clearly** state the stability restrictions.
+
+  **Prefix:** `elektra*`
+  (Note: rest of `libelektra-ease`, merged with `libelektra-meta` & `libelektra-utility`, plus some stuff from old core; cleanup needed)
```
markus2330: Any specific reason you want to get rid of `libelektra-meta`?

Did you consider to introduce `libelektra-array` for dealing with arrays?

markus2330 (commented):
Great job! A big thanks for starting the discussion how the libraries should be! :rocket:

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+- `libelektra-merge`:
+  The API for merging and/or detecting conflicts between two keysets.
+
+  **Prefix:** `elektraMerge`
+
+- `libelektra-pluginbase`:
+  The API _used by_ plugins.
+
+  **Prefix:** `elektraPlugin*`
+  (Note: previously `libelektra-plugin`)
```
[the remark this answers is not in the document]
kodebach: I actually think the name `libelektra-pluginbase` makes more sense, because it forms the base for all plugins. It contains functions like `elektraPluginExport` and `elektraPluginGetData`.

The `libelektra-plugin` meanwhile is for interacting with plugins (not just from tooling) so the name makes sense too. It will contain things like `elektraPluginCall2Arg` to call a function from a plugin, possibly from another one.

doc/dev/libraries.md
kodebach: I wasn't sure about that part. Either we depend on this library from `libelektra-kdb`, or we build the 1 or 2 loader functions into a static library that is linked into both `libelektra-kdb` and this library. Since the loader library should be very small in the `BUILD_SHARED` variant, I think this duplication would be acceptable. In the `BUILD_STATIC` and `BUILD_FULL` variants everything is bundled into one library anyway.

doc/dev/libraries.md
kodebach: Maybe as a compromise: 

- `libelektra-ease` **Prefix:** `elektraEase*`
- `libelektra-operations` **Prefix:** `elektraOps*`, or `elektraKey*`/`elektraKeyset*` when it doesn't conflict with `libelektra-core`
- `libelektra-lowlevel-c` **Prefix:** `elektraLlc*`, or `elektraKey*`/`elektraKeyset*` when it doesn't conflict with `libelektra-core`

The reason why I would allow `elektraKey*`/`elektraKeyset*` in `libelektra-operations` and `libelektra-lowlevel-c` are things like `elektraKeysetCut`. Calling it `elektraOpsKeysetCut` would hurt discoverability and development experience. It is much nicer IMO, if everything that does some sort of concrete operation on a `Key`/`Keyset` has the same prefix.

There shouldn't ever be a name conflict, because if a function is minimal it should be in the core, otherwise it is not. And between operations and lowlevel-c the conflict is resolved by looking at whether the function is useful outside the C world.

doc/dev/libraries.md
kodebach: `libelektra-meta` doesn't really have many useful functions IMO (I'd remove some of them like `key*Comment`). I also didn't find a good way to define what it is. AFAICT it's the same as ease but limited to metadata, which isn't a good enough distinction IMO.

> Did you consider to introduce `libelektra-array` for dealing with arrays?

Could be worth considering, but we'd have to extend the current API (or at least plan to) for that to make sense. With the current 6 functions (1 of which is redundant) some one which aren't that generally useful, a separate library doesn't make much sense to me.

doc/dev/libraries.md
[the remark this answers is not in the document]
markus2330: :+1: Ok, this makes sense. Please also add it to [doc/dev/libraries.md](https://github.com/ElektraInitiative/libelektra/pull/4245/files/6141c6adc9a7a166451f3292e3025db0c4ef3f48..44e6293f42f57d8dba0b397e01b0b61f5be0ccb1#diff-db061ad3ca3bb4e7f8806c2a6339502e77689bd7ecc8aa2bd79ba94345fb0948)

doc/dev/libraries.md
markus2330: I agree, the duplication makes sense in this case.

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+  **Names:** `elektra<TYPE>ToString` and `elektraKeyTo<TYPE>`
+  (Note: extracted from `libelektra-ease`)
+
+- `libelektra-ease`:
+  A collection of various other APIs that help when interacting with `ElektraKey` and `ElektraKeyset`.
+  Contains APIs that are not large, specific, or complex enough to justify a separate library and not minimal enough for `libelektra-core`, but can be implemented on top of the public API.
+  The APIs should not specifically target C and should be usable via bindings (if appropriate for the other language).
+
+  **Prefix:** `elektra*`
+  (Note: rest of old `libelektra-ease`, merged with `libelektra-meta` & `libelektra-utility`; cleanup needed)
```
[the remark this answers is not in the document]
markus2330: I would keep libelektra-core standalone (even if there might be some trivial function that might be useful within core).

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+
+- `libelektra-operations`:
+  Like `libelektra-ease`, this library contains various additional `ElektraKey` and `ElektraKeyset` APIs beyond the minimal API.
+  However, these APIs specifically need access to private APIs of `libelektra-core` to be implemented (efficiently).
+  The APIs should not specifically target C and should be usable via bindings (if appropriate for the other language).
+  All APIs must abstract over unstable details of private APIs and provide a stable API, or **very clearly** state the stability restrictions.
```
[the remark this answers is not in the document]
markus2330: > if a function has restrictions on stability, these must be clearly stated. This could simply be

It cannot be on function level. Maintainers expect that libraries with the same SOVERSION **are** compatible. So we can only have libraries that might be incompatible (SOVERSION might get pumped from time to time).

For things that are refactored by you, I wouldn't keep the option to pump SOVERSION as this is a **major** annoyance to users. Users want stability. They want their applications to continue working. This is our #1 goal, see doc/GOALS.md

> That's why https://github.com/ElektraInitiative/libelektra/pull/4246 proposes different headers. Then users need to explicitly opt in to non-stable APIs.

I agree that it must be very clear for users if there are any non-stable APIs. Can you clarify this in the decisions, please? Ideally we don't have any unstable libraries, though. Thus we really should watch out to make everything as simple&minimal as possible. Otherwise it would be impossible to make a beautiful API which lasts long.

doc/dev/libraries.md
markus2330: Yes, probably array&meta can go to ease (if it doesn't get too bloated then).

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+  **Prefix:**: `elektraPlugin*`
+  (Note: previously `libelektra-invoke` & `loader`)
+
+- `libelektra-mount`:
+  The API for manipulating mountpoints.
+
+  **Prefix:** `elektraMount*`
+
+- `libelektra-type`:
+  The API that defines Elektra's type system. (Alternative: The API that defines the type system used by `libelektra-highlevel-c`.)
```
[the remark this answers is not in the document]
kodebach: The keeping in sync part is mostly about the exact "string-to-type" conversion logic. If we find bugs or problems, there could be a need to change edge cases. This has to then be adapted to the various bindings.

doc/dev/libraries.md
kodebach: > If we link `libelektra-operations` by default

We can't really define what gets linked that's on the user side. But yes, I would include `libelektra-operations` (and maybe `libelektra-lowlevel-c`) in the pkgconfig and CMake configs that we provide.

> I would rather make exceptions for the high-level API, as this API is much more user-facing then these low level libraries.

Yes, that's why I didn't really like the `elektraHlc*` prefix. But maybe we can use `elektraConf*` for the high-level API. That's still a pretty good prefix for use in applications. We definitely need a prefix beyond `elektra*`, because otherwise it becomes hard to see what is part of the high-level API (i.e. intended for applications) and what is more low-level.

doc/dev/libraries.md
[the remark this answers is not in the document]
kodebach: It's not as simple as "this library is stable/unstable". We already agreed that e.g. `struct _Key` cannot be fully private (needed by `mmap`, will be used directly by `libelektra-operations` etc.), but also can't really be stable API. If we made it stable API there would be no point to hide the layout at all and we could just let users access `struct _Key` directly.

So their are certain things that are technically public (i.e. exported), but we don't consider public API and therefore don't provide stability guarantees.

A similar example would be

```c
// in elektra/foo.h
elektraFooMainFunc();

// in elektra/foo/unstable.h
elektraFooDoSomething();
elektraFooDoMoreStuff();
```

the public API of `libelektra-foo` is just `elektraFooMainFunc()`. If you only call that you get all the stability guarantees you expect.

However, for whatever reason (e.g. special use by other libs, or maybe new APIs that need more testing before we make them stable), `libelektra-foo` also exports the functions `elektraFooDoSomething()` and `elektraFooDoMoreStuff()`. Probably, these are part of the implementation of `elektraFooMainFunc()`. If you call `elektraFooDoSomething()` or `elektraFooDoMoreStuff()`, things may change, things may break, the functions may even be removed entirely, you need to deal with that.

Of course, if there is public API in another library of Elektra e.g.

```c
// in elektra/bar.h
elektraBarMainFunc();
```

that calls `elektraFooDoSomething()` or `elektraFooDoMoreStuff()` then you can call `elektraBarMainFunc` without any issues. `elektraBarMainFunc` is public stable API and we will make sure things don't break, no matter how it is implemented and what other APIs it calls.

> Users want stability. They want their applications to continue working. 

And that's why 99.999% of applications should only ever use the stable public API. But if you can tie your application to a specific version of Elektra (e.g. fully integrated embedded system) or your application isn't public and will always be updated when things break, then you could use the unstable APIs too.

> we really should watch out to make everything as simple&minimal as possible

"as possible" is doing a lot of work in that sentence... What is possible?

IMO the core should be small, so that when somebody needs it they can go to the bare minimum.

But for most other things, I think usability and developer experience is much more important than being minimal. No one will use a library, where you have to make 6 API calls to do a common task, when there could just be a single API. If the library get's used at all, people will create wrapper libraries to simplify things and that is IMO a failure of API design in the underlying library.

> beautiful API

APIs don't need to be beautiful; they need to be functional, usable, safe, etc. But I get want you mean.

---

With this system problems only come up in two cases:
1. You mix and match Elektra's `.so` files between `x.y.z` versions. That's clearly a broken install and shouldn't be supported.
2. Some external library/application depends on Elektra and used an unstable API, but wasn't updated for the newest release. That's a problem with the library/application and the author should avoid the unstable APIs.

doc/dev/libraries.md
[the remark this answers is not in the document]
markus2330: Yes, floating point parsing is also not so trivial to implement.

doc/dev/libraries.md
markus2330: >  We can't really define what gets linked that's on the user side. But yes, I would include libelektra-operations (and maybe libelektra-lowlevel-c) in the pkgconfig and CMake configs that we provide.

Yes, I mean the pkgconfig and CMake configs to be "default". Is there anyone not using them?

> elektraConf* 

Yes, that would be a possibility.

doc/dev/libraries.md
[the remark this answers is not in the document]
markus2330: > It's not as simple as "this library is stable/unstable".

Then **we** need to make it simpler. In the end of the day maintainers/application developer trust that libraries without changes in SONAME are compatible. Any other rules we might have can only be of internal nature (within the Elektra initiative).

> We already agreed that e.g. struct _Key cannot be fully private (needed by mmap, will be used directly by libelektra-operations etc.)

- in case of mmap cache files there is version information. Caches simply get discarded when `struct _Key` changes.
- libelektra-operations (and probably also other libs+plugins like "cache") is simply bound to be exactly the same version as libelektra-core. This can easily be done in package managers as they are both from the same source package.

> "as possible" is doing a lot of work in that sentence... What is possible?

Remove/Cleanup as much as possible for 1.0, see  #2987

> create wrapper libraries to simplify things and that is IMO a failure of API design in the underlying library.

I don't think it is a big deal in general. There are often some requirements like how to do threading, different naming conventions etc. that require wrappers. I often wrap KDB, parentKey and main KeySet to one struct to pass it around more easily. The wrapping code is super-small and written in a few minutes, so I would not worry about that for our low-level APIs. Only in more high-level APIs people of course expect that simple things (like getting one value) can be done with a few calls.

> With this system problems only come up in two cases:

There is a third:

3. We have stable libraries that exactly follow semantic versioning/SONAME (no breaks **at all**, i.e. ONLY what symbol versioning allows). We have other libs that are a bit more experimental. If people use them, they might get problems. We don't encourage to include such header files or to link against such libraries until we get them out of the experimental stage.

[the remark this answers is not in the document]
kodebach: > There are often some requirements like how to do threading, different naming conventions etc. that require wrappers.

Those are fine. These are things that Elektra leaves up to the user intentionally. The bad wrappers are those that work around shortcomings, inconveniences and other issues in Elektra's API.

> I often wrap KDB, parentKey and main KeySet to one struct to pass it around more easily.

If you do something like that _often_, then I would consider that a serious problem with Elektra's API. But in this case we already have a solution. The highlevel API has exactly this struct.

> The wrapping code is super-small and written in a few minutes

But every time somebody write one of these wrappers they make additional assumptions and muddy the use case of Elektra. It also makes Hyrum's Law worse.

> Then **we** need to make it simpler.

That's not how things work. We want to keep `struct _Key` private, but we want to have stuff outside `libelektra-core` that as access to `struct _Key`. Ergo we need to make `struct _Key` technically public, but cannot guarantee that it remains the same.

> In the end of the day maintainers/application developer trust that libraries without changes in SONAME are compatible.

Sure, but that's easy to fix. We just need to make sure that our SONAME changes when we break something. The easiest (but least useful) way is to include the full x.y.z version in the SONAME. 

> There is a third: [...]

Not sure how what you described is a problem that could happen with my proposed header/versioning system. IMO what you describe is basically the same as what I proposed by less detailed.

> libelektra-operations (and probably also other libs+plugins like "cache") is simply bound to be exactly the same version as libelektra-core.

I'd say this should apply to everything.

So the rules would be:

1. _Everything_ in this repo will be released together with the same version numbers.
2. Semantic versioning guarantees only apply between this repo and the outside, but not within this repo. An application built for `libelektra-core` v1.0.0 will work with `libelektra-core` v1.x.y. But any `libelektra-ABC` va.b.c _only works_ with `libelektra-XYZ` va.b.c. In other words, all of Elektra's packages must be installed with the same version. Like you said, any proper package manager setup would already enforce this.
3. Some libraries have headers `elektra/foo/unstable.h` that are exempt from semantic versioning.

doc/dev/libraries.md
kodebach: > Is there anyone not using them?

Who knows what kinds of proprietary internal build systems are out there.

To come back to the original issue: We want to keep `elektraKey*` and `elektraKeyset*` exclusive to `libelektra-core` while having a nice convention for `libelektra-operations`. How about using `elektraXKey*` and `elektraXKeyset*` (`X` for extension/extra)?

For potential other functions in `libelektra-operations` that are not extensions of the Key or Keyset APIs, I'd simply propose `elektraOps*`.

kodebach:
@markus2330 Summary of open questions:

1. [Prefix(es) for `libelektra-operations`](https://github.com/ElektraInitiative/libelektra/pull/4245#discussion_r953662047): I proposed `elektraXKey`, `elektraXKeyset` and `elektraOps`.
2. [Library stability](https://github.com/ElektraInitiative/libelektra/pull/4245#discussion_r935471511): I proposed these rules:
   > 1. Everything in this repo will be released together with the same version numbers.
   > 2. Semantic versioning guarantees only apply between this repo and the outside, but not within this repo. An application built for libelektra-core v1.0.0 will work with libelektra-core v1.x.y. But any libelektra-ABC va.b.c only works with libelektra-XYZ va.b.c. In other words, all of Elektra's packages must be installed with the same version. Any proper package manager setup should already enforce this.
    > 3. Some libraries have headers `elektra/foo/unstable.h` that are exempt from semantic versioning.

doc/dev/libraries.md
```diff
@@ -0,0 +1,84 @@
...
+  (Note: previously `libelektra-plugin`)
+
+- `libelektra-plugin`:
+  The API for _interacting with_ plugins.
+  This includes loading plugins and calling exported functions.
+
+  **Prefix:**: `elektraPlugin*`
+  (Note: previously `libelektra-invoke` & `loader`)
+
+- `libelektra-mount`:
```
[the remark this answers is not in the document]
kodebach: I think just `libelektra-mount` makes more sense. It is not certain that only "tools" need this. For example, an application may wish to create its own mountpoint at first startup. This use case would also need this library, even though the application is not a tool.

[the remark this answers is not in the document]
markus2330: Yes, in certain embedded contexts (where applications are run as root etc.) it might be like this.

So let us stay with `libelektra-mount`.

markus2330:
Thank you for continuing work on the decsision! :heart: 

> elektraE*

I am not a big fan of one-letter shortenings but if we do it, we should use it consistently for all libs.

In particular the capitalization is confusing, e.g. "elektraEKey" is misleading. elektra_e_ks would be better...

What I could imagine is that the prefix is elektrax, where x is a consonant (not a vowel, i.e., a, e, i, o, u) for each lib.
E.g. elektraxKey would be imho easier to read.

What exactly was wrong with elektraEase as prefix?

> elektraX*

Again, elektrax or elektra_x would imho look better. Furthermore, imho the library name  should somehow match with the prefix. So if we use "x" for extension, the lib should be called libelektra-extension.

>   Everything in this repo will be released together with the same version numbers.
>   Semantic versioning guarantees only apply between this repo and the outside, but not within this repo. An application built for libelektra-core v1.0.0 will work with libelektra-core v1.x.y. But any libelektra-ABC va.b.c only works with libelektra-XYZ va.b.c. In other words, all of Elektra's packages must be installed with the same version. Any proper package manager setup should already enforce this.

Yes, fully agree (the word "already" should be removed, though). Can you clarify this in doc/VERSION.md?

> Some libraries have headers elektra/foo/unstable.h that are exempt from semantic versioning.

This is not how reliable software can be. You can have private/unstable libraries but not private/unstable header files which are part of a seemingly stable library. The library would stop being stable if it provides such an interface. The reason is: Maintainers only know which libraries (in which versions) link against which applications, as this is information you have without studying the source code. On library upgrades, they must be able to safely assume that all applications will continue to work. (Unless the SOVERSION changed, which we do not want to do, at least within Elektra 1.*.)

kodebach:
> elektra_e_ks would be better

That would be a totally different naming scheme than we have right now...

> What I could imagine is that the prefix is elektrax, where x is a consonant (not a vowel, i.e., a, e, i, o, u) for each lib.

That would 1) be very limiting 2) could become quite confusing.

If you like it better, we can use `elektrax` instead of `elektraX` for `libelektra-operations`, but capitalising the X would be consistent with the camel case rules.

> What exactly was wrong with elektraEase as prefix?

I thought it could result in awkward and annoying names, depending on what functions end up in this library. It could happen that a function is not minimal (i.e. not in the core), but still very useful. With the rules currently described in the PR such a function would end up in ease, if it can be implemented with the public API. I could live with an additional E/X in the name, but adding always adding `Ease` sounds really awkward to me.

Maybe the solution is to change the rules of when something goes into `libelektra-operations` and when it goes into `libelektra-ease` (do we really need two libs?). Then the more common "extensions" could all use `elektraX*` and `elektraEase*` is only for rarely used stuff like SHA hashes or topological sorts.

> This is not how reliable software can be. [...] Maintainers only know which libraries (in which versions) link against which applications, as this is information you have without studying the source code.

That's not true, the one crucial piece that (distro/package) maintainers _do_ have is documentation from application developers. In that documentation application developers should state, if they access some private APIs and depend on specific versions of Elektra (not encouraged).

> On library upgrades, they must be able to safely assume that all applications will continue to work.

You _cannot_ assume that without also making some assumptions about what application developers do and don't do. Application developers could always go in and manipulate the contents of a `struct _Key` directly if they wanted to, even though we don't consider that public/stable API. There is nothing stopping them and things will break, if we change `struct _Key`.

If you want something that is super stable, you have to make _everything_ (including the contents and layout of all structs) part of the public API. But then why even have a `keyNew` function? Why return allocated `Key *`? You could just make things easier for everybody and let the API caller allocate everything (it would definitely be more minimal).


The important part about the `unstable.h` headers, however, is that I never intended for them to be used in applications. At best they might be used in external libraries, which should then clearly state that they need specific versions of Elektra to work, or take some other precaution to still present a stable API themselves. But mostly these headers should be used for cross-dependencies between Elektra's libraries. They could also be useful to introduce new APIs which may need some more work, before the can be fixed forever and made stable. Maybe the solution is to ship these headers in a separate `elektra-unstable` package that need to be installed separately. That way it should be very clear that using these could cause problems.

markus2330:
> I thought it could result in awkward and annoying names

I think you only need to get used to it. E.g. at the first look, the "elektra" prefixes were also surprising for me.

> goes into libelektra-operations and when it goes into libelektra-ease (do we really need two libs?)

I think their goals are quite different, so two libs make sense. "operations" really extends performant functionality, "ease" only makes some tasks easier to implement. From library size alone, a split is probably not needed (ease is only 1/3 of core or kdb).

Of course everything depends on a nice architecture of these two libs. Ideally, there should be many use cases that need only **either** operations or ease. (If everyone needs none or both, two libs obviously don't make sense.)

> Then the more common "extensions"

This doesn't work as you often don't know what is/will be common. What is common in some application/plugins might be esoteric in others.

>> This is not how reliable software can be. [...] Maintainers only know which libraries (in which versions) link against which applications, as this is information you have without studying the source code.

>  That's not true, the one crucial piece that (distro/package) maintainers do have is documentation from application developers. In that documentation application developers should state, if they access some private APIs and depend on specific versions of Elektra (not encouraged).

It would only work if **all** applications consistently document the use of private APIs (which is kind of shameful, as everyone knows you shouldn't use them) and all maintainers read this documentation. I personally never saw any application stating that it uses a private API of some lib.

What I already saw often (e.g. fontconfig, glibc, ...) is that libraries do incompatible changes without pumping SOVERSION. This is a recipe for total disaster because it is a serious bug in production code (usually segfaults etc.) that needs to be fixed in every application using the lib. We always need to remember that Elektra serves a purpose: it should make configuration easier. If our APIs are only slightly unstable, applications are **much** better off not using Elektra at all.

> If you want something that is super stable, you have to make everything (including the contents and layout of all structs) part of the public API.

No, actually you have to do the contrary: We have to shift around the contents of our private structs from time to time. Then nobody can rely on it. This is what Linux does with its internal ABIs. Unfortunately, this would be bad for our cache (but only once after upgrade).

kodebach:
> I think you only need to get used to it. E.g. at the first look, the "elektra" prefixes were also surprising for me.

(while writing the code) I think "elektra" is different. I can "ignore" that part pretty easily since I know I'm working with Elektra. But a secondary prefix for the individual library is different. IMO there are two case here: 1) It's obvious in which library a function lives so don't have to think about the prefix (e.g. "opts") 2) The function has no clear library it belongs to so I have to actively remember the correct prefix for a function (e.g. "ease" vs "operations")

Once the code is written, yes, "elektra" and another prefix are pretty much the same and I just ignore them when reading the code, unless the prefix actually contributes to the name of a function (e.g. "plugin").

> Ideally, there should be many use cases that need only either operations or ease. (If everyone needs none or both, two libs obviously don't make sense.)

I think this will be pretty hard to achieve, if we just go by: "operations" is for performance stuff or needs private API. That's why I originally suggested the somewhat nebulous approach of:

"operations" is for common extensions beyond the minimal API, "ease" is for rarely used stuff, (and to complete the set: "lowlevel-c" is for C specific things)

> This doesn't work as you often don't know what is/will be common. What is common in some application/plugins might be esoteric in others.

To some extent, we can estimate how common a use case will be. For example, `ksCut` is a much more common use case, than "sha256 hash of a KeySet". Even if we don't know the exact use cases. There will of course be cases where it is hard to decide, but in those cases I'd just go by function size. Simple, small functions go into "operations" since it is more likely that there is already a dependency on that one.

> use of private APIs (which is kind of shameful, as everyone knows you shouldn't use them

This is why I never called these APIs "private". We do also have private APIs that should definitely not be used. But these "unstable" APIs are just that. APIs without stability guarantees. We know they are public, you are not doing something naughty if you use them, but you have to know the risk and won't get all the stability.

> If our APIs are only slightly unstable, applications are much better off not using Elektra at all.

I think you are taking things out of proportion. The main API is still stable. The "unstable" headers would just be additional APIs. It must always be possible to use Elektra sensibly without relying on these unstable APIs.

kodebach:
@markus2330 ping

markus2330:
> The function has no clear library it belongs to so I have to actively remember the correct prefix for a function (e.g. "ease" vs "operations")

It is our responsibility to avoid this case. It should be clear to which lib some feature belongs. We need to improve the library design if it is not yet clear.

> "operations" is for common extensions beyond the minimal API

This description is probably too generic (it calls for overlap with others). Maybe we should restrict it to KeySet (data structure) operations, like union, cut, append (a whole KeySet).

> "ease" is for rarely used stuff

As already said we cannot use rare/often as distinguish feature as this is very dependent on use-case (see also below). Better criteria are topics (like cryptography), minimal/optional (minimal interface in core, other stuff outside) or technical criteria (like needing POSIX interfaces)

Btw. obviously, there should be use cases for everything in every library.

> (and to complete the set: "lowlevel-c" is for C specific things)

> ksCut is a much more common use case, than "sha256 hash of a KeySet"

I disagree. It can easily imagine some application or plugin where sha256 hash of KeySet(s) is very important and ksCut is not needed.

> But these "unstable" APIs are just that. APIs without stability guarantees.

Realistically there always will be some unstable APIs. This is acceptable as long as they are in unstable libraries.

>> If our APIs are only slightly unstable, applications are much better off not using Elektra at all.

> I think you are taking things out of proportion.

I mean it exactly as said. I wouldn't want to contribute to a library which doesn't have stability as important goal. If there is any weak link, there is no stability at all. It doesn't matter if an application crashes because one unstable function changed or the whole API was redesigned: the user experience for application developers is terrible. (Of course with constant redesigns even worse as you effect every application.) Elektra was one of these libraries being terrible to application developers (especially 0.1 -> 0.6) and as a result lost every user in that period.

doc/dev/libraries.md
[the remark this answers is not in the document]
markus2330: > We want to keep struct _Key private, but we want to have stuff outside libelektra-core that as access to struct _Key.

It might be outside of core but inside of Elektra.

> Ergo we need to make struct _Key technically public, but cannot guarantee that it remains the same.

No, we do not "need to". We could and probably should not even install the header describing how _Key looks like.

> Sure, but that's easy to fix. We just need to make sure that our SONAME changes when we break something. The easiest (but least useful) way is to include the full x.y.z version in the SONAME.

Exactly.

>  I'd say this should apply to everything.

Yes, I agree. An inside/outside Elektra world-view simplifies everything a lot.

> Everything in this repo will be released together with the same version numbers.

:+1:  Can you please add this do doc/VERSION.md?

> Some libraries have headers elektra/foo/unstable.h that are exempt from semantic versioning.

See also  #4245. We might have unstable libraries with SOVERSION pumps from time to time (ideally we won't but this is a compromise I am willing to take) but we would only create (huge) troubles if we do stuff nobody expects. These "unstable.h" things also do not solve a real problem. As long as we do not know how to offer some functionality, it is better not to offer it, and especially not in our core libraries.

kodebach:
AFAICT there are two issues we still disagree on. First the library split and the prefixes used and second the unstable APIs. I'll give answers for both, but I'd prefer if we could first focus on the prefix and the library split. Once we've agreed on that we can discuss what to do with unstable APIs. Discussing both at the same time will just lead to chaos. It should also keep future comments shorter, if we're just discussing one thing.

### Unstable APIs (for future discussion, ignore for now)

> Realistically there always will be some unstable APIs. This is acceptable as long as they are in unstable libraries.

That is just not how things work. To reiterate once again. 

1. `libelektra-core` **must** define `struct _Key` and `struct _KeySet`, otherwise it's not standalone
2. We want `libelektra-core` to be minimal, but (some) functions  **outside** of `libelektra-core` should still have access to the contents of `struct _Key` and `struct _KeySet`; for various reasons like performance and mmap
3. We don't want the contents of `struct _Key` and `struct _KeySet` to fall under stable public API guarantees, i.e. changing the contents should **not** require increasing the major version

Based on these constraints I only see two basic options A) break a constraint (1-3 below) or B) do what I proposed all along (4 below).

1. Put `struct _Key` and `struct _KeySet` into a separate _unstable library_. That would mean `libelektra-core` is no longer standalone and has a dependency on this new library.
2. Put everything that has access to `struct _Key` and `struct _KeySet` into `libelektra-core`. That would mean `libelektra-core` is not minimal anymore. We'd also have to do the same thing for any similar situation and `mmapstorage` wouldn't be allowed at all, unless `libelektra-core` also acts as a plugin itself.
3. Make `struct _Key` and `struct _KeySet` public stable API and increase the major version every time we change something. However, at this point we should really change the entire API to allow keys and keysets to be allocated by user code and not just via `keyNew`/`ksNew`.
4. Accept that some libraries have both a stable API (defined in one set of headers) and an unstable API (defined in another set of headers).

> **Note**: If it makes you more comfortable replace "unstable" with "internal". I just used the term "unstable" as "not stable but public". I'm also only using the term "stable" in terms of API guarantees here. If a function is not "stable" (i.e. "unstable") it doesn't mean that it is more likely to have bugs.

Another benefit of my proposed approach is that you can have a sort of experimental alpha/beta API. If we in introduce a new function `foo`, but we're not sure about the API yet, we can put it into the unstable header of a library. We can then start using it internally and maybe in some test apps. External people could use it too, but of course with the risk that they might need to change their code. When we like the API, we can release the stable `foo` simply by moving it from the unstable header to the stable one. It's not a breaking change, because we only break the unstable header, which we don't care about. Of course it could have been in a separate library and the same would be true, but then releasing `foo` means more work for users. If it just moves from one header to another, any thing that's complied already keeps working. In the source code, you might need to change the include, but it's likely you've already included the public header of `foo`s library anyway. You could also always include both headers. When making `foo` stable means moving it to a different library, you definitely need to recompile/re-link. 

> Elektra was one of these libraries being terrible to application developers (especially 0.1 -> 0.6) and as a result lost every user in that period.

Arguable a library that is still in version 0.1 to 0.6 (i.e. pre 1.0) shouldn't have many users. Such a library is clearly still in under development and it's developers weren't comfortable enough to say you can now rely on this and we won't change things out from under you.

### Prefixes/Library split

> It is our responsibility to avoid this case. It should be clear to which lib some feature belongs. We need to improve the library design if it is not yet clear.

It's not so much about it being clear where a function belongs. For `elektraKeysetCut` it is kinda obvious that it can only in `libelektra-core` or `libelektra-operations` depending on whether it is minimal API. But that's also kind of the point. We are adding an extra unnecessary prefix to make it `elektraOperationsKeysetCut` just because C doesn't have proper namespaces or modules.

That's why I like `elektraXKeysetCut`. `elektraKeyset*` is minimal API, `elektraXKeyset` is extra stuff. You can ignore the X really easily compared to "Operations". And if you want to be specific saying "Elektra X Keyset Cut" is much easier than "Elektra Operations Keyset Cut".

> I disagree. It can easily imagine some application or plugin where sha256 hash of KeySet(s) is very important and ksCut is not needed.

Of course you can imagine a plugin that does this. But is it really something that many plugins would do? However, I agree that discussing about how often something will be used is pointless.

My point was more that to use `calculateSpecificationToken` (SHA256 hash of keyset) you need a much more specific situation than for `ksCut`. That's simply because SHA256 is defined for byte sequences, so `calculateSpecificationToken` needs to make various assumptions about turning a `KeySet` into the byte sequence that will be hashed.

> Maybe we should restrict it to KeySet (data structure) operations, like union, cut, append (a whole KeySet).

I think I get the idea, but I don't really like the definition. It is very hard to say that is and what isn't a "data structure operation". 
Also we shouldn't exclude stuff that just operates on Key(s), like `keyIsBelow`.

This is probably still too general, but maybe we should aim for this:

- `libelektra-operations` contains extensions the the `Key` and `KeySet` API. These functions could be part of a larger `libelektra-core`, but are not, because we do not consider them minimal API.

I think this is similar to what yo mean by "data structure operation", I just don't know how really define this notion without being vague or relying on arbitrary examples.

As for `libelektra-ease`: Honestly, I have no idea how to define that. This lib has always been a bit of random pile of stuff. It's kind of our catch-all for things that don't fit elsewhere and aren't big or important enough for their own lib. However, one distinction I would definitely make is that things like "SHA hash of a KeySet", which don't really make sense in the world just `Key` and `KeySet` should never be in `libelektra-operations`. I think this also falls under your notion of "data structure operation".

doc/dev/libraries.md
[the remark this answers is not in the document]
kodebach: > See also https://github.com/ElektraInitiative/libelektra/pull/4245.

This is #4245... I assume you mean [this comment](https://github.com/ElektraInitiative/libelektra/pull/4245#issuecomment-1247818411).

As stated in [my response](https://github.com/ElektraInitiative/libelektra/pull/4245#issuecomment-1247923112), let's focus on the prefix and library split question first.

markus2330:
Nice trick to reiterate the same argument again and say that nothing should be replied to it :wink: Discussions get much easier if we stop making long treatises outside of the decision text. If you want separate discussion (and I would also would like it as obviously stating a main goal is not enough if there is no decision backing it up), please put it to different decisions (new PRs) or at least separate discussion threads.

There is a fundamental misunderstanding at the beginning of your argumentation: It is not a library that defines how structs look like but header files define the structs. Header files that do not get installed, are not part of the API. So we simply do not install these header files and use them only internally.

(Btw. if there is really a **need**, and I doubt it, we could provide some serialize functions for Key and KeySet so that people could write something like mmap-storages. But nobody will do it, as writing a cache probably is, together with deciding on the API, the most difficult piece of Elektra.)

> It's not so much about it being clear where a function belongs.

The purpose of this decision exactly is to find an architecture where it becomes clear! Long time ago it was also not so clear what belongs to "Key" and "KeySet". E.g. the iteration was part of "Key", which was obviously non-sense. See keyNext: https://manpages.debian.org/wheezy/elektra-doc/keyset.3.en.html

So our goal here is simply to avoid making non-nonsensical decisions, easy, isn't it? :wink: 

Btw. maybe we should split "core" to libelektra-key and libelektra-keyset, and maybe we simply should not have libelektra-operations, as even with all the proposed keyset operations, libelektra-keyset is probably still smaller than libelektra-key. And maybe we should also not have libelektra-ease but libraries of concrete topics where it is clear what belongs there.

I agree that with core/extensions there is easily some tension and confusion what should be where. I absolutely agree with your concerns on that.

kodebach:
> Nice trick to reiterate the same argument again and say that nothing should be replied to it 

To clarify: I had written the response already when I noticed how long it was. I didn't want my thoughts to get lost, but also wanted to "simplify" the discussion process. Yes, I could have created a separate PR, but that would just create even more parallel discussions. I wanted to tackle one thing after the other...

I also want to mention, the structs were just an example. I used them because this is a known issue. There are lots of other things currently in `kdbprivate.h`. Many of the functions are just "the API we needed for X", making those things public and stable with all the semver consequences could easily lead to problems. I won't go into the unstable stuff any further to avoid opening the discussion again.

> > It's not so much about it being clear where a function belongs.
> 
> The purpose of this decision exactly is to find an architecture where it becomes clear!

With "it" I referred to my statement "The function has no clear library it belongs to so I have to actively remember the correct prefix for a function", which I wanted to clarify.

Yes, the purpose of the decision is exactly to clarify these things. I know that, we agree on that. That's not the issue. The issue was that you seemingly misunderstood what wanted to say and now seemingly misunderstood my attempt at clarification.

> Btw. maybe we should split "core" to libelektra-key and libelektra-keyset,

Would that even be possible? Every key contains a metadata keyset and a keyset obviously contains keys. I think this would create a circular dependency and even if it doesn't right now, it certainly could affect future API.

Also, I don't see any use case where I'd want to use a `Key` without using a `KeySet`, so why split it into two libraries that are always used together.

> with all the proposed keyset operations, libelektra-keyset is probably still smaller than libelektra-key

I'm not sure about that keysets are complex. The hashmap stuff alone is quite a bit of code.

Also if I understand you correctly, this would essentially mean abandoning the idea of a minimal API. I'm not against that, but just to be sure I understand you.

> And maybe we should also not have libelektra-ease but libraries of concrete topics where it is clear what belongs there.

That part I agree with in theory, but that would probably lead to quite a few libraries with just one or two very basic functions, because they don't fit anywhere else.

markus2330:
> I wanted to tackle one thing after the other...

I welcome this approach. If you write in a PR/issue that it should be handled after some other PR/issue, I'll usually be able to follow that approach, as there are no worries that it will be forgotten.

> Would that even be possible?

Yes possible but as you mentioned we would need to revert the decision with meta data, as we would not be allowed to expose the KeySet.

A different mindset would be to break the 1:1 relation "class" <-> library, like it already happened in libelektra-core. And each class has it own name prefixes.

E.g.:
- core has key and keyset, methods are named as elektraKey and elektraKeyset
- operations has keyset eXtensions elektraKsXCut (so then you finally get your X)
- ease has the "class" array (calculateSpecificationToken would be in a class related to cryptography)

Then

- everything would be consistent.
- people only need to remember which class is in which method (and do not need to remember every individual method)

If we manage that there are not more than a handful classes per library it should work out quite fine.

kodebach:
> And each class has it own name prefixes. [...] If we manage that there are not more than a handful classes per library it should work out quite fine.

I like that approach. I'd use the term "(API) group" instead of "class" to avoid confusion. Each such group should also be documented as a separate group in doxygen (`@defgroup`/`@ingroup`, see [here](https://www.doxygen.nl/manual/grouping.html)).

> so then you finally get your X

I'm completely open to better suggestions. I just really really dislike the super unnecessarily long "operations".

The X actually comes from [AndroidX](https://developer.android.com/jetpack/androidx) which is the (technical) name for all the development libraries that Google provides outside the base Android SDK that is part of OS.

> elektraKsXCut 

That should probably be either `elektraXKeysetCut` or `elektraKeysetXCut` to match the `elektraKeyset` functions from `libelektra-core`.

> operations has keyset eXtensions

Maybe we should also call it `libelektra-extensions` or `libelektra-extra`.

kodebach:
> Maybe we simply pick another word. E.g. for ksCut: ksOpCut ksDoCut ksRoutineCut ksActionCut

IMO only `ksOpCut` is okay here. "routine" and "action" both have to much specific meaning in various software engineering contexts. "do" could make some very awkward sounding names.

Also IMO the names should be: `elektraOpKeysetCut`, `elektraDoKeysetCut`, `elektraRoutineKeysetCut`, `elektraActionKeysetCut`
In other words, the prefix for the library (`elektraOp`, etc.) should come first, then a prefix for a specific group like `Keyset` and then the actual name of the function. Think of it in terms of namespaces/classes in C++: `elektra::op::Keyset::cut`

> Extra is short.

`elektraExtraKeysetCut` is not that short anymore, but I'd be okay with it, since "extra" is a pretty nice name and very clearly tells the user that this additional stuff.

---

So do we go with `libelektra-extra` instead of `libelektra-operations` with prefix `elektraExtra*`? If so, I'll update the decision and we can hopefully merge this

markus2330:
Yes all of them sound good, please update.

kodebach:
@markus2330 I've moved the library list into the `library_split.md` decision and remove the two extra decisions for the new libraries. I hope this is now satisfactory for you.

doc/decisions/library_split.md
```diff
@@ -35,10 +36,112 @@ included by parts that need it
...
+- `libelektra-kdb`:
+  The main API for interfacing with the KDB.
+
+  **Prefix:** `elektraKdb*`
+
+- `libelektra-lowlevel-c`:
+  Additional C APIs that are useful when working with `ElektraKey` and `ElektraKeyset`, but not are not minimal API.
+  Specifically targets C and not intended for use via bindings.
+
+  **Prefix:** `elektraCKey*`, `elektraCKeyset*`
```
markus2330, line 64:
```suggestion
  **Prefixes:** `elektraCKey*`, `elektraCKeyset*`
```

doc/decisions/library_split.md
```diff
@@ -22,9 +22,10 @@ Only libelektra-core is supposed to access private data but this contradicts the
 
 ## Decision
 
-Also allow `libelektra-operations` library to access private Key/KeySet.
-Put struct definitions of Key/KeySet in a separate header file, which gets
-included by parts that need it
+Also allow `libelektra-extra` (and maybe others) to access private Key/KeySet.
```
markus2330, line 25:
```suggestion
Also allow `libelektra-extra` (and maybe other explicitly documented libraries) to access private Key/KeySet.
```

doc/decisions/library_split.md
```diff
@@ -35,10 +36,112 @@ included by parts that need it
...
+  Intended for use by applications.
+  Partially intended for use with code-generation.
+  Specifically targets C and not intended for use via bindings.
+
+  **Prefix:** `elektraHlc*`
+  (Note: Even though there isn't any other `highlevel` implementation, `c` suffix to show this lib is only for C)
+
+- `libelektra-opts`:
+  The API for parsing command-line arguments according to Elektra's spec.
+  **Prefix:** `elektraOpts*`
```
markus2330, line 77:
```suggestion

  **Prefix:** `elektraOpts*`
```

doc/decisions/library_split.md
```diff
@@ -35,10 +36,112 @@ included by parts that need it
...
+
+- `libelektra-type`:
+  The API that defines Elektra's type system.
+
+  **Names:** `elektra<TYPE>ToString` and `elektraKeyTo<TYPE>`
+  (Note: extracted from `libelektra-ease`)
+
+- `libelektra-extra`:
+  Contains extra APIs for `ElektraKey` and `ElektraKeyset` beyond the minimal API of `libelektra-core`.
+  These APIs are things that could be considered part of the "classes" for `ElektraKey` and `ElektraKeyset`, but which we do not consider minimal.
```
markus2330, line 124:
```suggestion
  These APIs are things that could be considered part of the classes for `ElektraKey` and `ElektraKeyset`, but which we do not consider minimal.
```

doc/decisions/library_split.md
```diff
@@ -35,10 +36,112 @@ included by parts that need it
...
+  The APIs should not specifically target C (see `libelektra-lowlevel-c` for that) and should be usable via bindings (if appropriate for the other language).
+
+  **Prefix:** `elektraExtra*`
```
markus2330, line 127: Where are the array functions? Wouldn't they be called `elektraArray`?

markus2330 (approved):
Looks very good now! Can be merged if you agree with my suggestions. I like the idea with the private libs very much! :rocket:  Mermaid diagram with deps would be very nice to have.

doc/decisions/library_split.md
kodebach: I'd just use `elektraExtraArray*` or maybe things like `elektraExtraKeyIsArrayName` and `elektraExtraKeysetGetArray`. Depends really what kind of array functions we'll have.

kodebach:
Let's not merge this until after #4187 is merged. The changes in `doc/decisions/README.md` will cause another conflict and it's easier to rebase this PR after #4187 than to deal with it in the huge #4187 again (especially since there is already a conflict here)

Issue #817, opened 2016-06-30
include <elektra/kdb.h>


markus2330:
We could install all header files also to `/usr/include/elektra`, which allows us to `include <elektra/kdb.h>` without any include path.

markus2330:
@domhof Is this a 1.0 issue?

sanssecours:
If the alternative “Macro magic to support both inclusion options“ means that we do not need to install each header file twice, then I am for that option.

How about creating a link `/usr/include/elektra/legacy` that points to `/usr/include/elektra`. Would that also work?

markus2330:
> means that we do not need to install each header file twice, then I am for that option.

What is so bad about installing the files twice? They are not really large... (388K)

> How about creating a link /usr/include/elektra/legacy that points to /usr/include/elektra. Would that also work?

You are right, actually we do not need to install them twice. We simply internally always use `<elektra/file.h>` and if the user sets the include path to `/usr/include/elektra` the internal `elektra/kdb.h` should still work?

Within the source repo, however, we would need to create a folder called `elektra`, so the include files would be in `src/include/elektra.

sanssecours:
> What is so bad about installing the files twice? They are not really large... (388K)

It’s not that bad, but I am usually a fan of keeping things small and tidy.

> Within the source repo, however, we would need to create a folder called elektra, so the include files would be in `src/include/elektra`.

Provided that this solution works, creating an additional subfolder does not sound that bad. I certainly could live with it.

markus2330:
lets do it

kodebach:
I think a big issue here is also that we often use e.g. `#include <kdbmacros.h>` with `<>`. AFAIK the C standard doesn't have an opinion on this, but most compilers will start only look in include paths with `<>`. But with `#include "kdbmacros.h"` they first look in the directory the current file is in, so that way it should work no matter where we put our headers, as long as the have the same relative layout as in the source tree.

markus2330:
As discussed in #3813 we do not need any legacy header files. The only supported way to include files should be `<elektra/NAME.h>`, where NAME usually does not start with "kdb", so e.g. it would be `<elektra/macros.h>`.

kodebach:
This issue is a bit outdated. We now have a few decisions that describe what the everything should look like:

[...]

PR #4246, opened 2022-02-15
[decisions] headers

kodebach:


doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,95 @@ The big problem is kdbprivate.h. It has two main problems:
...
+  1.  Some parts are truly private, i.e. shouldn't be used outside the library that defines them.
+      These things are only there because a library was split into multiple `.c` files.
+      This includes e.g. `splitNew` and `struct _Split`.
+
+      Symbols belonging to this category should not appear at all in the `symbols.map` file.
+
+  2.  Other things are truly private, but must be tested.
+      This includes e.g. most other `split*` functions or the `elektraKeyName*` functions.
+
+      Symbols belonging to this category should not appear in a public section of the `symbols.map` file.
```
markus2330, line 37: Do we need these symbols at all? In theory unit tests could simply include the .c files (with static functions).

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,95 @@ The big problem is kdbprivate.h. It has two main problems:
...
+  Without this constraint, we would constantly include things that aren't needed.
+  This is inconvenient as it clutters the auto-suggestions of IDEs.
+  It also slows down the compile process, since the compiler has to parse everything that was included (see [also](https://lore.kernel.org/lkml/YdIfz+LMewetSaEB@gmail.com/T/)).
+
+- Non-public API and public API should be kept in separate files
+- Some non-public APIs need to be accessible for testing, such headers should not be packaged/installed
+
 ## Assumptions
 
+- There are different categories of "private":
```
markus2330, line 26: This is a very important point, actually it is a decision on its own.

doc/decisions/header_include.md
```diff
@@ -0,0 +1,131 @@
...
+
+- Put all headers into one directory and use the same relative layout as will be installed. Then use `""`:
+
+  This makes for an inconvenient development experience. It is also pretty hard to achieve for plugins and private headers would clutter the include-directory.
+
+## Decision
+
+The rules for including headers are:
+
+- Including another header from the current directory is done with:
```
markus2330, line 81: I wouldn't introduce such a rule. It is very hard to enforce it. And what is the benefit?

doc/decisions/header_include.md
```diff
@@ -0,0 +1,131 @@
...
+  ```c
+  #include "header.h"
+  ```
+- Including another header from a subdirectory of the current one (e.g. from a third-party library whose code was copied into the repo) is done with:
+  ```c
+  #include "subdir/header.h"
+  ```
+- Including a header from another library within the Elektra codebase, is done with:
+  ```c
+  #include <elektra/otherlib/header.h>
```
markus2330, line 91: :+1: see also #817

markus2330 (commented):
Input before meeting.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,95 @@ The big problem is kdbprivate.h. It has two main problems:
 
 ## Constraints
 
+- Each library should have at least one separate header file:
+
+  Without this constraint, we would constantly include things that aren't needed.
+  This is inconvenient as it clutters the auto-suggestions of IDEs.
+  It also slows down the compile process, since the compiler has to parse everything that was included (see [also](https://lore.kernel.org/lkml/YdIfz+LMewetSaEB@gmail.com/T/)).
+
+- Non-public API and public API should be kept in separate files
```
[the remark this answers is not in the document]
kodebach: Non-public in this case includes things like the definitions of `struct _Key` et al. So in some cases it makes sense to install the non-public headers. In other cases (like the `fortests`) we probably don't need to install them.

doc/decisions/header_file_structure.md
kodebach: IMO including .c files is a bad idea.

1. It means that there will be a separate definition of the function in every file that includes the function. The performance aspect wouldn't matter for tests, but it would definitely break things like `static` variables (if we have those anywhere).
2. I think it's harder to define rules for when including `.c` files is okay and which files can be included where, than to just say "if you need to test a private function, make it non-static, add it to a `fortests` header and add it to the `fortests` section of the `symbols.map`".

doc/decisions/header_include.md
kodebach: > It is very hard to enforce it.

Code review is always an option, but it also shouldn't be too hard to write a script that checks this, or add a custom query to our lgtm.com build job.

> And what is the benefit?

1. It clearly shows whether or not a header is part of the same library or not.
2. You need to type less than. The alternative would be always having `<elektra/mylib/header.h>`.
3. Generally I'd say in C code the convention is `<>` is for external/system headers and `""` is for local headers. So when I see a `<>` I expect it to come from somewhere else.
4. When you use `""` your editor/IDE doesn't need to know about the correct include path. Since the header will be in the same directory the header can be found without extra setup. With `<>` an editor/IDE that isn't setup explicitly, will just look in the system folders. So it might find a version of the header, but it could be the wrong one (installed vs. local source).

Ideally I'd like to use `""` for including all Elektra headers from within the code base. Then there can't be any confusion. But this would require careful coordination between the source tree and the install locations and also the use of `../`, so I made this compromise.

doc/decisions/header_include.md
kodebach: Headers are already installed into `/usr/include/elektra`, but we don't properly use the `<>`/`""` right now. Which is why we need a custom include path.

doc/decisions/header_file_structure.md
kodebach: What exactly would the decision part of the decision be? I'm just stating facts here. The assumption is that all symbols can be put into one of these four groups. But I'm not sure there is anything to decide here. The decision to use these four groups to build our headers is what **this** decision is about. Not sure, what you want to separate here.

kodebach:
@markus2330 I added the results from the meeting today, please check if this can be merged already

doc/decisions/header_include.md
```diff
@@ -0,0 +1,123 @@
...
+>
+>    (that does not match one of the two previous forms) is permitted. The preprocessing tokens after include in the directive are processed just as in normal text. (Each identifier currently defined as a macro name is replaced by its replacement list of preprocessing tokens.) The directive resulting after all replacements shall match one of the two previous forms. The method by which a sequence of preprocessing tokens between a `<` and a `>` preprocessing token pair or a pair of `"` characters is combined into a single header name preprocessing token is implementation-defined.
+>
+> ### Definitions:
+>
+> - h-char: any member of the source character set except the new-line character and `>`
+> - q-char: any member of the source character set except the new-line character and `"`
+
+In short, the whole search process of the `#include` mechansim is implementation-defined.
+However, because `#include` works (mostly) like a literal copy-paste, deciding between `""` and `<>` within a public header file can affect user code.
```
markus2330, line 55: The problem should always describe a problem with Elektra. And it should be important enough, I don't think consistent use of `""` and `<>` is important enough for a decision or a rule in CODING.md.

doc/decisions/header_include.md
```diff
@@ -0,0 +1,123 @@
...
+  #include "header.h"
+  // or
+  #include "subdir/header.h"
+  ```
+  It should not be necessary, to include a private non-installed header from another directory (i.e. `../` shouldn't be needed).
+- Installed headers are included with their full path as if they are installed already:
+  ```c
+  #include <elektra/somelib.h>
+  // or
+  #include <elektra/somelib/header.h>
```
markus2330, line 88: :+1:

markus2330 (commented):
Please keep decisions short and to the point. Also make sure they are about important decisions within the architecture and framework of Elektra. I don't think that `""` vs. `<>` includes can fulfill these points.

doc/decisions/header_include.md
markus2330: I don't understand what you try to say. Files installed to `/usr/include` can be included via `<>` without any include path.

kodebach: I just wanted to clarify the relation to #817. The problem is not where the headers are installed, it is how we include them.

kodebach:
As an example of what I was aiming for with this decision, take a look at [tree-sitter](https://github.com/tree-sitter/tree-sitter/tree/master/lib). They use a top-level `include` folder for the public headers, which has the same structure as will be installed. All the non-public headers live directly in the `src` folder next to the source files. For `#include` the use `<>` with system headers, `"tree_sitter/..."` for public headers and `"./..."` for private headers.

Now obviously tree-sitter is only a single library unlike Elektra, but the ideas could be adapted.

1. Public headers live in `include/elektra/[libname]`
2. Private headers live next to the source files in e.g. `src/libs/core-c/` or `src/libs/kdb`.

For dependencies between public headers `#include <elektra/[libname]/header.h>` is always used, for other files following rules apply:

1. Including system/external headers uses `#include <header.h>`
2. Including private headers from the current library uses `#include "./header.h"` 
3. Including public headers from the current or another library uses `#include <elektra/[libname]/header.h>`
4. Including private from another library is not allowed.

AFAICT this approach really simplifies the include path config too. The include path would just point to the top-level `include` folder and that's it.

markus2330:
I am not convinced that there is really a problem that needs to be solved: I never had a problem that a wrong header file gets included. We definitely have a problem with installed header files (there are several kdb.h floating around and it would be great if `#include <elektra/kdb.h>` just works for applications without any include path hurdles. So I would concentrate on this problem, and if this is fixed we can also do internal changes. What do you think?

kodebach:
Changing the way the headers are stored in the repo would make fixing the include path much simpler. The goal would be that the headers are stored in the same layout as when installed. That means we just need to add a single directory to the include path (in our CMake) to mimic the installed behaviour.

For the `#include`s: If we go with the solution proposed above, we'd essentially have the rule "always use `<>` unless the target is a private and in the same folder". The advantage here would be that only headers that are public and get installed will be in added to the include path. So accidentally including something that isn't installed (I think we had that happen once a few years ago), isn't possible, because it won't compile.

markus2330:
> The goal would be that the headers are stored in the same layout as when installed.

I agree that this is useful.

kodebach:
@markus2330 Can we agree on this approach?

> 1. Public headers live in `include/elektra/[libname]`
> 2. Private headers live next to the source files in e.g. `src/libs/core-c/` or `src/libs/kdb`.
> 
> For dependencies between public headers `#include <elektra/[libname]/header.h>` is always used, for other files following rules apply:
> 
> 1. Including system/external headers uses `#include <header.h>`
> 2. Including private headers from the current library uses `#include "./header.h"`
> 3. Including public headers from the current or another library uses `#include <elektra/[libname]/header.h>`
> 4. Including private from another library is not allowed.

AFAICT the only remaining issue in this scheme was `#include "./header.h"`. A bit of further explanation for this:
- I would enforce the `./` prefix to make abundantly clear that this is header only present in the local directory.
- Since it may only be used for private headers (which will not be installed), these `#include`s can only appear in source (`.c`) files or in headers that are themselves private.
- Together with the other rules, this scheme should allow a very simple configuration of the include path.
   - When Elektra is installed in a default location like `/usr` no configuration is needed. The headers will be in e.g. `/usr/include/elektra/foo/foo.h` and `/usr/include` will on the default include path. So `#include <elektra/foo/foo.h>` will just work.
   - When Elektra is installed in a non-default location like `/home/user/stuff`, just one directory has to be added to the include path (e.g. `/home/user/stuff/include`).
   - When building Elektra the include path is pointed at `src/include`. This directory will mirror the installed layout (e.g. `src/include/elektra/foo/foo.h`). Because we use `#include "./header.h"`, we don't need to add extra include paths for private non-installed headers that live outside of `src/include`.


If we agree on this scheme, I will rewrite the decision documents so this PR can finally be merged.

markus2330:
>    I would enforce the ./ prefix to make abundantly clear that this is header only present in the local directory.

My only objection is: how can we check and enforce such a rule? If we don't find any way, it is probably a dead and confusing rule. Such rules should not exist, as then also important rules (e.g. which header files to install where, thread-safety etc.) might not be read or not taken serious.

If you your automatic reformatter changes `"./"` and `<>` accordingly, then: why not, I don't have opinions about it. I would allow anything that works. To be more clear what I find worthwhile to actually check in reviews and tests (which is a lot of effort and can only be done for user-facing decisions):

>    Since it may only be used for private headers (which will not be installed), these #includes can only appear in source (.c) files or in headers that are themselves private.
>    Together with the other rules, this scheme should allow a very simple configuration of the include path.
>        When Elektra is installed in a default location like /usr no configuration is needed. The headers will be in e.g. /usr/include/elektra/foo/foo.h and /usr/include will on the default include path. So #include <elektra/foo/foo.h> will just work.

Here I agree 100%. This we should have and is the main goal of the decision. Implication: We must test inclusion of all header files without include path.

>        When Elektra is installed in a non-default location like /home/user/stuff, just one directory has to be added to the include path (e.g. /home/user/stuff/include).

Yes, so we need to keep the package-config files. This implication should be also mentioned in the decision.

kodebach:
> how can we check and enforce such a rule?

Writing a script (maybe using clang-query) to run as a test case and in CI should be easy enough. As a first step, even this could work (although the false positive rate may be high)

```sh
# find #include "..." lines, within those exclude #include "./ lines
grep -E '^\s*#include\s+".+"' | grep -vE '^\s*#include\s+"\.\/'
```

> If you your automatic reformatter changes "./" and <> accordingly

I could probably make it do that, but a simple regex replace might be enough here.

markus2330:
> Writing a script (maybe using clang-query) to run as a test case and in CI should be easy enough.

If you write such a script, I do not have objections against the new rules of header files.

kodebach:
Sadly using `clang-query` doesn't work, since the Clang AST doesn't contain any nodes for preprocessor directives. But I checked against a `tree-sitter` query (tree-sitter's grammar has nodes for preprocessor directives), and found that the simple `grep` command from above is actually accurate.

So we can probably just use something like

```sh
find -type f -name "*.c" -print0 | xargs -0 grep -E '^\s*#include\s+".+"' | grep -vE '.*:\s*#include\s+"\.\/"'
```

At least until find a false positive. At that point we could add a whitelisting mechanism (e.g. adding a `// ignore include` comment at the end of the false positive line).

kodebach:
@markus2330 The decisions are now updated, please review again.

doc/decisions/header_file_structure.md
markus2330: It is probably more standard to use the word "protected" for something that is "public" within Elektra's repository but private for the outside world (point 3).

> The assumption is that all symbols can be put into one of these four groups.

Then write what you mean:

```suggestion
- We assume that all parts of the API belong to one of the following visibility categories:
```

> Not sure, what you want to separate here.

The separate decision would be about which parts belong to which "visibility category".  You give good examples here.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+Monolithic libraries should be small and bigger libraries with large APIs should be modularized.
+
+A monolithic library `foo` may have these headers (covering categories 3 & 4 from above):
+
+- `src/include/elektra/foo.h`:
+  Contains the full public API of `libelektra-foo`.
+  Will be installed as `<include-root>/elektra/foo.h`.
+- `src/include/elektra/foo/internal.h`:
+  Contains the internal (i.e., non-stable) API of `libelektra-foo`.
+  Will be installed as `<include-root>/elektra/foo/internal.h`.
```
markus2330, line 70: I don't think we should install this file.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+
+A modularized library `bar` may have these headers (covering categories 3 & 4 from above):
```
markus2330, line 72:
```suggestion
A modularized library `foo` may have these headers (covering categories 3 & 4 from above):
```

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+
+- `src/include/elektra/foo/public.h`:
+  Declares the public API of `libelektra-foo`, by including `#include <elektra/foo/*.h>`.
+  Will be installed as `<include-root>/elektra/public/foo.h`.
```
markus2330, line 76: This sounds weird to me. I wouldn't restructure the files during installation.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+
+- `src/include/elektra/foo/*.h`:
+  Additional public API header of `libelektra-foo`.
+  Will be installed as `<include-root>/elektra/foo/*.h`.
+
+  All of these installed headers `installed.h` must be included from `src/lib/foo/public.h` via a line `#include <elektra/foo/installed.h>`.
+
+- `src/include/elektra/foo/internal.h`:
+  Declares the internal (i.e., non-stable) API of `libelektra-foo`, by including `#include <elektra/foo/*.h>`.
+  Will be installed as `<include-root>/elektra/internal/foo.h`.
```
markus2330, line 88: Again, it is okay to have internal header files (to be used within other modules of Elektra) but they should not be installed.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+Additionally, all libraries may also have private headers:
+
+- `src/lib/foo/**/*.h`:
+  Additional headers may be present.
+  These headers may only be used by other files within `src/lib/foo`, according to the rules in [Including Headers](header_include.md).
+
+### Plugins
+
+Plugins do not declare their API via header files.
+Their headers are never installed and can be named any way the developer wants.
```
markus2330, line 107: :+1:

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+
+### Tests
```
markus2330, line 109: This info also needs to be in doc/TESTING.md

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+If a symbol is needed in only one file and for tests, option 1 should be preferred.
+For symbols that are used in multiple files, a header needs to exist anyway.
+In any case, these unit tests should not be installed and should statically link the library into the test executable.
+This way we don't pollute our `symbols.map` files and keep the number of exported symbols down.
 
 ## Rationale
 
+- This structure makes the `#include`s simple and works nicely with our directory structure.
+- Having a uniform naming convention simplifies things, both for developers writing Elektra and those using Elektra.
+- Requiring libraries must either be fully modularized (main headers are only `#includes`) or completely monolithic, avoids the situation where a library has some parts in `public.h` and other parts in extra headers.
```
markus2330, line 125: :+1:

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,128 @@ The big problem is kdbprivate.h. It has two main problems:
...
+- `src/include/elektra/foo.h`:
+  Contains the full public API of `libelektra-foo`.
+  Will be installed as `<include-root>/elektra/foo.h`.
+- `src/include/elektra/foo/internal.h`:
+  Contains the internal (i.e., non-stable) API of `libelektra-foo`.
+  Will be installed as `<include-root>/elektra/foo/internal.h`.
+
+A modularized library `bar` may have these headers (covering categories 3 & 4 from above):
+
+- `src/include/elektra/foo/public.h`:
```
markus2330, line 74: I don't like the word public here. What about  "foo/foo.h" to include everything?

doc/decisions/header_include.md
markus2330: Not really an update here?

markus2330 (commented):
Further discussion.

markus2330:
> At least until find a false positive. At that point we could add a whitelisting mechanism (e.g. adding a // ignore include comment at the end of the false positive line).

We should avoid further whitelisting mechanisms. Wouldn't it be possible to write it differently to not trigger a false positive?

doc/decisions/header_file_structure.md
kodebach: That's actually a typo. It was supposed to be

```suggestion
  Will be installed as `<include-root>/elektra/foo/public.h`.
```

doc/decisions/header_file_structure.md
kodebach: I'm not a massive fan of the repetition, but I guess it's better than `public`.

markus2330: I also considered "all" and "everything".

kodebach:
@markus2330 Let's move the discussion about "unstable" API from #4245 to this PR, since it also affects the header structure.

I have thought about it some more and I came to the conclusion that we shouldn't use the term "unstable" anymore. It is just causing confusion. Actually we should split what I used the term "unstable" for into two categories:

1. "internal": These APIs are technically public (i.e. exported symbols), but not meant for consumption out side of `libelektra-*` libraries. They exist for communication between libraries.
2. "experimental": These APIs are public and meant for consumption by external users, but they are not fully stable yet. None of the semantic versioning guarantees apply. There may be breaking changes even in a minor version change.

With this distinction I agree, that the `internal` headers should not be installed. However, I'd actually change the setup of `src/include` a bit to make this more obvious.

- `src/include/internal/foo.h` or `src/include/internal/foo/header.h` is for internal headers.
  These are not installed. Inside the repo they can in theory be included everywhere as e.g. `#include <internal/foo.h>`, since the include path points to `src/include`. To make sure we don't use these in an installed header, I would extend the script that checks `#include "./..."` to also ensure that `#include <internal/...` doesn't appear in `src/include/elektra`.
   Another advantage of moving internal headers out of `src/include/elektra` is during installation we simply copy _all of_ `src/include/elektra`.
- `src/include/elektra/experimental/foo.h` or `src/include/elektra/experimental/foo/header.h` is for experimental headers.
  These are installed (since they are in `src/include/elektra`). Theoretically they can also be included anywhere as e.g. `#include <elektra/experimental/foo.h>`. Here the script would also need to be extended to make sure there is no `#include <elektra/experimental/...` in `src/include/elektra` (except for `src/include/elektra/experimental`).

> **Note**: The third category of APIs is "stable" APIs in `src/include/elektra` (outside of `src/include/elektra/experimental/`). Therefore I use non-stable to refer to "internal" and "experimental" headers together.

_A bit of explanation why I think this is better than excluding whole libraries from semver guarantees:_
1. When a function moves from being experimental to stable (without API change), it remains in the same shared object. That means application don't need to be recompiled/relinked. Therefore, such a change would not be a breaking change and could be done in a minor version update. External source code probably needs to be updated for the new header (unless it is already included), but I think that's acceptable.
2. One advantage of having stable and non-stable libraries is that you can do e.g. `ldd /usr/bin/kdb` to check if an application uses non-stable libraries. But if we really consider that a feature, a stable library could never depend on a non-stable one, even if the stable library would abstract over all the non-stable APIs.
3. Even if all libraries are stable, you can't actually be sure that an application isn't using something that falls outside our semver guarantees. All the "internal" APIs are technically public and accessible to any app/library that knows the functions signatures (e.g. by copying the header from our repo). So the "use `ldd /usr/bin/kdb` to check for non-stable libs" idea isn't actually reliable.

markus2330:
I think it is better time-management-wise if we continue the discussion about experimental libraries when we know which of our libraries will stay experimental even within 1.0. Maybe we can get all functions in the important libraries non-experimental, and I think everyone agrees that this would be the best.

kodebach:
> Maybe we can get all functions in the important libraries non-experimental, and I think everyone agrees that this would be the best.

Yes, but I'd like to finish this PR at some point... So if we can at least agree on the part about "internal" headers that would be nice.

Then I'd update the decisions and just mention that the decision must be updated before we can have "experimental" APIs, i.e. APIs that are not stable but should still be public. In other words, anything that isn't moved to an "internal" header would be public stable API as soon as 1.0 releases.

markus2330:
> So if we can at least agree on the part about "internal" headers that would be nice.

What is missing? The PR doesn't seem updated.

> In other words, anything that isn't moved to an "internal" header would be public stable API as soon as 1.0 releases.

:+1:

kodebach:
> What is missing? The PR doesn't seem updated.

I wanted to know, if you agree with the ideas about "internal" headers from this comment https://github.com/ElektraInitiative/libelektra/pull/4246#issuecomment-1249185666. If so, I will update the decision files in the PR.

markus2330:
I already said I am not a fan of individual symbols that are excluded from stability guarantees. I agree it might be necessary if we are unable to refactor everything nicely (e.g. in src/include/kdbprivate.h are many public symbols just for tests) but I don't see why we should write a complicated rule-framework for something which is actually not something we want.

It is important that decisions focus on something. It is political and not technical if decisions about headers try to punch holes through totally unrelated library conventions.

Furthermore, urgent now is merging new-backend. I don't see a correlation of these decisions and the work on new-backend.

kodebach:
> Furthermore, urgent now is merging new-backend. I don't see a correlation of these decisions and the work on new-backend.

Yes, I know you want to merge `new-backend`. But I don't think it's ready and I don't think I can do much to make it ready. At least not without trampling over the work others are trying to do.

The reason why I asked, is because I thought "Can we agree on ...?" is a simple yes/no question. And in fact I thought we already agreed and just wanted clarification. Clearly that's not the case. I asked, because it would have been very quick for me to update the PR and the could be merged and we'd have one less thing to worry about...

To get back to the issue at hand:

> I already said I am not a fan of individual symbols that are excluded from stability guarantees. I agree it might be necessary if we are unable to refactor everything nicely (e.g. in src/include/kdbprivate.h are many public symbols just for tests) but I don't see why we should write a complicated rule-framework for something which is actually not something we want.

I have no idea what you are talking about.... Please don't even try to explain, it will not help the discussion.

Please just answer this very basic yes/no question: Do you agree that we need (at least) public headers which will be included in packages like `libelektra-dev`, and also internal headers, which exist and can be used in the repo, but will not be included in packages?

That is all I wanted to know, because the decision in this PR currently does not reflect this correctly. In the current file, a header is either included in packages or it is restricted to a single library. Ignore all the ideas about experimental/unstable headers that will be installed, I'll remove all that and replace it with something like "if we need an installed, but not stable API for 1.0, we need to make a new decision".

> It is important that decisions focus on something. It is political and not technical if decisions about headers try to punch holes through totally unrelated library conventions.

Again no idea where this came from... This is a very technical decision. It is about where in the repo headers live, what headers are included in packages, what headers can be included in which files, etc. In a more general sense, this is about API guarantees which might be a bit more philosophical but is still very important, because it has real impact on users of Elektra.

markus2330:
Yes, there must be header files that are internal to our repo, including but not limited to our struct definitions. It was wrong to install them, as this brings expectations that cannot and should not be fulfilled.

> This is a very technical decision.

Perfect, this is what I want it to be, too! :+1: 

> might be a bit more philosophical

Lets try to remove the philosophical stuff. :carpentry_saw:

kodebach:
Okay, so we do agree after all. I'll update the files in this PR, when I have some spare time.

kodebach:
@markus2330 I updated the PR

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,135 @@ The big problem is kdbprivate.h. It has two main problems:
...
+  Declares the public API of `libelektra-foo`, by including `#include <elektra/foo/*.h>`.
+  Will be installed as `<include-root>/elektra/foo/public.h`.
+
+  Such a header may only contain `#include <elektra/foo/*.h>` lines.
+
+- `src/include/elektra/foo/*.h`:
+  Additional public API header of `libelektra-foo`.
+  Will be installed as `<include-root>/elektra/foo/*.h`.
+
+  All of these installed headers `installed.h` must be included from `src/include/elektra/foo.h` via a line `#include <elektra/foo/installed.h>`.
```
markus2330, line 84: Sentence is broken

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,135 @@ The big problem is kdbprivate.h. It has two main problems:
...
+  Declares the internal API of `libelektra-foo`, by including `#include <internal/foo/*.h>`.
+  Will not be installed.
+
+  Such a header may only contain `#include <internal/foo/*.h>` lines.
+
+- `src/include/internal/foo/*.h`:
+  Additional internal header of `libelektra-foo`.
+  Will not be installed.
+
+  All of these installed headers `installed.h` must be included from `src/include/internal/foo.h` via a line `#include <internal/foo/installed.h>`.
```
markus2330, line 96: Also here.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,135 @@ The big problem is kdbprivate.h. It has two main problems:
...
+- This structure makes the `#include`s simple and works nicely with our directory structure.
+- Having a uniform naming convention simplifies things, both for developers writing Elektra and those using Elektra.
+- Requiring libraries must either be fully modularized (main headers are only `#includes`) or completely monolithic, avoids the situation where a library has some parts in `foo.h` and other parts in extra headers.
+  This is bad, because it encourages users to simply include `foo.h`, which defeats the point of modularized headers.
+
 ## Implications
 
+- The location of a header file within the source tree determines what API it contains:
+  - Public: Will be installed, must be stable, can be used anywhere
+  - Internal: Will not be installed, doesn't have to be stable, can be used anywhere, but public headers
```
markus2330, line 132:
```suggestion
  - Internal: Will not be installed, doesn't have to be stable, can be used anywhere but public headers
```

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,135 @@ The big problem is kdbprivate.h. It has two main problems:
...
+  - Private: Will not be installed, doesn't have to be stable, can only be used in the same library
```
markus2330, line 133: This private is different to the 4 private above?

doc/decisions/header_include.md
```diff
@@ -0,0 +1,136 @@
...
+  // etc.
+  ```
+
+Only `build/include`, which is a copy of `src/include` after CMake processing, is added to the include path.
+The entire `build/include/elektra` directory is installed as-is, with the exact same directory structure and without any further processing.
+
+This is enough for `#include <>`s to work.
+
+We will also enforce that the path in a `#include ""` always starts with a `./` and does not contain any `/../`.
+To do this, we will use a simple `grep` based script that runs as a test case and as an early part of the CI (like e.g., the formatting check).
```
markus2330, line 113: This is implication.

doc/decisions/header_include.md
```diff
@@ -0,0 +1,136 @@
...
+The decision highlights the difference between installed and non-installed headers.
+The main driving factor for using `""` at all was that including a non-installed private header with `<>` would be unexpected, since non-installed headers shouldn't be in the (standard) include-path.
+This distinction also makes it easier to ensure that non-installed headers are not accidentally included in installed ones.
+
+See also considered alternatives.
+
+## Implications
+
+- All installed headers (and only those) must be put into `src/include/elektra`.
+- There must not be any `#include <internal/...>`s anywhere within `src/include/elektra`.
```
markus2330, line 129: Also checked by grep script.

doc/decisions/header_include.md
```diff
@@ -0,0 +1,136 @@
...
+However, because `#include` works (mostly) like a literal copy-paste, deciding between `""` and `<>` within a public header file can affect user code.
+
+## Assumptions
+
+- Modern Compilers work like [GCC](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html):
+  - `#include "[header]"` treats `[header]` as file path relative to the current file (as defined by the standard, if such a file doesn't exist there is a fallback to the `<>` behavior)
+  - `#include <[header]>` treats `[header]` as file path relative to one of the pre-defined include-paths
+
+## Considered Alternatives
+
```
markus2330, line 64:
```suggestion

- Only use `#include <>` and add everything to include path.
```

markus2330 (commented):
Looks quite good. A few sentences need improvements, no replies needed.

I am not a fan of doc/decisions/header_include.md but I agree it would be an improvement to have consistency checks in the public headers.

doc/decisions/header_file_structure.md
kodebach: I don't see the problem...

Would this make it better:

```suggestion
  All of these "installed headers" (`installed.h`) must be included from `src/include/elektra/foo.h` via a line `#include <elektra/foo/installed.h>`.
```

doc/decisions/header_file_structure.md
kodebach: It's category 1. Truly private shouldn't be used outside library.

doc/decisions/header_file_structure.md
markus2330: It is a bit better but still: "installed headers" is plural, `installed.h` is one header. Maybe simply a "for example" is missing to fix the sentence but I am not sure what you want to say here at the moment.

kodebach: How about 

```suggestion
  Any one `installed.h` of these installed headers must be included from `src/include/elektra/foo.h` via a line `#include <elektra/foo/installed.h>`.
```

markus2330: Yes, better.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,135 @@ The big problem is kdbprivate.h. It has two main problems:
 
 ## Constraints
 
+- Each library should have at least one separate header file:
+
+  Without this constraint, we would constantly include things that aren't needed.
+  This is inconvenient as it clutters the auto-suggestions of IDEs.
+  It also slows down the compile process, since the compiler has to parse everything that was included (see [also](https://lore.kernel.org/lkml/YdIfz+LMewetSaEB@gmail.com/T/)).
```
markus2330, line 19: The link is misleading here. This was a huge undertaking and has little to do what is planned with this decision. Please move it down to Notes.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,135 @@ The big problem is kdbprivate.h. It has two main problems:
...
+- Non-public API and public API should be kept in separate files
+- Some non-public APIs need to be accessible for testing, such headers should not be packaged/installed
+
 ## Assumptions
 
+- There are different categories of "private":
+
+  1.  Some parts are truly private, i.e., shouldn't be used outside the library that defines them.
+      These things are only there because a library was split into multiple `.c` files.
+      This includes e.g., `splitNew` and `struct _Split`.
```
markus2330, line 30: Do they still exist?

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,135 @@ The big problem is kdbprivate.h. It has two main problems:
...
-    Here I would either create a separate kdbinternal.h or just keep kdbprivate.h, as long as it is a separate file from 4.
-    Symbols belonging to this category should only appear in the private section of the symbols.map file.
-4.  Finally, there is what I'd call unstable or pseudo-public. This includes e.g. ksFindHierarchy or elektraReadArrayNumber. These functions could be public, but for various reasons are not. Maybe they are not well-tested or maybe we just don't want to commit to the function yet. Many of these (but not all) can't be part of libelektra-ease, because the rely on something more private (e.g. directly access the array in a KeySet).
-    Here I would either create a separate kdbinternal.h or just keep kdbprivate.h, as long as it is a separate file from 3.
-    Symbols belonging to this category should only appear in the private section of the symbols.map file.
+### Libraries
```
markus2330, line 58: The title as-is is confusing.

```suggestion
### Header Files for Different Types of Libraries
```

markus2330 (approved):
1. I agree with header_file_structure. In the review I suggested only textual changes. Please furthermore make more clear which of the 4 types of header files get installed where. (also which of the 4 categories "internal" is could be more clear.)
2. Header include I accept with reservations: I have doubts if people will follow the rules and if it is worth the effort. Please write in the decision (or in assumption) that a tool will automatically do it and rules do not need to be known by developers (or similar).

Who wants to be second reviewer?

doc/decisions/header_file_structure.md
kodebach: No, but `backendsMerge` and `struct _BackendData` do, I'll change that.

doc/decisions/header_file_structure.md
kodebach: I find your title a bit verbose and it would also no longer match the next sections "Plugins" and "Tools".

How about, I add an explanation above e.g.

```suggestion
The decision is split into 4 subsections "Libraries", "Plugins", "Tools" and "Tests", because the different parts of Elektra have different requirements.

### Libraries
```

Note: I'll also add the section about tools. It will basically say, something similar to the subsection about plugins. Tools don't have a public API (at least not a C API), so no restrictions apply.

markus2330: Yes, makes it much clearer! :+1:

kodebach:
> Who wants to be second reviewer?

No idea who would be best, but you probably need to tag someone or we need to bring it up in the next meeting. Otherwise, probably nobody will respond.

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,154 @@ The big problem is kdbprivate.h. It has two main problems:
...
 ## Constraints
 
+- Each library should have at least one separate header file:
+
+  Without this constraint, we would constantly include things that aren't needed.
+  This is inconvenient as it clutters the auto-suggestions of IDEs.
+  It also slows down the compile process, since the compiler has to parse everything that was included.
+
+- Non-public API and public API should be kept in separate files
+- Some non-public APIs need to be accessible for testing, such headers should not be packaged/installed
```
atmaxinger, line 22: What a coincidence! I just stumbled upon this while creating the high-level change-tracking API. Currently I just added a function for creating a fake/mock ChangeTrackingContext into the public API, but I thought maybe we should have seperate `-testing` headers for such functionality? This then would have to be installed, as 3rdpartys may write plugins that use it.

kodebach: > This then would have to be installed, as 3rdpartys may write plugins that use it.

Well, if the API is intended for consumption by third parties, then it should be installed anyway and not a separate header for testing. If people need more, they should just copy the headers from the repo into their code.

doc/decisions/header_file_structure.md
atmaxinger: Hmm honestly I don't think we should do a seperate decision here. The examples that @kodebach listed serve illustrative purposes in this decision. 

Creating decisions just for documenting where which each function of each library should be placed is too much IMHO. That sounds very much like documentation for documentation's sake. Checking the placement is something that should be part of the PR review.

atmaxinger: If we apply the suggestion by @markus2330 then we must also define 'public' and 'internal' visibilities in this section.

atmaxinger:
Looks good to me 👍

doc/decisions/header_file_structure.md
```diff
@@ -12,31 +12,154 @@ The big problem is kdbprivate.h. It has two main problems:
...
+      Symbols belonging to this category should not appear at all in the `symbols.map` file.
+      Headers declaring such symbols must never be installed.
+
+  2.  Other things are truly private, but must be tested.
+      This includes e.g., most `backends*` functions or the `elektraKeyName*` functions.
+
+      Symbols belonging to this category should not appear in at all in the `symbols.map` file.
+      Headers declaring such symbols must never be installed.
+
+  3.  Some things are not part of the public API and will never be part of the public API.
```
atmaxinger, line 41: So this is visibility 'internal'?

kodebach:
@markus2330 Can this be merged now? If so, I will resolve the merge conflict.

markus2330:
I already approved before: Please write my reservation in the notes of [header_include.md](https://github.com/ElektraInitiative/libelektra/pull/4246/files#diff-d73b4e18afecf3f9b6fe38faf66f2151e657c678e43fa37acf126723d8daadf5) and rebase. :rocket:

kodebach:
> I already approved before

I made more changes after the review, that's why I re-requested the review.

> Please write my reservation in the notes of [header_include.md](https://github.com/ElektraInitiative/libelektra/pull/4246/files#diff-d73b4e18afecf3f9b6fe38faf66f2151e657c678e43fa37acf126723d8daadf5)

The decision already states in multiple places that a tool will check the new rules. If you want more, please make a concrete suggestion.

> rebase

I'll do that when I know the merge is imminent, because the conflict will just reappear when another decision PR is merged.

markus2330 (approved):
last changes LGTM

markus2330:
Please rebase after  #4635

kodebach:
@markus2330 rebased, please merge when CI is done

markus2330:
There are several broken builds. Please check CI first before mentioning me.

kodebach:
> Please check CI first before mentioning me.

Sorry, didn't have time to wait for the CI earlier and I hoped nothing would fail, since this is just text change and it worked before the rebase. However, it seems I forgot to update some links.

Not sure, though what the issue on Jenkins is... Seems like some Docker image can't be found. Definitely not something from this PR.

kodebach:
Seems like https://wiki.gnome.org/ is down. But that is clearly a temporary problem, so I think we can ignore the failed Link Checker.

kodebach:
@markus2330 I think this can be merged

markus2330:
@atmaxinger is your comment is an approval?

atmaxinger:
Yes

PR #4314, opened 2022-04-03
[CM T1] jna: added KDBTest

Gratla:
Added a test which checks, if both kdb.get implementations return the same keysets.


Gratla:
@kodebach I reformatted the code with the provided docker container, but the release notes markdown file seems to be still not valid. Sadly, I have not found the reason for this, may I ask for help?

kodebach:
If you formatted locally with the scripts and the [Jenkins formatting check](https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-4314/4/pipeline/604) doesn't complain either, you can ignore the failed `restyled` job. There seems to be an issue with this build job lately (a few people reported it already).

kodebach (changes requested):
Good work. Just one small suggestion.

Also, please rebase with `master` to fix the build server problems.

Issue #4357, opened 2022-05-21
Kotlin binding 1.0


markus2330:
Follow-up from #4294:

- [x] #4393
- [ ] An example application
- [ ] An example using plugin code (current example covers most functions)
- [ ]  Array handling functions
- [ ]  Plugin error creation (already discussed in https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140314385)

Maybe sometime in the future:

- [ ] Use Kotlin native to integrate with the C library directly
- [ ] Plugin extensions

Other work:

- [ ] README.md improvements: more explanations, guide through most important features, ...
- [ ] add link to binding: APP.HOME.LABEL.ORIENTATION.2.DETAILS in src/tools/website/public/assets/translations/en.json
- [ ] improvement of tests
- [ ] API docu missing/could be improved
- [ ] Missing full&working examples, ideally a small application and a plugin that uses Kotlin Serialization

markus2330:
Please tick what was done and write here what is still left undone as future work.

mandoway:
We can't tick the options so here is a list of future work:

- Musts:
  - Publishing (config is present)
- Optionals:
  - An example using plugin code (current example covers most functions)
  - Array handling functions (open discussion for the best fitting place for this functionality, see https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140327073)
  - Plugin error creation (already discussed in https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140314385)
- Maybe sometime in the future:
  - Use Kotlin native to integrate with the C library directly
  - Plugin extensions

markus2330:
Thank you for summarizing! I updated the top post!

PR #4352, opened 2022-05-12
[CM P1] JNA/Kotlin binding

mandoway:
## Basics

- [x] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [x] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [x] References to issues, e.g. `close #X`, are in the commit messages.
- [x] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [x] The PR is rebased with current master.

## Checklist

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned [every code](/.reuse/dep5) not directly written by me in [reuse syntax](https://reuse.software/)

## Review

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

- [ ] Add the "work in progress" label if you do not want the PR to be reviewed yet.
- [ ] Add the "ready to merge" label **if the basics are fulfilled** and no further pushes are planned by you.

---
Implements #4294

PR #4365, opened 2022-05-25
[CM P2] Kotlin binding refactorings

mandoway:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

---
Implements #4357 

- [x] README.md improvements: more explanations, guide through most important features, ...
- [x] src/bindings/README.md and APP.HOME.LABEL.ORIENTATION.2.DETAILS in src/tools/website/public/assets/translations/en.json should list Kotlin as binding
- [x] improvement of tests
- [x] API docu missing/could be improved
- [ ] Missing full&working examples, ideally a small application and a plugin that uses Kotlin Serialization
- [x] error handling? type conversion problems? (see [Comment below](https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140314101))
- [x] easy error&warning set/add (for plugins) (see [Comment below](https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140314385))
- [x] improvements in traversal and array handling (see [Comment below](https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140314936))
- [x] OrDefault should be removed (code-generation as alternative)
- [x] booleans broken https://www.libelektra.org/decisions/boolean (see [Comment below](https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140307418))
- [x] serialization should be generalized (not only JSON)
- [x] publish (see [Comment below](https://github.com/ElektraInitiative/libelektra/pull/4365#issuecomment-1140315437))


Gratla:
https://github.com/ElektraInitiative/libelektra/pull/4352#discussion_r878668638

We checked the behaviour of the JNA binding and there are only empty strings saved, if there is no given value.
Therefore, we are limited because we just mirror the behaviour.

mandoway:
https://github.com/ElektraInitiative/libelektra/pull/4352#discussion_r878703800

Regarding booleans in Keys:  
This is already a bug in the Java binding.  
Since the Kotlin binding depends on the Java binding, this should be fixed in the Java binding.  
Therefore, it is not fixed in this PR.

Gratla:
https://github.com/ElektraInitiative/libelektra/pull/4352#discussion_r878704095

Yes, the first parameter is "expected". In this case, we want to check, if the `key.get()` method has the right behaviour.

mandoway:
https://github.com/ElektraInitiative/libelektra/pull/4352#discussion_r878705038

As a clarification:
`key.byte` in Kotlin calls `Key.getByte()` from the JNA binding.  
So when we test `assertEquals(key.byte, key.get())`, we do test the underlying key value (`key.byte`) against the new implementation of the type inferring `get()`.  
The get method will return a value of type `Byte` automatically, since it inferred it from the usage.

mandoway:
Regarding type conversion problems:

We always use the getters and setters from the JNA binding, so there shouldn't be any type conversion problems, unless they are already present in JNA.

Regarding error handling:

Can you explain this a bit more detailed? What is missing in regards to error handling?  
We don't really see a problem right now.

Gratla:
Concerning the topic "easy error&warning set/add (for plugins)":

We looked into the plugin error handling in the JNA binding and think, that there is no meaningful improvement to be made.
The method `addErrorFunction.apply` only takes the error type and a message which are both required.
This method can also be used within the kotlin binding without restrictions and drawbacks.
Did we miss some important features corresponding plugin error handling?

mandoway:
Regarding array handling:

Since it is possible to convert a KeySet to any Kotlin collection (List, Set, Map), it would be best to convert the KeySet to a List, then manipulate this in Kotlin native ways and finally convert it back to a KeySet.

This also acts as an abstraction layer to the KeySet data structure and removes the need for extra KeySet manipulation functions.

mandoway:
Regarding publish:

We added the config for publishing but did not yet upload the artifact, since we thought this should be done after this PR is merged and all maintainers are happy with it :)

kodebach:
> Since it is possible to convert a KeySet to any Kotlin collection (List, Set, Map), it would be best to convert the KeySet to a List, then manipulate this in Kotlin native ways and finally convert it back to a KeySet.

Depending on what exactly you are trying to do, this might be a very wasteful operation. If for example, you just want to filter a `KeySet`, you definitely shouldn't convert the `KeySet` into a JVM `List`/`Set`/`Map` and then back. That would be totally unnecessary. However, since `KeySet` implements `SortedSet`, I think it should automatically be possible to do `keySet.filterTo(KeySet.create()) { ... }`.

> This also acts as an abstraction layer to the KeySet data structure and removes the need for extra KeySet manipulation functions.

Abstraction layers are fine, that's why e.g. `KeySet` implements the `SortedSet` interface. But relying on conversion to JVM data structures to achieve things is very bad for performance.

That being said, I think question about array handling was more about adding e.g.

```kotlin
fun KeySet.addToArray(arrayParent: String, key: Key) {
   // lookup arrayParent in `this`, check that has `meta:/array` and extract index
   // increment index, construct keyname from arrayParent and index and set on key
   // insert modified key into `this`
}

fun KeySet.insertIntoToArray(arrayParent: String, index: Long key: Key) {
   // similar to addToArray, but instead of adding to the of the array we
   // check that arrayParent has `meta:/array` and that index is in bounds
   // move other keys so that there is space at index
   // modify name of key and insert at index
}
```

For the cases above, the native JVM methods and structures don't help at all, since this is Elektra specific. To use methods like `List::add` and `List::insert` you'd have to do some very unnecessary copying of that.

In the end, however, I think that such methods are probably better implemented in C and than exposed via the binding. Manipulating the `KeySet` via the JNA binding also adds lots of unnecessary translation. Even if this is not implemented in C, it should be part of the Java binding, since none of this needs Kotlin to work.

@markus2330 I don't think adding methods like the ones above makes much sense. If you had something else in mind, please clarify.

> We added the config for publishing but did not yet upload the artifact

That's perfect. AFAIK publishing should happen with the next release. @tucek can you please check that the Maven config is correct.

markus2330:
Can you please create an issue about the bugs in the JNA binding? (null values and boolean look like bugs)

> What is missing in regards to error handling?

Which statement are you referring to? If it was about plugins, I might have been talking about a wrapper that handles language-specific exceptions and appends error/warnings in the case of unhandled exceptions. If a JNA-only solution suffices, then please add it to the JNA issue.

> Did we miss some important features corresponding plugin error handling?

Probably only an example plugin is missing showing how it works.

> Since it is possible to convert

Again, probably only docu/example missing.

> We added the config for publishing but did not yet upload the artifact, since we thought this should be done after this PR is merged and all maintainers are happy with it :)

Yes, perfect :rocket:

mandoway:
Added an issue regarding JNA bugs: https://github.com/ElektraInitiative/libelektra/issues/4368

mandoway:
> Please fix the spelling issues reported by Jenkins, so that the other tests can run.

One of these issues is actually in a code block and would be wrong if fixed (Json.encode() to JSON.encode())
Can we somehow ignore certain lines or should we still fix it, so the build server is happy?

markus2330:
We already had an issue that non-correct JSON spellings are required, so I tend to simply remove lines 69 and 70 in `scripts/sed`.

src/bindings/jna/libelektra-kotlin/src/main/kotlin/org/libelektra/dsl/KeySetDSL.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    fun <T> key(name: String, value: T? = null, vararg metaKeys: Key) {
+        keys.add(
+            Key.create(name, value).apply {
+                metaKeys.forEach {
+                    setMeta(it.name, it.string)
+                }
+            }
+        )
+    }
```
kodebach, line 33:
```suggestion
    fun key(name: String, initializer: KeyDSL.() -> Unit = {}) {
        keys.add(keyOf(name, initializer))
    }
```

I think that would be more consistent with the `keyOf` DSL function.

src/bindings/jna/libelektra-kotlin/src/main/kotlin/org/libelektra/keySetExt/KeySetSerializer.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.libelektra.KeySet
+import org.libelektra.dsl.keySetOf
+
+/**
+ * Basic serializer for KeySets
+ *
+ * Can be used to serialize KeySets in different formats, e.g. JSON
+ */
+@OptIn(ExperimentalSerializationApi::class)
+class KeySetSerializer : KSerializer<KeySet> {
```
kodebach, line 20: The docs should mention that this doesn't support metadata

PR #4399, opened 2022-07-04
[kdb record] Discussion man page

atmaxinger:
This is the preliminary man page for the `kdb record` tooling. Let's focus on discussing the user-facing parts of it. If we can agree on this feature-set we can then discuss the technical part of it in a separate issue.

### DIFF
--- a/doc/help/kdb-record.md
+++ b/doc/help/kdb-record.md
@@ -13,27 +13,27 @@ Where `<COMMAND>` is one of the following:
 - `delete [session_name]`: delete the recording session.
 - `assert [--meta meta_name] <key_name>`: marks a key to be used for value-based validation during import.
 - `list`: list all available recording sessions.
-- `name`: prints the name of the currently active session.
+- `name`: prints the name of the currently active recording session.
 - `status [session_name]`: shows a human-readable list of changes.
-- `export  [--format=replay|diff] [session_name]`: creates a file with the changes in the session.
+- `export  [--format=replay|diff] [session_name]`: creates a file with the changes of the recording session.
 - `import [--disable-validation] <file_name>`: imports files created by the `export` command.
 
 ## DESCRIPTION
 
 The `kdb record` utility is used to manage session recordings.
 Only a single recording session can be active per user at a time.
-A user can have multiple non-active sessions.
+A user can have multiple non-active recording sessions.
 
-If a session is active, every change to the keys, values and meta-data stored in Elektra will be logged into that session.
+If a recording session is active, every change to the keys, values and meta-data stored in Elektra will be logged into that recording session.
 
 ## ENVIRONMENT VARIABLES
 
-- `ELEKTRA_RECORD_ACTIVE_SESSION`: If this environment variable is set, the specified session will be the active session.
+- `ELEKTRA_RECORD_ACTIVE_SESSION`: If this environment variable is set, the specified recording session will be the active recording session.
 
 ## COMMON PARAMETERS
 
 Most commands allow specification of the name of the recording session they use.
-If no session name is provided, the active session will be used.
+If no recording session name is provided, the active recording session will be used.
 
 Notable exceptions are the `start` and `resume` commands.
 
@@ -42,38 +42,38 @@ Notable exceptions are the `start` and `resume` commands.
 ### `start [session_name]`
 The `start` command is used to start a new recording session.
 If the environment variable `ELEKTRA_RECORD_ACTIVE_SESSION` is set, this command has no immediate effect.
-As soon as the environment variable is unset, the session started with this command will be active.
+As soon as the environment variable is unset, the recording session started with this command will be active.
 
-This command will accept a single optional argument that is the name of the session.
+This command will accept a single optional argument that is the name of the recording session.
 If the name is not provided, a random name will be generated.
 
-The command will output an error if a session with the specified name already exists.
+The command will output an error if a recording session with the specified name already exists.
 
-On success, the name of the session will be printed to `stdout`, and the return code is `0`.
+On success, the name of the recording session will be printed to `stdout`, and the return code is `0`.
 
 ### `stop`
 The `stop` command is used to stop the active recording session.
-The session can then be resumed using the `resume` command with the appropriate session name.
+The recording session can then be resumed using the `resume` command with the appropriate recording session name.
 Stopping a recording session will not delete any recorded changes.
 
 ### `resume [session_name]`
-The `resume` command is used to resume an existing session.
+The `resume` command is used to resume an existing recording session.
 If the environment variable `ELEKTRA_RECORD_ACTIVE_SESSION` is set, this command has no immediate effect.
-As soon as the environment variable is unset, the session resumed with this command will be active.
+As soon as the environment variable is unset, the recording session resumed with this command will be active.
 
-This command will accept a single required argument that is the name of the session.
-If the name is not provided, or a session with the specified name does not exist, the command will output an error.
+This command will accept a single required argument that is the name of the recording session.
+If the name is not provided, or a recording session with the specified name does not exist, the command will output an error.
 
-On success, the name of the session will be printed to `stdout`, and the return code is `0`.
+On success, the name of the recording session will be printed to `stdout`, and the return code is `0`.
 
 ### `reset [session_name]`
 The `reset` command will clear the specified recording.
-If no session name argument is provided, the currently active session will be reset.
+If no recording session name argument is provided, the currently active recording session will be reset.
 
 ### `delete [session_name]`
 Deletes the specified recording session.
-It the optional argument is not provided, the current session will be deleted.
-If it is the currently active session, it will be stopped.
+It the optional argument is not provided, the current recording session will be deleted.
+If it is the currently active recording session, it will be stopped.
 
 ### `assert [--meta meta_name] <key_name>`
 This command marks a key to be used for validation when importing the recording.
@@ -87,25 +87,25 @@ Lists all known recording sessions.
 
 ### `name`
 Prints out the name of the currently active recording session.
-If no session is active, the return code is `1`, otherwise `0`.
+If no recording session is active, the return code is `1`, otherwise `0`.
 
 ### `status [session_name]`
-Shows all the changes in the specified session in human-readable format.
+Shows all the changes in the specified recording session in human-readable format.
 Outputs to `stdout`.
-If no session is specified, the currently active session is used.
+If no recording session is specified, the currently active recording session is used.
 
 ### `export  [--format=replay|diff] [session_name]`
-Exports the changes made in the specified session.
+Exports the changes made in the specified recording session.
 Outputs to `stdout`.
-If no session is specified, the currently active session is used.
+If no recording session is specified, the currently active recording session is used.
 
 The `export` command has the ability to output two different types: diffs and replays.
 
 The `--format=diff` argument (default) only exports the changes between the start and end states of the recording session.
-This means, e.g., that when you add a key and remove it in the same session, it will not be exported.
+This means, e.g., that when you add a key and remove it in the same recording session, it will not be exported.
 
-The `--format=replay` argument creates a file that contains every modifying operation made in the session in the order they were performed.
-This is regardless whether at the end of the session the value has really changed or is the same as before.
+The `--format=replay` argument creates a file that contains every modifying operation made in the recording session in the order they were performed.
+This is regardless whether at the end of the recording session the value has really changed or is the same as before.
 Files created using `--format=replay` serve as a transaction log.
 
 
--- a/doc/help/kdb-record.md
+++ b/doc/help/kdb-record.md
@@ -14,7 +14,7 @@ Where `<COMMAND>` is one of the following:
 - `list`: list all available recording sessions.
 - `name`: prints the name of the currently active session.
 - `changes [session_name]`: shows a human-readable list of changes.
-- `export  [--replay|--diff] [session_name]`: creates a file with the changes in the session.
+- `export  [--format=replay|diff] [session_name]`: creates a file with the changes in the session.
 - `import [--disable-validation] <file_name>`: imports files created by the `export` command.
 
 ## DESCRIPTION
@@ -82,17 +82,17 @@ Shows all the changes in the specified session in human-readable format.
 Outputs to `stdout`.
 If no session is specified, the currently active session is used.
 
-### `export  [--replay|--diff] [session_name]`
+### `export  [--format=replay|diff] [session_name]`
 Exports the changes made in the specified session.
 Outputs to `stdout`.
 If no session is specified, the currently active session is used.
 
 The `export` command has the ability to output two different types: diffs and replays.
 
-The `--diff` flag (default) only exports the changes between the start and end states of the recording session.
+The `--format=diff` argument (default) only exports the changes between the start and end states of the recording session.
 This means, e.g., that when you add a key and remove it in the same session, it will not be exported.
 
-The `--replay` flag creates a file that contains every modifying operation made in the session in the order they were performed.
+The `--format=replay` argument creates a file that contains every modifying operation made in the session in the order they were performed.
 This is regardless whether at the end of the session the value has really changed or is the same as before.
 Files created using `--replay` serve as a transaction log.
 
@@ -120,7 +120,7 @@ Marked system:/hosts/ipv6/localhost to be ::1
 Changed system:/hosts/ipv4/www.example.com from 1.2.3.4 to 8.8.8.8
 Added system:/hosts/ipv4/www.beispiel.de with value 4.4.4.4
 Removed system:/hosts/ipv4/www.ejemplo.es
-$ kdb record export --replay
+$ kdb record export --format=replay
 <export output. format not yet defined>
 $ kdb record list
 AddNewHosts
--- a/doc/help/kdb-record.md
+++ b/doc/help/kdb-record.md
@@ -6,7 +6,7 @@
 
 Where `<COMMAND>` is one of the following:
 
-- `start [session_name]`: start a new recording session.
+- `start [--parent key] [session_name]`: start a new recording session.
 - `stop`: stop the active recording session.
 - `resume <session_name>`: resume a stopped recording session.
 - `reset [session_name]`: remove all data from the recording session.
@@ -39,13 +39,16 @@ Notable exceptions are the `start`, `resume` and `delete` commands.
 
 ## COMMANDS
 
-### `start [session_name]`
+### `start [--parent key] [session_name]`
 The `start` command is used to start a new recording session.
 If the environment variable `ELEKTRA_RECORD_ACTIVE_SESSION` is set, this command has no immediate effect.
 As soon as the environment variable is unset, the recording session started with this command will be active.
 
-This command will accept a single optional argument that is the name of the recording session.
+This command will accept an optional argument that is the name of the recording session.
 If the name is not provided, a random name will be generated.
+
+The also optional argument `--parent key` specifies that only changes for the key `key` or below should be recorded.
+E.g., `start --parent user:/foo` will only record changes below `user:/foo`. The key may also be a cascading key, e.g. `--parent /foo`.
 
 The command will output an error if a recording session with the specified name already exists.
 
--- a/doc/help/kdb-record.md
+++ b/doc/help/kdb-record.md
@@ -8,9 +8,9 @@ Where `<COMMAND>` is one of the following:
 
 - `start [session_name]`: start a new recording session.
 - `stop`: stop the active recording session.
-- `resume [session_name]': resume a stopped recording session.
+- `resume <session_name>': resume a stopped recording session.
 - `reset [session_name]`: remove all data from the recording session.
-- `delete [session_name]`: delete the recording session.
+- `delete <session_name>`: delete the recording session.
 - `assert [--meta meta_name] <key_name>`: marks a key to be used for value-based validation during import.
 - `list`: list all available recording sessions.
 - `name`: prints the name of the currently active recording session.
@@ -35,7 +35,7 @@ If a recording session is active, every change to the keys, values and meta-data
 Most commands allow specification of the name of the recording session they use.
 If no recording session name is provided, the active recording session will be used.
 
-Notable exceptions are the `start` and `resume` commands.
+Notable exceptions are the `start`, `resume` and `delete` commands.
 
 ## COMMANDS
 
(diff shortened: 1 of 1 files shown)


doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+The `kdb record` utility is used to manage session recordings.
+Only a single recording session can be active per user at a time.
+A user can have multiple non-active sessions.
+
+If a session is active, every change to the keys, values and meta-data stored in Elektra will be logged into that session.
+
+## ENVIRONMENT VARIABLES
+
+- `ELEKTRA_RECORD_ACTIVE_SESSION`: If this environment variable is set, the specified session will be the active session.
```
kodebach, line 30:
```suggestion
- `ELEKTRA_RECORD_SESSION`: If this environment variable is set, the specified session will be the active session.
```

Personally, I prefer shorter names for environment variables, but `ELEKTRA_RECORD_ACTIVE_SESSION` would be fine too.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+## COMMON PARAMETERS
+
+Most commands allow specification of the name of the session they should work against.
+If no session name is provided, the active session will be used.
```
kodebach, line 35: And if no session is provided and there is no active session?

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+A notable exception is the `start` command.
+If no session name is provided, it will create a random name.
+
+## COMMANDS
+
+### `start [session_name]`
+The `start` command is used to start a new or resume an existing session.
+If the environment variable `ELEKTRA_RECORD_ACTIVE_SESSION` is set, this command has no immediate effect.
+As soon as the environment variable is unset, the session started or resumed with this command will be active.
```
kodebach, line 45: We should print a warning in this case, or make it an error case and require unsetting the env-var before `start` works.

The current behaviour has lots of potential for mistakes, e.g. when `kdb record start` is used within a script that assumes it starts a new session.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+This command will accept a single optional argument that is the name of the session.
+If the name is not provided, a random name will be generated.
+
+On success, the name of the session will be printed to `stdout`, and the return code is `0`.
```
kodebach, line 50: IMO even if it is allowed, in the case where the env-var is set, the exit code should also be non-zero. That way exit code 0 means session started and actually in active.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+### `stop`
+The `stop` command is used to stop the active recording session.
+The session can then be resumed using the `start` command with the appropriate session name.
```
kodebach, line 54: "Resuming a session" needs more explanation. What happens if the KDB is in an incompatible state when the session is resumed. For example:

1. Start session `X`
2. Add key `user:/foo = 123`
3. Stop session `X`
4. Delete key `user:/foo`
5. Resume session `X`
6. Add key `user:/foo = 456`

Up until including 5 everything is fine. But at 6 the recording will record that `user:/foo` is newly added, when the recording already contains an entry that says "newly add `user:/foo`". Depending on the implementations this may be trivial to handle. However, I think maybe taking a different approach is easier (i.e. has fewer edge cases):

Instead of "resuming a session", I would say you always need to start a new session, but there is a kind command that concatenates finished sessions into a new one e.g. `kdb record cat`. That way the command can just check for inconsistencies and throw an error, if the sessions cannot be combined.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+Stopping a recording session will not delete any recorded changes.
+
+### `reset [session_name]`
+The `reset` command will clear the specified recording.
+If no session name argument is provided, the currently active session will be reset.
+
+### `delete [session_name]`
+Deletes the specified recording session.
+It the optional argument is not provided, the current session will be deleted.
+If it is the currently active session, it will be stopped.
```
kodebach, line 64: IMO an optional argument doesn't make sense here. It should be required. Seems far too dangerous to accidentally delete the active session by forgetting the argument.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+### `mark [--meta meta_name] <key_name>`
+This command marks a key to be used for validation when importing the recording.
+By default, the values of a marked key will be compared when importing.
+The import will fail if the key has a different value.
+
+Using the `--meta` flag, one can specify to use the value of a specific meta instead of the value of the key.
```
kodebach, line 71:
```suggestion
Using the `--meta` flag, one can specify to use the value of a metakey instead of the value of the key.
```

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+### `name`
+Prints out the name of the currently active recording session.
+If no session is active, the return code is `1`, otherwise `0`.
+
+### `changes [session_name]`
+Shows all the changes in the specified session in human-readable format.
+Outputs to `stdout`.
+If no session is specified, the currently active session is used.
+
+### `export  [--replay|--diff] [session_name]`
```
kodebach, line 85:
```suggestion
### `export  [--format=replay|diff] [session_name]`
```

Using separate `--replay`/`--diff` makes it seem like both can be used at the same time.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+Exports the changes made in the specified session.
+Outputs to `stdout`.
+If no session is specified, the currently active session is used.
+
+The `export` command has the ability to output two different types: diffs and replays.
+
+The `--diff` flag (default) only exports what has actually changed in the recording session.
```
kodebach, line 92:
```suggestion
The `--diff` flag (default) only exports the changes between the start and end states of the recording session.
```

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+### `import [--disable-validation] <file_name>`
+Imports files created by the `export` command.
+
+By default we check whether the current values in the key database match the "old" values in the files.
+The values of marked keys will also be checked.
+Those checks can be disabled using the `--disable-validation` flag.
+
+
+## EXAMPLE
```
kodebach, line 108: IMO the examples should be listed with the descriptions of the individual commands. At least for `kdb record changes` the descriptions is kinda useless without an example.

kodebach (commented):
Overall this looks very promising. I agree with most of the functionality, but there are some edge cases we need to think about (e.g. when "resuming a session" and marking in diffs).

Also something I found missing: What about a rollback? `kdb record reset` clears the active session, but it doesn't seem to rollback the changes that I made. I think in general, a command to apply a diff/replay-log in reverse would be a good idea and shouldn't be too hard to do (just swap old/new values).

doc/help/kdb-record.md
atmaxinger: Then we should fail with an error message and an appropriate return code.

doc/help/kdb-record.md
atmaxinger: Yes, good catch!

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+The `reset` command will clear the specified recording.
+If no session name argument is provided, the currently active session will be reset.
+
+### `delete [session_name]`
+Deletes the specified recording session.
+It the optional argument is not provided, the current session will be deleted.
+If it is the currently active session, it will be stopped.
+
+### `mark [--meta meta_name] <key_name>`
+This command marks a key to be used for validation when importing the recording.
```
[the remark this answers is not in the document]
atmaxinger: I think we should only allow marking keys before any changes have occured in the session.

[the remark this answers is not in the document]
kodebach: Actually, I looked at it again and I think the only requirement would be that the key being marked does not occur yet in the current session. If the key is already in the session, marking shouldn't be needed, since the current value is already recorded in the session.

So I would suggest:

Marking a key that is already part of the active session is a no-op and will produce a warning.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+By default, the values of a marked key will be compared when importing.
+The import will fail if the key has a different value.
+
+Using the `--meta` flag, one can specify to use the value of a metakey instead of the value of the key.
```
kodebach, line 71: One more thing: There should also be a `--recursive` flag that marks the key and everything below. It shouldn't be hard to implement, since internally every entry in the session log will need to record a whole KeySet of old/new values anyway (since `kdbSet` only deals with KeySets and we cannot guarantee that only one key changed).

doc/help/kdb-record.md
markus2330: > Maybe even add an option to make it an error or just a warning.

Please don't go this way. Such problems are indications that the usability is bad.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+Exports the changes made in the specified session.
+Outputs to `stdout`.
+If no session is specified, the currently active session is used.
+
+The `export` command has the ability to output two different types: diffs and replays.
+
+The `--format=diff` argument (default) only exports the changes between the start and end states of the recording session.
+This means, e.g., that when you add a key and remove it in the same session, it will not be exported.
+
+The `--format=replay` argument creates a file that contains every modifying operation made in the session in the order they were performed.
```
markus2330, line 95: These functionalities are so different that they should be different commands. Maybe the "replay" feature should not even be within the whole "kdb record" framework. (It seems to me that the current description of this man page, with the exception of this `export --format=replay`, can be implemented as a base+ours+assert keyset. Only the replay needs an actual log of everything that happened?)

markus2330 (commented):
This is definitely an improvement! But I think we could avoid many discussions if you polish your descriptions more (write **complete** man pages and a tutorial perfectly describing your use cases). There are still many obvious inconsistencies which yield many opinions how these cases should be handled. These discussions lead to even more inconsistencies.

doc/help/kdb-record.md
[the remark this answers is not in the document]
atmaxinger: First of all 

> We could consider to use SUDO_USER (preferred to USER) to be used in our resolver. 

I'm fine with this. This would fix a lot of problems, especially with storing the recording session within Elektra itself. I will create an issue for this. If we do it this way, the `ELEKTRA_RECORD_ACTIVE_SESSION` will be dropped.


> It is unclear if [session_name] or ELEKTRA_RECORD_ACTIVE_SESSION will be used.

`ELEKTRA_RECORD_ACTIVE_SESSION` would be the active session.
If you specify the `[session_name]` argument for any command, you instruct that command NOT to use the active session. So that way `[session_name]` will be used.

> With the env var you don't have any control over how many sessions are open? 

Yes, this is a limitation of this approach. If we store the session in external files we don't have control over sessions that are controlled via `ELEKTRA_RECORD_ACTIVE_SESSION`. This is a trade-off. If we store them in Elektra like I initially proposed, we do actually have control over those sessions.

> It does not solve the sudo problem as sudo per default removes env vars.

Yes. But it can be instructed to keep them using the `-E` flag. This is something we should note in the manual.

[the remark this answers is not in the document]
kodebach: > It does not solve the sudo problem as sudo per default removes env vars.

True, but there is `sudo -E` to preserve the environment. So at least it is possible to solve it this way everything else that was suggested before didn't solve the sudo problem at all.

> We could consider to use `SUDO_USER` (preferred to USER) to be used in our resolver. 

I did not think of that. I admit, using `$SUDO_USER` in the `resolver` is probably a better solution than the env-var for the sudo problem. If we go down this root, we should definitely create a separate issue, since there is more to discuss here (e.g. permissions when creating files in `user:/` with `sudo`).

> With the env var you don't have any control over how many sessions are open? 

I assume you mean different shell sessions could use different `kdb record` sessions. IMO this would actually be an advantage, since you can't accidentally influence your recording from a different shell (or maybe from an active background process).

The currently  man page says "Only a single recording session can be active per user at a time.", but from a technical POV I think this can be seen as when a `kdbSet` happens exactly 0 or exactly 1 sessions are active.

> I think it is either/or: having a start/stop interface or using an env var.

Yes, probably just one would be better. Personally, I tend towards the env-var, but if we can solve all the problems, I'm fine with storing the active session in the KDB as well.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+- `start [session_name]`: start or resume a recording session.
+- `stop`: stop the active recording session.
+- `reset [session_name]`: remove all logged data from the recording session.
+- `delete [session_name]`: delete the recording session.
+- `mark [--meta meta_name] <key_name>`: marks a key to be used for validation.
+- `list`: list all available recording sessions.
+- `name`: prints the name of the currently active session.
+- `changes [session_name]`: shows a human-readable list of changes.
+- `export  [--replay|--diff] [session_name]`: creates a file with the changes in the session.
```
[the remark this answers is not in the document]
kodebach: > Export requires an extra argument for the format. Diff we can have one format.

Not sure what "export vs. diff" would mean. The options are replay and diff. Also both variants could in theory have multiple formats from which the user could choose.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+If no session is specified, the currently active session is used.
+
+The `export` command has the ability to output two different types: diffs and replays.
+
+The `--format=diff` argument (default) only exports the changes between the start and end states of the recording session.
+This means, e.g., that when you add a key and remove it in the same session, it will not be exported.
+
+The `--format=replay` argument creates a file that contains every modifying operation made in the session in the order they were performed.
+This is regardless whether at the end of the session the value has really changed or is the same as before.
+Files created using `--replay` serve as a transaction log.
```
kodebach, line 97:
```suggestion
Files created using `--format=replay` serve as a transaction log.
```

@atmaxinger You missed one ;)

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+Exports the changes made in the specified session.
+Outputs to `stdout`.
+If no session is specified, the currently active session is used.
+
+The `export` command has the ability to output two different types: diffs and replays.
+
+The `--diff` flag (default) only exports what has actually changed in the recording session.
+This means, e.g., that when you add a key and remove it in the same session, it will not be exported.
+
+The `--replay` flag creates a file that contains every modifying operation made in the session in the order they were performed.
```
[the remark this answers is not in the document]
kodebach: > Why would this be needed?

From a purely technical standpoint: There is no way of knowing what plugins do during `kdbSet`. Therefore, we cannot guarantee that a full replay of the individual `kdbSet` changes has the same result (especially in regard to  side effects) as a single collapsed diff.

Take for example this sequence of commands:

```
kdb set user:/foo 123
kdb record start
kdb rm user:/foo
kdb set user:/foo 123
kdb record stop
```

The full replay would remove and then newly create `user:/foo`. The collapsed diff on the other hand would be empty. Especially such a case could have big differences in side effects. Imagine, if the config is backed by an SQL database and there are triggers attached to the tables that store the config. The full replay and the collapsed diff would cause different triggers to run.

Now from a philosophical POV, you could probably argue that both variants should have the same effect. But I don't think we can assume that.

There is also a second technical reason: Recording the replay log and collapsing it into a diff is more efficient in terms of storage. It is similar to a Copy-on-Write filesystem. We only record what actually changed. And if we already record the log, we might as well provide an option to export and replay it. The other option would be, not recording the log and instead producing a diff based on saved snapshots. However, with the current setup (`kdb record` records changes to the whole KDB), every `kdb start` would have to make a complete copy of the whole KDB, so that we can later produce a diff.

> Isn't there an inconsistency with --format=replay vs. --replay?

I think that's a mistake, from a recent commit (possibly you also look at different versions of the file).

doc/help/kdb-record.md
kodebach: > Only the replay needs an actual log of everything that happened?

See above: https://github.com/ElektraInitiative/libelektra/pull/4399#discussion_r913869936

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
+# kdb-record(1) -- Session recording for Elektra
+
+## SYNOPSIS
+
+`kdb record <COMMAND>`
+
+Where `<COMMAND>` is one of the following:
+
+- `start [session_name]`: start or resume a recording session.
```
kodebach, line 9: I think optionally adding a `parentKey` might be nice. So this would start a session `mysession` that only records changes that happen below `user:/foo`:

```
kdb record start --parent user:/foo mysession
```

So below the first two lines and the last line are recorded, but the third line is not recorded in the session:
```
kdb set user:/foo/bar 123
kdb set user:/foo/foo 456
kdb set user:/bar 789
kdb rm user:/foo/bar
```

Alternatively, the `parentKey` could be given to the `export` command. But restricting at `kdb record start` would avoid storing unnecessary data.

doc/help/kdb-record.md
atmaxinger: Indeed, thank you :)

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+$ ELEKTRA_RECORD_ACTIVE_SESSION=JustATest kdb record name
+JustATest
+$ kdb record reset
+$ kdb record changes
+<no changes>
+$ kdb record import /tmp/invalid
+Error importing! 
+Existing value for key system:/hosts/ipv6/localhost does not match! (shoud be ::1, is ::2)
+$ kdb record import --disable-validation /tmp/invalid
+Successfully imported!
```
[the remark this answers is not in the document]
atmaxinger: I agree this example is not clear enough.

If `system:/hosts/ipv6/localhost` was only a marked key, its value has not changed. Otherwise it has the value from the imported file.

All other keys have the value from the imported file.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+
+By default we check whether the current values in the key database match the "old" values in the files.
+The values of marked keys will also be checked.
+Those checks can be disabled using the `--disable-validation` flag.
+
+
+## EXAMPLE
+
+```
+$ kdb record start AddNewHosts
```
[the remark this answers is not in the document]
atmaxinger: I'm not quite sure what you mean. Can you point me to an example?

Do you mean this:
grafik
?

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+- `stop`: stop the active recording session.
+- `reset [session_name]`: remove all logged data from the recording session.
+- `delete [session_name]`: delete the recording session.
+- `mark [--meta meta_name] <key_name>`: marks a key to be used for validation.
+- `list`: list all available recording sessions.
+- `name`: prints the name of the currently active session.
+- `changes [session_name]`: shows a human-readable list of changes.
+- `export  [--replay|--diff] [session_name]`: creates a file with the changes in the session.
+- `import [--disable-validation] <file_name>`: imports files created by the `export` command.
+
```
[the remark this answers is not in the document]
atmaxinger: Is this really part of this tool? If I remember correctly, we said that this was out of scope for now but we'll lay the technical foundations for this functionality if it's not too much of a hassle.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+By default we check whether the current values in the key database match the "old" values in the files.
+The values of marked keys will also be checked.
+Those checks can be disabled using the `--disable-validation` flag.
+
+
+## EXAMPLE
+
+```
+$ kdb record start AddNewHosts
+AddNewHosts
```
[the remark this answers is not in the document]
atmaxinger: My initial thought was to make it consistent with the behaviour when the user does not specify a name.
However I can see how someone might find this behaviour irritating.

doc/help/kdb-record.md
```diff
@@ -0,0 +1,144 @@
...
+- `start [session_name]`: start or resume a recording session.
+- `stop`: stop the active recording session.
+- `reset [session_name]`: remove all logged data from the recording session.
+- `delete [session_name]`: delete the recording session.
+- `mark [--meta meta_name] <key_name>`: marks a key to be used for validation.
+- `list`: list all available recording sessions.
+- `name`: prints the name of the currently active session.
+- `changes [session_name]`: shows a human-readable list of changes.
+- `export  [--replay|--diff] [session_name]`: creates a file with the changes in the session.
+- `import [--disable-validation] <file_name>`: imports files created by the `export` command.
```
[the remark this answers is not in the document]
markus2330: I am not a big fan of `--force` either but it makes sense to (additionally) support it as people will immediately look for it if something fails for a non-permanent reason. `validation` in specific, however, is misleading, maybe `--disable-assertions` will do?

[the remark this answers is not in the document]
atmaxinger: Yes, `--disable-assertions` would be a good compromise, as validation could entail much more, e.g. checking that the file format is correct.

doc/help/kdb-record.md
atmaxinger: I've added a new command `resume`. We can add those checks in this command.
I think this is more practical than starting a new session and concatenating. It enables us to throw an error before it's too late.

doc/help/kdb-record.md
atmaxinger: Yes, this could easily be checked for with the already existing `keyIsBelowOrSame` function. Seems like a no-brainer. Should we also allow cascading keys?

doc/help/kdb-record.md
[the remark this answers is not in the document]
kodebach: I'd suggest `--no-assertions` or `--without-assertions`. The `--disable` sounds a bit like it changes some config somewhere instead of just being a temporary flag.

doc/help/kdb-record.md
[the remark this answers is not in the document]
kodebach: But in general, I think we should define all potential features in this PR, and then you can decide what you do first, once you also know the technical details.

doc/help/kdb-record.md
kodebach: > Should we also allow cascading keys?

Yes, I would allow them. That makes it possible to isolate a session to the config for a single application (which might include `spec:/` keys).

doc/help/kdb-record.md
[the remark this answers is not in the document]
markus2330: Yes, good point.

We should watch out that this is consistent within the whole tooling. Currently `kdb` is inconsistent, we have --no-newline but --without-elektra and --with-recommends.

doc/help/kdb-record.md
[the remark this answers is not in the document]
kodebach: Re https://github.com/ElektraInitiative/libelektra/issues/4401#issuecomment-1179551907


> We would only store keys that changed/are relevant. So every change makes at most two keys: in base and in ours (or in assert).

I clearly misunderstood what you meant by base+ours+assert. I thought you want to collect base only once at the start, but since we don't know what will change, we need to collect everything. I explained this in my original response above ("The other option would be [...]").

Turns out your solution is actually very similar to the "collect a log of everything" solution. The only difference seems to be whether we remember intermediate values during an active session, or just the initial and final values. But why would that mean we need two separate implementations (suggested in https://github.com/ElektraInitiative/libelektra/issues/4401#issuecomment-1179526126)?

PR #4462, opened 2022-09-12
Discussion: record-elektra use cases

atmaxinger:
This PR should contain the major use cases/features for the following scenario:

- Export existing configuration into an Ansible playbook to be applied to other machines also running Elektra
- Record changes on any machine configured using above playbook
- Export those changes to yet another playbook
- Be able to apply those changes to all the other configured machines

To focus on the overall picture of this scenario I have currently intentionally left out all the different use cases we have already talked about for the record-elektra tooling itself - like session handling. See #4399.


doc/usecases/record_elektra/UC_change_notifications.md
```diff
@@ -0,0 +1,23 @@
+# Use Case: Notify plugins when data in the KDB changes
+
+## Summary
+
+- **Title:** Notify plugins when data in the KDB changes
+- **Scope:** Configuration/Logging
+- **Level:** Developer goal
+- **Actors:** Plugin developer
+- **Brief:** It should be possible for plugins to be notified when data in KDB changes.
```
kodebach, line 9: IMO this is to general. To me this reads like: "A plugin whose main purpose is something else (e.g. `type`) can be notified of changes via an extra API." I think that would lead to chaos.

IMO what we should have is:

> Some plugins are only executed when something changed in the KDB (*) and as an input receive the keys that changed in their old and new state.

Importantly, the plugins should already receive the changeset as an input. In other words, `libelektra-kdb` (or an extra component between `libelektra-kdb` and the plugins) does the change tracking not the plugins themselves. This is important, because otherwise every such "change plugin" would keep its own copy of all keys to keep track of the old values.

---

(*) This is already the case for standard mounted backend plugins. They are only called, if there is a changed key in their mountpoint. But they always receive all keys for the mountpoint and don't receive the old state.

kodebach (commented):
Looks good to me. I have a small note on what actually is recorded. I'd also like some clarification on the "notify plugins" thing.

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
```diff
@@ -0,0 +1,18 @@
+# Use Case: Fail application of ansible-libelektra if certain keys differ in value
+
+## Summary
+
+- **Title:** Fail application of ansible-libelektra if certain keys differ in value
```
markus2330, line 5: Useless to state the same title again.

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
```diff
@@ -0,0 +1,18 @@
...
+- **Scope:** ansible-libelektra
+- **Level:** User goal
+- **Actors:** User (usually sysadmin)
```
markus2330, line 8:
```suggestion
- **Actors:** System administrator
```

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
```diff
@@ -0,0 +1,18 @@
...
+- **Brief:** We should be able to mark certain keys in the tasks for a specific value. If the keys do have a different value on the target machine, ansible-libelektra should fail.
```
markus2330, line 9: It is still not clear to me why we the system administrator ansible-libelektra to fail? What is the goal? Give a bit background why the user needs this. Give examples.

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Scenarios
+
+- **Precondition:** The users marks a certain key in the task
```
markus2330, line 13:
```suggestion
- **Precondition:** The user marks a certain key in the task
```

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
```diff
@@ -0,0 +1,18 @@
...
+- **Main success scenario:** The key on the target machine has a different value or is not present -> fail the execution
+- **Alternative scenario:** The key on the target machine has the same value -> continue the execution
+- **Error scenario:** -
+- **Postcondition:** -
```
markus2330, line 17: Isn't it a postcondition that we only continue if the value is as wanted?

doc/usecases/record_elektra/UC_ansible-libelektra_merge_strategies.md
```diff
@@ -0,0 +1,18 @@
+# Use Case: Select merge strategy in ansible-libelektra
+
+## Summary
+
+- **Title:** Select merge-strategy in ansible-libelektra
+- **Scope:** ansible-libelektra
+- **Level:** User goal
+- **Actors:** User (usually sysadmin)
+- **Brief:** It should be possible to specify the merge strategy used, e.g. OURS, THEIRS, ABORT.
```
markus2330, line 9: Not only to specify but that ansible-libelektra actually does the merging

doc/usecases/record_elektra/UC_ansible-libelektra_merge_strategies.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** If the user has specified a strategy, this strategy is applied on merge conflicts.
+- **Alternative scenario:** If no strategy is specified, a sane (and documented) default strategy should be used.
+- **Error scenario:** -
```
markus2330, line 16: Describe what can go wrong in merging...

doc/usecases/record_elektra/UC_ansible-libelektra_remove_keys.md
```diff
@@ -0,0 +1,18 @@
+# Use Case: Possibility to delete keys in ansible-libelektra
+
+## Summary
+
+- **Title:** Possibility to delete keys in ansible-libelektra
+- **Scope:** ansible-libelektra
+- **Level:** User Goal
+- **Actors:** User (usually sysadmint)
```
markus2330, line 8: typo

doc/usecases/record_elektra/UC_ansible-libelektra_remove_keys.md
```diff
@@ -0,0 +1,18 @@
+# Use Case: Possibility to delete keys in ansible-libelektra
```
markus2330, line 1:
```suggestion
# Use Case: Remove keys in ansible-libelektra
```

doc/usecases/record_elektra/UC_ansible-libelektra_remove_keys.md
```diff
@@ -0,0 +1,18 @@
...
+- **Actors:** User (usually sysadmint)
+- **Brief:** It should be possible to remove keys from the KDB using ansible-libelektra
+
+## Scenarios
+
+- **Precondition:** User specifies which key to delete in the playbook
+- **Main success scenario:** The key exists on the target machine and will be deleted
+- **Alternative scenario:** The key does not exists, nothing will be deleted
+- **Error scenario:** -
+- **Postcondition:** The specified key does not exist on the target machine
```
markus2330, line 17: :+1:

doc/usecases/record_elektra/UC_ansible-libelektra_report_conflicts.md
```diff
@@ -0,0 +1,18 @@
+# Use Case: Report merge conflicts in ansible-libelektra
```
markus2330, line 1: One merge use case should be enough? This is the Error scenario of the merge use case?

doc/usecases/record_elektra/UC_ansible-libelektra_start_recording.md
```diff
@@ -0,0 +1,19 @@
...
+- **Actors:** User (usually sysadmin)
+- **Brief:** After a new elektra configuration has been rolled out via Ansible, automatically enable session recording
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** User deploys configuration via Ansible.
+- **Alternative scenario:** -
+- **Error scenario:** -
+- **Postcondition:** Session recording is enabled
```
markus2330, line 17:
```suggestion
- **Postcondition:** All locally changed keys get tracked.
```

doc/usecases/record_elektra/UC_ansible_export.md
```diff
@@ -0,0 +1,26 @@
...
+
+- **Title:** Exporting configuration as Ansible playbook
+- **Scope:** Configuration
+- **Level:** User goal
+- **Actors:** User (usually a sysadmin)
+- **Brief:** User exports certain configuration as Ansible playbook
+
+## Scenarios
+
+- **Precondition:** Some kind of configuration has to exist
```
markus2330, line 13: Isn't there some valid "empty" playbook?

doc/usecases/record_elektra/UC_ansible_export.md
```diff
@@ -0,0 +1,26 @@
...
+  
+- **Main success scenario:** 
+  - User specifies which configuration to include in playbook
+  - It must be possible to select only configuration that has changed (was recorded in session recording)
+  - Ansible playbook is generated
+  
+- **Alternative scenario:** When no configuration exists, an empty playbook is generated
+- **Error scenario:** -
+- **Postcondition:** 
+  - Ansible playbook exists
```
markus2330, line 23: This UC does not catch the situation that a playbook is updated (already existed before and gets updated with what was recorded).

doc/usecases/record_elektra/UC_ansible_specification.md
```diff
@@ -0,0 +1,27 @@
...
+- **Level:** User Goal
+- **Actors:** User (usually a sysadmin)
+- **Brief:** Based on [Ansible Export](UC_ansible_export.md), it should be possible to describe how keys in Elektra are transformed into tasks in an Ansible playbook.
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** 
+  - The user defines (e.g. using meta keys) how to translate keys in a mountpoint into a task in an Ansible playbook
+  - The user exports data of that mountpoints into an Ansible playbook
```
markus2330, line 16: Is this export also connected to UC_ansible_export.md?

doc/usecases/record_elektra/UC_ansible_specification.md
```diff
@@ -0,0 +1,27 @@
...
+- **Scope:** Configuration
+- **Level:** User Goal
+- **Actors:** User (usually a sysadmin)
+- **Brief:** Based on [Ansible Export](UC_ansible_export.md), it should be possible to describe how keys in Elektra are transformed into tasks in an Ansible playbook.
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** 
+  - The user defines (e.g. using meta keys) how to translate keys in a mountpoint into a task in an Ansible playbook
```
markus2330, line 15: Not clear to me.

doc/usecases/record_elektra/UC_ansible_specification.md
```diff
@@ -0,0 +1,27 @@
...
+  - The user exports data of that mountpoints into an Ansible playbook
+  - The correct task has been created
+  
+- **Alternative scenario:** 
+  - The user does not explicitly specify how to map keys to an Ansible task
```
markus2330, line 20: What happens?

doc/usecases/record_elektra/UC_cmerge_conflict_keys.md
```diff
@@ -0,0 +1,19 @@
+# Use Case: Get information which keys are conflicting in 3-way-merge
```
markus2330, line 1: Probably too low-level to be needed as UC. It is simply missing functionality, needed for the merge UC.

doc/usecases/record_elektra/UC_change_notifications.md
markus2330: I agree, we need this narrowed down, maybe even several UC (logging, change notification, ...)

markus2330 (commented):
Thank you for the PR! I am missing the high-level use cases (build server, nextcloud, olimex should be somehow derivable from the UC described here). What is the workflow you plan to do while doing the three UC? At least the "do local changes/updating playbook" UC is missing.

doc/usecases/record_elektra/UC_change_notifications.md
kodebach: > logging, change notification

I don't think that separation would make sense, since "logging" is an application/special case of "change notification". The plugin is notified of changes and just logs them somewhere, instead of doing something else (e.g. broadcasting via dbus)

However, it might make sense to make one UC for "global/hook plugins" in general and a second UC for a "logging plugin". But this may be a too technical distinction, as "global/hook plugins" could just be considered the implementation details of the current UC.

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
atmaxinger: This can be used as a safe guard to ensure the configuration is only applied to the correct machines. Probably won't be used by many, but it could be useful to some. This the feature needed for `kdb record mark` to make any sense.

markus2330: A UC should clearly state for whom, in which situation and why it is useful.

Obviously we cannot know how often features will actually be used. For now there should be at least one convincing example where we ourselves would use it (or at least can imagine using it).

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
atmaxinger: But the template clearly states so...

grafik

doc/usecases/record_elektra/UC_ansible-libelektra_report_conflicts.md
atmaxinger: Depends on how granular these use cases should be. But yes, we can move them into the same use case.

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
kodebach: I think one example would be a dependency between two applications. Let's say apps A and B work together. There could a key in A's config that may be set to tell A to work with B. Now we define a config for B and this config only works, when used with A. Therefore, we need A to configured to work with B.

We could of course just set A's key to the right value and include that in the recording. But IMO that would really be good practice. We'd be reconfiguring A when we apply the config for B. Instead a better solution would be just marking the the key in A's config and thereby recording that it must be set correctly for this recording to be applicable.

---

Another example that isn't covered by this UC, but which might be useful (and is probably more common): Let's say there is a key that can't be set automatically, but must be manually configured by the sysadmin before applying the config. This could for example be a path that is unique to the machine. In this case, we'd need away to record "this key must be set, but we don't care about the value".

However, this special case can probably be better achieved, by adding an additional task to the Ansible playbook.

---

The above are the most reasonable examples I could think of, but there are other more questionable cases. For example, if setting the key in question is not idempotent, e.g. because of some (less than ideal) plugin used for the mountpoint.

---

> **Note** On a technical level, the ability to record such "marked keys" falls directly out of the "record old & new values" format. For a marked key, the old and new value would just be identical. Therefore, what we are really deciding on with this use case is whether or not to expose this feature to the user.

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
markus2330: As you want :wink:

doc/usecases/record_elektra/UC_ansible_export.md
atmaxinger: Yes! we could just create that.

doc/usecases/record_elektra/UC_ansible_export.md
atmaxinger: Wouldn't this be another use case? Merging of playbooks?

doc/usecases/record_elektra/UC_ansible_specification.md
atmaxinger: Yes! The idea of this use case was/is that we could (sometimes in the future) generate "real" Ansible playbooks that work without ansible-libelektra. But I think this is something out of scope for now.

atmaxinger:
@markus2330 I have now created a new use case for the whole "do local changes/updating playbook" --> `UC_interactive_configuration.md`

doc/usecases/record_elektra/UC_interactive_configuration.md
```diff
@@ -0,0 +1,25 @@
+# Use Case: Interactively creating system config
+
+## Summary
+
+- **Title:** Interactively creating system config
+- **Scope:** Configuration
+- **Level:** User Goal
```
markus2330, line 7: This must be somewhere else, otherwise "Level:" is useless.

doc/usecases/record_elektra/UC_interactive_configuration.md
```diff
@@ -0,0 +1,25 @@
...
+- **Actors:** User (usually sysadmin)
```
markus2330, line 8: Be as specific as possible. Who do we target?

```suggestion
- **Actors:** System administrator
```

doc/usecases/record_elektra/UC_interactive_configuration.md
```diff
@@ -0,0 +1,25 @@
...
+- **Brief:** In order to interactively create a workable system configuration, sometimes multiple iterations are required. 
+It should be possible to modify the configuration on a single host, and then apply the changes onto all other hosts.
```
markus2330, line 10: Why limit to modifications on a single host? As long the changes can somehow be merged...

```suggestion
It should be possible to modify the configuration directly on hosts, and then apply the changes onto all other hosts.
```

doc/usecases/record_elektra/UC_interactive_configuration.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** 
+  - User connects to the host and changes settings in Elektra
+  - User exports a playbook that contains either
+    - only the changed keys
+    - the original playbook with the current changes applied
+- **Alternative scenario:** If the host was not configured using ansible-elektra, it is only possible to export a playbook with the changed keys
```
markus2330, line 21:
```suggestion
- **Alternative scenario:** If the host has no applied playbook, it is only possible to export a playbook with the changed keys
```

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
markus2330: Yes, I agree: limited/broken plugins, in particular read-only plugins, are a situation I could imagine it makes sense. There are some cases where plugins must be read-only by design (hardware-information, read-only file systems etc.) so in these cases assertions are the only way.

Btw. terminology-wise we used "assert" in #4399. This alone would make the UC much clearer.

markus2330 (commented):
This use case was very much needed, good work! What I am still missing is a UC that somehow would make multiple sessions necessary. From the current UCs actually a single global session would be enough.

doc/usecases/record_elektra/UC_interactive_configuration.md
atmaxinger: The problem is, if you change the configuration on multiple systems at the same time, we'd have to manually merge the playbooks - as in the user would have to copy all the playbooks from the machines and let a tool merge them. 

If we limit this to a single host at a time, the ansible export plugin could do this automatically at this host.

doc/usecases/record_elektra/UC_interactive_configuration.md
atmaxinger: Hmm we can't do it with every playbook. It must be a playbook that was specifically exported by our plugin to do this, as we need to copy the playbook to the remote hosts for the merging to work.

markus2330: I did not intend to change the meaning in this way. You can simply ignore the suggestion.

doc/usecases/record_elektra/UC_ansible-libelektra_marked_keys.md
atmaxinger: Another example would be to check if a certain Elektra plugin is installed, e.g. the `check` plugin.

doc/usecases/record_elektra/UC_change_notifications.md
kodebach: >  log read-only accesses to the kdb, maybe for audit purposes

Yeah, I think this would be very far out of scope. Logging read accesses a totally different can of worms.

What should be possible with the basic logging/notification hook is logging the parent key that was used in every `kdbGet`. Any logging/notification plugin would be called with the parent key anyway.

doc/usecases/record_elektra/UC_ansible-libelektra_merge_strategies.md
```diff
@@ -0,0 +1,18 @@
...
+- **Scope:** ansible-libelektra
+- **Level:** User goal
+- **Actors:** User (usually sysadmin)
+- **Brief:** It should be possible to specify the merge strategy used, e.g. OURS, THEIRS, ABORT. ansible-libelektra should then use this strategy when merging.
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** If the user has specified a strategy, this strategy is applied on merge conflicts.
+- **Alternative scenario:** If no strategy is specified, a sane (and documented) default strategy should be used.
```
markus2330, line 15: This is vague. It would be better if you decide.

doc/usecases/record_elektra/UC_ansible-libelektra_merge_strategies.md
```diff
@@ -0,0 +1,18 @@
+# Use Case: Select merge strategy in ansible-libelektra
+
+## Summary
+
+- **Title:** Select merge-strategy in ansible-libelektra
+- **Scope:** ansible-libelektra
+- **Level:** User goal
+- **Actors:** User (usually sysadmin)
+- **Brief:** It should be possible to specify the merge strategy used, e.g. OURS, THEIRS, ABORT. ansible-libelektra should then use this strategy when merging.
```
markus2330, line 9: Please list at least the currently known strategies we want to implement. Ideally the use case gives argumentation when to use what.

doc/usecases/record_elektra/UC_ansible-libelektra_start_recording.md
```diff
@@ -0,0 +1,18 @@
+# Use Case: Start session recording after Ansible run
+
+## Summary
+
+- **Title:** Start session recording after Ansible run
+- **Scope:** Recording
+- **Level:** User Goal
+- **Actors:** User (usually sysadmin)
+- **Brief:** After a new elektra configuration has been rolled out via Ansible, automatically enable session recording
```
markus2330, line 9:
```suggestion
- **Brief:** After a new elektra configuration has been rolled out via Ansible, automatically enable or reset session recording for the appropriate namespaces
```

doc/usecases/record_elektra/UC_ansible-libelektra_start_recording.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** User deploys configuration via Ansible.
```
markus2330, line 14: Sometimes there are dots, sometimes not?

```suggestion
- **Main success scenario:** User deploys configuration via Ansible
```

doc/usecases/record_elektra/UC_change_notifications.md
```diff
@@ -0,0 +1,23 @@
...
+- **Brief:** Some plugins are only executed when something changed in the KDB and as an input receive the keys that changed in their old and new state.
+
+## Scenarios
+
+- **Precondition:**
+  - Something in the KDB changed
+  - A plugin listening to such a hook is enabled
+- **Main success scenario:**
+  - After a change in the KDB the plugin(s) listening on the hooks are notified
+  - The plugins receive both the old and new state of the keys
```
markus2330, line 18: Isn't this implementation detail? Actually they want to know the difference? (Duplicate with Brief)

doc/usecases/record_elektra/UC_record_changes.md
```diff
@@ -0,0 +1,25 @@
...
+- **Brief:** Any changes that are made in the key database are recorded.
+
+## Scenarios
+
+- **Precondition:**
+  - Recording is enabled
+  - An active recording session exists
+- **Main success scenario:**
+  - User makes a change to the key database, e.g. add key, delete key, modify key, modify meta
+  - The old & new values for every chagned key and metakey are recorded.
```
markus2330, line 18:
```suggestion
  - The old & new values for every changed key and metakey are recorded.
```

doc/usecases/record_elektra/UC_record_changes.md
```diff
@@ -0,0 +1,25 @@
...
+- **Alternative scenario:** -
+- **Error scenario:** We log that the changes could not be recorded.
```
markus2330, line 20: I think we should make `kdb set` warn about this.

doc/usecases/record_elektra/README.md
```diff
@@ -0,0 +1,21 @@
+# record-elektra Use Cases
+
+The contained use cases are both from a developer side and user side.
+
+The main use case is [interactive configuration](UC_interactive_configuration.md).
+
+## Developer-oriented Use Cases
```
markus2330, line 7: move this section down

markus2330 (changes requested):
It looks much better now! The PR is nevertheless still marked as draft. Please undraft and set "ready to merge" so that we can get this merged eventually.

One UC I would find very interesting is to allow specific overrides https://github.com/ElektraInitiative/ansible-libelektra/issues/11 and nevertheless allow exports (i.e. we would need some way to tell if a change is specific to a host or should be applied to all hosts).

doc/usecases/record_elektra/UC_ansible-libelektra_start_recording.md
atmaxinger: What would 'the appropriate namespaces' be?

doc/usecases/record_elektra/UC_record_changes.md
atmaxinger: Yes, my bad. This is what I meant by 'We log'.

doc/usecases/record_elektra/UC_ansible-libelektra_merge_strategies.md
atmaxinger: It think we should use OURS as default. It's probably the intention in 99% of the use cases to set a new value when using Ansible.

markus2330: I agree!

doc/usecases/record_elektra/UC_ansible-libelektra_start_recording.md
markus2330: Probably all of them which had changes (any recording happened)?

doc/usecases/record_elektra/UC_ansible-libelektra_merge_strategies.md
kodebach: I'm not sure, using "ours" to resolve conflicts by default is a good idea. IMO the default should be to just abort. Otherwise things may already be broken before the user even notices that there was a merge.

Yes, most people will probably decide to use "ours", but it's not a safe default IMO. It's kind of as if the default for an SQL INSERT would be to replace a conflicting primary key instead of aborting.

markus2330: Yes, abort on conflicts is probably a better choice. Then people look at which changes were done by maintainers and manually select what is actually best.

Btw. we discarded the undo UC. It might help with unwanted merges but it is probably not worth the effort to keep previous recordings.

atmaxinger: I'm not even sure anymore if we need the merging for the Ansible Use Case. If I execute an Anisble playbook, my goal is that all the hosts I run them on are in the configured state. After all, an Ansible playbook is supposed to be a single source of truth.

atmaxinger: I mean I'll still implement this feature ... but for the Case Studies this shouldn't be needed.
Also, to be usable we need to supply the `base`. Otherwise, it will always come to conflicts (except keys that never change) if we use `ABORT` as default.

markus2330: As I envisioned it, Ansible runs would set a new BASE. Modifications of Ansible playbooks are OURS, modifications on the machine are THEIRS.

THEIRS can have two reasons:
1. manual interventions as supported by our recording approach
2. updates of software (i.e. new config files in /etc installed by packages)

Because of case 2. this UC is very much needed whenever you will use ansible-libelektra (regardless of if you also use record-libelektra). If 1. is done correctly (all changes exported and added to playbooks) there actually should be no conflict, as OURS would be THEIRS (both different from BASE). If someone forgot exporting and the playbook introduces incompatible changes, the strategy suggested from @kodebach, would loudly communicate that the sys-admins did something wrong, which looks like to be the proper way to do it. (Careful sys-admins would reset the machines to the previous BASE if they played around and nothing useful came out.)

Did this clarify the concerns?

atmaxinger: Yes this did clarify a few concerns. I totally forgot about the updates of software.

The only thing I still don't get is what you mean by 'Ansible runs would set a new BASE'. Where is this BASE stored? Within the playbook?

markus2330: It cannot be the playbook (otherwise you would not be able to reset to base locally) but where exactly is actually implementation detail. I thought that when `kdb record` gets reset, "/elektra/record" (or whatever we use as path for recording) is empty and with every changed key, the old values/metadata of that key get stored in "/elektra/record". So basically the merge from the current keys and (with higher priority) /elektra/record is the base. But let us discuss details later when you actually come to the implementation. For now hooks/global plugins have highest priority :rocket:

atmaxinger: This is actually a good idea - but remember: We don't track mounted configuration files. So if an update changes those directly we are currently out of luck.

I agree, let's get this merged and concentrate on the hooks.

markus2330 (approved):
Looks much better. Some parts need to be revisited later (which merge strategies, details of change notifications) but in general the UC goes to a good direction.

@atmaxinger For this PR: please have a look at the build server problems.

atmaxinger:

atmaxinger:
@markus2330 the BSD builds fail because of the wrong version of `libc`, and the jenkins build is flakey. For example, right now its running into a timeout when importig demo.kdb. Not much I can do about that in this PR.

markus2330:
Please report problems of Jenkins to #160.

@Eiskasten can you help out with the BSD failures?

markus2330:
@atmaxinger please report problems of Jenkins to https://github.com/ElektraInitiative/libelektra/issues/160.

@Eiskasten can you help out with the BSD failures?

atmaxinger:
jenkings build libelektra please

atmaxinger:

markus2330:
Yes, let us merge it. Great work! :fireworks:

Issue #4412, opened 2022-07-15
[new-backend] Discussion: Global Plugins


atmaxinger:
Quotes taken from [doc/decisions/global_plugins.md](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/global_plugins.md)

> Notification does not happen once after final commit, but for every plugin

What exactly does that mean? How should it work?

> Plugin interface should be the same. Many plugins, e.g. dbus, should work as global plugins w/o any change in code (i.e. only changes in contract)

How exactly would this work? How can we have a different contract with hooks specific for each type of plugin when there should not be any change in the code of the plugin? 

> Global plugins might depend on specific applications or specific mount points (it should be possible to enforce global plugins for specific applications).

So we need some filter where we can say "execute this plugin only for these parent keys"

Also, should it be possible that there are multiple active plugins of the same type?
- This might be important if we utilize the logging type for the recording plugin - otherwise you can only either have logs or session recording.

How do we configure which plugins are active if there are more than one of the same type?
- IIRC, there was some talk about symlinks, not sure whether I like this approach.

markus2330:
>  What exactly does that mean? How should it work?

For global plugins: It should work in a way that only if all backends successfully did their job, logging&notification happens.

> How exactly would this work?

As the global plugins have the same interface as other plugins, plugins suitable to be global plugins can be mounted in both ways. Only global plugins, however, can fulfill additional global properties (like doing something only if all plugins succeeded).

> So we need some filter where we can say "execute this plugin only for these parent keys"

This "filter" is exactly what happens with mounted plugins.

> Also, should it be possible that there are multiple active plugins of the same type?

I don't think it is needed. There were longer discussions about it. I do not think it is useful to start the same discussions again here. Difficult decisions should have a documented decision process. So if you (or someone else) think such a feature is useful, please open a PR with a decision template.

> This might be important if we utilize the logging type for the recording plugin - otherwise you can only either have logs or session recording.

This could be simply fixed by having "recording" (collecting ours, base and assert keys) as a different kind of global plugin than "logging". Btw. please think hard about terminology&names. It is very hard to change terminology later, as it goes across the whole docu.

> IIRC, there was some talk about symlinks, not sure whether I like this approach.

If we change it, we should consistently change it also for default storage and resolver. See change-resolver-symlink and change-storage-symlink for how it works in these cases.

[...]

Issue #4452, opened 2022-08-28
[new-backend] Change system:/elektra/mountpoints to singular


kodebach:
> system:/elektra/mountpoints

This should be changed to `system:/elektra/mountpoint` as we have in doc/DESIGN.md "We use singular for all names.".

_Originally posted by @markus2330 in https://github.com/ElektraInitiative/libelektra/issues/4442#issuecomment-1229415692_

atmaxinger:
Also `system:/elektra/modules/<plugin name>/exports`

Issue #4446, opened 2022-08-25
default config and mountpoints


markus2330:
In https://github.com/ElektraInitiative/libelektra/issues/4407#issue-1303786329 @kodebach proposed to make bootstrap storage != default storage. The proposal there, imho, moves too much complexity to the build system. I propose following alternative, reaching the same goal but reducing complexity from the build system.

An Elektra installation should come with following mountpoints by default (names all singular, see #4472 #4452):

- [ ] / (i.e. user: system: dir: spec:, see #4428)
- [ ] system:/info/elektra/constant/#0/current (constants plugin)
- [ ] system:/info/elektra/metadata/#0/current (METADATA.ini)
- [ ] system:/info/elektra/contract/#0/current (CONTRACT.ini)
- [ ] system:/info/elektra/uname/#0/current (uname plugin)
- [ ] system:/info/elektra/desktop/#0/current (desktop plugin)
- [ ] system:/sys/host/#0 (hosts plugin)
- [ ] system:/sys/fstab/#0
- [ ] /sw/elektra/kdb/#0/current (i.e. spec: (via improved specload, see #4032  #4444) user: system: dir:, see #4428)
- [ ] /sys/odbc/#0 (user: and system: /etc/odbc.ini and $HOME/.odbc.ini)
- [ ] probably some other files in /etc we have good support for
- [ ] enable all global plugins
- [ ] configuration for recording

A script should do this mountpoints during installation. A user is free to recall this script later, to get different file formats, file paths or file names for root (e.g. from `kdb mount-default toml default.ini`, `kdb mount-default xerces default.xml` or even `kdb mount-default sql <connection>` so that everything goes to an SQL-database.), only the bootstrap file is more-or-less fixed (only changeable via symlinks or similar).

With these mountpoints by default, it does not matter anymore what the default storage is, as it is unused. We could mount "error" by default (during bootstrap phase, with error message that `kdb mount-default` needs to be executed).

@kodebach @flo91 @hannes99  what do you think?

kodebach:
> it does not matter anymore what the default storage is, as it is unused

Not true, the default storage plugin is still used, if no storage plugin is defined during mounting. However, that is a tooling issue, so the default should be defined there.

> only the bootstrap file is more-or-less fixed

[...]

Issue #4436, opened 2022-08-12
Structure of Documentation


kodebach:
Continuing from https://github.com/ElektraInitiative/libelektra/issues/1164#issuecomment-1213156318

> > Maybe there should just be a quick overview of the required steps at the top.
>
> I think that in most of our problems in the docu additional overviews/links/text won't help (except for, e.g., beginner devs, there we probably really missing something like that). We need to fix vague/misleading sentences.

I think an overview in the form of a table of contents for the entire docs would be immensely helpful. Currently it is very hard to find anything unless, you already know that things exists and have a rough idea where they could be.

Further the current split of docs is also problematic IMO. In the GitHub repo we have:

- `decisions`: For internal reference only
- `dev`: The `README.md` says it is the "developer's manual", but it is entirely unclear what information can be found there and who it is addressed to.
- `help`/`man`: Manpages, inherently (end-)user-facing, but sometimes also contains documentation in the sense of (behavioural) specification
- `tutorials`: Tutorials, all of which are "user"-facing. But what a "user" is differs, sometimes it is for application developers, sometimes it is for plugin developers, sometimes it is even of contributing to Elektra. The only commonality is that all are address to some person doing something for the first time, while consuming some API or system (e.g. build system) provided by Elektra.
- and a random pile of `UPPERCASE.md` files in `doc`

On the website it is even worse, the top-level menu groups things semi-randomly into "Documentation" or "Development" and from this top-level menu it is absolutely unclear which items have further sub-menus.

---

IMO we need to totally restructure this and split the documentation into 3 main groups divided by the target audience:

1. User: Anything that is addressed to a person, who wants to install and then use Elektra. 
2. Developer: Stuff addressed to people using Elektra in applications and/or developing plugins, i.e. the consumers of Elektra's API.
3. Contributor: All the internal information targeted at people actually working on Elektra, i.e. the people that work on Elektra's API and functionality.

With this split some of the existing categories clearly belong only to one group. For example, "decisions" are only relevant for contributors, while manpages should be only relevant to users. Other categories, like tutorials, could be found in different groups depending on their target audience.

Yes, this system may require duplicating information. But this is intentional. For example, the full and detailed specification and explanation/reasoning of how key names work (`KEYNAMES.md`) or how `libelektra-kdb` works internally (some of the new docs in `backend` branch) are mostly only  interesting to contributors.

[...]

PR #4471, opened 2022-09-18
Basic implementation of "gopts" hook

atmaxinger:
This PR aims to provide a basic implementation for the "gopts" hook. In theory it works, but there is still one assert in the test cases that does not, as mentioned in #4412.

This PR should also provide an opportunity to discuss the architecture of hooks in general.

## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -5,7 +5,7 @@ Hooks are central points in the KDB lifecycle, where specialized plugins are cal
 ## Selecting which Plugin will be Used for a Specific Hook
 
 Currently, the names of the plugins are hard-coded.
-This decision was made, because these plugins are meant to fulfil very specific purposes.
+This [decision](../decisions/hooks.md) was made, because these plugins are meant to fulfil very specific purposes.
 A symlink replacing the shared library file of the plugin could be used to change the implementation.
 
 ## Interface of the hooks
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -6,6 +6,8 @@ Hooks are the new mechanism that will replace the old global plugins, [as specif
 ## Selecting which Plugin will be used for a specific hook
 
 Currently, the names of the plugins are hard-coded.
+This decision was made, because these plugins are meant to fulfil very specific purposes.
+A symlink replacing the shared library file of the plugin could be used to change the implementation.
 
 ## Interface of the hooks
 
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -11,9 +11,9 @@ A symlink replacing the shared library file of the plugin could be used to chang
 ## Interface of the hooks
 
 If a plugin should be able to act upon a hook, it must export all the functions that the hook requires.
-These exports are of the form `system:/elektra/modules/<plugin name>/exports/hooks/<hook name>/<hook function>`.
+These exports are of the form `system:/elektra/modules/<plugin name>/exports/hook/<hook name>/<hook function>`.
 
-For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hooks/gopts/get`.
+For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hook/gopts/get`.
 
 Other hooks (e.g. `spec`) might require multiple exported functions.
 
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -1,7 +1,6 @@
 # Hooks
 
 Hooks are central points in the KDB lifecycle, where specialized plugins are called.
-Hooks are the new mechanism that will replace the old global plugins, [as specified in this decision](../decisions/global_plugins.md).
 
 ## Selecting which Plugin will be used for a specific hook
 
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -1,0 +1,24 @@
+# Hooks
+
+Hooks are central points in the KDB lifecycle, where specialized plugins are called. 
+Hooks are the new mechanism that will replace the old global plugins, [as specified in this decision](../decisions/global_plugins.md).
+
+## Selecting which Plugin will be used for a specific hook
+
+Currently, the names of the plugins are hard-coded.
+
+## Interface of the hooks
+
+If a plugin should be able to act upon a hook, it must export all the functions that the hook requires.
+These exports are of the form `system:/elektra/modules/<plugin name>/exports/hooks/<hook name>/<hook function>`.
+
+For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefor has to export `system:/elektra/modules/<plugin name>/exports/hooks/gopts/get`.
+
+Other hooks might require multiple exported functions.
+
+## Lifecycle
+
+1. Hooks are initilized within `kdbOpen` after the contract has been processed. This includes loading the plugins.
+2. The appropriate hooks are called within each `kdbGet` and `kdbSet` call.
+3. Hooks are deinitialized within `kdbClose`. This includes unloading the plugins.
+
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -13,7 +13,7 @@ A symlink replacing the shared library file of the plugin could be used to chang
 If a plugin should be able to act upon a hook, it must export all the functions that the hook requires.
 These exports are of the form `system:/elektra/modules/<plugin name>/exports/hook/<hook name>/<hook function>`.
 
-For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hook/gopts/get`.
+For example, the `gopts` hook only requires the `get` function. A plugin that wants to act upon the `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hook/gopts/get`.
 
 Other hooks (e.g. `spec`) require multiple exported functions.
 
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -2,7 +2,7 @@
 
 Hooks are central points in the KDB lifecycle, where specialized plugins are called.
 
-## Selecting which Plugin will be used for a specific hook
+## Selecting which Plugin will be Used for a Specific Hook
 
 Currently, the names of the plugins are hard-coded.
 This decision was made, because these plugins are meant to fulfil very specific purposes.
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -15,7 +15,7 @@ These exports are of the form `system:/elektra/modules/<plugin name>/exports/hoo
 
 For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hook/gopts/get`.
 
-Other hooks (e.g. `spec`) might require multiple exported functions.
+Other hooks (e.g. `spec`) require multiple exported functions.
 
 ## Lifecycle
 
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -1,6 +1,6 @@
 # Hooks
 
-Hooks are central points in the KDB lifecycle, where specialized plugins are called. 
+Hooks are central points in the KDB lifecycle, where specialized plugins are called.
 Hooks are the new mechanism that will replace the old global plugins, [as specified in this decision](../decisions/global_plugins.md).
 
 ## Selecting which Plugin will be used for a specific hook
@@ -21,4 +21,3 @@ Other hooks might require multiple exported functions.
 1. Hooks are initilized within `kdbOpen` after the contract has been processed. This includes loading the plugins.
 2. The appropriate hooks are called within each `kdbGet` and `kdbSet` call.
 3. Hooks are deinitialized within `kdbClose`. This includes unloading the plugins.
-
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -16,7 +16,7 @@ These exports are of the form `system:/elektra/modules/<plugin name>/exports/hoo
 
 For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hooks/gopts/get`.
 
-Other hooks might require multiple exported functions.
+Other hooks (e.g. `spec`) might require multiple exported functions.
 
 ## Lifecycle
 
--- a/doc/dev/hooks.md
+++ b/doc/dev/hooks.md
@@ -14,7 +14,7 @@ A symlink replacing the shared library file of the plugin could be used to chang
 If a plugin should be able to act upon a hook, it must export all the functions that the hook requires.
 These exports are of the form `system:/elektra/modules/<plugin name>/exports/hooks/<hook name>/<hook function>`.
 
-For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefor has to export `system:/elektra/modules/<plugin name>/exports/hooks/gopts/get`.
+For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hooks/gopts/get`.
 
 Other hooks might require multiple exported functions.
 
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -580,7 +580,7 @@ int mountVersion (KDB * kdb, Key * errorKey);
 int mountGlobals (KDB * kdb, KeySet * keys, KeySet * modules, Key * errorKey);
 int mountBackend (KDB * kdb, const Key * mountpoint, Plugin * backend);
 int initHooks (KDB * kdb, KeySet * config, KeySet * modules, Key * errorKey);
-void freeHooks(KDB * kdb, Key * errorKey);
+void freeHooks (KDB * kdb, Key * errorKey);
 
 const Key * mountGetMountpoint (KDB * handle, Key * where);
 Plugin * mountGetBackend (KDB * handle, Key * key);
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -99,6 +99,8 @@ typedef int (*kdbSetPtr) (Plugin * handle, KeySet * returned, Key * parentKey);
 typedef int (*kdbErrorPtr) (Plugin * handle, KeySet * returned, Key * parentKey);
 typedef int (*kdbCommitPtr) (Plugin * handle, KeySet * returned, Key * parentKey);
 
+typedef int (*kdbHookGoptsGetPtr) (Plugin * handle, KeySet * returned, Key * parentKey);
+
 typedef Plugin * (*OpenMapper) (const char *, const char *, KeySet *);
 typedef int (*CloseMapper) (Plugin *);
 
@@ -319,6 +321,7 @@ struct _KeySet
 #endif
 };
 
+typedef struct _Hooks Hooks;
 
 /**
  * The access point to the key database.
@@ -362,7 +365,7 @@ struct _KDB
 			up their parts of the global keyset, which they do not need any more.*/
 
 	Plugin * globalPlugins[NR_GLOBAL_POSITIONS][NR_GLOBAL_SUBPOSITIONS];
-
+	Hooks * hooks;
 	KeySet * backends;
 };
 
@@ -409,6 +412,18 @@ struct _Plugin
 			up their parts of the global keyset, which they do not need any more.*/
 
 	KeySet * modules; /*!< A list of all currently loaded modules.*/
+};
+
+struct _HookPluginGopts
+{
+	struct _Plugin* plugin;
+	kdbHookGoptsGetPtr kdbHookGoptsGet;
+};
+
+struct _Hooks
+{
+	struct _HookPluginGopts* gopts;
+	bool goptsEnabled;
 };
 
 // FIXME (kodebach): document
@@ -564,6 +579,8 @@ int mountModules (KDB * kdb, KeySet * modules, Key * errorKey);
 int mountVersion (KDB * kdb, Key * errorKey);
 int mountGlobals (KDB * kdb, KeySet * keys, KeySet * modules, Key * errorKey);
 int mountBackend (KDB * kdb, const Key * mountpoint, Plugin * backend);
+int initHooks (KDB * kdb, KeySet * config, KeySet * modules, Key * errorKey);
+void freeHooks(KDB * kdb, Key * errorKey);
 
 const Key * mountGetMountpoint (KDB * handle, Key * where);
 Plugin * mountGetBackend (KDB * handle, Key * key);
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -365,8 +365,16 @@ struct _KDB
 			up their parts of the global keyset, which they do not need any more.*/
 
 	Plugin * globalPlugins[NR_GLOBAL_POSITIONS][NR_GLOBAL_SUBPOSITIONS];
-	Hooks * hooks;
 	KeySet * backends;
+
+	struct
+	{
+		struct
+		{
+			struct _Plugin* plugin;
+			kdbHookGoptsGetPtr kdbHookGoptsGet;
+		} gopts;
+	} hooks;
 };
 
 /**
@@ -412,18 +420,6 @@ struct _Plugin
 			up their parts of the global keyset, which they do not need any more.*/
 
 	KeySet * modules; /*!< A list of all currently loaded modules.*/
-};
-
-struct _HookPluginGopts
-{
-	struct _Plugin* plugin;
-	kdbHookGoptsGetPtr kdbHookGoptsGet;
-};
-
-struct _Hooks
-{
-	struct _HookPluginGopts* gopts;
-	bool goptsEnabled;
 };
 
 // FIXME (kodebach): document
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -575,7 +575,7 @@ int mountModules (KDB * kdb, KeySet * modules, Key * errorKey);
 int mountVersion (KDB * kdb, Key * errorKey);
 int mountGlobals (KDB * kdb, KeySet * keys, KeySet * modules, Key * errorKey);
 int mountBackend (KDB * kdb, const Key * mountpoint, Plugin * backend);
-int initHooks (KDB * kdb, KeySet * config, KeySet * modules, const KeySet * contract, Key * errorKey);
+int initHooks (KDB * kdb, KeySet * modules, const KeySet * contract, Key * errorKey);
 void freeHooks (KDB * kdb, Key * errorKey);
 
 const Key * mountGetMountpoint (KDB * handle, Key * where);
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -575,7 +575,7 @@ int mountModules (KDB * kdb, KeySet * modules, Key * errorKey);
 int mountVersion (KDB * kdb, Key * errorKey);
 int mountGlobals (KDB * kdb, KeySet * keys, KeySet * modules, Key * errorKey);
 int mountBackend (KDB * kdb, const Key * mountpoint, Plugin * backend);
-int initHooks (KDB * kdb, KeySet * modules, const KeySet * contract, Key * errorKey);
+int initHooks (KDB * kdb, const KeySet * config, KeySet * modules, const KeySet * contract, Key * errorKey);
 void freeHooks (KDB * kdb, Key * errorKey);
 
 const Key * mountGetMountpoint (KDB * handle, Key * where);
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -575,7 +575,7 @@ int mountModules (KDB * kdb, KeySet * modules, Key * errorKey);
 int mountVersion (KDB * kdb, Key * errorKey);
 int mountGlobals (KDB * kdb, KeySet * keys, KeySet * modules, Key * errorKey);
 int mountBackend (KDB * kdb, const Key * mountpoint, Plugin * backend);
-int initHooks (KDB * kdb, KeySet * config, KeySet * modules, Key * errorKey);
+int initHooks (KDB * kdb, KeySet * config, KeySet * modules, const KeySet * contract, Key * errorKey);
 void freeHooks (KDB * kdb, Key * errorKey);
 
 const Key * mountGetMountpoint (KDB * handle, Key * where);
--- a/src/libs/elektra/hooks.c
+++ b/src/libs/elektra/hooks.c
@@ -10,34 +10,23 @@
 
 void freeHooks (KDB * kdb, Key * errorKey)
 {
-	if (kdb->hooks == NULL)
+	if (kdb->hooks.gopts.plugin != NULL)
 	{
-		return;
+		elektraPluginClose (kdb->hooks.gopts.plugin, errorKey);
 	}
-
-	if (kdb->hooks->gopts != NULL)
-	{
-		elektraPluginClose (kdb->hooks->gopts->plugin, errorKey);
-		free (kdb->hooks->gopts);
-		kdb->hooks->gopts = NULL;
-	}
-
-	free (kdb->hooks);
-	kdb->hooks = NULL;
 }
 
-static struct _HookPluginGopts * initHooksGopts (Plugin * plugin)
+static int initHooksGopts (KDB * kdb, Plugin * plugin)
 {
 	if (!plugin)
 	{
-		return NULL;
+		return -1;
 	}
 
-	struct _HookPluginGopts * hook = elektraCalloc (sizeof (struct _HookPluginGopts));
-	hook->plugin = plugin;
+	kdb->hooks.gopts.plugin = plugin;
+	kdb->hooks.gopts.kdbHookGoptsGet = (kdbHookGoptsGetPtr) elektraPluginGetFunction (plugin, "hooks/gopts/get");
 
-	hook->kdbHookGoptsGet = (kdbHookGoptsGetPtr) elektraPluginGetFunction (plugin, "hooks/gopts/get");
-	return hook;
+	return 0;
 }
 
 static Plugin * loadPlugin (const char * pluginName, KeySet * config, KeySet * modules, Key * errorKey)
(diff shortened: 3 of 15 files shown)


src/libs/elektra/hooks.c
```diff
@@ -0,0 +1,96 @@
...
+{
+	if(!plugin)
+	{
+		return NULL;
+	}
+
+	struct _HookPluginGopts * hook = elektraCalloc (sizeof(struct _HookPluginGopts));
+	hook->plugin = plugin;
+
+	hook->kdbHookGoptsGet = (kdbHookGoptsGetPtr) elektraPluginGetFunction(plugin, "hooks/gopts/get");
```
atmaxinger, line 39: `elektraPluginGetFunction()` has return type `size_t`. Is there a particual reason why it doesn't return `void *`? This would save us from casting every time we use this function.

kodebach: AFAIK according to the C standard casting function pointers to non-function pointers like `void *` is undefined behaviour (presumably because in theory the Hardware could be Harvard architecture instead of Von-Neumann architecture). Compilers like GCC and Clang allow it and even treat `void (*) (void)` as a special type that is compatible with any function pointer. However all that is non-standard. I assume that's why `elektraPluginGetFunction` returns `size_t`.

At the same time, `dlsym` is defined as `void *dlsym(void *handle, const char *symbol)`, so we may as well use `void *` too. But you'd almost definitely still need a cast, otherwise the man page for `dlsym` wouldn't contain this wonderful snippet:

```c
   /* Writing: cosine = (double (*)(double)) dlsym(handle, "cos");
       would seem more natural, but the C99 standard leaves
       casting from "void *" to a function pointer undefined.
       The assignment used below is the POSIX.1-2003 (Technical
       Corrigendum 1) workaround; see the Rationale for the
       POSIX specification of dlsym(). */

   *(void **) (&cosine) = dlsym(handle, "cos");
```

src/libs/elektra/kdb.c
```diff
@@ -1874,6 +1889,23 @@ int kdbGet (KDB * handle, KeySet * ks, Key * parentKey)
 		clear_bit (parentKey->flags, KEY_LOCK_NAME | KEY_LOCK_VALUE);
 	*/
 
+	// Step 13: run gopts (if enabled)
+	keyCopy (parentKey, initialParent, KEY_CP_NAME);
```
kodebach, line 1893: As stated in the other comment, you should do:

```suggestion
	keyCopy (parentKey, initialParent, KEY_CP_NAME);
	keySetNamespace (parentKey, KEY_NS_CASCADING);
```

and then before step 15 and you may need

```c
keySetNamespace (parentKey, keyGetNamespace (initialParent));
```

src/libs/elektra/kdb.c
```diff
@@ -1006,6 +1007,14 @@ KDB * kdbOpen (const KeySet * contract, Key * errorKey)
 		goto error;
 	}
 
+	// TOOD (atmaxinger): improve
+	if (initHooks (handle, ksDup (elektraKs), handle->modules, errorKey) == -1)
+	{
+		ELEKTRA_SET_INSTALLATION_ERROR (errorKey, "Mounting hooks failed. Please see warning of concrete plugin");
+		ksDel (elektraKs);
+		goto error;
+	}
```
kodebach, line 1016: You should probably use the same trick as in `runSetPhase` to make sure the plugin can only create warnings (*), otherwise this `ELEKTRA_SET_INSTALLATION_ERROR` may actually result in a warning.

https://github.com/ElektraInitiative/libelektra/blob/6dc6bb8305ff8d9581f2ffa518d9120045f21eae/src/libs/elektra/kdb.c#L2058-L2063

https://github.com/ElektraInitiative/libelektra/blob/6dc6bb8305ff8d9581f2ffa518d9120045f21eae/src/libs/elektra/kdb.c#L2158-L2164

You may need to modify the `if` conditions a bit. The idea is simply that setting any value for `meta:/error` will block the creation of errors.

---

(*) In case you don't know, the function behind the `ELEKTRA_SET_*_ERROR` macros will create a warning instead of an error, if there is already an error. That's because according to our design, a `Key` may only contain a single error.

src/libs/elektra/kdb.c
```diff
@@ -378,6 +378,7 @@ static KDB * kdbNew (Key * errorKey)
 	handle->global =
 		ksNew (1, keyNew ("system:/elektra/kdb", KEY_BINARY, KEY_SIZE, sizeof (handle), KEY_VALUE, &handle, KEY_END), KS_END);
 	handle->backends = ksNew (0, KS_END);
+	handle->hooks = elektraCalloc (sizeof (struct _Hooks));
```
kodebach, line 381: As stated above, I'd get rid of all the `calloc`s for the hooks, but this one is extra unnecessary. AFAICT this will never be used and will always be freed by `initHooks`.

src/libs/elektra/hooks.c
```diff
@@ -0,0 +1,95 @@
...
+	KeySet * cut = ksCut (ksTemp, cutPoint);
+
+	goptsEnabled = ksGetSize (cut) > 0;
+
+	keyDel (cutPoint);
+	ksDel (ksTemp);
+	ksDel (cut);
+
+	return goptsEnabled;
+}
```
kodebach, line 76: This is a bit of a hack. What you should really do is pass the `KeySet * contract` from `kdbOpen` to this function and just check for some key that is added by `elektraGOptsContract`.



You could just check whether `system:/elektra/contract/mountglobal/gopts` is there:

```suggestion
static bool isGoptsEnabledByContract (KeySet * contract)
{
	return ksLookupByName (contract, "system:/elektra/contract/mountglobal/gopts", 0) != NULL;
}
```

The `elektraGOptsContract` will always add this key and the [docs for the contract](https://www.libelektra.org/devdocu/kdb-contracts#mounting-global-plugins) also explain that.

In the long it's probably better (not least because of the `mountglobal` name) to also modify `elektraGOptsContract`. In the new hooks system the generic "mount plugin X with config Y" API doesn't really make sense. However, I don't think we should do this now. Instead I'd wait until we have also updated `spec`, `internalnotification`, etc. and can fully remove the global plugin system.

src/libs/elektra/kdb.c
```diff
@@ -1006,6 +1007,14 @@ KDB * kdbOpen (const KeySet * contract, Key * errorKey)
 		goto error;
 	}
 
+	// TOOD (atmaxinger): improve
+	if (initHooks (handle, ksDup (elektraKs), handle->modules, errorKey) == -1)
```
kodebach, line 1011: `elektraKs` is not the correct parameter for `KeySet * config`. Since `config` is passed right through to `elektraPluginOpen`, it is supposed to be the config for the plugin you're loading, i.e. needs to be different for every plugin.

However, AFAIK none of current global plugins require any config, so you can just always pass `ksNew(0, KS_END)` to `elektraPluginOpen`.

kodebach, line 1011: `initHooks` only has a single `return 0`, this `if` is always false

src/libs/elektra/hooks.c
```diff
@@ -0,0 +1,95 @@
...
+{
+	if (!plugin)
+	{
+		return NULL;
+	}
+
+	struct _HookPluginGopts * hook = elektraCalloc (sizeof (struct _HookPluginGopts));
+	hook->plugin = plugin;
+
+	hook->kdbHookGoptsGet = (kdbHookGoptsGetPtr) elektraPluginGetFunction (plugin, "hooks/gopts/get");
```
kodebach, line 39: You should also check that this function exists (`!= NULL`) and if not produce an error/warning.

kodebach (changes requested):
The general idea looks great (it's pretty much what I had in mind when I started the `new-backend` branch). However, the way you changed `struct _KDB` requires quite a lot of allocations. I'm pretty sure we can avoid those. There are also some other minor issues.

src/libs/elektra/kdb.c
atmaxinger: Ha that one is actually required currently! When bootstrapping KDB `initHooks` isnt called. 🤷‍♂️. But I will inline the struct, so it will no longer be necessary.

src/libs/elektra/kdb.c
atmaxinger: For now ;)

src/include/kdbprivate.h
```diff
@@ -362,7 +365,7 @@ struct _KDB
 			up their parts of the global keyset, which they do not need any more.*/
 
 	Plugin * globalPlugins[NR_GLOBAL_POSITIONS][NR_GLOBAL_SUBPOSITIONS];
-
+	Hooks * hooks;
```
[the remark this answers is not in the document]
atmaxinger: The idea behind the `goptsEnabled` flag was that we could enable and disable hooks at runtime. But now that I think about it I don't think there is any mechanismn to do it currently.

src/libs/elektra/kdb.c
atmaxinger: it will now return -1 if any hooks fail to initialized

src/libs/elektra/plugin.c
```diff
@@ -167,7 +167,7 @@ size_t elektraPluginGetFunction (Plugin * plugin, const char * name)
 	plugin->kdbGet (plugin, exports, pk);
 	ksRewind (exports);
 	keyAddBaseName (pk, "exports");
-	keyAddBaseName (pk, name);
+	keyAddName (pk, name); // we may not want to escape for hooks exports
```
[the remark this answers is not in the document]
atmaxinger: Hmm I don't think just blocking names starting with `.` is enough. I could still pass `blahblah/../../..`. I think we should check whether the sequence `..` is contained in the finished key name.

src/libs/elektra/hooks.c
atmaxinger: Yeah let's just leave it with size_t :D -- or we maybe could create our own macro, so that the semantics of the return value are clearer. Anyway, that's not a topic for this PR.

´#define ELECTRA_FPTR size_t`

src/libs/elektra/plugin.c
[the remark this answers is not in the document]
markus2330: What is the problem with the escaping keyAddBaseName does? What is the benefit of the whole change?

"we **may** not" does not sound very convincing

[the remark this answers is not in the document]
atmaxinger: Previously, this method only worked for names that were of the format:
```
system:/elektra/modules/<module name>/exports/<function name>
```

However, hooks exports have a new format:
```
system:/elektra/modules/<module name>/exports/hooks/<hook name>/<function name>
```

So if we want to load a hook function, we have to pass `hooks/<hook name>/<function name>` as an argument to this function. If we use `keyAddBaseName`, the composed key will be 
`
system:/elektra/modules/<module name>/exports/hooks\/<hook name>\/<function name>
`
which is obviously not the same.

src/libs/elektra/kdb.c
markus2330: `kdb cache` already has support for disabling a global plugin, this feature should be available (in the same way) for all global plugins.

Important is that the user can always fix the problem somehow (e.g. disable the broken plugin). If we already fail in kdbOpen, then e.g. `kdb cache disable` probably will also fail, as it tries to write to KDB (and obviously needs to open it first).

So I think broken global plugins can only lead to a warning, not to an error.

atmaxinger: Yes, broken plugins only lead to warnings, this is implemented right now.

> kdb cache already has support for disabling a global plugin, this feature should be available (in the same way) for all global plugins.

We do. As I see, `kdb cache enable/disable` sets a key in the contract. This contract is also passed to `initHooks`. We already parse the contract to enable/disable the `gopts` plugin, so this is accounted for in the current architecture.

atmaxinger: > So I think broken global plugins can only lead to a warning, not to an error.

They already do.

> kdb cache already has support for disabling a global plugin, this feature should be available (in the same way) for all global plugins.

Agreed. The way `kdb cache` does it, it sets the key `system:/elektra/cache/enabled` to either `0` or `1.`.  We can just pass the config to `initHooks` and then check there for the respective keys.

One question that comes to my mind: What shall we do if those keys are not present for a plugin? Is it enabled or disabled by default?

markus2330: See doc/decisions/boolean.md a boolean is by default 0 but we can set all hooks to be 1 via a installation script, see #4446.

Btw. we should create a sub-hierarchy system:/elektra/hook/cache/enabled (if we settle to the global plugins called hooks, is this now fixed?)

src/libs/elektra/plugin.c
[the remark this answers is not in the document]
kodebach: > I think we should check whether the sequence `..` is contained in the finished key name.

Actually, it would have to be `/../` anywhere, `../` at the start or `/..` at the end, because things like `..a` and `a..` are fine in theory. One way to check that would be to make a copy of `pk` before `keyAddName` and then to a `keyIsBelow` check afterwards.

However, I think we could just say `name` is not allowed to contain the character `.` at all. Since we are talking about function names here, IMO that limitation is fine.

src/libs/elektra/kdb.c
atmaxinger: > Btw. we should create a sub-hierarchy system:/elektra/hook/cache/enabled (if we settle to the global plugins called hooks, is this now fixed?)

Sure. But do we want `hook` (singular) or `hooks` (plural) for the sub-hierarchy? The same would also apply for the `system:/elektra/modules/<plugin name>/exports/hooks/<hook name>/get`. Currently I use the plural form there to be consistent with `modules` and `exports`.

kodebach: > it will now return -1 if any hooks fail to initialized

What I meant was that there is already a failure case (couldn't load `gopts` plugin), but the function doesn't actually return an error code for that.

> this feature should be available (in the same way) for all global plugins.

I agree and disagree... The flag in the KDB doesn't make sense for all hooks. Especially for `gopts` it isn't the right solution. `gopts` can only be enabled via the `kdbOpen` contract. Therefore having a permanent flag in the KDB makes zero sense. It makes no sense to "enable gopts by default" and if the flag is used to override the contract it will just break applications that expect `gopts` to run.

For `cache` the flag makes sense of course. For `spec` I could see the flag as well, but in this case it is probably more about switching from errors to warnings. I don't really see a case where completely disabling `spec` would help. Finally, there is the notification stuff. AFAIK for `internalnotification` the situation is the same was with `gopts`. You enable it only via contract and therefore a flag in the KDB is useless. For `dbus` and `zeromq` enabling/disabling via the KDB would probably make sense again.

> system:/elektra/hook/cache/enabled

Yes, I'd do that too. But as stated above, I'd tailor the actual config to the individual hook. So for `spec` it might be something like `system:/elektra/hook/spec/loglevel`.

> we can set all hooks to be 1 via a installation script,

We could also switch from `cache/enabled` to `cache/disabled` which automatically switches the default to "cache enabled".

src/libs/elektra/hooks.c
```diff
@@ -0,0 +1,88 @@
...
+		keyDel (openKey);
+		return NULL;
+	}
+
+	return plugin;
+}
+
+static bool isGoptsEnabledByContract (const KeySet * contract)
+{
+	KeySet * dupContract = ksDup (contract); // We need to duplicate because contract is const, and ksLookupByName doesn't take const
```
kodebach, line 67: @markus2330 We should probably provide a variant of `ksLookup` that works with a `const KeySet *`.

AFAICT these are the reasons why we can't use `const`:
1. We may build the hashmap
2. `KDB_O_POP` could remove a key and `KDB_O_CREATE` could create a key
3. We may create a key from `meta:/default`
4. We update the internal cursor

AFAIK cases 2-4 will all be removed soon. IMO in cases like this one, a `const` variant could be useful to avoid the `ksDup`. The variant never builds the hashmap and just falls back to binary search, if the hashmap wasn't built yet.

src/libs/elektra/kdb.c
kodebach: > do we want `hook` (singular) or `hooks` (plural) for the sub-hierarchy?

`doc/DESIGN.md` says we use singular for all names

https://github.com/ElektraInitiative/libelektra/blob/ce6c6984a81037559332640258c7bd3774ce1036/doc/DESIGN.md?plain=1#L194

But it is not done very consistently, especially for keynames. I also think that often the plural feels more natural in keynames (since it refers to a collection of things), but always using singular is probably the easiest rule to have.

markus2330: Yes, let us stick to always singular. If you spot somewhere plural, we can fix it before 1.0. I created  #4472 to collect such issues.

In some cases it feels a bit weird (e.g. /etc/host) but I think it is only a matter of getting used to it.

markus2330: About enabling plugins: if it does not make sense to enable a plugin, we don't need to provide a flag (and the plugin will always be disabled except if enabled via contract). gopts is the only one, isn't it?

kodebach: > gopts is the only one, isn't it?

As stated above, I think `internalnotification` also needs to be enabled via contract, because it needs a function pointer to the callback(s).

atmaxinger: > Yes, let us stick to always singular. 

Ok, I have updated the newly added key in the exports.

src/libs/elektra/hooks.c
kodebach: From the C Standard section 6.7.3:

> If an attempt is made to modify an object defined with a const-qualified type through use of an
lvalue with non-const-qualified type, the behavior is undefined

Seems pretty clear to me.

Since this is part of our public API we should really do things properly and not go against the standard without good reason. Here we don't know where the `KeySet` originally comes from, so casting a `const KeySet *` to a `KeySet *` is definitely not safe.

src/libs/elektra/kdb.c
```diff
@@ -999,6 +999,29 @@ KDB * kdbOpen (const KeySet * contract, Key * errorKey)
 		goto error;
 	}
 
+	// TODO (atmaxinger): improve
```
markus2330, line 1002: I like the idea of blocking errors this way but I wouldn't call it "dummy" value. Please document the blocking in doc/METADATA.ini. Maybe it should even be its own error category, which is not allowed to be propagated to the user.

markus2330 (commented):
Beautiful coding style!

doc/dev/hooks.md
```diff
@@ -0,0 +1,24 @@
...
+## Selecting which Plugin will be used for a specific hook
+
+Currently, the names of the plugins are hard-coded.
+This decision was made, because these plugins are meant to fulfil very specific purposes.
+A symlink replacing the shared library file of the plugin could be used to change the implementation.
+
+## Interface of the hooks
+
+If a plugin should be able to act upon a hook, it must export all the functions that the hook requires.
+These exports are of the form `system:/elektra/modules/<plugin name>/exports/hooks/<hook name>/<hook function>`.
```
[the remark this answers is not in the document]
atmaxinger: I'd rather not do this in this PR. This would require changing all plugins which is not in the scope for this PR.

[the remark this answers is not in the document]
markus2330: Ok, please create an issue for what is still plural. (Or simply add it to #4452.)

doc/dev/hooks.md
```diff
@@ -0,0 +1,24 @@
...
+
+For example, the `gopts` hook only requires the `get` function. A plugin that wants to act as a `gopts` hook therefore has to export `system:/elektra/modules/<plugin name>/exports/hooks/gopts/get`.
+
+Other hooks (e.g. `spec`) might require multiple exported functions.
```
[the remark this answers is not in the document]
kodebach: But we do know. `spec` is called during `kdbGet` and `kdbSet`. That's at least two functions.

```suggestion
Other hooks (e.g. `spec`) require multiple exported functions.
```

src/libs/elektra/kdb.c
```diff
@@ -999,6 +999,29 @@ KDB * kdbOpen (const KeySet * contract, Key * errorKey)
...
+	bool existingError = keyGetMeta (errorKey, "error") != NULL;
```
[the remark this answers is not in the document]
kodebach: IMO that's more of a nice-to-know. It's good if @atmaxinger knows that in future the recommended way to access meta keys is `ksLookupByName (keyMeta (errorKey), "meta:/error", 0)` (or possibly `keyGetMeta (errorKey, "meta:/error")`), but in the end it really doesn't matter if my `tree-surgeon` scripts have to replace one more call.

So maybe always use `meta:/` in future, but IMO there is really no need to change this now.

src/libs/elektra/kdb.c
kodebach: > Maybe it should even be its own error category, which is not allowed to be propagated to the user.

I'm not so sure about this. Making this an official solution is a bit awkward. Just saying "is not allowed to be" doesn't really ensure anything. If it does end up being propagated to the user, you'll probably get very weird output, since none of the `meta:/error/*` keys is set. For example `kdb` would print (with verbose and debug output)

```
Sorry, module  issued the error :
: 
Mountpoint: 
Configfile: 
At: :

```

> Please document the blocking in doc/METADATA.ini.

Especially this part I'm not sure about.

To go by the terminology from #4436. Blocking an error like this should only be documented in "Contributor documentation" and nowhere else. It is an internal hacky solution, not something that should be encouraged.

If we want an official solution, we should create a "only warnings" error category with all the various `meta:/error/*` values. But if you use that to block errors, you'll need to use a `ksDel (ksCut (...))` to clear the block instead of a simple `keySetMeta`. So I'd think that's the best solution either.

src/libs/elektra/plugin.c
```diff
@@ -153,21 +153,29 @@ int elektraPluginClose (Plugin * handle, Key * errorKey)
...
  * @return        Pointer to function. NULL if function not found or not enough memory available
  */
 size_t elektraPluginGetFunction (Plugin * plugin, const char * name)
 {
 	ELEKTRA_NOT_NULL (plugin);
 	ELEKTRA_NOT_NULL (name);
 
+	if (strstr (name, "..") != NULL)
+	{
+		// The sequence ".." is contained in the key.
```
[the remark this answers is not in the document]
kodebach: This check technically limits what can be used as `name` beyond what is necessary. I think it's fine, but this check now also excludes (technically acceptable) things like `system:/elektra/modules/fooplugin/exports/foo..func`

See also https://github.com/ElektraInitiative/libelektra/pull/4471#discussion_r974115271

doc/dev/hooks.md
```diff
@@ -0,0 +1,24 @@
+# Hooks
+
+Hooks are central points in the KDB lifecycle, where specialized plugins are called.
+
+## Selecting which Plugin will be used for a specific hook
+
+Currently, the names of the plugins are hard-coded.
+This decision was made, because these plugins are meant to fulfil very specific purposes.
```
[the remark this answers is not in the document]
atmaxinger: Is there a decision file specifically for the hardcoding part?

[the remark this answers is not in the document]
markus2330: No, we only have doc/decisions/global_plugins.md

And you need to adapt the decision with what we are actually doing now, e.g. calling it hooks. So the file also needs to be renamed.

src/libs/elektra/kdb.c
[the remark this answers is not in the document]
markus2330: @kodebach if your tool can do these changes automatically it is of course also fine for now. But at some point we need to learn writing correct code (not relying on legacy code being in place), otherwise we will introduce wrong code again...

src/libs/elektra/plugin.c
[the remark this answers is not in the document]
markus2330: Yes, in any case we should document what we allow.

src/libs/elektra/kdb.c
markus2330: > If we want an official solution, we should create a "only warnings" error category with all the various meta:/error/* values.

Yes, good proposal. Probably we give it the number C01311 as it is an Internal error if the error would actually surface to the user?

IMHO such a feature is also relevant for, e.g. plugins, if they want to make sure that they can set an error later of a category they don't know yet.

Within kdb.c I already thought several times it would be good to have a way to say that all errors in a function should be emitted as a warning.

src/libs/elektra/hooks.c
atmaxinger: I have created issue #4474 for this. Let's not discuss this any further in this PR.

src/libs/elektra/hooks.c
```diff
@@ -0,0 +1,88 @@
...
+
+	if (result == 0)
+	{
+		ELEKTRA_ADD_INSTALLATION_WARNINGF (errorKey, "Plugin '%s' does not implement function '%s'", plugin->name, functionName);
+	}
+
+	return result;
+}
+
+static int initHooksGopts (KDB * kdb, Plugin * plugin, Key * errorKey)
```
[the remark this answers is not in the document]
atmaxinger: I've added some comments now, but I am notoriously bad at commenting my own code. If there is anything more that you'd like to know please let me know.

src/libs/elektra/plugin.c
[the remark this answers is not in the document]
atmaxinger: I'd still allow `.`, as we're not talking purely about function names in here, and `.` could be useful as a seperator.

src/libs/elektra/plugin.c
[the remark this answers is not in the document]
atmaxinger: I've added the log statement and it is now documented that we don't allow `..` in the name.

src/libs/elektra/kdb.c
atmaxinger: I've created issue #4475 to disucss this further.

src/libs/elektra/kdb.c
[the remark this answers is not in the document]
atmaxinger: So we keep it for now and let the tool of @kodebach handle it.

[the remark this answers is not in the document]
kodebach: Yes, you can keep this. But for future PRs it's probably better to always use `meta:/` for meta keys. Just to get into the habbit and because there are already cases, where you need to `meta:/`.

src/libs/elektra/hooks.c
kodebach: I noticed there is `ksDeepDup` in `ensureContract` (in `kdb.c`), so independent of the outcome of #4473, we can avoid this `ksDup`. We just need to pull the `ksDeepDup` (and the corresponding `ksDel`) out of `ensureContract` and into `kdbOpen`. Then we can move it before the `initHooks` call and just pass the deep-duped `KeySet`.

src/libs/elektra/kdb.c
```diff
@@ -999,6 +999,29 @@ KDB * kdbOpen (const KeySet * contract, Key * errorKey)
...
+
+	if (!existingError)
+	{
+		// set a dummy value to block errors
+		// any errors that occur will be converted into warnings
+		keySetMeta (errorKey, "error", "blocked");
+	}
```
kodebach, line 1010: I'd actually do this "error blocking" inside of `initHooks`, the `kdbOpen` function doesn't really care about this.

src/libs/elektra/kdb.c
```diff
@@ -999,6 +999,29 @@ KDB * kdbOpen (const KeySet * contract, Key * errorKey)
...
+
+	if (initHooks (handle, elektraKs, handle->modules, contract, errorKey) == -1)
+	{
+		ELEKTRA_SET_INSTALLATION_ERROR (errorKey, "Initializing hooks failed. Please see warning of concrete plugin");
```
kodebach, line 1014: You need to unblock the errors before setting this. Otherwise, you'll not get an error ;) That's actually another reason to move the blocking into `initHooks`. Then it would already be unblocked at this point.

atmaxinger: Good catch! 🎣

src/libs/elektra/hooks.c
```diff
@@ -0,0 +1,159 @@
...
+
+	if ((kdb->hooks.gopts.kdbHookGoptsGet = (kdbHookGoptsGetPtr) getFunction (plugin, "hook/gopts/get", errorKey)) == NULL)
+	{
+		return -1;
+	}
+
+	return 0;
+}
+
+static KeySet * getPluginConfigFromContract (const char * pluginName, const KeySet * contract)
```
kodebach, line 54: This function is complicated enough that it should probably have some documentation comment. Something simple like

```suggestion
/**
 * Extracts the config for a single plugin from the contract
 *
 * The config must be below system:/elektra/contract/mountglobal/<pluginName> and will be moved to below user:/
 *
 * @param pluginName The name of the plugin for which the config will be extracted
 * @param contract The contract from which the config is extracted
 */
static KeySet * getPluginConfigFromContract (const char * pluginName, const KeySet * contract)
```

atmaxinger: Yes, but on that topic - I'd like to create some tests for this function. Do you know of a way that I can use it within a test case without also exposing it through the binary? Currently it's static so no entry in the function table is created for it. 
I just don't want anyone outside of Elektra to call it.

I mean I could `#include` the .c file in the unit test ...

src/libs/elektra/kdb.c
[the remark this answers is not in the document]
atmaxinger: Do we have documentation for `keyMeta`? Because the API Doc still tells me that I should use `keySetMeta` and `keyGetMeta`.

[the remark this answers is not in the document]
kodebach: Beyond the API docs, I don't think `keyMeta` is documented yet. The docs are definitely outdated. `keyGetMeta`/`keySetMeta` might be removed (uncertain), but `keyMeta` definitely will be the main API for metadata in future, since you can already do everything necessary with it.

> Or is it just that I have to use the `meta:/` prefix for all keys I use with those functions?

When you use `keyMeta` then you definitely have to provide the `meta:/` namespace yourself. The fact that `keyGetMeta` and `keySetMeta` work without the `meta:/` prefix is a backwards compatibility thing. Nobody has yet wanted to update all uses, and my semi-automatic tool isn't ready yet.

The actual meta keys always use the `meta:/` prefix, because otherwise it wouldn't be a valid keyname. For example:

```c
printf ("%s\n", keyName (keyGetMeta (keyNew ("/foo", KEY_META, "type", "string", KEY_END), "type")));
// prints meta:/type
```

What you should do to get in the habit of things is write code with `meta:/` instead of without:

```c
Key * k = keyNew ("/foo", KEY_META, "meta:/type", "string", KEY_END);
keyGetMeta (k, "meta:/type");
keySetMeta (k, "meta:/type", "long");
```

This also makes it far less likely that you run into problems with one of the cases, where the `meta:/` is actually needed, e.g. things like:

```c
Key * k = keyNew ("/foo", KEY_META, "type", "string", KEY_END);

if (strcmp ("type", keyName (keyGetMeta (k, "type"))) == 0) {
   // doesn't work
}

if (strcmp ("meta:/type", keyName (keyGetMeta (k, "type"))) == 0) {
   // works
}
```

src/libs/elektra/hooks.c
kodebach: There is some info about such tests in #4246 (not merged yet): 

https://github.com/ElektraInitiative/libelektra/blob/2e89bfdd38f6cefb0deb17d0e677385f92367b33/doc/decisions/header_file_structure.md?plain=1#L109-L119

You can try to follow this idea (probably needs new CMake setup), or you just do what we did until now and:
1. remove the `static`
2. Add the function name to `libelektraprivate_1.0` in the right `symbols.map` (in your case `src/libs/elektra/symbols.map`), to actually export the symbol
3. make sure the test links against the right library

doc/dev/hooks.md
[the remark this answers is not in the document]
atmaxinger: I have modified and renamed it now.

kodebach (approved):
Great work, LGTM. I really like the way you wrote the tests. Very clean. I have a few other comments, but IMO the PR could be merged as-is.

I also noticed that there are [10 new tests that fail](https://github.com/ElektraInitiative/libelektra/pull/4471/checks?check_run_id=8475876489) compared to [`new-backend`](https://github.com/ElektraInitiative/libelektra/runs/8387201378) (see [also](https://cirrus-ci.com/task/5163487511969792)):
```
testshell_markdown_blacklist
testshell_markdown_conditionals
testshell_markdown_email
testshell_markdown_ipaddr
testshell_markdown_length
testshell_markdown_tutorial_validation
testkdb_error
testkdb_nested
testkdb_simple
```

Please have a quick look, if you find any obvious problems that are easy to fix. If you don't find anything or it's too much work, just leave it for another PR. I think the issue might be that you disable all "global plugins" including `spec`.

atmaxinger:
```
testkdb_error
testkdb_nested
testkdb_simple
```
Those seem to expect `0` as return from `kdbGet`. The only way this could happen is if there is no backend left. I don't think that has anything to do with my changes.

```
testshell_markdown_blacklist
testshell_markdown_conditionals
testshell_markdown_email
testshell_markdown_ipaddr
testshell_markdown_length
testshell_markdown_tutorial_validation
```
I can't run these for whatever reason (probably missing dependency), but if I interpret it correctly, these tests just execute the shell commands included in the README of those plugins?

kodebach:
> I can't run these for whatever reason (probably missing dependency),

I think some of these plugins don't have dependencies. You probably haven't enabled them in CMake. When you run `cmake`, set `-DPLUGINS=ALL` that enables all plugins for which you have the dependencies.

> these tests just execute the shell commands included in the README of those plugins?

Yes, all the `testshell_markdown_*` tests run shell scripts from some markdown file. If the suffix is a plugin name, then the script is the plugin's README. `testshell_markdown_tutorial_validation` is from another file. Of the top of my head I'm not sure which one...


> I don't think that has anything to do with my changes.

In some way it must be your changes, since the test worked in the latest run on `new-backend`. But like I said, if you don't find the problem, just leave it.

atmaxinger:
> In some way it must be your changes, since the test worked in the latest run on new-backend. But like I said, if you don't find the problem, just leave it.

Yes, the problem lies in `goptsActive`. Previously, it was checked like this:
```c
bool goptsActive = handle->globalPlugins[PROCGETSTORAGE][MAXONCE] != NULL;
```

Which doesn't really tell you if `gopts` is really activated, just if there is some plugin loaded there. Spoiler: the `list` plugin is. This with the fact that both backends are in the `proc:/` namespaces led to the following block executing:

https://github.com/ElektraInitiative/libelektra/blob/3e9f5ed10725627405fb36c7a1ea33fa7eb67a1e/src/libs/elektra/kdb.c#L1780-L1794

So this is more or less a bug/works-by-accident behaviour of the `new-backend` branch IMHO.

kodebach:
Hm yeah you're right. Let's leave it the way it is. We'll fix it in another PR at some point.

Although something needs to change, because the failing tests haven't changed for a long time. For example this one

https://github.com/ElektraInitiative/libelektra/blame/5ebc9a4ba5215950f990f7dca948ce310f764079/tests/kdb/testkdb_simple.cpp#L482

hasn't changed for at least 7 years. That's a lot older than `gopts` itself.

atmaxinger:
@markus2330 can we merge this so I can continue with spec and the other hooks?

markus2330:
@atmaxinger Thank you, great job! Is there already something to test for @Eiskasten?

Issue #4444, opened 2022-08-24
remove specload plugin


kodebach:
The current `specload` plugin is not really recommended, but it could be improved on the `new-backend` branch.

I propose 3 changes:
1. Turn `specload` into a standalone backend plugin
2. Remove the modifications/overrides feature.
3. Remove `file` mode
4. Add `symbol` mode, in addition to `app`/"executable" mode

### Standalone backend plugin

Instead of being a storage plugin for the file-based backend plugin `backend`, the `specload` plugin should be backend plugin itself. This makes the various restrictions it imposes a lot more natural and gets rid of some problems, e.g. in relation to the resolver.

To do this we need to do a few things:
1. `specload` should be a very minimal and restricted backend plugin similar to `modules`.
2. `specload` can only be mounted on `spec:/` parent keys.
3. Everything that can currently be configured via the plugin config, should instead be configured via the mountpoint definition and be processed in `init` instead of `open`.

### Remove modifications/overrides

`specload` will be read-only, i.e. like `modules` it doesn't provide a `set` function at all. All code related to modifications or the override file shall be removed.

The functionality implemented by `specload` should only be:
- in `open` create an instance of the plugin
- in `close` free all resources
- in `init` parse the mountpoint definition (previously done from config in `open`) and return `ELEKTRA_PLUGIN_STATUS_NO_UPDATE` (read-only)
- in `get`:
  - during `resolver` phase: return `ELEKTRA_PLUGIN_STATUS_SUCCESS` without doing anything
  - during `storage` phase: load the spec as configured in the mountpoint definition and return
  - during any other phase: return `ELEKTRA_PLUGIN_STATUS_NO_UPDATE` without doing anything

### Remove `file` mode

Remove `file` mode since, without the modifications feature, it doesn't provide any benefit over a normal file-based backend with `quickdump` as storage.

### New `symbol` mode

> **Note** `system:/` keynames below refer to the `KeySet * definition` passed to the `init` function.

The existing modes of the `specload` plugin are:
- "executable" mode: An executable specified as `system:/app` is called with the arguments defined in `system:/app/args/#`. The spec is read in `quickdump` format from the `stdout` of this executable.
- "file" mode: `specload` directly instructs `quickdump` to load the file defined in `system:/file` (removed above).

[...]

PR #4484, opened 2022-09-25
[new-backend] docs

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

need to be checked.

      (not in the PR description)

## Review

## Labels


doc/dev/backend-plugins.md
```diff
@@ -130,7 +130,7 @@ It is where validation, generation of implicit values and similar tasks happen.
 
 Finally, `libelektra-kdb` merges the keyset returned by the `poststorage` phase with the ones returned by other backend plugins for different mountpoints and then returns it to the user.
 
-TODO: Are modifications to `parentKey` visible to the user?
+<!-- TODO: Should modifications to `parentKey` (beyond errors/warnings) be visible to the user? -->
```
kodebach, line 133: input welcome

doc/dev/backend-plugins.md
```diff
@@ -237,7 +237,7 @@ This makes the `postcommit` phase mostly useful for logging.
 
 Finally, `libelektra-kdb` merges the keyset returned by the `postcommit` phase (which is still the same one that was returned by the `prestorage` phase) with the ones returned by other backend plugins for different mountpoints and then returns it to the user.
 
-TODO: Are modifications to `parentKey` visible to the user?
+<!-- TODO: Should modifications to `parentKey` (beyond errors/warnings) be visible to the user? -->
```
kodebach, line 240: input welcome

doc/dev/backend-plugins.md
```diff
@@ -268,4 +268,4 @@ In the `rollback` phase the backend plugin:
 
 Finally, `libelektra-kdb` will restore `ks` to the state in which the user provided it and return.
 
-TODO: Are modifications to `parentKey` visible to the user?
+<!-- TODO: Should modifications to `parentKey` (beyond errors/warnings) be visible to the user? -->
```
kodebach, line 271: input welcome

doc/dev/kdb-operations.md
```diff
@@ -1,9 +1,11 @@
 # Proposal: KDB Operations
 
 There are four main _operations_ in `libelektra-kdb`: `open`, `get`, `set` and `close`.
-For each of these there is a `kdb*` function the user calls to trigger the operation and plugins implement a `elektra<Plugin>*` function for each of the operations they support (at least `get`).
+For each of these there is a `kdb*` function the user calls to trigger the operation and plugins export a function for each of the operations they support (at least `get`).
 
-Additionally, plugins may implement `elektra<Plugin>Commit` and `elektra<Plugin>Error`.
+<!-- TODO: should commit and error really be separate? With the new elektraPluginGetPhase they could just be part of set... -->
```
kodebach, line 6: @markus2330 What do you think? IMO `commit`/`error` should be merged back into `set`.

The current solution just creates unnecessary extra code in both `libelektra-kdb`

https://github.com/ElektraInitiative/libelektra/blob/fac849659be65b3021ded570dcf33e2edb95c407/src/libs/elektra/kdb.c#L2028-L2035

and backend plugins

https://github.com/ElektraInitiative/libelektra/blob/fac849659be65b3021ded570dcf33e2edb95c407/src/plugins/backend/backend.c#L562-L589

(there is more and it is and is all basically copy-paste from the set code)

doc/dev/kdb-operations.md
```diff
@@ -61,13 +62,14 @@ Properties of `kdbGet()`:
 
 - After calling `kdbGet (kdb, ks, parentKey)`, the KeySet `ks` will contain _all keys_ (including their values) that are stored in _any backend_ with a mountpoint that is _below `parentKey`_.
 - After calling `kdbGet (kdb, ks, parentKey)`, below `parentKey` the KeySet `ks` will _only_ contain keys that are stored in a backend.
+  <!-- TODO: what about default:/ keys? Should they be kept to allow runtime defaults? -->
```
kodebach, line 65: input welcome

IMO it's cleaner, if `kdbGet` always reflects the KDB exactly. Defaults should be provided via the spec and if that isn't possible for some reason, you can always handle the `NULL` return value from `ksLookup`.

doc/dev/kdb-operations.md
```diff
@@ -133,53 +133,47 @@ The purpose of the `set` operation is to write data from a `KDB` instance into b
 
 Properties of `kdbSet()`:
 
+<!-- TODO: what about spec/copy? Should that really not be reflected? With the steps below the changes would show up... -->
```
kodebach, line 136: input welcome

doc/dev/kdb-operations.md
```diff
@@ -133,53 +133,47 @@ The purpose of the `set` operation is to write data from a `KDB` instance into b
...
    From now on ignore all backends that have not changed.
    From now on also ignore all backends that were initialized as read-only.
    Issue a warning, if a change was detected (via `KEY_FLAG_SYNC`) in a read-only backend.
+   > **Note**: Steps 4-6 might be combined into a single procedure that deep-copies only keys from changed backends into separate KeySets per backend
 7. Run the `resolver` and `prestorage` on all backends (abort immediately on error and go to e).
-<!-- TODO: can this merge+split be avoided?
-    do we really need to call spec for this?
-    Could we use metaspec:/ namespace and just remove all those keys?
- -->
+<!-- TODO: instead of this merge, spec/remove, split sequence, maybe we could add a metaspec:/ namespace and simply remove all those metakeys? -->
```
kodebach, line 157: Adding a `metaspec:/` namespace is pretty clean, but it is also a lot of effort and has some drawbacks (e.g. no more direct `keyCopyMeta`).

We can't call `spec/remove` separately for all mountpoints, because that would fail for e.g. `spec:/foo` and `user:/foo/bar` as mountpoints. The spec from `spec:/foo` wouldn't apply to keys below `user:/foo/bar`.

However, if `spec/remove` only ever modifies existing keys (which should be possible), we only need the merge and can simply not do the second split. We would just create a `KeySet` that combines alle the results from Step 7 and run `spec/remove` on that. Because this new `KeySet` and the ones from Step 7 share the same `Key`s, we could just use the ones from Step 7 after `spec/remove`.

doc/dev/mountpoints.md
```diff
@@ -32,7 +32,13 @@ The _mountpoint config_ is defined by the keys below `system:/elektra/mountpoint
 It is merged into the configuration keyset of every plugin of this mountpoint (including the backend plugin).
 Specifically, the mountpoint config is put into the `system:/` namespace, while the config from `system:/elektra/mountpoints/<mountpoint>/plugins/<ref>/config` is put into the `user:/` namespace of the keyset passed to a plugin.
 
-<!-- TODO: cover config/needs dynamic config; put into default:/ at runtime ? -->
+<!-- TODO: implement `config/needs` -->
```
kodebach, line 35: Currently, `config/needs` is read during `kdb mount` and into the mountpoint configuration and persisted. That means, if a plugin receives an update and changes the `config/needs` in it's contract, the changes will only apply for newly created mountpoints.

doc/dev/mountpoints.md
```diff
@@ -32,7 +32,13 @@ The _mountpoint config_ is defined by the keys below `system:/elektra/mountpoint
...
+
+Additionally, there is the `config/needs` part of a plugins contract.
+This part of the contract can be used by a plugin to provide configuration for another plugin it depends on.
+In particular, these keys are put into the `default:/` namespace, meaning this config can be overriden by both the _mountpoint config_ in `system:/elektra/mountpoints/<mountpoint>/config` and the plugin's config in `system:/elektra/mountpoints/<mountpoint>/plugins/<ref>/config`.
+
+<!-- TODO: swap defaul:/ and system:/ to mimic current behaviour? -->
```
kodebach, line 41: Currently, `config/needs` takes precedence over the mountpoint configuration provided to `kdb mount -c`. I'm not sure which way is better. On the one hand `config/needs` feels more like defaults to me, on the other hand, if the mountpoint configuration overrides it, the whole "load `config/needs` during `kdbOpen`" is pointless.

kodebach (commented):
@markus2330 There are some open questions. Input is welcome.

Also @atmaxinger I made up some names for the hooks for notifications and spec. If you have better ideas go ahead and suggest them. Also, if you work on spec, you may be interesting in comment about `metaspec:/`/`spec/remove`.

doc/dev/backend-plugins.md
markus2330: Yes, everything should be visible to the user. The string is by convention the "connection" to the underlying storage (file). But @flo91 can reconsider this. Anyway, it should be possible to pass out information by the backends.

doc/dev/kdb-operations.md
markus2330: This is up to @flo91 to decide. I thought it is a good idea to have commit+error, in particular for the database backend.

doc/dev/kdb-operations.md
markus2330: default keys should be created by the spec plugin

doc/dev/kdb-operations.md
markus2330: Please shortly describe why the changes would show up.

doc/dev/backend-plugins.md
kodebach: Ah yes, for things like `kdb file` you need the parentKey value of the backend. However, I think it might be better, if communicate this via metakeys e.g. `meta:/backends/#/storageid`. This would allow passing back the information about _all_ the backends that have been used.

> it should be possible to pass out information by the backends.

Do we need any information beyond that? My metadata idea would allow for limited extra info.

In any case, I think everything should be standardized and backends should be allowed to do whatever to the parentKey.

doc/dev/kdb-operations.md
kodebach: So this is the correct behaviour?

```c
Key * parentKey = keyNew ("/foo", KEY_END);
Key * x =  keyNew ("default:/foo/bar", KEY_END);
ksAppendKey (ks, x);
kdbGet (kdb, ks, parentKey);

// x is NEVER in ks, because x overlaps with the things loaded from the KDB
```

doc/dev/kdb-operations.md
kodebach:
```c
Key * parentKey = keyNew ("/foo", KEY_END);
kdbGet (kdb, ks, parentKey);
Key * x = keyNew ("user:/foo/new/key", KEY_END);
ksAppendKey (ks, x);
kdbSet (kdb, ks, parentKey);
```

If there is a `spec:/foo/new/key` that spec must of course be applied to `user:/foo/new/key` at the start of `kdbSet` to ensure `x` is validated correctly. The question is whether the actual key `x` should be modified, or whether `kdbSet` should not touch `ks` (and it's contents) at all and the spec data would only show up after the next `kdbGet`?

doc/dev/mountpoints.md
markus2330: Is it still the case that both configs are available and it is only about precedence? Imho it should be the other way round, when the user explicitly asked for some config, it should be preferred.

So the order of preference is (highest first):

- plugin specific config (kdb mount ... plugin `<config>`), could be in `dir:` namespace
- config passed by user (kdb mount `-c <config>` ...), could be in `user:` namespace
- config calculated from backend (`config/needs` and/or other config that might stem from backend), could be in `system:` namespace

@flo91 did you already touch that topic?

doc/dev/mountpoints.md
markus2330: Good point. I created #4494 to tackle this (amongst other) issues.

doc/dev/kdb-operations.md
markus2330: I don't get the comment. Is there a question? metaspec is definitely out of scope for new-backend so it also shouldn't be mentioned in the PR.

doc/dev/kdb-operations.md
markus2330: Actually it sounds like a very good idea that meta-data gets copied and the defaults get created in `elektraKdbSet` **if** the spec implementation doesn't get too complicated.

The alternative would be that we require applications to always use `elektraKdbGet` afterwards. And this would get quite complicated, as in the error case applications already need to call `elektraKdbGet` and then merge the ks.

```c
ret = elektraKdbSet(kdb, ks...);
if (ret == -1) {
    elektraKdbGet(kdb, ksdup...);
    // merge ks and ksdup to ks
   elektraKdbSet(kdb, ks, ...)
} else if (ret == 1) {
  // also on success we need elektraKdbGet but not if nothing was changed ret==0?
  elektraKdbGet(kdb, ks...);
}
// now "ks" is allowed to be used again by the application
```

IMHO examples/kdbset.c is already way too complicated!

doc/dev/kdb-operations.md
kodebach: Okay, but then the question is the same, but for `spec:/` and `proc:/`

If a `spec:/` backend exists at `spec:/foo` and a `proc:/` backend (e.g. for `poststorage` validation) exists at `proc:/foo`. Currently this would happen:

```c
Key * parentKey = keyNew ("/foo", KEY_END);
Key * x =  keyNew ("spec:/foo/bar", KEY_END);
ksAppendKey (ks, x);
Key * y =  keyNew ("proc:/foo/bar", KEY_END);
ksAppendKey (ks, y);
kdbGet (kdb, ks, parentKey);

// x and y are NEVER in ks, because x and y overlap with the things loaded from the KDB
// other Key * with the same name may exist in ks
```

So a hard-coded spec is possible, but you get either the entire hard-coded spec OR the entire spec from the KDB for one parent key of the spec. You can't have e.g. most of the spec from the KDB, but `spec:/foo/bar` is from the hard-coded spec.

Is that acceptable?

doc/dev/kdb-operations.md
kodebach: > the spec implementation doesn't get too complicated.

If you look at the operational sequences I wrote for `get` and `set`, both use the hook `spec/copy`. That was intentional, because AFAICT the exact same code could be used both `get` and `set`. In `set` just need an additional `spec/remove` hook called a bit later, to remove the `meta:/` keys before we store the config.

The question here is actually just, whether `spec/copy` should be called

1. before `ksDeepDup` and therefore make the updated `meta:/` keys visible outside `kdbSet`
2. or after `ksDeepDup` thereby keeping all `spec` related changes local to `kdbSet`

doc/dev/kdb-operations.md
kodebach: This is basically ideas about optimizing the operational sequence of `kdbSet`. Currently we need to split `ks` into the backends, do some work, then merge everything back into on `KeySet`, call `spec/remove` and split it back into the backends. With `metaspec:/` we wouldn't need the "merge, `spec/remove`, split" steps. Instead we could remove all `metaspec:/` keys directly in `kdbSet` and not have the `spec/remove` hook at all. Even without `metaspec:/` I think it may be possible (with a careful implementation of `spec/remove`) to only do "merge, `spec/remove`" and but not the "split" step.

I will just add a note that these are possible future optimizations. The only actual question is whether `metaspec:/` is still on the table. If not, I won't mention that part.

doc/dev/mountpoints.md
kodebach: Actually, I would tackle `config/needs` differently. It is simple enough to always load it at runtime. We don't need to remember the command for that. See comment below.

doc/dev/mountpoints.md
kodebach: > So the order of preference is (highest first):

True, we can just move the plugin config to `dir:/`. Then a plugin could add it's own `default:/` keys before calling `ksLookup (config, ...)` and have defaults always available.

> config calculated from backend (`config/needs` and/or other config that might stem from backend)

What other config?

Currently there is:

1. Plugin specific config (`kdb mount ... plugin <config>`) stored in `system:/elektra/mountpoints/<mountpoint>/plugins/<ref>/config`
2. Mountpoint config (`kdb mount -c <config> ...`) stored in `system:/elektra/mountpoints/<mountpoint>/config`
3. `config/needs` loaded from plugins contracts

I don't know of any other way to provide plugin config

doc/dev/kdb-operations.md
markus2330: I don't think that code duplication in backend plugins should be the decisive factor here. IMHO more important is the clarity in resolvers. But I am open to both ways, having plugins with only open/get/set/close is more elegant.

doc/dev/kdb-operations.md
```diff
@@ -22,33 +24,32 @@ The basic flow of this operation is:
 
 1. Create empty `KDB` instance
```
markus2330, line 25: I think this can be summarized more, otherwise it does not provide much benefit from reading the code.

doc/dev/kdb-operations.md
```diff
@@ -22,33 +24,32 @@ The basic flow of this operation is:
...
-5. Process contract
-6. Parse mountpoints
-7. Configure `KDB` instance with real mountpoints
-8. Add hard coded mountpoints to `KDB` instance
-
-> **Note:** Despite sharing a name, the `open` operation **does not** call the `elektra<Plugin>Open` function of any plugin (except within the bootstrap `get` operation).
+3. Run bootstrap `get` operation:
+   This loads the contents of `system:/elektra/mountpoints` so that the mountpoints can be configured.
+4. Process contract and set up plugins for hooks (see [Hooks](hooks.md))
+5. Parse mountpoints:
```
markus2330, line 30: Ideally give some links to the code (e.g. function name)

doc/dev/kdb-operations.md
```diff
@@ -22,33 +24,32 @@ The basic flow of this operation is:
...
-- explicit `default:/` mountpoints are illegal
-  They can only exist via cascading mountpoints.
-  This is so that default values can be checked and validated by plugins.
-- `spec:/` mountpoints may exist, but they are treated differently during `get` and `set`
-- `proc:/` mountpoints may exist, but they are read-only (see below)
-- `dir:/`, `user:/` and `system:/` mountpoints can be created without restrictions, except for the one below
+- `dir:/`, `user:/` and `system:/` mountpoints can be created without restrictions, except for the reserved sections listed below.
+- `spec:/` mountpoints can be created with the same restrictions, but they are also treated specially during `get` and `set`.
+- `proc:/` mountpoints are always read-only and receive special treatment during `get`
+- `default:/` mountpoints are read-only and receive special treatment during `get`, specifically they only go through the `poststorage` phase
```
markus2330, line 43: It would be interesting what the special treatment is.

doc/dev/kdb-operations.md
markus2330: Yes, I don't see a use case for individual hardcoded overrides. (Only the other way round I see a use case but even this case we want to eliminate, #4444). So the application either supplies full hardcoded proc or no proc and/or full hardcoded spec or no spec.

doc/dev/kdb-operations.md
markus2330: It sounds to me like a feature if the spec related changes are visible in the KeySet the application has. Then applications can simply fire off "elektraKdbSet" and, if they don't care if it is actually persisted, are done (The ks they have is consistent and complete). Or do I miss here something important?

doc/dev/kdb-operations.md
markus2330: It is on the table (if someone has the time!) but nevertheless I would not mention it. It is very important that documentation always refers to how it is, not how it might be. TODO markers for possible optimizations also have little value as optimizations are driven by profiling, not by looking at TODOs.

doc/dev/mountpoints.md
markus2330: Yes, that would be one problem less. But as said above: the docu should reflect what is, not what could be done. So if you also implement this before merging, I am a big fan of it :+1:

doc/dev/mountpoints.md
markus2330: > What other config?

Maybe some backends will provide extensions to the contracts or want to append other configs. But yes, this is hypothetical, these 3 categories are perfect for now and cover everything we support.

doc/dev/mountpoints.md
```diff
@@ -124,71 +130,68 @@ Here we will look at a few more.
...
 This example shows an entirely different type of backend.
 The hypothetical `db_backend` is backed by a database.
 In this case it is configured for a PostgreSQL database running on `127.0.0.1:5432` to which we connect as user `admin`.
 
 We also configured the `network` plugin to run in the `prestorage` phase of the `set` operation.
-Which phases can be used and how the must be configured of course depends on `db_backend`.
+Which phases can be used and how they must be configured of course depends on `db_backend`.
 
-One might also imagine there could be a (probably quite complicated) part of the mountpoint definition that defines how the relational tables of the database are mapped into the KDB.
+There might also be a (potentially quite complicated) part of the mountpoint definition that defines how the relational tables of the database are mapped into the KDB.
```
markus2330, line 189: Hypothetical, better don't write anything unless @flo91 confirms.

src/libs/elektra/kdb.c
```diff
@@ -1885,6 +1871,17 @@ int kdbGet (KDB * handle, KeySet * ks, Key * parentKey)
 		goto error;
 	}
 
+	/* TODO (kodebach): implement actual steps with new global plugins
```
markus2330, line 1874: Is this a marker for @atmaxinger?

markus2330 (approved):
Great work, hope to see this docu updates soon in new-backend!

doc/dev/backend-plugins.md
markus2330: > Ah yes, for things like kdb file you need the parentKey value of the backend. However, I think it might be better, if communicate this via metakeys e.g. meta:/backends/#/storageid. This would allow passing back the information about all the backends that have been used.

I don't see an immediate benefit of it. `kdb file`, the only customer of that feature, would get more complicated.

> In any case, I think everything should be standardized and backends should be allowed to do whatever to the parentKey.

I fully agree. Please document this.

kodebach: > > In any case, I think everything should be standardized and backends should be allowed to do whatever to the parentKey.
> 
> I fully agree. Please document this.

Oops that was a bad typo on my part... I meant to say "backends should **NOT** be allowed to do whatever to the parentKey"

> I don't see an immediate benefit of it.

It would be a bit more structured than the current solution, which seems a bit arbitrary. Especially, if you don't know about the internal behaviour of `kdbGet` and backend plugins.

> `kdb file`, the only customer of that feature, would get more complicated.

I thought `kdb file` could just print all the backends that were accessed, but I guess that's compatible with the current behaviour.

So probably we just keep the current behaviour. Adding a way to find all the backends that where accessed (both parent keys and filenames/storageids) could still be useful for diagnostic purposes. Currently, only `kdbGet` has this information. The caller doesn't even get to know whether the cache was used.

doc/dev/kdb-operations.md
kodebach: Another (more dramatic) option would be to add the current phase as a 4th parameter to the `get` and `set` functions of plugins:

```c
int elektraFooGet (Plugin * handle, ElektraGetPhase phase, KeySet * ks, Key * parentKey);
int elektraFooSet (Plugin * handle, ElektraSetPhase phase, KeySet * ks, Key * parentKey);
```

The it's also quite clear that `phase` should probably be handled in some way.

doc/dev/kdb-operations.md
kodebach: > if they don't care if it is actually persisted

Well they should also know that, because it should be persisted if `kdbSet` doesn't return an error. But of course there could (in theory) be things done by (less well designed) plugins that only show up after a `kdbGet`.

> Or do I miss here something important?

No, I also see this a feature for `kdbSet`.

doc/dev/kdb-operations.md
kodebach: > refers to how it is, not how it might be

I would only have mentioned it, if it would have been implemented soon-ish, e.g. during FLOSS. But I think that's unlikely so I'll leave it.

> TODO markers for possible optimizations also have little value as optimizations are driven by profiling, not by looking at TODOs.

Finding which optimizations are worth the implementation effort and finding real bottlenecks is determined by profiling. Marking locations where an obvious, but maybe not significant (enough), improvement can be made can be done in advance.

In this case: Not merging and re-splitting a `KeySet` is obviously faster since it's less work, what we don't know is if it actually matters.

doc/dev/mountpoints.md
kodebach: Probably not before merging, but before release. I'll explain what should happen, in case someone else wants to do it before I have time, and also for future reference for myself. But I'll clearly mark it as "not implemented yet".

doc/dev/mountpoints.md
kodebach: > Maybe some backends will provide extensions to the contracts or want to append other configs

That would need some changes, but should be doable (not for now though).

I'm thinking:

- Always load the backend plugin of a mountpoint last
- Add an API that let's plugins access the contract of other plugins
- Add an API that let's plugins get a list of other plugins on the same mountpoint
- Since the contract is a `KeySet*` created via function call anyway, the plugin could look at what other plugins are there and adapt it's contract (including `config/needs`) accordingly

doc/dev/kdb-operations.md
kodebach: This was intended as specification, i.e. the target audience is contributors. The goal was a specification so that we know what must happen in code. For developers a separate summarized version needs to be created.

doc/dev/kdb-operations.md
kodebach: The link is actually there, but the other way around. The `kdb*` functions have `// Step X` comments that refer to these steps.

kodebach:
@markus2330 I didn't have time to incorporate your review yet. But I did now look through most of the docs and removed some more things.

I didn't remove much and instead added clear warnings ("Outdated"). All these files should be updated before the release. I didn't remove them because they are all still mostly correct and I didn't want to rewrite them all (because just removing something the middle probably creates more confusion).

I didn't really look closely at the ALLCAPS.md files, since they all seems mostly unrelated to the new-backend changes.

For the manpages (especially `elektra-*`) I didn't do anything. I didn't just want to remove them because the might still contain some useful stuff. But anything related to mountpoints/backends probably needs to rewritten.

Another big reason why I removed so little is: I think the old stuff can be used as reference for the person writing the updated docs. Only the outdated parts really need to be replaced.

src/libs/elektra/kdb.c
kodebach: Actually, I just moved this down, because before it went "Step 14" then "Step 13". In fact this part is already gone because of #4495.

doc/dev/mountpoints.md
kodebach: The whole "Further Examples" section is purely hypothetical, but I'll remove it.

kodebach:
@markus2330 I have now updated and rebased this PR. Please to a quick, cursory check and merge the PR, unless there is some really major issue. I'll do more minor update in another PR

kodebach:
I merged this PR since there is still #4187 which can still be reviewed.

doc/dev/kdb-operations.md
markus2330: No, the phase can be communicated via the plugin handle better. Not every plugin is interested in the phase. In any case: there is doc/decisions/commit_function.md, which needs an update.

doc/dev/kdb-operations.md
markus2330: More important are the links from docu to code (traceability links) because usually people start looking at the docu and then would like to dive into the implementation.

doc/dev/kdb-operations.md
markus2330: Such forward references are not ideal, even less ideal if they are implicit.

doc/dev/kdb-operations.md
kodebach: > there is doc/decisions/commit_function.md, which needs an update.

Sure, but do we want a separate commit/error function or not. Because if not, I would just remove the file. I can create a new `kdb_phases.md` decisison, but I'm not sure we need it. It's already implemented and it is documented in `doc/dev/backend-plugins.md`

doc/dev/kdb-operations.md
kodebach: I'll add an "explained below", but IMO it is very obvious where to look already. The file clearly has one section per operations, so if I refer to the `get` operation that's where you should look.

I also know that forward references are not ideal, but there is no way around it here. Explaining all of `get` before `open` doesn't make sense. Explaining just the special treatment here is also not ideal, because that should clearly be explained in `get`. Duplicating the information will just lead to conflicting information in future.

doc/dev/kdb-operations.md
kodebach: Yes, and that is fine here IMO. This is contributor documentation. You read this and then go look at `kdbGet`/`kdbSet` code and will see references to the steps you already now.

There is also no better solution. Not every step is a separate function (*) and referring to line number is obviously bad. What I will do, however, is mention that the `// Step X` comments exist.

(*) Making every step a separate function would create unnecessary friction in the code IMO. You need good names for everything, sometimes you need to pass a lot of arguments, you need to encode different results into return codes and then check those instead of just doing the cleanup directly, etc.

markus2330:
> Do you have a specific date?

No, it is asap. People may fork their repo any time.

> IMO we should just merge what is there in new-backend ASAP

I fully agree!

> even if you are not fully satisfied with the docs. I'll do my best to make more updates in the next weeks.

Thank you! :heart: 

> NO. There is already lots of docs for the new-backend changes: doc/dev/kdb-operations.md, doc/dev/mountpoints.md, doc/dev/backend-plugins.md. In fact some of those were there before I implemented the changes. I honestly have no idea why you either purposefully ignore those files, or seem to forget them all the time.
>
> Based on those docs, many of the other less low-level docs can easily be updated by other people. For example, AFAICT doc/dev/algorithm.md is supposed be similar to what doc/dev/kdb-operations.md describes, but less like a specification.

E.g. doc/dev/kdb-operations.md is a useful description how to implement a backend. But there are many other viewpoints, e.g. the algorithms etc. which are not yet documented. We need easier understandable docu for new contributors, at least at the level as we had before.

> There are also multiple other people implementing things for new-backend. People who do that was part of their thesis, while I spend time on it, completely unrelated to my thesis topic...

It isn't a perfect match for anyone, as it is a huge left-over and there cannot be a thesis about "finishing others peoples job".

Anyway, I am very grateful for your amazing&huge contributions on the most important topics.

kodebach:
> > IMO we should just merge what is there in new-backend ASAP
>
> I fully agree!

The problem is it seems we disagree about the definition of "ASAP". IMO ASAP is basically now (maybe when I have disabled the failing tests). I think you want a bit more work done, or at least complete your review. I understand that, but any delay will lead to people starting to work on `master` and then the merge could be quite annoying for them.

> We need easier understandable docu for new contributors, at least at the level as we had before.

`doc/dev/algorithm.md` will still exist and that's why I didn't delete it. It just needs to be updated. I didn't have time for that, and I believe that updating it could also be done by someone else who has read the newly added docs. The general idea of the algorithm is still the same, only some details changed, but those changes should be obvious from conflicts between new and old docs.

I have now also updated the doxygen docs for `kdbGet` and `kdbSet` so there is even more information to find what has changed.

> It isn't a perfect match for anyone, as it is a huge left-over and there cannot be a thesis about "finishing others peoples job".

Of course, but if everyone does a bit, it's much easier for everyone.

Issue #4493, opened 2022-09-27
[new-backend] spec hook - default:/ backend


atmaxinger:
I've got the `spec` plugin and its copy hook in a state where it will generate the `default:/` keys according to spec.
However, `backendsDivide` fails because there is no backend for `default:/` keys.

Command:
`kdb meta-get spec:/example/port default`

spec.ini:
```ini
[port]
meta:/default = 8080
```

Expected result:
8080

Actual result:

Sorry, module kdb issued the error C01310:
Internal: Couldn't divide keys into mountpoints before poststorage. Please report this bug at https://issues.libelektra.org.

@kodebach can we just add a backend that is responsible for `default:/` keys? Or can we somehow prevent `default:/` keys from being passed to `backendsDivide`, but so that they are still part of the resulting keyset?

atmaxinger:
I worked around it by cutting all `default:/` keys before the split and then merging them in into the end result.
As there is not backend involved, it should be find if they are not contained in the post storage phase (?).

```c
Key * defaultCutpoint = keyNew ("default:/", KEY_END);
KeySet * defaults = ksCut (dataKs, defaultCutpoint);

// backendDivide
// other phases
// backendMerge

ksAppend (ks, defaults);
ksDel (defaults);
keyDel(defaultCutpoint);
```

kodebach:
> As there is no backend involved, it should be fine if they are not contained in the post storage phase (?).

As written in the docs (`doc/dev/kdb-operations.md`) backends for `default:/` are possible (not sure if they would work right now). This needed to run validation like the `type` plugin on `default:/` keys created by `spec`.

So the cutting and re-adding might work now, but it is not a permanent solution. Instead we should probably create a `default:/` backend for every `spec:/` backend.

markus2330:
So we should mount a backend containing `type` per default as `default:/` backend?

kodebach:
I don't know. Wouldn't that just raise questions why `type` is added by default to `default:/` but not other namespaces?

markus2330:
I think type should be added by default to all the namespaces, see #690

kodebach:
With #4187 it would ultimately be the responsibility of a backend plugin to call `type`. "Added by default" in this context just means, "call `type` even if the user didn't configure it". There are also exceptions like `specload` (see #4444) which obviously shouldn't call `type`. In fact calling `type` on `spec:/ keys doesn't make sense in general.

[...]

Issue #4061, opened 2021-09-21
Fix handling of array parents in `spec` plugin


qwepoizt:
The `spec` plugin changes *which* keys of an array it gets/sets, when a previously empty array becomes non-empty.

## Expected behavior
1. `cd` into `.../libelektra/examples/codegen/econf/`
2.  ``kdb mount `pwd`/spec.ini "spec:/sw/example/econf/#0/current" ni``
3. `kdb spec-mount "/sw/example/econf/#0/current"`
4. `kdb ls "spec:/sw/example/econf/#0"` yields:
```
spec:/sw/example/econf/#0/current
spec:/sw/example/econf/#0/current/format/#
spec:/sw/example/econf/#0/current/format/#/charset
spec:/sw/example/econf/#0/current/format/#/eofnewline
spec:/sw/example/econf/#0/current/format/#/eol
spec:/sw/example/econf/#0/current/format/#/indent/size
spec:/sw/example/econf/#0/current/format/#/indent/style
spec:/sw/example/econf/#0/current/format/#/linelength
spec:/sw/example/econf/#0/current/format/#/pattern
spec:/sw/example/econf/#0/current/format/#/tabwidth
spec:/sw/example/econf/#0/current/format/#/trim
spec:/sw/example/econf/#0/current/root
```
5. `kdb meta-set "user:/sw/example/econf/#0/current/format" array "#2"`
6. `kdb ls "spec:/sw/example/econf/#0"` yields:
```
spec:/sw/example/econf/#0/current
spec:/sw/example/econf/#0/current/format/#
spec:/sw/example/econf/#0/current/format/#/charset
spec:/sw/example/econf/#0/current/format/#/eofnewline
spec:/sw/example/econf/#0/current/format/#/eol
spec:/sw/example/econf/#0/current/format/#/indent/size
spec:/sw/example/econf/#0/current/format/#/indent/style
spec:/sw/example/econf/#0/current/format/#/linelength
spec:/sw/example/econf/#0/current/format/#/pattern
spec:/sw/example/econf/#0/current/format/#/tabwidth
spec:/sw/example/econf/#0/current/format/#/trim
spec:/sw/example/econf/#0/current/root
```

## Actual behavior:
Steps 1. to 5. like expected
6. Yields
```
spec:/sw/example/econf/#0/current
spec:/sw/example/econf/#0/current/format/#/charset
spec:/sw/example/econf/#0/current/format/#/eofnewline
[...]

PR #4495, opened 2022-09-27
[new-backend] spec hook

atmaxinger:
This is still a work in progress:
- [x] remove meta data
- [x] settings for enabling/disabling spec hook - will be discussed in #4499 
- [x] clarify wether `copy` should behave differently between get and set - will be discussed in #4500
- [x] clarify handling of `default:/` keys with regards to backend - see also  #4493

## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/src/libs/elektra/hooks.c
+++ b/src/libs/elektra/hooks.c
@@ -171,7 +171,7 @@ static bool isGoptsEnabledByContract (const KeySet * contract)
 static bool isSpecEnabledByConfig (const KeySet * config)
 {
 	// TODO: check for system:/elektra/hook/spec/enabled or system:/elektra/hook/spec/disabled or something else ... TBD
-	//       See this discussion: https://github.com/ElektraInitiative/libelektra/pull/4471#discussion_r974129215
+	//       See this discussion: https://github.com/ElektraInitiative/libelektra/issues/4499
 	return true;
 }
 
--- a/src/libs/elektra/hooks.c
+++ b/src/libs/elektra/hooks.c
@@ -73,7 +73,7 @@ static int initHooksSpec (KDB * kdb, Plugin * plugin, Key * errorKey)
 	kdb->hooks.spec.copy = (kdbHookSpecCopyPtr) getFunction (plugin, "hook/spec/copy", errorKey);
 	kdb->hooks.spec.remove = (kdbHookSpecRemovePtr) getFunction (plugin, "hook/spec/remove", errorKey);
 
-	if(kdb->hooks.spec.copy == NULL || kdb->hooks.spec.remove == NULL)
+	if (kdb->hooks.spec.copy == NULL || kdb->hooks.spec.remove == NULL)
 	{
 		return -1;
 	}
@@ -168,7 +168,7 @@ static bool isGoptsEnabledByContract (const KeySet * contract)
 	return isEnabled;
 }
 
-static bool isSpecEnabledByConfig(const KeySet * config)
+static bool isSpecEnabledByConfig (const KeySet * config)
 {
 	// TODO: check for system:/elektra/hook/spec/enabled or system:/elektra/hook/spec/disabled or something else ... TBD
 	//       See this discussion: https://github.com/ElektraInitiative/libelektra/pull/4471#discussion_r974129215
@@ -206,7 +206,7 @@ int initHooks (KDB * kdb, const KeySet * config, KeySet * modules, const KeySet 
 		goto error;
 	}
 
-	if(isSpecEnabledByConfig (config) &&
+	if (isSpecEnabledByConfig (config) &&
 	    initHooksSpec (kdb, loadPlugin ("spec", kdb->global, modules, contract, errorKey), errorKey) != 0)
 	{
 		goto error;
--- a/src/libs/elektra/hooks.c
+++ b/src/libs/elektra/hooks.c
@@ -20,15 +20,15 @@ void freeHooks (KDB * kdb, Key * errorKey)
 	{
 		elektraPluginClose (kdb->hooks.gopts.plugin, errorKey);
 		kdb->hooks.gopts.plugin = NULL;
-		kdb->hooks.gopts.kdbHookGoptsGet = NULL;
+		kdb->hooks.gopts.get = NULL;
 	}
 
 	if (kdb->hooks.spec.plugin != NULL)
 	{
 		elektraPluginClose (kdb->hooks.spec.plugin, errorKey);
 		kdb->hooks.spec.plugin = NULL;
-		kdb->hooks.spec.kdbHookSpecCopy = NULL;
-		kdb->hooks.spec.kdbHookSpecRemove = NULL;
+		kdb->hooks.spec.copy = NULL;
+		kdb->hooks.spec.remove = NULL;
 	}
 }
 
@@ -53,7 +53,7 @@ static int initHooksGopts (KDB * kdb, Plugin * plugin, Key * errorKey)
 
 	kdb->hooks.gopts.plugin = plugin;
 
-	if ((kdb->hooks.gopts.kdbHookGoptsGet = (kdbHookGoptsGetPtr) getFunction (plugin, "hook/gopts/get", errorKey)) == NULL)
+	if ((kdb->hooks.gopts.get = (kdbHookGoptsGetPtr) getFunction (plugin, "hook/gopts/get", errorKey)) == NULL)
 	{
 		return -1;
 	}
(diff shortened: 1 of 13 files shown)


src/plugins/spec/testmod_spec.c
```diff
@@ -608,15 +624,16 @@ int main (int argc, char ** argv)
...
-	test_default ();
-	test_assign_condition ();
-	test_wildcard ();
-	test_require ();
-	test_logMissing ();
-	test_array ();
-	test_require_array ();
-	test_array_member ();
-	// test_remove_meta ();
+	test_hook_copy_default (true);
```
atmaxinger, line 627: Should we test all of them for both the get and set operations? Should there be differences between the operations?
For now, e.g. the set operation does not create default:/ keys.

kodebach: The copy should behave the same in `kdbGet` and `kdbSet` that's why I used the same name.


That said, the current set from `spec` is a bit broken.

To make validation for newly added keys during `kdbSet` work properly, we need to copy `meta:/` keys from `spec:/` to other namespaces at the start of `kdbSet` (before `prestorage`). This should behave the same as in `kdbGet` and the new `meta:/` keys should be visible outside `kdbSet` I think (see https://github.com/ElektraInitiative/libelektra/pull/4484#discussion_r979454050).

However, we don't want to store the copied `meta:/` keys in the other namespace, so we need to remove them before the `storage` phase. To do this we should simply check for every metakey of every key in a namespace other than `spec:/` whether it (the metakey) is pointer-equal to the corresponding metakey on the equivalent `spec:/` key. So something like:

```c
KeySet * specMeta = keyMeta (specKey);
KeySet * meta = keyMeta (key);

for (elektraCursor it = 0; i < ksGetSize (specMeta); i++)
{
    Key * m = ksAtCursor (specMeta, it);
    if (ksLookup (meta, m, 0) == m)
    {
        keySetMeta (key, keyName (m), NULL);
    }
}
```

kodebach:
Haven't had time to look at this properly yet. But there are something that need to be changed about the `spec` plugin (see also https://github.com/ElektraInitiative/libelektra/pull/4495#discussion_r981678559). If you don't have the time right now, or don't want to do it right now. Just try to emulate what happens on `master` and leave some TODOs to point people in the direction of what still needs to be change in `kdb.c`.

---

The things that need to be changed in `spec` are:
- The copy  & remove stuff for `kdbSet` described in https://github.com/ElektraInitiative/libelektra/pull/4495#discussion_r981678559
- `spec` should always report errors/warnings for conflicts instead of the useless it does now.
- Instead of lots of `ksDup`/`ksCut` we should use `ksFindHierarchy` where possible.

src/plugins/spec/testmod_spec.c
atmaxinger: The differences between a set and get operation for the spec plugin were copied from the existing implementation:

- There can be different settings for the conflict handling 
  - This can probably be removed
- neither `assign/condition` nor `default` are processed in set 
  - We can probably also execute this on set
- set operation sets `internal/spec/remove` on `array`
  - remove or also do it in get?
- set removes `internal/spec/array/validated`
  - remove or also do it in get?
- set does this thing here, I'm not sure what it is
    - ???
```c
if (!isKdbGet)
{
	keySetMeta (specKey, "internal/spec/array/validated", NULL);

	if (keyGetMeta (specKey, "internal/spec/array") == NULL && keyGetMeta (specKey, "internal/spec/remove") == NULL)
	{
		ksAppendKey (returned, specKey);
	}
}
```

kodebach: > There can be different settings for the conflict handling
> * This can probably be removed

Yes, all the conflict handling settings should be removed entirely. During `get` all conflicts should be warnings during `set` they should be errors (unless @markus2330 thinks some conflicts should never be errors). That part could be handled from the `libelektra-kdb` side with the "error blocking trick", so the `spec` would be the same.


> neither `assign/condition` nor `default` are processed in set

Not sure about `assign/condition` and whether we actually want to keep that (@markus2330 ?), but `default` should definitely be processed in `set`. See also https://github.com/ElektraInitiative/libelektra/pull/4484#discussion_r979454050



> * set operation sets `internal/spec/remove` on `array`
> * set removes `internal/spec/array/validated`
> * set does this thing here, I'm not sure what it is

All of the stuff related to array handling is very complicated and weird. Part of the weirdness comes from the fact that on `master` the `spec` plugin is only called once during `kdbSet` and somehow there should still be validation during `kdbSet`, but also `spec` should cleanup the copied data (obviously not possible).

See also #2700 and #4061

---

However, please don't waste time on this. Like I said already: Just try to imitate what `spec` does on `master`. The existing tests obviously also expect this behaviour. Once that is done and merged, you can create another PR with the actual updates for `spec` (probably after `new-backend` is merged in `master`).

kodebach: Especially, the conflict handling should remain the same as on `master` for now. Changing this will probably break some tests and take more time than might expect.

markus2330 (approved):
Great work. Elegant tests. I did not check functionality, though.

atmaxinger:
New failing tests in this PR:

`testshell_markdown_kdb_get`: Strange, as the outputs actually match on my machine. (Note: I can not find the `testshell_markdown_*` scripts on my system. I just manually did the steps mentioned in doc/help/kdb-get.md)

`testshell_markdown_kdb_global_umount`: Expects `spec` to be in the list of global plugins, which it isn't anymore. Easy fix, just remove it from the expected values. But as the global mount/global unmount stuff should actually go away with hooks, maybe delete those tests entirely?

`testscr_check_kdb_internal_check`: fails with the error message `error: check of plugin spec with args '' failed`. This seems to somehow test all plugins via execution of `kdb plugin-check $PLUGIN_NAME`. This fails for the `spec` plugin with the following message: 

There are 1 warnings for this plugin
For high quality plugins there should be no warning
Warning #0: no placements information found

This seems about right, as spec does not have any placements anymore. Should we just ignore the spec plugin for this test?

src/libs/elektra/hooks.c
```diff
@@ -138,6 +168,13 @@ static bool isGoptsEnabledByContract (const KeySet * contract)
 	return isEnabled;
 }
 
+static bool isSpecEnabledByConfig (const KeySet * config)
+{
+	// TODO: check for system:/elektra/hook/spec/enabled or system:/elektra/hook/spec/disabled or something else ... TBD
```
atmaxinger, line 173: I don't think we need to solve this before merging this PR. Can be added later at any time.

src/plugins/spec/testmod_spec.c
atmaxinger: > However, please don't waste time on this. [...] Once that is done and merged, you can create another PR with the actual updates for spec.

I agree, this is probably the correct method, so I can continue with the other hooks. 

Please take a look at my comment below with the test cases. I think those still need to be resolved before we can merge it.

src/libs/elektra/hooks.c
kodebach: Agreed.

kodebach:
> testshell_markdown_kdb_get

If the issue is just the `#> got 2 keys` line then change it to whatever gets printed now. AFAIK that line simply prints the size of the keyset returned by `kdbGet`. That's not really something we make any guarantees about, so we can just change the test whenever the size of the keyset changes.

> testshell_markdown_kdb_global_umount

Just remove the test.

> testscr_check_kdb_internal_check

For now I think excluding `spec` and probably `gopts` from the test is the best solution. For the future we might want to introduce a `infos/placements = hook` to indicate the plugin is used for a kdb hook. But adding the probably is a good chunk of work, so just exclude the known hook plugins from these checks.

kodebach (approved):
Great work, LGTM. If you take care of the 3 tests, we can merge this IMO.

atmaxinger:
@markus2330 the tests are fixed now, you can merge this PR.

markus2330:
@atmaxinger please don't self-merge. Probably nobody looked at the test fixes now. In this case (for merges to new-backend) it is not a big deal, however, as it gets reviewed again.

src/plugins/spec/testmod_spec.c
markus2330: It looks like we got someone who will work on the spec plugin! :rocket:

So we should soon have a coordination meeting about spec-related topics.

markus2330: > Not sure about assign/condition and whether we actually want to keep that (@markus2330 ?)

No, we can remove it. Defaults are a superior way. It should be possible via `namespace` spec to give `default` higher priority than others, giving something like an assignment via defaults.

PR #4504, opened 2022-10-03
[new-backend] notification send hooks

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -332,7 +332,14 @@ typedef struct _SendNotificationHook
 	struct _Plugin * plugin;
 	struct _SendNotificationHook * next;
 
+	/**
+	 * Optional, may be NULL
+	 */
 	kdbHookSendNotificationGetPtr get;
+
+	/**
+	 * Optional, may be NULL
+	 */
 	kdbHookSendNotificationGetPtr set;
 } SendNotificationHook;
 
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -327,14 +327,14 @@ struct _KeySet
 #endif
 };
 
-struct _SendNotificationHook
+typedef struct _SendNotificationHook
 {
 	struct _Plugin * plugin;
 	struct _SendNotificationHook * next;
 
 	kdbHookSendNotificationGetPtr get;
 	kdbHookSendNotificationGetPtr set;
-};
+} SendNotificationHook;
 
 /**
  * The access point to the key database.
--- a/src/include/kdbprivate.h
+++ b/src/include/kdbprivate.h
@@ -104,6 +104,9 @@ typedef int (*kdbHookGoptsGetPtr) (Plugin * handle, KeySet * returned, Key * par
 typedef int (*kdbHookSpecCopyPtr) (Plugin * handle, KeySet * returned, Key * parentKey, bool isKdbGet);
 typedef int (*kdbHookSpecRemovePtr) (Plugin * handle, KeySet * returned, Key * parentKey);
 
+typedef int (*kdbHookSendNotificationGetPtr) (Plugin * handle, KeySet * returned, Key * parentKey);
+typedef int (*kdbHookSendNotificationSetPtr) (Plugin * handle, KeySet * returned, Key * parentKey);
+
 typedef Plugin * (*OpenMapper) (const char *, const char *, KeySet *);
 typedef int (*CloseMapper) (Plugin *);
 
@@ -324,6 +327,15 @@ struct _KeySet
 #endif
 };
 
+struct _SendNotificationHook
+{
+	struct _Plugin * plugin;
+	struct _SendNotificationHook * next;
+
+	kdbHookSendNotificationGetPtr get;
+	kdbHookSendNotificationGetPtr set;
+};
+
 /**
  * The access point to the key database.
  *
@@ -382,6 +394,8 @@ struct _KDB
 			kdbHookSpecCopyPtr copy;
 			kdbHookSpecRemovePtr remove;
 		} spec;
+
+		struct _SendNotificationHook * sendNotification;
 	} hooks;
 };
 
--- a/src/libs/elektra/CMakeLists.txt
+++ b/src/libs/elektra/CMakeLists.txt
@@ -84,7 +84,7 @@ if (BUILD_SHARED)
 
 	add_library (elektra-kdb SHARED ${KDB_FILES})
 	add_dependencies (elektra-kdb generate_version_script)
-	target_link_libraries (elektra-kdb elektra-core)
+	target_link_libraries (elektra-kdb elektra-core elektra-ease)
 
 	get_property (elektra-extension_LIBRARIES GLOBAL PROPERTY elektra-extension_LIBRARIES)
 
--- a/src/libs/elektra/CMakeLists.txt
+++ b/src/libs/elektra/CMakeLists.txt
@@ -84,7 +84,7 @@ if (BUILD_SHARED)
 
 	add_library (elektra-kdb SHARED ${KDB_FILES})
 	add_dependencies (elektra-kdb generate_version_script)
-	target_link_libraries (elektra-kdb elektra-core elektra-ease)
+	target_link_libraries (elektra-kdb elektra-core)
 
 	get_property (elektra-extension_LIBRARIES GLOBAL PROPERTY elektra-extension_LIBRARIES)
 
--- a/src/libs/elektra/hooks.c
+++ b/src/libs/elektra/hooks.c
@@ -8,6 +8,8 @@
 
 #include <kdbinternal.h>
 
+static Plugin * loadPlugin (const char * pluginName, KeySet * global, KeySet * modules, const KeySet * contract, Key * errorKey);
+
 /**
  * Uninitializes and frees all hooks in the passed KDB handle.
  *
@@ -29,6 +31,24 @@ void freeHooks (KDB * kdb, Key * errorKey)
 		kdb->hooks.spec.plugin = NULL;
 		kdb->hooks.spec.copy = NULL;
 		kdb->hooks.spec.remove = NULL;
+	}
+
+	if (kdb->hooks.sendNotification != NULL)
+	{
+		struct _SendNotificationHook * hook = kdb->hooks.sendNotification;
+		while (hook != NULL)
+		{
+			elektraPluginClose (hook->plugin, errorKey);
+			hook->plugin = NULL;
+			hook->get = NULL;
+			hook->set = NULL;
+
+			struct _SendNotificationHook * old = hook;
+			hook = hook->next;
+			elektraFree (old);
+		}
+
+		kdb->hooks.sendNotification = NULL;
 	}
 }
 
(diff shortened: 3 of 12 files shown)


src/libs/elektra/kdb.c
```diff
@@ -2465,6 +2472,17 @@ int kdbSet (KDB * handle, KeySet * ks, Key * parentKey)
 	// Step 12c: run postcommit phase
 	runSetPhase (backends, parentKey, KDB_SET_PHASE_POST_COMMIT, true, KDB_SET_FN_COMMIT);
 
+
+
+	struct _SendNotificationHook * sendNotificationHook = handle->hooks.sendNotification;
+	while(sendNotificationHook != NULL)
+	{
+		// TODO (atmaxinger): Is setKs really the correct KeySet?
+		// Unfortunately setKs removes the SYNC flag, so plugins can not reliably detect changed keys ...
```
atmaxinger, line 2481: This is something we need to tackle. As of now the plugins are called, but for example dbus does not correctly detect changed keys, as it relies on the `keeNeedsSync` function.

Maybe we change the contract and provide the plugins with the three KeySets (added, changed, removed) directly?

src/libs/elektra/CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ if (BUILD_SHARED)
 
 	add_library (elektra-kdb SHARED ${KDB_FILES})
 	add_dependencies (elektra-kdb generate_version_script)
-	target_link_libraries (elektra-kdb elektra-core)
+	target_link_libraries (elektra-kdb elektra-core elektra-ease)
```
atmaxinger, line 87: Not sure if I'm supposed to do create a dependency of `elektra-core` on `elektra-ease`, but otherwise I would have to duplicate the array handling code in `-core` ...

tests/ctest/test_hooks.c
```diff
@@ -168,6 +168,33 @@ static void test_initHooks_shouldInitAllHooksWithoutFailure (void)
 	elektraFree (kdb);
 }
 
+static void test_initHooksSendNotifications(void)
```
atmaxinger, line 171: Ignore this test case, this was just for debugging. Tests will be added when the general design has been approved,

src/libs/elektra/hooks.c
```diff
@@ -81,6 +101,64 @@ static int initHooksSpec (KDB * kdb, Plugin * plugin, Key * errorKey)
 	return 0;
 }
 
+static int initHooksSendNotifications (KDB * kdb, const KeySet * config, KeySet * modules, const KeySet * contract, Key * errorKey)
+{
+	struct _SendNotificationHook * lastHook = kdb->hooks.sendNotification = NULL;
+
+	Key * pluginsKey = keyNew ("system:/elektra/hook/notification/send/plugins", KEY_END);
+	KeySet * configuredPlugins = elektraArrayGet (pluginsKey, config);
```
kodebach, line 109: I **really** dislike `elektraArrayGet`. It is a super inefficient way of dealing with arrays. It copies the array part into a separate `KeySet` (which is almost never needed). On top of that it doesn't even use `ksCut` or some other smart way of finding the array. It just iterates _the whole_ `KeySet` and for every key checks if it is an array element.

A better way for iterating the array elements (without `libelektra-ease`) would be:

```c
for (elektraCursor end, it = ksFindHierarchy (config, pluginsKey, &end); it < end; it++)
{
    Key * cur = ksAtCursor (config, it);
    if (keyIsDirectlyBelow (pluginsKey, cur) == 0)
    {
        continue;
    }

    const char * pluginName = keyString (cur);
    // ...
}
```

This doesn't actually check whether the keys are array elements (e.g. `system:/elektra/hook/notification/send/plugins/foo` would work), but we don't really need to check that. If you do want to check, you can use `elektraArrayValidateName (cur) > 0`, but that needs `libelektra-ease` again.

src/libs/elektra/hooks.c
```diff
@@ -81,6 +101,64 @@ static int initHooksSpec (KDB * kdb, Plugin * plugin, Key * errorKey)
...
+
+	Key * cur;
+	while ((cur = ksNext(configuredPlugins)) != NULL)
+	{
+		const char * pluginName = keyString (cur);
+		Plugin * plugin = loadPlugin (pluginName, kdb->global, modules, contract, errorKey);
+
+		if (!plugin)
+		{
+			continue;
```
kodebach, line 126: There should be a warning that the plugin couldn't be loaded.

src/libs/elektra/hooks.c
```diff
@@ -81,6 +101,64 @@ static int initHooksSpec (KDB * kdb, Plugin * plugin, Key * errorKey)
...
+		}
+
+		kdbHookSendNotificationGetPtr getPtr = (kdbHookSendNotificationGetPtr) getFunction (plugin, "hook/notification/send/get", errorKey);
+		kdbHookSendNotificationSetPtr setPtr = (kdbHookSendNotificationSetPtr) getFunction (plugin, "hook/notification/send/set", errorKey);
+
+		if (!getPtr || !setPtr)
+		{
+			elektraPluginClose (plugin, errorKey);
+			continue;
+		}
```
kodebach, line 136: Should we really require both hooks? From a functionality standpoint it doesn't matter to `libelektra-kdb`. We don't expect to get any result from a notification plugin, so unlike with `spec` it's not necessarily wrong to just provide one of the hooks.

For example, there could be an implementation that just uses `hook/notification/send/set` and basically just broadcasts "hey someone wrote to the KDB". Or a very simple auditing mechanism that just records every single `kdbSet` call.

I'm not sure there is an example of just using `hook/notification/send/get`. But as I said, `libelektra-kdb` doesn't expect any particular outcomes so we don't really care which hooks are available.

src/libs/elektra/hooks.c
```diff
@@ -81,6 +101,64 @@ static int initHooksSpec (KDB * kdb, Plugin * plugin, Key * errorKey)
 	return 0;
 }
 
+static int initHooksSendNotifications (KDB * kdb, const KeySet * config, KeySet * modules, const KeySet * contract, Key * errorKey)
+{
+	struct _SendNotificationHook * lastHook = kdb->hooks.sendNotification = NULL;
```
kodebach, line 106:
```suggestion
	struct _SendNotificationHook * lastHook = kdb->hooks.sendNotification;
```

Why do we need to set it `NULL` here? Isn't this function called only after a `freeHooks`?

src/include/kdbprivate.h
```diff
@@ -324,6 +327,15 @@ struct _KeySet
...
 };
 
+struct _SendNotificationHook
+{
+	struct _Plugin * plugin;
+	struct _SendNotificationHook * next;
+
+	kdbHookSendNotificationGetPtr get;
+	kdbHookSendNotificationGetPtr set;
+};
```
kodebach, line 337:
```suggestion
typedef struct _SendNotificationHook
{
	struct _Plugin * plugin;
	struct _SendNotificationHook * next;

	kdbHookSendNotificationGetPtr get;
	kdbHookSendNotificationGetPtr set;
} SendNotificationHook;
```

I'd use a `typedef`, it makes using the type easier.

src/plugins/zeromqsend/zeromqsend.c
```diff
@@ -83,11 +83,13 @@ int elektraZeroMqSendGet (Plugin * handle ELEKTRA_UNUSED, KeySet * returned, Key
 	if (!strcmp (keyName (parentKey), "system:/elektra/modules/zeromqsend"))
 	{
 		KeySet * contract = ksNew (
-			30, keyNew ("system:/elektra/modules/zeromqsend", KEY_VALUE, "zeromqsend plugin waits for your orders", KEY_END),
+			32, keyNew ("system:/elektra/modules/zeromqsend", KEY_VALUE, "zeromqsend plugin waits for your orders", KEY_END),
```
kodebach, line 86: It's nice that you updated the count, but you actually don't have to. The way `ksNew (N, ...)` works is that:

- If `N == 0`: The other arguments `...` are completely ignored and an empty keyset is created.
- For `N != 0`: All of the following are done
  1. It allocates space for at least `N` keys
  2. It allocates space for at least 16 keys
  3. It allocates space for all the keys in `...` and adds them all
  
So the actual size of the keyset will be `N == 0 ? 0 : max (16, N, [no of keys in ...])`. Basically, unless you want to reserve space for extra keys added later, you can just use `ksNew (1, ...)`.

kodebach (changes requested):
The general concept looks good, but I have some notes on the details

src/libs/elektra/hooks.c
atmaxinger: Yes, I agree. I made both optional, and issue a warning when none of them is set.

src/libs/elektra/hooks.c
atmaxinger: Yes, this isn't actually needed.

src/libs/elektra/kdb.c
kodebach: > I thought about this too, but we'd also need to map it using the parent keys used to get the keysets. Otherwise we would only have the last retrieved keyset stored, and there is no way of enforcing that the user of the API will only perform `kdbGet` after a `kdbSet` of the previous read keyset.

I'm not entirely sure what you mean here. But regardless wouldn't any such issue exist in exactly the same way in a plugin?

It is already mandatory to call `kdbGet` before `kdbSet` and the `kdbGet` call must use the same or more backends as `kdbSet` needs.

There is also the `KeySet * keys` in each of the `struct _BackendData` in `kdb->backends`. I'd have to check, but maybe these already are or at least could be `keyDup`'ed and maybe they aren't cleared at the end of `kdbGet`. Then we could use that data at the start of `kdbSet` (before all the `KeySet * keys` are overwritten).


---

> This mostly works, but now all keys in `setKs` are marked as `SYNC`. Probably because the keys are marked somewhere in the process of `backendsDivide`. I think this is because of the `keyDup` calls here:

Yeah, I forgot about that. `ksDeepDup` actually restores the sync flag after calling `keyDup`.

> Maybe we should set the SYNC flag only when it is there on the original key?

So yes, we have to do something similar. Instead of just `keyDup (k, KEY_CP_ALL)` we need:

```c
Key * d = keyDup (k, KEY_CP_ALL);
clear_bit (d->flags, KEY_FLAG_SYNC);
d->flags |= k->flags & KEY_FLAG_SYNC;
```

kodebach (changes requested):

src/libs/elektra/kdb.c
atmaxinger: Yes, the same problem persists in the plugin. What I mean is this:

```
Key * keyA = ...;
Key * keyB = ...;

KeySet * a = ...;
KeySet * b = ...;

kdbGet(kdb, a, keyA);
kdbGet(kdb, b, keyB);

// Change keys in a

kdbSet(kdb, a, keyA);
```
If we only store the last KeySet that has been retrieved, we would have stored `b`, but now we are updatding KeySet `a`.

kodebach (approved):
LGTM now, great work, IMO this could be merged now (unless you want to do something more).

The remaining discussion about using different `KeySet`s with one `KDB` can be moved to a separate issue, since it is a problem with the current solution (even on master) anyway.

src/libs/elektra/kdb.c
```diff
@@ -2465,6 +2489,21 @@ int kdbSet (KDB * handle, KeySet * ks, Key * parentKey)
 	// Step 12c: run postcommit phase
 	runSetPhase (backends, parentKey, KDB_SET_PHASE_POST_COMMIT, true, KDB_SET_FN_COMMIT);
 
+	SendNotificationHook * sendNotificationHook = handle->hooks.sendNotification;
+	while (sendNotificationHook != NULL)
+	{
+		if (sendNotificationHook->set != NULL)
+		{
+			// TODO (atmaxinger): Is setKs really the correct KeySet?
+			// Unfortunately setKs removes the SYNC flag, so plugins can not reliably detect changed keys ...
```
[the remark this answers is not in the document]
atmaxinger: Ah yes it is

atmaxinger:
No, I think this hook is done. The next thing I'll probably tackle would actually be the tracking of the changes, as this is what I also need for the record plugin.

src/libs/elektra/kdb.c
atmaxinger: I created issue #4514 for this discussion.

markus2330:
It is a bit difficult to review for me as I don't know the changes in new-backend enough. @kodebach its best when you decide if this brings us closer to merge new-backend. I'll directly review new-backend, then.

@atmaxinger can you add some docu about how the hooks work, when they are called etc.. Or is it a straightforward implementation of the decision?

kodebach:
> kodebach its best when you decide if this brings us closer to merge new-backend. I'll directly review new-backend, then.

IMO we can merge this PR now.

> can you add some docu about how the hooks work, when they are called etc..

See https://github.com/ElektraInitiative/libelektra/blob/new-backend/doc/dev/hooks.md

We need more docs on the individual hooks. But I think this can wait until we finalize the entire design (i.e. all planned hooks work they way we want). There is no point in adding docs that might change again soon, if there is no consumer of the docs.

For example, it would be really stupid to document the way `spec` is called right now, when it isn't clear yet whether we will change this before the next release.

atmaxinger:
I will add some documentation to the `doc/dev/hooks.md`.

atmaxinger:
@markus2330 @kodebach I added some documentation for all of the current hooks. We can merge this PR now.

kodebach:
@markus2330 I merged this PR, if you still want to review the changes do so in #4187.

Issue #4463, opened 2022-09-12
Milestones kdb-record


atmaxinger:
This issue is for tracking my milestones

- [x] M1 - 1.10.2022 - finalized use cases/begin of implementation. Global plugins should work in new-backend branch.
- [ ] M2 - 12.12.2022 - working implementation/begin of case studies. Submitted introduction, methodology and description of case studies for review.
- [ ] M3 - 6.2.2023 First Draft of Thesis

markus2330:
Please also write for M2 which parts of the thesis you want to submit, e.g. introduction and methodology.

About Implementation: I assume everything related to new-backend is M1, everything else is M2 and M3 is for left-overs?

atmaxinger:
Hmm yes it would probably make sense to have the Methodology, Introduction and the description of the Case Studies finished by M2.

Regarding the Implementaion: originally everything including new-backend was planned for M2, but if we want to merge it by October then yes M1 would make much more sense for reparing the current global plugins.

Yes, at M3 implementation should be completely finished, including all bugs that I found during the execution of the case studies.

markus2330:
Sounds good to me! :rocket:

PR #4554, opened 2022-10-13
[decision] Change Tracking

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/decisions/change_tracking.md
+++ b/doc/decisions/change_tracking.md
@@ -1,0 +1,156 @@
+# Change Tracking
+
+## Problem
+
+Currently, a range of Elektra plugins are implementing some sort of change tracking for configuration data.
+This includes, but is not limited to, the [internalnotification](/src/plugins/internalnotification/README.md) and [dbus](/src/plugins/dbus/README.md) plugins.
+In the near future, Elektra shall also be extended with session recording.
+
+KDB itself also has some rudimentary change tracking (via the `keyNeedsSync` flag) to determine whether `kdbSet` needs to actually do something.
+
+These competing change tracking implementations create multiple problems:
+
+1. pointless waste of resources, as data is duplicated in each plugin,
+2. multiplication of code, generating a maintenance burden.
+3. various subtle differences in change tracking behavior, e.g., `kdbSet` might write a config file but notification is not sent out
+4. the current approach to change tracking in plugins is fragile, which is outlined in [a separate decision about valid kdbGet/kdbSet sequences](https://github.com/ElektraInitiative/libelektra/pull/4574).
+
+For `KeySet` we need to track which of the keys:
+
+- have been removed
+- have been added
+
+For `Key` we need to track:
+
+- original value of the key
+- size of the original value (for binary keys)
+- metadata, which is a combination of the tracking of keysets and keys
+- tracking should only be done on the following namespaces:
+  - `system:/`
+  - `user:/`
+  - `dir:/`
+  - `meta:/`
+  - `spec:/`
+
+## Constraints
+
+Change tracking must:
+
+- be transparent for applications using the public Elektra API
+- be transparent to users changing configuration data
+- if overhead is not negligible: only do tracking as required, i.e., a plugin specifically requests it
+- have negligible overhead if disabled
+- not duplicate data for each plugin that wants change tracking
+- work with all allowed sequences of `kdbGet` and `kdbSet` as [per this decision](https://github.com/ElektraInitiative/libelektra/pull/4574)
+
+The library `libelektra-core` must be kept minimal.
+
+## Assumptions
+
+- It is possible to do change tracking with reasonable memory and computation overhead
+- It is possible to design a single change tracking API that is useful
+  for all existing and future plugins
+
+## Considered Alternatives
+
+### Alternative 1 - Per-parent key tracking withn `libelektra-kdb`
+
+Do the per-parent-key tracking within `libelektra-kdb`, within the `kdbGet` and `kdbSet` operations.
+
+Essentially, we'd have a 'giant' hashmap in the form of (parent key)->(KeySet) for all parent keys that were used for a `kdbGet` operation in the lifetime of the `kdb` instance.
+We also need to update it on `kdbSet` so that a future `kdbSet` operation without a `kdbGet` will also work.
+
+When change tracking is enabled, this one will have the most memory overhead, as we need to deep-dup every key we have read or stored.
+
+### Alternative 2 - Per-parent key tracking with meta keys
+
+Do the per-parent-key tracking within `libelektra-kdb`, but with meta keys.
+
+Essentially the same approach as above, but instead of deep-duping, we add the original value
+as a meta-key to every key. Not yet clear how we handle changes to meta-data then.
+
+### Alternative 3 - Change tracking within a separate plugin
+
+Outsource the change tracking into a separate plugin.
+
+Essentially the same as (1), just that it is implemented within a plugin and not `libelektra-kdb`.
+This will be a hook plugin, and will be called within `kdbGet` and `kdbSet` accordingly.
+It will also need to export a hook-method to get the changeset.
+
+The following hooks will be needed:
+
+- `tracking/get`: wil be called at the end of `kdbGet`, directly before the result is returned.
+- `tracking/set/preliminary`: will be called at the beginning of `kdbSet` and after every step/phase in `kdbSet`.
+  We need this to have change tracking information as soon as possible, so that plugins in any step of the process can use this information.  
+   Also, plugins in every step of the process might change data, so it is important to call it after every step.
+- `tracking/set/final`: will be called in the `post-commit` phase, after all changes have been written to disk, but before the `notification/send` hook.
+  Represents the final, real changes to the KDB.
+- `tracking/changeset`: compute the changeset for the requested parent key and return it.
+
+### Alternative 4 - Copy-on-write change tracking within `libelektra-core`
+
+The idea here is that we extend the `KeySet` and `Key` structs with additional fields.
+
+We need to extend `KeySet` with the following info:
+
+- What keys have been removed
+- What keys have been added
+- Whether tracking is enabled for this KeySet
+
+We need to extend `Key` with the following info:
+
+- Original value of the key
+- Size of the original value (for binary keys)
+- Whether tracking is enabled for this key
+
+The tracking itself would be done within the `ks*` and `key*` methods, after checking if it is enabled.
+It would also transparently work for meta-data, as meta-data itself is implemented as a keyset with keys.
+
+Downsides of this approach:
+
+- It adds functionality to `libelektra-core` which may violate the constraint above.
+  It may, however, be debatable whatever 'minimal' means in this context.
+- Adding fields to the structs causes a slight memory overhead, even with tracking turned off.
+  While negligible for `KeySet` due to the low amount of keysets in typical applications, it may be noticable for `Key`.
+  On a 64-bit system we'd add 16 bytes to it.
+  8 bytes for the pointer to the original value, 8 bytes for the size of the original value.
+  To put this in perspective, the current size of the `Key` struct is 64 bytes, so we'd add 25% overhead to an empty key.
+  However, this percentage will be much lower in a real-world application, as the usefulness of an empty key is very low.
+- Another downside here is that it is not so easy to determine what the "original" value is.
+  Some part of `libelektra-kdb` would need to mark the keys as original (after transformations etc.)
+
+### Alternative 5 - Use `backendData->keys` for change tracking
+
+Use the `backendData->keys` for change tracking
+
+We already store which keys have been returned by `kdbGet` for each backend within KDB.
+Currently, however, this is not a deep duplication, as we are returning the keys from `backendData->keys` directly.
+This means we can not detect changes to the values or meta-data of keys right now.
+We can, however, rely on this for detecting removed and added keys in its current form.
+
+If we don't want to deep-dup it, we'd need to do something different to detect which keys have been modified.
+One possibility would be to add meta-data within the `key` functions (e.g. `meta:/elektra/original`), but that would violate the `libelektra-core` must be minimal constraint.
+
+There is already [another decision](internal_cache.md) which discusses adding general copy-on-write semantics to Elektra.
+We could use this together with `backendData->keys` deep-duped to do memory efficient change tracking.
+
+## Decision
+
+## Rationale
+
+## Implications
+
+-
+-
+-
+
+## Related Decisions
+
+- [Valid kdbGet/kdbSet sequences](https://github.com/ElektraInitiative/libelektra/pull/4574).
+- [Internal cache](internal_cache.md)
+- []()
+
+## Notes
+
+- Issue [#4514](https://issues.libelektra.org/4514) uncovered a problem in the current change tracking approach
+- Issue [#4520](https://issues.libelektra.org/4520) already explored some of the considered alternatives
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -229,7 +229,7 @@ This section keeps you up-to-date with the multi-language support provided by El
 
 ## Scripts
 
-- <<TODO>>
+- Added [automatic spelling corrections](https://master.libelektra.org/scripts/sed) for changeset. _(Maximilian Irlinger @atmaxinger)_
 - <<TODO>>
 - <<TODO>>
 - <<TODO>>
@@ -289,6 +289,7 @@ This section keeps you up-to-date with the multi-language support provided by El
 - decided [decision process](https://www.libelektra.org/decisions/decision-process) _(Markus Raab)_
 - draft for [man pages](/doc/decisions/man_pages.md) _(Markus Raab)_
 - <<TODO>>
+- Add decision for [change tracking](/doc/decision/change_tracking.md) _(Maximilian Irlinger @atmaxinger)_
 - <<TODO>>
 - <<TODO>>
 - <<TODO>>
@@ -478,6 +479,7 @@ The website is generated from the repository, so all information about plugins, 
 We are currently working on following topics:
 
 - Session recording and better Ansible integration _(Maximilian Irlinger @atmaxinger)_
+- Change tracking. _(Maximilian Irlinger @atmaxinger)_
 - <<TODO>>
 - <<TODO>>
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -289,7 +289,7 @@ This section keeps you up-to-date with the multi-language support provided by El
 - decided [decision process](https://www.libelektra.org/decisions/decision-process) _(Markus Raab)_
 - draft for [man pages](/doc/decisions/man_pages.md) _(Markus Raab)_
 - <<TODO>>
-- Add decision for [change tracking](/doc/decision/change_tracking.md) _(Maximilian Irlinger @atmaxinger)_
+- Add decision for [change tracking](/doc/decisions/change_tracking.md) _(Maximilian Irlinger @atmaxinger)_
 - <<TODO>>
 - <<TODO>>
 - <<TODO>>
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -75,6 +75,7 @@ Before you write your first decision, read [Decision Process](decision_process.m
 - [Notifications](notifications.md)
 - [Header File Structure](header_file_structure.md)
 - [Man Pages](man_pages.md)
+- [Change Tracking](change_tracking.md)
 
 ## Delayed
 
--- a/scripts/sed
+++ b/scripts/sed
@@ -59,6 +59,8 @@ s/\<([Mm])eta[- ][Dd]ata\>/\1etadata/g
 s/\<([Nn])ame [Ss]pace\>/\1amespace/g
 s/\<([Pp])lug-[Ii]n(s)?\>/\1lugin\2/g
 
+s/\<([Cc])hange[- ][Ss]et\>/\1hangeset/g
+s/\<([Cc])hange[- ]([Tt])racking\>/\1hange \2racking/g
 
 # key-value pair
 s:\<key/value:key-value:g


kodebach (changes requested):
As stated before, I don't think `libelektra-core` should do the change tracking directly by storing the old value. It is simply not part of `libelektra-core`'s goal. Also, while it is not relevant for the use-case in `libelektra-kdb` (*), if we do do change tracking in `libelektra-core`, it should also include the key name. That would make the whole size overhead a lot worse. I say this, because from `libelektra-core`'s POV there is no reason why only values should be tracked, but not names.

IMO if we do this in `libelektra-core` it would be far better, if both `KeySet` and `Key` had a field `changeCallback`. This would be a function pointer, that is called when the `Key`/`KeySet` is changed. Then  `libelektra-kdb` could set this callback and keep an internal record of the changes. However, I don't think this really is a good solution either, because it is still beyond what `libelektra-core` should do.

(*) for `libelektra-core` tracking names is not needed, because we only deal with `KeySet`s and when a `Key` is in a `KeySet` it's name cannot change.

---

I think you discredit option 1 far too quickly. It is not only the easiest to implement, but it is also not nearly as bad as you make it out to be. There are already constraints on the `kdbGet` and `kdbSet` relation. The constraint right now is specifically:

For `kdbSet (handle, setKs, setParent)` to succeed, `kdbGet (handle, getKs, getParent)` must have been called previously and the relation between `getParent` and `setParent` must be (at least) one of:
1. `setParent` is the same as or below `getParent`
2. both `getParent` and `setParent` are stored in the same backend

AFAICT the only required change would be to say "the _last_ `kdbGet` call using `handle` must have been `kdbGet (handle, getKs, getParent)`" instead of "`kdbGet (handle, getKs, getParent)` must have been called previously". This could easily be checked and we could return an error if this is not the case.

Yes, it has an impact on the developer using Elektra. But it is not a huge impact. You just need to use a separate `KDB` handle. That too has some memory (and computation) overhead, but in general it should be less than the overhead of e.g. option 3.

In addition to that, I believe that it would be very rare that one of the problematic sequences was created intentionally. It is far more likely that it happend through sharing of a `KDB` handle between unrelated components or even threads (which is explicitly unsupported).

Furthermore, I think this might not even have any additional memory overhead. In the `KDB` handle's `KeySet * backends` we already have a `struct _BackendData` for every backend that was used. These contain a `KeySet * keys`, with the keys for the backend. Importantly, those `keys` are already `keyDup`'d. I think that data could be used to detect changes between `kdbGet` and `kdbSet`. 

---

Finally, when talking about memory overhead you also have to keep in mind that Elektra is global. If you start a change tracking session, this should affect all applications using Elektra. Therefore, each of these applications will have the overhead of change tracking.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+   For `Key` we need to track:
+
+   - Original value of the key
+   - Size of the original value (for binary keys)
+   - Whether tracking is enabled for this key
+
+   The tracking itself would be done within the `ks*` and `key*` methods, after checking if it is enabled.
+   It would also transparently work for meta-data, as meta-data itself is implemented as a keyset with keys.
+
+   Two downsides of this method have been identified by @kodebach:
```
markus2330, line 123: Acknowledgements only in Notes please.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+
+   - It adds functionality to `libelektra-core` which is supposed to be minimal
```
markus2330, line 125: Add a constraint above and say here this violates the constraint.

doc/news/_preparation_next_release.md
```diff
@@ -466,7 +466,7 @@ The website is generated from the repository, so all information about plugins,
 
 We are currently working on following topics:
 
-- <<TODO>>
+- Change tracking. _(Maximilian Irlinger @atmaxinger)_
```
markus2330, line 469: Please link to decision(s).

markus2330 (changes requested):
Great job, it is much more clear to me now!

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+
+Now, this problem could be overcome by also remembering the key for which the KeySet was read, but none of the plugins do that.
+Changing this leads us back to problem number 2, the maintenance burden.
+
+## Constraints
+
+Change tracking must:
+
+- be transparent to developers using the public Elektra API
+- be transparent to users changing configuration data
```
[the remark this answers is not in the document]
atmaxinger: No. Developers = people elektrifying applications
Users = Users of those elektrified applications of the `kdb` command line utility

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+- only be done if it is required, i.e. a plugin specifically requests it
```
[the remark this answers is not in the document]
atmaxinger: Yes, but we don't know that yet.

doc/decisions/change_tracking.md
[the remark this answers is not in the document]
markus2330: I probably need an example to understand what you are aiming at. My thought was that users indirectly also use the public Elektra API, so if it is transparent to the API it must be transparent to the user.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,146 @@
...
+- At point `(C)` they will generate a wrong changeset for key `a`
+
+
+## Constraints
+
+Change tracking must:
+
+- be transparent to developers using the public Elektra API
+- be transparent to users changing configuration data
+- only be done if it is required, i.e. a plugin specifically requests it
```
markus2330, line 48:
```suggestion
- if overhead is not negligible: only do tracking as required, i.e., a plugin specifically requests it
```

doc/decisions/change_tracking.md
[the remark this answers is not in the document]
atmaxinger: Yes, Users indirectly use the public API. In that sense we could remove this constraint. But I think it doesn't hurt to specify it.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+- not duplicate data for each plugin that wants change tracking
+- work with arbitrary sequences of `kdbGet` and `kdbSet`
+- not require any changes in the plugin API for plugins without change tracking
+
+## Assumptions
+
+- Some kind of memory and computation overhead is acceptable
+  as long as it doesn't scale linearly (or worse) with the number of plugins
```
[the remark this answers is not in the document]
atmaxinger: Do you have any suggestions for those constraints?

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+4. Do the per-parent-key tracking within `libelektra-kdb`, but with meta-keys.
+
+   Essentially the same approach as above, but instead of deep-duping, we add the original value
+   as a meta-key to every key. Not yet clear how we handle changes to meta-data then.
+
+5. Outsource the change tracking into a separate plugin.
+
+   Essentially the same as (3), just that it is implemented within a plugin and not
+   `libelektra-kdb`. This will be a hook plugin, and will be called within `kdbGet` and `kdbSet` accordingly.
+   It will also need to export a hook-method to get the changeset.
```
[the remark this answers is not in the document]
atmaxinger: Pushed an update with more details.

doc/decisions/change_tracking.md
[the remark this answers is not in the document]
kodebach: I think is not worth our time to discuss whether a constraint/assumption is redundant...

IMO this way makes things more explicit which is a good thing in a decision. There should be as few unmentioned implications as possible.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+2. multiplication of code, generating a maintenance burden.
+
+Additionally, it was uncovered in issue [#4514](https://issues.libelektra.org/4514),
+that the current approach to change tracking in plugins is very fragile.
+Basically, what plugins do currently is:
+
+1. on `kdbGet` duplicate the returned keyset,
+2. on `kdbSet` calculate what changed inbetween.
+
+This basic approach will fail when there is another `kdbGet` operation for another key inbetween.
```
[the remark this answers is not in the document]
kodebach: I think it's not just about the "allowed sequences", but also about the intended meaning of these sequences. For the case where "neither A below B nor B below A" I couldn't come up with a sensible intended meaning. That's why I originally chose this example. But for other relations there can be sensible meaning depending on the contents of the keysets, e.g. "overwrite all", "remove parts" etc.

doc/decisions/change_tracking.md
```diff
@@ -39,79 +38,86 @@ kdbSet(kdb, a, keyA); // (C)
...
+
+- What keys have been removed
+- What keys have been added
+
+For `Key` we need to track:
+
+- Original value of the key
+- Size of the original value (for binary keys)
+
+Tracking should only be done on the following namespaces:
```
kodebach, line 51: You forgot `spec:/`

atmaxinger:
@markus2330 @kodebach I have updated the decision and addressed (hopefully) all your remarks. 

In particular, I have removed the first two constraints-violating alternatives, and added more detailed info about the hooks-based approach.

It would be greatly appreciated if we come to a consensus which approach we'll decide on soon, so that I can start implementing it as quickly as possible 🚀.

kodebach:
> It would be greatly appreciated if we come to a consensus which approach we'll decide on soon, so that I can start implementing it as quickly as possible .

Of course, but I think we need to decide #4574 first. If the sequences are limited, this decision becomes much simpler.

doc/decisions/change_tracking.md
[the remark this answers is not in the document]
markus2330: It shouldn't get worse. Compensations with other improvements is possible.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,160 @@
+# Change Tracking
+
+## Problem
+
+Currently, a range of Elektra plugins are implementing some sort of change-tracking for configuration data.
+This includes, but is not limited to, the [internalnotification](../../src/plugins/internalnotification/README.md)
```
markus2330, line 6:
```suggestion
This includes, but is not limited to, the [internalnotification](/src/plugins/internalnotification/README.md)
```

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,160 @@
...
+and [dbus](../../src/plugins/dbus/README.md) plugins. In the near future, Elektra shall also be extended with session recording.
```
markus2330, line 7:
```suggestion
and [dbus](/src/plugins/dbus/README.md) plugins. In the near future, Elektra shall also be extended with session recording.
```

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,160 @@
...
+
+Having multiple implementations for the same problem is problematic in multiple ways:
```
markus2330, line 9:
```suggestion
These competing change-tracking implementations create two problems:
```

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,160 @@
...
+
+1. pointless waste of resources, as data is duplicated in each plugin,
+2. multiplication of code, generating a maintenance burden.
+
+The current approach to change tracking in plugins is very fragile.
```
markus2330, line 14: "very" is subjective and not very scientific :wink: 

```suggestion
Furthermore, the current approach to change tracking in plugins is fragile.
```

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,160 @@
...
+Basically, what plugins currently do is:
+
+1. on `kdbGet` duplicate the returned keyset,
+2. on `kdbSet` calculate what changed in between.
+
+This basic approach will fail when there is another `kdbGet` operation for a different part of the KDB in between.
```
markus2330, line 20: I stopped reading here, as this duplicates with #4574

markus2330 (changes requested):
There is duplication with #4574. Please only request reviews when all work was done. Or is this wrongly shown by GitHub?

atmaxinger:
@markus2330 of course it duplicates to some extend with #4574, as the "illegal" sequences mostly only affect plugins that do change tracking. Creating another decision for it was your idea.

The thing is, and this is where I strongly disagree with @kodebach, we can actually decide this independently from #4574, as all of the alternatives presented in here allow us to work around the "illegal" sequences issue. 

In the case of alternatives 1-3, we can just store all keysets that were returned by `kdbGet`. This is an implementation detail that doesn't even need to be in a decision. If we decide then later decide on #4574, we can update the implementation to something that uses resources more efficiently based on the boundaries we decide there.

In the case of alternative 4, the change tracking is completely done on KeySet level, so the sequences of `kdbGet` and `kdbSet` don't even matter.

atmaxinger:
@kodebach RE `_BackendData->keys`: yes we could use that! It stores a deep-duped copy of each part of the read keyset. However, it gets (obviously) cleared at every `kdbGet` and `kdbSet` operation. So the outcome of #4574 will greatly effect this if we go into that direction.

kodebach:
> However, it gets (obviously) cleared at every `kdbGet` and `kdbSet` operation

Actually cleared or, rather replaced and updated?

Unless I'm mistaken, `backendsDivide` is the only function that clears the data. But it also immediately replaces the data with the current data. If there is still a problem, we could also split `backendsDivide` into two functions. One for clearing the data and one for writing the new data. Then we could just clear the data when it actually needs to be cleared.

markus2330:
Having deeply duplicated keys of everything is out of question. We need to go **down** with memory consumption and not double it! So it goes against the constraint to not use more memory.

What I can totally imagine (and is probably a brilliant idea) is that we always use COW semantics from the keys that are passed out from our API (also in the non-mmap case, where we actually already do it that way) and keep the duplicate inside.

Then we could:

- implement change tracking
- only pass out the keys as requested (internally there might be more keys, depending on where the backends are, but we can ksBelow exactly as requested)

Btw. there is already a draft decision on that topic: doc/decisions/internal_cache.md

@atmaxinger can you take over this decision, too?

kodebach:
> Having deeply duplicated keys of everything is out of question. We need to go down with memory consumption and not double it!

You misunderstood, my idea... I suspect, we _already have_ a deepDup'ed copy of the keys that we can use. We could probably get rid of this, but AFAIK it was already present in the split of the old backend system.

> COW semantics

General copy-on-write semantics are **very** complicated and I don't believe we can do that without API changes. Since the `KeySet` is provided by the caller (and changing that seems out of the question) it would definitely need changes in `libelektra-core`. I'm not totally against that, but the core should remain minimal _and_ it should not know anything about `libelektra-kdb`. Doing this also means thinking about _a lot of edge cases_.

IMO the only option to do COW properly, is extending the `mmap` specific code in the core to be full COW write support that can be used by anyone.

Also @markus2330 since you seem to have a vague idea about how this should work, could you maybe write that up? To me it is currently entirely unclear what exactly you mean by "always use COW semantics from the keys that are passed out from our API".

atmaxinger:
@kodebach actually cleared, on mutliple occasions. Yes, `backendsDivide` is one place, but directly in `kdbGet` it is also cleard. However, as mentioned, this shouldn't be too much of a problem. We can still use this to compute the changes done by the user.

@markus2330 I don't see a way to do change tracking without using more memory. After all, we need to store original values of changed keys. However, as mentioned this will only be done if a plugin explicitly requests it. 

Having a central code that does the change tracking can actually **lower** the memory consumption to the current state. Currently, if you use both the `dbus` and the `internalnotification` plugin, both will do the change tracking and keep a copy of the keysets. If we do it centrally, there will only be a single copy.


> There is already a draft decision on that topic: doc/decisions/internal_cache.md

I read it, and IMHO it really isn't adding anything to this. This decision is very vague and I don't understand its purpose. Actually it very much reminds me of stuff we already do in new-backend, especially the "keep duplicated keyset internally". Maybe @kodebach can correct me on that.

Also, the considered alternatives are only bullet points without further description. As with @kodebach, I have no idea what you mean with "COW semantics". Could you be so kind and write it up? Does it differ much from what I have written in this decision?

kodebach:
How would you access the already duplicated data in a plugin? I think using `backendData->keys` is the best solution, but I think needs to be done mostly in `libelektra-kdb`.

You could pass a "current KDB state" `KeySet` to a special plugin in `kdbGet` and `kdbSet`, and then calculate the diff in the plugin. But the calculation of this "current state" from the `backendData->keys` has to happen in `libelektra-kdb`.

One advantage of this, would be that all hook plugins are implemented as a single (main) plugin and not a list of plugins. Where multiple plugins could be active (e.g. notifications) the main hook plugin would call extra plugins.

atmaxinger:
`backendData->keys` is just a KeySet. Change-tracking would be a seperate hook, and I can define that function to just take two keysets (current and the duplicated one), so no problem there.

As for other plugins, they will just use the change-tracking API that I will provide. This API will be general enough that we can replace the implementation of the change tracking to whatever we like later.

```c
bool elektraChangeTrackingIsEnabled(KDB *kdb);
KeySet * elektraChangeTrackingGetAddedKeys(KDB* kdb, Key * parentKey);
KeySet * elektraChangeTrackingGetRemovedKeys(KDB * kdb, Key * parentKey);
KeySet * elektraChangeTrackingGetModifiedKeys(KDB * kdb, Key * parentKey);

bool elektraChangeTrackingValueChanged(KDB * kdb, Key * key, Key * parentKey);
bool elektraChangeTrackingMetaChanged(KDB * kdb, Key * key, Key * parentKey);

KeySet * elektraChangeTrackingGetAddedMetaKeys(KDB * kdb, Key * key, Key * parentKey);
KeySet * elektraChangeTrackingGetRemovedMetaKeys(KDB * kdb, Key * key, Key * parentKey);
KeySet * elektraChangeTrackingGetModifiedMetaKeys(KDB * kdb, Key * key, Key * parentKey);

Key * elektraChangeTrackingGetOriginalKey(KDB * kdb, Key * key, Key * parentKey);
Key * elektraChangeTrackingGetOriginalMetaKey(KDB * kdb, Key * key, const char * metaName, Key * parentKey);
```

kodebach:
> `backendData->keys` is just a KeySet.

It's one KeySet _per backend_, they can be merged into a single KeySet to get the one from `kdbGet`/`kdbSet`, but only works when they are merged in the right order (so that nested backends work).

I think the merging should just be done in `libelektra-kdb` (because we already have `backdendsMerge` to do it) and then that data can be passed to a plugin. Maybe that is what you wanted to do anyway...

> As for other plugins, they will just use the change-tracking API that I will provide

`KDB *` is currently not available in plugins, although it could be made available. For `ChangeTrackingContext *` I have even less idea, where plugins would get that from.

atmaxinger:
> I think the merging should just be done in libelektra-kdb (because we already have backdendsMerge to do it) and then that data can be passed to a plugin. Maybe that is what you wanted to do anyway...

Exactly

> KDB * is currently not available in plugins, although it could be made available

Can't we just add it to `struct _Plugin`? And then add a function to get it for those plugins that need it.

> For ChangeTrackingContext * I have even less idea, where plugins would get that from.

From `KDB`.

atmaxinger:
@kodebach turns out while yes we do deep duplication in `backendsDivide`, we then actually return those deep-duped keys in `kdbGet` via `backendsMerge`. So as long as this isn't a bug or an oversight, we can not use those keys and have to keep our own deep duped data in the change tracking.

markus2330:
**Please Note:** all relevant discussions must be accounted for within the decision. I will not read your discussion to have a fair view on the decision, as someone else will have when later reading the decision (e.g. in the second round).

> @markus2330 I don't see a way to do change tracking without using more memory. After all, we need to store original values of changed keys. However, as mentioned this will only be done if a plugin explicitly requests it.

In particular it is important that `kdbGet` alone does not do any deep duplication, as it also didn't do before. Performance requirements for `kdbSet` are that actually writing the files should be the dominating factor. Ideally you do some profiling on the situation before new-backend and also now to have a good idea what is going on and what is actually expensive.

> @markus2330 @kodebach if we don't reach a decision here in short time, I will implement a prototype based on option Nr 3 - doing it in a seperate plugin while reusing the already duplicated data.

The progress will be faster if your PRs are in a better state. Please finish #4515 before creating new PRs. If you are really stuck with everything, please write in #4463.

kodebach:
> **Please Note:** all relevant discussions must be accounted for within the decision.

Yes, of course the document needs to be updated. But until now the discussion wasn't completed, so there was no point in updating the document.

_Now_ @atmaxinger should update the document and add using `backendData->keys` as an option. It is similar to option 1, but slightly different. The "hashmap" would exist, but the keys wouldn't be all the `parentKey`s used with `kdbGet`. Instead the keys would be all the mountpoints and the "hashmap" would be the already existing `KeySet * backends`. When change tracking is active, this data would be kept separate from the data returned by `kdbGet`.

> In particular it is important that `kdbGet` alone does not do any deep duplication, as it also didn't do before.

Any additional duplicating that happens, will of course only happen when change tracking is active. However, when it is active the duplication **must** happen during `kdbGet`, because it must happen _before_ the user has a chance to modify keys. This simply cannot be avoided.

atmaxinger:
> Please Note: all relevant discussions must be accounted for within the decision. I will not read your discussion to have a fair view on the decision, as someone else will have when later reading the decision (e.g. in the second round).

You don't need to. But those discussions are relevant for reaching a consensus and updating the document. How else would you or @kodebach be able to share your thoughts if not with these discussions? There is no point to update anything in the file if we don't even agree on what to update.

> In particular it is important that kdbGet alone does not do any deep duplication, as it also didn't do before

Yes, while the function `kdbGet` in the file `kdb.c` didn't do any duplication directly by itself, the plugins that were called by it, i.e. `internalnotification` and `dbus` did a duplication! In the case of `internalnotification` it's even a deep duplication of the requested keys. (One could argue that that's marginally better, because only requested keys are deep-duped. But that could still be added to the general change-tracker).

The only way we don't deep-dup is when using option 4 - the COW approach. But you didn't event comment on that, nor did you provide an answer to my question whether this was what you ment with your "COW semantics". That said, this approach also has downsides, as already mentioned in the decision file.

> Please finish https://github.com/ElektraInitiative/libelektra/pull/4515 before creating new PRs.

This is a completey different PR which can be worked on in parallel. There is no need for it to be merged while doing this. In fact, change tracking is more important for the general record use-case. #4515 only deals with usability improvents. Still very important for the finished product, but change-tracking is the base of it all.

doc/decisions/change_tracking.md
atmaxinger: I removed it and referenced #4574 instead.

markus2330:
> Could you be so kind and write it up?

#4619 

> I'd also like to mention that for me this decision is slowly drifting into [bikeshedding](https://www.techtarget.com/whatis/definition/Parkinsons-law-of-triviality-bikeshedding).

Absolutely, thus my insistence on a clear problem, constraints, assumptions etc. This was not the case in my last review so obviously any discussion on such grounds are questionable.

Please inform me when I should reread.

> Optimizations can then always be done later on.

No, this statement is plain wrong. E.g. read #4619. We needed to discard the whole implementation because it is a dead end. At that time we obviously didn't put enough efforts in finding the right decision. So please let us not repeat the same.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,149 @@
...
+
+   Two downsides of this method have been identified by @kodebach:
+
+   - It adds functionality to `libelektra-core` which is supposed to be minimal
+   - Adding fields to the structs causes a slight memory overhead, even with tracking turned off.
+     While neglectable for `KeySet` due to the low amount of keysets in typical applications,
+     it may be noticable for `Key`. On a 64-bit system we'd add 8+8=16 bytes to it.
+     To put this in perspective, the current size of the `Key` struct is 64 bytes,
+     so we'd add 25% overhead to an empty key. However, this percentage will be much lower in
+     a real-world application, as the usefulness of an empty key is very low.
```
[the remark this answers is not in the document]
markus2330: @atmaxinger can you please elaborate which 16 bytes you mean? Isn't a single pointer to "change-tracking data" enough? And maybe it even can be in meta-data, not needing any extra bytes in _Key.

doc/decisions/change_tracking.md
```diff
@@ -0,0 +1,147 @@
+# Change Tracking
+
+## Problem
+
+Currently, a range of Elektra plugins are implementing some sort of change-tracking for configuration data.
```
markus2330, line 5: change-tracking still has multiple spellings. If the word itself is written with `-`, then `-` always must be present (e.g. title above) and you can add it to scripts/sed.

markus2330 (changes requested):
A few small things, otherwise I think the problem is described clear enough to merge the PR as draft.

doc/decisions/change_tracking.md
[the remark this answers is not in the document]
kodebach: 8 bytes for the pointer, 8 bytes for the size.

I won't comment on the rest, because this clearly goes into too much detail. If the problem is now clear, this PR should be merged and this should be addressed in the next PR.

kodebach (approved):
The problem is clear, IMO this can merged. @markus2330 What do you say?

atmaxinger:
@markus2330 I have rebased to current master and addressed all your remarks.

markus2330:
Thank you, great job! I agree the problem is clearly presented now! :heart:

PR #4562, opened 2022-10-16
decisions: decide process, improve manpages

markus2330:
## Basics

- [x] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [x] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [ ] The PR is rebased with current master.

## Checklist

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation (see [Documentation Guidelines](https://master.libelektra.org/doc/contrib/documentation.md))
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned [every code](/.reuse/dep5) not directly written by me in [reuse syntax](https://reuse.software/)

## Review

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

- [ ] Add the "work in progress" label if you do not want the PR to be reviewed yet.
- [ ] Add the "ready to merge" label **if the basics are fulfilled** and no further pushes are planned by you.

PR #4574, opened 2022-10-18
[decisions] sequences of kdbGet and kdbSet operations

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -76,7 +76,7 @@ Before you write your first decision, read [Decision Process](decision_process.m
 - [Header File Structure](header_file_structure.md)
 - [Man Pages](man_pages.md)
 - [Change Tracking](change_tracking.md)
-- [Operation Sequnces](operation_sequences.md)
+- [Operation Sequences](operation_sequences.md)
 
 ## Delayed
 
--- a/doc/decisions/README.md
+++ b/doc/decisions/README.md
@@ -76,6 +76,7 @@ Before you write your first decision, read [Decision Process](decision_process.m
 - [Header File Structure](header_file_structure.md)
 - [Man Pages](man_pages.md)
 - [Change Tracking](change_tracking.md)
+- [Operation Sequnces](operation_sequences.md)
 
 ## Delayed
 
--- a/doc/decisions/operation_sequences.md
+++ b/doc/decisions/operation_sequences.md
@@ -227,7 +227,7 @@ As can be seen, the change tracking within the `dbus` and `logchange` plugins wr
    This way, we have only one such algorithm in a central place, and plugin authors don't have to think about the sequences their plugins are called by developers.
    
    This approach can also be paired with [COW semantics](internal_cache.md), so that memory toll will be kept low.
-   A separate [decision for change tracking](https://github.com/ElektraInitiative/libelektra/pull/4554) is currently in progress.
+   A separate [decision for change tracking](change_tracking.md) is currently in progress.
    
    Should we observe this problem with use cases other than change tracking, we can provide general frameworks for those too.
 
@@ -243,7 +243,7 @@ As can be seen, the change tracking within the `dbus` and `logchange` plugins wr
 
 ## Related Decisions
 
-- [Change Tracking](https://github.com/ElektraInitiative/libelektra/pull/4554)
+- [Change Tracking](change_tracking.md)
 - [Internal KeySet Cache](internal_cache.md)
 - []()
 
--- a/doc/decisions/operation_sequences.md
+++ b/doc/decisions/operation_sequences.md
@@ -1,0 +1,253 @@
+# Operation sequences
+
+## Problem
+
+The current public API allows various kind of sequences of `kdbGet` and `kdbSet` calls.
+There is no real guideline of allowed and/or prohibited sequences, other than `kdbGet` for a certain parent key must be called before `kdbSet`:
+
+- initially after `kdbOpen`
+- after conflict errors in `kdbSet`
+
+We discovered that some sequences are currently legal but problematic for change tracking and maybe also other use cases.
+
+Currently, notification plugins assume that a `kdbSet` operation for a certain parent key always directly follows the `kdbGet` operation for that same key.
+However, this is not enforced in any way within Elektra, nor is it documented that it should.
+
+In applications, `kdbGet` and `kdbSet` of different parent keys might be interwoven in any way.
+This is problematic, as plugins may store the result of `kdbGet` and use the stored data in the following `kdbSet` to calculate a changeset.
+Plugins have to do this, as there isn't any other Elektra-provided mechanism currently to do change tracking for them.
+
+Note, that erroneous behaviour only occurs if plugins are used globally, i.e. as `notification-send` hook plugins.
+If they are mounted for specific backends (via `kdb mount`), this behaviour will not occur as plugins from different backends are isolated.
+
+### Reproducible Example
+
+The following example will demonstrate the problematic sequence as real, runnable code.
+This example will show the erroneous behaviour for the `dbus` and `logchange` plugins, if used as `notification-send` hooks.
+These both plugins just serve as a general demonstration, and there may be more plugins that do change tracking this way.
+
+First, run some setup steps:
+
+```sh
+# enable the dbus plugin as a notification-send hook
+kdb set system:/elektra/hook/notification/send/plugins/#0 dbus
+
+# enable the logchange plugin as a notification-send hook
+kdb set system:/elektra/hook/notification/send/plugins/#1 logchange
+
+# mount user:/a and user:/b as two backends
+kdb mount a.ecf user:/a
+kdb mount b.ecf user:/b
+
+# populate user:/a and user:/b with values
+kdb set user:/a/brightness 1
+kdb set user:/a/saturation 1
+kdb set user:/b/foreground 1
+kdb set user:/b/background 1
+```
+
+Then, run the following command to monitor the notifications by the `dbus` plugins:
+
+```sh
+dbus-monitor --session type='signal',interface='org.libelektra'
+```
+
+Finally, execute the following program:
+
+```c
+#include <kdb.h>
+#include <stdio.h>
+
+int main (void)
+{
+	Key * root = keyNew ("/", KEY_END);
+	KDB * handle = kdbOpen (NULL, root);
+
+	Key * keyA = keyNew ("user:/a", KEY_END);
+	Key * keyB = keyNew ("user:/b", KEY_END);
+
+	KeySet * a = ksNew (0, KS_END);
+	KeySet * b = ksNew (0, KS_END);
+
+	kdbGet (handle, a, keyA);  // (A)
+	kdbGet (handle, b, keyB);  // (B)
+
+	// we need to change *something*, otherwise the backend would assume nothing changed
+	// and never even execute all of kdbSet
+	ksAppendKey (a, keyNew ("user:/a/test", KEY_VALUE, "hi", KEY_END));
+
+	kdbSet (handle, a, keyA);  // (C)
+
+	ksDel (a);
+	ksDel (b);
+	keyDel (keyA);
+	keyDel (keyB);
+	keyDel (root);
+
+	kdbClose (handle, 0);
+	return 0;
+}
+```
+
+- At point `(A)` the plugins internally store the data below the parent `keyA`.
+- At point `(B)` the same thing happens for the data below the parent `keyB`.
+  This **replaces** the data from `(A)`.
+- At point `(C)` plugins that do change tracking need to generate a changeset.
+  The changeset will be calculated between the stored data and the new data in the KeySet `a`.
+  Because the stored data is now from `(B)` and **not** from `(A)`, this changeset will be wrong.
+  The calculation will wrongly assume everything in KeySet `b` was removed and everything in KeySet `a` is newly added.
+
+Right after `(B)`, the two keysets are filled with the following values:
+
+| `a`                  | `b`                  |
+| -------------------- | -------------------- |
+| `user:/a/brightness` | `user:/b/background` |
+| `user:/a/saturation` | `user:/b/foreground` |
+
+If a plugin calculates a changeset during `kdbSet` by comparing to the `KeySet` from the last `kdbGet` (as described above), the plugin will wrongly calculate the following changeset at `(C)`:
+
+| Added Keys           | Removed Keys         | Modified Keys |
+|----------------------|----------------------| ------------- |
+| `user:/a/brightness` | `user:/b/background` |               |
+| `user:/a/saturation` | `user:/b/foreground` |               |
+| `user:/a/test`       |                      |               |
+
+This is obviously wrong. Looking at the example above, the changeset should only contain `user:/a/test`, as nothing else has been changed in keyset `a`.
+So the correct changeset should look like:
+
+| Added Keys     | Removed Keys | Modified Keys |
+|----------------| ------------ | ------------- |
+| `user:/a/test` |              |               |
+
+The erroneous behaviour can be noticed by the output of `dbus-monitor`:
+
+```
+signal time=1666993066.749997 sender=:1.364 -> destination=(null destination) serial=2 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
+   string "user:/a/brightness"
+signal time=1666993066.750161 sender=:1.364 -> destination=(null destination) serial=3 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
+   string "user:/a/saturation"
+signal time=1666993066.750551 sender=:1.364 -> destination=(null destination) serial=4 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
+   string "user:/a/test"
+signal time=1666993066.750628 sender=:1.364 -> destination=(null destination) serial=5 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyDeleted
+   string "user:/b/background"
+signal time=1666993066.750866 sender=:1.364 -> destination=(null destination) serial=6 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyDeleted
+   string "user:/b/foreground"
+```
+
+The same behaviour is present in the `logchange` plugin. Notice its output onto `stdout`:
+
+```
+added key: user:/a/brightness
+added key: user:/a/saturation
+added key: user:/a/test
+removed key: user:/b/background
+removed key: user:/b/foreground
+```
+
+As can be seen, the change tracking within the `dbus` and `logchange` plugins wrongly calculates everything below `user:/a` are new keys that have been added, and everything below `user:/b` was removed.
+
+## Constraints
+
+1. As far as possible, we must check for illegal sequences and raise an error when an illegal sequence occurs
+
+## Assumptions
+
+1.
+2.
+3.
+
+## Considered Alternatives
+
+1. Document which sequences are allowed without raising errors in all illegal cases.
+   Least-effort approach, but at least offers some transparency for developers.
+   Developers utilizing Elektra can still ignore it, and may cause problems in certain setups without knowing.
+
+   (Violates constraint 1)
+
+2. Enforce that the `parentKey` used in `kdbSet` is below the one used in the last `kdbGet`.
+   If this is not the case, `kdbSet` will abort and report an error in `parentKey`.
+   Developers might still wrongly mix the sequences, but they will get an error and have to fix it.
+
+   - Adding a `Key * lastGetParent` to `struct _KDB`
+   - Doing `keyCopy (handle->lastGetParent, parentKey, KEY_CP_NAME)` in `kdbGet`
+   - And checking `keyIsSameOrBelow (handle->lastGetParent, parentKey) == 1` in `kdbSet`
+
+   ```c
+   KDB * kdb = ...;
+   
+   Key * keyA = ...;
+   Key * keyB = ...; 
+   
+   KeySet * a = ...;
+   KeySet * b = ...;
+   
+   // These sequences will work fine
+   assert (kdbGet (kdb, a, keyA) == 0);
+   assert (kdbSet (kdb, a, keyA) == 0);
+   assert (kdbGet (kdb, b, keyB) == 0);
+   assert (kdbSet (kdb, b, keyB) == 0);
+   
+   // These will not
+   assert (kdbGet (kdb, a, keyA) == 0);
+   assert (kdbGet (kdb, b, keyB) == 0);
+   assert (kdbSet (kdb, a, keyA) == -1); // Error! keyA is different from last used key in kdbGet (keyB)
+   ```
+
+3. Enforce that `ks` used in `kdbSet` is the same as in the last `kdbGet`
+   Similar to (2), but we check for the pointer of the `KeySet` instead of having a copy of the parent key, which saves some memory.
+
+   - Adding a `KeySet * lastGetKs` to `struct _KDB`
+   - Doing `handle->lastGetKs = ks` in `kdbGet`
+   - And checking `lastGetKs == ks` in `kdbSet`
+
+4. Change the API so that a single instance of KDB can only contain a single instance of a `KeySet`.
+   This solution completely eliminates problematic sequences.
+
+   ```c
+   // same as before but uses handle->data instead of the ks argument
+   int kdbGet (KDB * handle, Key * parentKey);
+   // same as before but uses handle->data instead of the ks argument
+   int kdbSet (KDB * handle, Key * parentKey);
+
+   KeySet * kdbData (KDB * handle) {
+       return handle->data;
+   }
+   ```
+
+5. Allow arbitrary sequences and let each plugin deal with it on a case-by-case basis.
+   This alternative would put most of the burden onto the plugin authors.
+   Depending on what the plugins do, every plugin may also need to deep-dup every keyset of every parent it ever receives via `kdbGet`.
+   This will increase memory usage. 
+   However, this could be paired with the [COW semantics](internal_cache.md) so the memory toll would not be that big of a deal.
+   The biggest problem with this approach would be the unnecessary duplication of the non-trivial change tracking algorithm.
+
+6. Don't restrict sequences further and provide a common framework to handle change tracking correctly.
+   
+   As the problem has only been observed with plugins doing their own change tracking, we could provide a general change tracking framework within Elektra.
+   This way, we have only one such algorithm in a central place, and plugin authors don't have to think about the sequences their plugins are called by developers.
+   
+   This approach can also be paired with [COW semantics](internal_cache.md), so that memory toll will be kept low.
+   A separate [decision for change tracking](https://github.com/ElektraInitiative/libelektra/pull/4554) is currently in progress.
+   
+   Should we observe this problem with use cases other than change tracking, we can provide general frameworks for those too.
+
+## Decision
+
+## Rationale
+
+## Implications
+
+-
+-
+-
+
+## Related Decisions
+
+- [Change Tracking](https://github.com/ElektraInitiative/libelektra/pull/4554)
+- [Internal KeySet Cache](internal_cache.md)
+- []()
+
+## Notes
+
+- Issue [#4514](https://issues.libelektra.org/4514) uncovered a problem in the current change tracking approach relating to sequences of `kdbGet` and `kdbSet`
+- Issue [#4512](https://issues.libelektra.org/4520) explored some possible solutions
--- a/doc/decisions/operation_sequences.md
+++ b/doc/decisions/operation_sequences.md
@@ -107,7 +107,7 @@ Right after `(B)`, the two keysets are filled with the following values:
 If a plugin calculates a changeset during `kdbSet` by comparing to the `KeySet` from the last `kdbGet` (as described above), the plugin will wrongly calculate the following changeset at `(C)`:
 
 | Added Keys           | Removed Keys         | Modified Keys |
-|----------------------|----------------------| ------------- |
+| -------------------- | -------------------- | ------------- |
 | `user:/a/brightness` | `user:/b/background` |               |
 | `user:/a/saturation` | `user:/b/foreground` |               |
 | `user:/a/test`       |                      |               |
@@ -116,7 +116,7 @@ This is obviously wrong. Looking at the example above, the changeset should only
 So the correct changeset should look like:
 
 | Added Keys     | Removed Keys | Modified Keys |
-|----------------| ------------ | ------------- |
+| -------------- | ------------ | ------------- |
 | `user:/a/test` |              |               |
 
 The erroneous behaviour can be noticed by the output of `dbus-monitor`:
@@ -174,19 +174,19 @@ As can be seen, the change tracking within the `dbus` and `logchange` plugins wr
 
    ```c
    KDB * kdb = ...;
-   
+
    Key * keyA = ...;
-   Key * keyB = ...; 
-   
+   Key * keyB = ...;
+
    KeySet * a = ...;
    KeySet * b = ...;
-   
+
    // These sequences will work fine
    assert (kdbGet (kdb, a, keyA) == 0);
    assert (kdbSet (kdb, a, keyA) == 0);
    assert (kdbGet (kdb, b, keyB) == 0);
    assert (kdbSet (kdb, b, keyB) == 0);
-   
+
    // These will not
    assert (kdbGet (kdb, a, keyA) == 0);
    assert (kdbGet (kdb, b, keyB) == 0);
@@ -217,18 +217,18 @@ As can be seen, the change tracking within the `dbus` and `logchange` plugins wr
 5. Allow arbitrary sequences and let each plugin deal with it on a case-by-case basis.
    This alternative would put most of the burden onto the plugin authors.
    Depending on what the plugins do, every plugin may also need to deep-dup every keyset of every parent it ever receives via `kdbGet`.
-   This will increase memory usage. 
+   This will increase memory usage.
    However, this could be paired with the [COW semantics](internal_cache.md) so the memory toll would not be that big of a deal.
    The biggest problem with this approach would be the unnecessary duplication of the non-trivial change tracking algorithm.
 
 6. Don't restrict sequences further and provide a common framework to handle change tracking correctly.
-   
+
    As the problem has only been observed with plugins doing their own change tracking, we could provide a general change tracking framework within Elektra.
    This way, we have only one such algorithm in a central place, and plugin authors don't have to think about the sequences their plugins are called by developers.
-   
+
    This approach can also be paired with [COW semantics](internal_cache.md), so that memory toll will be kept low.
    A separate [decision for change tracking](change_tracking.md) is currently in progress.
-   
+
    Should we observe this problem with use cases other than change tracking, we can provide general frameworks for those too.
 
 ## Decision
--- a/doc/decisions/change_tracking.md
+++ b/doc/decisions/change_tracking.md
@@ -13,7 +13,7 @@ These competing change tracking implementations create multiple problems:
 1. pointless waste of resources, as data is duplicated in each plugin,
 2. multiplication of code, generating a maintenance burden.
 3. various subtle differences in change tracking behavior, e.g., `kdbSet` might write a config file but notification is not sent out
-4. the current approach to change tracking in plugins is fragile, which is outlined in [a separate decision about valid kdbGet/kdbSet sequences](https://github.com/ElektraInitiative/libelektra/pull/4574).
+4. the current approach to change tracking in plugins is fragile, which is outlined in [a separate decision about valid kdbGet/kdbSet sequences](operation_sequences.md).
 
 For `KeySet` we need to track which of the keys:
 
@@ -41,7 +41,7 @@ Change tracking must:
 - if overhead is not negligible: only do tracking as required, i.e., a plugin specifically requests it
 - have negligible overhead if disabled
 - not duplicate data for each plugin that wants change tracking
-- work with all allowed sequences of `kdbGet` and `kdbSet` as [per this decision](https://github.com/ElektraInitiative/libelektra/pull/4574)
+- work with all allowed sequences of `kdbGet` and `kdbSet` as [per this decision](operation_sequences.md)
 
 The library `libelektra-core` must be kept minimal.
 
@@ -146,7 +146,7 @@ We could use this together with `backendData->keys` deep-duped to do memory effi
 
 ## Related Decisions
 
-- [Valid kdbGet/kdbSet sequences](https://github.com/ElektraInitiative/libelektra/pull/4574).
+- [Valid kdbGet/kdbSet sequences](operation_sequences.md).
 - [Internal cache](internal_cache.md)
 - []()
 
--- a/doc/decisions/internal_cache.md
+++ b/doc/decisions/internal_cache.md
@@ -148,7 +148,7 @@ ksRemoveByName (cowMeta, "meta:/type");
 
 @kodebach wrote:
 
-> Change the API and remove KeySet from kdbGet and kdbSet also option 4 in [#4574](https://pull.libelektra.org/4574).
+> Change the API and remove KeySet from kdbGet and kdbSet also option 4 in [the operation sequences decision](operation_sequences.md).
 > If the keyset is owned by the KDB handle, it should not be as big surprise, if there is extra data in there.
 > I certainly wouldn't try to asset anything on the contents of a KeySet that I don't own directly, unless the condition is explicitly documented somewhere.
 
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -291,6 +291,8 @@ This section keeps you up-to-date with the multi-language support provided by El
 - <<TODO>>
 - Add decision for [change tracking](/doc/decisions/change_tracking.md) _(Maximilian Irlinger @atmaxinger)_
 - <<TODO>>
+- Create [decision](../decisions/operation_sequences.md) for allowed and prohibited operation seqences _(Maximilian Irlinger @atmaxinger)_
+- <<TODO>>
 - <<TODO>>
 - <<TODO>>
 - <<TODO>>


doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,94 @@
+# KDB operation sequences
+
+## Problem
+
+The current public API allows various kind of sequences of `kdbGet` and `kdbSet` calls.
+There is no real guideline of allowed and/or prohibited sequences, other than `kdbGet` for a certain parent key must be called before `kdbSet` on the same key.
```
kodebach, line 6: Specifically, the current constraints are

https://github.com/ElektraInitiative/libelektra/blob/d1a8f828fea2ad77c612df62009764beb397d273/src/libs/elektra/kdb.c#L2194-L2196

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Assumptions
+
+1.
+2.
+3.
+
+## Considered Alternatives
+
+1. Document which sequences are allowed.
```
kodebach, line 42:
```suggestion
1. Document which sequences are allowed without raising errors in all illegal cases.
```

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,94 @@
...
+- At point `(B)` the plugins internally override the stored the KeySet with the KeySet for key `b`
+- At point `(C)` they will generate a wrong changeset for key `a`
+
+
+
+## Constraints
+
+1.
+2.
+3.
```
kodebach, line 32:
```suggestion
- As far as possible, we must check for illegal sequences and raise an error when a constraint is violated
```

kodebach (commented):
As stated previously in https://github.com/ElektraInitiative/libelektra/pull/4554#discussion_r998096541, I think this decision must also define the meaning for all allowed sequences. Basically, we just need to define a call `kdbSet(..., ks, parent)` **always** means _replace everything_ below `parent` with the contents of `ks` (below `parent`), modulo mountpoints (*). We use a more complex definition, but this covers the current situation and is pretty clear.

(*) To be technically accurate have to take mountpoints into account. Only where a mountpoint exists will data be replaced, everything that falls outside a mountpoint is ignored or causes an error.

---

IMO option 2 or 3 is the best in terms of effort-to-benefit ratio. Option 1 is easy, but also puts a lot of burden on developers using Elektra and allows easy mistakes. Option 4 might be the best overall IMO, but it is a lot of effort to implement.

atmaxinger:
@markus2330 what are your thoughts about this issue?

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+// assume that neither A below B nor B below A
+KeySet * a = ...;
+KeySet * b = ...;
+kdbGet(kdb, a, keyA); // (A)
+kdbGet(kdb, b, keyB); // (B)
+kdbSet(kdb, a, keyA); // (C)
+```
+
+- At point `(A)` the plugins internally store the KeySet for key `a`
+- At point `(B)` the plugins internally override the stored the KeySet with the KeySet for key `b`
```
markus2330, line 26: Sentence is broken.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+- At point `(C)` they will generate a wrong changeset for key `a`
```
markus2330, line 27: "they"? The plugins? Can you please give an example how it would be wrong.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Constraints
+
+- As far as possible, we must check for illegal sequences and raise an error when a constraint is violated
```
markus2330, line 31: Please stick to terminology, otherwise the sentences are becoming confusing. Give constraints a number so that you refer to it.

```suggestion
1. As far as possible, we must check for illegal sequences and raise an error when an illegal sequence occurs
```

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
+# KDB operation sequences
```
markus2330, line 1: Should be same as file. I think `Illegal Sequences` are a better name.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Problem
+
+The current public API allows various kind of sequences of `kdbGet` and `kdbSet` calls.
+There is no real guideline of allowed and/or prohibited sequences, other than `kdbGet` for a certain parent key must be called before `kdbSet`:
+
+- initially after `kdbOpen`
+- after conflict errors in `kdbSet`
+
+While researching ways to do change tracking, we discovered that some sequences are currently legal, but might be problematic.
```
markus2330, line 11: Comma before but usually only when there are two full sentences connected.

```suggestion
While researching ways to do change tracking, we discovered that some sequences are currently legal but might be problematic.
```

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+
+1.
+2.
+3.
+
+## Considered Alternatives
+
+1. Document which sequences are allowed without raising errors in all illegal cases.
+   Least-effort approach, but at least offers some transparency for developers.
+   Developers utilizing Elektra can still ignore it, and may cause problems in certain setups without knowing.
```
markus2330, line 43:
```suggestion
   Developers utilizing Elektra can still ignore it, and may cause problems in certain setups without knowing.
   (Violates Constraint 1)
```

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+
+2. Enforce that the `parentKey` used in `kdbSet` is below the one used in the last `kdbGet`.
+   If this is not the case, `kdbSet` will abort and report an error in `parentKey`.
+   Developers might still wrongly mix the sequences, but they will get an error and have to fix it.
+
+   - Adding a `Key * lastGetParent` to `struct _KDB`
```
markus2330, line 49: Don't like this.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+   - Doing `keyCopy (handle->lastGetParent, parentKey, KEY_CP_NAME)` in `kdbGet`
+   - And checking `keyIsSameOrBelow (handle->lastGetParent, parentKey) == 1` in `kdbSet`
+
+3. Enforce that `ks` used in `kdbSet` is the same as in the last `kdbGet`
```
markus2330, line 53: This ties a KDB to one ks, which is imho a too strong coupling.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,92 @@
...
+   // same as before but uses handle->data instead of the ks argument
+   int kdbGet (KDB * handle, Key * parentKey);
+   // same as before but uses handle->data instead of the ks argument
+   int kdbSet (KDB * handle, Key * parentKey);
+
+   KeySet * kdbData (KDB * handle) {
+       return handle->data;
+   }
+   ```
+
```
markus2330, line 72: IIRC there was also some idea of storing each previous parent key somewhere, so that we neither have to change the API nor have more illegal sequences as we currently have. Did I misunderstand?

markus2330 (changes requested):
Sorry, I still don't fully understand the problem. There were some inconsistencies with terminology, which distracted me. And an example would be very valuable (of concrete KeySets (a), (b) and (c) and the changesets). Please also specifically explain why the change set is calculated from the KeySet as passed from the user and not as received from the plugin.

atmaxinger:
@markus2330 the gist of the problem is, that plugins may (and in the case of notification plugins DO) assume that a `kdbSet` operation for a certain parent key always follows the `kdbGet` operation. However this is not enforced in any way within Elektra, nor is it documented that it should.

doc/decisions/operation_sequences.md
atmaxinger: Yes, this is a possibility. But we need to check every plugin whether it implicitly assumes that `kdbSet` always follows `kdbGet` for the same parent key. If it does, we then need to add this 'workaround'.

markus2330:
Btw. no need for explanations in the comments. Please rewrite and I will reread everything.

doc/decisions/operation_sequences.md
atmaxinger: It's not. This decision is not pureley about illegal sequences, but as @kodebach mentiones also about the semantical meaning of allowed sequences.

doc/decisions/operation_sequences.md
atmaxinger: I've now made this alternative 5.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,101 @@
...
+5. Allow arbitrary sequences and let each plugin deal with it on a case-by-case basis.
```
kodebach, line 80: This description make this option sound far too easy. Please explain more details (deep-dup per parent, in every plugin, etc.). You should also mention that we could keep the deepDup'ed data in `KDB` instead of separately in every plugin and provide an API for plugins to access this.

doc/decisions/operation_sequences.md
kodebach: > This ties a KDB to one ks, which is imho a too strong coupling.

Can you give more details on this opinion? IMO tied a `KDB` to exactly one `KeySet *` (via option 4) make things much simpler for everybody and makes the API very clear to users. The data is owned by the `KDB` instance, so you shouldn't expect anything that `KDB` doesn't tell you.

Are there specific use cases where you using a separate `KDB` per `KeySet` is not possible?

kodebach (commented):
@atmaxinger Can you please check, whether using `KeySet * keys` in the `struct _BackendData` values of the `KeySet * backends` in `struct _KDB` is possible?

I think before the first `backendsDivide` in the keys in those keysets should still be a deepdup copy of the data from the last `kdbGet` (divided by backend). If I remember correctly, those keys are replaced during `kdbGet` and `kdbSet`, but only below the `parentKey`. So (if I'm correct) in the sequence

```c
Key * keyA = ...;
Key * keyB = ...;
// assume that neither A below B nor B below A
KeySet * a = ...;
KeySet * b = ...;
kdbGet(kdb, a, keyA); // (A)
kdbGet(kdb, b, keyB); // (B)
kdbSet(kdb, a, keyA); // (C)
```

At (C) we should see the keys loaded by (A) and (B), with (B) overwriting any overlapping parts. If that's the case, we can use the data from all the `KeySet * keys` (with a `backendsMerge`) before the first `backendsDivide` to know what was read during `kdbGet`.

doc/decisions/operation_sequences.md
markus2330: Of course it would be always possible if performance and memory does not matter. One problem is that kdbOpen is expensive.

The other problem is: It is quite often the case to have two different configs (or more), e.g., one which might get written (could be called state) and another one only read at startup. I think it is a nice design if these different configs are actually in different KeySets.

doc/decisions/operation_sequences.md
markus2330: I don't like that we get too much state/too many illegal sequences. To e.g. force the same parent key or keyset breaks the whole idea of separating these 3 classes.

markus2330:
> You've asked this question now multiple times. I since I tried to give some explanations previously, I don't think I understand the question. @markus2330 Can you maybe explain what you want to know here? Of course the change-set must be based on the data passed to kdbSet where else would the data come from?

Yes, obviously the changeset calucation algorithm uses the data from the user in the case of kdbSet. My question is of which relevance the KeySet passed from the user in the case of kdbGet is? Isn't only data as emitted from the storage plugin of relevance? This should be clearly explained in the problem.

doc/decisions/operation_sequences.md
markus2330: I am not sure if this is a good idea. If there is also unclear semantics, this should be in separate decisions. Are these semantics also of relevance for #4554? It is important that we find out the connections between the decisions, so that we solve them in the right order.

Btw. you are doing a great job here. This is really an operation in the heart of Elektra. :heart_decoration:

kodebach:
> My question is of which relevance the KeySet passed from the user in the case of kdbGet is? Isn't only data as emitted from the storage plugin of relevance?

The data _passed into_ `kdbGet` is not relevant, that's correct.

But from "_the_ storage plugin" is also wrong, because in all likelihood there are multiple backends involved and the data may change in `poststorage`.

However, we do need to know the data that was returned by the `kdbGet` corresponding to the current `kdbSet`. So in the example, during (C) we need to know what data was returned by (A). At least that is how all the existing notification plugins operate (with the assumption that (B) never exists).

Of course we could define differently, how the diff is calculated (e.g. between `storage` phases instead of between "end of get" and "start of set"). But this wouldn't change the fact that in the example (B) can mess with the data in the `KDB` handle and we need to deal with that somehow. That's what this decision is about, and that is why I brought up that we also need to define semantics. Some modifications to `KDB` may result in acceptable semantics (e.g. keys loaded by (A) and (B) use the newer state from (B) as "old value" in the diff), while others are just broken (e.g. anything not loaded at (B) is considered as "newly added" during (C)).

doc/decisions/operation_sequences.md
kodebach: See my recent response in the root thread of this PR. The "semantics" don't refer to the semantics of `kdbGet`/`kdbSet` on their own, but purely to the semantics of sequences, i.e. how they interact when used on the same handle.

doc/decisions/operation_sequences.md
kodebach: It is not too bad. If you always do a `kdbGet`, then modify the keyset and then do `kdbSet` it works fine. In fact reducing the time between `kdbGet` and `kdbSet` is good for avoiding conflicts anyway. There just can't be an unrelated `kdbGet` in between.

However, I agree, if there is a better solution, we should avoid additional restrictions.

atmaxinger:
As this seems to get nowhere, I just openly propose to do option Nr 1 - document that it isn't allowed to do `kdbSet` with a different parent key than that of the previous `kdbGet`. It seems like this hasn't been an issue until now, so we shouldn't waste any more time with this.

markus2330:
About a week ago I said that the problem in both decisions are still unclear. I am still waiting for you to correct it and inform me when it is done.

> It seems like this hasn't been an issue until now, so we shouldn't waste any more time with this.

We didn't have change tracking up to now, which is a big difference. Please investigate carefully of what internal and external notification does, what the differences to your proposed change tracking are, how a single algorithm could serve as basis for all three and which consequences this has on various parts of the framework.

Right decisions in the framework is the center of your thesis. I'll definitely not agree to any rash decisions here. The process is quite short (two iteration rounds) but it obviously depends on clearly written decisions, which simply was not the case when I read this decision the last time.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,101 @@
...
+KeySet * a = ...;
+KeySet * b = ...;
+kdbGet(kdb, a, keyA); // (A)
+kdbGet(kdb, b, keyB); // (B)
+kdbSet(kdb, a, keyA); // (C)
+```
+
+- At point `(A)` the plugins internally store the KeySet for key `a`
+- At point `(B)` the plugins internally override the stored KeySet with the KeySet for key `b`
+- At point `(C)` plugins that do change-tracking will generate a wrong changeset for key `a` if they assume that the previous `kdbGet` operation was also for key `a`
```
kodebach, line 32:
```suggestion
- At point `(A)` the plugins internally store the data below the parent `keyA`
- At point `(B)` the same thing happens for the data below the parent `keyB`.
  This **replaces** the data from `(A)`.
- At point `(C)` plugins that do change-tracking need to generate a changeset.
  The changeset will be calculated between the stored data and the new data in the KeySet `a`.
   Because the stored data is now from `(B)` and **not** from `(A)`, this changeset will be wrong.
   The calculation will wrongly assume everything in KeySet `b` was removed and everything in KeySet `a` is newly added.
```

markus2330:
I wrote quite a few specific points where I want clarification. As said in https://github.com/ElektraInitiative/libelektra/pull/4554#issuecomment-1292437985 I need to refuse to read any explanations, otherwise the decisions will not get into a state where it is understandable by anyone except the ones who participated in the discussions. We want to avoid documentation where external information is needed to understand what is written.

If it is unclear to @atmaxinger what to correct, it makes more sense if he asks me directly.

kodebach:
> If it is unclear to atmaxinger what to correct, it makes more sense if he asks me directly.

True, but I had to chime in, because I've had pretty much the same issue with you in the past.

> I wrote quite a few specific points where I want clarification.

AFAICT we addressed everything you asked for. Not in the decision file. That's true. But like I said, it would be really helpful, if you could just read the comments and say "Yes, that explains it. That needs to be in the decision file" or "No, it's still unclear".

In particular, I think you should say whether [this suggestion](https://github.com/ElektraInitiative/libelektra/pull/4574#discussion_r1005608186) is good enough, or you need more.

> As said in https://github.com/ElektraInitiative/libelektra/pull/4554#issuecomment-1292437985 I need to refuse to read any explanations, otherwise the decisions will not get into a state where it is understandable by anyone except the ones who participated in the discussions. We want to avoid documentation where external information is needed to understand what is written.

To use a positive example from the past as illustration: In https://github.com/ElektraInitiative/libelektra/pull/4246#discussion_r1002442705 I suggested a new sentence to clear up some confusion. You said, "yep, that's good" and I knew and I can just add that information to the file.

atmaxinger:
@markus2330 I think the major issue in this decision is, that neither I nor @kodebach know what is unclear for you. For me, it is clearly stated that currently there can occur problematic sequences of `kdbGet` and `kdbSet`, that will break plugins right now.

We have also provided a good, minimal example showcasing such a problematic sequence, and added a text that explains in detail what goes wrong, where and why. From my POV we can't make it anymore obvious.

> I am still waiting for you to correct it and inform me when it is done.

The only thing I could find where you had a specific question was the textual explaination of the example. And yes, that has been updated.

For your other comments to the alternatives ... "Don't like this" is a fine and valid comment, but I have no clue on how to address this? It's a considered alternative, so we put that in there. But there is really nothing I can do to address "Don't like this".

> We didn't have change tracking up to now, which is a big difference. 

Yes we did. It was implemented multiple times in the `internalnotification` and `dbus` plugins.

> Please investigate carefully of what internal and external notification does, what the differences to your proposed change tracking are, how a single algorithm could serve as basis for all three and which consequences this has on various parts of the framework.

That's what I'm doing currently. I even replaced the current change-tracking implementations in `internalnotification` and `dbus` and it seems to work fine so far. But I still need to do some more testing.

doc/decisions/operation_sequences.md
atmaxinger: > It is quite often the case to have two different configs (or more), e.g., one which might get written (could be called state) and another one only read at startup. I think it is a nice design if these different configs are actually in different KeySets.

There is nothing preventing you from doing this in this alternative. You only have to call `kdbGet` for your state immediately (*)  before you call `kdbSet` to write it, that's all.

I think you both are mixing this up with alternative 4 .... (you commented on alternative 3)

(*) by immediately I mean that you can not do a `kdbGet` for another part of the KDB in the meantime

kodebach:
> This is the nature of something being unclear. As long as I don't understand what you understand, I could only give interpretations of what you want to say. This way you would learn what I think, and not how to think/speak more clearly yourself. So I don't do it.

This is absolute and utter bullshit. Sorry to be so blunt, but it's true...

**We don't know** what is unclear to you, so we can only guess and add random extra bits of information that may or may not help. Worst case the extra info confuses even more and adding it was a massive waste of time.

**You**, however, can tell us **much more** than you are doing.

- Almost certainly, you have understood at least some parts of the document. You can at least tell us, which parts are clear/make sense to you and which are not clear/need more explanation.
- Even if you think, you haven't understood enough to make the distinction between good and bad parts, you can still say what you currently think the document says. Maybe you understood everything correctly and there is something missing. Maybe there was a misunderstanding and information needs to be corrected. These are two very different problems, and we simply cannot know which one is the case, if you refuse to be helpful.
- Finally, in the very unlikely case that you got literally zero useful information from the document, at least state that explicitly. Then clearly the entire thing has to be rewritten.

"unclear" is just a far too vague word to help anyone. What would you do, if I just told you `doc/dev/algorithm.md` is unclear to me, but don't tell you anything else?

Also, your idea, of "I don't tell you what I understood, because then you'll just learn to explain things to me and not be more clear in general" sounds good, but is bullshit. First of all, there are already two people (atmaxinger and I) who understand the current text. So unless there somebody else reads the text and also doesn't understand it, _you_ are the one we need to cater to. Therefore, we actually need to know how you think, because clearly it's different from how we think. Second, we need to know what the readers (including you) interpreted from the text, to learn what we need to improve. Finally, the idea is basically the same as a compiler printing only "Syntax Error" without any context and expecting the programmer to fix it.

> I don't require you to proof the absence of problematic sequences. I mainly requested a complete example of what exactly is broken in which situations. This shouldn't be too difficult. It is like writing a bug report. And I asked that this gets updated in the PR, not said in discussions. I assumed that you will inform me when it is done.

There is already one example in the document that shows a problematic sequence. This was always there. Since your last review, it has been expanded a bit. **That** was the suggestion (https://github.com/ElektraInitiative/libelektra/pull/4574#discussion_r1005608186) I made and you refused to comment on. @atmaxinger has merged this suggestion since then and even told you in his last comment about it:

> We have also provided a good, minimal example showcasing such a problematic sequence, and added a text that explains in detail what goes wrong, where and why. From my POV we can't make it anymore obvious.

Honestly, I'm starting to believe you want to make us suffer and intentionally only read half of the comments.

> I should have stopped reading after the unclear problem.

**NO**, oh god no. Stopping to read doesn't help at all. Reading further is a good thing. Maybe the order of the text is just a bit wrong and there things get clearer when you read further. And if you do stop at some point. At least tell us, where you stopped so we know the problem is before that point.

atmaxinger:
@markus2330 I think we have now clarified the example a bit better. I have also illustrated it with a new example. I have also added a short text to explain WHY plugins may choose to store the result of `kdbGet`.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,143 @@
...
+
+`keyA` will be `user:/a`, `keyB` will be `user:/b`.
+Right after `(B)`, the two keysets are filled with the following values:
+
+| `a`                  | `b`                  |
+|----------------------|----------------------|
+| `user:/a/brightness` | `user:/b/background` |
+| `user:/a/saturation` | `user:/b/foreground` |
+
+For a plugin that does change-tracking as described above, the plugin will then wrongly calculate the following changeset at `(C)`:
```
markus2330, line 61: > change-tracking as described above

There is not really a change tracking described above.

Please always use same spelling for "change tracking".

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,143 @@
...
+
+| Added Keys           | Removed Keys         | Modified Keys |
+|----------------------|----------------------|---------------|
+| `user:/a/brightness` | `user:/b/background` |               |
+| `user:/a/saturation` | `user:/b/foreground` |               |
+
+This is obviously wrong. Looking at the example, the changeset should be empty, as nothing has been changed in keyset `a`.
```
markus2330, line 68: Did you actually test this, e.g. with:

```
kdb mount a.ecf user:/a logchange
kdb mount b.ecf user:/b logchange
kdb set user:/a/brightness 1
kdb set user:/a/saturation 1
kdb set user:/b/background 1
kdb set user:/b/foreground 1
```

Or do you mean with other change tracking? Please be more specific and say what you actually checked.

markus2330 (changes requested):
We are slowly coming to an understandable problem. Several points, and in particular also scope are still unclear.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,143 @@
...
+
+Currently, plugins may (and in the case of notification plugins actually do) assume that a `kdbSet` operation for a certain parent key always directly follows the `kdbGet` operation for that same key. 
+However, this is not enforced in any way within Elektra, nor is it documented that it should.
+
+So in reality, as illustrated by the example, `kdbGet` and `kdbSet` of different parent keys might be interwoven.
+
+This is problematic, as plugins may store the result of `kdbGet` and use the stored data in the following `kdbSet` to calculate a changeset.
+Plugins have to do this, as there isn't any other Elektra-provided mechanism currently to do change-tracking for them.
+
+This problem is illustrated by the following example:
```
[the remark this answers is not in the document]
kodebach: So you want a concrete examples of for `keyA`/`keyB`, the contents of `a`/`b` and the existing mountpoints?

doc/decisions/operation_sequences.md
kodebach: > e.g. with: [...]

This sequence would NOT be the sequence given above. Each `kdb set` is a `kdbGet` immediately followed by a `kdbSet` with the same KeySet and `parentKey`. That is the one case that always works without issues.

However, I can tell we don't need to test it. This is just a logical conclusion from what is stated above.

1. In line 56-59 we defined the contents of `a` and `b`
2. The last `kdbGet` at (B) used `b`, so that is "theirs"
3. The the current `kdbSet` at (C) uses `a`, so that is "ours"
4. Since "theirs" and "ours" are distinct, the changeset is just:
    - remove everything from "theirs"
    - add everything from "ours"

That's obviously wrong, because "theirs" is simply the wrong data. It should be the contents of `a` at (A), which would give the correct empty changeset.


All this follows from one single fact: Plugins store data internally that is entirely replaced during each `kdbGet` and then compared against during `kdbSet`.

atmaxinger: @markus2330 I have added a reproducible example

doc/decisions/operation_sequences.md
atmaxinger: Yes this was meant.

atmaxinger:
@markus2330 I have now updated the decision with all your remarks. I also included real, runnable code (including setup setps) that exaclty reproduce the problem as described.

markus2330 (commented):
One specific problem is now clear but the presentation of the problem still could be better. Scope is still not clear, why is it not specific to dbus, what would need to be changed that it would work with dbus (as @kodebach requested something like this for 5.) and how would a framework, which implements this fix, look like?

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,223 @@
...
+signal time=1666993066.749997 sender=:1.364 -> destination=(null destination) serial=2 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
+   string "user:/a/brightness"
+signal time=1666993066.750161 sender=:1.364 -> destination=(null destination) serial=3 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
+   string "user:/a/saturation"
+signal time=1666993066.750551 sender=:1.364 -> destination=(null destination) serial=4 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyAdded
+   string "user:/a/test"
+signal time=1666993066.750628 sender=:1.364 -> destination=(null destination) serial=5 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyDeleted
+   string "user:/b/background"
+signal time=1666993066.750866 sender=:1.364 -> destination=(null destination) serial=6 path=/org/libelektra/configuration; interface=org.libelektra; member=KeyDeleted
+   string "user:/b/foreground"
```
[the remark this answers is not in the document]
kodebach: Since we also know **why** `dbus` behaves like this (and in fact knew that before the example), we can already say that it affects all plugins that store the `KeySet` during `kdbGet` and compare against `kdbSet`. The exact behaviour will depend on the plugin.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,223 @@
...
+This is problematic, as plugins may store the result of `kdbGet` and use the stored data in the following `kdbSet` to calculate a changeset.
+Plugins have to do this, as there isn't any other Elektra-provided mechanism currently to do change tracking for them.
+
+Assume KDB contains the following keys:
+
+```
+user:/a/brightness
+user:/a/saturation
+user:/b/background
+user:/b/foreground
```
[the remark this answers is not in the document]
kodebach: I disagree. If we remove this, the reproducible example must be extended. I want to know what the problem is **without** having to run some code to see what happens. Also as you said yourself, the reproducible example only shows the problem for `dbus`. This explains the problem in general and shows it is not just an issue in `dbus`. 

Maybe swapping the order would be better. First show the concrete example, _then_ explain the general issue behind.

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
markus2330: The problem is not that either of us doesn't know it. The problem is that it is not properly written in the decision.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,223 @@
...
+   // same as before but uses handle->data instead of the ks argument
+   int kdbSet (KDB * handle, Key * parentKey);
+
+   KeySet * kdbData (KDB * handle) {
+       return handle->data;
+   }
+   ```
+
+5. Allow arbitrary sequences and let each plugin deal with it on a case-by-case basis.
+
```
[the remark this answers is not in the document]
kodebach: This is just option 1, with the additional bit that all sequences are allowed.

The change tracking part is being discussed in #4554. _You_ wanted to split the two decisions...

Also: Technically "arbitrary sequences" is not possible anyway, because we already have a constraint:

Before calling `kdbSet (kdb, ksX, parentX)` there must have been a call to `kdbGet (kdb, ksY, parentY)`, such that
- `parentX` is below `parentY` OR
- `parentX` and `parentY` are stored in the same backend.

That's needed so that the backends are correctly initialized.

So what this option should say is to make the above constraint the only one.

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
markus2330: In the reproducible example exactly the same keys are listed but as full command that can be copy&pasted (if someone wants). Saying several times the same does not add clarity. But I agree that some statements in the first example need to be kept.

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
kodebach: > The problem is not that either of us doesn't know it.

Your question was phrased in a way that I wasn't sure, whether you know the answer or not. 

> The problem is that it is not properly written in the decision.

IMO the general section ("Assume KDB contains the following keys:" etc.) shows pretty well that this is a general problem.

But as I said in another comment, maybe swapping the other (concrete example first, general issue afterwards) would make it more clear.

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
kodebach: True, listing the same keys as used in `kdb set` isn't needed. But an explanation of what the C code does, would be needed, if there is no general explanation of the problem.

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
markus2330: I agree, I corrected my suggestion to not contain "arbitrary sequences".

> Before calling kdbSet (kdb, ksX, parentX) there must have been a call to kdbGet (kdb, ksY, parentY), such that

Not necessarily. We could have a kdbSet which internally does a kdbGet. But I don't think it should be done in the low-level API. Instead it should be in a higher-level API, also doing necessary merges on conflicts.

doc/decisions/operation_sequences.md
markus2330: I agree it is the least-annoying restriction. But IIUC it always needs a ksDup on the user-side, so that a 3-way merge can be done if kdbSet fails.

@atmaxinger can you please give a small example how you imagined normal kdbSet usage (with conflict handling) to be with this restriction?

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,223 @@
...
+```
+user:/a/brightness
+user:/a/saturation
+user:/b/background
+user:/b/foreground
+```
+
+The following sequence will lead to change tracking problems:
+
+```c
```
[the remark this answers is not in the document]
atmaxinger: The reproducible example has too much stuff in it, that is non-essential for this problem, e.g. memory management code. I think having the general description of the problem with some pseudo-code and then the reproducible example works better for understanding the problem.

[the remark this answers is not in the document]
atmaxinger: I have removed it in 5e9a662b

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
atmaxinger: I show that the same behavior is also present in `logchange` in 7784ea82

doc/decisions/operation_sequences.md
atmaxinger: > But IIUC kdbSet would always need a ksDup on the user-side, so that a 3-way merge can be done 

Where are you getting this idea from?
What does `ksDup` have to do with this alternative?
Why do you bring the 3-way merge into this?

Please answer this questions, I don't have any ideas on how to address this in the decision file.

> (with conflict handling)

What conflicts do you mean with that? 

> as it only decreases the likelyhood but doesn't give you any guarantee.

Aggree, but IMHO discussing merge conflict strategies in this decision is a bit off-topic.

atmaxinger: I have added a sample for kdbSet usage in 49c20d33

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
atmaxinger: I've added this alternative.

doc/decisions/operation_sequences.md
kodebach: > IIUC kdbSet would always need a ksDup on the user-side, so that a 3-way merge can be done

I'm not sure how the 3-way merge would be different for this option. From what I understand, you need at least **two** `ksDup` and possibly even a `ksDeepDup` to be on the safe side. To do a 3-way merge you need 3 keysets: base, ours and theirs. AFAIK "ours" is what we gave the `kdbSet` that failed. Then we do another `kdbGet` which gives us "theirs". And "base" is basically what this whole decision is about. For things to be correct, it must the data returned by the last `kdbGet` that **matched** the failed `kdbSet`.

> What conflicts do you mean with that?

I think in this case, it refers to the kinds of conflicts that make a `kdbSet` fail. The recommended way to resolve these is with a 3-way merge. Although I don't think we have a full example with explanations anywhere that definitely works in all cases.

The simplest case here is, if a file is edited on disk between `kdbGet` and `kdbSet`.

>  IIUC we normally have failures anymore in kdbSet but a successful kdbGet would tell us about conflicts...

No idea what this is supposed to mean...

markus2330: I am referring to examples/kdbset.c, what we currently recommend to do when someone wants to call `kdbSet`.

We are discussing about restrictions in sequences, so for me the question comes up if these examples still would work.

> What does ksDup have to do with this alternative?

See line 57 of examples/kdbset.c

https://github.com/ElektraInitiative/libelektra/blob/664bc1d21410afe92dfc3eb9b5921a017176e8fa/examples/kdbset.c#L57

with the current pattern (no kdbGet needed before), one could easily move the ksDup inside the `if` (if we use `if` instead of `while`, which we should probably recommend anyway).

markus2330: > No idea what this is supposed to mean...

A "won't" was missing in the sentence. I'll reexplain in more detail: Currently, earlier kdbSet from other processes will make kdbSet from our own process failing. If you call `kdbGet`, however, nothing fails and you simply pull in the changes from the other process.

Previous situation:

```
Process A                |           Process B
kdbSet() -> succeeds     |
                         |  kdbSet() -> fails with conflict
```

IIUC new situation:


```
Process A                |           Process B
kdbSet() -> succeeds     |
                         |  kdbGet() -> pulls in updates from Process A
                         |  kdbSet() -> succeeds
```

But there is no guarantee for that to happen, obviously also following could happen, even if kdbGet is directly before kdbSet:


```
Process A                |           Process B
                         |  kdbGet() -> NOP
kdbSet() -> succeeds     |
                         |  kdbSet() -> fails with conflict
```

So conflict handling would become more complicated, as there are two scenarios.

kodebach: > We are discussing about restrictions in sequences, so for me the question comes up if these examples still would work.

This option would put a restriction on the `parentKey`. That would not affect the 3-way merge example, because clearly you'll always use the same `parentKey` (at least the same name, if not pointer-equal).

> one could easily move the ksDup inside the if

Pretty sure the `ksDup` only exists to make cleanup easier.

> if we use if instead of while, which we should probably recommend anyway)

No, the `while` make sense. There could be repeated conflicts (e.g. if multiple process try to write at the same time).

doc/decisions/operation_sequences.md
atmaxinger: Done

kodebach:
To address the 3-way merge use case in general: AFAICT only option 3 and 4 might affect the 3-way merge case.

For option 4 (internal keyset and `kdbData`), I think the solution would simply be to create a 
```c
// tries to do a kdbSet
// on conflict uses internal data in kdb to get base for merge then calls kdbGet to retrieve theirs
int kdbMerge (KDB * kdb, KeySet * ours, Key * parent, int strategy);
```

In fact creating such a function would be possible and might make sense for all options.

For option 3, 3-way merge also works. But the code needs to be slightly different. Here is a pseudocode example:

```c
kdbGet (kdb, data, parent);
base = ksDup(data);

// do other things, may include kdbGet with other keyset
// and modify data, e.g.
ksAppendKey (data, newKey);

kdbSet (kdb, data, parent);

// Note: conflict could also be kdbGet with different keyset inbetween
if (has_conflict) {
  ours = ksDup (data);

  while (should_merge) {
    kdbGet (kdb, data, parent);
    theirs = ksDup (data);
    merged = elektraMerge (ours, theirs, base, parent, strategy, parent);

    // move from merged to data    
    ksClear (data);
    ksAppend (data, merged);

    kdbSet (kdb, data, parent);

    // not done in the current example, but replacing `ours` with `merged` and 
    // replacing `base` with `theirs` at this point might make sense
  }
}
```

I may have missed something, but basically the trick is the `ksClear`&`ksAppend` to move the data into the keyset that was used for `kdbGet`. Unless I've missed something, that's the same number of `ksDup` as in the current example, but we need the extra move. That could be avoided, if `elektraMerge` would write to e.g. `ours` instead of creating a new keyset.

doc/decisions/operation_sequences.md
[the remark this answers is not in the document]
markus2330: Actually best would be to have the test code somewhere in tests/ (DISABLED_ as it currently fails).

doc/decisions/operation_sequences.md
kodebach: > So conflict handling would become more complicated, as there are two scenarios.

I don't see what changed. Everything you described should also apply equally to the current situation and to the new situation using option 2 or 3 from this decision. The only difference is in the "new situation" there is a new way for `kdbSet` to fail. When the keyset/parentKey doesn't match. But that is resolved just like the existing conflicts you described. Call `kdbGet` again (this time use the correct keyset) and try `kdbSet` once more.

See also https://github.com/ElektraInitiative/libelektra/pull/4574#issuecomment-1297391738

> It is wrong as it is, as the loop only makes sense if kdbSet failed because of an conflict, which isn't necessarily the case in this code.

But that doesn't actually make the `while` wrong. That only means there is a missing `if (hasConflict)` _around_ the `while`.



> But there are even two ksDup in this example:
> 
> https://github.com/ElektraInitiative/libelektra/blob/664bc1d21410afe92dfc3eb9b5921a017176e8fa/examples/kdbset.c#L63
> 
> And this second one is needed to have theirs and ours.

1. This comment chain is still on option 2, where the `parentKey` must match the previous `kdbSet`. Which makes the solution much simpler than for option 3. But again, see https://github.com/ElektraInitiative/libelektra/pull/4574#issuecomment-1297391738 for a solution that should also work for this option 2.
2. I'm 99% sure in the current `master` situation the example doesn't need this `ksDup` at all. At least not for 3-way merge reasons. It could just be `theirs = ksNew (0, KS_END)` if `kdbGet` would work as expected (It doesn't see [`doc/decisions/internal_cache.md`](https://github.com/ElektraInitiative/libelektra/blob/514319e26cfb62c3577df99d167985ecf9036d13/doc/decisions/internal_cache.md#fewer-keys)).

> We couldn't get it within the loop if we would have done the kdbGet already before kdbSet, at least in situation (1) of process examples above.

I think https://github.com/ElektraInitiative/libelektra/pull/4574#issuecomment-1297391738 should also cover this.

markus2330: Very messy discussion (also my fault), @atmaxinger please update the decision and let us go from there.

kodebach: A good example of why I suggested in https://github.com/ElektraInitiative/libelektra/pull/4562#discussion_r1010366283 that longer discussions should be moved to the root thread.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,253 @@
...
+
+The current public API allows various kind of sequences of `kdbGet` and `kdbSet` calls.
+There is no real guideline of allowed and/or prohibited sequences, other than `kdbGet` for a certain parent key must be called before `kdbSet`:
+
+- initially after `kdbOpen`
+- after conflict errors in `kdbSet`
+
+We discovered that some sequences are currently legal but problematic for change tracking and maybe also other use cases.
+
+Currently, notification plugins assume that a `kdbSet` operation for a certain parent key always directly follows the `kdbGet` operation for that same key.
```
kodebach, line 13:
```suggestion
Currently, notification plugins assume that a `kdbSet` operation for a certain parent key always follows the `kdbGet` operation for that same key, without a different `kdbGet` in between.
```

"directly" suggests `kdbGet (...); kdbSet (...);` without _any code_ in between, which obviously doesn't do anything.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,253 @@
...
+However, this is not enforced in any way within Elektra, nor is it documented that it should.
+
+In applications, `kdbGet` and `kdbSet` of different parent keys might be interwoven in any way.
+This is problematic, as plugins may store the result of `kdbGet` and use the stored data in the following `kdbSet` to calculate a changeset.
+Plugins have to do this, as there isn't any other Elektra-provided mechanism currently to do change tracking for them.
+
+Note, that erroneous behaviour only occurs if plugins are used globally, i.e. as `notification-send` hook plugins.
+If they are mounted for specific backends (via `kdb mount`), this behaviour will not occur as plugins from different backends are isolated.
```
kodebach, line 21:
```suggestion
If they are mounted for specific backends (via `kdb mount`), this behaviour will not occur as plugins from different backends are isolated and backends are treated as an atomic unit with a single parent key.
```

Just isolating them wouldn't be enough, if two calls could still have different parent keys or do different partial reads.

doc/decisions/operation_sequences.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Reproducible Example
+
+The following example will demonstrate the problematic sequence as real, runnable code.
+This example will show the erroneous behaviour for the `dbus` and `logchange` plugins, if used as `notification-send` hooks.
+These both plugins just serve as a general demonstration, and there may be more plugins that do change tracking this way.
```
kodebach, line 27:
```suggestion
These plugins both just serve as a general demonstration, and there may be more plugins that exhibit the problematic behaviour.
```

Could be something other than change tracking in theory.

kodebach (approved):
The problem is clear to me now (always was, but I think the text also makes it very clear now). I have few minor corrections, but IMO this can be merged now.

@markus2330 What do you think? Is the problem clear enough that this PR can be merged?

doc/decisions/operation_sequences.md
atmaxinger: Oh god ... where to even begin with summarizing this ...

doc/decisions/README.md
```diff
@@ -76,6 +76,7 @@ Before you write your first decision, read [Decision Process](decision_process.m
 - [Header File Structure](header_file_structure.md)
 - [Man Pages](man_pages.md)
 - [Change Tracking](change_tracking.md)
+- [Operation Sequnces](operation_sequences.md)
```
markus2330, line 79:
```suggestion
- [Operation Sequences](operation_sequences.md)
```

markus2330 (approved):
Great, very clear now! Should we first merge this or #4632?

mpranj (approved):
Very nice and detailed description! 👍 

I think it's best we merge and move further discussion to a new PR.

markus2330:
Any reason why this is still draft?

atmaxinger:
@markus2330 no, I fixed it now

atmaxinger:
@markus2330 who merges this?

markus2330:
Actually nothing should be merged as the builds fail now and errors can creep in easily.. Help for #4627 is very much appreciated. Furthermore, I still didn't get an answer if #4632 should be merged first.

But as you seemingly feel that you are blocked by this, I merge it.

doc/decisions/operation_sequences.md
markus2330: > Oh god ... where to even begin with summarizing this ...

This is why it is so important to constantly update and also lead discussions to directions that actually clarify questions relevant for you.

Issue #3497, opened 2020-09-22
General validation/normalization/conversion plugin


kodebach:
This was just an idea I had, and its not very concrete, but I'll post it anyway (@markus2330 maybe this is a topic for a thesis)

The idea of Elektra's plugins is very much the Unix philosophy of KISS. A plugin should just do a single job. For example the `ipaddr` plugin checks that a key is a valid IP address, the `macaddr` plugin does the same for MAC addresses and `rgbcolor` deals with colors. All of these plugins will share some basic code.

Some plugins also normalize values into a standard representation, so that the config file can be more flexible without application code (e.g. convert HTML colors like `#aabbcc` into unsigned integers). This is generally a lot more complicated than simple validation. And again all of these plugin will share some code.

My proposal now is to create a generic validation and normalization plugin that deals with all the general logic and delegates to other specialized plugins for the actual validation and normalization of key values.

The `get` and `set` functions of specialized plugins would be simple NOPs, since the generic plugin will call some other functions the plugin exports.

We could also extend the generic plugin to allow more general conversion plugins (which don't convert to a normalized form directly). For example `dns` plugin, that resolves DNS names into IP addresses, which are then validated by the `ipaddr` plugin. The `dns` plugin would simply tell the generic plugin that it outputs IP addresses and the `ipaddr` plugin indicates that it outputs a normalized value. Of course that makes the plugin very complex, but it also makes it very powerful.

---

**NOTE:** this issue should probably done post-1.0. Feel free to close it an mark the issue appropriately.

mpranj:
> The get and set functions of specialized plugins would be simple NOPs, since the generic plugin will call some other functions the plugin exports.

What is the benefit of implementing this as a (classical) Elektra plugin at all? If the plugin does nothing in get or set, maybe it can be a library or a different kind of plugin interface?

markus2330:
It is a good idea and also good grounds for a Master thesis. So if you you are sure this should be your topic, go ahead.

[...]

PR #4621, opened 2022-10-28
[decisions] Transformations

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)
- [ ] I fixed all affected documentation (see [Documentation Guidelines](https://master.libelektra.org/doc/contrib/documentation.md))

## Review

## Labels

### DIFF
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -7,7 +7,7 @@ Those transformations include, but are not limited to:
 
 - Changing the names of the keys inbetween `kdbGet` and `kdbSet` operations
 - Changing values back and forth for normalization, e.g. `true` -> `1`, `1` -> `true`
-- Adding and removing metadata (e.g. the `spec` plugin)
+- Adding and removing metadata
 
 While these features are useful, they do create feature-interaction problems.
 More specifically, problems have been observed in conjunction with the following (overlapping) types of plugins:
@@ -92,6 +92,7 @@ This is a bit of a mixture between changing key names and changing values.
 
 Require plugins that rename keys to remove the `keyNeedsSync` flag.
 This would require making the 'key sync' APIs public.
+Letting user code modify these internal flags may lead to serious bugs.
 This does not solve the problem of changing values.
 
 ### Use change tracking to replace the `keyNeedsSync` flag
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -15,7 +15,7 @@ More specifically, problems have been observed in conjunction with the following
 - notification plugins
 - plugins that do change tracking
 
-The problem, in general, can be described as: Which phase of the KDB should be used for notifications/change tracking?
+The problem, in general, can be described as: Which representation of the KDB should be used for notifications/change tracking?
 
 We differentiate between:
 
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -74,6 +74,7 @@ Plugins relying on change tracking plugins (e.g. notification plugins) will howe
 ## Assumptions
 
 1. False positives for change tracking algorithms are only a minor problem.
+   False positives are that changes are detected, even though nothing changed.
 2. There is no reason to modify or delete existing `meta:/` keys.
 3. Newly generated `meta:/...` keys can
    - either stay and get permanently stored during `kdbSet`
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -73,10 +73,9 @@ Plugins relying on change tracking plugins (e.g. notification plugins) will howe
 
 ## Assumptions
 
-1. 
-2. False positives for change tracking algorithms are only a minor problem.
-3. There is no reason to modify or delete existing `meta:/` keys.
-4. Newly generated `meta:/...` keys can
+1. False positives for change tracking algorithms are only a minor problem.
+2. There is no reason to modify or delete existing `meta:/` keys.
+3. Newly generated `meta:/...` keys can
    - either stay and get permanently stored during `kdbSet`
    - or be written as `meta:/generated/...`.
      The `meta:/generated/...` metakeys are never stored and are automatically removed during `kdbSet` before `storage` is called.
@@ -122,6 +121,16 @@ Something similar could be done for the value of a key as well.
 
 We could also introduce a new phase between before/after storage exclusively for transformations.
 Then we can just do a "fake" call to that phase to get back the transient names for change tracking.
+
+### Call value transformation plugin(s) when `keySetValue` / `keySetString` is called
+
+After a value has been set by the user, call the transformation plugin.
+We could store those callbacks as metakeys, i.e. `meta:/generated/transformation/value/callback/#1`.
+
+In `keySetRaw` we then call all those callbacks.
+We also need to provide a simple API to plugins to register such callbacks for a key.
+
+As the transformations will be applied before `kdbSet`, this will elimate lots of possible false positives in changetracking.
 
 ## Decision
 
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -67,13 +67,13 @@ Plugins relying on change tracking plugins (e.g. notification plugins) will howe
 
 ## Constraints
 
-1.
-2.
+1. We want to enable some kind of transformations
+2. Renaming keys should be possible, at least in storage plugins
 3.
 
 ## Assumptions
 
-1. We want to enable some kind of transformations
+1. 
 2. False positives for change tracking algorithms are only a minor problem.
 3. There is no reason to modify or delete existing `meta:/` keys.
 4. Newly generated `meta:/...` keys can
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -127,11 +127,41 @@ Then we can just do a "fake" call to that phase to get back the transient names 
 
 After a value has been set by the user, call the transformation plugin.
 We could store those callbacks as metakeys, i.e. `meta:/generated/transformation/value/callback/#1`.
+Alternatively, we could implement a linked list or other data structure for the callbacks.
 
-In `keySetRaw` we then call all those callbacks.
+```c
+struct callback {
+    callback_fn function;
+    struct callback * next;
+}
+```
+
+There will be 3 possibilities where transformations take place:
+
+1. For keys with a registered callback: 
+   - Within `keySetValue` / `keySetString` the callbacks are called.
+2. For new keys, i.e. keys without a callback: 
+   - Within `kdbSet`, as it is now.
+3. For new keys that override keys with an intact callback:
+   - Within `ksAppend` and/or `ksAppendKey` the callbacks are copied from the original key and then executed on the new key.
+
+
+```c
+kdbGet (kdb, ks, parent);
+// ks contains two keys: 
+//  - user:/limits/openfiles
+//  - user:/background/color
+
+Key * background = ksLookupByName(ks, "user:/background/color", 0);
+keySetString(background, "#ff00ff"); // (1) -> callbacks are called.
+ksAppendKey (ks, keyNew ("user:/my/key", KEY_VALUE, "1234", KEY_END); // (2) -> transformation will happen inside kdbSet
+ksAppendKey (ks, keyNew ("user:/limits/openfiles", KEY_VALUE, "23", KEY_END); // (3) -> replaces existing key, callbacks of existing key are copied and then executed
+kdbSet (kdb, ks, parent);
+```
+
 We also need to provide a simple API to plugins to register such callbacks for a key.
 
-As the transformations will be applied before `kdbSet`, this will elimate lots of possible false positives in changetracking.
+As the transformations for existing keys will be applied before `kdbSet`, this will elimate false positives in changetracking.
 
 ## Decision
 
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -53,7 +53,7 @@ As new keys by design have the 'key needs sync' flag set, the plugins that imple
 
 ### Observed problems with changing values (normalization)
 
-Suppose we have a plugin that changes the value of a key to the hard-coded value `1` during the storage phase of the `set` operation.
+Suppose we have a plugin that changes the value of a key to the hard coded value `1` during the storage phase of the `set` operation.
 If a plugin does change tracking, this will lead to false positives.
 
 ```
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -80,7 +80,11 @@ This is a bit of a mixture between changing key names and changing values.
 
 1. We want to enable some kind of transformations
 2.
-3.
+3. There is no reason to modify or delete existing `meta:/` keys.
+4. Newly generated `meta:/...` keys can
+    - either stay and get permanently stored during `kdbSet`
+    - or be written as `meta:/generated/...`.
+    The `meta:/generated/...` metakeys are never stored and are automatically removed during `kdbSet` before `storage` is called.
 
 ## Considered Alternatives
 
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -17,7 +17,7 @@ More specifically, problems have been observed in conjunction with the following
 
 The problem, in general, can be described as: Which phase of the KDB should be used for notifications/change tracking?
 
-We differntiate between three types:
+We differentiate between three types:
 
 - *stored* name/value/metadata: How it is actually stored, i.e. the state returned by and passed to the `storage` plugins.
 - *runtime* name/value/metadata: How it is at runtime, i.e. what is returned by `kdbGet` and passed to `kdbSet`.
@@ -99,9 +99,10 @@ In theory the same issue with the 'key needs sync' flag based change tracking ap
 
 ## Related Decisions
 
-- [Hooks](hooks.md)
-- [Change Tracking](https://github.com/ElektraInitiative/libelektra/pull/4554)
-- [Internal Cache](internal_cache.md)
+- [Hooks](../4_partially_implemented/hooks.md)
+- [Change Tracking](../0_drafts/change_tracking.md)
+- [Operation Sequences](../0_drafts/operation_sequences.md)
+- [Internal Cache](../3_decided/internal_cache.md)
 
 ## Notes
 
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -108,6 +108,26 @@ This allows notification and change tracking functionality to work determine and
 
 A similiar thing was already attempted for values, i.e. `meta:/origvalue`.
 
+### Create an API for transformations
+
+Plugins must use a special function to transform key names, e.g.:
+
+```c
+typedef const char * (*ElektraNameTransform)(const Key *);
+
+int elektraApplyNameTransform (KeySet * ks, ElektraNameTransform transform);
+```
+
+How the `elektraApplyNameTransform` function marks the original name is an internal implementation detail.
+May be as `meta:/` or something else entireley.
+
+Something similar could be done for the value of a key as well.
+
+
+### Introduce a new phase for transformations
+
+We could also introduce a new phase between before/after storage exclusively for transformations.
+Then we can just do a "fake" call to that phase to get back the runtime names for change tracking.
 
 ## Decision
 
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -19,9 +19,9 @@ The problem, in general, can be described as: Which phase of the KDB should be u
 
 We differentiate between three types:
 
-- *stored* name/value/metadata: How it is actually stored, i.e. the state returned by and passed to the `storage` plugins.
-- *runtime* name/value/metadata: How it is at runtime, i.e. what is returned by `kdbGet` and passed to `kdbSet`.
-- *intermediate* name/value/metadata: Any state inbetween the two.
+- _stored_ name/value/metadata: How it is actually stored, i.e. the state returned by and passed to the `storage` plugins.
+- _runtime_ name/value/metadata: How it is at runtime, i.e. what is returned by `kdbGet` and passed to `kdbSet`.
+- _intermediate_ name/value/metadata: Any state inbetween the two.
 
 ### Observed problems with changing key names
 
@@ -34,16 +34,16 @@ For example, here is a configuration file with a hypothetical format:
 
 ```
 /DISPLAY/BRIGHTNESS = 100
-``` 
+```
 
 As can be seen, the keys are in UPPERCASE within the configuration file.
-In Elektra keys are case-sensitive. 
-As operations on keysets such as `ksLookup` operate with *runtime* data after the post-storage phase, `kdb get /DISPLAY/BRIGHTNESS` will fail.
+In Elektra keys are case-sensitive.
+As operations on keysets such as `ksLookup` operate with _runtime_ data after the post-storage phase, `kdb get /DISPLAY/BRIGHTNESS` will fail.
 For Elektra, the key `/DISPLAY/BRIGHTNESS` does not exist, as the `rename` plugin transformed this into the lowercase `/display/brightness`.
 
 This leads to problems with the notification plugins.
 As notification plugins are executed after the post-storage phase of the `set` operation, they will receive a keyset with the already transformed keys.
-In this example, the notification plugins will receive all-UPPERCASE keys, and send out notifications with those all-UPPERCASE keys. 
+In this example, the notification plugins will receive all-UPPERCASE keys, and send out notifications with those all-UPPERCASE keys.
 An application listening to those notifications will not be able to query Elektra for those keys, as for Elektra those UPPERCASE keys do not exist.
 
 Apart from the problems with notifications, the way key name changing plugins work also breaks change tracking in plugins like `dbus`.
@@ -82,9 +82,9 @@ This is a bit of a mixture between changing key names and changing values.
 2. False positives for change tracking algorithms are only a minor problem.
 3. There is no reason to modify or delete existing `meta:/` keys.
 4. Newly generated `meta:/...` keys can
-    - either stay and get permanently stored during `kdbSet`
-    - or be written as `meta:/generated/...`.
-    The `meta:/generated/...` metakeys are never stored and are automatically removed during `kdbSet` before `storage` is called.
+   - either stay and get permanently stored during `kdbSet`
+   - or be written as `meta:/generated/...`.
+     The `meta:/generated/...` metakeys are never stored and are automatically removed during `kdbSet` before `storage` is called.
 
 ## Considered Alternatives
 
@@ -123,7 +123,6 @@ May be as `meta:/` or something else entireley.
 
 Something similar could be done for the value of a key as well.
 
-
 ### Introduce a new phase for transformations
 
 We could also introduce a new phase between before/after storage exclusively for transformations.
@@ -149,5 +148,5 @@ Then we can just do a "fake" call to that phase to get back the runtime names fo
 ## Notes
 
 - [Issue #404](https://issues.libelektra.org/404) - dbus and rename plugin do not work together
-- [Issue #955](https://issues.libelektra.org/955) - dbus: non-UTF8 key names 
+- [Issue #955](https://issues.libelektra.org/955) - dbus: non-UTF8 key names
 - [Issue #3626](https://issues.libelektra.org/3626) - origvalue
--- a/doc/decisions/0_drafts/transformations.md
+++ b/doc/decisions/0_drafts/transformations.md
@@ -25,9 +25,9 @@ We differentiate between three types:
 
 ### Observed problems with changing key names
 
-For example, the [`rename` plugin](../../src/plugins/rename/README.md) supports different transformations for `get` and `set`.
+Suppose there is a plugin that changes key names.
+It converts to lowercase for the runtime representation, and to uppercase for the stored representation.
 The plugin is executed in the post-storage phase for the `get` operation and the pre-storage phase for the `set` operation.
-We can instruct it to convert the keynames to lowercase in the `get` operation, and to UPPERCASE in the `set` operation.
 This results in the keys being in UPPERCASE in the configuration files, but they are presented in lowercase to other plugins and applications using the Elektra API.
 
 For example, here is a configuration file with a hypothetical format:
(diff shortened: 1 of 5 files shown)


doc/decisions/transformations.md
```diff
@@ -0,0 +1,98 @@
...
+Apart from the problems with notifications, the way the `rename` plugin works also breaks change tracking in plugins like `dbus`.
+This is because the `rename` plugin does the transformation by deleting the original key and adding a new key with the transformed name.
+The `dbus` plugin implements change tracking by checking the 'key needs sync' flag instead of comparing the values.
+As new keys by design have the 'key needs sync' flag set, the plugins that implement change tracking via the flag will always erroneously detect transformed keys as changed.
+
+### Observed problems with normalization
+
+TBD ... 
+
+In theory the same issue with the 'key needs sync' flag based change tracking approach as above.
```
markus2330, line 53: See above: don't center the problem around "key needs sync" flag. This is an implementation detail. This specific problem could be easily solved within the rename plugin but it still wouldn't solve the whole problem.

Maybe we can even drop the whole idea of "key needs sync" and let the change tracking algorithm decide about if changes happened. This is actually a pure performance topic to be answered by a benchmark.

markus2330 (changes requested):
Thanks for the write-up, great job so far!

doc/decisions/transformations.md
```diff
@@ -0,0 +1,98 @@
...
+
+This leads to problems with the notification plugins.
+As notification plugins are executed after the post-storage phase of the `set` operation, they will receive a keyset with the already transformed keys.
+In this example, the notification plugins will receive all-UPPERCASE keys, and send out notifications with those all-UPPERCASE keys. 
+An application listening to those notifications will not be able to query Elektra for those keys, as for Elektra those UPPERCASE keys do not exist.
+
+Apart from the problems with notifications, the way the `rename` plugin works also breaks change tracking in plugins like `dbus`.
+This is because the `rename` plugin does the transformation by deleting the original key and adding a new key with the transformed name.
+The `dbus` plugin implements change tracking by checking the 'key needs sync' flag instead of comparing the values.
+As new keys by design have the 'key needs sync' flag set, the plugins that implement change tracking via the flag will always erroneously detect transformed keys as changed.
```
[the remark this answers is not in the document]
kodebach: It is a constraint right now, since none of the "key sync" APIs are public.

doc/decisions/transformations.md
kodebach: Using the change tracking to detect changes should definitely be investigated. I'm not sure how feasible it is, but could be worth a try. It should definitely be more reliable, because it is not an extra flag that needs to be considered and could be forgotten somewhere.

doc/decisions/transformations.md
```diff
@@ -0,0 +1,98 @@
...
+For example, the `rename` plugin supports different transformations for `get` and `set`.
+The plugin is executed in the post-storage phase for the `get` operation and the pre-storage phase for the `set` operation.
+We can instruct it to convert the keynames to lowercase in the `get` operation, and to UPPERCASE in the `set` operation.
+This results in the keys being in UPPERCASE in the configuration files, but they are presented in lowercase to other plugins and applications using the Elektra API.
+
+For example, here is a configuration file with a hypothetical format:
+
+```
+/DISPLAY/BRIGHTNESS = 100
+/DISPLAY/COLOR = #FFFFFF
```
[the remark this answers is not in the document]
atmaxinger: > and avoid repetition of rename docu.

Honestly I think it makes the problem better understandable, if I repeat the essential stuff that leads to the described problem. 

If I just link to the documentation of the rename plugin, it is not guaranteed that it still works that way in a year or two. This would make understanding the rationale behind this decision much harder then.

doc/decisions/transformations.md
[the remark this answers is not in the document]
atmaxinger: > We can easily require plugins like rename to remove the "key needs sync" flag.

Yes we could, but part of the current problem is that we don't. So this could be a considered alternative.

doc/decisions/transformations.md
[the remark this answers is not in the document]
kodebach: I haven't had time to properly review this, but from a cursory glance I think the issue is not just that you repeat the `rename` docs. I'm with you repeating the essentials is important, because the problem must remain understandable, even if `rename` changes and even if the problem doesn't exist anymore.

In fact, I think you could even totally ignore the fact that `rename` is an actual plugin and just say "suppose there is a plugin that changes key names". The problem would still be there and since the plugin is hypothetical, it would be obvious that is not a bug in the specific implementation.

doc/decisions/transformations.md
markus2330: Yes, another problem we currently have is that irrelevant changes also cause changes to be triggered, e.g. assuming the key `k` has the content `orig`:

```c
keySetString(k, "abc"); // actually not even needed to trigger the problem
keySetString(k, "orig"); // triggers `k` to need sync
ksAppend(ks, k);
kdbSet(kdb, ks, p); // would write keys even though nothing has changed!
```

Expected would be that `kdbSet` is a NOP as no changes occurred.

doc/decisions/transformations.md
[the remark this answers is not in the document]
markus2330: > If I just link to the documentation of the rename plugin, it is not guaranteed that it still works that way in a year or two. 

I agree.

> In fact, I think you could even totally ignore the fact that rename is an actual plugin and just say "suppose there is a plugin that changes key names". The problem would still be there and since the plugin is hypothetical, it would be obvious that is not a bug in the specific implementation.

Yes, good solution. :+1:

doc/decisions/transformations.md
atmaxinger: > Expected would be that kdbSet is a NOP as no changes occurred.

Yes, good observation! I'll add replacing `keyNeedsSync` flag with change tracking as an alternative.

atmaxinger:
Okay so I updated this PR because some of the edgecases in here are relevant for changetracking.

I think I may have found a solution for the key renaming problem - we could just require that transformation plugins that rename the key in `kdbSet` store the original "elektra-visible" keyname as part of a meta key.

"Elektra-visible" keys are those keys that are returned by `kdbGet`. I don't know if there is a better word for it.

This way, plugins that crucially require "elektra-visible" keys, like changetracking and notification plugins, can still be executed after those transformations.

doc/decisions/transformations.md
```diff
@@ -0,0 +1,102 @@
...
+/DISPLAY/BRIGHTNESS = 100
+``` 
+
+As can be seen, the keys are in UPPERCASE within the configuration file.
+In Elektra keys are case-sensitive. 
+As operations on keysets such as `ksLookup` operate after the post-storage phase, `kdb get /DISPLAY/BRIGHTNESS` will fail.
+For Elektra, the key `/DISPLAY/BRIGHTNESS` does not exist, as the `rename` plugin transformed this into the lowercase `/display/brightness`.
+
+This leads to problems with the notification plugins.
+As notification plugins are executed after the post-storage phase of the `set` operation, they will receive a keyset with the already transformed keys.
```
kodebach, line 39: Don't know how to phrase this properly, but the issue is not "when" the plugins are executed, but which data they key. Executing them after `poststorage` is fine. You could still give them the "elektra-visible" data. You have to return that data from `kdbSet` after all, so the data does exist.

kodebach (commented):
I've said for a long time that the `poststorage` phase in `get` (and therefore the `prestorage` phase in `set`) are chaos and need more structure and rules (e.g. #3497). But with new-backend done, the best chance for that is gone.

However, just for change tracking I believe the solution is quite simple. We just say change tracking works with "elektra-visible" state (names, values, metadata). Then we calculate the diff **at the start** of `kdbSet` (against the end of `kdbGet`). We keep that diff around until the end of `kdbSet` and then we call the plugins.

kodebach:
> we could just require that transformation plugins that rename the key in `kdbSet` store the original "elektra-visible" keyname as part of a meta key.

Bascially, the same idea as `meta:/origvalue` for value transformations and basically the same problem... What if more than one plugin does name transformations?

> "Elektra-visible" keynames are the keynames that are returned by `kdbGet`. I don't know if there is a better word for it.

For values we used the terms "normalization" and "normalized value", but that's also not ideal.

How about:
- **stored name/value/metadata**: How it is actually stored, i.e. the state returned by and passed to the `storage` phases.
- **runtime name/value/metadata**: How it is at runtime, i.e. what is returned by `kdbGet` and passed to `kdbSet` (What you called "elektra-visible")
- **intermediate name/value/metadata**: Any state in between the two, e.g. if there are multiple transformation plugins.

atmaxinger:
> What if more than one plugin does name transformations?

As I mentioned, this is only allowed for the first plugin that does the transformation. I.e., the plugin that changes the **runtime name** (I like that wording btw).

> How about stored [...] runtime [...] intermediate [...]

Oh that sounds great!

doc/decisions/transformations.md
atmaxinger: Yes, I thougt about this too. Basically, we do a deep-dup of the data before we hand it off to the backends? And use this data to return from `kdbSet` and for the notification plugins.

doc/decisions/transformations.md
```diff
@@ -0,0 +1,102 @@
...
+## Considered Alternatives
+
+1. Require plugins that rename keys to remove the `keyNeedsSync` flag.
+   This would require making the 'key sync' APIs public.
+
+2. Use change tracking to detect whether something has changed, replacing the `keyNeedsSync` flag.
+
+3. For key name transformations - require that plugins set a meta key (e.g. `meta:/elektra/visiblename`) with the name before the transformation in the `kdbSet` phase.
+   This _must not_ be done if the meta key already exists, i.e. another plugin already tranformed it beforehand.
+   This allows notification and changetracking functionality to work with the "Elektra-visible" name of the key.
```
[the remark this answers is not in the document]
atmaxinger: We won't need it if we do it like suggested above - i.e. have a deep-dup of the keyset as passed in by the user. 
(Provided that the keyname change is really the last thing that runs before writing it to disk).

doc/decisions/transformations.md
kodebach: I'm not sure what you mean. [What currently should happen is](https://github.com/ElektraInitiative/libelektra/blob/master/doc/dev/kdb-operations.md#set-operation):

- The `KeySet` passed to `kdbSet` is not modified (expect for adding of metadata by `spec`)
- Backends receive parts of a deep-dup copy of this `KeySet`
- The deep-dup copy is discarded at the end

doc/decisions/transformations.md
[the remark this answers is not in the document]
kodebach: > Provided that the keyname change is really the last thing that runs before writing it to disk

You can't guarantee that.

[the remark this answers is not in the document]
atmaxinger: Well then we might need it, if plugins after the keyname change should also be able to use change tracking.

atmaxinger:
I'm not really sure how to continue with this decision. I think I at least made the problem more clear. But when coming up with alternatives, I always kinda hit the wall that I find a solution for part of the problem, but not for the other.

This might actually be two or three completely separate issues 
1. changing key names
2. changing key values
3. changing key meta data

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
...
+## Constraints
+
+1.
+2.
+3.
+
+## Assumptions
+
+1. We want to enable some kind of transformations
+2.
```
kodebach, line 82: I would also assume that the "false positives" for change tracking that you described are a very minor problem. They can also be easily solved, if the change tracking API provides both the old and new value for any changed key. Then a plugin that wants to avoid false positives can do an additional value comparison to see, if the actual value has changed.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
...
+### Enable setting and unsetting of the `keyNeedsSync` flag
+
+Require plugins that rename keys to remove the `keyNeedsSync` flag.
+This would require making the 'key sync' APIs public.
+This does not solve the problem of changing values.
+
+### Use change tracking to replace the `keyNeedsSync` flag
+
+Eliminate the need of the `keyNeedsSync` flag by utilizing the planned change tracking API.
+This does not solve the problem of changing values and false positives.
```
kodebach, line 96: I would actually make it an assumption that `keyNeedsSync` is replaced with the new change tracking implementation.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Store original names and values in metadata.
+
+For key name transformations require that transformation plugins set a metakey (e.g. `meta:/elektra/runtimename`) with the runtime name before they do any transformations in the `kdbSet` phase.
+This _must not_ be done if this metakey already exists, i.e. another plugin already tranformed it beforehand.
+This allows notification and change tracking functionality to work determine and work with the runtime name of the key.
+
+A similiar thing was already attempted for values, i.e. `meta:/origvalue`.
+
+
```
kodebach, line 106: We could also introduce a new phase between before/after `storage` exclusively for transformations. Then we can just do a "fake" call to that phase to get back the runtime names for change tracking.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
...
+This does not solve the problem of changing values.
+
+### Use change tracking to replace the `keyNeedsSync` flag
+
+Eliminate the need of the `keyNeedsSync` flag by utilizing the planned change tracking API.
+This does not solve the problem of changing values and false positives.
+
+### Store original names and values in metadata.
+
+For key name transformations require that transformation plugins set a metakey (e.g. `meta:/elektra/runtimename`) with the runtime name before they do any transformations in the `kdbSet` phase.
```
kodebach, line 100: I'd prefer, if we say plugins must use a special function to transform key names, e.g.:

```c
typedef const char * (*ElektraNameTransform)(const Key *);

int elektraApplyNameTransform (KeySet * ks, ElektraNameTransform transform);
```

This `elektraKeyTransformName` could set the `meta:/` key or mark the key in some other way. That makes things easier, if we discover the `meta:/` isn't enough and also avoids potential mistakes (like changing an existing `meta:/` key). The function could also use "hacky tricks" to avoid creating new keys, if the `KeySet * ks` holds the only reference.

We could have a similar function for the value of a key.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
...
+
+1.
+2.
+3.
+
+## Assumptions
+
+1. We want to enable some kind of transformations
+2.
+3.
```
kodebach, line 83:
```suggestion
3. There is no reason to modify or delete existing `meta:/` keys.
4. Newly generated `meta:/...` keys can
    - either stay and get permanently stored during `kdbSet`
    - or be written as `meta:/generated/...`.
    The `meta:/generated/...` metakeys are never stored and are automatically removed during `kdbSet` before `storage` is called.
```

I really can't think of a use case for metadata transformations that don't fulfil these assumptions.

kodebach (commented):
The problem is clear to me. But I don't have a good solution either. It is a **very** complicated issue and this decision doesn't even cover everything.

There is also the issue of _when_ a plugin that changes key names is executed. If e.g. the `reference` plugin is used, all renaming has to be done before `reference` runs. But AFAIK there is currently no way to ensure this.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
+# Transformations
+
+## Problem
+
+Certain plugins, e.g. `rename`, perform extensive transformations on keys and keysets within Elektra.
+Those transformations include, but are not limited to:
+
+- Changing the names of the keys inbetween `kdbGet` and `kdbSet` operations
+- Changing values back and forth for normalization, e.g. `true` -> `1`, `1` -> `true`
+- Adding and removing metadata (e.g. the `spec` plugin)
```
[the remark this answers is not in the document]
atmaxinger: Yes that's true, but was the only plugin from the top of my head that changed metadata

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
...
+If a plugin does change tracking, this will lead to false positives.
+
+```
+user:/limits/openfiles = 1
+```
+
+If the user changes the value, e.g. using `kdb set user:/limits/openfiles 23`, plugins will observe the new value `23`.
+However, the value-changing plugin will reset that value back to `1`.
+So in practice, the configuration has not been changed.
+Plugins relying on change tracking plugins (e.g. notification plugins) will however think that it has.
```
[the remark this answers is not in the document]
atmaxinger: Yes, if the change tracking is only based on the `keyNeedsSync` flag. If we do value-based change tracking `kdb set user:/limits/openfiles 1` would not lead to a detected change.

doc/decisions/0_drafts/transformations.md
atmaxinger: On second thought, replacing this flag entirely is not a good idea. Detecting changes will always be more compute-intensive than just checking a flag. This might be overkill for plugins that do not really care about the changes.

doc/decisions/0_drafts/transformations.md
atmaxinger: Good idea! I added it.

doc/decisions/0_drafts/transformations.md
atmaxinger: Great idea!

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,128 @@
...
+2.
+3.
+
+## Considered Alternatives
+
+### Enable setting and unsetting of the `keyNeedsSync` flag
+
+Require plugins that rename keys to remove the `keyNeedsSync` flag.
+This would require making the 'key sync' APIs public.
+This does not solve the problem of changing values.
```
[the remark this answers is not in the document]
atmaxinger: True. Added this to the description of the alternative aswell.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
+# Transformations
+
+## Problem
+
+Certain plugins, e.g. `rename`, perform extensive transformations on keys and keysets within Elektra.
```
markus2330, line 5: The extensiveness is not the problem.

```suggestion
Certain plugins, e.g. `rename`, perform transformations on keys and keysets within Elektra.
```

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+Those transformations include, but are not limited to:
+
+- Changing the names of the keys inbetween `kdbGet` and `kdbSet` operations
```
markus2330, line 8: Wouldn't this "inbetween" also be applicable to the others (value and metadata)?

```suggestion
- Changing the names of the keys back and forth
```

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+- Changing values back and forth for normalization, e.g. `true` -> `1`, `1` -> `true`
+- Adding and removing metadata
```
markus2330, line 10: Can you give an example here to make more clear why this is a transformation.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+
+While these features are useful, they do create feature-interaction problems.
+More specifically, problems have been observed in conjunction with the following (overlapping) types of plugins:
+
+- notification plugins
+- plugins that do change tracking
+
+The problem, in general, can be described as: Which phase of the KDB should be used for notifications/change tracking?
```
markus2330, line 18: I don't understand this sentence. It also seems to me to be more specific and not more general?

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+
+We differentiate between three types:
```
markus2330, line 20: Please avoid "types" here, IMHO it doesn't bring clarity:

```suggestion
We differentiate between:
```

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+
+- _stored_ name/value/metadata: How it is actually stored, i.e. the state returned by and passed to the `storage` plugins.
```
markus2330, line 22:
```suggestion
- _persistent_ name/value/metadata: How it is actually stored, i.e. the state returned by and passed to the `storage` plugins.
```

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+- _runtime_ name/value/metadata: How it is at runtime, i.e. what is returned by `kdbGet` and passed to `kdbSet`.
```
markus2330, line 23:
```suggestion
- _transient_ name/value/metadata: How it is at runtime, i.e. what is returned by `kdbGet` and passed to `kdbSet`.
```

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+
+Suppose we have a plugin that changes the value of a key to the hard-coded value `1` during the storage phase of the `set` operation.
+If a plugin does change tracking, this will lead to false positives.
+
+```
+user:/limits/openfiles = 1
+```
+
+If the user changes the value, e.g. using `kdb set user:/limits/openfiles 23`, plugins will observe the new value `23`.
+However, the value-changing plugin will reset that value back to `1`.
```
markus2330, line 65: Little style suggestion (can be ignored, see Elements of Style by Jr. Strunk, William):

```suggestion
The value-changing plugin, however, will reset that value back to `1`.
```

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+So in practice, the configuration has not been changed.
+Plugins relying on change tracking plugins (e.g. notification plugins) will however think that it has.
+
+### Observed problems with adding and removing metadata
+
+This is a bit of a mixture between changing key names and changing values.
```
markus2330, line 71: Better to leave this away from this decision if we don't really have an example for it.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Constraints
+
+1.
+2.
+3.
+
+## Assumptions
+
+1. We want to enable some kind of transformations
```
markus2330, line 81: At least renames are a must-have (so it should be constraints). It could be implemented in storage plugins, though.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+2. False positives for change tracking algorithms are only a minor problem.
```
markus2330, line 82: Please define false positives. If we notify on every `kdbSet` many keys even though actually nothing ever changed, we have a big problem.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+3. There is no reason to modify or delete existing `meta:/` keys.
```
markus2330, line 83: :+1:

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+4. Newly generated `meta:/...` keys can
+   - either stay and get permanently stored during `kdbSet`
+   - or be written as `meta:/generated/...`.
```
markus2330, line 86: Isn't this the internal we already have?

doc/decisions/0_drafts/transformations.md
markus2330: I am not so sure if there is a big difference. strcmp on such little values is very fast. Comparing 4/8 byte values is **maybe** nearly as fast as checking a flag (benchmark needed).

But I agree with @atmaxinger: if we already have an optimization, there should be a good reason to throw it.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+typedef const char * (*ElektraNameTransform)(const Key *);
+
+int elektraApplyNameTransform (KeySet * ks, ElektraNameTransform transform);
+```
+
+How the `elektraApplyNameTransform` function marks the original name is an internal implementation detail.
+May be as `meta:/` or something else entireley.
+
+Something similar could be done for the value of a key as well.
+
```
markus2330, line 125: What would such plugins then do differently that the problem does not occur?

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,152 @@
...
+### Introduce a new phase for transformations
+
+We could also introduce a new phase between before/after storage exclusively for transformations.
+Then we can just do a "fake" call to that phase to get back the runtime names for change tracking.
```
markus2330, line 129:
```suggestion
Then we can just do a "fake" call to that phase to get back the transient names for change tracking.
```

markus2330 (approved):
Good job, problem is quite clear, I did minor suggestions. We can merge this as draft or problem_clear.

doc/help/elektra-glossary.md
```diff
@@ -102,6 +102,15 @@
 - [Metadata](elektra-metadata.md):
   Allows us to describe configuration settings.
 
+- **stored name/value/metadata**:
```
atmaxinger, line 105:
```suggestion
- **persistent name/value/metadata**:
```

doc/help/elektra-glossary.md
```diff
@@ -102,6 +102,15 @@
...
+  How it is actually stored, i.e. the state returned by and passed to the `storage` plugins.
+
+- **runtime name/value/metadata**:
```
atmaxinger, line 108:
```suggestion
- **transient name/value/metadata**:
```

doc/decisions/0_drafts/transformations.md
atmaxinger: I think "representation" is the better word for "phase" here.

doc/decisions/0_drafts/transformations.md
atmaxinger: Well I think we can leave out metadata completely from this decision.

doc/decisions/0_drafts/transformations.md
atmaxinger: Yes, but that is exactly what happens. But I think I found a workable solution for this.

doc/decisions/0_drafts/transformations.md
atmaxinger: If they use this function to transform the name, we internally know that it has an old name. 
This knowledge can be utilized by changetracking and notification plugins to detect false positives.

atmaxinger:
I have adapted the decision with all the requested changes.

I think I also found a solution for the value transformation problems. We can just transform the values before `kdbSet`, using callbacks registered in the meta keys of a key. You can find it in the considered alternatives section.

doc/decisions/0_drafts/transformations.md
kodebach: Similar semantics, but a separate prefix of generated data makes more sense IMO. Also `meta:/internal` is split by plugin and each `meta:/internal/<plugin>/foo` should only be used by `<plugin>` and ignored by everything else (at least that's how I understand it).

doc/decisions/0_drafts/transformations.md
kodebach: Okay maybe making it an assumption is too much. But replacing `keyNeedsSync` should still be an option we explore. However, I think it's mostly independent of what we discuss here.

I brought it up, because the `keyNeedsSync` flag is actually not as straightforward as it seems. It doesn't cover all cases and even in `libelektra-kdb` (which has more information than e.g. plugins) we need to do extra work. IMO `keyNeedsSync` may still be useful for `libelektra-kdb`, but shouldn't be public API and plugins/applications should use the new change tracking instead.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,157 @@
...
+
+If the user changes the value, e.g. using `kdb set user:/limits/openfiles 23`, plugins will observe the new value `23`.
+The value-changing plugin, however, will reset that value back to `1`.
+So in practice, the configuration has not been changed.
+Plugins relying on change tracking plugins (e.g. notification plugins) will however think that it has.
+
+## Constraints
+
+1. We want to enable some kind of transformations
+2. Renaming keys should be possible, at least in storage plugins
```
kodebach, line 71: Adding "at least in storage plugins" makes this constraint meaningless. Before the end of the `get/storage` phase, `libelektra-kdb` doesn't know about any keys. So storage plugins can do whatever transformations they want to. Any such transformation would be considered part of the semantic translation between the storage format and Elektra.


I guess what we want is "user-defined renaming" without the need to build the whole functionality into every storage plugin separately. I.e. either have separate renaming plugins, or have a generic way through which _all_ storage plugins can support renaming.

doc/decisions/0_drafts/transformations.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Call value transformation plugin(s) when `keySetValue` / `keySetString` is called
+
+After a value has been set by the user, call the transformation plugin.
+We could store those callbacks as metakeys, i.e. `meta:/generated/transformation/value/callback/#1`.
+
+In `keySetRaw` we then call all those callbacks.
+We also need to provide a simple API to plugins to register such callbacks for a key.
+
+As the transformations will be applied before `kdbSet`, this will elimate lots of possible false positives in changetracking.
```
kodebach, line 134: 1. This should be done for name and value, not just value to be consistent.
2. I don't like that it makes `libelektra-core` more complex.
3. The name `meta:/generated/transformation/value/callback/#1` is very long. Using metakeys with such long names uses a lot of memory compare to the naive solution of adding a linked list 
    ```c
    struct callback {
        callback_fn function;
        struct callback * next;
    }
    ```
    to the key struct.
4. It's a neat idea, but doesn't work. There is a very simple case that isn't covered:
    ```c
    kdbGet (kdb, ks, parent);
    ksAppendKey (ks, keyNew ("user:/foo", KEY_VALUE, "new value", KEY_END);
    kdbSet (kdb, ks, parent);
    ```
    The new key doesn't have the callback and therefore isn't transformed before `kdbSet`.

kodebach (changes requested):

doc/decisions/0_drafts/transformations.md
atmaxinger: Good points, but

1. Does this actually make sense for names? IMHO names should be handled as close to the storage plugins as possible. But yes, nothing is preventing us from doing it for names also.
2. This is true, but we already do similar things. i.e. `meta:/binary`.
3. Yes, true. How we implement it is not very important to me.
4. It works to eliminate false positives. 
   - A new key is always a change so that it would get detected correctly. 
   - The transformation plugins could be required to handle those cases too, e.g. if no callback is on that key, do the transformation in `kdbSet` as it is now.

kodebach: Ad 4: What exactly would the callback do?

To me it sounded like the callback would perform the transformation. But that would mean that a newly added key is never transformed, unless the plugin also checks during `kdbSet`. But then we're back to the same old problem.

To use a concrete example, take the `user:/limits/openfiles` example you gave in the "Problem" section.

How would this solution work in this snippet?

```c
kdbGet (kdb, ks, parent);
ksAppendKey (ks, keyNew ("user:/limits/openfiles", KEY_VALUE, "23", KEY_END);
kdbSet (kdb, ks, parent);
```

IMO this is not solved at all. There is no callback, because it is a new key and if the plugin "simply runs"  during `kdbSet`, we haven't solved anything compare to the problem statement.

atmaxinger: Yes, there would be 2 different possibilities where transformations happen:
1. For keys returned by `kdbGet`, i.e. with the callback set --> in the callback
2. new keys, i.e. keys that do not have a callback --> in `kdbSet`

In the case of the example, the transformation on `user:/limits/openfiles` would be in `kdbSet`.

As this is a new key, it doesn't matter that the transformation is done later for notification and changetracking plugins. They should detect this key as modified anyway.

But for existing keys, the transformation would take part long before `kdbSet`, so here we can avoid false positives.

kodebach: > They should detect this key as modified anyway.

But wasn't that the point of your example? The key value wouldn't actually be modified, because the plugin changes the value back to `1`.

atmaxinger: In the case of a new key, there wouldn't be a previous value?

atmaxinger: If you are hinting at the case, that `kdbGet` already returns a key `user:/limits/openfiles` and we are overriding this key with `ksAppendKey` with a new key: then we also need to check if the overridden key has a callback and the new one does not. 

If this is the case, then copy the callback to the new case, and execute the callback. So this would be a 3rd possibility where the transformation happens.

kodebach: > copy the callback to the new key, and execute the callback

So `ksAppendKey` would have to do that?

This solution is getting really complicated with a lot of moving parts...

atmaxinger: > So ksAppendKey would have to do that?

Yes, both `ksAppendKey ` and `ksAppend`.

> This solution is getting really complicated with a lot of moving parts...

I don't think it's _that_ complicated, but sure. However, this is the first solution that would really eliminate false positives. I'm always open to new solutions for this problem.

atmaxinger: I updated the decision to include what has been discussed here.

atmaxinger:

atmaxinger:

atmaxinger:
@markus2330 @kodebach what is still required for this decision to be merged?

markus2330:
It is a nice draft decision! I think the only conclusion to this problem can be a set of restrictions we have on transformation plugins. So the decision should basically describe:

- which plugins you can rescue and make work nicely with your change tracking (hopefully the type plugin with normalization on boolean values etc. as this is also a great feature for better merging)
- which plugins we remove as they cause problems (e.g. rename)

kodebach:
It may have gotten lost somwhere, AFAICT **any** kind of transformation should work fine, as long as it happens _during_ the `storage` phase and is done correctly. "Correctly" means:
1. The transformation must reversible.
2. During `get/storage` the transformation happens in the forward direction (e.g. "to lowercase") after the storage plugin returns
3. During `set/storage` the transformation happens in the reverse direction  (e.g. "to uppercase"), then the storage plugin runs and then the transformation happens in the forward direction again (e.g. "to lowercase").

I'm not saying, we should do that for everything. But it is certainly possible, if the backend plugin supports it.

So my suggestion would be this:
1. As markus2330 suggested get as much as possible working with the default `backend` plugin
2. Remove/Deprecate everything that doesn't work with `backend`
3. In the decision explicitly state that more transformations may be supported by other backend plugins. The default `backend` intentionally provides no special support, so that anything that works with `backend` should work with other backend plugins too.

PR #4573, opened 2022-10-18
Update new-backend docs

kodebach:
Incorporates the reviews from #4187

## Basics

- [x] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [x] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [x] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [x] The PR is rebased with current master.

## Checklist

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation (see [Documentation Guidelines](https://master.libelektra.org/doc/contrib/documentation.md))
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://master.libelektra.org/doc/CODING.md))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned [every code](/.reuse/dep5) not directly written by me in [reuse syntax](https://reuse.software/)

## Review

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

- [ ] Add the "work in progress" label if you do not want the PR to be reviewed yet.
- [ ] Add the "ready to merge" label **if the basics are fulfilled** and no further pushes are planned by you.

PR #4622, opened 2022-10-31
logchange: add notification-send hooks to contract

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/news/_preparation_next_release.md
+++ b/doc/news/_preparation_next_release.md
@@ -65,9 +65,9 @@ The following text lists news about the [plugins](https://www.libelektra.org/plu
 
 - Removed the `list` plugin. _(Maximilian Irlinger @atmaxinger)_
 
-### <<Plugin>>
+### logchange
 
-- <<TODO>>
+- Made logchange a notification-send hook plugin _(Maximilian Irlinger @atmaxinger)_
 - <<TODO>>
 - <<TODO>>
 
--- a/src/plugins/logchange/logchange.c
+++ b/src/plugins/logchange/logchange.c
@@ -35,8 +35,10 @@ int elektraLogchangeGet (Plugin * handle, KeySet * returned, Key * parentKey ELE
 			keyNew ("system:/elektra/modules/logchange/exports", KEY_END),
 			keyNew ("system:/elektra/modules/logchange/exports/get", KEY_FUNC, elektraLogchangeGet, KEY_END),
 			keyNew ("system:/elektra/modules/logchange/exports/commit", KEY_FUNC, elektraLogchangeCommit, KEY_END),
-			keyNew ("system:/elektra/modules/logchange/exports/hook/notification/send/get", KEY_FUNC, elektraLogchangeGet, KEY_END),
-			keyNew ("system:/elektra/modules/logchange/exports/hook/notification/send/set", KEY_FUNC, elektraLogchangeCommit, KEY_END),
+			keyNew ("system:/elektra/modules/logchange/exports/hook/notification/send/get", KEY_FUNC, elektraLogchangeGet,
+				KEY_END),
+			keyNew ("system:/elektra/modules/logchange/exports/hook/notification/send/set", KEY_FUNC, elektraLogchangeCommit,
+				KEY_END),
 			keyNew ("system:/elektra/modules/logchange/exports/close", KEY_FUNC, elektraLogchangeClose, KEY_END),
 #include ELEKTRA_README
 			keyNew ("system:/elektra/modules/logchange/infos/version", KEY_VALUE, PLUGINVERSION, KEY_END), KS_END);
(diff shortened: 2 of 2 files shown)


atmaxinger:
Enables the `logchange` plugin to run as hook plugin

kodebach (approved):
LGTM, not sure if we need this, but why not.

That said, I believe following the README and adding `logchange` via `kdb mount` to a single mountpoint only should also work. That way there also shouldn't be the problems discussed in #4554 and #4574. However, if a key is moved from one mountpoint to another (*), one would see the key as removed and the other as newly added.

(*) This can happen by adding a new nested mountpoint with a `kdbSet` that includes more than just `system:/elektra` (i.e. it can't be done via the `kdb` tool)

```c
// assume `system:/foo/bar` exists
parent = keyNew ("system:/");
kdbGet (kdb, ks, parent);
// add mountpoint for `system:/foo` to ks
kdbSet (kdb, ks, parent);
// now `system:/foo/bar` will belong to the new mountpoint
// because ks included the data for the old and new mountpoint, as well as the mountpoint definitions,
// the kdbSet automatically moves the data to the nested mountpoint
```

atmaxinger:
> I believe following the README and adding logchange via kdb mount to a single mountpoint only should also work. 

This still works! I have just added the ability for `logchange` to also be used globally via hooks.

> However, if a key is moved from one mountpoint to another (*), one mountpoint would see the key as removed and the other as newly added.

You mean the mountpoints would see it as such, but the `logchange` plugin wouldn't?

kodebach:
> This still works! I have just added the ability for `logchange` to also be used globally via hooks.

I just wasn't sure we need the hook support here. But I then saw you used it as an example in the other PR, so makes sense.

> You mean the mountpoints would see it as such, but the `logchange` plugin wouldn't?

If `logchange` is added to mountpoint and not used via the notification hook, it
- SHOULD NOT exhibit the problematic behaviour discussed in #4554 and #4574
- WOULD see keys that just move from one mountpoint to the other as added/removed, while the hook version would see them as unchanged. That is simply a consequence of the isolation between mountpoints. Basically this is the reason that we need to have global/hook plugins at all. Adding a plugin globally (as a hook) and adding a plugin to all mountpoints is not the same thing.

atmaxinger:
> I just wasn't sure we need the hook support here. 

Apart from needing it as an example in #4574, the README of the logchange plugin states "The purpose of this plugin is to demonstrate how one can be notified of every removed, added or changed key easily". 

So it more or less acts as a template on how to write notification plugins. I think it just makes sense to also showcase how a hook notification plugin can be developed.

> SHOULD NOT exhibit the problematic behaviour discussed in https://github.com/ElektraInitiative/libelektra/pull/4554 and https://github.com/ElektraInitiative/libelektra/pull/4574

Correct.

> WOULD see keys that just move from one mountpoint to the other as added/removed, while the hook version would see them as unchanged. 

Also correct.

markus2330 (approved):

markus2330:
So hooks are (also) plugins after all? We need a rename of `global` in `infos/status` to hook to adapt to the new terminology.

Please also update the decision PRs with the updates which cases are problematic and which work.

kodebach:
> So hooks are (also) plugins after all? 

The difference is the API they use. Hook plugins are not called via the normal get/set methods.

markus2330:
> SHOULD NOT exhibit the problematic behaviour discussed in https://github.com/ElektraInitiative/libelektra/pull/4554 and https://github.com/ElektraInitiative/libelektra/pull/4574

@atmaxinger Please add this information there, this is one of the problems I had with understanding the scope of your decision. And please let us avoid such discussions in unrelated PRs. (Which would not be a big deal if the decisions would be fully updated but it seems like they were not?)

> The difference is the API they use. Hook plugins are not called via the normal get/set methods.

@kodebach This should be described in doc/plugins/README.md with a link to doc/dev/hooks.md. And there is another difference (the missing isolation) you wrote about here. Can you please add both these points in #4573?

kodebach:
> add this information there

By "there" you mean in the PRs #4554 and/or #4574? Since the split this information definitely has no place in #4574. In #4554 a note about mountpoint plugins vs. hook plugins for change tracking could be more sensible. However, I don't think it has a place there either. #4554 is about hook plugins, because that is the correct way to do change tracking. Change tracking on a mountpoint level has unintuitive behaviour that doesn't match what applications actually see (my second point about `logchange`):

>> WOULD see keys that just move from one mountpoint to the other as added/removed, while the hook version would see them as unchanged.

&nbsp;

> And please let us avoid such discussions in unrelated PRs. (Which would not be a big deal if the decisions would be immediately updated but it seems like they were not?)

No idea what your problem is now. This was a discussion about the behaviour of `logchange` in a PR about `logchange`. If it doesn't belong here, I don't know where it would belong.

> This should be described in doc/plugins/README.md

1. That file doesn't exist
2. `src/plugins/README.md` is the wrong place for this information IMO. It might work on the website, but in the repo I would expect information about the plugin API to be somewhere in `doc`.
3. `doc/dev/hooks.md` quite clearly describes the interface that different hooks use. The part about the isolation is not explicitly there, but mountpoints aren't mentioned either so you can't assume isolated mountpoints either.

Issue #3976, opened 2021-07-31
Remove Internal Iterator


lawli3t:
remove functions related to internal iterator:
- `ksRewind`
- `ksNext`
- `ksCurrent`
- `ksGetCursor`
- `ksSetCursor`
- `ksHead`
- `ksTail`

change `ksAtCursor` -> `ksAt`
document new way of iterating

kodebach:
This list should also include the versions for the metadata keyset, e.g. `keyNextMeta` and `keyCurrentMeta`.

markus2330:
@flo91 which parts are still open in this issue?

PR #4623, opened 2022-11-01
[FLOSS PX] Elixir bindings

0x6178656c:
This PR is to provide bindings to Elixir.

The only user-facing documentation is in `src/bindings/erl_nif/elixir/README.md`.

## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)
- [x] I fixed all affected documentation (see [Documentation Guidelines](https://www.libelektra.org/devgettingstarted/documentation))

## Review

## Labels


markus2330:
The example is very nice but there are still many important parts for a proposal unspecified. What is clear:

- that you want to implement the three modules Key, KeySet and KDB
- that you will have a CMake integration (which already seems to be, at least partly, done :+1:)

What is missing:

- which Erlang-specific features you want to provide. E.g. have a look at https://github.com/ElektraInitiative/libelektra/tree/master/src/bindings/jna/libelektra-kotlin for ideas
- concurrency: how can several KDB etc. be used safely? In other languages we require one KDB per thread, how to do this in Erlang?
- how can a user work with the native data types of Erlang?
- how can a user work with composite types (e.g. using serialization).

0x6178656c:
@markus2330 Thanks a lot for the suggestions.

>     * which Erlang-specific features you want to provide. E.g. have a look at https://github.com/ElektraInitiative/libelektra/tree/master/src/bindings/jna/libelektra-kotlin for ideas

I have added some ideas. In Elixir there exists the concept of "protocols" which are similar to abstract base classes. For instance, a KeySet is something which should implement the `Enumerable` protocol. Once that is done, you have access to many functions which work on enumerable types.

>     * concurrency: how can several KDB etc. be used safely? In other languages we require one KDB per thread, how to do this in Erlang?

I think this is fairly straightforward in Erlang/OTP. I think that the idiomatic way is to implement all the Elektra classes as generic servers. A generic server can be thought of as an isolated thread. Other processes/threads can communicate only via message passing and the generic server processes the message queue synchronously.

>     * how can a user work with the native data types of Erlang?

One very nice thing that I realized recently: In Elixir all strings are UTF-8 encoded binaries. I had a look and this now seems to be also the case in Erlang. (Previously in Erlang strings were lists of codepoints, I think.) So instead of using the NIF functions that deal with strings (and only supports Latin-1 encodings), I can simply pass the raw binary data from Elektra to Elixir and it works perfectly.

>     * how can a user work with composite types (e.g. using serialization).

This is a good point. I think it should be feasible to provide a conversion in both directions.

kodebach:
> In Elixir there exists the concept of "protocols" which are similar to abstract base classes. For instance, a KeySet is something which should implement the `Enumerable` protocol. 

Sounds like Java's interfaces. The `jna` binding for example implements `NavigableSet<Key>` for it's `KeySet` class. Take a look a the other classes, maybe you find some more equivalents for Elixir.

>  In Elixir all strings are UTF-8 encoded binaries.

An important note here is that in Elektra neither string values nor key names are currently required to be UTF-8. Any 8-bit encoding is technically accepted. However, many components (e.g. `toml` plugin, Java binding) already assume UTF-8. So you don't need to change anything, just add a note that the binding only supports UTF-8.

> I think it should be feasible to provide a conversion in both directions

Take a look at the Kotlin binding. It has some serialization support. Not sure how much of that can be done in Elixir, but it can at least be an inspiration of what could be done.

src/bindings/erl_nif/elixir/kdb.ex
```diff
@@ -0,0 +1,277 @@
...
+    raise "ks_current/1 not implemented"
+  end
+
+  def ks_get_cursor(_ks) do
+    raise "ks_get_cursor/1 not implemented"
+  end
+
+  def ks_set_cursor(_ks, _cursor) do
+    raise "ks_set_cursor/2 not implemented"
+  end
```
markus2330, line 260: Old style iterators not needed to implement, see #3976.

```suggestion
```

src/bindings/erl_nif/elixir/kdb.ex
```diff
@@ -0,0 +1,277 @@
...
+    raise "key_copy_all_meta/2 not implemented"
+  end
+
+  def key_get_meta(_key, _meta_name) do
+    raise "key_get_meta/2 not implemented"
+  end
+
+  def key_set_meta(_key, _meta_name, _new_meta_string) do
+    raise "key_set_meta/3 not implemented"
+  end
```
markus2330, line 70: API probably not needed, as key_meta gives you the keyset


```suggestion
```

src/bindings/erl_nif/elixir/kdb.ex
```diff
@@ -0,0 +1,277 @@
...
+    raise "key_new/1 not implemented"
+  end
+
+  def key_copy(_dest, _source, _flags) do
+    raise "key_copy/3 not implemented"
+  end
+
+  def key_clear(_key) do
+    raise "key_clear/1 not implemented"
+  end
```
markus2330, line 38: API probably not needed as the same can be done via key_copy(.., NULL)


```suggestion
```

src/bindings/erl_nif/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+This binding provides the possibility to access Elektra from Erlang.
+
+## Module name
+
+When compiling the module one needs to set a module name.
+This is because the naming conventions differ between Erlang and Elixir.
+The standard module name `Elixir.Elektra` is for an Elixir binding.
+
+The module name can be set with the CMake option `ELEKTRA_NIF_MODULE_NAME`.
```
markus2330, line 11: Is this really needed?

markus2330 (approved):
Reduction in API.

src/bindings/erl_nif/elixir/kdb.ex
0x6178656c: I see. Is the recommended way using an `elektraCursor` and iterating from `0` to `ksGetSize(ks) - 1` using `ksAtCursor`?

src/bindings/erl_nif/README.md
0x6178656c: Well, if you want to make the binding available for both Erlang and Elixir then you probably would have to provide two separate libraries.

I can think of the following options:

* Provide an option so the user can choose a name (current implementation)
* Using CMake `check_binding_included` allow setting of both `erlang` and `elixir` and compile the library with an appropriate module name.
* Only provide Elixir bindings as this PR only focuses on that

markus2330: Yes, it is better if you concentrate on Elixir for this project. The ideal solution would be if both libraries drop out in one build (like we used to have for e.g. python2 and python3). Please collect such further work in an issue.

flo91:

flo91:

src/bindings/erl_nif/elixir/README.md
```diff
@@ -0,0 +1,170 @@
+# Elixir bindings
+
+The functions of the underlying C API are exposed by the modules in `Elektra.System`
```
markus2330, line 3:
```suggestion
The functions of the underlying C API are exposed by the modules in `Elektra.System`.
```

src/bindings/erl_nif/elixir/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+- `Elektra.System`
+
+However, see [`src/bindings/erl_nif/README.md`](../README.md) for an overview of the differences to the C API and the limitations.
```
markus2330, line 7:
```suggestion
See [`src/bindings/erl_nif/README.md`](../README.md) for an overview of the differences to the C API and the limitations.
```

src/bindings/erl_nif/elixir/README.md
```diff
@@ -0,0 +1,170 @@
...
+|> Enum.each(fn %{name: name, value: value} ->
+  IO.puts("#{name}, #{value}")
+end)
+```
+
+## Examples
+
+### Hello Elektra with `Elektra.System`
+
+The analogue of the [`helloElektra.c`](../../../../examples/helloElektra.c) example is given by the following snippet
```
markus2330, line 63:
```suggestion
The analogue of the [`helloElektra.c`](/examples/helloElektra.c) example is given by the following snippet
```

src/bindings/erl_nif/elixir/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+```elixir
+defmodule Main do
+  use Elektra
+
+  def main do
+    error_key = Elektra.System.key_new("user:/error/key")
```
markus2330, line 70: Maybe use "/" as default key?

```suggestion
    error_key = Elektra.System.key_new("/")
```

src/bindings/erl_nif/elixir/README.md
```diff
@@ -0,0 +1,170 @@
...
+    Elektra.Kdb.close(handle)
+  end
+end
+
+Main.main()
+```
+
+## Installation
+
+To have access to the Elixir module you need to compile Elektra with `elixir` added to the CMake option `BINDINGS`.
```
markus2330, line 146: Please also describe if it is available via EXPERIMENTAL etc.

src/bindings/erl_nif/elixir/lib/elektra/system.ex
```diff
@@ -0,0 +1,277 @@
...
+  @on_load :load_nifs
+
+  def load_nifs do
+    :erlang.load_nif(Path.join(:code.priv_dir(:elektra), "libnif_kdb"), 0)
+  end
+
+  # KDB
+
+  def kdb_open(_contract, _parent_key) do
+    raise "kdb_open/2 not implemented"
```
markus2330, line 11: Who sees that error and when does it occur? Isn't it implemented in [src/bindings/erl_nif/lib/nif_kdb.c](https://github.com/ElektraInitiative/libelektra/pull/4623/files#diff-8e1b76f104e84f36aa85921e5cd0d1dc91f8c585effc73a9c44034397fa05398)? So is a better error maybe that the installation is broken or a system module is missing?

markus2330 (approved):
Great work!

src/bindings/erl_nif/elixir/lib/elektra/system.ex
0x6178656c: This is a good point. With the current configuration this error should never occur.

If the module cannot be loaded this is an error in itself and would cause an exception to be raised before this one. (The failure would be in the `load_nifs` function a few lines above which gets called when the module is loaded for the first time.)

The idea behind the wording of the error message is that one might have different NIF libraries/implementations to choose from and not all of them need to implement every function.

In the present case this error message would only occur if there is a mismatch between the functions specified in the two files. I will adapt the message accordingly.

0x6178656c:
@markus2330

I am not sure why there are failing tests and am a bit skeptical if they are related to my PR at all. (Not because I am confident that there aren't any problems, but rather because I would be surprised if the necessary tests would have been generated automatically.)

The test "testcpp_kdb – projectroot.src.bindings.cpp.tests" failing seems to be also failing in the (unrelated) PR #4705: 
https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-4705/9/tests

The other tests "testkdb_allplugins – projectroot.tests.kdb" seem to be similar to the tests failing in #4800: https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/PR-4800/1/tests

To me it seems more likely that some of the tests are flaky for some other reason.

What do you think? Am I even looking in the right places?

markus2330:

markus2330:
Yes, these two tests are not reliable, see #4642

markus2330:
Great job!

PR #4639, opened 2022-11-06
decisions: improving process

markus2330:
## Basics

- [ ] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [ ] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [ ] The PR is rebased with current master.

## Checklist

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation (see [Documentation Guidelines](https://www.libelektra.org/devgettingstarted/documentation))
- [ ] I fixed all affected decisions (see [Decision Process](https://www.libelektra.org/decisions/decision-process))
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://www.libelektra.org/devgettingstarted/coding))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned [every code](/.reuse/dep5) not directly written by me in [reuse syntax](https://reuse.software/)

## Review

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

- [ ] Add the "work in progress" label if you do not want the PR to be reviewed yet.
- [ ] Add the "ready to merge" label **if the basics are fulfilled** and no further pushes are planned by you.

PR #4641, opened 2022-11-06
[decision] internal cache

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


atmaxinger:
I did a bit of a deep-dive now for this decision.

I found 1 problem with the exisitng COW approach, in that we can not track how many keys point to the same data. This means that the approach needs a constraint on the lifetime of copied keys. I added that to the cons.

I also added a new alternative for a full-blown COW implementation that does not come with any usage restrictions. This alternative, however, will require major refactorings and I'm not sure if I'm willing to make them right now.

So my favorite would be the mmap-style COW approach, with the constraint that copied keys must not be used after the original key has been deleted.

atmaxinger:
Sorry there was a rebase problem. I updated it with the correct decision now.

markus2330:
> I did a bit of a deep-dive now for this decision.

:+1: 

> I found 1 problem with the exisitng COW approach, in that we can not track how many keys point to the same data. This means that the approach needs a constraint on the lifetime of copied keys. I added that to the cons.

Is this problem also with mmap? Can you give an API example how a user can trigger that problem?

> I also added a new alternative for a full-blown COW implementation that does not come with any usage restrictions. This alternative, however, will require major refactorings and I'm not sure if I'm willing to make them right now.

Maybe you can do it together with @mpranj? Especially if the problem affects mmap.

> So my favorite would be the mmap-style COW approach, with the constraint that copied keys must not be used after the original key has been deleted.

If we can give clear user guidelines how to avoid that problem, this solution should work, too. What would the guidelines be?

(As always but as we are all new to the decision process: All answers, except the one about coordination of work, only to be answered directly in the decision.)

atmaxinger:
@markus2330 

> Is this problem also with mmap? Can you give an API example how a user can trigger that problem?

> If we can give clear user guidelines how to avoid that problem, this solution should work, too. What would the guidelines be?

[74575a9](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/74575a9f1ca097b62ad4c683089622ed87af0e84)

> Maybe you can do it together with @mpranj? Especially if the problem affects mmap.

Yes that could work.

kodebach (commented):
> Is this problem also with mmap?

AFAICT yes. AFAIK the data allocated by `mmapstorage` is only valid until `kdbClose` is called. For example:

```c
kdbGet (kdb, ks, parent);
kdbClose (kdb, parent);

// AFAIK ks (or at least it's contents) now points to freed memory
```

---

I think there is also another option for COW. We might be able to finally make use of `elektraMalloc` and build COW into the allocator. Then COW could work for everything. I'm not sure how that would look like (probably needs a new `elektraCopyAlloc`), but if we can work it out, it might be more flexible and may need fewer code changes overall.

> I'm not sure if I'm willing to make them right now

This is my opinion, but IMO @atmaxinger should focus on implementing change tracking under the assumption that only "correct" sequences of `kdbGet`/`kdbSet` are used. What the correct sequences are and how this is achieved can be determined and implemented afterwards. AFAICT the only condition for this approach is that the change tracking plugin receives a pre-calculated added/changed/removed diffset from `libelektra-kdb`. For now this could be calculated naively like `dbus` does it and we only test with the sequences where this works. Then we can just improve the diffset calculation later and everything should still work.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -157,9 +197,98 @@ ksRemoveByName (cowMeta, "meta:/type");
...
+
+    uint16_t refs;
+    uint16_t reserved;
+};
+```
+
+### Changes to `KeySet`
 
-Not yet decided.
+For `KeySet`, we need to split out everything to do with the stored keys into a separate datastructure.
```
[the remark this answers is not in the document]
atmaxinger: Two things,

1. If we delete a copied KeySet, we don't known which KeySet is the original, so we couldn't decrement the counter. This could be dealt with storing a pointer to the original KeySet.

2. If the original KeySet is deleted, we don't know which other KeySets point at the data, so updating their count would not work

3. In similar fashion, if you update the original KeySet, the copied KeySets will also contain the new data (if the memory address does not change). This is unexpected behaviour.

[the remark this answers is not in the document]
atmaxinger: [9c035ac](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/9c035ac17415843d155e1cca7159b1ff39dedc93)

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -157,9 +197,98 @@ ksRemoveByName (cowMeta, "meta:/type");
...
+### Changes to `Key`
+
+For the `Key`, we need to extract everything for the data and name into their own structs.
+This is done for memory-management reasons, as we need to track how many keys point to the same data and/or name.
+
+In the data structures below, an empty key does have 96 bytes.
+An empty key of the current implementation has 64 bytes.
+
+A copied, non-modified key with the data structure below does always have 32 bytes.
+A copied, non-modified key of the current implementation has at least 64 bytes (for an empty key), but in reality much more as the name and the data are also copied.
```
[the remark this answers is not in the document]
atmaxinger: [512a805](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/512a805f77aef6db20676cd56a0eed600b76115e)

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -136,6 +136,46 @@ ksRemoveByName (cowMeta, "meta:/type");
...
+  
+  assert (keyString(copiedKey) == keyString(originalKey));
+  keySetString (originalKey, "new value!");
+  
+  keyString(copiedKey); // Error! Original value has been deleted. Pointer to data in copiedKey points to potentially freed memory
+  ```
+
+  The same problems in principle exist for `mmapstorage`, however this is just a storage plugin and the user does not have any API to create COW keys themselves.
+  We can still let users access the flag `ELEKTRA_CP_COW`, we just need to clearly document what is forbidden.
+  Maybe set the `KEY_FLAG_RO_VALUE` on the original key, so that the API itself detects the error.
```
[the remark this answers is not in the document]
atmaxinger: Yes, but we will clearly state that in the documentation.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -136,6 +136,46 @@ ksRemoveByName (cowMeta, "meta:/type");
...
 
+**Cons:**
+
+- Lifetime of a copied COW key MUST be less than the key it was copied from.
+  We can not track how many keys point to the same data this way, so we can only free data if the key does not have the COW flag.
+  If the original key gets deleted, using a COW key that points to the same data will lead to corrupt data.
+  The same is true for updating values of the original key.
+  
+  This is only problematic if we want to use COW for keys outside of `KDB`.
+  If it is only for use within `KDB`, especially for usage as internal cache and in change tracking, we always know that the original keys are going to last as long as the `KDB` instance.
```
kodebach, line 147:
```suggestion
  If it is only for use within `KDB`, especially for usage as internal cache and in change tracking, we could always guarantee that the original keys are going to last as long as the `KDB` instance.
```

It is not automatically guaranteed that keys last until `kdbClose`. For change tracking it would probably make much more sense if the keys were only valid until the next `kdbGet`/`kdbSet`. Otherwise we would have to keep all versions of keys internally until `kdbClose`.

Also I just checked and it turns out that data returned by `mmapstorage` is only valid until the next `kdbSet`, because that's when `munmap` is called. But I think (at least for the cache) we could move that to `close`.

Sorry for the confusion because of my earlier comments, should have checked the code first.

kodebach:
I've had another idea as a variation of the "Full-blown copy-on-write" option.

1. We already have flags for mmap keys. Reuse those to mark COW name & value instead.
2. Use a struct like this:
    ```c
    struct RcBuffer {
      void * data;
      uint16_t refs;
    };

    struct Key {
      union {
        char * data;
        struct RcBuffer * rc;
      } name;
      // same for value
      keyflag_t flags;
      // other fields
    };
    ```
3. Only read the name via `keyName` implemented as:
    ```c
    const char * keyName (Key *  key) {
        if (test_bit (key->flags, KEY_FLAG_COW_NAME)) {
            return key->name.rc->data;
        } else {
            return key->name.data;
        }
    }
    ```
   Do the same for value via `keyValue`.
4. When deleting name/value do something like:
    ```c
    void keyDelName (Key * key) {
        if (test_bit (key->flags, KEY_FLAG_COW_NAME)) {
            elektraRcBufferDel (key->name.rc);
        } else {
            elektraFree (key->name.data);
        }
    }

    void elektraRcBufferDel (struct RcBuffer * rc) {
        if (rc->refs < UINT16_MAX) rc->refs--;
        if (rc->refs == 0) {
            elektraFree (rc->data);
            elektraFree (rc);
        }
    }
    ```
    Note: The `if (rc->refs < UINT16_MAX)` is for `mmapstorage` and similar use-cases where `elektraFree` should never be called. By setting `rc->refs = UINT16_MAX`, `elektraRcBufferDel` becomes a no-op.
5. For setting a new name/value essentially delete name/value and then set a new one. But of course this should be optimized and use `elektraRealloc` where possible.

The advantage here, is that because of the `union` in `struct Key`. There is no memory overhead, if the key is not COW.

doc/decisions/0_drafts/internal_cache.md
```diff
@@ -1,185 +0,0 @@
-# Internal KeySet Cache
```
markus2330, line 1: Did you use `git mv` to move this file?

kodebach:
> more complex than the currently proposed full-blown COW approach as we need all that code + all the cases where the key is not COW

Kind of. Since we already have the mmap stuff, most places where the COW needs to be accounted for should already have an `if` (with the mmap flag). There will be some places that need to be changed, because you can't access e.g. `key->ukey` directly anymore. But replacing those with the correct function call shouldn't be too hard.

But yes, the currently proposed version results in "more uniform" code with fewer `if` checks. However, overall I think this would still be more changed code in the end. The effort of implementing the change should be about the same in both cases I think.

> How does a key get COW with this approach?

Depends. If we want a public API, I would add a `KEY_CP_COW` flag for `keyCopy`. For `mmapstorage` the key has to be constructed manually or via a private API that sets `rc->refs = UINT16_MAX`. But `mmapstorage` already is very lowlevel so that shouldn't be an issue.

---

I've actually had another idea.

```c
struct Key {
  union {
    char * data;
    struct Key * other;
  } name;
  // same for value
  keyflag_t flags;
  // other fields
};
```

Basically, a key either has its own name `key->name.data` or it uses the name from another key that has it's own name `key->name.other->name.data`.

IMO this is actually the best variant, but it doesn't work because of the way our reference counting is set up. If we had a system with just `keyBorrow` (increments ref count, i.e. creates new reference) and `keyRelease` (releases reference, i.e. decrement ref count and `free` if last reference), this should work: 

```c
// returns the current name of key
const char * keyName (Key * key) {
    return test_bit (key, KEY_FLAG_COW_NAME)? key->name.other->name.data : key->name.data;
}

// frees the current name of key
// not a real function, would be part of keyDel
void keyDelName (Key * key) {
    if (test_bit (key, KEY_FLAG_COW_NAME)) {
        keyRelease (key->name.other);
    } else {
        elektraFree (key->name.data);
    }
}

// copy name from src to dest as COW
// not a real function, would be part of keyCopy
void keyCopyNameCow (Key * dest, Key * src) {
    keyDelName (dest);
    if (test_bit (src, KEY_FLAG_COW_NAME)) {
        dest->name.other = keyBorrow (src->name.other);
    } else {
        dest->name.other = keyBorrow (src);
    }
    set_bit (dest, KEY_FLAG_COW_NAME);
}
```

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+	KeySet * ks2 = ksNew (0, KS_END);
+
+	kdbGet (kdb, ks1, keyNew("/somewhere", KEY_END));
+	succeed_if (ksLookupByName (ks1, "/somewhere/key", 0) != NULL, "should find key (1)");
+	kdbGet (kdb, ks2, keyNew("/somewhere", KEY_END));
+	succeed_if (ksLookupByName (ks2, "/somewhere/key", 0) != NULL, "should find key (2)");
+}
+```
+
+> It actually outputs should find key (2) so the assertion fails.
```
markus2330, line 53: This is also said in the next sentence.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+Improve documentation to make people more aware of these two problems:
+
+- add a tutorial about `kdbGet` semantics
+- add full examples how to correctly work with `kdbGet`
+
+### Cachefilter Plugin
+
+Naively one would simply cache the whole keyset and use `ksBelow` to always get the keyset.
+
+This idea was implemented and later on discarded: a3d95f07160d792fdd0f169d8543138c32a2f580
```
markus2330, line 76: Please add a link here.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+
+**Cons:**
+
+- invalidation of OPMPHM
+
+### MMAP Cache without parent key
+
+We make the mmap cache non-optional and only use a single cache, caching everything.
+We remove the parent key of `kdbGet` and `kdbSet` and always return the keyset of the whole KDB.
+
```
markus2330, line 96:
```suggestion
Advantage: then also symlinks and constraints for keys outside of the parentKey would work.

```

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+  Key * copiedKey;
+  keyCopy (copiedKey, originalKey, ELEKTRA_CP_COW);
+  
+  assert (keyString(copiedKey) == keyString(originalKey));
+  keySetString (originalKey, "new value!");
+  
+  keyString(copiedKey); // Error! Original value has been deleted. Pointer to data in copiedKey points to potentially freed memory
+  ```
+
+  The same problems in principle exist for `mmapstorage`, however this is just a storage plugin and the user does not have any API to create COW keys themselves.
```
markus2330, line 178: Not only principle, I think @kodebach is right that `kdbClose` will unmap everything, so the KeySet afterwards is broken.

I think this is currently also not properly documented, so we definitely must do something about it.

```suggestion
  The same problems exist for `mmapstorage` where `kdbClose` frees (munmap) the KeySet.
```

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+  We can still let users access the flag `ELEKTRA_CP_COW`, we just need to clearly document what is forbidden.
+  Maybe set the `KEY_FLAG_RO_VALUE` on the original key, so that the API itself detects the error. 
+  There is, however, no flag for `keyDel` that we could set.
+
+### Data restrictions
```
markus2330, line 183: I would put this earlier so that it better grouped by topics.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+    } data;
+
+    size_t dataSize;
+
+    uint16_t refs;
+};
+
+struct _KeyName {
+    char * key;
+    size_t keySize;
```
markus2330, line 230: Actually we want to get rid of this, see ./2_in_progress/store_name.md

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+
+    char * ukey;
```
markus2330, line 232: I have an additional proposal here: What about having an extra namespace, several (ref-counted) pointers to namespace, dirname and basename. The dirname is what we have as parent key. Then we need to store the parent-key (usually the largest portion) only once, and for all the other keys only the basename.

This way should save plenty of memory (instead of using more). The problem is that we cannot have a `keyName` anymore.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+
+    uint16_t refs;
+};
+```
+
+The following calculations are based on the AMD64 platform:
+
+In the data structures above, an empty key does have
+- with `_keyData` and `_keyName` allocated: 96 bytes.
+- with `_keyData` and `_keyName` `NULL`: 32 bytes.
```
markus2330, line 252: :+1: please also write another common case: keyName allocated but no _keyData

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+    OpmphmPredictor * opmphmPredictor;
+
+    uint16_t refs; /**< Reference counter */
+};
+
+struct _KeySet {
+    struct _KeySetData * data;
+
+    struct _Key * cursor; /**< Internal cursor */
+    size_t current;		  /**< Current position of cursor */
```
markus2330, line 310: As @kodebach already wrote this should be removed.

```suggestion
```

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,350 @@
...
+```
+
+The following calculations are based on the AMD64 platform:
+
+An empty keyset with the datastructure above has 80 bytes.
+An empty keyset with the current implementation has 64 bytes.
+
+A copied, non-modified keyset with the datastructure above has always 32 bytes.
+A copied, non-modified keyset with the current implementation has at least 64 bytes (for an empty keyset).
+
```
markus2330, line 325: Sizes are not the only criteria, a disadvantage of this approach is the probably higher number of allocations.

markus2330 (changes requested):
Great job!

markus2330:
> focus on implementing change tracking

No, it is much better if first the more fundamental problems are tackled.

> I've actually had another idea.

What would be the advantage? Doesn't it require the names to be read-only?

atmaxinger:
> IMO this is actually the best variant

It's not possible to change the value of the original key, but that is a constraint we can have, as long as the scope for COW is just within KDB.

> but it doesn't work because of the way our reference counting is set up

What is stopping us from just implementing `keyBorrow` and `keyRelease`?

```c
void keyRelease (Key * key)
{
  key->refs --;
  if (key->refs == 0)
  {
    // free key
  }
}

Key * keyBorrow (Key * key)
{
  key->refs ++;
  return key;
}
```

doc/decisions/0_drafts/internal_cache.md
atmaxinger: No, Nautilus.

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: What do you mean by "this"? The "Data restrictions" and "API restrictions" alternatives?

doc/decisions/0_drafts/internal_cache.md
kodebach: `git mv` doesn't actually do anything special (see https://stackoverflow.com/a/1094392). The file is no longer detected as moved, because too much has changed.

kodebach:
> No, it is much better if first the more fundamental problems are tackled.

The thing is, _this is not a fundamental problem_. I even described how change tracking can be implemented in such a way that changing the internals is simple. In fact, there is so much theory crafting going on here, that I fear we'll end up with a solution that works in theory but in practice turns out to be impossible because we overlooked something. IMO it is much better to start writing code somewhere to see what works.

It is of course @atmaxinger's choice, but I think we should not force him to work out all the little details first, when he would rather start coding now.

> > I've actually had another idea.
>
> What would be the advantage? Doesn't it require the names to be read-only?

Yeah, that's true. It would be good enough for both `mmapstorage`/cache and change tracking use cases, because all those keys are part of a keyset and therefore have read-only names. But it wouldn't work for values then. So probably we can ignore that solution.

> What is stopping us from just implementing keyBorrow and keyRelease?

There was a whole decision and long PR(s) on reference counting. I don't even wanna dig it out to link to it. Basically the issue is that with the `keyBorrow`/`keyRelease` approach things like `ksAppendKey (ks, keyNew (...))` and more importantly `ksNew (1, keyNew (...), ...))` wouldn't work.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: > higher number of allocations.

That is an issue I was concerned with too. But do we actually have any data to show that this is a problem? Is it really the high number of individual allocations or is it actually the larger number of indirections?

If the problem is just a large number of small allocations, then we could maybe implement a pool allocator. Then there would just be one (or multiple) bigger pools for each type of struct.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: Would that really be a good thing? Reach up out of the mountpoint breaks isolation in any case. It can also only work, within hook plugins, because mountpoints are still isolated for mounted plugins.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: As stated elsewhere, I checked and `munmap` actually happens in `kdbSet` for `mmapstorage`. Note: I have not checked when this would be called for the cache.

kodebach:
> No, it is much better if first the more fundamental problems are tackled.

To expand upon my [previous comment](https://github.com/ElektraInitiative/libelektra/pull/4641#issuecomment-1305843939) above: It seems this decision is heading more and more in the direction of API changes (at least internal APIs, if not public APIs). This would not only fit much better with my thesis than with @atmaxinger's, but it will also almost certainly lead to long discussions. I really don't think @atmaxinger should spend more time on this than he actually wants to, and he definitely shouldn't have to wait for this decision to start work on the actual change tracking (which can be doen completely independently of this).

doc/decisions/1_in_discussion/internal_cache.md
kodebach: Storing the namespace separately could make sense (for different reasons), but that's just a single byte.

Regarding `dirname` & `basename`:
There are many problems with this. The first and biggest is where do we split between `dirname` and `basename`? Finding the optimal split would be quite difficult and would change over time. How would the API for key creation and key name changes look like?

If we really want to go in that direction, I think we actually need to rethink the whole API (again) and see, if we can find a better way to do things.

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: [3dda5d5](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/3dda5d5bfa9b15dc7e48d4218e10f6acede49b12)

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: [bbc8302](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/bbc83023b7dd2d5b31684eb01416ec4d61a06852)

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -157,9 +197,98 @@ ksRemoveByName (cowMeta, "meta:/type");
...
+    uint16_t refs; /**< Reference counter */
+    uint16_t reserved; /**< Reserved for future use */
+};
+
+struct _KeySet {
+    struct _KeySetData * data;
+
+    struct _Key * cursor; /**< Internal cursor */
+    size_t current;		  /**< Current position of cursor */
```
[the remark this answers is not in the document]
atmaxinger: [c81d549](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/c81d5493f6c3b5a5962f08fa2bb1b4bc1714eefb)

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: [c81d549](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/c81d5493f6c3b5a5962f08fa2bb1b4bc1714eefb)

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: [0ca9f78](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/0ca9f785b7d7ce8c70c8c81bebedcd8e3f20675f)

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: [0734b46](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/0734b46b19d14ec65636f3428c65435d6f2bb06f)

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: What exactly are our concerns with the number of allocations? @markus2330 can you formulate a research question for this? I can then do some research and benchmarking to find the answer for it.

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: [217dff4](https://github.com/ElektraInitiative/libelektra/pull/4641/commits/217dff470ccf40b28993628aaf6d91eabdf2d545)

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: Okay. But what does "in progress" mean in that case? Is somebody actively working on it?

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: Nice proposal, but let's try to avoid doing everything at once in this decision.

atmaxinger:
I have to agree with @kodebach. As I originally proposed - I'm for building the `mmap`-style COW. This requires the least changes, and can be developed as quickly as possible.

As we found out, the constraints are already present with the cache, and it wasn't much of an issue. If we ever need the full-blown copy-on-write, this can be added without breaking changes to apps using Elektra, as it does not add any new constraints.

The change tracking can then use a simpler approach, knowing that we do copy-on-write in the background. 

If @markus2330 is fine with this, I will formulate this as the decision.

doc/decisions/0_drafts/internal_cache.md
markus2330: So we should separate the moving of the file and changing the content? Or is it a GitHub problem and moves are detected differently than `git` would do?

a1e7bf8aafa711d719aad8693cad99eeb5b3c1f3

atmaxinger: Also locally the moved file still has all the history of the "old" file.

doc/decisions/1_in_discussion/internal_cache.md
markus2330: Yes, @mpranj worked on it.

doc/decisions/1_in_discussion/internal_cache.md
markus2330: symlinks are not affected by the problem you describe. I updated the suggestions to make this more clear. I agree that this is actually a separate problem but it would be solved by some mmap solutions proposed here.

doc/decisions/1_in_discussion/internal_cache.md
markus2330: @mpranj can you help out here?

(That `kdbSet` does it might be true but is totally uninteresting for the API user.)

doc/decisions/0_drafts/internal_cache.md
kodebach: Seems like a GitHub UI issue in the PR UI. I think the treat the whole PR as a single commit to generate the diff. If you look at the git blame of the moved file, the old history is still there: https://github.com/ElektraInitiative/libelektra/blame/330dc5383aedca4aa0c935e9abdd93f3868203ac/doc/decisions/1_in_discussion/internal_cache.md

doc/decisions/1_in_discussion/internal_cache.md
kodebach: AFAIK there has been zero progress on removing the key names. Also "In Progress" is _before_ "Decided" so that's extra confusing and more reason why I suggested to remove "In Progress" in #4639.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: What exactly do you mean by "symlinks"?

`kdb get -a` is exactly the issue I describe. I normally expect mountpoints to be isolated. Having a key `system:/foo/bar` refer to `../../abc` (in whatever way) when `system:/foo` is a mountpoint is always fishy. The other way around e.g. `system:/abc` referring to `system:/foo/bar` (assuming `system:/abc` is _not_ a mountpoint) would be less fishy, but still a bit weird.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: > That `kdbSet` does it might be true but is totally uninteresting for the API user

It's not phrased perfectly. But it should tell the API user, that the keyset is only valid until `kdbSet` is called.

In any case, this document is not addressed to API users, so it doesn't matter if it is interesting to API users or not, if it is interesting to the audience of the decision. Which it is, because the only interesting bit here is where `munmap` happens.

atmaxinger:
So, I have now tried to formulate a decision based on the mmap COW approach.

When formulating how to integrate this with `kdbGet` and `kdbSet` I stumbled upon an old problem ... The problem with operation sequences.
We can not use this approach as-is, as every `kdbGet` deletes the keys from the previous `kdbGet`. This leaves us with dangling pointers.

From my POV we have three options:
1. (easiest) restrict the squences of `kdbGet` and `kdbSet`.
2. (medium) add a pointer(s) to the "source" key(s) (potentially 2 - one for the name and one for the data). Increase reference count for every cow-copied key on a key. When the COW-copied key is deleted or its data/name is changed, decrease the reference and potentially keyDel the original key.
3. implement the full-blown COW.

kodebach:
> I have to agree with kodebach.

I'm not 100% certain we are on the same page. The fact that you formulated a decision already, rather indicates we are not.

My suggestion was, you completely ignore this PR and this decision for now, because it will probably take quite a while to come to a proper decision here.

Instead you just implement the change tracking and notification hook the way we originally thought of. Simply `ksDeepDup` the keyset at the end of `kdbGet`, then create a diff at the start of `kdbSet` and pass that diff to the notification hook plugins. Yes, that approach has the issue with the sequences. But just ignore that. Assume only working sequences are used.

At a later point we can then either restrict the allowed sequences, implement this cache stuff, or use some other way to properly generate the diff in `kdbSet`. None of this matter for the `kdb record` use case. The plugin receives the diff from `libelektra-kdb` and just assumes this diff is correct. I think this will be much more productive in the short term and lead to results much quicker.

kodebach:
Another idea: Use the cache data for change tracking.

I'm not entirely sure this is possible (@mpranj may know more), but the way I understand it the cache should be updated at the end of every `kdbGet` that was a cache miss. So during `kdbSet` (assuming there is no external modification, i.e. conflict) the on-disk data of the cache should always be up-to-date. My idea would be to just read the cached keyset from disk and diff against the current keyset in `kdbSet`.

This would mean enabling change tracking also enables the cache (or at least updating the cache, we don't have to use it in `kdbGet`). If that's not wanted or if the cache data cannot be used directly for some other reason, the same approach could still be used. We'd just have to write the keyset to disk during `kdbGet` and use it during `kdbSet`. Since disk space is far less precious than RAM, we could even create separate files for every parent key. If we do go down this route, `kdbClose` should cleanup the files created by this KDB instance to avoid wasting disk space.

This approach wouldn't be very performant (since it uses disk IO), but especially if we can use the cache data, it should be pretty easy to do.

doc/decisions/1_in_discussion/internal_cache.md
markus2330: > What exactly are our concerns with the number of allocations? @markus2330 can you formulate a research question for this?

The concerns are basically covered with CPU+memory overhead. I was throwing this in as it is easy to calculate during the decision phase. Many more allocations are definitely not favorable for CPU+memory overhead and give a more complete picture than simply looking at bytes.

Ideally you could make a table with the different COW options and their estimated resource usage (bytes+nr of allocations). For the thesis this will be nice; if it will already help for the decision, I have to admit that I don't know.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: > It is not a single byte. With COW is N*M bytes, N the number of namespaces, M the shared portion of the keyname.

Assuming we remove the escaped key name, the namespace is stored literally as a single byte. There is no point in doing anything other than just copying it.

> I suggested the parent-key above.

But what is _the parent-key_? The term is context-dependent. There is the parent key of a mountpoint, the parent key used in `kdbGet`, etc.

Even if one of those is chosen (parent of mountpoint would be sensible), this still doesn't answer how `keyNew` or `keySetName` would now about this.

> I don't see a connection between "not having keyName (and an iterator instead)" and "need to rethink the whole API".

And that's honestly, why you shouldn't make these decisions. I've dealt with key name changes before. I know it affects _literally everything_. If there is any point to rethink the API that is the point. It is almost impossible to find an API change that would affect more code than key name changes.

And like I explained above, but you may not have realized yet, the changes you proposed **require** API changes. Not just that, but API changes to pretty fundamental functions used all over the code base. This really would be the point to rethink the whole API.

> mpranj can you add these possibilities to the store_name.md decision?

I seriously doubt mpranj wants to be involved in this. AFAIK the task isn't even related to mpranj's previous work. IMO it is a very bad habit of yours to assign such tasks to random people.

markus2330:
@kodebach please stop trying to supervise other people. Your statements ["completely ignore this PR and this decision for now"](https://github.com/ElektraInitiative/libelektra/pull/4641#issuecomment-1307463565) and ["I seriously doubt mpranj wants to be involved in this. AFAIK the task isn't even related to mpranj's previous work. IMO it is a very bad habit of yours to assign such tasks to random people."](https://github.com/ElektraInitiative/libelektra/pull/4641#discussion_r1016914103) are totally unacceptable. You are sabotaging Elektra's goals here (like memory consumption in this case, which we know is terrible with this approach and cannot be implemented as after-thought). Decisions are also totally not the place to discuss who does what in which order, see also #4639. We already have several flawed change tracking algorithms, your proposal to implement one more without any backing of a decision is astounding. Please don't give a fast answer to this, you will do us both a favor if you sleep over it. I welcome your feedback very much but it must be also clear that we must work together and not against each other as happened here.

Why these decision(s) must be done first is simply answered: We already learned that Elektra's core currently doesn't have the capability to do efficient change tracking other than "if anything has changed" (via needs sync flag etc.). So we must **first** clarify the fundamentals (what must be change in the core), so that efficient change tracking is even possible. I am very much in favor of @atmaxinger's approach to unify the different change tracking algorithms, which is a delicate operation at the heart of Elektra, desperate in need for these decisions.

> To expand upon my https://github.com/ElektraInitiative/libelektra/pull/4641#issuecomment-1305843939 above: It seems this decision is heading more and more in the direction of API changes (at least internal APIs, if not public APIs). This would not only fit much better with my thesis than with @atmaxinger's,

Yes it is. Again: the decision by itself does not say who will implement what. It looks like that almost certainly a much bigger step towards improving Elektra can be done if @atmaxinger, @kodebach and @mpranj together work towards a single solution. Obviously what the solution is, must be decided first. I created #4645 to track which decisions are needed for this. There we can also discuss time-lines and who does what (decisions+implementations).

> There was a whole decision and long PR(s) on reference counting.

A quite incomplete decision, most likely not understandable for anyone not involved in the discussions, is there: doc/decisions/4_partially_implemented/reference_counting.md
Would be great if this gets improved. I hope we can avoid such decisions in the future.

I couldn't review this decision (internal_cache) properly due to lots of disturbing noise I couldn't ignore anymore. But to give @atmaxinger quick feedback, the decision:

1. doesn't address API issues enough. I propose that keyDup and ksDup always creates COW keys/key sets. (And we maybe add ksDeepDup to the public API if we find a more efficient way to do deep duplication with COW keys than a loop of keyDup. And a more efficient solution is probably needed if we use ksDeepDup in kdbGet. But for this, I agree, we need a benchmark first.).
2. should review by @mpranj (at least the mmap parts)

We might nevertheless merge the decision soon (only CI is not satisfied), as it is annoying to not see a diff to the previous version and several comments here are questionable.

atmaxinger:
@markus2330 

> I propose that keyDup and ksDup always creates COW keys/key sets. 

Yes, this actually falls in line with my thinking. However, this requires the full-blown approach. 
From a technical POV this is actually my favorite, because it is completely transparent and AFAICT does not have any weird edge cases. However, it requires a lot of changes to the `mmapstorage` plugin, which I am not willing to make. If @mpranj is willing to do so, I'm open to implementing the full-blown approach.

kodebach:
The technical parts first.

> it is annoying to not see a diff to the previous version

You can select a range of commits to view on GitHub. In the [Files tab](https://github.com/ElektraInitiative/libelektra/pull/4641/files) at the top right select "Changes from all commits". There is also several places in the Conversation tab that link to a pre-selected range of commits (e.g. the "View Changes" button next to "New changes since you last viewed", if available).

> I propose that keyDup and ksDup always creates COW keys/key sets

I agree for `keyDup`. What to do with `ksDup` is another question. COW for keysets isn't really needed for change tracking AFAICT. Implementing it would be a bonus and definitely something that could be decided later.

> add ksDeepDup to the public API if we find a more efficient way to do deep duplication with COW keys than a loop of keyDup

This is where we get into the rethinking of the API part. All the memory usage and COW things get a bit simpler IMO, if the allocator was a bit smarter, or at least provided more information. For example, if we had a way to put all the data of one `KeySet` into a contiguous region of memory, a deep copy would just be a `memdup`. But doing that would need API changes.

> I have another proposal 
> _from https://github.com/ElektraInitiative/libelektra/pull/4641#discussion_r1015571608_

Along the lines of splitting key names and using a better allocator there is also the idea of string interning. Also something that could dramatically improve memory use, but almost certainly couldn't be implemented without API changes.

> cannot be implemented as after-thought

Simply not true. I described how we could start on other parts of `kdb record` and use a temporary changeset calculation until the decison(s) are done.

---

Now the organizational parts:

I had a bigger response, but positing it is not going to help. We've had enough discussions that I know you won't understand right a way and that it would need a few more days of back and forth. It's not worth that time...

Just a few quick responses:

"completely ignore this PR and this decision for now" was just a suggestion. It was not an instruction or an assignment and definitely not an attempt to supervise. I wanted to get things moving and to get back on track. The original goal was `kdb record`, but now we are 3 PRs deep into semi-related stuff. I wanted to explain how `kdb record` can be started in a way that allows this decision here to be made at a later date. That is also why I clarified again, after atmaxinger updated the decision.

"I seriously doubt mpranj wants to be involved in this." was clearly a misguided comment. I'm sorry for that. The gist of it is, you've done that to me before. You've assigned tiny tasks from conversations I wasn't involved in to me. I was always quite annoyed when that happend.
My recommendation for cases like this, would be do create an issue and assign that to someone, if you don't have the time to do it yourself. Then it can't happen that the comment goes overlooked and we forget the task. #4645 is a start, but the issue should also state what needs to be done. If we don't think about it for a few weeks, no one will remember what was supposed to happen with store_name.md.

kodebach:
@atmaxinger @markus2330 I also want to make sure this idea doesn't get overlooked: https://github.com/ElektraInitiative/libelektra/pull/4641#issuecomment-1307498849

I actually think it is the most promising solution, apart from the nuclear option of making everything COW all the time.

markus2330 (approved):
For me this decision already looks very good. A few (small) last changes, then we can merge for the next round.

markus2330:
@kodebach There is no need anymore to use political tricks like repeating the same opinion again and again or trying to be the loudest (like shouting "Simply not true"). It is enough to request changes and then the opinion must be accounted for (at least in Notes or Considered Alternatives). I tried to clarify in my review the MMAP approaches. Good that your refrained from posting a long text.

It is much more helpful for the decision author to give concrete suggestions of what should be changed in the text. To me it is unclear if https://github.com/ElektraInitiative/libelektra/pull/4641#issuecomment-1307498849 differs from the two MMAP approaches that are already considered alternatives and if so, how.

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: Is there a resolution to this?

atmaxinger:
@markus2330 I integrated your latest bunch of suggestions.
@kodebach I integrated your idea with the mmap-cache for change tracking.

I also added some pseude-code for the keyset API in regards to the full-blown COW approach.

atmaxinger:
Are the decision also included in the manual? Because it looks like my pseudo code is what breaks LaTeX.

doc/decisions/1_in_discussion/internal_cache.md
markus2330: I updated the suggestion already before, it explains symlinks now.

The other stuff is imho another decision and not really important as symlinks already show the whole problem.

kodebach: You still haven't addressed why this would be actually be a good default behaviour. You just claim it is an advantage without any explanation.

Like I said, I would normally expect mountpoints to be isolated. If symlinks work like this, the isolation is partially broken. One could argue that the problem right now is that such "broken" symlinks are not prevented. But your solution doesn't completely fix the problem either.

Take a variant of my previous example:
- `system:/foo` is a mountpoint, there are no other mountpoints
- `spec:/foo/bar` refers to `system:/abc` via `meta:/override`

Now in a plugin that is part of the mountpoint `system:/foo` doing a `ksLookupByName (ks, "/foo/bar", 0)` would **NOT** use `system:/abc`, because that key is **never** part of the keyset passed to this plugin.

However, in an application that used `system:/foo` as the parent, the `meta:/override` **would work** with your proposal. To me that seems like very confusing behaviour, because both the application and the plugin seemingly use the same parent key.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: Since you marked my comment as disruptive, I'll repeat the purely technical and absolutely correct parts again. There is no need to answer it here and now. But if you expect this approach to be added to the decision or to be discussed elsewhere, we need an answer to these questions.

> It is not a single byte. With COW is N*M bytes, N the number of namespaces, M the shared portion of the keyname.

Assuming we remove the escaped key name, the namespace is stored literally as a single byte. There is no point in doing anything other than just copying it.

> I suggested the parent-key above.

But what is _the parent-key_ here? The term is context-dependent. There is the parent key of a mountpoint, the parent key used in kdbGet, etc.

Even if one of those is chosen (parent of mountpoint would be sensible), this still doesn't answer how keyNew or keySetName would now about this. I don't see how this would work without API changes to very basic APIs used all over the code base.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,411 @@
...
+	kdbGet (kdb, ks2, keyNew("/somewhere", KEY_END));
+	succeed_if (ksLookupByName (ks2, "/somewhere/key", 0) != NULL, "should find key (2)");
+}
+```
+
+It was found unexpected that the second assert - should find key (2) - will fail.
+
+## Constraints
+
+- memory consumption must be low for `kdbGet`, see [4. Goal: Performance](/doc/GOALS.md), in particular, deep duplication is too expensive
```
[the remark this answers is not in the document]
kodebach: > For non-optional parts of Elektra, also non-POSIX systems must be supported

We should clarify what "non-POSIX" means. AFAIK we only support POSIX and MinGW (subset of POSIX). At least that is what we test for. Last I checked, even `libelektra-core` did not compile on a pure Windows system, so unless there is something I missed, I don't think we can claim Elektra runs on "anything with libc".

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,411 @@
...
+A `ksDeepDup()` of a keyset with COW keys will create a keyset with deep-copied non-COW keys.
+Internally we may need a `ksCowDup()` function to create a keyset with copy-on-write keys from another keyset.
+Whether this function will be part of the public API is a point for discussion.
+
+### Full-blown copy-on-write implementation
+
+Make Elektra's `Key` and `KeySet` data structures copy-on-write.
+This requires some major refactoring of code within `libelektra-core`.
+Code that does only interact with the data structures via the public `libelektra-core` API should not notice any differences.
+The `mmapstorage` plugin will need a major refactoring.
```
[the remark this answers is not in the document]
kodebach: > The `mmapstorage` already needs to be fixed.

Why? What exactly is broken? AFAIK it works perfectly fine, [even as the default storage plugin](https://build.libelektra.org/blue/organizations/jenkins/libelektra/detail/master/858/pipeline/1062).

The `cache` plugin on the other hand, is broken and needs to be fixed because of the `new-backend` changes. Maybe that's what you're thinking of.

> **Note** `cache` and `mmapstorage` are separate plugins. `cache` depends on `mmapstorage`, but `mmapstorage` works without `cache`. Using the terms interchangeably is confusing and wrong.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,501 @@
...
+
+    char * ukey;
+    size_t keyUSize;
+
+    uint16_t refs;
+};
+
+struct _Key {
+    struct _KeyData * keyData;
+    struct _KeyName * keyName;
```
mpranj, line 314: Some thoughts regarding moving name and data to separate structures:

1. If they are a separate entity, `mmapstorage` will need a flag once again for each of those. This is used to mark whether the data is in an mmap region or not. (or we find some bit somewhere that we can steal for this purpose)
2. Adding more indirections is probably not going to help performance. (I understand that we save memory here)

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,501 @@
...
+> To change the data you need to append an entirely new key to replace the existing one.
+> Then we just need to keep a shallow copy internally.
+
+### API restrictions
+
+@kodebach wrote:
+
+> Change the API and remove KeySet from kdbGet and kdbSet also option 4 in [the operation sequences decision](../0_drafts/operation_sequences.md).
+> If the keyset is owned by the KDB handle, it should not be as big surprise, if there is extra data in there.
+> I certainly wouldn't try to asset anything on the contents of a KeySet that I don't own directly, unless the condition is explicitly documented somewhere.
```
kodebach, line 116:
```suggestion
> I certainly wouldn't try to assert anything on the contents of a KeySet that I don't own directly, unless the condition is explicitly documented somewhere.
```

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,501 @@
...
+
+- If we delete a copied KeySet, we don't know which KeySet is the original, so we couldn't decrement the counter.
+  This could be dealt with storing a pointer to the original KeySet.
+- If the original KeySet is deleted, we don't know which other KeySets point at the data, so updating their count would not work
+- In similar fashion, if you update the original KeySet, the copied KeySets will also contain the new data (if the memory address does not change).
+  This is unexpected behaviour.
+
+```c
+struct _KeySetData {
+    struct _Key ** array;
```
mpranj, line 372: While we're making these fundamental changes...

One thing I was curious about, but I thought these kind of changes were off the table:

```c
struct _KeySetData {
    struct _KeyName ** array;
    struct _KeyData ** array;

    size_t keySize;
    size_t keyDataSize;

    [...]
}
```

In that case we could do the `mmap` pointer update using vector extensions and probably speed it up significantly.

kodebach:
> There is no need anymore to use political tricks like repeating the same opinion again and again or trying to be the loudest (like shouting "Simply not true").

1. I'm not using political tricks. I don't have an agenda beyond making my opinion heard, which is the whole point of this collaborative decision process.
2. If I repeat myself, it is almost always because I think I was misunderstood, or because I think my comment was missed entirely.
3. I don't shout anything. If I stated things like "simply not true", that is exactly what I mean. I'm already overusing things like "IMO", "IMHO" and "I think" to clearly mark opinions.

> To me it is unclear if https://github.com/ElektraInitiative/libelektra/pull/4641#issuecomment-1307498849 differs from the two MMAP approaches that are already considered alternatives and if so, how.

That is a good question. From your review I saw that you thought my suggestion was the same as an existing one. Probably also why I thought I was using "political tricks" and repeating stuff.

Maybe I misunderstood the existing option. But AFAICT all existing options keep the data in RAM between `kdbGet` and `kdbSet`. Even the "MMAP Cache with parent key" option says "always have a keyset [...] internally", which suggests "in RAM".

What I suggested would to load the data **from disk** during `kdbSet`. I thought that was made clear by "disk space is far less precious than RAM". If the "MMAP Cache" options are meant to work this way, that is not at all clear from the decision text.

doc/decisions/1_in_discussion/internal_cache.md
kodebach: Yeah, I've also though about similarly radical changes in the past, but I've never thought we could actually agree to do something as big as that.

markus2330: Big changes that'll be done in the near future by you before 1.0 are totally possible.

As performance is the lowest of our 4 goals, and memory usage was in the last benchmark a **much** bigger problem than CPU, I don't think a decision that has a focus to speed without a clear usability benefit, has any chance.

If the change is an improvement for e.g. the API user and the speed-up is only a side effect, I cannot wait for your proposal in form of a decision PR.

doc/decisions/1_in_discussion/internal_cache.md
markus2330: This is a separate decision. Can you please create one?

doc/decisions/1_in_discussion/internal_cache.md
kodebach: > Big changes that'll be done in the near future by you before 1.0 are totally possible.

Not the ones I've thought of... Even if I had the time to implement them. Basically, I've again and again come to the conclusion that our biggest limitation is the separate `Key` structure (or at least the fact that it is public API). More specifically, that `Key` is actually a whole (key,value) pair despite its name.

For example a [Radix tree](https://en.wikipedia.org/wiki/Radix_tree) (or [HAT-trie](https://en.wikipedia.org/wiki/HAT-trie)) would be prefect for `KeySet`, but it just doesn't work. You'd effectively store the name twice, because you still need the full name in `Key`.

> I think this proposal might break COW for keys.

The much bigger problem here (as alluded to above) is that there is no `Key *` in the sturcture. This immediately makes this snippet impossible

```c
Key * k = keyNew ("user:/foo", KEY_END);
ksAppendKey (ks, k);
assert (k == ksLookupByName (ks, "/foo", 0));
```

More generally: How would all the `ks*` functions that return a `Key *` work at all?
You'd have to allocate a new `Key *` or take a `Key * dest` as argument or use some fixed `Key *` that is only valid until the next call. All of which would be API changes so big and complex that you'd almost have to rewrite the whole codebase and definitely not something that could be automated.

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: I added this discussion to the decision.

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: I've added this to the discussion

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -0,0 +1,587 @@
...
+## Memory comparison of COW approaches
+
+The following calculations are based on the AMD64 platform.
+All results are in bytes unless stated otherwise.
+
+Example key: `user:/hosts/ipv6/example.com = 2606:2800:220:1:248:1893:25c8:1946`
+
+We want to measure the following properties:
+
+- Empty KeySet: size of a simple malloc of the keyset struct
```
kodebach, line 481: Either remove the KeySet column, or add a second table similar to the one for Keys that fully shows the situation for KeySets, i.e., "Empty KeySet", "Empty KeySet + extra structs", "KeySet with 1 key", etc.

The current table suggests the "full-blown COW" needs less memory for a keyset, when this is not the case. It's only the case, when the keyset shares it's keys with another.

kodebach:
I have looked at the actual problem in this decision again, and I think we have gotten far, **far** off track.

1. "Internal Cache" is a very bad name for the problem that is described
2. All the COW stuff is a neat idea, but it has very little to do with the initial problem. If we had COW, it might be a way to solve this problem and others like change-tracking, but it really is an entirely separate decision.
3. The simplest solution to the actual problem described here would be:
   1. _"More keys"_: Upon returning from `kdbGet`/`kdbSet`, set the keyname of `parentKey` to the key that actually is parent of the data that was loaded. I.e. to the mountpoint of the backend that contains `parentKey`.
   2. _"Fewer keys"_: IMO this is simply a bug in the "nothing changed" logic. I don't think we need some complicated COW scheme to solve this. I haven't had a chance to check this yet, but I think it is just a matter of copying the keys from `backendData->keys`, so they are actually there and we don't just assume they are there.

doc/decisions/1_in_discussion/internal_cache.md
atmaxinger: Agree, I added another table for the keyset.

kodebach:
> MHO it is even more unexpected if the parentKey also changes.

That is true, it is surprising, if you are used to the current situation. But we already use the value of `parentKey` to return information and in the error case we use the metadata. So `parentKey` is already an in-out type of argument. I think for someone who doesn't know the current `kdbGet`, it wouldn't be too surprising, if the name also changes.

> Suppose you use the same parentKey for kdbSet, then depending on the structure of mountpoints "outside" keys get written (or not).

But that's already the case isn't it? Either that, or not every key given to `kdbSet` is saved.

Backends are atomic. You can only save the whole thing or nothing at all. So if we have the mountpoints `system:/foo` and `system:/foo/bar/baz` and someone calls `kdbSet` with `system:/foo/bar`, we must decide between: (1) saving everything below `system:/foo` OR (2) saving only the things below `system:/foo/bar/baz`.

AFAIK we currently do (1), i.e. the same as `parentKey` to backends translation as in `kdbGet`. Therefore, I would actually add the same `parentKey` name change to `kdbSet` too.

> > "Fewer keys": IMO this is simply a bug in the "nothing changed" logic. I don't think we need some complicated COW scheme to solve this. I haven't had a chance to check this yet,
> 
> Then better don't say it yet!

I said it, because somebody else might have time to check before me... If you don't want my help and my ideas, just say so. I'll just focus on exactly what I need for my thesis and won't help with anything else.

In any case, I have now tried it and it works for some cases. The easiest mistakes should be covered, but the data is still shared and there can still be some oddities. See [`3694bb8`](https://github.com/ElektraInitiative/libelektra/pull/4573/commits/3694bb885fe04c638a59f5716db9c1124e52b60b).

> And to make this efficient and working as expected for change tracking you need COW, thus is the decision as it is.

Maybe saying "it has very little to do with the initial problem" was wrong. It is related, but you need a few extra steps. Just adding COW doesn't solve the problem. 

IMHO the whole COW stuff is definitely a separate (related) decision. Yes, one solution here is "have some form of COW". But the details of how COW is done are a separate decision. Additionally, COW is also useful for other things, so it should be separated to make linking to the decision simpler and because in theory we could implement COW for other things, but use a different solution for this problem.

Note: This would also make the decision file smaller an easier to manage.

atmaxinger:
@markus2330 I have now added some calculations about the COW approaches.

I also tried to measure allocation impact, however I'm not quite sure about the approach yet.

What I have done is to create a program allocating 100k keys 25 times for the current implementation and the full-blown COW approach.

The current implementation does a `malloc` of `struct _Key`.
The full-blown COW approach does `malloc` of `struct _Key`, `struct _KeyData` and `struct _KeyName`.

I took measurments in nanoseconds for every batch of 100k keys.

| Approach | Mean (ns) | Mean (ms) | Standard Derivation (ns) | Standard Derivation (%) |
|---|--:|--:|--:|--:|
| Current    | 11803630 | 11.8 | 1197351 | 10.14 |
| Full-blown | 23702684 | 23.7 | 1711740 | 7.22 |

Interestingly, it's only a doubling although we allocate 3 times as much.

image

kodebach:
100k keys is **a lot**. I doubt any application is going to use that many keys. Maybe try with 100, 1k, 10k as well to get some idea of how the performance scales.

But more importantly, I think we also need some idea of how long a `kdbGet` takes. Now that's much harder to measure of course, because there are much more variables at play. However, the `kdbGet` numbers are only to see, whether the `malloc` times are significant at all.

From the <25ms for 100k keys, I suspect the `malloc` time won't ever be significant. If I had to guess I would say a `kdbGet` for 100k keys will take at least 1s. 

> Interestingly, it's only a doubling although we allocate 3 times as often.

Could be CPU cache related.

But when fully using COW, it will save some allocations too. Currently every `keyCopy`/`keyDup` creates new allocations for struct, name, value and all the metadata. With COW, we would only need new allocations for the structs, the actual data would remain the same. So the relationship isn't as simple as "3 times as many allocations" anyway.

kodebach:
Also one more thing, I think some insight into the effect of the additional indirections would be nice. I'm not 100% sure, how to measure that without fully implementing the COW approach, but maybe this could be a start:

1. Allocate a bunch of keys and add names to them. So not just `malloc` for the struct, but actually `malloc` for the name put something in their (can be any string)
2. Now try to read and replace the name of **random keys** over and over again and measure that.
3. Repeat for both approaches

> **Note** That keys probably shouldn't be read/written in any fixed order, and definitely not sequentially. Otherwise, CPU cache, branch prediction and whatnot, would probably mask the effect of the indirections somewhat. Since we can't reliable say that keys will be accessed in any particular order, the benchmark should rely on something like that.

My guess is that the additional indirections will probably look similar to the `malloc` results. In isolation the change will be significant, but in absolute terms the will still be so small that it will never be significant in a whole `kdbGet`.

markus2330:
@atmaxinger fascinating work, I love it! :heart:  For me it is clear and we can merge this and go to the decision. :rocket: 

> AFAIK we currently do (1), i.e. the same as parentKey to backends translation as in kdbGet

Good point, this is actually a third bug that we can cover with this decision. ("Wrong Keys" next to "Fewer Keys" and "More Keys").

> Just adding COW doesn't solve the problem.

Yes, this is true. The decision, however, is already too long and I hope it is obvious for you how to use the cache to fix the 3 problems described in this decision.

> IMHO the whole COW stuff is definitely a separate (related) decision.

I don't think there is anyone disagreeing that what we discuss here would be much better split in more decisions with better names and more consistent style. But it is clear that it is a lot of work and it is unfair to put all the work to @atmaxinger. Please step ahead and create new decisions. I suggest:

- one about the datastructure to be used for keyset
- about how strong the isolation of backends should be, e.g. symlinks, constraints implemented in hooks etc. (already suggested earlier)

> Note: This would also make the decision file smaller an easier to manage.

Any work on this is very much appreciated :+1:

atmaxinger:
@markus2330 I can split out the COW part of this into its own decision. I'd do it right in this PR if this is okay. However, I don't think splitting it into a "KeySet" and "Key" COW decision. Those should be in the same decision.

atmaxinger:
@markus2330 I have split it out now. As the internal cache issue in it self is not of interest to me, I propose someone else takes it over, and I push the copy-on-write decision forward.

kodebach:
> I can split out the COW part of this into its own decision. I'd do it right in this PR if this is okay. However, I don't think splitting it into a "KeySet" and "Key" COW decision is a good idea. Those should be in the same decision.

Thank you, that is basically all I wanted. To be clear, I'm not trying to push work on to you, as markus2330 seems to think. I just didn't want to push commits to other people's PRs. I'm happy to take over the COW decision, if you want to do other work for change tracking in the meantime. But since markus2330 rejected my idea to start change tracking without this decision, I'm not sure how that would go...

> Please step ahead and create new decisions. I suggest:

I will create new decisions, but I too have other things to do outside of Elektra as well and I like to think about decisions before starting them. Especially, since I have multiple ideas right now and don't want to with another 10 decision PRs that are all stalled. I need to decide myself, what to propose first.

---

> Good point, this is actually a third bug that we can cover with this decision. ("Wrong Keys" next to "Fewer Keys" and "More Keys").

I'm not sure how this is a third bug or what "wrong keys" would be. I just said that `kdbGet` and `kdbSet` have the same "issue". Since backends are an atomic unit, both need to decide which backends to use. That means you cannot directly use `parentKey` as the limit.

> **Note**: You often say things like "good point" or "I agree" and then state something different from the previous comment and add some new ideas, without explaining yourself. I didn't call it "a third bug", so you should explain why you think it's "a third bug", not just expect people to know your thought process. Then we could avoid some back and forth and save both of us some time.

atmaxinger:
> I'm happy to take over the COW decision

Hmm I'd like use some of this for my thesis, so I'm not sure if it's a good idea if you do it.

atmaxinger:

kodebach:
> I'd like use some of this for my thesis

Yeah, I think you'd just be blocked waiting on the outcome for COW anyway. So probably better if

- you take COW
- the internal cache decision stays undecided until COW is done
- when COW is done I can take over internal cache, if nobody else wants to

markus2330:
@kodebach please state the reason why you still block merging this decision. I already requested a review some time ago.

For collaboration/coordination please use #4645, for the decision process there is #4639.

I think we should merge this asap, textual improvements, forgotten stuff etc. can be done in the next round. Imho the current state of the decision is further ahead than its step.

kodebach:
> please state the reason why you still block merging this decision. I already requested a review some time ago.

I'm sorry I have other things to do in my life.... Until yesterday I was waiting on whether the decision will be split in two. Then I didn't have time to review the current state. I'll post a review today.

doc/decisions/1_in_discussion/internal_cache.md
```diff
@@ -148,18 +150,27 @@ ksRemoveByName (cowMeta, "meta:/type");
...
 
-Not yet decided.
+We keep a duplicated keyset in-memory and tag the keys as copy-on-write (COW).
+From this keyset, we use `ksBelow` to return the correct keyset.
+If the user tries to change the value or metadata of these keys, the data gets duplicated.
+I.e. the original keyset is not changed.
+The name is not relevant.
+It is always read-only, because the key is in at least one keyset (the internal one).
+
+Possible copy-on-write implementations are described in [another discussion](../1_in_discussion/copy_on_write.md).
```
kodebach, line 171:
```suggestion
Possible copy-on-write implementations are described in [another decision](../1_in_discussion/copy_on_write.md).
```

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
+# Copy On Write
+
+## Problem
+
+One of Elektra's core goals is low memory usage.
+Currently, there are many places within Elektra where keys and keysets are duplicated and copied around.
+Most of those copied keys are never modified.
```
kodebach, line 7:
```suggestion
Most of those copied keys are never modified, but are required to be detached from the lifetime of the original instance.
```

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+We want to introduce an in-memory copy-on-write mechanism to lower our memory usage.
+
+In the near future, Elektra will also gain facilities for [change tracking](../0_drafts/change_tracking.md) and session recording, both of which will potentially again duplicate keys.
+There are also aspirations to create a new, simple [internal cache](../1_in_discussion/internal_cache.md) that would also benefit from a copy-on-write mechanism.
+
+## Constraints
+
+1. Elektra's public API should require no changes
```
kodebach, line 15:
```suggestion
```

Currently we do not make any API guarantees, so all changes are possible. The only restriction is that we must be able to update the whole codebase to the new API with reasonable effort. However, that is always an implied constraint, so we don't need to mention it.

A few more points regarding API changes:
1. My thesis is about migrating Elektra to a new API, so I would help with API changes
2. As part of that I'm working on a tool to automate the migrations. So if I can automate the API updates, even big API changes are possible.
3. There are already quite a few proposed API changes, including a somewhat stalled decision to change even `keyNew`. Please don't hesitate to propose something that needs big API changes. There might be a way to make it work.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+2. Modifying keys and keysets returned by `kdbGet` should work as expected
```
kodebach, line 16: What is expected? Without defining that the constraint is meaningless.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+assert (keyName(key) == keyName(key_dup)); // stays always valid
+```
+
+This is already implemented for the MMAP cache, so the implementation should be straightforward:
+Do the same COW duplications as done for MMAP but with a different flag.
+
+@kodebach wrote:
+
+> What I wanted to say is that mmap already does COW, so we can reuse the code and probably the flag.
+> If there is some code that is only needed for mmap and not COW, we could make mmap set two flags one for mmap and one for the general COW code.
```
kodebach, line 43:
```suggestion
```

Don't think that's needed

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+
+  keyString(copiedKey); // Error! Original value has been deleted. Pointer to data in copiedKey points to potentially freed memory
+  ```
+
+  The same problems in principle exist for `mmapstorage` where `kdbSet` frees (`munmap`) the keyset.
+  We can still let users access the flag `ELEKTRA_CP_COW`, we just need to clearly document what is forbidden.
+  Maybe set the `KEY_FLAG_RO_VALUE` on the original key, so that the API itself detects the error.
+  There is, however, no flag for `keyDel` that we could set.
+
+#### Changes to `libelektra-core`
```
kodebach, line 105: Unless anybody sees significant advantages in this approach over the "full-blown COW" approach, I think we can consider this a "discarded idea" and mark it as such. Unlike the `mmap` idea above, here it is not obvious that the idea has downsides, or why it is inferior to the "full-blown COW". It sounds like this idea would solve everything.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+A `ksDeepDup()` of a keyset with COW keys will create a keyset with deep-copied non-COW keys.
+Internally we may need a `ksCowDup()` function to create a keyset with copy-on-write keys from another keyset.
+Whether this function will be part of the public API is a point for discussion.
+
+### Full-blown copy-on-write implementation
+
+Make Elektra's `Key` and `KeySet` data structures copy-on-write.
+This requires some major refactoring of code within `libelektra-core`.
+Code that does only interact with the data structures via the public `libelektra-core` API should not notice any differences.
+The `mmapstorage` plugin needs to be fixed.
```
kodebach, line 173:
```suggestion
The `mmapstorage` plugin needs to be updated.
```

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+  if `dest->data->refs` == 0 (meaning no other keysets points to it):
+    pop the last key directly on the keys
+  else if `dest->data->refs` > 0: (there are other keysets sharing the same data):
+    decrease `dest->data->refs` by 1
+    create a new `dest->data`
+    copy over all but the last keys from the old `dest->data`
+  ```
+
+- `ksLookup()`:
+  - if the `DEL` or `POP` flag is specified, do the COW-stuff as described multiple times now in the operations above
```
kodebach, line 333: Those flags will be removed soon. The replacement is `elektraKsPopAtCursor` (but will be renamed).

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+- `ksCut(dest, key)`:
+  ```
+  if `dest->data->refs` == 0 (meaning no other keysets points to it):
+    perform the cutting algorithm directly on these keys
+  else if `dest->data->refs` > 0: (there are other keysets sharing the same data):
+    decrease `dest->data->refs` by 1
+    create a new `dest->data`
+    copy over the non-cut keys into `dest->data` from the old `dest->data`
+  ```
+- `ksPop(dest)`:
```
kodebach, line 321: Will be removed soon

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+    create a new `dest->data`
+    copy over all but the last keys from the old `dest->data`
+  ```
+
+- `ksLookup()`:
+  - if the `DEL` or `POP` flag is specified, do the COW-stuff as described multiple times now in the operations above
+
+#### Optimizations
+
+This approach requires more allocation than previously.
```
kodebach, line 337:
```suggestion
This approach requires more allocations than previously.
```

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+We have not benchmarked whether this is big of an issue.
```
kodebach, line 338:
```suggestion
We have not fully benchmarked whether this is a big issue.
```

There are some benchmarks below...

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -0,0 +1,474 @@
...
+One optimization could be an expanding "pool" of `_KeySetData`, `_KeyData` and `_KeyName`.
+We could then allocate multiple of them at the same time, and borrow and give back instance from and to the pool.
```
kodebach, line 340: Good idea, but I don't see how this works without major API changes. Where does the pool come from?

kodebach (approved):
Looks good, I think this can be merged now.

All the comments from the review can be addressed in the next PR. The "change tracking" and "operation sequences" decisions should probably also be moved to "in discussion". If you want to stop working on any of the decisions and let someone else take over, we should probably record somewhere that there are leftover comments here.

Also regarding the COW decision. I think we all agree that the "full blown" approach is the best idea, since it cover all the requirements and also has additional benefits (e.g. COW will be used in application code too). But I'm not yet sure, the currently proposed implementation is the best we can do. Therefore, I propose we either move the other ideas to the "Notes" section as discarded ideas, or we mark them in some other way as completely discarded. The decision would become about "How do we do full-blown COW?" instead of "What are the options for COW".
This can also be done (and discussed further in a new PR).

atmaxinger:
@markus2330 I think we can merge this now

markus2330:
@atmaxinger there is a CI problem

atmaxinger:
@markus2330 it's all green now 😄

markus2330:
Great! :heart_on_fire:

PR #4676, opened 2022-11-14
[decisions] reference counting

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


kodebach:
I reopened this decision, because I believe we need a good and uniform reference counting system for COW. The COW decision will most likely introduce multiple new structures, all of which need reference counting. The current refcount decision was made with only `Key` and `KeySet` in mind. In particular, I believe the current system is not well suited for the new structures.

For now I left the current implementation in the "Decision" section, but I expanded the description. Additionally, I added some more alternatives. I did not yet have time to consider, how each of the alternatives work with `ksAppendKey`/`ksNew` et al. w.r.t. inline `keyNew` calls. However, since both the constructor functions (#4244) and the API in general are a bit "up in the air" right now, I believe we will find a way for each of the alternatives. IIRC last time the biggest concern (against changes), was that we would either need big API changes, or would have to check a lot of code for correct use of the new implementation.

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+A `Key` can be owned by many different `KeySet`s.
+Both `Key` and `KeySet` can be used in bindings.
+If the languages of those bindings have automatic memory management, this also implies a shared ownership.
+
+To solve this issue of shared ownership, we need reference counting.
+
+An adjacent problem is that the names of `Key`s must be read-only while the `Key` is part of a `KeySet`.
+This can also be solved by a secondary counter just for `KeySet`s.
+
+There has also been [discussion](copy_on_write.md) about adding further data structures, which would need reference counters as well.
```
atmaxinger, line 15:
```suggestion
There has also been [discussion](../1_in_discussion/copy_on_write.md) about adding further data structures, which would need reference counters as well.
```

Can we refer to other decisions with the directory they are in? This makes it **much** easier to move individual decisions, as updating links is just a single find/replace for `old_folder/decision_name.md` with `new_folder/decision_name.md".

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+  elektraRelease(ks);
+}
+```
+
+## Decision
+
+General decisions regardless of implementation:
+
+- `Key` has two counters.
+  One that blocks deletion, another to prevent name changes.
```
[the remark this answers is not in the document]
kodebach: No it hasn't yet. But it was already decided that we'll use a second counter with this exact "reentrant lock" idea to automatically lock/unlock the key when it is/isn't in a KeySet. In that case, I don't think there are many alternatives, since it is the prime example use case for a reentrant lock.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: Good point, @markus2330 should add that to #4639 as well: "Links to decision should always contain the directory of the decision, even if the link is from another decision in the same directory."

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+
+typedef struct {
+  Destructor destroy;
+  unsigned int count;
+} ObjectRecord;
+
+static ObjectRecord *toRecord(void *object) { return (ObjectRecord*)(object) - 1; }
+static void *fromRecord(ObjectRecord *record) { return record + 1; }
+
+void * elektraMalloc(size_t size, ) {
```
atmaxinger, line 181:
```suggestion
void * elektraMalloc(size_t size) {
```

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+  ObjectRecord *record = malloc (sizeof (ObjectRecord) + size);
+  if (record == NULL) {
+    return NULL;
+  }
+  record->count = 1;
+  record->destroy = destroy;
```
atmaxinger, line 187: Where is `destroy` coming from?

atmaxinger: Oh that's probably the missing parameter above

doc/decisions/1_in_discussion/reference_counting.md
atmaxinger, line 181:
```suggestion
void * elektraMalloc(size_t size, Destructor destroy) {
```

doc/decisions/1_in_discussion/reference_counting.md
kodebach: > Oh that's probably the missing parameter above

Yup

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+
+// Note: should NOT call free() on pointer
+typedef void (*Destructor)(const void*);
+
+typedef struct {
+  Destructor destroy;
+  unsigned int count;
+} ObjectRecord;
+
+static ObjectRecord *toRecord(void *object) { return (ObjectRecord*)(object) - 1; }
```
atmaxinger, line 178:
```suggestion
static ObjectRecord *objectToRecord(void *object) { return (ObjectRecord*)(object) - 1; }
```

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+static void *fromRecord(ObjectRecord *record) { return record + 1; }
```
atmaxinger, line 179:
```suggestion
static void *recordToObject(ObjectRecord *record) { return record + 1; }
```

doc/decisions/1_in_discussion/reference_counting.md
atmaxinger: Wait ... is this really `object - 1`? Shouldn't it be `object - sizeof(ObjectRecord)`?

doc/decisions/1_in_discussion/reference_counting.md
kodebach: I'm gonna ignore those name change suggestions because these are just placeholders anyway. Some are even just copied from the sources.

atmaxinger: That's up to you, but in the code below you use the other names. Would make it a bit more clear if they are consistent.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: Yes, it works because of the parentheses. `(ObjectRecord*)(object - 1)` would be wrong, but by casting first and then subtracting the pointer arithmetic is done in terms of `sizeof(ObjectRecord)`. One of the weird and wonderful properties of C. See also the first answer of the StackExchange link a few lines above.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: Oh in that case, I'll change it

kodebach:
> What if we want to allocate stuff that does not need reference counting?

Could be handled by passing `NULL` for the destroy function. However, a separate function e.g. `eletktraAllocObject` could also work.

> I'd have nothing against adding just adding `uint16_t refs` fields to every new struct we need for COW, just as they are in `KeySet` and `Key` right now.

In principle I don't have an issue with that either. But what I want is a system that
1. is uniform, not e.g. some ref counts starting at 0, other at 1
2. works for all use cases, including `mmapstorage`
    If we copy the current approach, for the new structs for COW, we would need to add the flags to indicate mmap somewhere as well. We could use special values for the refcount to encode that (e.g. `UINT16_MAX`), but again that should be uniform.
3. makes sense semantically
    The current system makes sense, but it should probably called a "reentrant deletion lock" instead of a "reference counter.

One thing I dislike about the current system is that it doesn't force you to think about the refcount and almost encourages bad code. For example, you can call `keyDel` as much as you want, as long as the key is in a `KeySet`.

So take this code, which works fine with `keyDel` acting as a no-op:

```c
// external function
// code not known to caller
// docs do not state where the returned key comes from
Key * getKey(struct foo * foo) {
   return ksLookupByName (foo->ks, "/foo", 0);
}

void foo_fun(struct foo * foo) {
   Key * key = getKey (foo);
   // there could be lots of code in this function, but somewhere there is:
   keyDel (foo);
}
```

But if `struc foo` and `getKey` now change, everything breaks and we get use-after-free problems:

```c
// external function
// code not known to caller
// docs do not state where the returned key comes from
Key * getKey(struct foo * foo) {
   return foo->key;
}

void foo_fun(struct foo * foo) {
   Key * key = getKey (foo);
   // there could be lots of code in this function, but somewhere there is:
   keyDel (foo);
}
```

With a different system, where `keyDel`/`keyRelease` decrements the refcount and deletes, people would be forced to think about the reference counting more. Then it would be obvious that you only call `keyRelease`, when you called `keyRetain`, or when the docs of another function told you that you have to call `keyRelease`.

Now, I know calling `keyDel` unnecessarily is bad. But the thing is with the current system, it is very hard to notice an unnecessary `keyDel`, because it acts as a no-op.

kodebach:
BTW I will add the opinions from the previous two comments to the file, when I find time.

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
+# Reference Counting
+
+## Problem
+
+Elektra's data structures are shared between multiple owners.
+A `Key` can be owned by many different `KeySet`s.
+Both `Key` and `KeySet` can be used in bindings.
+If the languages of those bindings have automatic memory management, this also implies a shared ownership.
```
markus2330, line 8:
```suggestion
Both `Key` and `KeySet` are used in bindings, often also implying shared ownership.
```

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+
+To solve this issue of shared ownership, we need reference counting.
+
+An adjacent problem is that the names of `Key`s must be read-only while the `Key` is part of a `KeySet`.
```
markus2330, line 12:
```suggestion
An adjacent problem is that the names of `Key`s must be read-only while the `Key` is part of ordered data structures like the `KeySet`.
```

doc/decisions/1_in_discussion/reference_counting.md
markus2330: Good idea, I added d94ff201d0d6826e4a3d277fcd26c21ad8ecb792

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+This can also be solved by a secondary counter just for `KeySet`s.
+
+There has also been [discussion](copy_on_write.md) about adding further data structures, which would need reference counters as well.
+
+## Constraints
+
+- Deleting data structures that still have references must be prevented
```
markus2330, line 19:
```suggestion
- Deleting (parts of) data structures that still have references must be prevented. (Valgrind must not detect any problems.)
```

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+- The APIs for `ksAppendKey` and `ksNew` must remain ergonomic.
+  The reference counting must not prevent the inline creation of a `Key` while adding it to a `KeySet`.
```
markus2330, line 21: Better move this to assumptions for now (as long as we don't have use cases).

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Assumptions
+
+- The reference counter doesn't need to be thread-safe.
+- A reference counter of size `uint16_t` is large enough for all use cases.
+  Further restricting the counter by a small amount to allow for error value is also acceptable.
```
markus2330, line 27: The current C++ binding increments the reference counter for usage of `Key` in blocks. So in the case of recursion, `uint16_t` might not be enough.

I agree that we should assume some fixed size to be enough, maybe even `uint16_t` (and fix the C++ Binding).

markus2330 (commented):
Only looked at problem for now.

doc/decisions/1_in_discussion/reference_counting.md
atmaxinger: I mean we could go all the way and just use `size_t`, then it's physically impossible to have more than that.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: I don't like the Valgrind part... Yes, it's a simple test, but if someone finds an issue that is not detected by valgrind, that's still a problem.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: What do you mean with "as long as we don't have use cases"?

```c
keyAppendKey (ks, keyNew ("/foo", KEY_END));
ksNew (1, keyNew ("/foo", KEY_END), KS_END);
```

That's what I'm talking about. A very real use case and integral part of the current API. The API might change, so maybe we remove the specific references to the functions, but I think we all agree that we want to keep this option. The alternative would be something far more verbose like (placeholder API):

```c
Key * k  = keyNew ("/foo", KEY_END);
keyAppendKey (ks, k);
keyDel (k);

Key * k1  = keyNew ("/foo", KEY_END);
ksNew (1, k1, KS_END);
keyDel (k);
```

I also don't know how this could be phrased as an assumption.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: > The current C++ binding increments the reference counter for usage of Key in blocks. So in the case of recursion, `uint16_t` might not be enough.

Do you mean something like this:

```cpp
void foo(Key k, int depth) {
   if (depth > 0) {
      foo(k);
   }
}
```

Because that can be solved without even touching the binding (I think):

```cpp
void foo(Key & k, int depth) {
   if (depth > 0) {
      foo(k);
   }
}
```

In any case, IMO assuming `uint16_t` for now is fine. We can always change it to e.g. `uint32_t` without much trouble.

kodebach: > I mean we could go all the way and just use `size_t`, then it's physically impossible to have more than that.

I think `uint32_t` should be more than enough. I wouldn't go to `size_t`, because that is also the maximum alignment. So having a `size_t refs` and e.g. `int flags` in the same struct has more padding than with `uint32_t refs` or `uint16_t refs`.

markus2330:
> Now, I know calling keyDel unnecessarily is bad.

Why? (please elaborate in the decision) E.g. free(3) also is a NOP on NULL.

> BTW I will add the opinions from the previous two comments to the file, when I find time.

Please request a review when it was done.

> [Discussions about size uint32_t/uint16_t...]

Signed/unsigned clearly is a part of this decision to discuss. 16 vs. 32 vs. size_t needs use cases first. Also if the ref counting is even public should have a use case. If there is no public interface for the ref counters, the ref counters also could be a typedef (that people might change as they need), with a sensible default that supports all our use cases.

doc/decisions/1_in_discussion/reference_counting.md
markus2330: Valgrind will detect any problem if there is a test case for this situation. I updated:

```suggestion
- Deleting (parts of) data structures that still have references must be prevented. (Valgrind must not detect any problems for any test case.)
```

doc/decisions/1_in_discussion/reference_counting.md
markus2330: The question is if we want to support the use case to be able to construct any keyset with a single function call. (Looking at our API we probably want, but we don't have documented use cases for that.)

doc/decisions/1_in_discussion/reference_counting.md
kodebach: Still don't like it. I wasn't suggesting that valgrind might miss a memory problem (although that is possible too, e.g. when `mmap` is used). I was talking about exactly the case that no test case exists. So I suggest

```suggestion
- Deleting (parts of) data structures that still have references must be prevented. This includes, but is not limited to, running all (compatible) tests via `valgrind` without issues.
```

My issue was that just adding the `valgrind` part made it sound like just running `valgrind` without problems would be enough to fulfil the constraint. If you really want to mention `valgrind`, we need to clarify that the constraint goes beyond that.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: Still, either we leave it as a constraint or remove it completely. It doesn't make sense as an assumption. IMO assumptions are statements that simplify the problem space. This is exactly the opposite. It would be stupid of us to assume something that makes our lives more difficult.

kodebach:
> > Now, I know calling keyDel unnecessarily is bad.
> 
> Why? (please elaborate in the decision) 

Sorry, that was I typo (now fixed above too). I meant to say it **is not** bad. But that's kind of the point. It's not bad, _if_ it is unnecessary.

> E.g. free(3) also is a NOP on NULL.

I've said this before, but I will elaborate on it again (when I find time to incorporate all the other comments, in the decision too): **This example is bad**, it is a COMPLETELY different thing. In fact, we do have an equivalent to `free(NULL)` and that is `keyDel(NULL)`. I'm totally fine with that being a no-op and even without an error code.

The big difference is `NULL` is a known fixed value. `keyDel(key)` may or may not actually be a no-op. It's non-trivial to find out without doing what `keyDel` does and checking the refcount.


> Signed/unsigned clearly is a part of this decision to discuss.

I won't even consider making it signed. There is absolutely no point, the ref count can never be negative.

> 16 vs. 32 vs. size_t needs use cases first.

The general use case is already explained in the "Problem" section. What we need to decide this question are real world usage scenarios. Talking theory of what might be is useless. 

So unless you can give me a specific example, where a reference count larger than 65k could occur and there is no easy way around that, I still say let's assume `uint16_t` and change when necessary. It is the most trivial thing to change and I really don't want to spend weeks discussing potential use cases because of this.

> **Note** Even if you find a concrete example, for >65k, I seriously doubt that `uint32_t` would ever be too small. Just having 2^32 pointers on 64-bit is already 32GiB of memory and that's without any other data around that.

> Also if the ref counting is even public should have a use case.

Sentence grammar is broken and I can't tell want you wanted to say.

> If there is no public interface for the ref counters, the ref counters also could be a typedef (that people might change as they need), with a sensible default that supports all our use cases.

1. It should be a CMake config not a typedef. The size needs to be defined when `libelektra-core` is compiled and not when a header is used.
2. I'm against this idea. It just complicates everything, without a clear benefit. Any benchmark would need a footnote "with the default refcount size", and I mean **any** benchmark because some sort of refcount would always be used with Elektra. It also creates more conditions for compatibility between libraries. If a library accesses the struct data directly (e.g. `libelektra-extras`) it must not just be the same version, but also the same refcount size. Theres probably more. I don't see this as ever being a good idea.

> > BTW I will add the opinions from the previous two comments to the file, when I find time.
> 
> Please request a review when it was done.

@markus2330 Before I do change the decision file, can you please answer the question I posed in https://github.com/ElektraInitiative/libelektra/pull/4676#discussion_r1025070341

kodebach:
@mpranj This PR would have more effect on `mmapstorage` than #4683. Whether it's worth waiting for a result here, before tackling the COW changes, I don't know...

kodebach:
> it depends on the implementation.

That's what I meant. Some variants could have a lot of impact (e.g. if there is a `free` pointer, or the allocator refcounts everything). In #4683 I'm rather certain, that every variant would have a low impact on `mmapstorage`.

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+
+Additional details:
+
+- `inc` returns an error on reference count overflow
+- `dec` does not return an error on underflow, instead the count stays at zero
+
+## Rationale
+
+- Adding a second reference counter to Key and reducing the size of both significantly (`size_t` to `uint16_t`)
+  actually saves memory (32 vs 64bit on 64-bit machines) compared to the previous solution.
```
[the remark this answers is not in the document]
kodebach: This is actually a left over bit from the old version of this decision. I'll probably split out the stuff about locking the key name into a separate decision. Maybe it'll end up in #4689. The second counter looks like our current ref counter, but it really has a completely different use case.

> it's not clear without implementation detail whether this really saves memory.

In the old version the decision stated that we go from one `size_t` to two `uint16_t` right next to each other in `struct _Key`. In that case it could save memory or stay the same, but it definitely wouldn't need more.

doc/decisions/1_in_discussion/reference_counting.md
markus2330: > Do you mean something like this:

Yes.

> Because that can be solved without even touching the binding (I think):

People might use call by value `Key` and still expect this to work. Call by value is now much more popular due to move semantics.

> I think uint32_t should be more than enough

Having fixed-size types has the advantages:

- same behavior on all systems
- mmap files are more likely to be compatible

But then we need to document how big arrays etc. can be. Maybe even its own little decision.

kodebach: > due to move semantics

A move constructor shouldn't change the refcount, because it just _moves_ the reference. If it's not certain that a move constructor would be used, then (especially for a deeply recursive function) pass-by-value should not be used for stack size reasons alone.

However, I've already said that we can discuss a larger refcount in future. But the size of the field has basically no impact on the decision. We should not discuss it now. All I wanted to say with the assumption (and maybe I'll rephrase it) is that (1) `uint16_t` is the absolute minimum and (2) for most or even all cases, it also good enough.

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
...
+
+Elektra's data structures are shared between multiple owners.
+A `Key` can be owned by many different `KeySet`s.
+Both `Key` and `KeySet` can be used in bindings.
+If the languages of those bindings have automatic memory management, this also implies a shared ownership.
+
+To solve this issue of shared ownership, we need reference counting.
+
+An adjacent problem is that the names of `Key`s must be read-only while the `Key` is part of a `KeySet`.
+This can also be solved by a secondary counter just for `KeySet`s.
```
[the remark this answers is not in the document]
kodebach: Ignore these two sentences about read-only names. I don't know why I added them. It doesn't really come up in the rest of the file. It's mentioned briefly in the "Decision" and "Rationale" section but never really explained. I think I just copied that from the old decision.

**This PR** is supposed to be about the general reference counting used by `Key` & `KeySet`, currently done via `inc`/`dec`(/`del`).

For the read-only names, I'll create a separate decision. It's much simpler for that case, because a re-entrant lock is clearly the only solution.

doc/decisions/1_in_discussion/reference_counting.md
markus2330: Ok, I agree, let us keep it as constraint. We cannot write use cases for every detail.

doc/decisions/1_in_discussion/reference_counting.md
```diff
@@ -0,0 +1,286 @@
+# Reference Counting
+
+## Problem
+
+Elektra's data structures are shared between multiple owners.
+A `Key` can be owned by many different `KeySet`s.
+Both `Key` and `KeySet` can be used in bindings.
+If the languages of those bindings have automatic memory management, this also implies a shared ownership.
+
+To solve this issue of shared ownership, we need reference counting.
```
markus2330, line 10: This is AFAIK already solved in the current API. See also https://github.com/ElektraInitiative/libelektra/pull/4805#discussion_r1083989993

doc/decisions/1_in_discussion/reference_counting.md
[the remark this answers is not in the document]
markus2330: > This PR is supposed to be about the general reference counting used by Key & KeySet, currently done via inc/dec(/del).

We already have reference counting, so the decision must be about what is wrong in what we have now.

doc/decisions/1_in_discussion/reference_counting.md
markus2330: Please simply change the underlying file(s).

doc/decisions/1_in_discussion/reference_counting.md
markus2330: > However, I've already said that we can discuss a larger refcount in future.

I don't think there is any sensible way to change anything in refcounting after 1.0.

doc/decisions/1_in_discussion/reference_counting.md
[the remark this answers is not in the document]
markus2330: We talked talked past each other when we discussed this. I never thought of the second ref counter as good idea but argued that we should try to simplify the problem (e.g. stay read-only, stay locked etc.)

markus2330 (commented):
Some comments to resurrect this decision (at least the problem should be clarified or decided that there is no problem.)

doc/decisions/1_in_discussion/reference_counting.md
kodebach: Well, if there is no `keyGetRefCount()` function then I don't see an issue. Changing the type of the refcount would be a purely internal change with no affect on the API. It would still change behaviour, so the question is how strict we want to be, but I'd argue it would just be a minor `x.1.0` change and not a major breaking change.

If there is a `keyGetRefCount()` then changing the return type would be a breaking change. But we could technically do a hacky solution like:

```c
uint16_t keyGetRefCount(Key * key) {
    if (key->refcount > UINT16_MAX) return UINT16_MAX;
    return key->refcount;
}

uint32_t keyGetRefCountLarge(Key * key) {
    return key->refcount;
}
```

All this is to say: For the IMO _highly unlikely_ case that `uint16_t` turns out to be too small, there are ways to go forward. So IMHO using `uint16_t` is fine. If it is possible without increasing the struct sizes, we could also use `uint32_t`. But I really wouldn't sacrifices extra bytes for every `struct _Key`, just of the hypothetical case where `uint16_t` isn't enough.

doc/decisions/1_in_discussion/reference_counting.md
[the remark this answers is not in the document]
kodebach: Oh okay, you should've been more clear on that.... The old decision was already in "Partially implemented" and had "add second counter to Key" in the "Decision" section.

I don't think this is the right place to discuss it. Since this PR is not at all about the locking of keynames. (I will remove this note about two counters). The keyname locking stuff should be discussed in the next PR for https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/0_drafts/readonly_keynames.md

doc/decisions/1_in_discussion/reference_counting.md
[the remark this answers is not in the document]
kodebach: As noted in https://github.com/ElektraInitiative/libelektra/pull/4805#discussion_r1085399408, this I introduced this decision as a means to re-examine the current refcounting implementation and to explore the different options, since there was no documentation on why the current solution was chosen.

As always "rejecting" the decision is an option. Although in this case for documentation purposes, I would actually still move the decision to "implemented" and just choose the current solution as the "Decision".

So in short: I'm not saying there is necessarily a problem with the current implementation. All I want is that _we_ (not just myself) think about whether another option could be better, _without_ always falling back to "we already have an implementation". I have absolutely no issue, if we decide "the current implementation is good enough" or "another option is better, but not so much that the effort of changing it is worth it". However, before we make that decision, we _have to_ discuss all options equally.

doc/decisions/1_in_discussion/reference_counting.md
kodebach: See https://github.com/ElektraInitiative/libelektra/pull/4676#discussion_r1088275260

kodebach:
@markus2330 As hinted in our meeting today: The goal of this decision is mostly to document the status quo of reference counting and also explore the other options. IMO the decision is low priority, and in the end it will probably come down to keep things the way they are now. Even if that might not be the best solution, it works and changing it is too much effort.

Regarding the part about "locking key names": That should be removed from this decision. It's a separate issue that needs a separate decision.

Issue #4645, opened 2022-11-08
Finalize decisions


markus2330:
- [x] uses cases of Elektra's API
- [x] #4717 @kodebach
- [x] #4744 @kodebach 
- [x] #4678
- [x] #4689
- [x] change tracking @atmaxinger  
- [x] 0_drafts/notifications.md @atmaxinger 
- [x]  0_drafts/operation_sequences.md @atmaxinger 
- [x] #4641
- [x] #4716  @kodebach
- [x] #4745 @kodebach
- [ ] 1_in_discussion/spec_expressiveness.md 5_implemented/default_values.md @tmakar 
- [ ] #4715  @kodebach
- [ ] 0_drafts/commit_function.md @flo91 (no decision PR yet)
- [ ] 0_drafts/man_pages and maybe improvements of 5_implemented/error_message_format.md @hannes99  (no decision PR yet)
- [ ] ksCut vs. ksFindHierarchy?
- [ ] symlinks+context lookup
- [ ] "binary" metadata

markus2330:
This issue is to foster collaboration between @mpranj, @kodebach and @atmaxinger for restructuring core data structures. Together we can tackle a quantum leap not possible if everyone works alone.

The [currently proposed idea](https://github.com/ElektraInitiative/libelektra/pull/4641#issuecomment-1307634773), with some extensions from me is, that
- first @atmaxinger implements key/keyset changes to allow efficient change tracking, and then
- @kodebach does the needed refactorings of API
- @mpranj fixes the mmap cache/storage, with improvements as discussed in store_name.md.

We want to avoid that anyone needs to do redo time-consuming fixes (like fixing the cache or doing API renames), so we must clarify the basics in decisions first.

I am looking forward on seeing this becoming reality! :heart:

kodebach:
The first question, before we can make any plans is:

@mpranj do you have the time to work on this right now? If not, there is no point in trying to involve you.

mpranj:
I think I can make time.

The fixes to `mmapstorage` regarding just keynames here are probably a smaller undertaking than the new backend stuff.

Is there a bigger task here that I am overlooking?

kodebach:
The changes from `store_name.md` are not much, no. But depending on the outcome of #4641 and all the other decisions spawned from change tracking, there could be bunch more changes needed. I wouldn't underestimate the time requirement, especially since it is still entirely unclear what we even want to do.

markus2330:
[...]

PR #4678, opened 2022-11-14
[decision] copy-on-write (round 2)

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


kodebach (changes requested):
Not much to review here...

Like I said before, I think the decision should now be about "How to do full-blown COW". We need to discuss:
- How will the structures look like?
- What is needed to make `mmapstorage` work?
- (related to #4676) How do we do the reference counting?
- Is there even any point in having non-COW copies, or can we just always do COW?

One concrete question I also have:
- Looking at the currently proposed structs, should `KeySet * meta` in `struct _Key` really be allocated separately? Would it make more sense to embed the struct directly? Now that the structs are smaller this makes a lot more sense IMO.

atmaxinger:
> I  think the decision should now be about "How to do full-blown COW"

Yes.

> How will the structures look like?

There are already some proposals in the file. Fell free to directly comment on them.

> What is needed to make mmapstorage work?

Good point.

> (related to https://github.com/ElektraInitiative/libelektra/pull/4676) How do we do the reference counting?

Agree, that needs to be addressed. Honestly I'd just do the same what we now do for `Key` and `KeySet`, because form experience we'll again need a month to come to an agreement on how to properly do reference counting.

Changing the way we do reference counting afterwards should not be a big deal.

> Is there even any point in having non-COW copies, or can we just always do COW?

My plan is to always do COW. I can't think of a good argument to have both COW and non-COW copies. 
Semantically, COW copies behave exactly the same as non-COW copies, with the added benefit of lower memory consumption.
The copying itself is also faster with COW.
The only (most likely irrelevant) performance hit is when making changes to a copied key.

> Would it make more sense to embed the struct directly? 

I guess that depends on how often metadata is used. If we know that the majority of keys is gonna have metadata, then yes. Embedding the metadata KeySet might gain us some performance. 

However, if we don't use metadata often, it will lead to unnecessary memory consumption. 

But IMHO this is some preliminary optimization that could easily be added after COW is already implemented and we have done some benchmarks on the real implementation.

kodebach:
> Changing the way we do reference counting afterwards should not be a big deal.

Agreed, under one condition: You can't copy the current solution exactly, or if you do you have to be careful how you use it.

> I'd just do the same what we now do for `Key` and `KeySet`

Yes, we can copy it. But for the COW structs it will never make sense to have a refcount of zero. That would mean they exist standalone, but are not used anywhere.

So, if you copy the current system, I think you should do:

```c
static void keySetValue(Key * key, void * value, size_t size) {
  // [...] removal of current value from key
  struct _KeyData data = keyDataNew (value, size);
  keyDataIncRef (data);
  key->data = data;
}

static void keyDel(Key * key) {
  keyDataDecRef (key->data);
  keyDataDel (key->data);
  // [...] other cleanup
}
```

Basically, always increment the refcount after creation and always decrement before deletion, so that the refcount is never zero.

> Semantically, COW copies behave exactly the same as non-COW copies

Not 100% true, but IMO the only case where it is relevant doesn't matter that much anyway.

```c
Key * key;
struct foo myFoo = {
  .x = 0
};
keySetBinary (key, &myFoo, sizeof(myFoo));

Key * dup = keyDup (key);

((struct foo *)keyValue (key))->x = 1;

// with COW
assert (((struct foo *)keyValue (dup))->x == 1);
// without COW
assert (((struct foo *)keyValue (dup))->x == 0);
```

The solution is simple. Clear documentation that changing the data stored in a `Key` without calling `keySetValue` could affect other keys sharing the same data. Both using `keySetValue` to update the data and doing `keySetValue (keyValue (key))` and then changing data directly would be fine.

In addition we could add a function like `keyDetach` (outside the core; better name welcome), which would create separate copies of all COW structs for the key.

> I guess that depends on how often metadata is used.

True. This definitely would need benchmarks. I think because of the padding/alignment requirements, embedding the struct would actually use less memory, fewer allocations and fewer indirections, IF metadata is used. If there is no metadata, the pointer of course uses less memory, but only 8 bytes (on 64-bit). The question is really, how much effect the extra indirections/allocations have on performance.

The actual configuration keys used in applications will almost certainly have metadata, because of most applications will have a specification. However, there will also be temporary keys and keys used for other purposes (`parentKey`, `errorKey`, etc.).

In short probably, best to keep it is a separately allocated pointer for now, but definitely worth mentioning as a potential future optimization.

atmaxinger:
All of them good points, I added them to the decision.

atmaxinger:
All of them good points, I added them to the decision.

> function like keyDetach (outside the core; better name welcome), 

Why would you add it outside of the core? IMHO this is basic functionality that should be in the same library as the datastructures themselves.

It may even depend on private fields that are opaque to libraries outside of the core. (I know that currently all of Elektra basically just includes `kdbprivate.h` but to me this seems like a big antipattern).

kodebach:
> Why would you add it outside of the core? IMHO this is basic functionality that should be in the same library as the datastructures themselves.

Assuming I haven't missed anything, the function would just be:

```c
// WARNING: pseudo-code, actual code slightly more complicated
static inline void keyDetach (Key * key) {
    keySetName (key, keyName (key));
    keySetValue (key, keyValue (key));
    // maybe do something for metadata, but I don't think it's needed
}
```

One option would be to add a `static inline` function like above to a public header, assuming the function is actually this simple and we are 100% certain the detaching will always work like that.

Otherwise, I'd put it outside the core (*), because it is not minimal. The caller can just manually do the `keySetName`, `keySetValue` calls.

(*) eventually, probably into `libelektra-extras`, but for now you can put it into the core or ease. Lots of things will need to be moved anyway.

> I know that currently all of Elektra basically just includes kdbprivate.h but to me this seems like a big antipattern

See #4246, and the decisions [library_split.md](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/3_decided/library_split.md), (maybe) [library_directory_structure.md](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/3_decided/library_directory_structure.md) and also [private_api.md](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/3_decided/private_api.md)

kodebach:
> There are already some proposals in the file. Fell free to directly comment on them.

Can't comment on the code, because there are no changes there...

But one thing I think definitely needs to be considered, is whether we need different structs for key name and value. Once the escaped name is gone, both are just a pointer, a size and a refcount. Both behave exactly the same w.r.t. COW and refcounting. I think we probably can use a single struct, which would also allow creating internal APIs for the refcounting and COW stuff with less duplication.

So I'd propose:

```c
typedef struct {
    ElektraBuffer buffer;
    uint16_t refs;
} RcBuffer;

struct _Key {
    RcBuffer * uname;
    RcBuffer * ename; // will be removed ASAP
    RcBuffer * value;
    // [...]
}
```

See also #4683 for `ElektraBuffer`, although that is optional here. Could be separate `void * data` and `size_t size` fields too.

atmaxinger:
> Can't comment on the code, because there are no changes there...

Oh no, who thought it was a good idea to restrict this? 😥

> Once the escaped name is gone, both are just a pointer, a size and a refcount.

True, but when is "once"? I really want to avoid another `new-backend` situation. I'd much rather keep the new implementation as close to the current datastructures as possible. If someone then wants to remove the escaped name we can still refactor it. IMHO small(ish) iterations are better than trying to do everything at once in the same branch. That's why I also advocate for keeping the public API as unchanged as possible. (Adding a `keyDetach` function is OK from my POV)

kodebach:
> True, but when is "once"?

It should be ASAP. But AFAIK nobody has started with it yet.

> I really want to avoid another new-backend situation. [...]

That's exactly why I proposed a separate `RcBuffer * ename` inside the `struct _Key`. Then removing it is very simple and we still can use a single `struct` for everything. However, it was just a suggestion. Maybe the duplication for having different structs for COW is not so bad, especially if we go with #4683. 

> That's why I also advocate for keeping the public API as unchanged as possible

There are a few public API changes that have already been decided (like removing the escaped name). I agree changes should be either kept to a minimum _or_ be combined with the already decided changes.

atmaxinger:
I added the variant with RcBuffer to the decision file. If we still have two names, it has worse memory usage than the "normal" full-blown implementation.

If we remove the escaped name from both, they are (suprise, suprise) exactly the same. 


The `RcBuffer` can only be used for `Key`, as `KeySetData` other requirements. For example, we need the hashmap also, and we need to store how much is allocated and how much elements there are.

atmaxinger:
@markus2330 do you have any inputs for this decision?

kodebach:
> I added the variant with RcBuffer to the decision file.  [...]

I wouldn't worry about memory consumption with the escaped name. It has been decided to remove the escaped name. The only question is when and how exactly we do it. So for memory consumption questions I'd assume there is only one name.

> The `RcBuffer` can only be used for `Key`, as `KeySetData` other requirements. For example, we need the hashmap also, and we need to store how much is allocated and how much elements there are.

There would be ways of resusing it with some casting and allocation tricks, but I don't any benefit there.

atmaxinger:
> The only question is when and how exactly we do it. 

Yes, this is a good question. Because AFAICT it requires huge changes to the overall codebase, so this not something that can be done in parallel.

> There would be ways of resusing it with some casting and allocation tricks, but I don't any benefit there.

Please let's not make this more complicated than it has to be. The code should be maintainable after all.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -12,9 +12,6 @@ There are also aspirations to create a new, simple [internal cache](../3_decided
 
 ## Constraints
 
-1. Elektra's public API should require no changes
```
markus2330, line 15: I think we should add an constraint that the life-time of KeySets also must be as expected. (And not be related to kdbClose).

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -327,12 +324,101 @@ API notes:
 - `ksLookup()`:
   - if the `DEL` or `POP` flag is specified, do the COW-stuff as described multiple times now in the operations above
 
-#### Optimizations
+#### Reference Counting
+
+We need reference counting for the internal COW datastructures.
+For the beginning, we can just do it the same way reference counting currently works for `Key` and `KeySet`.
```
markus2330, line 330: Please avoid such statements: we don't want to do something "For the beginning" and later differently.

So we also need to decide reference counting first.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -327,12 +324,101 @@ API notes:
...
+};
+keySetBinary (key, &myFoo, sizeof(myFoo));
+
+Key * dup = keyDup (key);
 
-This approach requires more allocations than previously.
-We have not fully benchmarked whether this is a big issue.
-One optimization could be an expanding "pool" of `_KeySetData`, `_KeyData` and `_KeyName`.
-We could then allocate multiple of them at the same time, and borrow and give back instance from and to the pool.
+((struct foo *)keyValue (key))->x = 1;
```
markus2330, line 391: This is undefined behavior, the memory returned by `keyValue` must not be modified. This can be simply added as constraint.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -327,12 +324,101 @@ API notes:
...
+
+// with COW
+assert (((struct foo *)keyValue (dup))->x == 1);
+// without COW
+assert (((struct foo *)keyValue (dup))->x == 0);
+```
+
+This edge case can be accounted for by providing a function, i.e. `keyDetach`, that forces that the key has its very own copy of the data.
```
markus2330, line 399: I don't think we need this.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -327,12 +324,101 @@ API notes:
...
+
+- This approach requires more allocations than previously.
+  We have not fully benchmarked whether this is a big issue.
+  One optimization could be an expanding "pool" of `_KeySetData`, `_KeyData` and `_KeyName`.
+  We could then allocate multiple of them at the same time, and borrow and give back instance from and to the pool.
+
+- Embed the `KeySet * meta` directly in `struct _Key`.
+  This may help with performance in cases we need metadata.
+  It will, however, increase memory usage.
+  This should only be considered after some benchmarking shows this is a real issue.
```
markus2330, line 421: Exactly, I don't think this needs further investigations now.

markus2330 (changes requested):
The variants seem to be so similar, is this even something that we need to decide now?

Please improve the problem section, so that it is clear what we decide here.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -351,11 +437,12 @@ We want to measure the following properties for the key:
...
+|:------------------------------------------------------------------|----------:|----------------------:|-----------------------------:|-------------------:|--------------------------:|---------------------------:|
 | Current Implementation                                            |        64 |                    64 |                           64 |                153 |                       306 |                        459 |
 | mmapstorage-like COW implementation (without additional pointers) |        64 |                    64 |                           64 |                153 |                       217 |                        281 |
 | mmapstorage-like COW implementation (with additional pointers)    |        80 |                    80 |                           80 |                169 |                       249 |                        329 |
 | Full-blown COW implementation                                     |        32 |                    72 |                           96 |                185 |                       217 |                        249 |
```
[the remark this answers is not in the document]
atmaxinger: The have a name ... It's the same as the section titles in "Considered Alternatives"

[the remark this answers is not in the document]
markus2330: I understood that this was the idea but I cannot navigate easily between the different variants. Maybe a table of contents would also solve my issue.

atmaxinger:
> The variants seem to be so similar

What? There are two completely different approaches to do COW. 
One that tries to fit into the mmapstorage semantics, and one that is actually a full-blown copy-on-write implementation without any constraints.

 > is this even something that we need to decide now?

IIRC we actually decided on doing the full-blown approach in the old PR. This decision was just split out for documentation purposes.

The only thing we still need to decide is @kodebach s idea of using the same datastructure for key name and key value. But if the escaped name will be removed soon (TM), there is actually no difference between the "vanilla" and "Variant 1" of the full-blown COW implementation.

markus2330:
> I think the decision should now be about "How to do full-blown COW". We need to discuss:

I fully agree. Detailed technical documentation of non-fundamental decisions can directly be in contrib/, no decision needed.

>   What is needed to make mmapstorage work?

This is maybe only an issue. If I understood @mpranj correctly, he already cannot determine what to fix without trying to fix it. So probably "damage" done to mmapstorage/cache is not a constraint. @mpranj please tell us if you found constraints.

>   Is there even any point in having non-COW copies, or can we just always do COW?

Yes, this should be clearly answered in the decision. I also don't think there is a need for non-COW. If the user pays for the indirection anyway, it should be always COW.

> Changing the way we do reference counting afterwards should not be a big deal.

If this is really the case, we don't need to decide this now and simply use current ref counting. Imho we also shouldn't mix public visible decisions with internal structures in one decision. And I agree we should avoid large PRs.

But: we should first decide if we need writable key names at all, lets discuss this in #4645.

markus2330:
> IIRC we actually decided on doing the full-blown approach in the old PR. This decision was just split out for documentation purposes.

Documentation later goes to doc/contrib/ anyway.

> The only thing we still need to decide is @kodebach s idea of using the same datastructure for key name and key value.

Read-only key names are also open, see #4645.

> But if the escaped name will be removed soon (TM), there is actually no difference between the "vanilla" and "Variant 1" of the full-blown COW implementation.

This is what I meant with "The variants seem to be so similar". I didn't say the considered alternatives are so similar.

doc/decisions/1_in_discussion/copy_on_write.md
```diff
@@ -12,9 +12,6 @@ There are also aspirations to create a new, simple [internal cache](../3_decided
 
 ## Constraints
 
-1. Elektra's public API should require no changes
-2. Modifying keys and keysets returned by `kdbGet` should work as expected
```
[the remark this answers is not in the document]
kodebach: I think it was removed, because I said in the last PR that "expected" is too vague. Without stating what is expected, this constraint is meaningless, everybody could add their own expectations.

doc/decisions/1_in_discussion/copy_on_write.md
kodebach: Again, what is "expected"?

I'd say: "The liftime of a `Key` and a `KeySet` must be unaffected by COW".

doc/decisions/1_in_discussion/copy_on_write.md
kodebach: I don't think it is technically undefined behaviour, but the cast from `const void *` returned by `keyValue` to `struct foo *` is suspicious. However, this code _currently_ works. Therefore, the decision should address it.

doc/decisions/1_in_discussion/copy_on_write.md
kodebach: The function is not just for this case, but for the general case, when you want a non-COW copy for whatever reason. However, I did already say that I'd probably put the function into `libelektra-extras` or similar, if we add it at all, since it would just be something like:

```c
// WARNING: pseudo-code, actual code slightly more complicated
static inline void keyDetach (Key * key) {
    keySetName (key, keyName (key));
    keySetValue (key, keyValue (key));
    // maybe do something for metadata, but I don't think it's needed
}
```

kodebach:
> > What is needed to make mmapstorage work?
> 
> This is maybe only an issue

That is very surprising to me. I already know multiple different approaches for making `mmapstorage` work. For example, some of the approaches in #4676 could be used.

IIRC @mpranj was only talking about the need for extra flags to mark keys `mmapstorage` and that the plugin itself needs to be adapted for the new structure. However, this does not address the issue that `mmapstorage` keys become invalid currently during `kdbSet`, but at the latest during `kdbClose` (with the current plugin). To make the data last longer, there would have to be a callback from `keyDel` (or the equivalent function for the COW stuff) back into `mmapstorage` to tell it when data is no longer needed.



> > Is there even any point in having non-COW copies, or can we just always do COW?
> 
> Yes, this should be clearly answered in the decision.

AFAIK it already is, that's why `keyDetach` is there.



> Imho we also shouldn't mix public visible decisions with internal structures in one decision. 

Hard to separate in some cases. Also I think at some point it just becomes to much bureaucracy. When the decision is done an implemented, we can move the publically visible and the internal details into different documentation. But during the discussion it is much easier, if everything is in one place.

mpranj:
> address the issue that mmapstorage keys become invalid currently during kdbSet, but at the latest during kdbClose (with the current plugin).

I'm not sure what you mean with invalid here. Do you mean cache invalidation due to updated keys in kdbSet?
Other than that: mmapstorage keys have been the same as any other keys, just with different memory management.

> If I understood @mpranj correctly, he already cannot determine what to fix without trying to fix it. So probably "damage" done to mmapstorage/cache is not a constraint. @mpranj please tell us if you found constraints.

I was trying to say that it's simpler to just start fixing it, instead of going through it theoretically, as most of it is just implementation detail. I was not saying that it is impossible to determine what to fix.

If we do change the internal data structures it makes much more sense to fix the cache and mmapstorage afterwards (or in tandem). Otherwise I'll be rewriting it at least twice.

The most important constraint for mmap is this: any structure (or bytes) that is an allocation unit (e.g. we malloc() the bytes needed for KeySet struct, so this is an unit) needs to have a flag to determine whether those bytes are actually malloc()ed or they are mmap()ed.
Thus all the newly added structures as proposed will need some kind of an mmap flag.

kodebach:
> I'm not sure what you mean with invalid here. 

I mean, `munmap` is called during `kdbSet`, or if we change the plugin at the latest during `kdbClose`. Unless I'm completely wrong accessing the memory after `munmap` is called is undefined behaviour and the memory might contain arbitrary data.

> The most important constraint for mmap is this: any structure (or bytes) that is an allocation unit (e.g. we malloc() the bytes needed for KeySet struct, so this is an unit) needs to have a flag to determine whether those bytes are actually malloc()ed or they are mmap()ed.
> Thus all the newly added structures as proposed will need some kind of an mmap flag.

Not entirely true. We don't need a specific flag field. We just need **some way** to know whether the key was allocated via `mmap` or `malloc`. More precisely (and even more general), we need to know _how_ to resize/realloc or free the memory.

See below for some ideas.

> Do we need the flag in the newly added structures themselves, or would it suffice if we have that flag on the key? We already can store on the key whether the name or data is in MMAP or not.

If you are sure, you can keep that in sync, it may work. But I'm not 100% certain it works, and I think there are better ways.

---

I have hinted at some ideas for this stuff in #4676.

1. If we have `free` function-pointer along side the refcount, `mmapstorage` (and also other plugins with different allocators) could set it to their own implementation. To mimic the current behaviour of `mmapstorage` this would point to a no-op function. However, we could also improve things and keep track of when all data has been freed and only then call `munmap`.

2. Another simpler way to avoid the flag, which doesn't really allow for further improvements, would be using the refcount. `mmapstorage` could set the refcount to a value that is otherwise illegal. This would allow us to detect the keys. Depending on the refcount implementation good values would probably be `0` or `UINT16_MAX`. The special value would have to ignored by **all** refcounting functions (inc, dec, del) and turn the functions into no-ops.


If we implement a generic refcounting approach, I'd go with option 1 here.
If we want to implement COW first, I'd go with option two.

mpranj:
> I mean, munmap is called during kdbSet

No, it is not.

> or if we change the plugin at the latest during kdbClose. Unless I'm completely wrong accessing the memory after munmap is called is undefined behaviour and the memory might contain arbitrary data.

`mmapstorage` only calls `munmap` in some error cases, so basically `munmap` is almost never done and the keyset is never invalidated.

When the `mmapstorage` was designed/implemented, not all structures had refcounters, so there was no way to know when a munmap is safe. This was simply out of scope at that point in time.

If refcounting is now implemented for all structures, we might be able to properly `munmap` in future.

mpranj:
>     If we have free function-pointer along side the refcount, mmapstorage (and also other plugins with different allocators) could set it to their own implementation. To mimic the current behaviour of mmapstorage this would point to a no-op function. However, we could also improve things and keep track of when all data has been freed and only then call munmap.

This sounds elegant, but probably needs too many bytes as we need an additional pointer.

>     Another simpler way to avoid the flag, which doesn't really allow for further improvements, would be using the refcount. mmapstorage could set the refcount to a value that is otherwise illegal. This would allow us to detect the keys. Depending on the refcount implementation good values would probably be 0 or UINT16_MAX. The special value would have to ignored by all refcounting functions (inc, dec, del) and turn the functions into no-ops.

A bit hacky, but it would save quite some bytes.
IMHO it makes sense in general that we minimize the size of the internal data structures.

I think the best is if we decide on which variant will be used regarding the COW approach. Then we can iron out little details like whether we use a flag or the refcount.

https://github.com/ElektraInitiative/libelektra/blob/626286e8d023a28036854476b44410d6105f35d1/doc/decisions/1_in_discussion/copy_on_write.md?plain=1#L359-L366

@atmaxinger kudos for finding this abstraction. Sounds a bit less readable, but I am in favor of this, as it is super simple.

Please not though that due to the separation of the structures we also lose 48 bits to padding per RcBuffer structure. (on 64 bit machines, if I am not mistaken?)
Since we have this structure 2-3x per key this is not insignificant..

atmaxinger:
> @atmaxinger kudos for finding this abstraction. Sounds a bit less readable, but I am in favor of this, as it is super simple.

Well this was the idea of @kodebach 

> I think the best is if we decide on which variant will be used regarding the COW approach. Then we can iron out little details like whether we use a flag or the refcount.

Good idea. I have actually already implemented the "vanilla" approach (without RcBuffers) today because that was easier to integrate into the existing code - the field names stay the same, so it was just a matter of adding `->keyData`, `->keyName` or `->data` to where they were used.

Now this is done, we can refactor it to however we like.

While implementing, I also found some funny edgecases with the opmpmHashmap and COW that we probably would never have thought of ... will add some insights to the decision tomorrow.

kodebach:
Regarding `munmap`: This call doesn't look like an error case to me. Maybe it's just written a bit oddly or I missed something...

https://github.com/ElektraInitiative/libelektra/blob/e9f2cab6f743f63746bfb800a3b4d14a29fe7df6/src/plugins/mmapstorage/mmapstorage.c#L1599

> This sounds elegant, but probably needs too many bytes as we need an additional pointer.

> Please not though that due to the separation of the structures we also lose 48 bits to padding per RcBuffer structure. (on 64 bit machines, if I am not mistaken?)

> Since we have this structure 2-3x per key this is not insignificant.

With COW it is no longer as simple as n-times per key. Many of the keys we have are created by `keyDup`, or are for example meta-keys copied by `spec`. I would assume with the de-duplication from COW we would have quite a bit less memory usage. Then the padding might not matter that much (we also can't really do much about it, unless we e.g. switch from `size_t` to `uint32_t`).

For the extra function pointer, I actually think the added possibilities in terms of functionality, balance out some of the memory impact.

However, we definitely need benchmarks to come to a conclusion here.

---

AFAICT there is only one really open question here: The #4689 thing. However, from my POV the only impact that would have is that COW for names wouldn't be needed, because they can't change. Instead names could just be shared directly, of course still via a ref-counted buffer. That means we'd just remove some code, not actually change anything. Therefore, I think the read-only name stuff can be done later, if we want to.

All in all, I would say (unless @markus2330 still has some input) this decision can be moved to "Decided" (with "Full-blown COW" being the "Decision"). The finer implementation details can be discussed in a PR with the actual code, where we can see the changes and test them.

mpranj:
> Regarding munmap: This call doesn't look like an error case to me. Maybe it's just written a bit oddly or I missed something...

Ah, sorry. I understand why this looks confusing, but this mmap/munmap pair in kdbSet is irrelevant:

During kdbSet the storage plugins always write to a temp file, due to how the resolver works.
We also don't need to mmap the temp file here: when doing kdbSet we already have the KeySet at hand, mmap-ing it is not needed at all, because we have the data. We just want to update the cache file.
The mmap/munmap in kdbSet are just so we can write the KeySet to a file in our format. (`mmap()` is just simpler, but we could also `malloc()` a region and then `fwrite()` the stuff)

Therefore the **only** case where we return a mmap()ed KeySet should be in kdbGet.

> All in all, I would say (unless @markus2330 still has some input) this decision can be moved to "Decided" (with "Full-blown COW" being the "Decision"). The finer implementation details can be discussed in a PR with the actual code, where we can see the changes and test them.

I tend to agree.
Maybe it is possible to also revise/update a decision if new knowledge comes up during implementation. @markus2330 does it make sense for you regarding #4639, to have some form of feedback loop, so we are able to update decisions?

markus2330:
I am with you: I also like the decision full-blow COW best. I think it has huge potential to speed up many places, with reduced memory consumption. See also my portability concerns above, which is IMHO the main argument against mmap as internal cache.

> Maybe it is possible to also revise/update a decision if new knowledge comes up during implementation.

Absolutely, updates in decisions are very much welcomed and then you do not need to do the whole process again. I clarified in de35d9b2c9c986b2f8e4c0d5dc2f52750a0f8613. If you only want to update text without changing the direction of decisions, no decision PR is needed. Simply update the decisions along with other PRs. Such a non-decision PR modifying decisions already exists, see #4690.

To get this PR #4678 merged what was said here must end up in documentation text of this PR. I don't think that workings of mmap cache were widely known before. For now it is okay it is put in the decision, later it should be in `doc/contrib`, though.

@atmaxinger please request my review when everything is updated.

doc/decisions/1_in_discussion/copy_on_write.md
atmaxinger: No, we need this. During the implementation I had to create such functions for both Key and KeySet to avoid code duplication. We can argue whether or not they should be part of the public API though.

kodebach:
> Ah, sorry. I understand why this looks confusing, but this mmap/munmap pair in kdbSet is irrelevant:

Ah, so the important point is: The `mmap`ed data returned my `kdbGet` is _never_ `munmap`ed. It stays valid until process death.

I assume this is fine, because even if multiple `kdbGet` cause multiple `mmap`s the OS will deduplicate that?

> > Maybe it is possible to also revise/update a decision if new knowledge comes up during implementation.
> 
> Absolutely, updates in decisions are very much welcomed and then you do not need to do the whole process again.

I actually think, the implementation details should be move out of the decision file somewhere into `doc/contrib`, as soon as an implementation exists. Any minor future changes that are found necessary should then be explained there. This avoids the whole "is this PR modifying a decision a decision-PR" question.

Only when the actual decision is called into question, i.e. "would a different approach have been better", should the decision file be modified. And this is clearly a new decision-PR.

doc/decisions/1_in_discussion/copy_on_write.md
kodebach: To come to a conclusion here:

We can leave the note in the file. It just says this would be a possible solution for the edge case. It doesn't say that we will introduce this.

Whether we do need it can be discussed in the PR implementing everything, where we see the code. If we really do need it, we can then discuss whether we make it public API and via which library.

markus2330:
```suggestion
This edge case can be accounted for by providing a private function `keyDetach()`, that forces that the key has its very own copy of the data.
```

Yes, let us introduce it as private API and we can later decide if it must be public if we know more about the use cases of Elektra's API.

markus2330:
>  I actually think, the implementation details should be move out of the decision file somewhere into doc/contrib, as soon as an implementation exists. Any minor future changes that are found necessary should then be explained there.

I fully agree. As the mmap cache already exists, it could be in doc/contrib. But as it is/gets broken now, it is a gray area. In case of doubt it is better in a decision and @mpranj can then rewrite to proper contrib docu.

>This avoids the whole "is this PR modifying a decision a decision-PR" question.

You cannot really avoid touching the decisions during/after the implementation. What we plan to do should be in decisions. Plans obviously will not be exactly of what later is. Only what already is should be in doc/contrib.

kodebach:
I agree with all you said. What I wanted to say is: We now have an idea of what we plan to do and what needs fixing. We don't know the exact details, but there is no point in working out all the fixes in theory, when directly iterating the code until it works would be easier.

So what I propose is, merge this decision (more or less) as-is. _In the PR that adds the implementation_, move the implementation details to `doc/contrib` and update as necessary.

Note, this is specific to _this decision_, not my general opinion. It is different here, because the decision is very low-level and needs to discuss unusually many implementation details upfront.

kodebach (approved):
IMO this is decided. Whether we need to merge it and open a 3rd PR just to move it to `3_decided` or not to fulfil the formalities, is for @markus2330 to decided.

atmaxinger:
So in what phase is this decision now? Stil "in discussion"?

atmaxinger:
@markus2330 I think I addressed all of your concerns (except for the TOC one, I'm not sure how I'd do this tbh).

I also documented the stuff @mpranj told us about the `mmapstorage` plugin, and how it interacts with a possible copy-on-write implementation.

markus2330:
Great job. I also think it is decided, so let us move the decision there in the next PR.

PR #4683, opened 2022-11-15
[decision] Buffer structs

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/decisions/0_drafts/buffer_struct.md
+++ b/doc/decisions/0_drafts/buffer_struct.md
@@ -1,0 +1,189 @@
+# Buffer Struct
+
+## Problem
+
+For this decision a buffer is a fixed-length sequence of arbitrary bytes.
+In other words, it is a pointer to data combined with the size of this data.
+
+Many parts of Elektra use such buffers, but there is currently no uniform API for all of these.
+
+## Constraints
+
+## Assumptions
+
+## Considered Alternatives
+
+### Public Buffer struct
+
+Introduce a new struct that is part of the public API and therefore declared in a public header.
+The struct should be usable everywhere, so it should be declared by `libelektra-core`.
+
+```c
+typedef struct {
+    void * data;
+    size_t size;
+} ElektraBuffer;
+```
+
+> **Note**: It is not clear yet, whether `void * data` or `const void * data` is preferrable.
+
+This struct can be embedded into other structs, e.g.:
+
+```c
+struct _Key
+{
+	ElektraBuffer name;
+    ElektraBuffer value;
+
+    KeySet * meta;
+	keyflag_t flags;
+	uint16_t refs;
+};
+```
+
+The public API for accessing the data could remain unchanged:
+
+```c
+const char * keyName (ElektraKey * key) {
+    return key->name.data;
+}
+
+const char * keyNameSize (ElektraKey * key) {
+    return key->name.size;
+}
+```
+
+It could also return the buffer in a single function.
+Either by value:
+
+```c
+const ElektraBuffer keyName (ElektraKey * key) {
+    return key->name;
+}
+
+void foo(ElektraKey * key) {
+    const ElektraBuffer name = keyName (key);
+    printf("size %zd, first part: %s", name.size, (const char*)name.data);
+}
+```
+
+or as a pointer:
+
+```c
+const ElektraBuffer * keyName (ElektraKey * key) {
+    return &key->name;
+}
+
+void foo(ElektraKey * key) {
+    const ElektraBuffer * name = keyName (key);
+    printf("size %zd, first part: %s", name->size, (const char*)name->data);
+}
+```
+
+A [short investigation](https://godbolt.org/z/r8q7YfqaW) into the generated assembly code indicates that the pointer version is likely superior.
+
+The buffer struct could also be used as an argument in the public API.
+For simple functions that would only take one such `ElektraBuffer` argument this there is little benefit.
+Especially on the caller side, the code becomes less readable:
+
+```c
+void fooNoBuffer(void * data, size_t size);
+void fooWithBuffer(ElektraBuffer buffer);
+void fooWithBufferPtr(ElektraBuffer * buffer);
+
+void caller(void) {
+    size_t size1 = 42;
+    void * data1 = malloc(size1);
+
+    fooNoBuffer(data1, size1);
+    fooWithBuffer((ElektraBuffer){.data = data1, .size = size1});
+    fooWithBufferPtr(&(ElektraBuffer){.data = data1, .size = size1});
+
+    ElektraBuffer buffer; // buffer from somewhere else
+    fooNoBuffer(buffer.data, buffer.size);
+    fooWithBuffer(buffer);
+    fooWithBufferPtr(&buffer);
+
+    ElektraBuffer * bufferPtr; // buffer pointer from somewhere else
+    fooNoBuffer(bufferPtr->data, bufferPtr->size);
+    fooWithBuffer(*bufferPtr);
+    fooWithBufferPtr(bufferPtr);
+}
+```
+
+However, for functions that need multiple buffer-like inputs the API should be more user-friendly:
+
+```c
+Key * keyNew(const void * name, size_t nameSize, const void * value, size_t valueSize);
+Key * keyNewWithBuffer(const ElektraBuffer name, const ElektraBuffer value);
+Key * keyNewWithBufferPtrs(const ElektraBuffer * name, const ElektraBuffer * value);
+```
+
+Furthermore, new APIs could be introduced that operate on such buffer structs.
+For example, extensive key name manipulation APIs could be created:
+
+```c
+typedef ElektraBuffer ElektraKeyname;
+
+const char * elektraKeynameEscape(const ElektraKeyname * keyname);
+int elektraKeynameUnescape(const char * escapedName, ElektraKeyname * unescapedName);
+
+void elektraKeynameAddPart(ElektraKeyname * keyname, const char * part);
+void elektraKeynameAddEscaped(ElektraKeyname * keyname, const char * escapedSuffix);
+
+const char * elektraKeynameLastPart(const ElektraKeyname * keyname);
+
+// examples below assume these functions exists
+ElektraKeyname * elektraKeyname(ElektraKey * key);
+void elektraKeySetName(ElektraKey * key, const void * name, size_t size);
+
+void caller(ElektraKey * key) {
+    printf("name: %s\n", elektraKeynameEscape(elektraKeyname(key)));
+
+    ElektraKeyname newName;
+    if (elektraKeynameUnescape("system:/foo/new", &newName) == 0) {
+        elektraKeySetName(key, newName.name, newName.size);
+    }
+
+    ElektraKeyname * name = elektraKeyname(key);
+    assert (strcmp(elektraKeynameLastPart(name), "extra") == 0);
+
+    elektraKeynameAddPart(name, "extra");
+    assert (strcmp(elektraKeynameLastPart(name), "extra") == 0);
+
+    elektraKeynameAddEscaped(name, "even/more");
+
+    assert (strcmp(elektraKeynameLastPart(name), "more") == 0);
+    assert (strcmp(elektraKeynameEscape(name), "system:/foo/new/extra/even/more") == 0);
+}
+```
+
+The main advantage of these APIs would be that they work both independently of a `Key` and with names on the stack.
+
+### Private Buffer struct
+
+Same as above, but the struct would be private and opaque.
+Since the struct is so simple, this would have very few benefits.
+However, it has a few downsides.
+
+If the struct should be used in the public API and provided by the user, it would have to be separately allocated on the heap via a constructor.
+Additionally, we'd need accessor functions to manipulate the data in a buffer struct.
+
+### Keep current situation
+
+Pointer to data and size as separate fields in structs.
+
+This of course means, no struct for the public API and separate accessors for data pointers and data sizes.
+
+## Decision
+
+## Rationale
+
+## Implications
+
+## Related Decisions
+
+- The buffer struct could be embedded in the various [Copy On Write](../1_in_discussion/copy_on_write.md) structures.
+  By using a common struct, internal APIs for common task (e.g. resizing) can be created.
+
+## Notes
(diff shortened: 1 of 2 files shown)


doc/decisions/0_drafts/buffer_struct.md
```diff
@@ -0,0 +1,189 @@
+# Buffer Struct
+
+## Problem
```
kodebach, line 3: > **Note**: The problem section is very sparse, because there isn't really a huge problem here. This is more of a proposal than decision for a solution to a problem.

doc/decisions/0_drafts/buffer_struct.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+The public API for accessing the data could remain unchanged:
+
+```c
+const char * keyName (ElektraKey * key) {
+    return key->name.data;
+}
+
+const char * keyNameSize (ElektraKey * key) {
```
atmaxinger, line 51:
```suggestion
const size_t keyNameSize (ElektraKey * key) {
```

doc/decisions/0_drafts/buffer_struct.md
```diff
@@ -0,0 +1,189 @@
...
+void foo(ElektraKey * key) {
+    const ElektraBuffer * name = keyName (key);
+    printf("size %zd, first part: %s", name->size, (const char*)name->data);
+}
+```
+
+A [short investigation](https://godbolt.org/z/r8q7YfqaW) into the generated assembly code indicates that the pointer version is likely superior.
+
+The buffer struct could also be used as an argument in the public API.
+For simple functions that would only take one such `ElektraBuffer` argument this there is little benefit.
```
atmaxinger, line 86:
```suggestion
For simple functions that would only take one such `ElektraBuffer` argument there is little benefit.
```

doc/decisions/0_drafts/buffer_struct.md
```diff
@@ -0,0 +1,189 @@
...
+    void * data1 = malloc(size1);
+
+    fooNoBuffer(data1, size1);
+    fooWithBuffer((ElektraBuffer){.data = data1, .size = size1});
+    fooWithBufferPtr(&(ElektraBuffer){.data = data1, .size = size1});
+
+    ElektraBuffer buffer; // buffer from somewhere else
+    fooNoBuffer(buffer.data, buffer.size);
+    fooWithBuffer(buffer);
+    fooWithBufferPtr(&buffer);
```
atmaxinger, line 105: Not sure about this. We have to copy the buffer internally or else the data is corrupt when `buffer` goes out of scope.

atmaxinger (commented):
In general, I think Elektra could benefit from a unified buffer datastructure.

There are some things that we should discuss before of its introduction. For example, who owns the created buffers, i.e. who is responsible for deallocating them.

Also how would such buffers fit in with reference counting.

As for the usage in COW, I think we should try to not do everything at once here. Let's first implement COW as close to the current implementation as possible, and then we can still refactor it to using these buffers. We should try to avoid another `new-backend` situation, where there is just 1 huge branch where everything happens all at once.

doc/decisions/0_drafts/buffer_struct.md
kodebach: Totally depends on what `fooWithBufferPtr` does. Take this for example:

```c
void fooWithBufferPtr(ElektraBuffer * buffer) {
    buffer->size = 0;
}
```

Not sure when this could be used, but it is totally possible to have a "correct" `fooWithBufferPtr`.

Of course a `keySetName` with a buffer would need to make a copy. But `keySetName` already makes a copy, so that's nothing new.

The example is intentionally vague, because the main point was showing that `fooNoBuffer` is clearer.

kodebach:
> who owns the created buffers, i.e. who is responsible for deallocating them.

That depends. In most cases, I think the struct itself will either be embedded in another struct, or it will be stack allocated. In both cases, there is no question of ownership. If there are cases, where structs are allocated on their own, the ownership would have to be defined on a case by case basis.

As for the data within the buffer, the answer is simple: Whoever would own the data, if the `ElektraBuffer` struct did not exist. For example, the key name and value are owned by the `Key` they belong to (Note: currently, without COW). 

> Also how would such buffers fit in with reference counting.

In general, `ElektraBuffer` is independent of reference counting. Instances could be stack allocated, where reference counting makes no sense. In other cases the struct will be embedded in another struct. This other struct may have reference counting. For example: https://github.com/ElektraInitiative/libelektra/pull/4678#issuecomment-1315807895

> Let's first implement COW as close to the current implementation as possible, and then we can still refactor it to using these buffers. We should try to avoid another new-backend situation, where there is just 1 huge branch where everything happens all at once.

We should at least discuss, what the long-term plan is. The last thing I want is, one PR that changes all the `key*` and `ks*` functions to implement COW and then another one that changes the same functions to implement `ElektraBuffer` (or the other way arround).

I agree that another `new-backend`-like branch wouldn't be good. But the changes here should be much more contained. Ideally only internals of `libelektra-core` change. The first PR that introduces `ElektraBuffer` should definitely not touch the public API.

doc/decisions/0_drafts/buffer_struct.md
```diff
@@ -0,0 +1,189 @@
+# Buffer Struct
+
+## Problem
+
+For this decision a buffer is a fixed-length sequence of arbitrary bytes.
+In other words, it is a pointer to data combined with the size of this data.
+
+Many parts of Elektra use such buffers, but there is currently no uniform API for all of these.
```
markus2330, line 8: This by itself is not a problem. We should not provide an API for everything (e.g. strings, float serialization, etc.) but only for our use cases #4686.

But I agree that this area has problems and improvements are possible. Please specify what you aim for:

- optimizations? the size could be part of the allocated buffer, then it wouldn't need space if the buffer is NULL.
  Also the size field could grow as needed (with a bit indicating if further bytes also hold the size).
- less bugs? buffer resizing etc. is error-prone and we basically have no tests in this area

Also the urgency should be described. I assume it would be good to have for COW? @atmaxinger what are your thoughts on that?

kodebach: I know this is not really a problem (see https://github.com/ElektraInitiative/libelektra/pull/4683#discussion_r1023212527). I would have just made a proposal issue, but I wanted to avoid the inevitable "make a decision PR, please".

I have to ask, have you read the rest of the file, or are you waiting for a "clear problem"? If you haven't read the file, I would really encourage you to do so. There isn't really a clear problem here, but I think having the `ElektraBuffer` struct would make many things better.

I think when you read the whole document and see how much effort is put into the different options, it also becomes much more obvious that this is just a proposal.

> atmaxinger what are your thoughts on that?

He already left a review, see https://github.com/ElektraInitiative/libelektra/pull/4683#pullrequestreview-1181540342

atmaxinger: > the size could be part of the allocated buffer

Honestly I think this adds a lot of complexity for a very minor gain. 

> buffer resizing etc. is error-prone 

I don't see how this would make it any better. It just groups the size of the buffer with the buffer itself. Maybe we could provide reusable methods for buffer resizing for this, then the logic only has to be implemented and tested once. But that would require that all usecases for buffer resizing are the same.

> I assume it would be good to have for COW

I'm not sure whether we gain much. Functionality wise, it has not impact whatsoever. The only way it would matter if theses buffer structs would also have reference counting, then we could just use those. But as @kodebach mentioned, reference counting is orthogonal to this. So I don't see a big correlation between this an COW.

kodebach: I tried to ignore the ideas for "what you aim for", because there is no goal, but some further comments:

1. Size as part of buffer: _Always_ doing this, is absolutely **NOT** a goal. It completely breaks many of the use cases I had in mind with the public struct.
    However, there are options, e.g.:
    ```c
    ElektraBuffer * elektraBufferAlloc(size_t size) {
        if (size == 0) return NULL;
        ElektraBuffer * buffer = malloc (sizeof (ElektraBuffer) + size);
        buffer->size = size;
        buffer->data = buffer + 1; // pointer arithmetic in terms of sizeof(ElektraBuffer)
        return buffer;
    }
    ```
    I _intentionally_ did not mention this in the file, because buffers created like this are a special use case of `ElektraBuffer` and providing such an API would just lead to confusion. However, for COW -- because it would all be private API -- we could do _roughly_:

    ```c
    typedef struct {
       ElektraBuffer buffer;
       uint16_t refcount;
    } RcBuffer;

    static RcBuffer * rcBufferAlloc(size_t size) {
        if (size == 0) return NULL;
        RcBuffer * rc = elektraMalloc (sizeof (RcBuffer) + size);
        rc->buffer.size = size;
        rc->buffer.data = rc + 1; // pointer arithmetic in terms of sizeof(RcBuffer)
        rc->refcount = 1;
        return rc;
    }
    
    void keySetName (Key * key, void * name, size_t size) {
        rcBufferRelease(key->name);
        key->name = allocRcBuffer(size);
        memcpy (key->name->buffer.data, name, size);
    }
    ```
2. Buffer resizing: Again not intentionally omitted. Private allocation/resizing APIs using `ElektraBuffer` are of course possible. But for a public API this could be confusing, because with the public struct this is possible:
   ```c
    char data[] = "some data";
    ElektraBuffer buffer = {.data = data, .size = sizeof(data)};
   ```
   In such a case using a resizing API `resizeBuffer(&buffer)` would obviously fail. The only way I can see having such things as public API would be with e.g.
   ```c
   typedef ElektraBuffer ElektraHeapBuffer;
   ```

3. COW: As stated, there are some ways of making the internal private APIs simpler and maybe removing some duplication. But I agree that this can definitely be done after COW is done and we shouldn't delay COW because of this decision.

doc/decisions/0_drafts/buffer_struct.md
markus2330: Especially abstractions should have a use case and a clear problem. E.g. avoiding code duplication is a common use case for abstraction. If it doesn't solve a problem we definitely shouldn't do it.

markus2330:
> We should at least discuss, what the long-term plan is.

We have clear guidelines for a decision process. Discussions, long-term plans, etc. are without commitment and were usually waste of time in the past.

kodebach:
> > We should at least discuss, what the long-term plan is.
> 
> We have clear guidelines for a decision process. Discussions, long-term plans, etc. are without commitment and were usually waste of time in the past.

Please don't take comments out of context and don't try to invalidate responses that were not directed at you. Instead try to provide some constructive feedback for the actual proposal.

I responded to atmaxinger's suggestion to implement COW as close to the current non-COW implementation and then do further refactors. With long-term plan I just meant the how to proceed with the currently proposed stuff. I wasn't talking about long-term as in "the coming years" or "post-1.0". It was just "long-term" in contrast to "right now".

Also, note that the comment was written **before** #4691 was opened. Now that #4691 exists and the implementation of COW is almost done, I agree with atmaxinger to do it like this and refactor later. Even if that will touch much of the same code again.

doc/decisions/0_drafts/buffer_struct.md
kodebach: Can you please just read my proposal and give an opinion instead of wasting both our time with criticising the process.... Why do you think I added this note when I created the PR? I know this doesn't fit the decision process. But if I had created a proposal issue, you'd have complained as well...

You can always invent some problem or come up with some abstract or even hypothetical use case. I even tried to come up with something instead of just leaving the section empty. You still found a reason to ignore the proposal.

doc/decisions/0_drafts/buffer_struct.md
kodebach: > This by itself is not a problem. We should not provide an API for everything

I've just read our comment again. Maybe this will help, if it does, please confirm:

```suggestion
Many parts of Elektra use such buffers, but there is currently no uniform way of using them throughout the codebase.
```

To clarify further (if I have to, I'll add something like this to the file too):
- With API I **did not** mean public API. If I did I would have written "public API".
- Yes, I do suggest possible uses in the public API and I believe using these structs in the public API would be good. But that is not what this decision is about. Stating a problem for that would have been much simpler. This decision is just about the introduction of a uniform way of handling buffers, because currently we handle buffers in whatever way the author of a particular piece of code felt like.
- This is not just about string-like buffers. It's more about the fact that C doesn't remember the size of an array. The decision could also be applied to the `Key *` array in a `KeySet`. Whether or not that would be a good idea, is a different question. But there are enough string-like buffers that the decision still makes sense without it.

doc/decisions/0_drafts/buffer_struct.md
markus2330: I already read the problem several times: it is simply not convincing that anything should be done. I would concentrate our efforts on something that brings benefit for Elektra's user and not do internal restructuring just to have a more uniform API, i.e. less internal functions (which is currently the only reason given). Having a more generic internal API/less functions by itself **isn't better**, hence no problem and hence nothing to do. It would be very different if you want to improve the public API...

I already gave you several hints which problems might be worth solving with buffers.  If you cannot add any problem in the problem section, rejecting it as the obvious choice. More comments or notes will not help as this does not lead anywhere. Please don't reply unless you have **questions** how to proceed.

Please let us focus on problems we all agree exist, there are more than enough of those. We should be happy if something works well and no extra work is needed, shouldn't we?

Btw. in any case I appreciate that you wrote this, even if we reject it as there is **currently** no immanent benefit. Then we have documented how we could introduce buffers if they are needed at some time (e.g. public API is clumsy to use).

kodebach: > I already read the problem several times

**Did you read the rest?**

I know this is not the protocol for a decision. But I clearly stated that this is more of a proposal and doesn't fit the "we have a problem, let's decide a solution" process...

> It would be very different if you want to improve the public API...

**That is a goal here**

There are some examples further down in the file.... I just didn't want to design a whole public API based on a public buffer struct, only to find out, e.g. "no we don't want public structs".

> Then we have documented how we could introduce buffers if they are needed at some time (e.g. public API is clumsy to use).

Why do you think I wrote this? The current API is okay, but with #4715 anything that needs a keyname will become pretty annoying, if we need to pass a size everywhere.

> Please don't reply unless you have questions how to proceed.

Okay, I have a question: How can I convince you to be less stubborn and get you to just spend maybe five minutes reading the whole file?

kodebach:
@markus2330 This decision was mostly just an idea. I formed the basis of `ElektraName` and `ElektraValue` in #4805. Although I think none of the options here match exactly what I ended up with. If we do keep `ElektraName` and `ElektraValue` in the new API, we need some version of this decision. But until that is clear, this decision is very low priority (I labelled the PR).

kodebach:
I don't think this decision is useful anymore

PR #4689, opened 2022-11-17
[decision] Read-only key names

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/decisions/0_drafts/readonly_keynames.md
+++ b/doc/decisions/0_drafts/readonly_keynames.md
@@ -3,7 +3,8 @@
 ## Problem
 
 It might be a good idea to prevent changes to the keyname after creation.
-Currently, key names are automatically read-only when they are contained in a keyset.
+Currently, the name of a key is automatically made read-only, when it is currently part of or at some point in the past was part of at least one `KeySet`.
+There are discussions about making key names writable again, when the key is removed from all `KeySet`s.
 
 There may also be situations where changing the name of a key after its creation is required.
 In some situations, it may be wise to reuse a key instead of deleting it and creating a new one.
@@ -22,7 +23,9 @@ In some situations, it may be wise to reuse a key instead of deleting it and cre
 
 ## Considered Alternatives
 
-### Alternative A
+### Separate API for keynames
+
+Use the proposed `ElektraBuffer` struct to create a separate API for keynames independent of the `Key` API.
 
 ### Alternative B
 
--- a/doc/decisions/0_drafts/readonly_keynames.md
+++ b/doc/decisions/0_drafts/readonly_keynames.md
@@ -27,7 +27,41 @@ In some situations, it may be wise to reuse a key instead of deleting it and cre
 
 Use the proposed `ElektraBuffer` struct to create a separate API for keynames independent of the `Key` API.
 
-### Alternative B
+### Read-only keynames
+
+The key name should be permanently read-only after creation.
+
+TBD: How would we dynamically create keynames with this approach?
+
+### Re-entrant lock for the key name
+
+Since we assume a single threaded context, this can be implemented as a simple counter.
+
+```c
+struct _Key {
+    // [...] other stuff
+   uint16_t nameLock; // if zero, name is writable, otherwise name is readonly
+};
+
+void keyLockName (Key * k) {
+    k->nameLock++;
+}
+
+void keyUnlockName (Key * k) {
+    if (k->nameLock == 0) return;
+    k->nameLock--;
+}
+
+void ksAppendKey (KeySet * ks, Key * k) {
+    keyLockName (k);
+    // [...]
+}
+
+void ksRemove (KeySet * ks, elektraCursor cursor) {
+    keyUnlockName (ks->array[cursor]);
+    // [...]
+}
+```
 
 ### Alternative C
 
--- a/doc/decisions/0_drafts/readonly_keynames.md
+++ b/doc/decisions/0_drafts/readonly_keynames.md
@@ -45,3 +45,5 @@ In some situations, it may be wise to reuse a key instead of deleting it and cre
 - []()
 
 ## Notes
+
+- [Issue 2202](https://issues.libelektra.org/2202) talks about how unexpected it is that keys will be readonly once in a keyset, but they don't get unlocked when removing them from a keyset.
--- a/doc/decisions/0_drafts/readonly_keynames.md
+++ b/doc/decisions/0_drafts/readonly_keynames.md
@@ -1,0 +1,47 @@
+# Readonly Keynames
+
+## Problem
+
+It might be a good idea to prevent changes to the keyname after creation.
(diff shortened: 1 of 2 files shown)


atmaxinger:
Okay so I added a barebones decision, beacuse I'm not quite sure about the problem myself.

doc/decisions/0_drafts/readonly_keynames.md
```diff
@@ -0,0 +1,49 @@
...
+
+It might be a good idea to prevent changes to the keyname after creation.
+Currently, key names are automatically read-only when they are contained in a keyset.
+
+There may also be situations where changing the name of a key after its creation is required.
+In some situations, it may be wise to reuse a key instead of deleting it and creating a new one.
+
+## Constraints
+
+1. `keyAddName` et al. still have to work up to a certain point
```
kodebach, line 13:
```suggestion
```

doc/decisions/0_drafts/readonly_keynames.md
```diff
@@ -0,0 +1,49 @@
+# Readonly Keynames
+
+## Problem
+
+It might be a good idea to prevent changes to the keyname after creation.
+Currently, key names are automatically read-only when they are contained in a keyset.
```
kodebach, line 6:
```suggestion
Currently, the name of a key is automatically made read-only, when it is currently part of or at some point in the past was part of at least one `KeySet`.
There are discussions about making key names writable again, when the key is removed from all `KeySet`s.
```

doc/decisions/0_drafts/readonly_keynames.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Assumptions
+
+1.
+2.
+3.
+
+## Considered Alternatives
+
+### Alternative A
```
kodebach, line 25:
```suggestion
### Separate API for keynames

Use the proposed `ElektraBuffer` struct to create a separate API for keynames independent of the `Key` API.
```

Intentionally not very detailed, because the buffer decision is still in progress.

kodebach (changes requested):
I absolutely do not think this is the correct moment for this decision. Apart from the fact that I don't see why this would decision would have any real benefits. There are much better ways of achieving similar goals.

As stated in #4683 I would create an API for keynames independent of the key API. However, I would not make the name inside a `Key` readonly. In fact as you can see in #4683, I would make it arguably _less_ read-only because the name is no longer returned as a const-pointer.

As I've proposed a long time ago, I not use the read-only flag for the automatic locking. I'd use a separate counter, or more precisely a re-entrant lock.

doc/decisions/0_drafts/readonly_keynames.md
```diff
@@ -0,0 +1,49 @@
+# Readonly Keynames
+
+## Problem
+
+It might be a good idea to prevent changes to the keyname after creation.
+Currently, key names are automatically read-only when they are contained in a keyset.
+
+There may also be situations where changing the name of a key after its creation is required.
+In some situations, it may be wise to reuse a key instead of deleting it and creating a new one.
```
markus2330, line 9: This was the original hope/idea. But with COW I doubt that reusing a key is worth the trouble, simply doing `keyDup` seems to be the natural choice without much performance/memory impact?

So, I suggest as "Considered Alternatives" that the key name should be permanently read-only after creation. If we don't have a better idea with the `keyLock` API.

markus2330: The benefit is the simplicity: all what you describe above is not needed (no 2nd ref counter etc.).

This is also the problem to be described here: the problem of how to find out that the key name can be changed again. I already mentioned #2202 which describes the problem in more detail.

Rationale for my proposal: `Key` from its name already says that the key name is its identity. So to get a different name, you need to duplicate, as a key with a different key name must be a different object of `Key`.

kodebach: I get the general idea and I see how a read-only name would seem simpler in theory. However, it is a huge semantic and API change. It would probably be one of the most tedious things to implement, because you need to fix every newly broken piece of code.

The issue is that by making the key name totally read-only, a huge chunk of possible use cases becomes illegal. In contrast, automatic locking and unlocking of the key name _creates_ new use cases, but keep all existing ones.

If you really think this would be a good idea, then please think it through and write up a "Considered Alternative". I think anybody other than you sees how this approach would work. For example, how do you create a key with a dynamic name? How do you rename a key? Basically, how would you replace this:

```c
const char * mp;
Key * k = keyNew ("system:/elektra/mountpoints", KEY_END);
keyAddBaseName (k, mp);

Key * k2 = keyDup (ksLookupByName (ks, "/foo", 0)); 
keySetName (k2, "/bar");
ksAppendKey (ks, k2);
```

I hope you're not suggesting to go back to basic `strcat` like operations.

> This is also the problem to be described here: the problem of how to find out that the key name can be changed again

I don't see where the problem is here. The concrete use case we have is: The name of a `Key` must not change, while the `Key` is part of at least one `KeySet`.

I think the real issue here is that we use the term "second reference counter". This was wrong. That's not what it is (mainly because the first refcounter is already a bit non-standard). What we actually want is a re-entrant lock for the key name. Since we assume a single threaded context, this can be implemented as a simple counter.

Maybe describing it like this would help:

```c
// Note: error handling omitted for brevity

struct _Key {
    // [...] other stuff
   uint16_t nameLock; // if zero, name is writable, otherwise name is readonly
};

void keyLockName (Key * k) {
    k->nameLock++;
}

void keyUnlockName (Key * k) {
    if (k->nameLock == 0) return;
    k->nameLock--;
}

void ksAppendKey (KeySet * ks, Key * k) {
    keyLockName (k);
    // [...]
}

void ksRemove (KeySet * ks, elektraCursor cursor) {
    keyUnlockName (ks->array[cursor]);
    // [...]
}
```

This looks totally clear to me. It also makes it clear that any other structure/caller/etc. could use `keyLockName`, `keyUnlockName` to lock the name as long as they need to.


> Rationale for my proposal: `Key` from its name already says that the key name is its identity.

You **do not** want to go down this road.... I've thought about this before and there is a reason I never brought it up. The only thing this path leads to is the conclusion that a `Key` should not exist outside a `KeySet`. Thinking of keys that way means a key is an entry in a `KeySet` and should be fully and solely owned by one `KeySet`. In essence a `Key` would be what a `Map.Entry` is to a `Map` in Java.

Sounds good in principle and would totally be how I'd design the API from the start. However, making this change now is simply impossible. It totally annihilates many of our API pattern. You can't have a `parentKey` or an `errorKey`, because those by definition exist standalone and in isolation from `KeySet`s.

atmaxinger:
@markus2330 I have updated the alternatives with all mentioned alternatives. I'd say we merge this and let @kodebach take over this decision.

markus2330:
I interpreted "thumbs up" from kodebach that it is okay to merge, even though there were still changes requested.

PR #4697, opened 2022-11-21
Use cases core & KDB

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


doc/usecases/kdb/UC_load_config.md
```diff
@@ -0,0 +1,31 @@
...
+- **Main success scenario:**
+  - Application requests current configuration from KDB
+  - Elektra loads from backends
+  - Elektra returns configuration
+- **Alternative scenario:**
+  - Application requests current configuration from KDB
+  - Elektra detects configuration hasn't changed and is already present in memory (of current instance)
+  - Elektra returns previously loaded configuration
+- **Error scenario:**
+  - Application requests current configuration from KDB
```
markus2330, line 23:
```suggestion
```

doc/usecases/kdb/UC_store_config.md
```diff
@@ -0,0 +1,28 @@
+# Use Case: Storing application configuration
+
+## Summary
+
+- **Title:** Storing application configuration
```
markus2330, line 5: I don't think this use case is needed. "Modifying application configuration" is enough.

doc/usecases/kdb/UC_modify_config.md
```diff
@@ -0,0 +1,28 @@
...
+- **Actors:** Application, Elektra
+- **Brief:** Application modifies configuration
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:**
+  - [Application loads configuration](UC_load_config.md)
+  - Application modifies configuration in memory
+  - [Application stores configuration](UC_store_config.md)
```
markus2330, line 17: You can be more low-level, e.g. I don't see any problem to say which functions get called:

```suggestion
  - `kdbSet()` is called immediately afterwards.
```

doc/usecases/kdb/UC_modify_config.md
```diff
@@ -0,0 +1,28 @@
...
+- **Alternative scenario:** -
+- **Error scenario:** -
+  > **Note**: The error scenarios from the references use cases still apply.
+  > However, the in-memory modification part itself has no error scenario.
+- **Postcondition:**
+  - The modified configuration MUST be stored into the KDB exactly.
```
markus2330, line 23: Here we come to the interesting discussions. E.g. this would influence what `spec` needs to do:

```suggestion
  - The modified configuration MUST be stored into the KDB exactly.
    Metadata from the specification gets added to the configuration.
```

@tmakar

doc/usecases/kdb/UC_validate_config.md
```diff
@@ -0,0 +1,37 @@
+# Use Case: Ensuring valid configuration
```
markus2330, line 1: I think this is simply a post-condition of "Modifying application configuration".

markus2330 (commented):
Great job! This draft already provides insightful and interesting use cases.

> but could you please look at this PR and tell me, if this is what you meant with use cases?

It is exactly what I meant. These use cases should have been written years ago!

Good that you are finally on it! :heart: 

Looking forward to more :rocket:

E.g.:

- KeySet as generic data structure
- config shared between applications (fallback, override, ...)
- maybe something about modifying specifications (specload etc.)

doc/usecases/kdb/README.md
```diff
@@ -0,0 +1,7 @@
+# Use cases for KDB
+
+This folder contains the use cases for the Key Database (KDB).
```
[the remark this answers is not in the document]
kodebach: I intentionally did not mention API. IMO it should be rare that a use case talks about API at all, most use cases should be on a more abstract, higher level.

doc/usecases/kdb/README.md
```diff
@@ -0,0 +1,7 @@
...
+These primarily are the basis for `libelektra-kdb`, but may also influence other things like, e.g., the high-level API `libelektra-highlevel`.
```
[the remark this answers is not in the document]
kodebach: Obvious to you and me, but not to somebody new, who may start with reading use cases. In general it would also be possible to have a completely different API that isn't built on another existing (public) API.

doc/usecases/kdb/UC_load_config.md
kodebach: Why? If the Elektra is a library not a server/daemon, if there is no request/call from an application nothing at all happens.

doc/usecases/kdb/UC_modify_config.md
kodebach: Again IMO a use case is inherently abstract and high-level. Mentioning a specific function is definitely out of scope. Everything would be the same, if we renamed `kdbSet` to `foofoo`, so why mention it? The information that `kdbSet` implements this use case should be in the documentation for `kdbSet`.

doc/usecases/kdb/UC_modify_config.md
kodebach: > Metadata from the specification gets added to the configuration.

What do you mean here?

1. `spec` should **remove** the metadata it added during `kdbSet`, so that it is **not stored** with the configuration
2. If specification is used at all here (which the use case intentionally doesn't mention because a specification is not required), then it would have to exist as a precondition.

kodebach:
> * KeySet as generic data structure

That's what I'll add to the `core` folder. Beyond that, I currently have no plans to write more use cases.

> * config shared between applications (fallback, override, ...)
> * maybe something about modifying specifications (specload etc.)

I tried to focus on the things that either affect the current decisions, or the API change for my thesis. I think these two are interesting, but don't fall into those categories. So I probably won't write use cases for that soon. I can of course provide input, if somebody else (e.g. @tmakar) wants to write them.

doc/usecases/kdb/README.md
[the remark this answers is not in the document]
markus2330: There are different "levels" for use cases. Of course there are also high-level use cases (sometimes called cloud or kite) but use cases can also be quite low-level.

doc/usecases/kdb/UC_load_config.md
markus2330: I simply don't think it needs to be repeated in the Error scenario.

doc/usecases/kdb/README.md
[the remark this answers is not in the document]
kodebach: Based on the template and the other existing use cases, _our_ use cases are high-level. If we want something similar on a lower level we should use a different name (and probably a different format).

doc/usecases/kdb/UC_load_config.md
```diff
@@ -0,0 +1,31 @@
+# Use Case: Loading application configuration
+
+## Summary
+
+- **Title:** Loading application configuration
```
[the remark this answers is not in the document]
kodebach: I'm not describing use cases for the `kdb` tool here. I'm describing the Key Database (KDB) itself and that needs to accommodate for both long-lived applications that want up-to-date config (like a web-server) and short-running applications (like `kdb`).

Therefore, this describes the simplest use case of loading configuration once. The polling/notification stuff would be a separate use case as I said.

doc/usecases/kdb/UC_modify_config.md
kodebach: Still not sure what you mean. Remember this should describe the ideal case "how it should be" not "how it is". I know the current `spec` leaves some stuff around, but in principle `spec` should remove everything it generated.

In any case this is definitely out of scope here.

doc/usecases/core/UC_keyset_create.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:**
+  - Caller requests to create a new `KeySet`
+  - Core instantiates an emtpy `KeySet` and returns it
+- **Alternative scenario:**
+  - Caller requests to create a new `KeySet`, with known content
+  - Core instantiates an appropriately sized `KeySet`, fills it and returns it
```
kodebach, line 19: I was intentionally vague here. I did not want to require this a single `ksNew` function that takes everything as arguments. Instead I wanted to still allow using multiple function calls to achieve this, e.g.

```c
KeySet * ks = ksCreate (15);
ksInsert (ks, key1);
ksInsert (ks, key2);
// ...

// OR

KeySet * ks = ksInsert (ksInsert (ksCreate (2), key1), key2);

// OR

KeySet * ks = ksInsertAll (ksCreate (15), key1, key2, ...);
```

One thing I may still change, is that the `ksNew`-like function could exist outside `libelektra-core`.

doc/usecases/core/UC_keyset_lookup_cascading.md
```diff
@@ -0,0 +1,35 @@
...
+  - Core does not find specification
+  - Core tries equivalent name in all appropriate namespaces in correct order
+  - If a matching `Key` is found, Core returns the index
+  - Otherwise, Core returns a value indicating "Not found".
+    No valid index is returned, because `Key`s with cascading names cannot be inserted.
+- **Alternative scenario:**
+  - Caller asks Core to look up `Key` by cascading name in `KeySet`
+  - Core checks for specification matching desired `Key`
+  - Core finds specification
+  - Core tries links from specification and equivalent names in appropriate namespaces in correct order
```
kodebach, line 28: It's unclear whether handling the specification links should even be the core's responsibility. It's not ideal, e.g. adding new kinds of links requires modifying the core. IMO this is also more part of a separate group of use cases concerned with specifications in general.

If we remove this alternative scenario here, I would make this cascading lookup case simply an alternative scenario of the direct lookup.

doc/usecases/kdb/UC_modify_config.md
kodebach: Also applies above, but we also have to keep in mind that most use case (and these ones especially) should apply independent of language binding. Therefore, we should avoid referencing specific APIs.

kodebach:
@markus2330 I'd prefer if we could merge this quickly and not focus on to many details. I really don't want to waste time on this and I'd much rather work on actual decisions. The use cases (especially for `Key` and `KeySet`) are intentionally very abstract, because many things are what I'd consider "technical details". IMO a use case should be language and implementation agnostic.

doc/usecases/core/UC_key_value.md
```diff
@@ -0,0 +1,23 @@
...
+
+- **Precondition:**
+  - [`Key` has been created](UC_key_create.md)
+- **Main success scenario:**
+  - Caller requests current value of `Key`
+  - Core returns current value of `Key`
+- **Alternative scenario:**
+  - Caller requests to change value of `Key` to new value
+  - Core stores new value in `Key`
+- **Error scenario:** -
```
markus2330, line 21: If not string/binary as requested.

doc/usecases/core/UC_key_value.md
```diff
@@ -0,0 +1,23 @@
+# Use Case: `Key` Value
+
+## Summary
+
+- **Title:** Create `Key`
+- **Scope:** `libelektra-core`
+- **Level:** Developer Goal
+- **Actors:** Core, Caller
+- **Brief:** Access to and manipulation `Key`'s value
```
markus2330, line 9:
```suggestion
- **Brief:** Access to and manipulation `Key`'s value. Strings cannot have null values nor  the value can be null pointer.
```

doc/usecases/core/UC_keyname_hierarchy.md
```diff
@@ -0,0 +1,29 @@
+# Use Case: `Key` Hierarchy
+
+## Summary
+
+- **Title:** Create `Key`
+- **Scope:** `libelektra-core`
+- **Level:** Developer Goal
+- **Actors:** Core, Caller
+- **Brief:** Hierarchy comparison between `Key`'s
```
markus2330, line 9: Please give name of function, otherwise one needs to guess what this is about.

doc/usecases/core/UC_keyset_create.md
markus2330: Use cases are also decisions. We decide here if we want this feature in the core.

The reason for the size argument was that the signature `ksNew(...)` with only `Key`s is not possible. So if we want to get rid of the use case to construct a `ksNew` in one go in the core, we should also get rid of this size.

Imho it is okay to have a `ksNew()` (without arguments) in the core. `ksCreate()` with the vaargs can be in an extra lib, it is not really minimal.

doc/usecases/core/UC_keyset_delete.md
```diff
@@ -0,0 +1,28 @@
+# Use Case: Delete `Key` from `KeySet`
```
markus2330, line 1: Please take care to be consistent with our terminology doc/help/elektra-glossary.md

We currently call this operation `pop`. But I am open to change it (e.g. by using remove for both removing from the keyset and removing from KDB). Please also change doc/help/elektra-glossary.md

doc/usecases/core/UC_keyset_delete.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Summary
+
+- **Title:** Delete `Key` from `KeySet`
+- **Scope:** `libelektra-core`
+- **Level:** Developer Goal
+- **Actors:** Core, Caller
+- **Brief:** Caller removes `Key` from existing `KeySet`
```
markus2330, line 9: Also be consistent within the documents. If it is called remove, it must be called remove everywhere.

doc/usecases/core/UC_keyset_delete.md
```diff
@@ -0,0 +1,28 @@
...
+  - Caller requests to remove `Key` by index from `KeySet`
+  - Core removes given index from `KeySet`
+- **Alternative scenario:**
+  - Caller [looks up hierarchy in `KeySet`](UC_keyset_lookup_prefix.md) to get index range
+  - Caller requests to remove `Key`s by index range from `KeySet`
+  - Core removes given index range from `KeySet`
+- **Error scenario:** -
+- **Postcondition:**
+  - The deleted `Key`(s) MUST NOT be part of `KeySet` anymore.
+    A lookup for that/those `Key`s MUST NOT return any results.
```
markus2330, line 27: We should also be clear in the use cases that we really have an array-like access.

```suggestion
    A lookup for that/those `Key`s MUST NOT return any results.
    Indexes of the `Key`s after the removed key in the `KeySet` change by one.
```

doc/usecases/core/UC_keyset_insert.md
```diff
@@ -0,0 +1,29 @@
...
+- **Actors:** Core, Caller
+- **Brief:** Caller inserts new `Key` into existing `KeySet`
+
+## Scenarios
+
+- **Precondition:**
+  - [`KeySet` has been created](UC_keyset_create.md)
+- **Main success scenario:**
+  - Caller requests to insert a new `Key` with non-cascading name into `KeySet`
+  - If necessary Core resizes `KeySet`
```
markus2330, line 17: This says it would not be okay for an implementation to fail after inserting a fixed number of keys. Probably okay to require this, as nowadays also most embedded systems have proper memory allocation.

doc/usecases/core/UC_keyset_lookup_cascading.md
markus2330: I was thinking about lookup plugins a lot and currently don't think it is worth the effort. I think the current lookup can be used as fixed integrated part of the core.

doc/usecases/core/UC_keyset_lookup_direct.md
```diff
@@ -0,0 +1,24 @@
+# Use Case: Direct lookup in `KeySet`
```
markus2330, line 1: I don't think this needs an extra use case. IMHO it is simply a scenario of the lookup.

doc/usecases/core/UC_keyset_lookup_prefix.md
```diff
@@ -0,0 +1,25 @@
+# Use Case: Lookup hierarchy in `KeySet`
+
+## Summary
+
+- **Title:** Lookup hierarchy in `KeySet`
+- **Scope:** `libelektra-core`
+- **Level:** Developer Goal
+- **Actors:** Core, Caller
+- **Brief:** Caller looks up `Key` hierarchy in existing `KeySet`
```
markus2330, line 9: Not clear to me what this means. I am not sure if we need this usecase. Please also add an example.

doc/usecases/core/UC_keyset_retrieve.md
```diff
@@ -0,0 +1,27 @@
+# Use Case: Retrieve `Key` from `KeySet`
```
markus2330, line 1: Again, I think this is only a scenario of lookup.

doc/usecases/kdb/UC_modify_config.md
markus2330: I agree the use cases are also the basis for the bindings. But I don't think that the name we use in the C API will be disturbing. We recommend for APIs to use the same names, maybe except of prefix or similar, e.g., `Elektra::Kdb.set()` but it is still clear that it is elektraKdbSet.

doc/usecases/kdb/UC_store_config.md
markus2330: > You could also store without loading first. I know it's not supported by kdbSet, but that doesn't matter. We're not documenting the API here, we're listing use cases.

If we don't support set without prior get, it shouldn't be present in the use cases.

doc/usecases/kdb/UC_validate_config.md
markus2330: Then the name is misleading. So you mean something like:

```suggestion
# Use Case: Configuration with specification
```

markus2330 (changes requested):
Much better!

> I'd prefer if we could merge this quickly and not focus on to many details.

As these are the most important use cases of Elektra, we definitely need a second review here (by someone else).

> I really don't want to waste time on this and I'd much rather work on actual decisions.

Use cases are the most important decisions. They set the direction and scope.

> The use cases (especially for Key and KeySet) are intentionally very abstract, because many things are what I'd consider "technical details". IMO a use case should be language and implementation agnostic.

Exactly, it must consider reimplementations and bindings of Elektra. But it should be **as specific as possible** (e.g. we must be clear that our KeySet has access like to an array etc.) and describe in good detail what every implementation and binding must offer.

When we are too abstract here (like we have been for KeySet in the past) it is maybe good for alternative implementations but it terrible for users, as they cannot benefit from e.g. random access.

markus2330:
Maybe Eiskasten and/or 0x6178656c are good reviewers as they both write bindings (didn't mention them on purpose, as it is better if you do the suggested fixes first).

doc/usecases/core/UC_key_meta.md
```diff
@@ -0,0 +1,21 @@
+# Use Case: `Key` Metadata
+
+## Summary
+
+- **Title:** Create `Key`
```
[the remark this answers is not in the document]
kodebach: Totally expected that I had copy-paste errors. Should I remove it from the template as well?

doc/usecases/core/UC_key_value.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Scenarios
+
+- **Precondition:**
+  - [`Key` has been created](UC_key_create.md)
+- **Main success scenario:**
+  - Caller requests current value of `Key`
```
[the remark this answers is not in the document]
kodebach: No it should not IMO. There is a semi-finished decision that says the core treats all values as untyped byte sequences (I'm still working on updating the decision). However, more importantly it really doesn't contribute anything to the use case. Also many bindings provide getters for other types too, not just strings or binary data. So the safest thing is to just say: A Key has a value, in general that's a sequence of arbitrary bytes with a known length.

doc/usecases/core/UC_key_value.md
kodebach: See above. This is actually another reason why I avoided the distinction. It creates unnecessary error cases. The type error is on a higher level than this use case IMO. If you want I can write a separate use case for typed value access. Which types are supported would be a technical implementation detail.

doc/usecases/core/UC_keyname_hierarchy.md
kodebach: Once again, I intentionally avoid function names, because it is an implementation detail or if you want, it is on "decision level" not "use case level". The function name doesn't change anything about the use case. Different bindings will use different function names.

doc/usecases/core/UC_keyset_create.md
kodebach: IMHO this discussion is far too detailed already for this PR. I'll remove the alternative scenario and we'll discuss the details of creating keysets in #4244.

doc/usecases/core/UC_keyset_delete.md
kodebach: Pop is a completely stupid name for what it does, because pop has a pretty specific meaning when it comes to data structures. If had to guess, I'd say at some point the pop operation removed the key that was added last, or maybe the last key in keyset order.

In any case, I'm open to renaming this use case to "Remove". But I don't think we need consistent terminology here. In fact, I think for use cases it makes a lot more sense to use _common terminology_ instead of introducing Elektra specific terms. That's also why I used terms like "associative array" and "prefix lookup". IMO the Elektra specific terminology should introduced on a decision level. In fact, deciding the terminology we will use is inherently part of a decision.

doc/usecases/core/UC_keyset_delete.md
kodebach: Like I also said above, IMO a use case is a more informal, prose description. Unless explicitly stated or in a completely different context, you should be able to assume that normal English synonyms (like "remove" and "delete") have the same meaning.

In some sense, I see the use cases as a description of what Elektra can do. When I read them, I don't want to study a Glossary first to know what's going on.

However, as already stated above, in this case I'm open to renaming the use case to "Remove".

doc/usecases/core/UC_keyset_delete.md
kodebach: No, IMO that's not necessary. The lookup use case already states that indices are only valid until a key is added/removed. That's enough IMO. That there is actually just a single array behind and the index changes by one, is an implementation detail.

I've been very careful in how I describe the data structure of a `KeySet`. If we go into too much detail, the use case already precludes the decision for how a `KeySet` is implemented. For example, I'm sure there is a way to build `KeySet` as a Radix tree or a Judy array.

doc/usecases/core/UC_keyset_insert.md
kodebach: Not sure what you mean here, a functioning `malloc` is a requirement for all of Elektra.

doc/usecases/core/UC_keyset_lookup_cascading.md
kodebach: I wouldn't actually create separate lookup plugins. IMO the cascading lookup makes sense. But the `meta:/fallback` et al. links are a bit out of place in the core IMO. It is basically the only place where metadata has an actual effect on a core function (which kinda makes it no longer **meta**data).

What I was thinking about was more like having a function pointer somewhere that points to a function in the spec plugin (i.e. the currently active implementation of the spec hook) when the `KeySet` comes from `kdbGet`. In other `KeySet`s it doesn't point anywhere, but I don't think the link stuff has any use for other keysets.

Some options for where to put the function pointer (just OTTOMH, no thought put into options):
- A pointer in `KeySet`; called when a cascading lookup finds no result; arguments are the `KeySet *` and the key we're looking for
- In `meta:/callback/lookup` instead of the current links; called when the key is found in a cascading lookup; arguments are the `KeySet *` and the key we're looking for
- Add new namespace `KEY_NS_LOOKUP`; has highest priority in cascading lookup; always contains function pointer; called when the key is found in a cascading lookup; arguments are the `KeySet *` and the key we're looking for

Like I said, I just find it a bit weird that the core knows about this stuff, when it makes almost no sense, if you don't know it's used for configuration data.

doc/usecases/core/UC_keyset_lookup_direct.md
kodebach: Like I said, IMO it depends on whether the links handling is part of the core. If it is, I find it easier to separate them, because a direct lookup and a cascading lookup are a bit different in nature.

doc/usecases/core/UC_keyset_lookup_prefix.md
kodebach: > Not clear to me what this means. I am not sure if we need this usecase.

It's basically `ksFindHierarchy`, which is a slightly restricted kind of "prefix lookup", i.e., finding all items that start with a given prefix. The restriction is that the prefix must end with a zero-terminator that is significant for the lookup, i.e., we only find keys that are below. Although I did not and do not want to use that terminology, because that requires extra Elektra-specific explanations.

> Please also add an example.

Where, there is not really a place for this in the template. Maybe more explanation will help instead?

doc/usecases/core/UC_keyset_retrieve.md
kodebach: No it's not. If you look carefully, the lookups return an index and this retrieves the key given a known index. This use case is also what makes iteration possible. No other use case says you should be able to access `Key`s by index.

In the current API the lookup is `ksSearch` and this is `ksAtCursor`.

doc/usecases/kdb/UC_modify_config.md
kodebach: Hm okay, let me phrase it this way. The proper order (that you yourself suggested) would be: (1) use case, (2) decision, (3) implementation.

I don't think at the use case stage API names should be defined. That should be part of the decision, or maybe even an implementation detail (for internal APIs). So if we imagine this use case was written before the API existed, then it the name `kdbSet` wouldn't even exist yet.

While we could make an exception here, because the name does exist and we can use it. I think that would be a bad idea. These use cases are the core part that everybody contributing to Elektra needs to understand. So if they want to write new use cases, they will look at these (because they'll understand the contents already and can focus on the format). Using API names here, may suggest they should be used in other use cases too.

So far that's not a big problem. But IMO the issue is then twofold:
1. By using names, the use case stage will be unnecessarily prolonged, because there will be an inevitable discussion about API names. This discussion contributes nothing to the development of the use case and just steals peoples focus.
2. If for whatever reason, we decide to change the name later (during the decision or implementation stage), then we would have to update the use case. For no reason other than to fix the name. I don't think this would ever be worth it, unless there is a very niche use case that is impossible to understand without concrete API names.

doc/usecases/kdb/UC_store_config.md
kodebach: I disagree. The use case should define what we want. It is not the place to define limitations. The limitations of different implementations should be weighed in the decision stage. That is where we have a proper process to decide what limitations are acceptable. It's just that in this case, we are doing things backwards.

So in this specific case:
- Use case: We want to be able to store configuration data.
- Decision: Requiring a load before a store is acceptable. We can always load an empty config and then store the new config, if we don't have config yet. If we want to store a config regardless of whether one already exists, just load the existing one and discard it in memory. Then store the new config.
- Implementation: `kdbGet` & `kdbSet`

kodebach:
> Maybe Eiskasten and/or 0x6178656c are good reviewers as they both currently write bindings

Not sure, I had hoped, we can make quick progress here without drawn out discussions and reviews, since all of this is just documenting already decided and implemented stuff anyway.

> it is better if you do the suggested fixes first

Then it seems like this will be a long PR... It seems we have a different under standing of what a use case should look like...

I also have to ask the question: Do we really need to be so painstakingly meticulous here? There were some pretty pedantic comments in your review (e.g. "delete" vs. "review").

doc/usecases/core/UC_key_meta.md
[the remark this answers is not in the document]
markus2330: Yes, I think it is useless to have the title twice.

markus2330:
>  Not sure, I had hoped, we can make quick progress here without drawn out discussions and reviews, since all of this is just documenting already decided and implemented stuff anyway.

Didn't know that you only want to document status-quo. This sounds like a good idea: let us merge this with status-quo asap, and then create a new PR with the differences of your API restructure. E.g. hierarchy lookup etc. needs to be removed then. Iteration is missing.

> Do we really need to be so painstakingly meticulous here?

I don't have the impression I were.

> There were some pretty pedantic comments in your review (e.g. "delete" vs. "[remove]").

This is not pedantic. In German it might be okay to switch between similar words, in technical writing (English) this is an absolute NO-GO. This is also practice for the thesis, where this must be flawlessly correct (with basically any supervisor I know). remove/delete are usually very different things in technical documents.

kodebach:
> E.g. hierarchy lookup etc. needs to be removed then. Iteration is missing.

IMO iteration is covered by "Retrieve `Key` from `KeySet`". But I can add a separate use case too.

Hierarchy Lookup already exists (`ksFindHierarchy`, `ksBelow`, `ksCut`) so I won't remove it. However, what I will do in general, is rewrite the use cases from the index based forms to be more like what we have now (return keys directly).

> in technical writing (English) this is an absolute NO-GO

Like I said in another comment above, I consider the use cases more of an informal prose text than a rigorous specification. IMO they should be understandable _without_ the need for a glossary and exact definition of terms. Specific terms and the glossary should be decided at a later stage (also: remember we're doing things backwards here).

kodebach:
@markus2330 I have now incorporated most of your feedback. In particular, I have switched from "delete key" to "remove key" everywhere and I rewrote the "hierarchy lookup" to better fit the existing `ksCut` (technically `ksBelow` because removal is not mentioned, but I think we can gloss over that).

Here is also a list of which use case is implemented by which function:

- `UC_key_create.md`: implemented by `keyNew`
- `UC_key_meta.md`: implemented by `keyMeta`
- `UC_keyname_hierarchy.md`: implemented by `keyIsBelow` et al.
- `UC_key_name.md`: implemented by `keyName`, `keyUnescapedName`, `keySetName` et al.
- `UC_keyname_ordering.md`: implemented by `keyCmp`
- `UC_key_namespace.md`: implemented by `keyGetNamespace`
- `UC_keyset_create.md`: implemented by `ksNew`
- `UC_keyset_insert.md`: implemented by `ksAppendKey` & `ksAppend`
- `UC_keyset_lookup_cascading.md`: should be implemented by `ksSearch` (return index), but currently only implemented by `ksLookup` (returns `Key *`)
- `UC_keyset_lookup_direct.md`: implemented by `ksSearch`
- `UC_keyset_lookup_hierarchy.md`: implemented by `ksBelow` and with extra removal by `ksCut`
- `UC_keyset_remove.md`: implemented by `elektraKsPopAtCursor` (not public, only main scenario); combined lookup and removal implemented by `ksLookup` (only main scenario); alternative scenario implemented by `ksCut` in limited fashion
- `UC_keyset_retrieve.md`: implemented by `ksAtCursor`
- `UC_key_value.md`: implemented by `keyString`, `keyValue`, `keySetString` et al.


If we want this in the repo, I'd put it somewhere in `doc/contrib`.

markus2330:
> If we want this in the repo, I'd put it somewhere in doc/contrib.

Why didn't you put it to the use cases? Would be much easier to write&read?

kodebach:
> Why didn't you put it to the use cases?

Because (once again) we are doing things backwards. The use case should have been written before the API was designed. Therefore, the use case cannot refer to the API.

Yes, here we could refer to the API because we're doing things backwards. But these will likely be the examples that people look at first, so we shouldn't make exceptions here.


The order should be (and I think we agree on that):
1. Terms (if needed)
2. Use cases
3. Decisions (including API design)
4. Implementation
5. Additional documentation

doc/usecases/core/UC_keyset_lookup_hierarchy.md
```diff
@@ -0,0 +1,23 @@
+# Use Case: Lookup hierarchy in `KeySet`
```
markus2330, line 1: I see a few problems here not solvable within one review iteration (which we both want). I think we should remove the use case for now and add it later when we discuss about the API changes. (Simply remove the file, and we revert it later.)

doc/usecases/core/UC_keyset_remove.md
```diff
@@ -0,0 +1,27 @@
+# Use Case: Remove `Key` from `KeySet`
```
markus2330, line 1: I am fine with calling it remove, please update doc/help/elektra-glossary.md

doc/usecases/core/UC_keyset_retrieve.md
markus2330: Then please call it `at` instead of `retrieve`.

doc/usecases/kdb/UC_load_config.md
```diff
@@ -0,0 +1,30 @@
+# Use Case: Loading application configuration
```
markus2330, line 1: Please also rename file to `UC_get_configuration.md`

```suggestion
# Use Case: Get Configuration
```

doc/usecases/kdb/UC_store_config.md
```diff
@@ -0,0 +1,27 @@
+# Use Case: Storing application configuration
```
markus2330, line 1: Please also rename file to `UC_set_configuration.md`

```suggestion
# Use Case: Set configuration
```

doc/usecases/kdb/UC_modify_config.md
markus2330: It is not only about "defining names" (but for new APIs of course also that) but my main reason for the wish now is that it is simply practical. For the future: Ideal would be a link from the use case to the implementation (traceability links) and to the API documentation (for convenience). This obviously can only be done where the implementation already exist.

For now https://github.com/ElektraInitiative/libelektra/pull/4697#issuecomment-1332014001 was enough, thank you for writing it down.

> By using names, the use case stage will be unnecessarily prolonged, because there will be an inevitable discussion about API names. This discussion contributes nothing to the development of the use case and just steals peoples focus.

I strongly disagree. The name is so important and so strongly connected to its use case, that it is actually a good idea to develop it together.

> If for whatever reason, we decide to change the name later

This will probably never be done again after your refactoring. And your refactoring tool hopefully will be able to also change the names in the docu and use cases, won't it?

doc/usecases/core/UC_keyset_remove.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Scenarios
+
+- **Precondition:**
+  - [`KeySet` has been created](UC_keyset_create.md)
+  - [`Key` has been inserted into `KeySet`](UC_keyset_insert.md)
+- **Main success scenario:**
+  - Caller [looks up `Key` in `KeySet`](UC_keyset_lookup_basic.md) to get index
+  - Caller requests to remove `Key` by index from `KeySet`
+  - Core removes given index from `KeySet`
```
markus2330, line 18:
```suggestion
  - Core removes given `Key` at index from `KeySet`
```

doc/usecases/core/UC_keyset_remove.md
```diff
@@ -0,0 +1,27 @@
...
+- **Alternative scenario:**
+  - Caller [looks up hierarchy in `KeySet`](UC_keyset_lookup_prefix.md) to get index range
+  - Caller requests to remove `Key`s by index range from `KeySet`
+  - Core removes given index range from `KeySet`
```
markus2330, line 22: This is also a new feature which shouldn't be added here.

```suggestion
```

doc/usecases/kdb/UC_store_config.md
```diff
@@ -0,0 +1,27 @@
...
+- **Level:** Developer Goal
+- **Actors:** Application, Elektra
+- **Brief:** Application stores configuration into KDB
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:**
+  - Application asks to store new configuration into KDB
+  - Elektra stores in backends
```
markus2330, line 15: This is confusing, unclear what you want to say here.

```suggestion
  - Elektra passes `KeySet` to backends, responsible to set the configuration as requested
```

doc/usecases/kdb/UC_uptodate_config.md
```diff
@@ -0,0 +1,24 @@
...
+- **Actors:** Application, Elektra
+- **Brief:** Application wants up-to-date configuration data
+
+## Scenarios
+
+- **Precondition:** -
+- **Main success scenario:** Notifications
+  - Application requests asynchronous callbacks via the notification system
+  - [Application loads configuration](UC_load_config.md)
+  - Every time the configuration changes afterwards, Elektra sends a notification to Application
```
markus2330, line 16:
```suggestion
  - Every time the configuration changes, Elektra sends a notification to Application(s)
```

doc/usecases/kdb/UC_uptodate_config.md
```diff
@@ -0,0 +1,24 @@
...
+  - When receiving a notification, Application may need to [load the configuration](UC_load_config.md) again
+- **Alternative scenario:** Polling
+  - [Application loads configuration](UC_load_config.md) repeatedly in regular intervals
+- **Error scenario:** -
+- **Postcondition:** -
+- **Non-functional Constraints:**
+  - Notifications SHOULD use standard communication systems
+  - Notifications MUST be fast to send and receive, because there SHOULD NOT be a rate limit
```
markus2330, line 24:
```suggestion
  - there SHOULD NOT be a rate limit of notifications
```

doc/usecases/kdb/UC_validate_config.md
```diff
@@ -0,0 +1,36 @@
...
+
+- **Scope:** `libelektra-kdb`, `spec`
+- **Level:** Developer Goal, User Goal
+- **Actors:** Elektra, Application, User
+- **Brief:** After loading configuration, Application can assume the configuration is valid
+
+## Scenarios
+
+- **Precondition:**
+  - A specification defining what a valid configuration for Application looks like exists.
```
markus2330, line 13:
```suggestion
  - A specification, defining how a valid configuration for Application looks like, exists.
```

markus2330 (changes requested):
Great job, looks already very promising! This is hopefully the last review round. Comments are quite specific and should be easy to fix.

doc/usecases/core/UC_key_create.md
```diff
@@ -0,0 +1,25 @@
...
+## Summary
+
+- **Scope:** `libelektra-core`
+- **Level:** Developer Goal
+- **Actors:** Core, Caller
+- **Brief:** Caller creates new `Key` instance
+
+## Scenarios
+
+- **Precondition:** -
```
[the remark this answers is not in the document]
kodebach: I will describe the error scenario. Although IMO this is a limitation of the current API and not something imposed by the use case. That's why I didn't mention it.

If the API looked like this:

```c
Key * keyNew (elektraNamespace namespace, const char * unescapedName, size_t nameSize);
```

Then there cannot be an invalid name.

doc/usecases/core/UC_keyset_lookup_cascading.md
kodebach: > the current solution that the lookup key has an undocumented function pointer as value is questionable.

Where is that the case?

> Can you please open a decision about that.

Not right now no. There are enough open decisions and we're making zero progress. This is not high priority stuff IMO.

doc/usecases/core/UC_keyset_lookup_hierarchy.md
kodebach: What problems?

This describes `ksBelow` exactly. If you want, I'll add the bit about removing the keys, then it is `ksCut`. That is already public API, so there isn't really anything to discuss, if we want to document the use cases for the **current** API.

doc/usecases/kdb/UC_modify_config.md
kodebach: > For the future: Ideal would be a link from the use case to the implementation (traceability links) and to the API documentation (for convenience).

IMO it would be better to have a file in `doc/contrib` that describes how the various use cases are covered by the API. Use cases might not map 1:1 onto the API. One use case might be multiple functions or one function may cover multiple use cases. It's easier if there is a separate file that describes the mapping.

> The name is so important and so strongly connected to its use case, that it is actually a good idea to develop it together.

Yes, the name is important. That's why IMO the API (including the names) should be developed during the decision stage, where it can be discussed and reviewed. However, I don't think it is useful to discuss the names with the use cases.

At the use case stage, the name would just be a placeholder anyway. It shouldn't matter, whether the function will be called `foo` or `bar`. What matters for the use case is what the function does.

> > If for whatever reason, we decide to change the name later
> 
> This will probably never be done again after your refactoring.

Skilfully ignored the examples in the parentheses, I see...

I didn't mean change the name of a function that is already merged and in use. That hopefully won't happen and would be a big breaking change anyway. I meant changing the name after the use case is merged, but before the implementation is written.

> your refactoring tool hopefully will be able to also change the names in the docu and use cases, won't it?

Highly unlikely. Updating code snippets in markdown files should be possible. They should be tagged as <code>\`\`\`c</code>, so I could look for that and try to parse the syntax tree. But for function names mentioned randomly, maybe not even in backticks (<code>\`</code>), I couldn't do any better than a standard regex replace.

doc/usecases/kdb/UC_store_config.md
kodebach: "responsible to set the configuration as requested" is even more confusing IMO and provides basically zero information.

Actually there is just a word missing here:

```suggestion
  - Elektra stores configuration in backends
```

doc/usecases/kdb/UC_uptodate_config.md
kodebach:
```suggestion
  - Every time the configuration changes, Elektra sends a notification to Application
```

The "(s)" is wrong. I used "Application" as a proper noun here. It's the name of the actor, i.e. the placeholder for _the one application_ that's "talking" to Elektra in this use case. That's why it's capitalized and there is no "the".

doc/usecases/core/UC_key_create.md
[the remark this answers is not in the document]
kodebach: I ended up adding it as a precondition after all. I intentionally just said "must be valid" without saying what "valid" means. That way it still covers all cases, because it may just be that all names are "valid".

kodebach:
@markus2330 I have now rewritten all the use cases to match the **current public API** as close as reasonably possible. I also added files to `doc/contrib/usecases_api` that describe how the use cases are implemented.

This also switching lookups from index-based to the current `Key *`-based approach and rewriting the "remove" use case to match the only public API (`ksLookup` with `KDB_O_POP`). I didn't rename "retrieve" to "at", because frankly "At `Key` from `KeySet`" is a stupid title. "At" is not a verb so it doesn't work to describe an action. I used "index access" instead, which can at least be turned into "to access by index" if a verb is needed.

I **really** hope this PR is now in a state where you agree that it can be merged.

PR #4714, opened 2022-11-24
[decision]: Changetracking (round 2)

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/decisions/1_in_discussion/change_tracking.md
+++ b/doc/decisions/1_in_discussion/change_tracking.md
@@ -91,7 +91,7 @@ Have a global keyset with deep-duped keys that is purely used for changetracking
 
 ### Store changes as meta keys
 
-When something changes for the first time, store the original value as a metakey to every key. 
+When something changes for the first time, store the original value as a metakey to every key.
 Not yet clear how we handle changes to metadata then.
 Also not really possible for keysets.
 
--- a/doc/decisions/1_in_discussion/change_tracking.md
+++ b/doc/decisions/1_in_discussion/change_tracking.md
@@ -53,16 +53,16 @@ The library `libelektra-core` must be kept minimal.
 - It is possible to do change tracking with reasonable memory and computation overhead
 - It is possible to design a single change tracking API that is useful for all existing and future plugins
 - False positivies are okay
+
   - this may happend when some keys may have been changed by the user, but have subsequentially been "unchanged" by a transformation plugin
     Scenario: hypothetical plugin that converts all values to UPPERCASE.
-      - user gets key `system:/background` with value `RED`
-      - user changes it to `red`
-      - changetracking detects that value has been changed, because `RED` != `red`
-      - plugin changes value to UPPERCASE `RED`
-      - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
+    - user gets key `system:/background` with value `RED`
+    - user changes it to `red`
+    - changetracking detects that value has been changed, because `RED` != `red`
+    - plugin changes value to UPPERCASE `RED`
+    - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
 
 - False negatives are not okay
-
 
 ## Considered Alternatives
 
--- a/doc/decisions/1_in_discussion/change_tracking.md
+++ b/doc/decisions/1_in_discussion/change_tracking.md
@@ -63,7 +63,7 @@ The library `libelektra-core` must be kept minimal.
     - plugin changes value from `false` to `0`
     - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
 
-- False negatives are not okay
+- False negatives (missed changes) are not okay
 
 ## Considered Alternatives
 
--- a/doc/decisions/1_in_discussion/change_tracking.md
+++ b/doc/decisions/1_in_discussion/change_tracking.md
@@ -55,11 +55,12 @@ The library `libelektra-core` must be kept minimal.
 - False positivies are okay
 
   - this may happend when some keys may have been changed by the user, but have subsequentially been "unchanged" by a transformation plugin
-    Scenario: hypothetical plugin that converts all values to UPPERCASE.
-    - user gets key `system:/background` with value `RED`
-    - user changes it to `red`
-    - changetracking detects that value has been changed, because `RED` != `red`
-    - plugin changes value to UPPERCASE `RED`
+    Scenario: plugin that converts `false`<->`0` and `true`<->`1`
+    - `system:/background` is stored with value `false`
+    - user gets key `system:/background` with value `0` (after conversion by plugin)
+    - user changes it to `false`
+    - changetracking detects that value has been changed, because `false` != `0`
+    - plugin changes value from `false` to `0`
     - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
 
 - False negatives are not okay
--- a/doc/decisions/1_in_discussion/change_tracking.md
+++ b/doc/decisions/1_in_discussion/change_tracking.md
@@ -128,12 +128,11 @@ Implement changetracking as a hooks plugin that will be called within `kdbGet` a
 The following hooks will be needed:
 
 - `tracking/get`: will be called at the end of `kdbGet`, directly before the result is returned.
-- `tracking/set/preliminary`: will be called at the beginning of `kdbSet` and after every step/phase in `kdbSet`.
-  We need this to have change tracking information as soon as possible, so that plugins in any step of the process can use this information.  
-   Also, plugins in every step of the process might change data, so it is important to call it after every step.
-- `tracking/set/final`: will be called in the `post-commit` phase, after all changes have been written to disk, but before the `notification/send` hook.
-  Represents the final, real changes to the KDB.
+- `tracking/set`: will be called at the beginning of `kdbSet`.
 - `tracking/changeset`: compute the changeset for the requested parent key and return it.
+
+As every hook plugin can define its own contract, in theory all storage forms mentioned in the previous chapter should be possible to implement.
+We could just point the plugin to `backendsData->keys` or the internal cache if we go down that route.
 
 ## Solutions - Query
 
--- a/doc/decisions/1_in_discussion/change_tracking.md
+++ b/doc/decisions/1_in_discussion/change_tracking.md
@@ -139,13 +139,13 @@ Downsides of this approach:
 Use the `backendData->keys` for change tracking
 
 We already store which keys have been returned by `kdbGet` for each backend within KDB.
-Currently, however, this is not a deep copy, as we are returning the keys from `backendData->keys` directly.
+Currently, however, this is not a deep copy, as we are returning the internally stored `Key` instances directly.
 This means we can not detect changes to the values or metadata of keys right now.
 We can, however, rely on this for detecting removed and added keys in its current form.
 
 As we now implement copy-on-write for all keys and keysets within Elektra, we can deep copy what we return in `kdbGet`.
 
-Another problem with this approach is that `backendData->keys` get cleared every time `kdbGet` is called.
+Another problem with this approach is that the internally stored keys are recreated as new instances every time `kdbGet` is called.
 
 ## Decision
 
--- a/doc/decisions/1_in_discussion/change_tracking.md
+++ b/doc/decisions/1_in_discussion/change_tracking.md
@@ -43,28 +43,43 @@ Change tracking must:
 - not duplicate data for each plugin that wants change tracking
 - work with all allowed sequences of `kdbGet` and `kdbSet` as [per this decision](operation_sequences.md)
 
+We only want to track changes that are done by the user, not changes that are done by plugins.
+I.e. the scope of change tracking is on what happens _outside_ of `kdbGet` and `kdbSet`.
+
 The library `libelektra-core` must be kept minimal.
 
 ## Assumptions
 
 - It is possible to do change tracking with reasonable memory and computation overhead
-- It is possible to design a single change tracking API that is useful
-  for all existing and future plugins
+- It is possible to design a single change tracking API that is useful for all existing and future plugins
+- False positivies are okay
+  - this may happend when some keys may have been changed by the user, but have subsequentially been "unchanged" by a transformation plugin
+    Scenario: hypothetical plugin that converts all values to UPPERCASE.
+      - user gets key `system:/background` with value `RED`
+      - user changes it to `red`
+      - changetracking detects that value has been changed, because `RED` != `red`
+      - plugin changes value to UPPERCASE `RED`
+      - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
+
+- False negatives are not okay
+
 
 ## Considered Alternatives
 
-### Alternative 1 - Per-parent key tracking within `libelektra-kdb`
+### Alternative 1 - Tracking withn `libelektra-kdb`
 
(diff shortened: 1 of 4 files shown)


atmaxinger:
I've mainly updates some alternatives with regards to full copy-on-write within Elektra. 

One more thing that we need to discuss, is what we do with false positives - i.e. keys that are detected as changed even though practically they aren't. This can (and will) happen with plugins that do transformation of the key values. 

As of now, I've added to the assumptions that false positives may be okay. We might be able to provide some advanced API that plugins can even filter out false positives.

markus2330:
> One more thing that we need to discuss, is what we do with false positives - i.e. keys that are detected as changed even though practically they aren't. This can (and will) happen with plugins that do transformation of the key values.

Please elaborate (ideally within the decision, maybe in notes). If I change a value and then back it is obviously not the same reference (needs more memory) but a string-comparison still shows it is unchanged. Doesn't the comparison avoid any false positives?

atmaxinger:
@markus2330 It is elaborated in the decision, under "Assumptions". It has nothing to do on how we compare values, but that plugins may change the values before writing to disk.

markus2330:
Thx, I see! Couldn't this problem solved as well if we require users to call `kdbSet` after any changes in the KeySet and the transformation happens there? For this decision I would simply assume that transformations are done correctly and discuss how the transformations should work in the transformations decision. #4621

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -43,28 +43,43 @@ Change tracking must:
...
+    - changetracking detects that value has been changed, because `RED` != `red`
+    - plugin changes value to UPPERCASE `RED`
+    - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
+
+- False negatives are not okay
 
 ## Considered Alternatives
 
-### Alternative 1 - Per-parent key tracking withn `libelektra-kdb`
+### Alternative 1 - Tracking withn `libelektra-kdb`
```
markus2330, line 69:
```suggestion
### Alternative 1 - Tracking within `libelektra-kdb`
```

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -43,28 +43,43 @@ Change tracking must:
...
 
-Do the per-parent-key tracking within `libelektra-kdb`, within the `kdbGet` and `kdbSet` operations.
+Do the tracking within `libelektra-kdb`, within the `kdbGet` and `kdbSet` operations.
 
-Essentially, we'd have a 'giant' hashmap in the form of (parent key)->(KeySet) for all parent keys that were used for a `kdbGet` operation in the lifetime of the `kdb` instance.
+Essentially, we'd have a 'giant' keyset for all keys that were used returned by a `kdbGet` operation in the lifetime of the `kdb` instance.
```
markus2330, line 73:
```suggestion
Essentially, we'd have an internal cache for all keys that were used returned by a `kdbGet` operation in the lifetime of the `kdb` instance.
```

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -124,15 +139,13 @@ Downsides of this approach:
 Use the `backendData->keys` for change tracking
 
 We already store which keys have been returned by `kdbGet` for each backend within KDB.
-Currently, however, this is not a deep duplication, as we are returning the keys from `backendData->keys` directly.
+Currently, however, this is not a deep copy, as we are returning the keys from `backendData->keys` directly.
```
markus2330, line 142: `backendData->keys` is an implementation detail. Is it the internal cache or something else?

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -124,15 +139,13 @@ Downsides of this approach:
...
 This means we can not detect changes to the values or metadata of keys right now.
 We can, however, rely on this for detecting removed and added keys in its current form.
 
-If we don't want to deep-dup it, we'd need to do something different to detect which keys have been modified.
-One possibility would be to add metadata within the `key` functions (e.g. `meta:/elektra/original`), but that would violate the `libelektra-core` must be minimal constraint.
+As we now implement copy-on-write for all keys and keysets within Elektra, we can deep copy what we return in `kdbGet`.
 
-There is already [another decision](../1_in_discussion/copy_on_write.md) which discusses adding general copy-on-write semantics to Elektra.
-We could use this together with `backendData->keys` deep-duped to do memory efficient change tracking.
+Another problem with this approach is that `backendData->keys` get cleared every time `kdbGet` is called.
```
markus2330, line 148: Again, please avoid internals like `backendData->keys` (or introduce them, as not everyone knows the details of new-backend yet).

markus2330 (changes requested):
Good to see progress here!

doc/decisions/1_in_discussion/change_tracking.md
atmaxinger: If we leave out this detail, then yes it would probably be the same as Alternative 1. This distinction was made before copy-on-write, as `backendData->keys` already stores the keys returned, whereas for Alternative 1, we'd have to have another duplication.

With copy-on-write, this distinction gets much more fuzzy.

atmaxinger:
> Couldn't this problem solved as well if we require users to call kdbSet after any changes in the KeySet and the transformation happens there?

I don't see how this would solve anything. The big problem is that something _other than the user_ can change the values.

This means that plugins using the change tracking API after the transformation step (e.g. notifications plugins and in particular the record plugin) will potentially get other change tracking results than plugins before the step.

From my POV, the "truest" change tracking results could be determined by:

- Storing a copy of the keys in `kdbGet` directly after the keys have been read from disk. Before any value transformations have been applied but after keyname transformations have been done
- Comparing it with the keys in `kdbSet` after everything is written to disk, but we somehow still have access to the keyname before the transformation.

This way, we would only detect if something truly changed. But this also means change tracking would be useless for all plugins other than notification hooks and the record hook. Those are the only plugins that are executed AFTER everything is done.

kodebach:
> From my POV, the "truest" change tracking results could be determined by:

That goes in the direction of what I suggested in #3497 a long time ago. But doing such a split on the `libelektra-kdb` level would be another (smaller) new-backend change.

I actually think, the "true" diff would be calculated between what is returned by `kdbGet` and what is passed to `kdbSet`. That's the data that is visible to applications, that is "Elektra's view of the world". The whole idea is that Elektra gives you a uniform view of the config, no matter what is actually stored, so that's what we should use.

> This means that plugins using the change tracking API after the transformation step (e.g. notifications plugins and in particular the record plugin) will potentially get other change tracking results than plugins before the step.

I don't quite understand that statement. I'd assume the diff is calculated once in `kdbSet` and stored as long as `kdbSet` is running. If a plugin requests the diff via the API, it just gets this one diff, no matter when the plugin is executed.

We could of course also provide an API for calculating diffs (should be a separate library though). But that API should IMO take 2 `KeySet`s as input and produce a diff. So if plugins used that API for some reason, they'd have to provide both `KeySet`s. Then it is obvious why plugins get different results, if the have different `KeySet`s

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -43,28 +43,43 @@ Change tracking must:
...
+- It is possible to design a single change tracking API that is useful for all existing and future plugins
+- False positivies are okay
+
+  - this may happend when some keys may have been changed by the user, but have subsequentially been "unchanged" by a transformation plugin
+    Scenario: hypothetical plugin that converts all values to UPPERCASE.
+    - user gets key `system:/background` with value `RED`
+    - user changes it to `red`
+    - changetracking detects that value has been changed, because `RED` != `red`
+    - plugin changes value to UPPERCASE `RED`
+    - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
```
kodebach, line 63: Let's use something that really exists

```suggestion
    Scenario: `type` plugin configured for boolean conversion with `false`/`true`
    - `system:/background` is stored with value `false`
    - user gets key `system:/background` with value `0` (after conversion by `type`)
    - user changes it to `false`
    - changetracking detects that value has been changed, because `false` != `0`
    - `type` plugin changes value to `0`
    - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
```

doc/decisions/1_in_discussion/change_tracking.md
kodebach:
```suggestion
Currently, however, this is not a deep copy, as we are returning the internally stored `Key` instances directly.
```

We don't need to say where they are stored, just that the keys are returned and therefore can be changed externally.

doc/decisions/1_in_discussion/change_tracking.md
kodebach:
```suggestion
Another problem with this approach is that the internally stored keys are recreated as new instances every time `kdbGet` is called.
```

Not quite true I think, but good enough here.

kodebach (commented):
I do agree with markus2330 that we should assume transformations are done correctly (whatever that may mean) in this decision. At least for now. If we find that there is no correct way to do them, then we may need to revisit this.

However, I also think false positives **for values** are not an issue at all. False negatives would be a much bigger issue, because that means we miss changes. If there is a use case where false positives must be avoided, we can just do a kind of re-check.

The change tracking plugin should have access to the current data. It can just check if the diff matches the current data and if not update the diff.

That said, I also believe that in your example the user using `red` (or in my example `false`) can simply be called incorrect. Then the reason for the false positive is simply that the user made an incorrect change. I say "incorrect", because the user should use the format returned by `kdbGet`. So if `RED` is returned then they should use `RED` and not `red`. In fact, you could even argue that the plugin that does the uppercasing should raise an error during `kdbSet`, if it sees a non-uppercase value.

doc/decisions/1_in_discussion/change_tracking.md
atmaxinger: Funny, in #4621 it was suggested that we don't name a plugin specifically 😅

kodebach: Hm, I guess there are pros and cons. I just think giving it a name makes this example more readable. I wouldn't even have to be a specific existing plugin. Also I think the `false` -> `0` conversion would be much more common and therefore it would be easier to understand why this is useful/important.

So maybe we go with:

```suggestion
    Scenario: plugin that converts `false`<->`0` and `true`<->`1`
    - `system:/background` is stored with value `false`
    - user gets key `system:/background` with value `0` (after conversion by plugin)
    - user changes it to `false`
    - changetracking detects that value has been changed, because `false` != `0`
    - plugin changes value from `false` to `0`
    - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
```

Note also the additional line at the beginning explaining how the value is stored.

markus2330:
Please rerequest the review when I should have a look.

atmaxinger:
I took the liberty and restructured this decision a bit. I have found that there are 3 largely independent things that we need to decide on here:

- Where to store changetracking data
- Where to implement changetracking
- How plugins and applications can interact with the changetracking mechanism.

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -41,68 +41,62 @@ Change tracking must:
...
 ## Assumptions
 
 - It is possible to do change tracking with reasonable memory and computation overhead
-- It is possible to design a single change tracking API that is useful
-  for all existing and future plugins
+- It is possible to design a single change tracking API that is useful for all existing and future plugins
+- False positivies are okay
 
-## Considered Alternatives
+  - this may happend when some keys may have been changed by the user, but have subsequentially been "unchanged" by a transformation plugin
```
kodebach, line 57:
```suggestion
  - this may happend when some keys have been changed by the user, but have subsequentially been "unchanged" by a transformation plugin
```

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -41,68 +41,62 @@ Change tracking must:
...
+    Scenario: plugin that converts `false`<->`0` and `true`<->`1`
+    - `system:/background` is stored with value `false`
+    - user gets key `system:/background` with value `0` (after conversion by plugin)
+    - user changes it to `false`
+    - changetracking detects that value has been changed, because `false` != `0`
+    - plugin changes value from `false` to `0`
+    - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
```
kodebach, line 64:
```suggestion
    - consumers of the changetracking API will now get a false positive if they query whether `system:/background` has been changed
    - We assume consumers of the changetracking API have access to both the previous and the new value of a changed key.
      Therefore, they could detect these false positive cases, if the need to.
```

kodebach (commented):
LGTM, but if you think the three parts are independent, it might make sense to split them into 3 decision files as well.

I'm not 100% sure it makes sense, since they are still quite related and share constraints, assumptions and most of the problem. But it may be easier to move forward, if we can decide the parts independently.

atmaxinger:
Well, I don't think they are 100% independent. What I mean is that those are 3 areas we need to decide on, but it's still a single decision. Otherwise, I would have to repeat some stuff in each alternative (e.g., if the implementation is in a plugin, then it could still be possible to use `backends->keys` but also plugin-internal storage)

I just want to present some alternatives for each area, and then we pick the most suitable ones from each.

atmaxinger:

atmaxinger:
@markus2330 could you please look over the decision?

doc/decisions/1_in_discussion/change_tracking.md
```diff
@@ -0,0 +1,181 @@
...
+
+bool elektraChangeTrackingValueChanged (ChangeTrackingContext * context, Key * key);
+bool elektraChangeTrackingMetaChanged (ChangeTrackingContext * context, Key * key);
+
+KeySet * elektraChangeTrackingGetAddedMetaKeys (ChangeTrackingContext * context, Key * key);
+KeySet * elektraChangeTrackingGetRemovedMetaKeys (ChangeTrackingContext * context, Key * key);
+KeySet * elektraChangeTrackingGetModifiedMetaKeys (ChangeTrackingContext * context, Key * key);
+
+Key * elektraChangeTrackingGetOriginalKey (ChangeTrackingContext * context, Key * key);
+const Key * elektraChangeTrackingGetOriginalMetaKey (ChangeTrackingContext * context, Key * key, const char * metaName);
```
markus2330, line 160: Please add some context why this would be needed (connect to use case etc.). Isn't three KeySets with all added/removed/modified (meta)keys enough?

markus2330 (approved):
Definitely looks fine for in_discussion. Before it can be in solutions_clear please add how a plugin would provide "Solutions - Query".

doc/decisions/1_in_discussion/change_tracking.md
atmaxinger: Good catch! Yes, if `elektraChangeTrackingGetModifiedKeys` / `elektraChangeTrackingGetModifiedMetaKeys` return the *old* keys, then these functions are not required.

atmaxinger:
@markus2330 I have added an entry as to how the query mechanism could be implemented in a plugin. Do note, however, that I actually intended to only have the API within a library (i.e. `libelektra-kdb`). This API wouldn't care whether changetracking is implemented as a plugin or not.

markus2330:
I think we can merge this but there are build errors.

atmaxinger:
The build error is a timeout in `npm install`. I triggered a rerun of the failed job.

atmaxinger:

atmaxinger:
@markus2330 @kodebach CI is green, we can merge now.

Issue #4686, opened 2022-11-17
Write use cases for Elektra's API


markus2330:
Before finalizing decisions #4645, we should write use cases for Elektra's API to get a common understanding what Elektra's API should offer.

> Something similar to doc/usecases/record_elektra but for applications and plugins using Elektra's API. More or less a summary of requirements/usecases that LCDproc or some plugins (you wrote) had.

> I hope this gets wider contributions, as this has intersections to basically anyone working with Elektra. In particular I hope that @lawli3t will also contribute. In retrospection this should have actually have been done before his RQ2.

_Originally posted by @markus2330 in https://github.com/ElektraInitiative/libelektra/issues/4645#issuecomment-1317407631_

- [ ] LCDproc @kodebach 
- [ ] KDE/GNOME @mpranj 
- [ ] kdb tools @hannes99 
- [ ] bindings @Eiskasten 
- [ ] mounting @flo91 
- [ ] escaping of key names (for users/admins to refer to keys, for storage plugins, for bindings like KDE/GNOME, ...)

PR #4715, opened 2022-11-26
[decisions] keyname

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


kodebach:
@markus2330 I put the decision back into draft status, but I hope we can make quick work of this and decide it ASAP. That's also why the decision is fully filled out and there is already a suggestion in the "Decision" section. I don't think there is any doubt left that only the unescaped name should be stored and I also think it is obvious that for some APIs taking the namespace and rest of the name as separate arguments makes sense.

It is not explained in the decision, because I can't link to the decision in #4683, but I think a separate namespace field works very well with the buffer struct API.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+The question now is, which representations should be used by `libelektra-core` and how.
+
+## Constraints
+
+- Because `KeySet` is ordered by name and stores `Key`, the order comparison between the name of two `Key`s must be fast.
+
+## Assumptions
+
+- In most cases the escaped name is used for developer convenience not, because of actual requirements.
+- The most common requirement for using the escaped name is UI: reading names from or displaying them in a user interface (e.g., `kdb` CLI)
```
atmaxinger, line 26: Also, debugging. Deciphering the unescaped name in debuggers is horrible. But this could be solved by providing [custom pretty-printers](https://undo.io/resources/gdb-watchpoint/here-quick-way-pretty-print-structures-gdb/) for gdb and other debuggers.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+However, in the API the name could always be a single `char *`, making for a very easy to use API.
+
+The biggest problem with this approach is that comparing two escaped names is not trivial.
+The comparison needs to account for namespaces, parts and escaping.
+
+Similarly, iterating over the individual parts of a name (and/or manipulating them) is non-trivial, because it requires logic to handle escape sequences.
+
+### Only unescaped name
+
+The unescaped name contains zero bytes.
```
atmaxinger, line 47:
```suggestion
The unescaped name contains zero-bytes (`\0`).
```

Took me an embarrassingly long time to make sense of this sentence 😅

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+Internally, this could be combined into one buffer, stored as a single pointer and size, again.
+However, depending on the API there can also be benefits to keeping the namespace as a separate field.
+
+Even with a separate namespace field, the benefits of "Only unescaped name" are retained.
+The memory consumption is near minimal (alignment padding can cause a difference).
+Comparisons are exactly the same, just with an additional namespace byte comparison beforehand.
+
+### Both escaped and unescaped name
+
+To combine the advantages of escaped and unescaped name, both could be used.
```
atmaxinger, line 75: This is the current approach, right?

atmaxinger (approved):
LGTM, make sense from a memory standpoint 🤗 👍

doc/decisions/0_drafts/keyname.md
kodebach: Ah yes, that was badly phrased :sweat_smile:

doc/decisions/0_drafts/keyname.md
kodebach: Yes

doc/decisions/0_drafts/keyname.md
kodebach: In GDB (and probably others) you can do:

```
p *key->ukey@key->keyUSize
```

This prints `key->ukey` as a fixed-length string of length `key->keyUSize`, e.g.:

```
$1 = "\006\000abc"
```

Had to google that syntax myself, but I was 99.9% sure it existed.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+The unescaped name contains zero bytes.
+It therefore must be represented as a pointer and a size.
+
+This can make for less convenient API, but there are mitigation strategies using additional types.
+Using unescaped names in code can be inconvenient, especially regarding the namespace.
+Without a namespace a name could be written as e.g., `"foo\0bar"`.
+But with a namespace it would be something like `"\1\0\foo\0bar"` and developers would need to remember what namespace `\1` is.
+Using the `KEY_NS_*` constants like this is not easily possible.
+
+Both order and hierarchy comparisons are very simple in this case and can be implemented with little more logic than a few `memcmp`.
```
markus2330, line 56: As this is a very common operation, performance is very important here. Not using `memcmp` definitely is much too slow (already tried it an earlier version). If "little more logic than a few `memcmp`" is too slow I cannot say, currently it is a single `memcmp`.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+Iterating over the individual parts is also trivial, since all parts are separated by zero bytes.
+
+### Only unescaped name, with separate namespace
+
+To solve the namespace issue of the above solution, the namespace could be separated.
+
+In the API it is an obvious solution.
+Instead of taking a single pointer and size, which receive values like `"\1\0foo\0bar"` and `10`, the API would take a namespace, a pointer, and a size, with values like `KEY_NS_CASCADING`, `"foo\0bar"` and `8`.
+
+Internally, this could be combined into one buffer, stored as a single pointer and size, again.
```
markus2330, line 66: Please more details and be more clear how

- the public API would look like and
- how the internal storage would look like.

Also be more clear about the advantages of the separation. Doesn't keyGetNamespace solve the problem?

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+If the API uses escaped names a lot (because it is more convenient), then this essentially degrades into the "Both escaped and unescaped name" solution.
+Even if APIs exist for both escaped and unescaped names, the convenience benefit, will lead to more use of escaped names.
+This means the escaped name will be stored for many keys and therefore the benefit of the lazy allocation is negated.
+
+Without the lazy allocation benefit, the only difference to "Both escaped and unescaped name" is that we have fewer pointers and sizes in structs.
+This saves some amount of memory and allocations, but makes internal code more difficult to write and understand.
+
+## Decision
+
+**Suggestion:**
```
markus2330, line 115: If there is a suggestion, it shouldn't be draft. Imho it could target "decided" or at least "in_progress" as it was before. Other than the separation of the namespace, everything is clear to me.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Go with "Only unescaped name, with separate namespace" from above
+- Leave whether namespace is stored separately TBD, until API requirements are worked out
+
+## Rationale
+
+- Largest memory savings
```
markus2330, line 122: It is not the largest, there are possibilities to split and reuse parts of key names. I would say it is well-balanced.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+- Still useful API
```
markus2330, line 123: Proposed API not clear to me.

markus2330 (changes requested):
Great work, looks very good. Only separation of namespace not clear.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+A `Key` in Elektra is identified by its name, which consists of a namespace and a number of name parts.
+There are two common representations for the name: escaped and unescaped.
+
+The unescaped form is essentially a single namespace byte plus and arbitrary-length sequence of arbitrary bytes, in which a zero byte separates parts.
+The escaped name is a more complicated representation as a zero-terminated string that maps 1:1 onto unescaped names.
+More details can be found in [the relevant docs](/doc/KEYNAMES.md).
+
+There is a conflict between these two forms in terms of API convenience and efficient execution.
+Generally, the escaped form is more convenient to use, since it is entirely human-readable and just a "normal" string.
+When it comes to execution, however, the unescaped name often makes things more efficient.
```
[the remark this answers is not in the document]
kodebach: Uh no. The unescaped name is actually harder to use in many cases. There are even some examples in this file. The comment really was just about the fact that the unescaped name makes things like sorting keys more efficient. I guess you mean that the internal implementation of things like `keyCmp` is easier with the unescaped name. But then we need a different wording. You can't say the unescaped name is easier to use in general.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+For example, order comparisons between unescaped names are much simpler.
```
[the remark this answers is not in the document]
kodebach: Haha... Now that's the opposite to above. Yes, it's more efficient too. But a single `memcmp` is clearly much simpler than whatever `strchr`/`strncmp` mess you'd have to do with just the escaped name.

doc/decisions/0_drafts/keyname.md
kodebach: How is that "rationale", it hard explains the decision and "easier debugging" really shouldn't affect an API related decision. I'll add it to the notes though.

doc/decisions/0_drafts/keyname.md
kodebach: I can add it, but I don't know, will anybody really care about that in a few months/years? When all this is done and implemented, nobody will care what the previous implementation was IMO.

doc/decisions/0_drafts/keyname.md
```diff
@@ -0,0 +1,136 @@
...
+The memory consumption is near minimal (alignment padding can cause a difference).
+Comparisons are exactly the same, just with an additional namespace byte comparison beforehand.
+
+### Both escaped and unescaped name
+
+To combine the advantages of escaped and unescaped name, both could be used.
+
+The API could largely rely on the escaped name, while e.g., comparisons can use the unescaped name.
+
+The issue with this approach is the insane memory consumption.
```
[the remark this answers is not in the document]
kodebach: > I would have written this already in the problem but it is also okay like it is.

Hm, not exactly like this, but I'll add a note about memory consumption to the problem.

> Another common problem worth mentioning is that people try to iterate the escaped name

Good point. I'll add more about the difficulties of using the escaped name correctly too.

doc/decisions/0_drafts/keyname.md
kodebach: `keyCmp` for order comparision is a single `memcmp`. The hierarchy comparison `keyIsBelow` is a tiny bit more. And if you want to be really technically, both at least to `NULL` check which are also a tiny bit of logic.

doc/decisions/0_drafts/keyname.md
kodebach: Like I said in https://github.com/ElektraInitiative/libelektra/pull/4715#issuecomment-1328036714 moving to draft was a formality, in case we got stuck clarifying the problem definition. Let's finish the reviews and then I'll move it to the right place before we merge.

The suggestion was mostly for the PR review. If it had looked like need a merge as draft or discussion, I'd have removed it.

doc/decisions/0_drafts/keyname.md
kodebach:
```suggestion
- Largest memory savings among proposed options.
```

I actually forgot to add the assumption about having a single pointer to the name. Then it should make sense again.

doc/decisions/0_drafts/keyname.md
kodebach: I'll explain it some more soon. The "Decision" and "Rationale" are not very detailed, because I wasn't sure how quick this PR would progress...

doc/decisions/0_drafts/keyname.md
kodebach: I didn't want to propose specific API here, because that's a lot to think about an discuss. Either I'll create a separate decision for the API, or it'll fall out of another decision (e.g. #4244) as a kind of corollary. The same goes for internal storage, where #4683 and #4676 are important decisions.

What this decision should be about IMO is the question posed in the "Problem" section: "Which representations should be used by `libelektra-core` and how?"

Maybe I could make this more clear, but the difference between the two options in question is:
- "Only unescaped name": Unescaped name is stored internally, and only the unescaped name (as a whole) is used in the API. So `keyGetNamespace (key)` wouldn't be allowed here. It should be replaced with `keyName (key)[0]`
- "Only unescaped name, with separate namespace": Similar, but namespace is separated. It is intentionally left open, whether the unescaped name is stored as one field or two fields (namespace and rest) internally. The important bit is that using the namespace separately in the API is not just allowed, but actually encouraged where it makes sense. A good example is: Something like `keySetName` would probably take the namespace as a separate argument, so the `KEY_NS_*` constants can be used easily.

doc/decisions/0_drafts/keyname.md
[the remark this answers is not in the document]
atmaxinger: Huh, can't you `memcmp` a string too?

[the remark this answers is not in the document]
kodebach: Sure you can, but we want this order:

```c
user:/
user:/a
user:/a/b
user:/a+
```

Because `user:/a < user:/a+` and everything below `user:/a` should be grouped.
But a simple lexicographical sort (i.e. sort by `strncmp`/`memcmp`) gives this order:

```c
user:/
user:/a
user:/a+
user:/a/b
```

`user:/a+ < user:/a/b` because at index 3 we have `+ < /`. With the unescaped name this is not an issue, because the names would be:

```c
\x06\x00
\x06\x00a
\x06\x00a\x00b
\x06\x00a+
```

and `\x00` is the smallest value, it always comes first. Therefore, `\x06\x00a\x00b < \x06\x00a+` since at index 3 we have `\x00 < /`.

> **Note** Yes, `strcmp` on escaped names works for alphanumeric names, and _some_ other symbols, but not all. That's why I used `a+` and not e.g. `ab`.

kodebach:
@markus2330 I hope this is now more clear. I also updated the "Decision" and "Rationale" sections. So, if you agree, I would move the file to `3_decided` (together with the required rebase).

markus2330 (changes requested):
Yes, this can directly target decided as the first decision round is already done.

One clarification is needed, though: that escaped keys API is only removed from the core, and not totally from Elektra. The details of this can be decided later.

kodebach:
Have you reviewed all the changes? I'd like to know you agree with everything, before I do the rebase and move the file, just so we don't loose any open threads...

> One clarification is needed, though: that escaped keys API is only removed from the core, and not totally from Elektra. The details of this can be decided later.

It's mentioned in both the "Decision"

https://github.com/ElektraInitiative/libelektra/blob/bead31f20a0c22dc4e7ab02d4bf1656de79b66bb/doc/decisions/0_drafts/keyname.md?plain=1#L132

and "Rationale"

https://github.com/ElektraInitiative/libelektra/blob/bead31f20a0c22dc4e7ab02d4bf1656de79b66bb/doc/decisions/0_drafts/keyname.md?plain=1#L141

markus2330:
As requested I looked at "Decision" and "Rationale". I don't think a full review is needed again. The basics are clear, the details are not yet decided anyway.

Feedback from @mpranj could be valuable.

kodebach:
@markus2330 I think this is done now. CI fails because Jenkins ran out of space again and there is a broken link in the CLion docs.

mpranj:
I'm sorry, but seems like I am missing some crucial information. (Thus I did not review/comment earlier)

Are we positive we actually **need** to have escaping at all?

Let's say we try to simplify and limit key names such that `/` can not be a key name part, but only the true separator character.
What functionality do we lose with this limitation, except that keynames can not contain this one character?

If this is something major, can you please add this to the decision?

kodebach:
> Are we positive we actually **need** to have escaping at all?

Assuming, key names are not limited to printable strings, we need a printable representation. `libelektra-core` doesn't have to know about that representation, but that's part of the point here. 

> Let's say we try to simplify and limit key names such that `/` can not be a key name part, but only the true separator character. What functionality do we lose with this limitation, except that keynames can not contain this one character?

Oh god, please let's not open that can of worms again....

Last time we discussed it @markus2330 was very insistent that we need to support arbitrary arguments for `keySetBaseName` and through some logic this lead to the conclusion we need to support arbitrary key name parts.
That's a bit philosophical and I still think, we could just make the caller do the escaping if they have arbitrary data as key name parts. After all POSIX paths have the same limitation and while people do complain about them, more often than not people think POSIX paths are not restricted enough.

Anyway, disallowing arbitrary parts would cause a few issues immediately:
1. Comparing names would be more complicated, a single `memcmp` wouldn't work. See https://github.com/ElektraInitiative/libelektra/pull/4715#discussion_r1032840316. I'll add that to the file as well.
2. Since `new-backend` we use the escaped name of the parent key as one of the parts in a mountpoint configuration in, i.e. `system:/elektra/mountpoints/user:\/foo/...`.

markus2330:
Escaping in key names is a use case we did not write yet, I added it to #4686 . The use case should be required as constraint, then it is clear what the decision discusses (and what not).

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,166 @@
...
+## Constraints
+
+- Because `KeySet` is ordered by name and stores `Key`, the order comparison between the name of two `Key`s must be fast.
+- We need a single pointer to a single buffer that contains the entire name of a `Key`.
+  While there are other options, some of which could even save memory (e.g., split into parts and deduplicate), much of the `KeySet` internals rely on the fact that the name is a single buffer.
+  Changing this would require major redesigns.
+
+## Assumptions
+
+- In most cases the escaped name is used for developer convenience and not because of actual requirements.
```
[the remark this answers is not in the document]
kodebach: Hm, yeah I'll rephrase it slightly. It certainly is for convenience, but probably not for developers.

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,166 @@
...
+In the escaped name, this task requires correct handling of escape sequences, whereas in the unescaped name parts are always delimited by a `\0` byte.
+
+Of course a naive way to solve this dilemma is to say "why not both" and store both versions.
+However, this can result in a lot of memory use, so it is not immediately the best option.
+
+The question now is, which representations should be used by `libelektra-core` and how.
+
+## Constraints
+
+- Because `KeySet` is ordered by name and stores `Key`, the order comparison between the name of two `Key`s must be fast.
```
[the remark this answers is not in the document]
kodebach: Do we have any data, that shows anything slower than a `memcmp` is too slow?

If not, I'd rather leave the constraint as "must be fast", but maybe we add an assumption that it must be comparable to a single `memcmp`.

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,166 @@
...
+Because the escaped name is a simple `\0`-terminated string, it can be represented as a single `char *`.
+
+Storing the name as a single `char *` would be the most space efficient.
+But resizing would require counting the length every time.
+Therefore, for storage the better solution may be a `char *` and a `size_t`.
+
+However, in the API the name could always be a single `char *`, making for a very easy to use API.
+
+The biggest problem with this approach is that comparing two escaped names is not trivial.
+The comparison needs to account for namespaces, parts and escaping.
```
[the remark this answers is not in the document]
kodebach:
```suggestion
The biggest problem with this approach is that comparing two escaped names is not trivial.
The comparison needs to account for namespaces, parts and escaping.
Previous benchmarks showed that it is very hard or even impossible to make the comparison of escaped names fast enough for our use cases.
```

IMO "not trivial" is the root problem. The previous benchmarks also can only show that the one implementation we had was too slow. Benchmarks cannot show that it is impossible to find a different faster algorithm.

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,166 @@
...
+- Convenience functions using escaped names, will be provided via other libraries
+- Where appropriate the API will take the namespace as a separate argument to allow using `KEY_NS_*` constants.
+- Whether namespace is stored separately in `struct _Key` will be decided at a later point, when the scope of all API changes and changes to `struct _Key` is clear.
+
+## Rationale
+
+- Largest memory savings among the proposed options
+- Option to use separate namespace argument leads to more convenient API (`KEY_NS_*` constants).
+- Simple internal code
+- Escaped name requirements can easily be solved by additional library (e.g., `libelektra-ease`), because not every caller will need those.
```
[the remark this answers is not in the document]
kodebach: > Probably better operations or even a dedicated library for escaping (if it is big enough).

IMO `libelektra-ease` makes more sense than `libelektra-extra`. Since escaped names should mostly be used for UI things, I think this would fit the descriptions in the `library_split.md` decision. I see the escaped names closer to the SHA256 Hash of a `KeySet` than something like a `ksCut` operation.

A completely separate library could make sense too. But it would be a pretty small library. The API would basically be just 4 functions:

```
isValidEscapedName()
escapeName()
cannonicalizeEscapedName()
unescapeName()
```

All together the code for those function would be only about 500-600 lines, based on the current code in `keyname.c`. So I certainly wouldn't say it is "big enough", but you could argue it is an important enough standalone use case to have it's own library. If we expect many people to need escaped names, but otherwise can just use `libelektra-core` then a standalone library certainly makes sense.

> Bindings should be required to also implement functions of this library (similar to `kdb`).

That would also be a reason to make it a standalone library. Otherwise it might be awkward to require that bindings implement just a part of a libraries API, but the rest is optional.

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,166 @@
...
+Both order and hierarchy comparisons are very simple in this case and can be implemented with a single `memcmp` and a tiny amount of extra logic (e.g., to handle cascading names).
+Iterating over the individual parts is also trivial, since all parts are separated by `\0` bytes.
+
+### Only unescaped name, with separate namespace
+
+In the above solution, the entire unescaped name (including the namespace) would always be considered one unit.
+As such, there would only be a single pointer and a size in an API that needs a name.
+This can be inconvenient, because it makes using the `KEY_NS_*` constants more difficult.
+
+This solution enhances the above, by considering the namespace a separate thing.
```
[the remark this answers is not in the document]
kodebach: I mean that's what the following sentences are meant to explain....

Not necessarily all the way separate as in #4732, but separate enough that the passing namespace and rest of name as separate arguments is allowed.

I'll try to explain it a bit better, and also incorporate some of #4732.

kodebach:
@markus2330 I incorporated your latest review, I hope everything is now covered and this can be merged. For the exact API I will open a new decision.

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,180 @@
+# Namespace and Name of Keys
+
+## Problem
+
+A `Key` in Elektra is identified by its name, which consists of a namespace and a number of name parts.
+There are two common representations for the name: escaped and unescaped.
+
+The unescaped form is essentially a single namespace byte plus and arbitrary-length sequence of arbitrary bytes, in which a `\0` byte separates parts.
+The escaped name is a more complicated representation as a `\0`-terminated string that maps 1:1 onto unescaped names.
```
markus2330, line 9: Sentence hard to read:

```suggestion
The escaped name is a `\0`-terminated string that maps 1:1 onto unescaped names.
```

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,180 @@
...
+There is a conflict between these two forms in terms of API convenience and efficient execution.
+Generally, the escaped form is more convenient to use, since it is entirely human-readable and just a "normal" string.
+Implementing many tasks (like order comparisons) is, however, much simpler when using the unescaped name.
+Additionally, using the unescaped name often results in more performant code as well.
+
+A particularly common example that highlights the difficulties in handling escaped names is splitting the name into parts.
+In the escaped name, this task requires correct handling of escape sequences, whereas in the unescaped name parts are always delimited by a `\0` byte.
+
+Of course a naive way to solve this dilemma is to say "why not both" and store both versions.
+However, this can result in a lot of memory use, so it is not immediately the best option.
```
markus2330, line 21:
```suggestion
However, this resulted in too much memory use, so we need to find another solution.
```

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,180 @@
...
+
+There is a conflict between these two forms in terms of API convenience and efficient execution.
+Generally, the escaped form is more convenient to use, since it is entirely human-readable and just a "normal" string.
+Implementing many tasks (like order comparisons) is, however, much simpler when using the unescaped name.
+Additionally, using the unescaped name often results in more performant code as well.
+
+A particularly common example that highlights the difficulties in handling escaped names is splitting the name into parts.
+In the escaped name, this task requires correct handling of escape sequences, whereas in the unescaped name parts are always delimited by a `\0` byte.
+
+Of course a naive way to solve this dilemma is to say "why not both" and store both versions.
```
markus2330, line 20:
```suggestion
Before this decision, we stored both versions in every `Key`.
```

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,180 @@
...
+- We need a single pointer to a single buffer that contains the entire name of a `Key`.
+  While there are other options, some of which could even save memory (e.g., split into parts and deduplicate), much of the `KeySet` internals rely on the fact that the name is a single buffer.
+  Changing this would require major redesigns.
+
+## Assumptions
+
+- In most cases the escaped name is used for convenience and not because of actual requirements.
+- The most common case for using the escaped name is UI: reading names from or displaying them in a user interface (e.g., `kdb` CLI)
+- In the constraint about order comparisons above, we assume that "fast enough" means "comparable to a single `memcmp`".
+  This is an assumption, because we do not have data showing this is a necessary requirement.
```
markus2330, line 38:
```suggestion
  In profiling non-`memcmp`-based implementations, up to now, it was obvious that most of the CPU time is spent on the comparisons in several benchmarks (like adding keys to a keyset).
```

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,180 @@
...
+  It may be possible to find a slower solution that is still "fast enough".
```
markus2330, line 39:
```suggestion
  It may be nevertheless possible to find a slower solution that is still "fast enough".
```

doc/decisions/3_decided/keyname.md
```diff
@@ -0,0 +1,180 @@
...
+This means the escaped name will be stored for many keys and therefore the benefit of the lazy allocation is negated.
+
+Without the lazy allocation benefit, the only difference to "Both escaped and unescaped name" is that we have fewer pointers and sizes in structs.
+This saves some amount of memory and allocations, but makes internal code more difficult to write and understand.
+
+## Decision
+
+Go with "Only unescaped name, with separate namespace" from above:
+
+- Store only unescaped name inside `struct _Key`
```
markus2330, line 144:
```suggestion
- Store only unescaped name with size inside `struct _Key`
```

markus2330 (approved):
Build server does not pass, please rerun.

doc/decisions/3_decided/keyname.md
kodebach: I don't like your suggestion. What is "non-`memcmp`-based"? Where are those benchmarks? If we can't link them here it's still "we don't have the data".

How about:

```suggestion
  Profiling for previous implementations, not based on a single `memcmp` of unescaped names, showed the comparison as a bottleneck, while the current single-`memcmp` implementation does not show the bottleneck.
```

doc/decisions/3_decided/keyname.md
kodebach:
```suggestion
  That said, it may be possible to find a solution slower than the current one that is still fast enough to avoid the previous bottleneck.
```

kodebach:
@markus2330 can this be merged?

PR #4716, opened 2022-11-26
[decisions] errors

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


kodebach:
@markus2330 Like in #4715 the decisions are fully filled in drafts, because I hope to get this decided quickly. Both of these are essentially "keep status quo" decisions with some extra guidelines for the future, so there is not that much to discuss here.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+For release builds, the `assert` won't be there.
+Instead, we return `NULL`, so the caller can do something and doesn't have to deal with the internal bug.
+
+This overlapping of error and success values has especially positive effects, when we talk about boolean functions:
+
+```c
+// returns 1 if key is "foo", 0 if key is not "foo", and -1 if key == NULL
+int keyIsFooInt (Key * key);
+
+// returns true if key is "foo", false otherwise (including if key == NULL)
```
markus2330, line 127: Seems to me quite unrelated.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+**Suggestion:** Go with "Pragmatic solution"
+
+## Rationale
+
+Under the current assumptions this is the obvious solution.
+Every other solution borders on over-engineering.
+
+## Implications
+
+- Some functions (e.g., `keyIsBelow`) should have simplified return values
```
markus2330, line 214: I don't think this is an implication of that decision. If we use boolean for functions returning a boolean is imho a separate decision. I am open to that, I agree that it is weird to introduce a third case just for passing NULL.

doc/decisions/0_drafts/errors_plugins.md
```diff
@@ -0,0 +1,103 @@
...
+### Simple guidelines
+
+Create simple guidelines for plugin authors and clearly document them.
+Rely on plugin authors adhering to the guidelines and testing their plugins themselves.
+Plugins that do not adhere to the guidelines can be marked down via `infos/status`.
+
+The basic guidelines will be (but may be extended):
+
+1. During `kdbGet` only emit warnings, unless the error is entirely unrecoverable:
+   Whatever other guidelines say, during `kdbGet` we plugins shall only emit warnings to allow the caller to fix issues via `kdbSet`.
```
markus2330, line 38:
```suggestion
   Whatever other guidelines say, during `kdbGet` plugins shall only emit warnings to allow the caller to fix issues via `kdbSet`.
   If possible, the emitted warnings should be identical to the errors set in `kdbSet()`.
```

doc/decisions/0_drafts/errors_plugins.md
```diff
@@ -0,0 +1,103 @@
...
+
+Leads to inconsistent behaviour and hard to diagnose interaction issues.
+
+### Detailed ruleset with conformance tests
+
+Write a detailed formal ruleset and generic conformance tests for all plugins.
+
+Defining the ruleset, let alone writing the tests, for all possible errors in plugins is borderline impossible.
+
+### Simple guidelines
```
markus2330, line 29: I fully agree with this decision. You can directly place the guidelines to `doc/contrib` in the next decision round.

doc/decisions/0_drafts/errors_plugins.md
```diff
@@ -0,0 +1,103 @@
...
+
+Create simple guidelines for plugin authors and clearly document them.
+Rely on plugin authors adhering to the guidelines and testing their plugins themselves.
+Plugins that do not adhere to the guidelines can be marked down via `infos/status`.
+
+The basic guidelines will be (but may be extended):
+
+1. During `kdbGet` only emit warnings, unless the error is entirely unrecoverable:
+   Whatever other guidelines say, during `kdbGet` we plugins shall only emit warnings to allow the caller to fix issues via `kdbSet`.
+   Only if errors are entirely unrecoverable, should an error be emitted.
```
markus2330, line 39:
```suggestion
   Only if errors are entirely unrecoverable, an error should be emitted.
```

doc/decisions/0_drafts/errors_plugins.md
```diff
@@ -0,0 +1,103 @@
...
+      Emit a warning.
+      We may still be able to overwrite the file in `kdbSet`.
+   2. File-based backend cannot access file, because path is invalid (e.g., on Windows `C:\foo\CON`)
+      Emit an error.
+      A `kdbSet` could recover, but the mountpoint definition needs to change.
+      To do this, a new `kdbOpen` is needed.
+
+2. Prefer errors to warnings in `kdbSet`:
+   Most validation detect either valid or invalid values.
+   Only in the rare case, where something is valid, but might not be correct should a warning be used.
```
markus2330, line 52: Can you please give an example?

```suggestion
   Only in the rare case, where something is valid, but might not be correct a warning should be used.
```

doc/decisions/0_drafts/errors_plugins.md
```diff
@@ -0,0 +1,103 @@
...
+
+   For very common mistakes which are still technically valid configuration, both emitting an error and emitting a warning are valid options.
+   But in both cases, there should be a way to suppress or avoid the error/warning.
```
markus2330, line 55: I wouldn't recommend that as it would require config options to most plugins. We should be careful with such warnings, if we even need them (looking forward to examples).

doc/decisions/0_drafts/errors_plugins.md
```diff
@@ -0,0 +1,103 @@
...
+3. Don't stop after the first error:
+   Additional errors, are automatically converted into warnings.
+   Report as many errors as possible in one go, so the user can fix as much as possible in one go.
+
+4. Specifications are exact:
+   Both the checked config must conform **exactly** to the specification.
+   Plugins should not ignore unknown values or values they don't understand.
+
+5. Reject unknowns:
+   To be on the safe side, always reject any unknowns that are direct at the plugin.
```
markus2330, line 66: "direct at"?

doc/decisions/0_drafts/errors_plugins.md
```diff
@@ -0,0 +1,103 @@
...
+   This mostly applies to storage plugins.
+   For example, if a plugin handles `meta:/foo`, it should emit an error, if encounters an unknown value for `meta:/foo`.
+
+   Of course, plugins can decide that they are not handling certain keys or metakeys.
+   But that should only be based on the keyname, never the value.
+
+6. The `storage` phase should handle everything during `kdbSet`:
+   A plugin executed in the `storage` phase of `kdbSet` should store the entire `KeySet` they receive, such that in a future `kdbGet` they will load an equal `KeySet`.
+   If such a plugin cannot handle the entire `KeySet` semantics (e.g., no metadata), the unsupported data should be removed in an earlier phase.
+   Either by the plugin itself, or by another plugin.
```
markus2330, line 78:
```suggestion
   Either by the plugin itself, or by another plugin that is required by said plugin.
```

markus2330 (changes requested):
I am not sure if we are on the same page in errors_core. About boolean is for me a totally unrelated decision.

errors_plugins is for me a no-brainer: I totally agree with the decision.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
+# Errors without `errorKey`
+
+## Problem
+
+We often use a `Key * errorKey` argument to return error information via metadata (see related decisions).
+In `libelektra-core`, however, that would be inappropriate and sometimes even impossible (e.g., `keyNew`).
+
+Therefore, we need a different concept for errors, when an `errorKey` cannot be used.
```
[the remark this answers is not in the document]
kodebach: I see your point, but I'll phrase it differently. I don't think it is as big a problem as your suggestion makes it sound like.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+  Note, however, this does not mean checking for an error should be enforced.
+  There are many cases, where the caller knows an error is impossible, because of the previous code path.
+  The requirement is analogous to the `Key * errorKey` argument.
+  The presence of the argument makes it obvious that errors _can_ happen, but you can easily omit the error check when you want to.
+- Adding error indications should not make the API less convenient to use.
+
+## Assumptions
+
+- In many cases, it is possible to avoid error cases through the preceding code path.
+- When an error occurred, the kind error can be distinguished after the fact in many cases.
```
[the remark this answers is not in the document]
kodebach: Well that depends on what "the problem" is defined as... If it is "how to report errors, so the user can distinguish them?", then yes, this assumption invalidates that question. But the problem is more general: "how to report errors, without using `Key * errorKey`?".

The assumption just says, it is not always necessary to report separate errors values for every error case to let the caller distinguish between errors. Maybe, I need to rephrase it...

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+
+Use a field in thread local storage (TLS) to store errors similar to `errno`.
+
+This leaves the API design entirely open.
+
+TLS was only standardized in C11, while Elektra currently only requires C99.
+
+This has all the same issues as `errno`: needs to be cleared to detect errors, can only store one error (per thread), etc.
+
+### Pragmatic solution
```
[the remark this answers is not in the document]
kodebach: 1. "next to memory errors" doesn't make it clear whether every function has 1 (for everything) or 2 (for memory and other) error values.
2. "Redesign API" is a new idea, which I'll add. But I believe it would be unnecessarily restrictive. For example, you couldn't have a `elektraCheckEscapedName` function that returns different values for different problems with the name (illegal escape, invalid namespace, etc.). You could argue those are not "error cases", but then what's the point of the restriction in the first place.
3. "Require users to check preconditions" goes partly in the direction of "Pragmatic Solution". I don't quite understand how it would work though. So the user has to do all the precondition checks. But what happens if they don't? the part about "against our principle" suggests, the APIs function would not check the preconditions. So do we just try and hope there isn't a segfault?
   If that's the suggestions, then sure I can add that. But that seems pretty stupid and I don't really think adding it as an option helps in anyway.
   If the API functions should still check the preconditions they obviously have to react somehow, which probably still means returning an error, but maybe just always the same error. In that case, it's basically "Pragmatic Solution".
   So I think I won't add this one...

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+
+The pragmatic solution here is to simply assume "everything will be fine".
+
+Many functions only return errors for simple precondition violations, like passing a `NULL` pointer or otherwise invalid argument.
+Especially for `libelektra-core` (the only place where `Key * errorKey` truly cannot be used), all functions must be side effect free.
+Therefore, the only possible cause of errors are invalid arguments.
+
+If we now assume the function is written such that it fails before it changes any argument's data, then we can infer that only a single error indicating return value is needed.
```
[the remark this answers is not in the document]
kodebach: Probably should phrase it differently move it to be an actual assumption. It's not just about API design (which we control, yes). What this is about is more that we assume, it is always _possible_ to write the function such that it detects errors and returns without permanent changes.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+Because the arguments are unchanged and the only possible cause of errors, we can distinguish between errors after the fact.
+
+As stated, many functions' only error cases are precondition violations.
+If the preconditions are simple enough, they can be checked beforehand if it is not clear whether the precondition holds.
+
+Take for example some hypothetical `keyValue` function, in a world where we have a `key->value` container for COW that should always be present.
```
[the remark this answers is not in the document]
kodebach: I'll try to come up with something. But it will probably still be hypothetical. I wanted simple cases, where we can focus more on the preconditions and less on what the function does. But I also didn't want to use something, where we're already planning to change the API.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+Looking just at the error checks, without even considering the function signature:
+
+```c
```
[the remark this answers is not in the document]
kodebach: No, actually in this case it is about what the function does. This part is about how we treat different kinds of errors, so we need to see inside the function.

Also a big part of the argument here is that, the user side matters far less than the current API suggests. I strongly believe the in huge numbers of cases, the caller won't even check the return value, and if the do, most of the time they just need to know _if_ there was an error not _what kind_ of error.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+  return key->value->data;
+}
+```
+
+We return `NULL` for both error cases and even let the errors overlap with a legitimate success value.
+In the first case, it's fine for multiple reasons:
+
+- Many callers won't care, if the value inside the key is `NULL` or the key itself is `NULL`.
+  For example, `keyValue (ksLookup (ks, name)) == NULL` is probably fine either way.
+- If the caller does care, they can easily check `if (key == NULL)` before _or_ after the `keyValue` call.
```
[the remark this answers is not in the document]
kodebach: No, that's cover further down with the `ksLookupByName` example. It has the same "is the escaped name valid" problem, which I assume is what you're referring to?

I didn't want to use `keyNew` or `keySetName`, since they are likely to change with the whole (un)escaped name stuff. Especially `keyNew` shouldn't use the escaped name anymore IMO (to be discussed in #4244).

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+
+Essentially, we do nothing, "everything is fine":
+
+```c
+const void * keyValue (Key * key) {
+  if (key == NULL) {
+    return NULL;
+  }
+
+  assert(key->value != NULL);
```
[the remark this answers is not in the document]
kodebach: You probably missed it and I probably should have used a different function name. In this hypothetical example, `key->value` is not the value itself, but the COW container. And also just in the hypothetical example (not in the real world) this should always be there.

Again, I wanted a very simple example to showcase this "internal precondition" case. I didn't use the name, because that cannot be `NULL`. So `NULL` is completely free to be used as an error value and doesn't clash (something I also wanted to show). Maybe I'll split it into multiple examples. I ended up with more examples than expected anyway...

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,222 @@
...
+
+- Many callers won't care, if the value inside the key is `NULL` or the key itself is `NULL`.
+  For example, `keyValue (ksLookup (ks, name)) == NULL` is probably fine either way.
+- If the caller does care, they can easily check `if (key == NULL)` before _or_ after the `keyValue` call.
+
+The second case has a different reason.
+The error is not the caller's fault.
+It is an internal issue the caller can do nothing about, `key->value` should always be there.
+That's why we add an `assert` before the check.
+This will trigger in a debug build, where the internal error can be fixed.
```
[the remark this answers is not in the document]
kodebach: I'm not sure what you mean here....

Yes, that's the point. Only Elektra's debug builds should check these internal conditions. The debug builds are used in our CI (at least I hope so) to run all our tests. That should guard against these issues.

In a release build I think it is bad form for a library to call `abort()` and kill the process. Especially, since we definitely could do something else (like `return NULL`) in this case.

A further note (which I probably should add) the extra `if` check after the `assert` should only be there, if it prevents something like a segfault (like in this case). If it's something more benign like "if the assert doesn't hold, we leak memory" then there shouldn't be an extra check.

However, this is the part of the suggested solution I'm most flexible about. I do think we should have separate rules for internal conditions that the user may not even be able to check. But I'm open to different solutions for them.

doc/decisions/0_drafts/errors_core.md
kodebach: The connection to this decision is: Not all functions need a specific error value. Sometimes it is good enough or maybe even better, to use a default value for things like `NULL` arguments.

doc/decisions/0_drafts/errors_plugins.md
kodebach: With "next round" do you mean the next PR, once this one is merged, or still within this PR?

doc/decisions/0_drafts/errors_plugins.md
kodebach: Maybe not phrased well, but it doesn't have to be a config option. And IMO a config option would be a bad choice because it is too global. It could simply be an additional metakey, or maybe a different value, to indicate you know what you're doing.

One example (which I will mention in the file) I know for a case like this is in the new `backend` plugin:

https://github.com/ElektraInitiative/libelektra/blob/8fe1340ea7f7bc37aa2f14bc0d331d954cf0e216/src/plugins/backend/backend.c#L235-L243

Technically a correct configuration, but in most cases probably not what the user intended. If they did want to do it this way, there is a clear way to suppress the warning.


Other cases could maybe be in the `spec` plugin. For example, if we do add a check to protect against mixed array and non-array children. There could be an exotic use case where this is still needed. It would make sense to allow adding something like `meta:/array/mixed = allow` to the array parent `spec:/foo/#` to avoid the warnings on every `kdbGet` and to allow a `kdbSet`.

doc/decisions/0_drafts/errors_plugins.md
kodebach: Should be "directed at"

```suggestion
   To be on the safe side, always reject any unknowns that are directed at the plugin.
```

I needed a general term for "things the plugin should handle". Not all plugins are triggered by presence of metadata. Some (like storage plugins) try to handle everything, others get a parent from somewhere, etc.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
+# Errors without `errorKey`
+
+## Problem
+
+We often use a `Key * errorKey` argument to return error information via metadata (see related decisions).
```
markus2330, line 5: Either directly link (if the decisions must be known to understand this decision) or it is enough if related decisions are linked below in its section.

```suggestion
We often use a `Key * errorKey` argument to return error information via metadata.
```

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+  Note, however, this does not mean checking for an error should be enforced.
+  There are many cases, where the caller knows an error is impossible, because of the previous code path.
+  The requirement is analogous to the `Key * errorKey` argument.
+  The presence of the argument makes it obvious that errors _can_ happen, but you can easily omit the error check when you want to.
+- Adding error indications should not make the API less convenient to use.
+
+## Assumptions
+
+- In many cases, it is possible to avoid error cases through the preceding code path.
+  For example, a very common case is that functions return an error, if they receive a `NULL` pointer.
```
markus2330, line 31:
```suggestion
  For example, a ubiquitous case is that functions return an error, if they receive a `NULL` pointer.
```

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+  Often this case can be excluded by static analysis, i.e., because of the preceding code it is impossible that the pointer is `NULL`.
+
+  ```c
+  int foo (KeySet * ks, Key * key) {
+    if (key == NULL) {
+      return -1;
+    }
+
+    // [...] could lots of code, as long as the Key * key doesn't change
```
markus2330, line 40:
```suggestion
    // [...] could be lots of code, but the code doesn't change `Key * key`
```

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+      // no need to worry about key == NULL, because we checked earlier
+      // also no need to worry about ks == NULL, because the ksLookupByName would have returned NULL
+      const Key * found2 = ksLookup (ks, key, 0);
+    }
+  }
+  ```
+
+- It is not always necessary to report separate error codes, for _every_ error case.
+  In many cases, the caller can distinguish after the fact, e.g., `keySetName` fails when the given name is invalid and when the given `Key` has a read-only name.
+  This can be distinguished after the fact, by checking whether the `Key` has a read-only name with `keyIsLocked`.
```
markus2330, line 53: I agree it can be distinguished but it still might be a problem. I already saw many errors where it was unclear which is the case because the application failed to distinguish these cases and debugging was needed to find out.

A failed KeySetName because of the name could be an input error but the read-only name would be an logic error (wrong understanding of when KeySetName can be called).

I don't think we should put this under the carpet with an assumption but should search for a solution with the best API usability.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+  For example, only `NULL` is an invalid pointer and sometimes even that is a legitimate non-error result.
+- It should be easy to detect error cases.
+- It must be possible to distinguish between errors, once an error is detected.
+- It should be hard to miss the fact that errors are possible.
+  Note, however, this does not mean checking for an error should be enforced.
+  There are many cases, where the caller knows an error is impossible, because of the previous code path.
+  The requirement is analogous to the `Key * errorKey` argument.
+  The presence of the argument makes it obvious that errors _can_ happen, but you can easily omit the error check when you want to.
+- Adding error indications should not make the API less convenient to use.
+
```
markus2330, line 27:
```suggestion
- It should be [hard to use the API the wrong way](doc/DESIGN.md).

```

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+- `nameNew(name) -> name_object` only has the error case of invalid name
+- `keyNew(name_object)` won't have error cases anymore
+- `keySetName(name_object)` only has the error case of read-only names
+
+This can be very limiting and in some cases makes it very hard to find out what the exact problem was.
+Reducing a function to a single error value, makes sense if the errors are easy to distinguish.
+But for more complex preconditions, this can be annoying and inefficient to use.
+While the above example can be solved, by simple checking with `keyIsLocked`, there could less straightforward cases.
+For example, a function `foo` that takes two keynames, both of which must be valid.
```
markus2330, line 99: Please use examples from our API.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Redesign API
+
+Redesign the API, so that every API function only has maximal one error case next to memory errors.
+
+- `nameNew(name) -> name_object` only has the error case of invalid name
+- `keyNew(name_object)` won't have error cases anymore
+- `keySetName(name_object)` only has the error case of read-only names
+
+This can be very limiting and in some cases makes it very hard to find out what the exact problem was.
```
markus2330, line 95:
```suggestion
The advantage is that it is hard to be used wrongly.
Users can only call keySetName if they have a valid key name.
If keySetName fails nevertheless, we know that it must be because the name was read-only.
No further debugging would be needed.

This can be very limiting and in some cases makes it very hard to find out what the exact problem was.
```

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+
+If caller are interested in the specific errors, they need to do extra checks, e.g., before calling `keySetName`, they call `elektraCheckName(name)` or `keyIsLocked`.
+
+This is not so intuitive and is against our principle to make it hard to use the API wrong.
+
+It's also unclear, what the function would do, if the caller did not check the preconditions.
+
+### Pragmatic solution
+
+We assume the functions in question are side-effect-free and the only possible cause of errors are invalid arguments.
```
markus2330, line 114: This is a constraints for the core, please put it to constraints.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,292 @@
...
+We also assume that functions _can_ always be written such that they fail before any permanent changes to the arguments are made.
```
markus2330, line 115: Please put to assumptions.

doc/decisions/0_drafts/errors_plugins.md
markus2330: Hard to say. Not so ideal that these two decisions are in one PR.

markus2330 (changes requested):
Thank you for the improvements! The "errors" decision becomes too abstract for me, I didn't do a full review due to lack of time today. Impact on the (actual) API should be much more clear. Probably we should first clarify if we introduce key_names objects (constructed from namespace, unescaped key and size), then it is probably much easier to decide this here.

doc/decisions/0_drafts/errors_plugins.md
kodebach: > Not so ideal that these two decisions are in one PR.

I didn't expect this much discussion... I thought the decisions are both very straightforward.

doc/decisions/0_drafts/errors_core.md
kodebach: I would, but there is currently no such function that I know of. Definitely no function takes two keynames and I can't think of one that takes two other arguments with complex preconditions.

I still wanted to give the example, because it shows that this is a valid option, even if we add a function with such complex preconditions in the future.

doc/decisions/0_drafts/errors_core.md
kodebach: It actually already is an assumption in the "Assumptions" section above. I can't make it a constraint, because it's only a constraint for `libelektra-core` not necessarily all APIs that don't (want to) use `Key * errorKey`.

I just mentioned it here as a reminder of the assumption and so the following conclusions are easier to understand, i.e., instead of saying "Because of our assumptions above" and I repeat the assumptions.

doc/decisions/0_drafts/errors_core.md
kodebach: Same as above, it already is in "Assumptions". It's just repeated for easier reading.

doc/decisions/0_drafts/errors_core.md
kodebach: > I already saw many errors where it was unclear which is the case because the application failed to distinguish these cases and debugging was needed to find out.

That sounds like an application problem not an API problem.

> A failed KeySetName because of the name could be an input error but the read-only name would be an logic error (wrong understanding of when KeySetName can be called).

I see two ways this could happen (assuming there is an error check at all):
1. The application developer didn't read the docs of `keySetName` properly and assumed an error always means invalid keyname. As long as our error cases are documented properly, this is clearly an issue with the application, not with the API.
2. The application developer did do everything properly, but _after_ they wrote their application, we added a new error case to `keySetName` and reused the same error value. _That_ is our problem. But one could argue that this is an API breaking change. However, this issue also happens with any other solution, if we reuse error values for unrelated errors _after the fact_. 

There is also the obvious issue that can always happen in C: The application developer might not be checking for errors at all. That's a problem we cannot avoid. So we simply can't write an API that forces callers to check for errors.

> [We] should search for a solution with the best API usability.

That's exactly my point. We can't make a general statement about the API. To get the best API, it has to be decided on a case by case basis.

That's also my argument against the "Redesign API" and "Require caller to check preconditions" options. IMO we will just regret such strict rules at some point and we'll have to revisit the decision, because there is no good option left.

kodebach:
> The "errors" decision becomes too abstract for me, I didn't do a full review due to lack of time today.

The decision is intentionally abstract. I want to decide on general rules, before I start to think about concrete cases. If we start with the existing API and only talk about changes, we get rules that are specific to the current situation. As soon as some comes up with a new API that's a bit different, we'll have to do this all over again. If we decide general rules, we can hopefully use them for a long time.

BTW that's also why I started #4683, #4676 and #4715. All of them are rather abstract and general. Based on those general decision, we can then make specific decisions about parts of the API. I hoped this would lead to less discussion and I still hope that at least the decisions about specific API changes will be quick (if needed at all), because we'll already have rules to guide us.

> Impact on the (actual) API should be much more clear.

If I would list the entire API and the changes for every function this would just be chaos. I'm not going to do that. We need a general solution that can then be applied to the API in further decisions. These decision will be split into coherent parts of the API (e.g. one decision for the keyname API, one for lookups, etc.).

Also the "actual API" is still very much undecided. So I couldn't even list it if I wanted to, without starting discussions about unrelated API changes.

> Probably we should first clarify if we introduce key_names objects (constructed from namespace, unescaped key and size), then it is probably much easier to decide this here.

No, this has nothing to do with this decision. **Please, stop focusing on `keySetName` so much.** We need a general solution of errors, not just for `keySetName`.

----

@markus2330 To make the `errors_core` decision a bit more concrete, I will add a list of guidelines similar to what we have in `errors_plugins`. I hope this will help.

kodebach:
@markus2330 Hopefully, you find the `errors_core` decision a bit more agreeable now.

As a side note, since you seem so hyper-focused on `keySetName`: `keySetName` is actually one of the least problematic functions. It can just return an `int` and use separate error codes for everything. The only problem is that the current API tries to return the size of the new keyname, which very few people will care about and if they do, they can just call `keyGetNameSize`.

kodebach:
@markus2330 Just in case you already started your review: I added another guideline. Please take that into account.

markus2330:
> We need a general solution of errors, not just for keySetName.

I think we don't need a general solution. We better fix the few cases where the return values are not enough.
Outside of `core` we can use the `errorKey` solution (or something similar like ElektraError). So in this decision we don't need general rules but rules specific to the problems in the core. But we need a specific list of functions where there is actually a problem to solve. keySetName is the most important one, hence my focus on it.

If you want to write rules for the whole API, doc/DESIGN.md is the place. These general rules, however, must be much more concise, otherwise it is impossible to apply them in future to new APIs. Please don't try to reinvent the wheel, it is years of efforts to bring the level of consistency we have now in the API. Only problems should be fixed, and not a new design invented. Everything that is just a matter of style/taste (like plural/singular) should stay as it is, it would be simply a waste of time to change this.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -278,12 +287,77 @@ if (found == NULL)
...
+#### Check preconditions
+
+Functions must check all preconditions.
+A (bug-free) function should never segfault, abort the process or otherwise cause unexpected behavior on the caller side.
+All conditions that could cause such behavior must be listed as preconditions and checked by the function as a safety measure.
+
+#### `NULL` arguments
+
+A `NULL` pointer argument should never be treated as a separate error case (unless it is the only error case).
+If a `NULL` pointer constitutes an invalid argument, the function should return an existing error (might be used for other precondition checks) or a default fallback value.
```
markus2330, line 318: Unclear what this means. If we have such a problem, e.g. in the core, the decision should clearly point out the problems.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -278,12 +287,77 @@ if (found == NULL)
...
+- are computationally expensive to check (i.e., duplicating the check would be bad)
+
+should always use separate error codes for each error case.
+The error code may not be shared with other complex preconditions, but can be shared with simple preconditions or `NULL` arguments, since those are easy to distinguish for the caller.
+
+#### Restructure API
+
+To avoid complex preconditions and the need for multiple error codes, the API can be restructured.
+This should, however, be done with caution.
+The API should still remain ergonomic to use and there should not be extra memory allocations creating opaque objects just to guard against possible errors.
```
markus2330, line 339: I agree this is what we both want to reach with this decision. But without details how it is done, the "rule" is worthless. (Imho it is not a rule but the goal of the decision.)

doc/decisions/0_drafts/errors_core.md
```diff
@@ -278,12 +287,77 @@ if (found == NULL)
...
+
+#### Error arguments
+
+If a function needs multiple error codes, return them as an `int` or enum value.
+Do not use an error argument (like `Key * error`).
+Instead, use an output pointer argument (e.g., `Foo * output`) for the data produced by the function.
+This makes it much more obvious to the caller that errors can occur.
```
markus2330, line 346: Rules must be more specific. And as you say yourself in the next paragraph, this cannot be a general rule, as it is library-depend. Please let us focus on core in this decision, otherwise we won't reach anywhere.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -352,6 +352,37 @@ This is allowed, because it has the same expressiveness and use case as a `Key *
 
 Internal errors that are never the fault of the caller and always caused by a bug in Elektra, should be prevented with an `assert`.
 
+#### Secondary return values
+
+In the success case, only return data the caller explicitly asked for by calling the function.
+Do not use secondary information as the return value.
+If there is secondary information, the caller may care about, use an (optional) output pointer argument.
```
markus2330, line 359: I don't understand what you want to say here. What is a "secondary return"?

markus2330 (changes requested):
Thank you for your tedious work. Unfortunately, this decision went astray. It is too unspecific. Please focus on a specific problem (error handling in core) with solutions for the specific functions where there are problems.

doc/decisions/0_drafts/errors_core.md
kodebach: What I wanted to say is: IMO you should not force yourself to invent an error value just for "illegal `NULL` argument". Try to handle it some other way, or combine it with other precondition failures. If "`NULL` argument" is the only error case and you already return an `int`, you can of course use an error value for `NULL`. But it should even at that point be clear, that the same error code will be used if additional preconditions are added in the future.

Some examples:
```c
//////// combines NULL with (potential) other errors
// returns 0 on success
// returns -1 on precondition violations, including key == NULL
int foo(Key * key);

/////// defines logic for NULL, i.e. NULL is not an invalid argument
// returns whether other is below key
// NULL is considered below no key, and no key is considered below NULL, i.e. returns false if at least one argument is NULL
bool keyIsBelow (Key * key, Key * other);

//////// uses default/fallback value for NULL
// returns namespace of key, for key == NULL the fallback value KEY_NS_NONE is returned
elektraNamespace keyGetNamespace(Key * key);
```

> the decision should clearly point out the problems.

Again, no concrete problems. Just a guideline to base APIs on.

doc/decisions/0_drafts/errors_core.md
kodebach: I'm not sure what you want to say here.

This guideline is meant convey that splitting a function into multiple smaller functions with only one failure case each, **can** be an option. But in most cases to the API would only be safe to use, if additional opaque structs are introduced, which are then passed from one function to the next. If this structs aren't used internally somehow and are just temporary allocations, they shouldn't exist. It is better to have a single function returning an `int` with multiple failure cases and avoid the temporary allocations.

doc/decisions/0_drafts/errors_core.md
kodebach: > Rules must be more specific.

Which is why I called them guidelines and even explicitly stated that we don't define strict rules.

A rule should cover all cases and should be followed unless there is a specific exception.

A guideline is just a guide. You should follow it as much as possible, but it is very much expected that you may have to deviate a bit.

> And as you say yourself in the next paragraph, this cannot be a general rule, as it is library-depend. Please let us focus on core in this decision, otherwise we won't reach anywhere.

This is actually already pretty concrete. But if you want I can make it **very** concrete and also library-independent:

---

If a function needs multiple error codes, prefer returning them as an `int` or enum value.

```c
// returns 0 on success
// returns -1 if key == NULL, or a precondition is violated
// returns -2 if error X happens
int foo (Key * key);
```

> **Note:** Sorry, it's late I couldn't come up with a better example, but I think this first part is pretty clear even without a concrete example.

Error arguments should only be used, if the errors are likely to be shown to a user.
Only `Key * errorKey` or a struct which follows the same error concept (e.g. `ElektraError` from the high-level API) should be used as an error argument.
The `Key * errorKey` concept was explicitly designed with the intention of having errors that can be shown to users, therefore it should be used as designed.

```c
/*
 * @retval 0	on success, this is the only case in which @p ks will be modified
 * @retval -1	on error, the error will be set as metadata in @p errorKey
 * @retval 1	if the help option `--help` was found, use elektraGetOptsHelpMessage() access the
 * 		generated help message
 */
int elektraGetOpts (KeySet * ks, int argc, const char ** argv, const char ** envp, Key * parentKey);

/**
 * @param error Pointer used to report errors
 * @retval True if the checks were successful (or not required).
 * @retval False if the checks were required but unsuccessful.
 */
kdb_boolean_t checkSpec (Key * const parentKey, KeySet * contract, ElektraError ** error)
```

If the function produces data, but also needs multiple error codes, prefer returning the error code and using output pointer argument(s) for the data.
By not returning a value other than the data the caller asked for, it is more clear that the value may convey error information.

```c
/*
 * @param variable pointer to the output variable
 * @retval 0 on success
 * @retval -1 on precondition violation (key == NULL, variable == NULL)
 * @retval -2 if key has the wrong meta:/type
 * @retval -3 if the value of key could not be converted
 */
int elektraKeyToChar (const Key * key, kdb_char_t * variable);
```
---

doc/decisions/0_drafts/errors_core.md
kodebach: Basically, that the `keySetName` API (and almost all other APIs that return `ssize_t`) is stupid.

There is no reason to return a size. The caller did not ask for a size. They wanted to change the name. That's an operation with no inherent result, so in theory it should be `void keySetName (...)`. But it can fail, so we should `int keySetName (...)`.

The bit about secondary information is meant to say: If the new size really is valuable information then it should be returned differently, since it is only secondary information and not the primary result of the function:

```c
// on success, if outNewSize != NULL it will be set to the new size of key's name
int keyAddName (Key * key, const char * suffix, size_t * outNewSize);

// if outNewSize != NULL it will be set to the new size of ks
void ksAppendKey (KeySet * ks, Key * key, size_t * outNewSize);
```

kodebach:
> I think we don't need a general solution. We better fix the few cases where the return values are not enough.

**NO** hard disagree on that one. Focusing just on the problems we have is not helpful. It just means when somebody wants to add a function the future and has the same problem, we have to discuss everything from scratch again. If we decide on a general solution future decision are much easier. But of course since we cannot predict the future, I'm heavily in favour of being pragmatic and just defining guidelines. We'll still have to decide the detailed error handling for a new API in the future, but with the guidelines we at least have something to start with.


> Outside of `core` we can use the `errorKey` solution (or something similar like ElektraError).

Not true. There are lots of cases outside `libelektra-core` where a `Key * errorKey` makes no sense. The easiest example is `libelektra-extra`, which is where we want to move `ksCut`. Using an `errorKey` for `ksCut` would be quite weird. Another example are the conversion functions `elektraKeyTo*`. Those are just off the top of my head, I'm sure I can come up with more examples, if I think longer. The `errorKey` concept is pretty advanced and pretty much only makes sense, if the error is likely to be shown to a user. Otherwise, it is just a waste of memory and makes checking for errors more complex. Basically anything where the error is not directly intended to be shown to a human has no reason to use an `errorKey`. The whole concept of `errorKey`s is designed for messages shown to humans.

> But we need a specific list of functions where there is actually a problem to solve.

I can't give you such a list, because we have yet to decide what the future API will look like. I wanted to have general decision on errors, so we have on thing less to discuss for each an every new API function. If we do not have a general concept, I will always think "Is it a good idea proposing the new function this way? Will this just cause discussion about error handling?". I mean, yes, I have already thought about error handling in general, so I will just follow my proposed rules anyway. But the issues, I'll have to argue for and explain the reasoning behind the error handling for every function separately, instead of just saying "I followed the agreed upon guidelines".

> keySetName is the most important one, hence my focus on it.

As I said already, but you still seem to miss/overlook/ignore: It's one of the easiest to fix. Just return an `int` instead of the useless `ssize_t`. In fact the `ssize_t` is even a problem in and of itself. Internally it's a `size_t`. While unlikely to happen, there are **a lot** of possible keynames where `size_t` can hold the size but `ssize_t` would overflow.

> If you want to write rules for the whole API, doc/DESIGN.md is the place. These general rules, however, must be much more concise, otherwise it is impossible to apply them in future to new APIs.

So what is the decision process for then? I thought we want a structured process to decide things, exactly to avoid long drawn out discussions in PRs that modify things like `doc/DESIGN.md`.

> it is years of efforts to bring the level of consistency we have now in the API.

:laughing:

Sorry, but Elektra's API is one of the least consistent designs I've seen, apart from things that have been thrown together with no intention of lasting for a long time. `libelektra-core` on it's own may be fairly consistent, but that's mostly because up until recently nobody dared to touch them and add or change anything. If you look beyond, you find all manor's of APIs, all designed in whatever way the author felt like that day. There are even attempts at functional APIs, which is just utterly ludicrous in C.

> Only problems should be fixed, and not a new design invented. Everything that is just a matter of style/taste (like plural/singular) should stay as it is, it would be simply a waste of time to change this.

IMO the design/concept **is** the problem. That's what I tried to say with the problem statement of the decision. We don't have a concept for "errors without `Key * errorKey`". **That** is the problem.

kodebach:
@markus2330 Can you please review this again. I haven't made any changes. But I think if you consider my response above **and** also take #4805 into account (where I applied the guidelines laid out here), you may come to a less negative conclusion than before.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,403 @@
...
+
+## Problem
+
+We often use a `Key * errorKey` argument to return error information via metadata.
+In `libelektra-core`, however, that would be inappropriate and sometimes even impossible (e.g., `keyNew`).
+
+Not using a `Key * errorKey` argument can be limiting, however.
+Depending on the solution, there can be limited error values, i.e., it may not be possible to use a different error value for every possible error case.
+A prime example of this, a function that returns a pointer and wants to indicate an error via that pointer, only has to option to return `NULL`.
+So if there is more than one possible error, the error cases have to be distinguished another way.
```
markus2330, line 11: The question is if this is actually the case, and in which functions. This would be a decision specific enough. But I already said something along the lines below.

doc/decisions/0_drafts/errors_core.md
markus2330: > We can't make a general statement about the API. To get the best API, it has to be decided on a case by case basis.

But we also cannot do arbitrary changes. There must be concepts behind the API, that are fulfilled consistently to the point. E.g. the return value 1 means success, 0 no change, -1 error. To change something like this, a decision can be a good method, so it is a good idea to have such a decision. At the moment the decision is just to unfocused to be useful.

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,403 @@
...
+
+Always limiting a function to a single error code can be very limiting and in some cases and makes it very hard to find out what the exact problem was.
+Reducing a function to a single error value, makes sense if the errors are easy to distinguish.
+But for more complex preconditions, this can be annoying and inefficient to use.
+For example, a function `foo` that takes two keynames, both of which must be valid.
+To find out which of the names was invalid, the caller would have to call something like `elektraCheckName(name)` on at least one name.
+That's a non-trivial check that is not cheap in terms of runtime.
+More importantly, it is a check that `foo` already did, and the result could just be propagated to the caller.
+
+### Redesign API
```
markus2330, line 98: Such redesign is better done in #4805. The decision here simply would be "Limit to single error value" then.

doc/decisions/0_drafts/errors_core.md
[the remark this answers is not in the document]
markus2330: > For example, you couldn't have a elektraCheckEscapedName function that returns different values for different problems with the name (illegal escape, invalid namespace, etc.). 

Something like this should be written in the decision. (Btw. it was just by accident that I opened this hidden/resolved conversation.)

doc/decisions/0_drafts/errors_core.md
```diff
@@ -0,0 +1,403 @@
...
+This is not so intuitive and is against our principle to make it hard to use the API wrong.
+
+It's also unclear, what the function would do, if the caller did not check the preconditions.
+
+### Pragmatic solution
+
+We assume the functions in question are side-effect-free and the only possible cause of errors are invalid arguments.
+We also assume that functions _can_ always be written such that they fail before any permanent changes to the arguments are made.
+
+If we assume that _all_ functions are indeed written that way, then we can conclude, that in theory a single error indicating return value is always enough.
```
markus2330, line 129: No assumption needed, it would be the new guideline that functions must be that way. (And reviewed that way, see #4426)

markus2330 (changes requested):
Thanks for the reminder. A few answers.

doc/decisions/0_drafts/errors_core.md
kodebach: For example `ksLookup`. The return value `NULL` could mean either "key not found" or "invalid arguments". Technically it may not apply 100%, since I wouldn't consider "key not found" an error case. However, conflating a non-error return value with error cases is arguably even worse.

If we go with the more general case (not just limited to pointer return value), then there is also `ksSearch`. Where `-1` means basically the same as `NULL` for `ksLookup`.

If you want a very clear-cut example in the current API then just look at `keySetName`. It definitely conflates multiple error cases into `-1` (read-only key, invalid name, etc.).

doc/decisions/0_drafts/errors_core.md
kodebach: > There must be concepts behind the API, that are fulfilled consistently to the point.

That's the whole point of this decision: Coming up with these concepts/rules.

I'm pretty sure we are on the same page, here and are just "talking past each other". I'll resolve the thread.

doc/decisions/0_drafts/errors_core.md
kodebach: Not sure what you want to say here. I can rename the solution, if you like, but not sure what this is going to change...

```suggestion
### Limit to single error value
```

doc/decisions/0_drafts/errors_core.md
kodebach: If you read on, this solution actually argues **against** introducing such a strict rule, because it could introduce unwanted restrictions.

I'll give you an example that shows how tedious such a rule could become. (I'll add something like this to the file as well).

Let's first assume there is a function `foo()` that needs a valid escaped keyname as an argument. There are multiple reasons why an escaped keyname could be invalid. The user may want to differentiate between those reasons. With the strict "1 error value" rule, the result of `foo()` can't tell us what the issue with the keyname is. But that's fine, we'll just add a `checkName()` function. However, that still must follow the "1 error value" rule. So what we would actually need is `checkNameProblem1()`,  `checkNameProblem2()`, etc. each of which just checks 1 potential problem and the user has to call all of them.

Of course, I assume, that is not what you actually want. So I once again think we are on the same page here. The rule should be: "Try to use only 1 error value, but a useful API is more important" That is more or less, what this solution argues for, just in lot's more detail, because I wrote down all the cases I thought about.

> **Note**: I know the example is somewhat abstract and doesn't use examples from the API. But since we are talking about potential rules for unknown future API, we should think abstractly and not limit ourselves to only the cases we have right now.

kodebach:
@markus2330 Not sure how to best proceed here. Maybe we merge these decision as drafts ASAP and then see where things go.

The goals of the decisions are:
- `errors_core`: Partly document status quo, but also decide how errors are handled in the new API (#4805). The decision is medium to low priority. IMO it might be easier to design the API and then document, how errors are handled and why we do it like that.
- `errors_plugins`: Document the status quo of how plugins should handle errors. The end goal is to have a set of guidelines for plugin authors.

kodebach:
@markus2330 Can I rebase and merge this as-is?

Issue #3954, opened 2021-07-21
spec: tags for defaults in arrays/tables/maps


kodebach:
## Status Quo

> **Note:** if you already know about the problems with the `_` and `#` wildcards in specifications, skip to the next section

When writing a specification, there are 2 basic kinds of keys you can specify:
1. Fixed Name Keys
    For keys that have a fixed name (e.g. `server/port`) you can just define the specification directly.
    ```ini
    # ni format
    [server/port]
    type = unsigned_short
    default = 8080
    ```
2. Dynamic Name Keys
    Sometimes there are cases where you don't know the full key name when writing the specification. Some parts of the name are chosen by the user. Currently we have `_` (single non-array part) and `#` (single array part) as wildcards for these cases.
    ```ini
    # ni format
    [subdomains/_/port]
    type = unsigned_short
    description = "port to be used for a subdomain, i.e. to route foo.example.com to port 8081, set subdomains/foo/port = 8080"
    
    [locales/#/country]
    type = string

    [locales/#/language]
    type = string
    ```

However, the current implementation has some serious limitations (especially for `_`) when it comes to default values. As long as you don't use `default`, both `_` and `#` work fine. The spec metadata is copied over to the existing keys and those will be checked by various plugins. But if there is no key matching the wildcard and a key should be generated from the `default` spec, things get tricky. The `_` wildcard cannot be used at all for generating keys (because we don't know which keys to generate in advance). The `#` has some limitations as well. If a default array size was specified, then we can generate that number of keys. But if there is no default, or we want something other than the default size, then the desired array size has to be stored somewhere. That means storage plugins need to be able to store the `array` metadata. Another issue is, if we want the default for most array elements, but a few of them should be overriden. That would require storing "arrays with holes".

All of this leads to some awkward specifications where the `#` is used in a very specific ways to circumvent all the problems.

## Differentiating use cases

I see a few different use cases for the `#` and `_` wildcards:

- A simple array of values (`foo/#`): Sometimes you need a simple array of strings/number/etc. For example, this could be a list of files that should be read.
- A simple map of values (`foo/_`): An array is not always sufficient, sometimes you need a set of key-value pairs, or mappings. For example, in an embedded application you may need to map keys on a keypad to functionality.
[...]

PR #4743, opened 2022-12-03
Decision: Spec Expressiveness

tmakar:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels

### DIFF
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -18,12 +18,7 @@ A solution to this is `globbing` i.o. using wildcards in key names.
 The thing to be aware of is to define configuration keys which are necessary for every service.
 In this example we use `port` and `url`.
 
-The key starts with `/sw/org/app-name/#0/development`.
-`sw` defines that this is a configuration for a software.
-`org` is the organization name.
-`app-name` is the application name we are writing the specification for.
-`#0` is the configuration version.
-`development` is the profile (it could also be `production`) 
+The parent key for our specification is `/sw/org/app-name/#0/development`.
 
 ```ni
 [service/_/port]
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -36,7 +36,7 @@ The `_` would match every `service` we are using in our application e.g. `databa
 
 This specification uses `_` but we also have `#` as wildcard in `globbing`. 
 The problem we face here is overlapping specifications.
-A example for this is if we used `[service/#/port]` and `[service/_/port]` in the same specification.
+A example for this is if we used `[service/#/port]` and `[service/shtg/port]` or `[service/_/port]` in the same specification.
 This is currently undefined behaivour in the spec plugin.
 
 ### Problem 2
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -8,6 +8,8 @@ We need to find minimal requirements to implement a sane spec plugin.
 
 ## Constraints
 
+TODO
+
 ## Assumptions
 
 ## Considered Alternatives
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -2,9 +2,33 @@
 
 ## Problem
 
-Currently, you can easily come into wrong assumptions that something would work in the specification.
+The current implementation of the spec plugin is copying metadata from the specification
+(`spec` namespace) to all other namespaces. By providing certain metadata e.g. 
+`type` or `description`, it can be assured that keys adhere to a given specification. 
 
-We need to find minimal requirements to implement a sane spec plugin.
+For some applications there are keys we don't want to define for every application explicitly.
+This is when `globbing` expressions in key names comes into place. By providing certain wildcard
+characters e.g. `_` or `#` one can match and specify a number of keys at once without the need
+to explicitly state the key in every specification.
+
+To describe the use case where `globbing` expressions make sense, we will use an application
+based on microservice architecture as sample.
+
+Imagine we have microservices and each of them communicates to a database to persist data. We
+will need some configuration so the applications can connect to the database e.g. `port`. Lets
+say we are specifying our configuration by `database/<DIALECT>/...`, where `...` is any name
+for a key. If we have multiple dialects e.g. a microservice uses more than one database e.g. `mysql`, `mssql`, etc. it is inconvenient to copy and paste the configuration for every 
+dialect we use.
+
+By using `globbing` expression we can overcome this problem.
+
+The key starts with `/sw/org/app-name`.
+```ni
+[database/_/port]
+meta:/require = true
+meta:/description = "The database port."
+```
+The `_` would match every `dialect` we are using in our application e.g. `postgresql`, `mssql`, etc.
 
 ## Constraints
 
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -2,14 +2,14 @@
 
 ## Problem
 
-The current implementation of the spec plugin is copying metadata from the specification
-(`spec` namespace) to all other namespaces. By providing certain metadata e.g. 
-`type` or `description`, it can be assured that keys adhere to a given specification. 
+The current implementation of the spec plugin is copying metadata to all other namespaces.
+Metadata e.g. `type` or `description` describes a certain key.
 
-For some applications there are keys we don't want to define for every application explicitly.
-This is when `globbing` expressions in key names comes into place. By providing certain wildcard
-characters e.g. `_` or `#` one can match and specify a number of keys at once without the need
-to explicitly state the key in every specification.
+Sometimes there are keys with similar or identical specifications.
+We want to define those in a single place and not have to manually copy metadata.
+`Globbing` expressions in key names comes into place at that time. 
+By providing certain wildcard characters e.g. `_` or `#` one can match multiple keys.
+This removes the need to explicitly define specification for keys.
 
 To describe the use case where `globbing` expressions make sense, we will use a REST application.
 This REST application uses multiple services which the user is not aware of at time of configuration.
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -19,7 +19,7 @@ meta:/description = "The port on which this application is accessible."
 Here one can see `default` and `description` used as metadata. 
 
 The current implementation of the spec plugin is copying this metadata to all other namespaces.
-It also assures that if keys do not exist and have `default` metadata, it creates a cascading key.
+It also assures that if keys do not exist and have `default` metadata, it creates a key in the default namespace.
 
 Assume we have a REST application we want to write a specification for.
 This REST application has to use multiple services.
@@ -80,7 +80,7 @@ meta:/description = "The url of the service."
 ```
 The `_` would match every `service` we are using in our application e.g. `database`, `message broker`, etc.
 
-But the current implementation has some undefined behaivour when using `globbing`.
+A naive implementation has some undefined behaivor when using `globbing`.
 
 This specification uses `_` as wildcard character.
 Unfortunately `_` has a shortcoming, as we do not know the key names in advance.
@@ -93,7 +93,7 @@ A fixed array size can be defined in the specification, which overcomes the shor
 The problem we face with those wildcard characters is overlapping specifications.
 A example for this is if we used `[service/#/port]` and `[service/shtg/port]` or `[service/_/port]` in the same specification.
 We don't know what metadata should be copied to other namespaces in this case.
-Both `[service/#/port]` and `[service/_/port]` existing produces conflicting specification.
+Defining, both `[service/#/port]` and `[service/_/port]` produces conflicting specification.
 
 For example `[service/#/port]` having as type `meta:/type = string`.
 `[service/shtg/port]` having `meta:/type = unsigned_short`.
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -56,12 +56,7 @@ The validation part of the `spec` plugin can be kept as it is now (`required key
 
 ## Considered Alternatives
 
-## Decision
-
-- no defaults for keys ending with globbing keys `sw/_` (**Problem 3**)
-- no defaults for `sw/_/keys` (default will not work for `ksLookup(/sw/sthg/key)`)
-- The spec plugin should yield should yield errors to user on overlapping keys (**Problem 1**)
-- The spec plugin should yield errors and warnings when it detects such situations (**Problem 2**)
+## Decision 
 
 ## Rationale
 
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -2,8 +2,24 @@
 
 ## Problem
 
-The current implementation of the spec plugin is copying metadata to all other namespaces.
-Metadata e.g. `type` or `description` describes a certain key.
+Specifications can help users define certain details about keys and make sure only valid keys are used.
+
+One example of such as specification can be found below:
+
+```ni
+[public/ip]
+meta:/default = 127.0.0.1
+meta:/description = "The IP via which this application is accessible."
+
+[public/port]
+meta:/default = 8080
+meta:/description = "The port on which this application is accessible."
+```
+
+Here one can see `default` and `description` used as metadata. 
+
+The current implementation of the spec plugin is copying this metadata to all other namespaces.
+It also assures that if keys do not exist and have `default` metadata, it creates a cascading key.
 
 Sometimes there are keys with similar or identical specifications.
 We want to define those in a single place and not have to manually copy metadata.
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -14,7 +14,7 @@ This removes the need to explicitly define specification for keys.
 To describe the use case where `globbing` expressions make sense, we will use a REST application.
 This REST application uses multiple services which the user is not aware of at time of configuration.
 It could use a database, message broker, etc. but this is not yet known.
-A solution to this is `globbing` i.o. using wildcards in key names.
+A solution to this is `globbing` i.e. using wildcards in key names.
 The thing to be aware of is to define configuration keys which are necessary for every service.
 In this example we use `port` and `url`.
 
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -138,11 +138,7 @@ But this adds another key to give us a meaningful name for a given service.
 
 This is why we want to have the wildcard character `_` working correctly.
 
-Besides this problem the `spec` plugin also creates default values.
-A plugin responsible for copying metadata to other namespaces should not create new keys.
-In this case it creates cascading keys which are then found by `ksLookup`.
-In my opinion a separate plugin for handling `default values` should exist.
-The validation part of the `spec` plugin can be kept as it is now (`required keys` and `array size validation`).
+Besides this problem the `spec` plugin also creates default values if the keys do not exist, which is not what someone would expect from this plugin.
 
 ## Constraints
 
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -21,16 +21,48 @@ Here one can see `default` and `description` used as metadata.
 The current implementation of the spec plugin is copying this metadata to all other namespaces.
 It also assures that if keys do not exist and have `default` metadata, it creates a cascading key.
 
-Sometimes there are keys with similar or identical specifications.
-We want to define those in a single place and not have to manually copy metadata.
-`Globbing` expressions in key names comes into place at that time. 
-By providing certain wildcard characters e.g. `_` or `#` one can match multiple keys.
-This removes the need to explicitly define specification for keys.
+Assume we have a REST application we want to write a specification for.
+This REST application has to use multiple services.
 
-To describe the use case where `globbing` expressions make sense, we will use a REST application.
-This REST application uses multiple services which the user is not aware of at time of configuration.
-It could use a database, message broker, etc. but this is not yet known.
-A solution to this is `globbing` i.e. using wildcards in key names.
+If we have `database`, `broker` and `mailer` services we could use the specification:
+
+```ni
+[service/database/port]
+meta:/require = true
+meta:/default = 8080
+meta:/description = "The port of the database service."
+
+[service/database/url]
+meta:/require = true
+meta:/description = "The url of the database service."
+
+[service/broker/port]
+meta:/require = true
+meta:/default = 8080
+meta:/description = "The port of the broker service."
+
+[service/broker/url]
+meta:/require = true
+meta:/description = "The url of the broker service."
+
+[service/mailer/port]
+meta:/require = true
+meta:/default = 8080
+meta:/description = "The port of the mailer service."
+
+[service/mailer/url]
+meta:/require = true
+meta:/description = "The url of the mailer service."
+```
+
+This is already very repetitive and error prone. 
+But writing the specification becomes completely impossible when the set of services is not known in advance. 
+To solve this we need some way to allow for dynamic/unknown parts in the keyname.
+
+`Globbing` expressions is the solution to this problem. 
+By providing certain wildcard characters e.g. `_` or `#` one can match multiple keys.
+This removes the need to explicitly define specification for keys in our example for every service. 
+
 The thing to be aware of is to define configuration keys which are necessary for every service.
 In this example we use `port` and `url`.
 
@@ -50,8 +82,15 @@ The `_` would match every `service` we are using in our application e.g. `databa
 
 But the current implementation has some undefined behaivour when using `globbing`.
 
-This specification uses `_` but we also have `#` as wildcard in `globbing`. 
-The problem we face here is overlapping specifications.
+This specification uses `_` as wildcard character.
+Unfortunately `_` has a shortcoming, as we do not know the key names in advance.
+This prevents us from creating those keys in namespace they do not exist or copying metadata in advance.
+
+But besides `_` we also have `#` as wildcard character.
+This wildcard character is used with arrays.
+A fixed array size can be defined in the specification, which overcomes the shortcoming of `_`.
+
+The problem we face with those wildcard characters is overlapping specifications.
 A example for this is if we used `[service/#/port]` and `[service/shtg/port]` or `[service/_/port]` in the same specification.
 We don't know what metadata should be copied to other namespaces in this case.
 Both `[service/#/port]` and `[service/_/port]` existing produces conflicting specification.
@@ -59,6 +98,45 @@ Both `[service/#/port]` and `[service/_/port]` existing produces conflicting spe
 For example `[service/#/port]` having as type `meta:/type = string`.
 `[service/shtg/port]` having `meta:/type = unsigned_short`.
 The spec plugin can not know which of those specification to use.
+
+To address `#` again, it has a shortcoming when compared to `_`.
+
+Lets take a look at this example:
+
+```ni
+[service/#/port]
+meta:/require = true
+meta:/default = 8080
+meta:/description = "The port of the service."
+
+[service/#/url]
+meta:/require = true
+meta:/description = "The url of the service."
+```
+
+We completely loose the possibility to have a meaningful name for our key here.
+`service/database/port` is more expressive to a user than `service/#0/port`.
+
+We can overcome this problem by having a specification like this:
+
+```ni
+[service/#/port]
+meta:/require = true
+meta:/default = 8080
+meta:/description = "The port of the service."
+
+[service/#/url]
+meta:/require = true
+meta:/description = "The url of the service."
+
+[service/#/name]
+meta:/default = ""
+meta:/description = "The name of the service."
+```
+
+But this adds another key to give us a meaningful name for a given service.
+
+This is why we want to have the wildcard character `_` working correctly.
 
 Besides this problem the `spec` plugin also creates default values.
 A plugin responsible for copying metadata to other namespaces should not create new keys.
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -11,28 +11,49 @@ This is when `globbing` expressions in key names comes into place. By providing 
 characters e.g. `_` or `#` one can match and specify a number of keys at once without the need
 to explicitly state the key in every specification.
 
-To describe the use case where `globbing` expressions make sense, we will use an application
-based on microservice architecture as sample.
+To describe the use case where `globbing` expressions make sense, we will use a REST application.
+This REST application uses multiple services which the user is not aware of at time of configuration.
+It could use a database, message broker, etc. but this is not yet known.
+A solution to this is `globbing` i.o. using wildcards in key names.
+The thing to be aware of is to define configuration keys which are necessary for every service.
+In this example we use `port` and `url`.
 
-Imagine we have microservices and each of them communicates to a database to persist data. We
-will need some configuration so the applications can connect to the database e.g. `port`. Lets
-say we are specifying our configuration by `database/<DIALECT>/...`, where `...` is any name
-for a key. If we have multiple dialects e.g. a microservice uses more than one database e.g. `mysql`, `mssql`, etc. it is inconvenient to copy and paste the configuration for every 
-dialect we use.
+The key starts with `/sw/org/app-name/#0/development`.
+`sw` defines that this is a configuration for a software.
+`org` is the organization name.
+`app-name` is the application name we are writing the specification for.
+`#0` is the configuration version.
+`development` is the profile (it could also be `production`) 
 
-By using `globbing` expression we can overcome this problem.
-
-The key starts with `/sw/org/app-name`.
 ```ni
-[database/_/port]
+[service/_/port]
 meta:/require = true
-meta:/description = "The database port."
+meta:/default = 8080
+meta:/description = "The port of the service."
+
+[service/_/url]
+meta:/require = true
+meta:/description = "The url of the service."
 ```
-The `_` would match every `dialect` we are using in our application e.g. `postgresql`, `mssql`, etc.
+The `_` would match every `service` we are using in our application e.g. `database`, `message broker`, etc.
+
+### Problem 1
+
+This specification uses `_` but we also have `#` as wildcard in `globbing`. 
+The problem we face here is overlapping specifications.
+A example for this is if we used `[service/#/port]` and `[service/_/port]` in the same specification.
+This is currently undefined behaivour in the spec plugin.
+
+### Problem 2
+
+Errors and warnings are strictly defined in elektra.
+The spec plugin is currently not 
+
+### Problem 3
+
+Default values on wildcard globbing is undefined.
 
 ## Constraints
-
-TODO
 
 ## Assumptions
 
@@ -40,10 +61,10 @@ TODO
 
 ## Decision
 
-- no defaults for `sw/_/key` specifications (default will not work for `ksLookup(/sw/sthg/key)`)
-- plugins are not allowed to create keys (may change in future; depends on plugin positions)
-
-The spec plugin should yield errors when it detects such situations.
+- no defaults for keys ending with globbing keys `sw/_` (**Problem 3**)
+- no defaults for `sw/_/keys` (default will not work for `ksLookup(/sw/sthg/key)`)
+- The spec plugin should yield should yield errors to user on overlapping keys (**Problem 1**)
+- The spec plugin should yield errors and warnings when it detects such situations (**Problem 2**)
 
 ## Rationale
 
--- a/doc/decisions/1_in_discussion/spec_expressiveness.md
+++ b/doc/decisions/1_in_discussion/spec_expressiveness.md
@@ -32,20 +32,23 @@ meta:/description = "The url of the service."
 ```
 The `_` would match every `service` we are using in our application e.g. `database`, `message broker`, etc.
 
-### Problem 1
+But the current implementation has some undefined behaivour when using `globbing`.
 
 This specification uses `_` but we also have `#` as wildcard in `globbing`. 
 The problem we face here is overlapping specifications.
 A example for this is if we used `[service/#/port]` and `[service/shtg/port]` or `[service/_/port]` in the same specification.
-This is currently undefined behaivour in the spec plugin.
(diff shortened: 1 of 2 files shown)


tmakar:
This PR is currently not introducing any new changes. Before starting to make any decisions, I want to clarify things first.

tmakar:
> implement a sane spec plugin.

What do you define as `sane`?

tmakar:
The current implementation states:
> If a spec key has the metakey default set and the key does not exist in other namespaces, we create a special (cascading) key that will be found by ksLookup. This key has the default value as its value. We also copy over metadata as always.

The suggested decision states:
> plugins are not allowed to create keys (may change in future; depends on plugin positions)

So, the creation of cascading keys should not be done in the new implementation if I understand correctly?

> depends on plugin positions

What is a plugin position?

tmakar:
@markus2330 @kodebach Maybe you can answer this question. Thank you already for your time!!

kodebach:
> What is `sthg` here?

"sthg" is an abbreviation of "something", no idea why it was abbreviated however, or why we didn't use "foo" or other common placeholders.

> As far as I have understood, `_` has the same meaning as `*`

Not quite. `#` is for arrays, `_` for everything else and `*` covers both (even mixed). Generally, we recommend using `#` or `_`, to avoid mixing array and non-array children.

Some examples:

```
# matched by foo/#
foo/#0
foo/#1
foo/#2

# matched by foo/_
foo/abc
foo/def
foo/a#2

# matched by foo/*
foo/abc
foo/def
foo/a#2
foo/#0
foo/#1
foo/#2
```

> The current implementation states:

Like much of the documentation for `spec`, this is outdated. We actually create a key in the `default:/` namespace.


---

This entire decision seems quite badly written and very very incomplete. It would probably best, if you just ignore it and start from scratch.

I'm not sure how big the scope of possible changes is here. Maybe you come up with a much better system for doing specifications. Who knows...

However, what the current `spec` plugin was meant to do, but isn't quite able to achieve is this:

- A specification is written my adding `spec:/` keys with metadata (`meta:/` keys)
- A specification can contain some special keys, to describe multiple keys at once. A key with a `#` part is used to describe the entries of an array. A key with a `_` part is used to describe other children, you can think of it as describing the entries in a map/dictionary.
- During `kdbGet` the `spec` plugin copies the metadata from the `spec:/` keys to all matching keys in other namespaces. It also creates `default:/` keys where no matching key exists, but `meta:/default` was set. This is done so that other plugins and the application just need to look at the metadata of a key to know it's specification.
- During `kdbSet` the `spec` plugin removes the metadata it copied again, so we don't end up storing it (it should only be in the specification).
- Additionally, the `spec` plugin does some checks itself. This includes `meta:/require` for required keys, but also conflicting metadata and some validation for arrays.

Most of this works, but there are a few issues:

First of all, I'm not 100% sure removing metadata during `kdbSet` works correctly. But that shouldn't be too hard to fix.

Second, the default configuration of `spec` is pretty stupid. It just logs validation problems in a place that nobody checks. It should properly emit warnings and errors, if a validation it does fails.

Most importantly, while both `#` and `_` work well for existing keys, there a are some intricacies if no matching key exists and `meta:/default` is set. With `#` we just check for `meta:/array` and create an array of the appropriate size (all with default values). With `_` we can't do anything, because we don't know what keys should exist. You may ask, why this is a problem at all, but the answer is simple with an example. With this spec

```ini
[foo/_]
default = 7

[bar/#]
default = 8
```

a user may expect that `kdb get /.../foo/x` would return `7` and `kdb get /.../bar/#0` would return `8`. But that's not the case, because `spec` does not what keys the user will lookup. It only knows which keys already exist.

For arrays the explanation is that the array is empty and therefore `bar/#0` doesn't exist. If you call e.g. `kdb meta-set user:/.../bar array #2` then `kdb get /.../bar/#0` will indeed return `8` (*). But for `_` there is no equivalent here.

Things get even more complicated when it comes to more complex specifications:

```ini
[foo/_/abc]
default = 6

[foo/_/def]
default = 7

[bar/#/abc]
default = 8

[bar/#/def]
default = 9
```

For `#` things still work the same. You need to set `meta:/array` on `bar` to have keys generated. For `_` there is still no solution.

I did propose a possible solution in https://github.com/ElektraInitiative/libelektra/issues/3954, but there are certainly other ways to do it. It's been awhile so I can't really tell you what exactly I wrote back then, without reading it myself.

(*) this only works if the storage plugin used for `user:/.../bar/#0` is able to store metadata. Many plugins can't do that and the `kdb meta-set` will just do nothing.

---

PS. You may want to look at the highlevel API with it's codegenerator (docs in `doc/tutorials/highlevel.md` and tests in `tests/shell/gen/highlevel`). It uses the specification stuff quite heavily and basically how we stumbled upon all the problems with `spec`. There are also some additional examples in `example/codegen`. Also in my bachelor's thesis I used all this to elektrify LCDproc. The results of which can be found in https://github.com/kodebach/lcdproc. It has some rather complex specifications and I had to use various tricks to make things work. I heavily used arrays and references instead of using `_`.

tmakar:
> 

Thank you for the very clear explanation!!!!

tmakar:
> A specification is written my adding spec:/ keys with metadata (meta:/ keys)?

You mean this, right (achieved by using `meta-set` command)? If yes, I'm little bit confused, because you wrote `However, what the current spec plugin was meant to do, but isn't quite able to achieve is this:`. Maybe I'm also confusing things.

```ni
# [server/port]
#  meta:/check/port =
#  meta:/type = unsigned_short
#  meta:/example = 8080
#  meta:/description = port of the REST server that runs the application
#  meta:/default = 8080
```
(Taken from https://www.libelektra.org/tutorials/writing-a-specification-for-your-configuration).

tmakar:
> Most importantly, while both # and _ work well for existing keys, there a are some intricacies if no matching key exists and meta:/default is set.

But if the key does not exist and the `spec:/` does not require the key to exist, is the `meta:/default` still applied?

tmakar:
A question regarding the spec language. Are those meta keys (`check`, `type`, `example`, `description`, `default` and `required)` the only ones that exist?

```ni
# meta:/check/port =
#  meta:/type = unsigned_short
#  meta:/example = 8080
#  meta:/description = port of the REST server that runs the application
#  meta:/default
```

kodebach:
> You mean this, right (achieved by using `meta-set` command)? [...]

That is one option. You can also `kdb import` the snippet you posted with a `spec:/` parent key.

>  If yes, I'm little bit confused, because you wrote

What I meant is that of the following 5 points, the `spec` plugin doesn't do everything perfectly as expected.

> But if the key does not exist and the `spec:/` does not require the key to exist (has no `meta:/required` set to true), is the `meta:/default` still applied?

`meta:/default` and `meta:/require` are mutually exclusive. If there is a default value the key cannot be required. If it is required, there cannot be a default **and** the key must be set in at least one namespace which can be found during a cascading lookup (`system:/`, `user:/`, `dir:/`, `proc:/`).

> A question regarding the spec language.

The current "spec language" (if you want to call it that) is `meta:/` keys on `spec:/` keys. The common INI-like format with the `[keyname]` section is what the `ni` plugin uses. The `meta:/` keys that can be used for specification should all be listed in `doc/METADATA.ini` together with the plugin that is responsible for handling the metakey. Similarly, the `README.md` of every plugin should list the metakeys it handles.

tmakar:
> > You mean this, right (achieved by using `meta-set` command)? [...]
> 
> That is one option. You can also `kdb import` the snippet you posted with a `spec:/` parent key.
> 
> > If yes, I'm little bit confused, because you wrote
> 
> What I meant is that of the following 5 points, the `spec` plugin doesn't do everything perfectly as expected.
> 
> > But if the key does not exist and the `spec:/` does not require the key to exist (has no `meta:/required` set to true), is the `meta:/default` still applied?
> 
> `meta:/default` and `meta:/require` are mutually exclusive. If there is a default value the key cannot be required. If it is required, there cannot be a default **and** the key must be set in at least one namespace which can be found during a cascading lookup (`system:/`, `user:/`, `dir:/`, `proc:/`).
> 
> > A question regarding the spec language.
> 
> The current "spec language" (if you want to call it that) is `meta:/` keys on `spec:/` keys. The common INI-like format with the `[keyname]` section is what the `ni` plugin uses. The `meta:/` keys that can be used for specification should all be listed in `doc/METADATA.ini` together with the plugin that is responsible for handling the metakey. Similarly, the `README.md` of every plugin should list the metakeys it handles.

Thank you!

tmakar:
The high-level API doc uses a sample for the specification where `meta:/` is omitted e.g.
```ni
[mydouble]
type = double
```
whereas the Writing a specification for your configuration uses `meta:/type = double`. 

Is this on purpose - which I would expect?

kodebach:
> Is this on purpose - which I would expect?

In an earlier version of Elektra (now pretty long ago), metakeys didn't use the `meta:/` prefix. Nowadays the `meta:/` prefix should be used everywhere in source code.

In spec and config files like the example you posted, it depends on the format and the storage plugin that reads that format. Here the plugin in question is `ni`. It supports metakeys both with and without the `meta:/` prefix, because it is already clear from the other syntax that you want a metakey. So you can omit the `meta:/` prefix for brevity, but most newer documentation does use `meta:/` for clarity and to avoid confusing newcomers.

tmakar:
> In an earlier version of Elektra (now pretty long ago), metakeys didn't use the `meta:/` prefix. Nowadays the `meta:/` prefix should be used everywhere in source code.

Ok, I see, thanks!

tmakar:
@markus2330 @kodebach 

I have now written done some description for the actual spec plugin and "use case" for globbing. If you find some time, take a look 😉 

> NOTE: This is how I understand what the spec plugin is for and what globbing could solve (currently). If something is not how I imagined it to be than please correct me.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,12 +2,38 @@
...
+
+By using `globbing` expression we can overcome this problem.
+
+The key starts with `/sw/org/app-name`.
+```ni
+[database/_/port]
+meta:/require = true
+meta:/description = "The database port."
+```
+The `_` would match every `dialect` we are using in our application e.g. `postgresql`, `mssql`, etc.
```
kodebach, line 31: The example is a bit odd IMO, because it only makes sense, if one instance of the application uses multiple databases at the same time. If it just needs a single database, but lets the user choose which database the want to use, the specification could just be:

```ni
[database/port]
meta:/require = true
meta:/description = "The database port."

[database/dialect]
meta:/require = true
meta:/description = "The database dialect."
```

Maybe a better example, would be a service that acts as a reverse proxy:

```ni
[proxy/_/url]
meta:/require = true

[proxy/domain]
meta:/require = true
```

So this config

```
proxy/domain = example.com

proxy/bing/url = bing.com
proxy/google/url = google.com
```

would proxy requests for `bing.example.com` to `bing.com` and requests for `google.example.com` to `google.com`.

Or another example would be a mail client, which supports checking multiple accounts:

```ni
[account/#/email]
meta:/require = true

[account/#/password]
meta:/require = true
```

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,12 +2,38 @@
...
+(`spec` namespace) to all other namespaces. By providing certain metadata e.g. 
+`type` or `description`, it can be assured that keys adhere to a given specification. 
+
+For some applications there are keys we don't want to define for every application explicitly.
+This is when `globbing` expressions in key names comes into place. By providing certain wildcard
+characters e.g. `_` or `#` one can match and specify a number of keys at once without the need
+to explicitly state the key in every specification.
+
+To describe the use case where `globbing` expressions make sense, we will use an application
+based on microservice architecture as sample.
```
markus2330, line 15: Please describe use-cases separately (in another PR). For web-UI there is already doc/usecases/elektra_web/UC_key_validation.md

We also need use-cases directly for the spec lang, e.g., for `#` and `_` (we probably don't need `*`).

kodebach (changes requested):
Looks pretty good, seems you understood the core function. I have some minor suggests.

Also two general notes:
1. The `spec_expressiveness.md` file should be moved to `doc/decisions/0_drafts`. To conform the rules for decisions. Maybe you can also come up with a better name, once you defined the concrete problem.
2. The text you wrote so far doesn't really explain a problem. I would move it to a file in `doc/dev`. In `spec_expressiveness.md` I would assume people already know how specifications work and just explain the problem we have.

Neither of these has to be done immediately, but they should be done at some point.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,12 +2,38 @@
 
 ## Problem
 
-Currently, you can easily come into wrong assumptions that something would work in the specification.
-
-We need to find minimal requirements to implement a sane spec plugin.
+The current implementation of the spec plugin is copying metadata from the specification
+(`spec` namespace) to all other namespaces. By providing certain metadata e.g.
```
markus2330, line 6: Please use line breaks as described in doc/contrib/documentation.md

markus2330 (changes requested):
Looks good, you are already diving into the topic!

tmakar:
@markus2330 @kodebach 

I have defined the problem and my decision. Please review when you find time.

@markus2330 I have also created a use case PR for `globbing` (example with `_` and `#`).

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,9 +2,56 @@
...
+
+This specification uses `_` but we also have `#` as wildcard in `globbing`. 
+The problem we face here is overlapping specifications.
+A example for this is if we used `[service/#/port]` and `[service/_/port]` in the same specification.
+This is currently undefined behaivour in the spec plugin.
+
+### Problem 2
+
+Errors and warnings are strictly defined in elektra.
+The spec plugin is currently not
```
kodebach, line 50: Incomplete sentence

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,9 +2,56 @@
...
+This is currently undefined behaivour in the spec plugin.
+
+### Problem 2
+
+Errors and warnings are strictly defined in elektra.
+The spec plugin is currently not 
+
+### Problem 3
+
+Default values on wildcard globbing is undefined.
```
kodebach, line 54: What does that mean?

kodebach (changes requested):
The "Problem" section needs more work. It is currently quite repetitive, a bit out of order (e.g. mentions globbing, later introduces globbing as something new again) and still doesn't really explain the problem.

IMO you have two options here:
1. Assume the current globbing solution as a fixed basis set in stone:
    In this case, the "Problem" section should first explain (briefly), how the `_` and `#` wildcards for globbing work in the current `spec` plugin. Then it should describe the problems this approach has and give some example for each of those problems.
2. Make the decision more general; look into totally different solutions:
    If you want to make the decision more general, the "Problem" section should introduce what we want to achieve with the `spec` plugin in general (i.e. having a specification in `spec:/` that gets used by all plugins and the application for keys in all namespaces). The main problem would simply be "How to best achieve these goals of `spec`?". The current globbing approach would then just be one of the "Considered Alternatives" (i.e. possible solutions).

I think we should also remove the contents of the "Decision" section for now to follow the [Decision Process](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/3_decided/decision_process.md).

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,9 +2,56 @@
...
+meta:/description = "The url of the service."
+```
+The `_` would match every `service` we are using in our application e.g. `database`, `message broker`, etc.
+
+### Problem 1
+
+This specification uses `_` but we also have `#` as wildcard in `globbing`. 
+The problem we face here is overlapping specifications.
+A example for this is if we used `[service/#/port]` and `[service/_/port]` in the same specification.
+This is currently undefined behaivour in the spec plugin.
```
markus2330, line 45: There can also be `service/sthg/port`.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,9 +2,56 @@
...
+
+### Problem 2
+
+Errors and warnings are strictly defined in elektra.
```
markus2330, line 49:
```suggestion
Errors and warnings are strictly defined in Elektra.
```

markus2330 (changes requested):
There should be only one problem per decision. Please split up in several decisions. Start with clarifications what exactly the problem is before writing solutions.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -14,10 +61,10 @@ We need to find minimal requirements to implement a sane spec plugin.
...
 ## Decision
 
-- no defaults for `sw/_/key` specifications (default will not work for `ksLookup(/sw/sthg/key)`)
-- plugins are not allowed to create keys (may change in future; depends on plugin positions)
-
-The spec plugin should yield errors when it detects such situations.
+- no defaults for keys ending with globbing keys `sw/_` (**Problem 3**)
+- no defaults for `sw/_/keys` (default will not work for `ksLookup(/sw/sthg/key)`)
+- The spec plugin should yield should yield errors to user on overlapping keys (**Problem 1**)
+- The spec plugin should yield errors and warnings when it detects such situations (**Problem 2**)
```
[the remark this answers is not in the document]
tmakar: Yes, exactly.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,9 +2,56 @@
...
+A solution to this is `globbing` i.o. using wildcards in key names.
+The thing to be aware of is to define configuration keys which are necessary for every service.
+In this example we use `port` and `url`.
+
+The key starts with `/sw/org/app-name/#0/development`.
+`sw` defines that this is a configuration for a software.
+`org` is the organization name.
+`app-name` is the application name we are writing the specification for.
+`#0` is the configuration version.
+`development` is the profile (it could also be `production`)
```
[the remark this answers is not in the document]
tmakar: Ok, I understand your point. It thought it would be good to add in case anyone comes across and does not remember the convention anymore.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,9 +2,56 @@
...
+Sometimes there are keys with similar or identical specifications.
+We want to define those in a single place and not have to manually copy metadata.
+`Globbing` expressions in key names comes into place at that time. 
+By providing certain wildcard characters e.g. `_` or `#` one can match multiple keys.
+This removes the need to explicitly define specification for keys.
+
+To describe the use case where `globbing` expressions make sense, we will use a REST application.
+This REST application uses multiple services which the user is not aware of at time of configuration.
+It could use a database, message broker, etc. but this is not yet known.
+A solution to this is `globbing` i.o. using wildcards in key names.
```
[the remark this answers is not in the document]
tmakar: Done.

doc/decisions/1_in_discussion/spec_expressiveness.md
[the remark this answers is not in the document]
tmakar: Done.

tmakar:
> There should be only one problem per decision. Please split up in several decisions. Start with clarifications what exactly the problem is before writing solutions.

I thought it would make sense to keep those problems in one document as the title of this document states `Spec Expressiveness`.

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: Done.

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: yes, thanks.

tmakar: Done.

tmakar:
@markus2330 @kodebach I have decided to go with the `globbing` solution.

I have adapted the problem section. Hopefully it is a bit clearer now. 
@kodebach I also removed the decision section for now as you stated.

kodebach:
> > There should be only one problem per decision. Please split up in several decisions. Start with clarifications what exactly the problem is before writing solutions.
> 
> I thought it would make sense to keep those problems in one document as the title of this document states `Spec Expressiveness`.

I agree, this is just one problem. It just has multiple aspects. The core problem (depending on POV) is "How to make `_` globbing work?". I think the new "Problem" section already makes this more clear.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,61 @@
...
-Currently, you can easily come into wrong assumptions that something would work in the specification.
+The current implementation of the spec plugin is copying metadata to all other namespaces.
+Metadata e.g. `type` or `description` describes a certain key.
 
-We need to find minimal requirements to implement a sane spec plugin.
+Sometimes there are keys with similar or identical specifications.
+We want to define those in a single place and not have to manually copy metadata.
+`Globbing` expressions in key names comes into place at that time. 
+By providing certain wildcard characters e.g. `_` or `#` one can match multiple keys.
+This removes the need to explicitly define specification for keys.
```
kodebach, line 12: Weird order. I'd start with the "Why?" then explain the "How?". So first give an example of specifications and then explain that the current `spec` achieves this by copying metadata.

You should probably also start this without mentioning globbing. A specification without globbing still makes sense and is easier to understand. In your example you could for example have

```ni
[public/ip]
meta:/default = 127.0.0.1
meta:/description = "The IP via which this application is accessible."

[public/port]
meta:/default = 8080
meta:/description = "The port on which this application is accessible."
```

tmakar:
@markus2330 @kodebach 

I also have a question regarding conflicts in the current implementation of the `spec` plugin. As I have understood this is how the spec plugin displays validation errors (by creating those conflict keys). Is this used anywhere else? In Elektra core or any other plugin?

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,61 @@
...
+[service/_/port]
+meta:/require = true
+meta:/default = 8080
+meta:/description = "The port of the service."
+
+[service/_/url]
+meta:/require = true
+meta:/description = "The url of the service."
+```
+The `_` would match every `service` we are using in our application e.g. `database`, `message broker`, etc.
```
kodebach, line 33: If I want to explain the need for globbing IMO it is best to show that it is worse (or impossible) without globbing. In your example you could start like this:

````markdown
If we have `database`, `broker` and `mailer` services we could use the specification:

```ni
[service/database/port]
meta:/require = true
meta:/default = 8080
meta:/description = "The port of the database service."

[service/database/url]
meta:/require = true
meta:/description = "The url of the database service."

[service/broker/port]
meta:/require = true
meta:/default = 8080
meta:/description = "The port of the broker service."

[service/broker/url]
meta:/require = true
meta:/description = "The url of the broker service."

[service/mailer/port]
meta:/require = true
meta:/default = 8080
meta:/description = "The port of the mailer service."

[service/mailer/url]
meta:/require = true
meta:/description = "The url of the mailer service."
```

This is already very repetitive and error prone. But writing the specification becomes completely when the set of services is not known in advance. To solve this we need some way to allow for dynamic/unknown parts in the keyname. That is what we call Globbing.

[...]
````

File: doc/decisions/1_in_discussion/spec_expressiveness.md [hunk 13]
```diff
@@ -2,22 +2,61 @@
...
+
+But the current implementation has some undefined behaivour when using `globbing`.
+
+This specification uses `_` but we also have `#` as wildcard in `globbing`.
```
kodebach, line 37: I would first explain the issues with `_` alone, i.e. that we cannot know the keynames that will be used with `ksLookup`, so we cannot create keys or copy metadata in advance.

Then you could mention that `#` also exists, and that because it is restricted to arrays it does _not_ have this problem. We only need to know the array size to create keys and copy metadata. We can simply enforce that the array size is defined.

Only then would I mentioned that `#` doesn't solve all problems. Overlapping specification is still and issue, but that's easy to fix. I'd just mention that any solution must report errors for overlapping specifications, something the current `spec` fails to do.

Finally, you should also mention the disadvantages of using `#` over `_`. Take your example, but with `#`:

```ni
[service/#/port]
meta:/require = true
meta:/default = 8080
meta:/description = "The port of the service."

[service/#/url]
meta:/require = true
meta:/description = "The url of the service."
```

Here we loose the unique name for the service that was previously covered by the `_` part. Now we just get an index in the array. Still a unique identifier, but much harder to handle. Also it's not easily possible to give meaning to this index. Whereas before using `database` for `_` could indicate something to the application (and the user), you now only have something like `service/#0/port`. You can't really give meaning to `#0`.

That's why you'll often end up with things like:

```ni
[service/#/port]
meta:/require = true
meta:/default = 8080
meta:/description = "The port of the service."

[service/#/url]
meta:/require = true
meta:/description = "The url of the service."

[service/#/name]
meta:/default = ""
meta:/description = "The name of the service."
```

Now we have a name again. But that is still more annoying to use. You have to set more keys. The key `service/#0/url` still doesn't tell you much. You still have to look at `service/#0/name` to find out what the URL is for, etc.

That is why we would like to have a `_` that works at least somewhat decently.

> **Note**: Using `service/#/name` to define what kind of service this array entry is, also has an advantage over using the `_` part for this. The value of `service/#/name` can be restricted with a `meta:/check/enum`, while `_` is completely arbitrary. The tags idea in #3954 could address that.

kodebach (changes requested):
The problem section looks much better. But IMO the problem could still be explained a bit better. In particular, I think reordering some parts and adding more examples could improve things. I made some concrete suggestions.

kodebach:
> I also have a question regarding conflicts in the current implementation of the `spec` plugin. As I have understood this is how the spec plugin displays validation errors (by creating those conflict keys). Is this used anywhere else? In Elektra core or any other plugin?

AFAIK it is not used anywhere. It can be removed as far as I'm concerned. The plugin should simply add errors/warnings to the `parentKey`. If we need machine readable information anywhere, we should discuss it only when we have a concrete use case with clear requirements.

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: Thank you, this is much better.

tmakar:
> > I also have a question regarding conflicts in the current implementation of the `spec` plugin. As I have understood this is how the spec plugin displays validation errors (by creating those conflict keys). Is this used anywhere else? In Elektra core or any other plugin?
> 
> AFAIK it is not used anywhere. It can be removed as far as I'm concerned. The plugin should simply add errors/warnings to the `parentKey`. If we need machine readable information anywhere, we should discuss it only when we have a concrete use case with clear requirements.

Alright, thank you!!

tmakar:
@kodebach Thanks again for the great examples!

tmakar:
@kodebach @markus2330 

I also have a question regarding the `C interface`.

```
kdbOpen() calls elektraPluginOpen() of every plugin to let them do their initialisation.
kdbGet() requests elektraPluginGet() of every plugin in the queried backends to return a key set.
kdbSet() usually calls elektraPluginSet() of every plugin in the queried backends to store the configuration.
kdbSet() also calls elektraPluginError() for every plugin when an error happens. Because of elektraPluginError(), plugins are guaranteed to have their chance for necessary cleanups.
kdbClose() makes sure that plugins can finally free their own resources in elektraPluginClose()
```

The `spec` plugin uses:

* int elektraSpecGet (Plugin * handle, KeySet * ks, Key * parentKey);
* int elektraSpecCopy (Plugin * handle, KeySet * returned, Key * parentKey, bool isKdbGet);
* int elektraSpecRemove (Plugin * handle, KeySet * returned, Key * parentKey);

It is no-where specified how and who calls `elektraSpecCopy` and `elektraSpecCopy`. Do you know how this is done?

kodebach:
> It is nowhere specified how and who calls `elektraSpecCopy` and `elektraSpecRemove`. Do you know how this is done?

This is a recent change by @atmaxinger. We made this change, because `spec` is not actually a "normal" plugin. It is a so-called "hook plugin". Each of these hook plugins has a very specific purpose and is called at very specific times (differently to normal plugins), to allow them to do there work. For `spec` this is the case, because it must be called at the correct points to make other plugins work, i.e. the specification must be copied _before_ (some) other plugins are called.

For the functions of `spec`:

- `elektraSpecGet` does basically nothing. It only exists to provide the contract (`if (!elektraStrCmp`) at the start. Because of this contract, the `get` function is currently mandatory for all plugins. We may move this contract elsewhere at some point to make the `get` function optional like the others. Then the would remove `elektraSpecGet`.
- `elektraSpecCopy` should copy metadata from `spec:/` keys to other keys, create `default:/` keys, check for problems with the spec etc. Basically this is the main logic of `spec`. The `isKdbGet` parameters tells the function, whether we are currently in a `kdbGet` or not. This is because, during a `kdbGet` only warnings should be created, i.e. any error that is normally created should be replaced with a warning when `isKdbGet == true`.
- `elektraSpecRemove` should basically undo/remove everything that was created in `elektraSpecCopy`. We need this, because we want to keep the specification only in `spec:/` keys, we don't actually want to write e.g. the copied metadata to disk when `kdbSet` is called.

When exactly these functions are called shouldn't matter. As long as they do want the are supposed to do, things should work.

But for completeness sake: `elektraSpecCopy` is called in `kdbGet` and at the start of `kdbSet`. `elektraSpecRemove` is called later in `kdbSet`, before we write anything to disk. The extra copy in `kdbSet` is so that we can correctly do validation for e.g. `meta:/type` during `kdbSet` even, if the key was newly created after `kdbGet` was called.

https://github.com/ElektraInitiative/libelektra/blob/6f97507417ec5940dc6cccf6d14491d256a6cb68/src/libs/elektra/kdb.c#L1897
https://github.com/ElektraInitiative/libelektra/blob/6f97507417ec5940dc6cccf6d14491d256a6cb68/src/libs/elektra/kdb.c#L2386
https://github.com/ElektraInitiative/libelektra/blob/6f97507417ec5940dc6cccf6d14491d256a6cb68/src/libs/elektra/kdb.c#L2468

tmakar:
> > It is nowhere specified how and who calls `elektraSpecCopy` and `elektraSpecRemove`. Do you know how this is done?
> 
> This is a recent change by @atmaxinger. We made this change, because `spec` is not actually a "normal" plugin. It is a so-called "hook plugin". Each of these hook plugins has a very specific purpose and is called at very specific times (differently to normal plugins), to allow them to do there work. For `spec` this is the case, because it must be called at the correct points to make other plugins work, i.e. the specification must be copied _before_ (some) other plugins are called.
> 
> For the functions of `spec`:
> 
> * `elektraSpecGet` does basically nothing. It only exists to provide the contract (`if (!elektraStrCmp`) at the start. Because of this contract, the `get` function is currently mandatory for all plugins. We may move this contract elsewhere at some point to make the `get` function optional like the others. Then the would remove `elektraSpecGet`.
> * `elektraSpecCopy` should copy metadata from `spec:/` keys to other keys, create `default:/` keys, check for problems with the spec etc. Basically this is the main logic of `spec`. The `isKdbGet` parameters tells the function, whether we are currently in a `kdbGet` or not. This is because, during a `kdbGet` only warnings should be created, i.e. any error that is normally created should be replaced with a warning when `isKdbGet == true`.
> * `elektraSpecRemove` should basically undo/remove everything that was created in `elektraSpecCopy`. We need this, because we want to keep the specification only in `spec:/` keys, we don't actually want to write e.g. the copied metadata to disk when `kdbSet` is called.
> 
> When exactly these functions are called shouldn't matter. As long as they do want the are supposed to do, things should work.
> 
> But for completeness sake: `elektraSpecCopy` is called in `kdbGet` and at the start of `kdbSet`. `elektraSpecRemove` is called later in `kdbSet`, before we write anything to disk. The extra copy in `kdbSet` is so that we can correctly do validation for e.g. `meta:/type` during `kdbSet` even, if the key was newly created after `kdbGet` was called.
> 
> https://github.com/ElektraInitiative/libelektra/blob/6f97507417ec5940dc6cccf6d14491d256a6cb68/src/libs/elektra/kdb.c#L1897
> 
> 
> https://github.com/ElektraInitiative/libelektra/blob/6f97507417ec5940dc6cccf6d14491d256a6cb68/src/libs/elektra/kdb.c#L2386
> 
> 
> https://github.com/ElektraInitiative/libelektra/blob/6f97507417ec5940dc6cccf6d14491d256a6cb68/src/libs/elektra/kdb.c#L2468

Thank you!

markus2330:
> @markus2330 @kodebach I have decided to go with the globbing solution.

Decisions are about to find good decisions together. First you need to clarify what the problem is, and then we will see if globbing is a good solution.

> I also have a question regarding conflicts in the current implementation of the spec plugin. As I have understood this is how the spec plugin displays validation errors (by creating those conflict keys). Is this used anywhere else? In Elektra core or any other plugin?

There is no decision giving any support for that solution. So it shouldn't be used in the spec plugin.

> It is nowhere specified how and who calls elektraSpecCopy and elektraSpecRemove. Do you know how this is done?

@kodebach Can you please describe this somewhere in the documentation, and not only here were it will be lost soon?

The goal would be that on such questions we can point to documentation.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,155 @@
...
+
+[public/port]
+meta:/default = 8080
+meta:/description = "The port on which this application is accessible."
+```
+
+Here one can see `default` and `description` used as metadata. 
+
+The current implementation of the spec plugin is copying this metadata to all other namespaces.
+It also assures that if keys do not exist and have `default` metadata, it creates a cascading key.
```
markus2330, line 22:
```suggestion
It also assures that if keys do not exist and have `default` metadata, it creates a key in the default namespace.
```

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,155 @@
...
+
+Assume we have a REST application we want to write a specification for.
+This REST application has to use multiple services.
```
markus2330, line 25: What is this first example for?

```suggestion
```

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,155 @@
...
+[service/mailer/url]
+meta:/require = true
+meta:/description = "The url of the mailer service."
+```
+
+This is already very repetitive and error prone. 
+But writing the specification becomes completely impossible when the set of services is not known in advance. 
+To solve this we need some way to allow for dynamic/unknown parts in the keyname.
+
+`Globbing` expressions is the solution to this problem.
```
markus2330, line 62: The problem part should only contain the problem.

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,155 @@
...
+meta:/default = 8080
+meta:/description = "The port of the service."
+
+[service/_/url]
+meta:/require = true
+meta:/description = "The url of the service."
+```
+The `_` would match every `service` we are using in our application e.g. `database`, `message broker`, etc.
+
+But the current implementation has some undefined behaivour when using `globbing`.
```
markus2330, line 83: I think the actual problem starts here.

```suggestion
A naive  implementation has some undefined behavior when using `globbing`.
```

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,155 @@
...
+This prevents us from creating those keys in namespace they do not exist or copying metadata in advance.
+
+But besides `_` we also have `#` as wildcard character.
+This wildcard character is used with arrays.
+A fixed array size can be defined in the specification, which overcomes the shortcoming of `_`.
+
+The problem we face with those wildcard characters is overlapping specifications.
+A example for this is if we used `[service/#/port]` and `[service/shtg/port]` or `[service/_/port]` in the same specification.
+We don't know what metadata should be copied to other namespaces in this case.
+Both `[service/#/port]` and `[service/_/port]` existing produces conflicting specification.
```
markus2330, line 96: existing?

doc/decisions/1_in_discussion/spec_expressiveness.md
```diff
@@ -2,22 +2,155 @@
...
+
+For example `[service/#/port]` having as type `meta:/type = string`.
+`[service/shtg/port]` having `meta:/type = unsigned_short`.
+The spec plugin can not know which of those specification to use.
```
markus2330, line 100: Again: this is already solution.

Btw. Of course some behavior could be defined, so the spec plugin can know what to use.

markus2330 (changes requested):
As already said please separate the problems. Problems and solutions shouldn't be mixed.

markus2330:
@tmakar it probably makes sense that you first write the use cases.

tmakar:
> @tmakar it probably makes sense that you first write the use cases.

What use cases are you talking about? I created a globbing use case.

tmakar:
> As already said please separate the problems. Problems and solutions shouldn't be mixed.

For me this does not make any sense to split. This decision is about spec expressiveness and I as a user would expect all decisions on the `spec` plugin to be found here. IMO by splitting it up in multiple files we lose the benefit of finding documentation about something quickly. But lets discuss this in more detail in the meeting.

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: Thanks!

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: Thanks!

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: I started this decision to first explain how something works and then slowly going towards to actual problem so everyone understands why this problem exists.

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: I already have just written down the problems in a previous version. I was told to back it up with some example, which was a very good input IMO, as I think people reading this document immediately understand what this problem is about without the need of knowing to much details.

doc/decisions/1_in_discussion/spec_expressiveness.md
tmakar: > Btw. Of course some behavior could be defined, so the spec plugin can know what to use.

So you are saying that this is not a problem at all? Having both `[service/#/port]` and `[service/shtg/port]` with different `meta:/type` can be configured the way it works?

kodebach (changes requested):
I think there is now a bottleneck issue here.

Ideally, the decision file would just reference the (relevant) use cases about spec and point out the problem. Before you created use cases, it made sense to give examples in the decision. But now that's no longer needed. For example, you should reference the use case from #4812 and then point out that when using `_` we cannot properly handle `meta:/default` right now. That way the "Problem" section is much shorter and more focused.

The issue you now have is that the use cases aren't merged yet, so you can't properly reference them. I think the best way forward is to leave this PR for now and wait until the use case PRs are finished.

tmakar:
> I think there is now a bottleneck issue here.
> 
> Ideally, the decision file would just reference the (relevant) use cases about spec and point out the problem. Before you created use cases, it made sense to give examples in the decision. But now that's no longer needed. For example, you should reference the use case from #4812 and then point out that when using `_` we cannot properly handle `meta:/default` right now. That way the "Problem" section is much shorter and more focused.
> 
> The issue you now have is that the use cases aren't merged yet, so you can't properly reference them. I think the best way forward is to leave this PR for now and wait until the use case PRs are finished.

Agree, I will come back to this when all of the use case PRs are merged!

FYI: @markus2330

PR #4797, opened 2023-01-01
[use cases] Index based use cases for core

kodebach:
## Basics

      contains `_(my name)_`)
  - [ ] fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


doc/usecases/core/UC_keyset_lookup_direct.md
```diff
@@ -14,9 +14,9 @@
 - **Main success scenario:**
   - Caller asks Core to look up `Key` by non-cascading name in `KeySet`
   - Core searches for `Key` with same name in `KeySet`
-  - If a matching `Key` is found, Core returns a `Key *` to it.
-    The name of the `Key` will be read-only, otherwise it is modifiable.
-  - Otherwise, Core returns `NULL`
+  - If a matching `Key` is found, Core returns its index.
+  - Otherwise, Core returns a negative value, which indicates "not found".
+    Additionally, the negative value can be converted (with a well-defined formula) into the index at which the searched `Key` would be inserted.
```
Eiskasten, line 19: What is this "well-defined formula"? I assume it is something like `(returned value + 1) * (-1)`?

doc/usecases/core/UC_keyset_lookup_hierarchy.md
```diff
@@ -1,23 +1,23 @@
...
 - **Actors:** Core, Caller
-- **Brief:** Caller cuts (=looks up and removes) a `Key` hierarchy from an existing `KeySet`
+- **Brief:** Caller looks up a `Key` hierarchy from an existing `KeySet`
 
 ## Scenarios
 
 - **Precondition:**
   - [`KeySet` has been created](UC_keyset_create.md)
 - **Main success scenario:**
   - Caller asks Core to cut a `Key` hierarchy descendant from a given `Key` R from a `KeySet` KS.
```
Eiskasten, line 15:
```suggestion
  - Caller asks Core to lookup a `Key` hierarchy descendant from a given `Key` R from a `KeySet` KS.
```

Eiskasten (commented):
Thank you for this improvement. Very precisely written. Is there a specific reason why removing a single/multiple keys is a single use case while accessing is separated?

kodebach:
> Is there a specific reason why removing a single/multiple keys is a single use case while accessing is separated?

The different return values for "index access" (key vs keyset) make it hard to consider one the "main" and the other the "alternative" scenario, IMO. Probably, I should separate the removing use cases too.

Eiskasten:
> > Is there a specific reason why removing a single/multiple keys is a single use case while accessing is separated?
> 
> The different return values for "index access" (key vs keyset) make it hard to consider one the "main" and the other the "alternative" scenario, IMO. Probably, I should separate the removing use cases too.

I see. Maybe altering the removing use case in such a way that both functions return more than just a status code (e.g. `Key *` and `Key **`) would make separating them much more obvious.

doc/usecases/core/UC_keyset_index_access_many.md
```diff
@@ -0,0 +1,26 @@
...
+  - Core creates new `KeySet` KS1
+  - Core adds all `Key`s in KS with index `i >= f && i < l` to KS1.
+  - Core returns KS1
+- **Alternative scenario:** -
+- **Error scenario:**
+  - Caller requests `Key` at invalid index range (`f < 0 || f >= size || l < f || l > size`) from `KeySet`
+  - Core returns `NULL`
+- **Postcondition:**
+  - The initial `KeySet` KS MUST be unmodified.
+- **Non-functional Constraints:** -
```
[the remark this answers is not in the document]
kodebach: Good question. I'd say it should be `O(N)` where `N` is the initial number of keys in KS. Which would be the same for removing a single key, because in both cases, we at minimum need to `memmove` a part of `ks->array` in the `KeySet`.

That's actually (partly) the reason for introducing this API, because  currently you need to "remove single key" M-times to remove M keys, which give `O(M*N)`.

doc/usecases/core/UC_keyset_lookup_direct.md
kodebach: Yes, that's currently the plan, I just didn't want to specify it here, because for the use case it doesn't matter what the exact formula is. It's even a bit much to specify a negative value.

Maybe it would be better to say:

```markdown
- Otherwise, Core returns a value, which indicates "not found".
  Additionally, the index at which the searched `Key` would be inserted can be determined by Caller without any additional calls to Core.
```

In theory the API could look like this:

```c
typedef struct {
    size_t index : (sizeof(size_t)*CHAR_BIT-1);
    bool found : 1;
} LookupResult;

LookupResult ksLookup(KeySet * ks, Key * k);
```

Or this:

```c
// returns true, if found and false if not found
// *index is set to index of key if found, or index where key would be inserted if not found
bool ksLookup (KeySet * ks, Key * k, size_t * index); 
```

doc/usecases/core/UC_keyset_index_access_many.md
```diff
@@ -0,0 +1,26 @@
...
+- **Actors:** Core, Caller
+- **Brief:** Caller accesses contents of `KeySet` by index
+
+## Scenarios
+
+- **Precondition:**
+  - [`KeySet` has been created](UC_keyset_create.md).
+  - [`Key`(s) have been inserted into `KeySet`](UC_keyset_insert.md).
+- **Main success scenario:**
+  - Caller requests `Key`s in valid index range (`[f, l)` with `0 <= f < size && f <= l <= size`) from `KeySet` KS
```
markus2330, line 16: This description doesn't motivate why the caller needs this. Probably an example would help.

doc/usecases/core/UC_keyset_index_access_many.md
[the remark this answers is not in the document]
markus2330: Then please add this info. I am wondering why someone even wants to request keys in an index range (as core feature).

doc/usecases/core/UC_keyset_lookup_direct.md
markus2330: If we already know that we want negative values, it is easier to write it down. No need to keep a UC abstract if within Elektra it needs to be specified anyway. If an alternative implementation could do it differently, it would be a different story.

doc/usecases/core/UC_keyset_lookup_hierarchy.md
```diff
@@ -1,23 +1,23 @@
...
 - **Precondition:**
   - [`KeySet` has been created](UC_keyset_create.md)
 - **Main success scenario:**
-  - Caller asks Core to cut a `Key` hierarchy descendant from a given `Key` R from a `KeySet` KS.
-  - Core searches for the first `Key` in KS that is part of the desired hierarchy, i.e., the first `Key` in KS that is a descendant of R according to [`Key` hierarchy comparison](UC_keyname_hierarchy.md).
-  - If not found, Core returns an empty KS
-  - If found, Core searches for the last `Key` that is not part of the desired hierarchy, i.e., Core tries all `Key`s in KS in [keyname order](UC_keyname_ordering.md) until it finds a `Key` that is not a descendant of R.
-  - Core removes all `Key`s of the desired hierarchy from KS and puts them into a new `KeySet` KS1
-  - Core returns KS1.
+  - Caller asks Core to lookup a `Key` hierarchy descendant from a given `Key` R from a `KeySet` KS.
```
markus2330, line 15: Please describe what an "hierarchy descendant" is (probably in glossary). Please motivate why a caller needs this (e.g. example).

doc/usecases/core/UC_keyset_lookup_hierarchy.md
```diff
@@ -1,23 +1,23 @@
...
+  - Core searches for the first `Key` in KS that is part of the desired hierarchy, i.e., the first `Key` F in KS that is a descendant of R according to [`Key` hierarchy comparison](UC_keyname_hierarchy.md).
+  - If not found, Core returns an out-of-range index range.
+  - If found, Core searches for the first `Key` L that is not part of the desired hierarchy.
+    This is equivalent to Core trying all `Key`s in KS after F in [keyname order](UC_keyname_ordering.md) until it finds a `Key` that is not a descendant of R.
+  - Core returns an index range `[f, l)` (start inclusive, end exclusive), where `f` is the index of `Key` F and `l` the index of L.
```
markus2330, line 20: Please describe why we get an index range if we only wanted the "hierarchy descendant".

doc/usecases/core/UC_keyset_remove.md
```diff
@@ -12,13 +12,17 @@
...
-  - If a matching `Key` is found, Core removes it from `KeySet` returns a `Key *` to it.
-    The name of the `Key` will be read-only, otherwise it is modifiable.
-  - Otherwise, Core returns `NULL`
-- **Alternative scenario:** -
-- **Error scenario:** -
+  - Caller requests to remove `Key` by index from `KeySet`
+  - Index is in-range for `KeySet`
+  - Core removes `Key` at index and resizes `KeySet`.
+- **Alternative scenario:**
+  - Caller requests to remove a range of `Key`s by index range from `KeySet`
```
markus2330, line 19: This probably can be main scenario, as the range simply would be 1. If there is a convenience function for that, is imho not really relevant for the use case. But again, there should be a motivation why a user wants this. (E.g. cut out a hierarchy).

markus2330 (changes requested):
Great job! Especially great that you also updated older use cases, too.

Main problem (that earlier use cases also already had) is that the use cases sometimes don't really motivate what exactly a user actually wants to achieve.

Examples might help but higher-level UC that include these lower-level UC might be a much better solution.

doc/usecases/core/UC_keyset_index_access_many.md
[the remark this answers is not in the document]
kodebach: As stated in #4805 whether the index range functions are in `libelektra-core` or `libelektra-extra` is debatable. However, there are two clear use cases (which I will add to the file):
1. Getting all the `Key`s in a key hierarchy, i.e. what `ksCut`/`ksBelow` does right now. We need this index access function, because the key hierarchy is only returned as an index range (for consistency with lookup and flexibility).
2. Array manipulation. If you have an array `foo/#` and want to get, move or delete a section (e.g. `#4` until `#7`), then the index range stuff allows doing this. With just the hierarchy based APIs (like `ksCut`) and array is always one atomic block.

doc/usecases/core/UC_keyset_lookup_direct.md
kodebach: > If we already know that we want negative values

Do we? I actually think the `LookupResult` from above is an equally good solution. [The struct uses 8 bytes](https://godbolt.org/z/ahaWxGvcK) just like a `size_t`/`ssize_t` would. It might even be better, because it's much clearer how to interpret the result.

> No need to keep a UC abstract if within Elektra it needs to be specified anyway.

IMO there has to be a balance. Making it unnecessarily abstract and harder to read is bad. But making it overly specific and thereby limiting what can be done on the decision/implementation level without changing the use case is also bad.

doc/usecases/core/UC_keyset_index_access_many.md
kodebach: See https://github.com/ElektraInitiative/libelektra/pull/4797#discussion_r1064486145

doc/usecases/core/UC_keyset_lookup_hierarchy.md
kodebach: There my be a comma missing here... It is supposed to read

```suggestion
  - Caller asks Core to lookup a `Key` hierarchy, descendant from a given `Key` R, from a `KeySet` KS.
```

So "descendant from a given `Key` R" is more or less what defines a hierarchy. I'll add it to the glossary nonetheless.

doc/usecases/core/UC_keyset_lookup_hierarchy.md
kodebach: Like above, we want the "hierarchy, which is descendant ...".

Returning an index range is
1. Consistent with lookup, which also only returns an index.
2. A lot more flexible than returning a `KeySet *`. The `KeySet *` needs to be allocated and a whole array of pointers has to be created (*). That would be totally unnecessary, if all the caller wants to do is iterate over the `Key`s in the hierarchy, e.g. to print them. It would be even worse with just a `ksCut` and not a `ksBelow`. Then you have to reinsert the keys you wanted to iterate too.

(*) The array allocation part could be solved by adding an `size_t offset` field to `KeySet` and letting `KeySet`s share arrays with different offsets. But that makes things more complicated than they need to be.

doc/usecases/core/UC_keyset_remove.md
kodebach: That's true, I'll swap main and alternative scenario.

But do we actually need a detailed motivation here? Removing range of elements in a sorted data structure seems pretty basic too me. For example, the entire `KeySet` might be an array (e.g. extracted from elsewhere) and you want to remove a section of the array.

doc/usecases/core/UC_keyset_lookup_direct.md
markus2330: Ok, then let us keep it open in the use case.

doc/usecases/core/UC_keyset_remove.md
markus2330: If we were happy without it up to now, there should be definitely some motivation now. The best motivation would be something from LCDproc or a plugin (something practical in the real world); together with a non-functional requirement.

kodebach: The thing is "happy until now" is hard to define. You can always call the function "remove Key at index X" repeatedly in a loop to remove N keys. So maybe somebody already encountered a practical use case. But OTTOMH I don't know a motiviation for general ranges.

But for hierarchies, we just need to look at were `ksCut` is used. All of those become a use case for "remove index range", with the index range returned from "find hierarchy". That the range doesn't have to be hierarchy is just neat (and free) bonus IMO. I'll add a note about that.

The non-functional requirement (which I'll add) is already covered in https://github.com/ElektraInitiative/libelektra/pull/4797#discussion_r1060901517. I just noticed that over there I actually explain "remove range" not "access index range". But the logic is the same, the numbers are just a bit different. Asymptotically the range functions are better than calling the "normal" functions in a loop.

kodebach:
@markus2330 There is still one open question here: Should the range-based functions be part of `libelektra-core`?

My opinion is: Yes, because there is no good alternative anymore. IMO it doesn't make sense to keep "remove range" in `libelektra-core` and put other range based functions elsewhere. Now that "remove range" is the main (and only) scenario for "Remove `Key` from `KeySet`", we have one use case that definitely uses ranges in `libelektra-core`. Therefore, I say we keep the rest too.

If we want to move the range-based functions to e.g. `libelektra-extra`, we would need a separate implementation for "remove single" and "remove range".

On a non-technical level, I think it does make sense for a sorted data structure to consider range-based functions part of the core use cases. Especially for a sorted _and_ hierarchical data structure it makes sense, since every hierarchy is a range and hierarchies are clearly a core part of the data structure.

doc/help/elektra-glossary.md
```diff
@@ -102,6 +102,34 @@
...
+  e.g.: `system:/`, `user:/`, `/`
+
+- **Child** (of key K):
+  A key whose name starts the same as that of K, but has exactly one extra part.
+
+- **Parent** (of key K):
+  A key P where K is a child of P.
+  A namespace root key has no parent, every other key has exactly one parent.
+
+- **Descendant** (of key K):
```
markus2330, line 117: This means we need to change all occurrences (inclusive the API) of "below"

doc/help/elektra-glossary.md
```diff
@@ -102,6 +102,34 @@
...
+  A key whose name starts the same as that of K, with some extra parts.
+  In other words: A child of K, or a child of a child of K, etc.
+
+- **Ancestor** (of key K):
+  A key X where K is a descendant of X.
+  In other words: The parent of K, or the parent of the parent of K, etc.
+
+- **Key Hierarchy**:
+  A key hierarchy is a collection of keys, which are all descendants of a single key R.
+  This key R is called the _root key_ of the hierarchy.
```
markus2330, line 127: Unclear if a key hierarchy is all keys of a namespace or everything descendant of a specific key k. If it is the second, we don't need the term, as we already have descendant? (Btw. collection also would need to be defined).

If it is the first:

```suggestion
- **Key Hierarchy**:
  A key hierarchy are all keys, which are all descendants of the *namespace root key*.
```

doc/usecases/core/README.md
```diff
@@ -21,3 +21,11 @@ To support the hierarchical and ordered nature of a `KeySet` there are two funda
...
 
 The individual use cases provide details on these data structures and the operations it supports.
+
+## Why index based?
+
+The use cases are written in an "index based" fashion.
+Most use cases for `KeySet` deal only with indices.
+Only a select few use cases, are about turning these indices into `Key`s (or new `KeySet`s).
+
+The reason behind this is the flexibility it gives the caller.
```
markus2330, line 31:
```suggestion
The reason behind this is efficiency and the flexibility it gives the caller.
```

doc/usecases/core/UC_keyset_index_access_many.md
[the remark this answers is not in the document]
markus2330: > which I will add to the file

I didn't find them.

doc/usecases/core/UC_keyset_index_access_many.md
markus2330: Also on a second read it is quite difficult for me to relate to the text as a use case. It reads more like a description of semantics of a function than a use case.

doc/usecases/core/UC_keyset_insert.md
```diff
@@ -13,15 +13,20 @@
...
 - **Alternative scenario:**
   - Caller requests to insert all `Key`s contained in one `KeySet` A into another `KeySet` B
+  - Type of A matches type of B.
   - If necessary Core resizes B
   - If necessary Core removes all existing `Key`s contained in A and B from B
   - Core copies all `Key`s from A to B
-- **Error scenario:** -
+- **Error scenario:**
+  - Caller requests to insert a new `Key` with non-cascading name into `KeySet`
+  - Namespace of `Key` is prohibited by type of `KeySet`.
```
markus2330, line 28: type of `KeySet` also needs to be in glossary, if we need it. Otherwise we could simply say:

```suggestion
  - Namespace of `Key` is prohibited in this `KeySet`.
```

doc/usecases/core/UC_keyset_lookup_hierarchy.md
```diff
@@ -1,23 +1,23 @@
...
 - **Precondition:**
   - [`KeySet` has been created](UC_keyset_create.md)
 - **Main success scenario:**
-  - Caller asks Core to cut a `Key` hierarchy descendant from a given `Key` R from a `KeySet` KS.
-  - Core searches for the first `Key` in KS that is part of the desired hierarchy, i.e., the first `Key` in KS that is a descendant of R according to [`Key` hierarchy comparison](UC_keyname_hierarchy.md).
-  - If not found, Core returns an empty KS
-  - If found, Core searches for the last `Key` that is not part of the desired hierarchy, i.e., Core tries all `Key`s in KS in [keyname order](UC_keyname_ordering.md) until it finds a `Key` that is not a descendant of R.
-  - Core removes all `Key`s of the desired hierarchy from KS and puts them into a new `KeySet` KS1
-  - Core returns KS1.
+  - Caller asks Core to look up a `Key` hierarchy, which descends from a given `Key` R, from a `KeySet` KS.
```
markus2330, line 15: If R is the root key, it doesn't need be given? (Maybe use K if it is any key?)

markus2330 (changes requested):
Thank you! There are some issues in the terminology.

markus2330:
> Should the range-based functions be part of libelektra-core?

I agree, it makes sense to have the range-based functions in the core and the ones that can e.g. delete individual keys in some extra/helper lib.

doc/help/elektra-glossary.md
markus2330: It is not only a lot of work for you but also for me (lecture materials etc.). So if there isn't a good reason, we definitely shouldn't do it.

kodebach: > This means we need to change all occurrences (inclusive the API) of "below"

No it doesn't. It means, we to explain "below" as well. I don't see why we couldn't have synonyms. However, if you want exactly one term for everything (*) then "below" is clearly a worse choice, since AFAIK there is no noun form of "below".

(*) Having exactly one term can be good for precise specifications and we if we have synonyms we should decide on a "specification term" for everything. But specifications can be quite tedious to read. Most forms of documentation become better, if they are written in a less formal prose style. In that case it helps, if you have some synonyms to use.

doc/help/elektra-glossary.md
kodebach: > collection also would need to be defined

I'm not going to start a dictionary. I'm a software engineer, not a linguist.

I did not use "collection" with any specific definition in mind (e.g. `java.util.Collection`). It's just a group/set/assembly/... of things.

> Unclear if a key hierarchy is all keys of a namespace or everything descendant of a specific key k.

The definition says very clearly what a "key hierarchy" is.

All keys in a namespace would be an example of a key hierarchy, where the root key is the namespace root.

doc/usecases/core/UC_keyset_lookup_hierarchy.md
kodebach: What? How does the core know what to look for, if the caller doesn't give a key?

kodebach:
> I agree, it makes sense to have the range-based functions in the core and the ones that can e.g. delete individual keys in some extra/helper lib.

I don't think that's a good solution.

1. Most users will want the single key versions. So most users would have to add the additional library.
2. The range-based and single key versions will almost often share large parts of the implementation. To make that possible there would have to be private functions exposed by the core, which the additional library can use.
3. Some things only make sense as single key versions (e.g. (cascading) lookup). It would be quite weird to sometimes have a single key function and sometimes only a range-based function in the core.

Issue #4761, opened 2022-12-09
Vision for new API


kodebach:
I have now thought about the upcoming changes to the API as a whole. The details below are based on the decision PRs I have opened in the last few weeks. In particular, this implements the suggested outcomes of #4715 and #4745, uses the error guidelines from #4716, and is consistent with #4744 and #4717. I have chosen to mostly ignore the generic buffer struct idea from #4683 in favor of separate **public** (non-opaque) buffer structs for keyname and keyvalue.

Re refcounting (#4676) I have chosen an approach that is different from the one we use right now. I did this to show the single expression creation of a keyset is still possible. It should be fairly trivial, however, to introduce separate `*Del` functions and switch to the refcounting approach that is used right now.

This issue is just meant to showcase the general vision I have. I haven't had time yet to divide this into smaller chunks. Once I've done that, I will open new decision PRs (or update existing ones) with the concrete API change proposals.

### Private

```c
/* opaque */ struct ElektraKeynameCow {
    struct ElektraKeyname name;
    
    uint16_t refs;

    bool shouldFree : 1;
    int reserved : 15;
};

/* opaque */ struct ElektraKeyvalueCow {
    struct ElektraKeyvalue value;
    
    uint16_t refs;

    bool shouldFree : 1;
    int reserved : 15;
};

struct ElektraKeyset;

/* opaque */ struct ElektraKey {
    struct ElektraKeynameCow * name;
    struct ElektraKeyvalueCow * value;
    struct ElektraKeyset * meta;

    uint16_t refs;
    uint16_t nameLock;

    bool shouldFree : 1;
    bool needsSync : 1;
    int reserved : 14;
};

/* opaque */ struct ElektraKeyset {
    struct ElektraKeysetDataCow * data;

    uint16_t refs;

    bool shouldFree : 1; // used for mmap
    int reserved : 15;
};
```

### Public

```c
// indicates special rules for constness, see comments at use-site for details
#define ELEKTRA_CONST const

typedef uint8_t ElektraNamespace;

typedef struct {
    // constness follows containing struct
    // if the containing struct is a `const ElektraKeyname *` then this field must not be modified
    // if the containing struct is a `ElektraKeyname *` or a non-pointer `ElektraKeyname` it is okay to modify
    ELEKTRA_CONST char * name;
    size_t size;
    ElektraNamespace ns;
} ElektraKeyname;

typedef struct {
    const void * value;
    size_t size;
} ElektraKeyvalue;

[...]

PR #4805, opened 2023-01-03
add preview proposal of new api for core

kodebach:
NOT INTENDED FOR MERGING

This PR is intended to discuss the ideas from #4761.


doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+
+### Note on names
+
+From the [use cases](/doc/usecases/core/README.md) we know that there is a collection structure which contains elements with a name and value (and metadata).
+
+The current API uses these names for those concepts:
+
+- Collection: `KeySet`
+- Element: `Key`
+- Name of element: —
```
markus2330, line 13:
```suggestion
- Name of element: `char *`
```

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+- Element: `ElektraKey`
+- Name of element: `ElektraKeyname`
+- Value of element: `ElektraKeyvalue`
+
+Another (maybe less confusing) option would be:
+
+- Collection: `ElektraSet` or `ElektraMap`
+- Element: `ElektraEntry`
+- Name of element: `ElektraKey` or `ElektraName`
+- Value of element: `ElektraValue`
```
markus2330, line 28: I don't think this is a good idea. This would need a convincing usecase.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+The API proposed below uses (to keep as much as possible from above):
+
+- Collection: `ElektraKeyset`
+- Element: `ElektraKey`
+- Name of element: `ElektraKeyname`
+- Value of element: `ElektraKeyvalue`
+
+Another (maybe less confusing) option would be:
+
+- Collection: `ElektraSet` or `ElektraMap`
```
markus2330, line 25: :+1: ElektraSet is nearer to current name, I don't see a reason to change to Map.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+- Element: `ElektraEntry`
```
markus2330, line 26: :+1: This is a huge change but probably a good idea as the "Key" for a key/value/meta was very confusing.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+- Name of element: `ElektraKey` or `ElektraName`
```
markus2330, line 27: Probably `ElektraKey` if we keep KDB.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+- Value of element: `ElektraValue`
+
+### Note on public structs
+
+The public structs `ElektraKeyname` and `ElektraKeyvalue` defined below are not problematic or limiting in terms of forward compatibility.
```
markus2330, line 32: I disagree. E.g. order or if namespace is separate should be implementation detail (only shared with mmap).

Anyway, we should have a decision about why we have opaque pointers, this is a common question and not so well known.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+    size_t size;
+} ElektraKeyvalue;
+
+typedef struct ElektraKey ElektraKey;
+typedef struct ElektraKeyset ElektraKeyset;
+
+// 0 always indicates success, other values depend on function
+typedef int ElektraErrorCode;
+
+// Note: breaking change compared to current values
```
markus2330, line 65: Not really relevant for pre-1.0

```suggestion
```

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+
+typedef struct {
+    const void * value;
+    size_t size;
+} ElektraKeyvalue;
+
+typedef struct ElektraKey ElektraKey;
+typedef struct ElektraKeyset ElektraKeyset;
+
+// 0 always indicates success, other values depend on function
```
markus2330, line 62: API usage would imho much easier if more than that is known.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+
+// Note: breaking change compared to current values
+typedef enum {
+    ELEKTRA_KEY_CP_NAME   = 1<<0,
+    ELEKTRA_KEY_CP_VALUE  = 1<<1,
+    ELEKTRA_KEY_CP_META   = 1<<2,
+    ELEKTRA_KEY_CP_ALL    = ELEKTRA_KEY_CP_NAME | ELEKTRA_KEY_CP_VALUE | ELEKTRA_KEY_CP_META,
+} ElektraKeyCopyFlag;
+
+void elektraKeynamePushPart (ElektraKeyname * name, const char * part);
```
markus2330, line 73: Imho add/append makes more sense.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+ElektraKey * elektraKeyRetain (const ElektraKey * key);
+void elektraKeyRelease (const ElektraKey * key);
+
+const ElektraKeyname * elektraKeyGetName (const ElektraKey * key);
+// Note: creates copy of name
+ElektraErrorCode elektraKeySetName (ElektraKey * key, const ElektraKeyname * name);
+// increments nameLock
+void elektraKeyLockName (ElektraKey * key);
+// decrements nameLock
+void elektraKeyUnlockName (ElektraKey * key);
```
markus2330, line 86: I am not a fan of having two ref counters. Imho it is over-engineering. Needs use case and decision.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+bool elektraKeyIsNameLocked (const ElektraKey * key);
```
markus2330, line 87: We are still missing any indication if we should take care of having a small amount of functions. Please make such a benchmark. (e.g. size change if getting rid of all *Is* functions)

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+
+const ElektraKeyvalue * elektraKeyGetValue (const ElektraKey * key);
+// Note: creates copy of value
+ElektraErrorCode elektraKeySetValue (ElektraKey * key, const ElektraKeyvalue * value);
+
+/**
+ * @returns a `const ElektraKeyset *` that can safely be cast to `ElektraKeyset *`, iff a non-const `ElektraKey *` was passed as @p key
```
markus2330, line 94: ?

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+ */
+ELEKTRA_CONST ElektraKeyset * elektraKeyGetMeta (const ElektraKey * key);
+
+ElektraErrorCode elektraKeyIsBelow (const ElektraKey * first, const ElektraKey * second);
```
markus2330, line 98: This looks weird as if only an error code is returned.

```suggestion
ElektraReturnCode elektraKeyIsBelow (const ElektraKey * first, const ElektraKey * second);
```

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+ElektraErrorCode elektraKeysetInsertAll (ElektraKeyset * ks, ElektraKeyset * other);
+static inline ElektraErrorCode elektraKeysetInsertAndRelease (ElektraKeyset * ks, ElektraKey * key)
+{
+    ElektraErrorCode error = elektraKeysetInsert (ks, key);
+    elektraKeyRelease (key);
+    return error;
+}
+
+ElektraKey * elektraKeysetGet (ElektraKeyset * ks, size_t index);
+ElektraKeyset * elektraKeysetGetRange (ElektraKeyset * ks, size_t start, size_t end);
```
markus2330, line 127: This would need some use case.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+
+void elektraKeysetRemove (ElektraKeyset * ks, size_t index);
+void elektraKeysetRemoveRange (ElektraKeyset * ks, size_t start, size_t end);
```
markus2330, line 130: I don't think this is minimal API.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+static inline void elektraKeysetClear (ElektraKeyset * ks)
+{
+    elektraRemoveRange (ks, 0, elektraKeysetSize (ks));
+}
+
+size_t elektraKeysetLookup (ElektraKeyset * ks, const ElektraKeyname * name);
+
+size_t elektraKeysetFindHierarchy (ElektraKeyset * ks, const ElektraKeyname * root, size_t * end);
```
markus2330, line 138: Your favorite function :wink: 

Is the `Find` necessary?

```suggestion
size_t elektraKeysetHierarchy (ElektraKeyset * ks, const ElektraKeyname * root, size_t * end);
```

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+### Usage Examples
+
+```c
+// Create key (Note: better ways below via additional APIs)
+ElektraKey * k1 = elektraKeyNew (&(ElektraKeyname){
+    .ns = ELEKTRA_NS_SYSTEM,
+    .name = "foo\0bar\0baz",
+    .size = 14
+});
+
```
markus2330, line 196: Looks terrible.

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+    .size = 4
+};
+
+// ... using existing name
+ElektraKey * k2 = elektraKeyNew (&name2);
+
+// manipulate name
+elektraKeynamePushPart (&name2, "boo");
+ElektraKey * k3 = elektraKeyNew (&name2);
+
```
markus2330, line 213: example too long, can you please split up to small examples that show specific features of the API?

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+ * @param errorPtr If != NULL, will be a pointer into @p escaped, which points to the location of the first error, i.e. the error which is described by the return value
+ *
+ * @returns 0 if @p escaped is a valid escaped name (other values for different kinds of errors)
+ */
+ElektraErrorCode elektraKeynameProcessEscaped (
+    ElektraKeyname * dest,
+    const char * escaped,
+    char ** canonical,
+    size_t * canonicalSize,
+    const char ** errorPtr
```
markus2330, line 352: I hope we can do better. (Unless too many functions are too much size-overhead.)

doc/contrib/new_api.md
```diff
@@ -0,0 +1,510 @@
...
+
+```c
+int a = 7;
+
+ElektraKey * key = elektraKeyNew (&ELEKTRA_KEYNAME(ELEKTRA_NS_SYSTEM, "foo")); // refs = 1
+
+ElektraKeyset * ks = ELEKTRA_KEYSET (
+    // elektraKeysetBuild consumes reference, doesn't retain itself
+    // we call elektraKeyRetain, so we can keep our own reference
+    elektraKeyRetain (key), // refs = 2
```
markus2330, line 497: Doesn't look easy to use.

markus2330 (changes requested):
Many great ideas and improvements. I don't see a problem in merging this, especially if you add links to use cases and decisions, it would be a **great** overview.

doc/contrib/new_api.md
kodebach: Well "Map, Entry, Key, Value" would be a pretty standard combination of terms.

Although changing the meaning of "Key" like that will be confusing for a while. It also means we actually need to update **all the docs** in more or less one go, instead of adding a notice "we changed terms from X to Y" somewhere and updating over a longer period of time.

If we want to use "Set", I actually prefer the combination: "Set, Entry, Name, Value".

doc/contrib/new_api.md
kodebach: We could go with EDB for ElektraDatabase. Would be another huge change, but no bigger than "Key -> Entry" really.

doc/contrib/new_api.md
kodebach: I don't quite understand the problem here... Basically what this say is:
1. The term for "the value of an Element" is: _value_
2. If we need to represent it as a type it is: `ElektraValue`

We might not use `ElektraValue` in C (let's discuss that in the structs section below), but in some bindings it could make sense. For example in Haskell:

```haskell
module Elektra (
  Value
) where

import Data.ByteString

newtype Value = Data.ByteString
```

doc/contrib/new_api.md
kodebach: 1. Whether namespace is separate or not is IMO actually decision we need to make. IMO (and I've stated that before), namespace and name should be entirely separate things. Now that I have thought more about the API, I actually don't see **any benefit** in storing the namespace as the first byte of the name field. IMO it just complicates things.

2. Yes, field order could be reason to make the struct opaque. But in this case there will always be the same 3 fields, a pointer, a size and a byte (for namespace). I don't see why we would ever want to change the order. In any case, I don't think field order outweighs all the benefits that having these public structs brings.

> we should have a decision about why we have opaque pointers, this is a common question and not so well known.

Then **you** need to write one. You are the only one who has been with Elektra long enough to know the actual reason.

doc/contrib/new_api.md
kodebach: IMO we should have rules/guidelines somewhere in `doc/contrib` so that our error codes are mostly uniform. So for example, "Illegal null pointer is -1". I just wouldn't document that in the user/developer facing headers, because then we'd make it an actual guarantee. Better to leave the option for some weird edge case function in a library somewhere that needs to use non-standard error codes.

kodebach: PS. I also thought I about having some predefined values, that could be but don't have to be used by functions. That would just end up like `errno`, where every `E*` constant has a definition, but you (often) still need to check the individual function to see what the error actually means.

Again, we could have those constants in private headers to simplify writing code, but I wouldn't make that public.

doc/contrib/new_api.md
kodebach: Yeah, I know I just added at for completeness's sake.

doc/contrib/new_api.md
kodebach: Yes, "append" makes more sense. But there is no good single word counter part for removing. We could use "pop", but IMO "push"/"pop" as a pair makes more sense. Elektra is the only place where I've seen "append"/"pop" used a pair.

I didn't go with "add"/"remove", because that looses the information that we're only manipulating the last part. It would have to be the longer:

```c
void elektraKeynameAddLastPart (ElektraKeyname * name, const char * part);
// or maybe: void elektraKeynameAppendPart (ElektraKeyname * name, const char * part);
void elektraKeynameRemoveLastPart (ElektraKeyname * name);
```

doc/contrib/new_api.md
kodebach: **PLEASE** stop thinking of this as "two ref-counters". Describing it as that was the biggest mistake I/we ever made.

The name is locked with a re-entrant lock instead of a single bit flag.

doc/contrib/new_api.md
kodebach: I have no idea what kind of benchmark you want here....

Also, how would you replace `elektraKeyIsNameLocked`? You can emulate it with other functions of `libelektra-core`.

Is the question just, whether to keep it in `libelektra-core` or to put it into `libelektra-extra`? If so, I really don't think right now is the time to discuss this.

doc/contrib/new_api.md
kodebach: The constness works like for e.g. `strchr`:

```c
const char * cs;
char * s;

// OKAY
const char * x = strchr(cs);
char * x = strchr(s);
const char * x = strchr(s);

// ILLEGAL CAST, non-const pointer into const string
char * x = strchr(cs);
```

Basically what it means that, this is correct API use:

```c
ElektraKey * key;

ElektraKeyset * meta = (ElektraKeyset *) elektraKeyGetMeta (key);
```

But this is still wrong, because the metadata of a const `Key` should also be const:

```c
const ElektraKey * constKey;

ElektraKeyset * meta = (ElektraKeyset *) elektraKeyGetMeta (constKey);
```

> **Note** I didn't go with the "const arg, non-const return" variant from `strchr`, because the illegal case with `strchr` is not detected by the compiler (the illegal cast would actually be within `strchr`). With `elektraKeyGetMeta` the caller needs to make the cast explicitly otherwise there will be a compiler warning/error. Also I believe, the main reason why `strchr` uses `const` the way it does, is because the original API was written before `const` existed in C and changing it now would be a breaking change.

doc/contrib/new_api.md
kodebach: See #4797. It is the basis for the cut functions described below.

doc/contrib/new_api.md
kodebach: Can you give a reason?

If you want we can put the two `*Range` functions into `libelektra-extra`.

doc/contrib/new_api.md
kodebach: > Your favorite function :wink:

The function is simply much more useful than `ksCut` IMO. It also makes a lot of sense with the `*Range` functions.

> Is the `Find` necessary?

I don't like function names without a verb. If I read code that uses APIs that I don't know, I find it much easier to understand, if every function name has a verb that tells me what's happening.

In this case, "find"  IMO indicates:
1. The operation is non-destructive.
2. There is actual work happening. Unlike a "get", a "find" indicates to me that I shouldn't call the function more than I need to and should rather cache results. As a concrete example, calling `elektraKeyGetName` multiple times instead of creating a local variable won't have much real world difference. But calling `elektraKeysetFindHierarchy` (on the same unchanged keyset) over and over again, will waste CPU time.
3. It is similar to a "lookup", but not quite the same.

doc/contrib/new_api.md
kodebach: That's a very personal opinion. I actually find it much more pleasing to look at than the current var-arg stuff in `keyNew`. Not least, because this has a very clear way of formatting it properly.

doc/contrib/new_api.md
kodebach: That would just repeat the setup over and over again. I know the example is long, but that's why I added so many comments...

doc/contrib/new_api.md
kodebach: How do you want to do better? What is your problem with this function? Do you have any concrete ideas?

Just saying "I hope we can do better" helps nobody...

doc/contrib/new_api.md
kodebach: Manual reference counting is never easy to use. Manual memory management in general is always hard to use. That's why all modern languages have some form of automatic memory management....

At least with this retain/release system it's clear when you should call retain/release. You retain, when you want a new reference for yourself that you hold. You release when you're done with that reference.

With the current inc/dec/del system it's just chaos. You inc when things broke and you dec as soon as possible. You call del when ever, because calling it too much doesn't matter. This sort of chaos is just cause trouble. It's also much easier to create memory leaks. If retain/release a segfault is much more likely than a memory leak and segfaults lead to crashes and issue reports. Memory leaks are silent and easily go undetected.

Anyway, the reference counting is the part of this proposal I'm least fixated on. I'm open to using any solution that works. I just used retain/release, because **I** like it best and I wanted to show that it is a good fit for Elektra (which has been questioned in the past).

kodebach:
@markus2330 Please respond to the comments I left. There are some answers I need from you, before I can work on this further. However, since you seem to be generally in favour of the changes, I see this as a lower priority than all the decision PRs.

> I don't see a problem in merging this, [...] it would be a great overview.

Maybe right now. But I don't see a point in keeping this, once the headers are updated and new examples are written.

doc/contrib/new_api.md
markus2330: We first need to benchmark the too-many-functions "problem". First the problem must be clear, then we can talk about solutions.

doc/contrib/new_api.md
markus2330: You can assume the setup to be done, or (I would prefer) compilable C in separate files where you can have a function/macro for common setup.

In general I think it would be a good idea if the whole API and examples are compilable (maybe even implemented to some degree, e.g. for ref counting). This would lower the barrier to actually try out modifying your examples.

doc/contrib/new_api.md
markus2330: I thought also rough and unspecific feedback is welcomed. I don't understand the reply, where exactly should I look?

doc/contrib/new_api.md
markus2330: I don't have troubles with EDB¹ and I don't think that the EDB change is as difficult, as *every* occurrence from KDB would be EDB afterwards. The trouble with Key is that there are many occurrences of Key not having the meaning of "ElektraEntry". But you probably know that very well anyway.

¹ historically there was the idea that "Elektra" is only a marketing name but not used in technical context. But I think we are already way beyond that point.

doc/contrib/new_api.md
markus2330: > instead of adding a notice "we changed terms from X to Y" 

This doesn't work anyway (even with FLOSS it would never be finished).

> If we want to use "Set", I actually prefer the combination: "Set, Entry, Name, Value".

Probably better to avoid `Key` anyway, see https://github.com/ElektraInitiative/libelektra/pull/4805#discussion_r1066061990

doc/contrib/new_api.md
kodebach: Again. This part here is not about introducing `ElektraValue` as a type or identifier **anywhere**. This text is basically just a preview for the glossary. It just says we call this thing "value" (no other special name). The `ElektraValue` part basically just says **if** we found a reason to introduce a type, `ElektraValue` would be time. Essentially we reserve the name for this purpose.

doc/contrib/new_api.md
kodebach: > This doesn't work anyway (even with FLOSS it would never be finished).

It is not intended as a permanent solution. But exactly because of FLOSS/CM having such a notice might be useful. Otherwise people could get confused, while the change happens, or we are limited to making the change when there is no FLOSS/CM happening.

Also we could always accidentally miss an occurrence in the update. The notice could also be put into the release notes. If we reuse "Key" we end up with something like:

```
Key was renamed to Entry and Keyname was renamed to Key.
```

Which is ambiguous. You can't tell whether "Key" is the old meaning and we forgot to update, or it is the new meaning.

doc/contrib/new_api.md
kodebach: > I thought also rough and unspecific feedback is welcomed

Yes, "looks terrible" is not feedback at all. I can't make use of this in anyway. I don't know which part you don't like, or what you'd like better.

> I don't understand the reply, where exactly should I look?

I don't understand the question...

What I personally said was, I find this `elektraKeyNew` call more less terrible (i.e. "more pleasing to look at") than the current `keyNew (..., KEY_VALUE, ... KEY_META, ...)` stuff.

doc/contrib/new_api.md
kodebach: Everything should compile already. Just put the API definition in a header (or into your test `.c` file) and try writing code. The _linker_ will of course complain because the functions aren't defined anywhere, but compiling should work...

More than that is not on the table right now. I won't invest the time to create an implementation that might not be used at all.

doc/contrib/new_api.md
kodebach: > AFAIK we have less problems since keyDel is a NOP on wrong use. 

No it is not. "NOP on **wrong** use" is probably impossible and might be equivalent to solving the halting problem.

What you mean is that: Calling `keyDel` on a key with non-zero reference counter is a NOP.
But IMO that is not actually "wrong use" according to the current API and semantics.

In any case, wrong use definitely also includes a case like this:

```c
Key * k = keyNew (...);
keyDel (k);
keyDel (k);
```

And here obviously neither `keyDel` is a NOP.

> free(NULL) is also a NOP (maybe for that reason).

Please stop using this example, it is not a valid comparison. The equivalent for `keyDel` is `keyDel (NULL)`, which works exactly the same as `free (NULL)` (i.e. a NOP). The same would be true for `elektraKeyRelease (NULL)` which would also be a NOP.

> please let clarify such things only in decisions.

I tried in #4676, which went nowhere. Since you didn't even look at the solutions.

And like I said, the ref counting is the one part of the API proposal where I'm most open to other solutions. I just responded to once again, correct your misunderstandings/incorrect statements about the current `keyDel`.

doc/contrib/new_api.md
markus2330: Back to topic: can you maybe create a second PR with this new naming schema? Maybe the look&feel gives a clearer picture.

I think "Key" -> "Entry" is a bit like our ref counting: a historical "problem", where the "problem" is that we do it differently then the rest of the world.

So these are border-line decisions (not very convincing problem but also not totally unconvincing): we should only tackle them if you are sure you can actually fully migrate&fix everything.

kodebach: > we should only tackle them if you are sure you can actually fully migrate&fix everything

Renaming in C code is not an issue. Renaming in most other languages should be doable too. The issue is with documentation. I fear that has to be done manually. 

> I think "Key" -> "Entry" is a bit like our ref counting:

The difference is
1. ref counting is a technical issue with pros and cons on both sides
2. ref counting is a somewhat advanced topic that most people shouldn't need to worry about.

Using "Key" for something that is clearly not just a key, meanwhile, is just plain confusing to people who don't already know.

> Can you maybe create a second PR with this new naming schema?

I'll create a self-contained commit in this PR. If we like it we keep, if not I'll just revert the commit. Creating a new PR, unnecessarily splits the conversation. (It's unlikely the discussion in the new PR would really just focus on the names and nothing else).

doc/contrib/new_api.md
kodebach: See https://github.com/ElektraInitiative/libelektra/pull/4797/commits/1f2e0b902011e4a37e683749520eb9f258583e76 in #4797

doc/contrib/new_api.md
markus2330: The benchmark would be to compare:

- a version with very few functions and many arguments each, e.g. `elektraKeySetBit(int operation, int whatDoLock, ...)`
- a version with a minimal set of functions which have only little arguments each (more or less what we have now in the case of locking the key, asking for the sync bit etc.)

At the moment we are a bit inconsistent about this more functions vs. more arguments.

doc/contrib/new_api.md
markus2330: > What you mean is that

Important is what you mean, so there should be a very clear description of the problem in a decision.

> Please stop using this example, it is not a valid comparison.

If we are not open to different points of views, we will obviously struggle. I try to understand your viewpoint but failed up to now.

> Since you didn't even look at the solutions.

Because the described problem there (that we need ref counting) is already solved by the current API (=status quo). Decisions must be a problem of status quo, which might exist but is not described. Otherwise we only endlessly refactor based on different tastes.

doc/contrib/new_api.md
markus2330: No general conclusion needed. It is simply about what is better for Elektra, with the constraint that the core stays small in size.

doc/contrib/new_api.md
kodebach: > If we are not open to different points of views, we will obviously struggle.

I **am** open to different points of view. I'm just not open to ignoring or even accepting factually incorrect statements.

I have no pointed out to you on multiple occasions that your comparison of `keyDel` and `free` is flawed. You can compare `free (NULL)` and `keyDel (NULL)`. But you cannot compare `free (NULL)` with `keyDel (k)` where `k` is a key that's already deleted.

> > What you mean is that
> 
> Important is what you mean,

_No_, this was about what (I believe) **you** meant with "keyDel is a NOP on wrong use". You didn't mean "wrong use" in general, you just meant "calling keyDel on a key with non-zero reference counter". At least I assume that is what you mean, since it is the only NOP case for `keyDel` other than the trivial `keyDel (NULL)`.


> Decisions must be a problem of status quo, which might exist but is not described. Otherwise we only endlessly refactor based on different tastes.

Just as a sidenote, no idea to response: Once again, I started decisions for things that are not technically a problem with the status quo, because we are undeniably in the middle of a big API redesign. I already committed to doing that. I don't see why we should not at least think about doing more than strictly necessary, if it might result in further improvements. There is a "rejected" state for decisions. I don't understand, why you refuse to even start a discussion, when there is a clear option of "okay we explored our options, let's keep what we already have".

doc/contrib/new_api/core/public.h
```diff
@@ -0,0 +1,122 @@
...
+
+typedef enum
+{
+	ELEKTRA_KEY_CP_NAME = 1 << 0,
+	ELEKTRA_KEY_CP_VALUE = 1 << 1,
+	ELEKTRA_KEY_CP_META = 1 << 2,
+	ELEKTRA_KEY_CP_ALL = ELEKTRA_KEY_CP_NAME | ELEKTRA_KEY_CP_VALUE | ELEKTRA_KEY_CP_META,
+} ElektraEntryCopyFlag;
+
+void ElektraNamePushPart (ElektraName * name, const char * part);
```
markus2330, line 55: I don't think we should use stack terminology for building up a path (rather append/add/concat/join or similar). Do we even need a "pop"-equivalent?

markus2330 (changes requested):
ElektraSet/ElektraEntry looks good to me. Looking forward to other opinions.

@kodebach you can also invite others to have a look after you changed the main problems: public structs, ref counting (or created decisions for it).

doc/contrib/new_api/README.md
```diff
@@ -0,0 +1,110 @@
...
+- Element: `ElektraEntry`
+- Name of element: `ElektraName`
+- Value of element: `ElektraValue`
+
+### Note on public structs
+
+The public structs `ElektraName` and `ElektraValue` defined here are not problematic or limiting in terms of forward compatibility.
+This is because, they essentially define what a "keyname" and a "key value" are for Elektra.
+That is why they are used both as part of the public API **and** in the definitions of the `*Cow` structs, which define how stuff is actually stored.
+Any change to these structs would mean changing that definition and therefore would be major breaking change, even if the structs where not public.
```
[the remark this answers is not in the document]
kodebach: Now that I have explained my reasons, can you maybe explain why this is not convincing for you?

Can you give an example, where a breaking change is needed with this API

```c
const ElektraName * elektraEntryGetName (const ElektraEntry * entry);
```

but could be avoided with an API without structs like below?

```c
ElektraNamespace elektraEntryGetNamespace (const ElektraEntry * entry);
const char * elektraEntryGetName (const ElektraEntry * entry);
size_t elektraEntryGetNameSize (const ElektraEntry * enty);
```

doc/contrib/new_api/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+> **Note**: The examples compile with `gcc -c examples.c -Wall -pedantic`, but they do not link, since there is no implementation of the `elektra*` functions yet.
+
+For completeness, here are some more example of unsafe and illegal operations.
+These operations compile and "work" without reporting errors at runtime, but will cause various problems.
+An "illegal" operation, immediately break something and will cause wrong results, segfaults, etc. down the line.
+An "unsafe" operation, does not immediately break anything, but it is still an incorrect use of the API and often creates new "unsafe" or even "illegal" operations for the future.
+
+```c
+ElektraEntry * k1 = ElektraEntryNew (&(ElektraName){
```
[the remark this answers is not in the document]
kodebach: 1. The change to uppercase function names was by mistake. It seems I didn't enable "case preserving" when replacing key -> entry.
2. It's not so bad for this function, but for other functions you may get multiple `const char*` and especially `size_t` arguments. With a wrapper struct it's much clearer which arguments belong to each other. Also once again, these structs **are not internal**.

> 2 just counts the `\0`

Interesting idea, but I think that would be (1) exceedingly confusing to newcomers (2) makes the internals much less efficient. We'd need to iterate the string to find it's length (skipping over the given number of `\0`) and only then could we `memcpy` it. It also means you need an extra function (or field in `ElektraName`), which tells you the number of parts (`\0`s) in an entry's name.

doc/contrib/new_api/builders/public.h
```diff
@@ -0,0 +1,69 @@
...
+		.value = (s), .size = strlen ((s))                                                                                         \
+	}
+#define ELEKTRA_KEYVALUE_PTR(v)                                                                                                            \
+	(ElektraValue)                                                                                                                     \
+	{                                                                                                                                  \
+		.value = &(v), .size = sizeof ((v))                                                                                        \
+	}
+
+/**
+ * CAUTION releases reference to @p meta
```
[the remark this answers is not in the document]
kodebach: Well it's what makes most sense with the refcounting scheme used here. If we don't consume the reference the examples in `builders/examples.c` would leak. You would have to build all the metadata sets in separate variables use those in the build calls and then release all the metadata sets afterwards. This defeats the point of the builders, which is to have a single function that builds an entire set.

doc/contrib/new_api/builders/examples.c
```diff
@@ -0,0 +1,24 @@
+#include "public.h"
+
+int main (void)
+{
+	int a = 7;
+
+	ElektraEntry * key = ElektraEntryNew (&ELEKTRA_KEYNAME (ELEKTRA_NS_SYSTEM, "foo")); // refs = 1
```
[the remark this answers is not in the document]
kodebach: Again, uppercase names were accidental. I don't know what you mean with "several APIs for the same"?

doc/contrib/new_api/core/public.h
kodebach: I used stack terminology, because we only support appending exactly one part to the end or removing exactly one part from the end. That is exactly LIFO behaviour. 

Like I said already, "append" is a good alternative for "push", but I don't see a good alternative for "pop" (*). In that case I much prefer using push/pop, a common pair, rather than e.g. append/pop, which may confuse people because it seems like a mismatch.

(*) "remove"/"delete"/etc. all convey removing something, but not that it is explicitly the last element that gets removed.

> Do we even need a "pop"-equivalent?

Well, if you can only add to a name, there would be no way to replace the current `keySetBaseName` (without rebuilding the entire name).

doc/contrib/new_api/README.md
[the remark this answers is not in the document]
kodebach: A half-way option could be introducing only a public struct

```c
typedef struct {
    char * string;
    size_t size;
} ElektraString;
```

I think we can all agree that this is something that will never change. Then we can write the API like this:

```c
ElektraNamespace elektraEntryGetNamespace (const ElektraEntry * entry);
const ElektraString * elektraEntryGetName (const ElektraEntry * entry);

ElektraReturnCode ElektraSetNamespace (ElektraEntry * key, ElektraNamespace ns);
ElektraReturnCode ElektraSetName (ElektraEntry * key, const ElektraString * name);

const ElektraString * elektraEntryGetValue (const ElektraEntry * entry);
ElektraReturnCode ElektraSetValue (ElektraEntry * key, const ElektraString * value);
```

That way we still avoid all the separate `char *` and `size_t` arguments that could get mixed up.

> **Note**: If we go this way, I'll have to rethink whether passing by ref (`ElektraString *`) is still better than by value (`ElektraString`).

kodebach:
@markus2330 I fixed the function names (now proper camelCase). I also added examples for the two reference counting schemes:

1. `ref_inc_dec.c` shows the scheme of the current API with `inc`/`dec`/`del`
2. `ref_retain.c` shows a retain/release scheme, where "insert" moves an existing reference into the set
3. `ref_retain2.c` shows retain/release, but with two "insert" variants. One that creates a new reference and another that creates a new internal reference, but also destroys the external reference.

The original proposal used option 3, which IMO is the simplest option from a usage POV. However, I fully agree that changing the refcounting scheme is out of scope for use. That's why the proposal now uses a variation of option 1. The only change compared to the current API is that `inc`/`dec` now return their argument to allow inline use.

> **Note**: I need to rethink what we do about `const` in the API. I still like the idea behind `ELEKTRA_CONST`, but it's a bit annoying to use.

kodebach:
I will reopen this PR in a clean state with a few different options for the new API.

PR #4841, opened 2023-02-07
[decision] Changetracking (round 3)

atmaxinger:
## Basics

- [ ] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which
      contains `_(my name)_`)
      **Please always add something to the release notes.**
- [ ] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildservers are happy. If not, fix **in this order**:
  - [ ] add a line in `doc/news/_preparation_next_release.md`
  - [ ] reformat the code with `scripts/dev/reformat-all`
  - [ ] make all unit tests pass
  - [ ] fix all memleaks
- [ ] The PR is rebased with current master.

## Checklist

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation (see [Documentation Guidelines](https://www.libelektra.org/devgettingstarted/documentation))
- [ ] I fixed all affected decisions (see [Decision Process](https://www.libelektra.org/decisions/decision-process))
- [ ] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://www.libelektra.org/devgettingstarted/coding))
- [ ] I updated all meta data (e.g. README.md of plugins and [METADATA.ini](https://master.libelektra.org/doc/METADATA.ini))
- [ ] I mentioned [every code](/.reuse/dep5) not directly written by me in [reuse syntax](https://reuse.software/)

## Review

- [ ] Documentation is introductory, concise, good to read and describes everything what the PR does
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to [our Coding Guidelines](https://master.libelektra.org/doc/CODING.md)
- [ ] APIs are conforming to [our Design Guidelines](https://master.libelektra.org/doc/DESIGN.md)
- [ ] Code is consistent to [our Design Decisions](https://master.libelektra.org/doc/decisions)

## Labels

- [ ] Add the "work in progress" label if you do not want the PR to be reviewed yet.
- [ ] Add the "ready to merge" label **if the basics are fulfilled** and no further pushes are planned by you.

PR #4843, opened 2023-02-22
Changetracking implementation

atmaxinger:
## Basics

      contains `_(my name)_`)
  - [x] fix all memleaks

## Checklist

      (not in the PR description)
- [x] I fixed all affected documentation (see [Documentation Guidelines](https://www.libelektra.org/devgettingstarted/documentation))
- [x] I fixed all affected decisions (see [Decision Process](https://www.libelektra.org/decisions/decision-process))
- [x] I added code comments, logging, and assertions as appropriate (see [Coding Guidelines](https://www.libelektra.org/devgettingstarted/coding))

## Review

## Labels


atmaxinger:
This PR is accompanying the decision in PR #4841. I decided it would be helpful to see whether it is even feasible to implement as-is.

As of now, I also updated the `dbus` and the `internalnotification` plugins to use the new API. 

I think the naming of the methods could be better. They got quite some long names currently. 

Also, I introduced a new data type called `KeySetDiff`, which - unsurprisingly - holds the difference between two KeySets. I think this might be something useful even outside the changetracking context. Maybe we can make it more "official" and drop the `elektraChangeTracking` prefix of its accompanying methods?

src/plugins/internalnotification/internalnotification.c
```diff
@@ -241,6 +243,70 @@ static int keySetContainsSameOrBelow (Key * check, KeySet * ks)
 	return 0;
 }
 
+void elektraInternalnotificationNotifyChangedKeys (Plugin * plugin, const KeySetDiff * diff)
+{
+	PluginState * pluginState = elektraPluginGetData (plugin);
+	ELEKTRA_ASSERT (pluginState != NULL, "plugin state was not initialized properly");
+
+	KeySet * modifiedKeys = elektraChangeTrackingGetModifiedKeys (diff);
+	// TODO: do we also consider added and removed keys as modified here?
```
atmaxinger, line 252: This is something that I'm not sure yet.

src/libs/plugin/plugin.c
```diff
@@ -152,6 +152,32 @@ void * elektraPluginGetData (Plugin * plugin)
...
+ * @return KDB
+ */
+KDB * elektraPluginGetKdb (Plugin * plugin)
+{
+	if (plugin->global == NULL)
+	{
+		return NULL;
+	}
+
+	Key * kdbKey = ksLookupByName (plugin->global, "system:/elektra/kdb", 0);
```
atmaxinger, line 168: AFAICT this is key is always present, so I guess it is okay to rely on this?

src/plugins/dbus/dbus.c
```diff
@@ -90,31 +82,15 @@ int elektraDbusCommit (Plugin * handle, KeySet * returned, Key * parentKey)
...
-	// because elektraLogchangeGet will always be executed before elektraLogchangeCommit
-	// we know that oldKeys must exist here!
-	KeySet * oldKeys = pluginData->keys;
+	KDB * kdb = elektraPluginGetKdb (handle);
+	const ChangeTrackingContext * context = elektraChangeTrackingGetContextFromKdb (kdb);
 
-	KeySet * addedKeys = ksDup (returned);
-	KeySet * changedKeys = ksNew (0, KS_END);
-	KeySet * removedKeys = ksNew (0, KS_END);
+	// we do a fresh calculation here, because we want the most current diff to avoid false positives as much as possible
```
atmaxinger, line 88: This comment is here because I initially experimented with caching the diffs within the `ChangeTrackingContext`. I've refrained from this as of now. The reason being that we want to eliminate false positives as much as possible, and currently we have no control over when transformations are being applied.

src/plugins/dbus/dbus.c
```diff
@@ -90,31 +82,15 @@ int elektraDbusCommit (Plugin * handle, KeySet * returned, Key * parentKey)
 	ElektraDbusPluginData * pluginData = elektraPluginGetData (handle);
 	ELEKTRA_NOT_NULL (pluginData);
 
-	// because elektraLogchangeGet will always be executed before elektraLogchangeCommit
-	// we know that oldKeys must exist here!
-	KeySet * oldKeys = pluginData->keys;
+	KDB * kdb = elektraPluginGetKdb (handle);
+	const ChangeTrackingContext * context = elektraChangeTrackingGetContextFromKdb (kdb);
```
markus2330, line 86: I don't think we should offer a way to get `kdb` (unless there is a valid use case for this). Better to have a getter directly for getting the change tracking.

src/plugins/dbus/dbus.c
markus2330: We should fixate what transformation plugins are allowed in the decision. A more performant change tracking is much more important than crazy transformation features.

src/plugins/internalnotification/internalnotification.c
markus2330: Don't the unit tests check this? Unit tests might be disabled. If not, please add some. Intuitively, both should trigger a notification, as the added or removed key might be a key the user is subscribed to? (Removing would go back to default once the spec plugin works properly.)

src/libs/plugin/plugin.c
markus2330: Once we document that the key must be there, we can assume it. (And it would be a bug if it isn't.)

markus2330 (approved):
Great work, good to see progress here!

> I think the naming of the methods could be better. They got quite some long names currently.

Yes, I agree. But let us rename it later when the rest of the API gets more clear.

> Also, I introduced a new data type called KeySetDiff,

Why not directly call it Diff? We don't have other diffs?

> which - unsurprisingly - holds the difference between two KeySets. I think this might be something useful even outside the changetracking context. Maybe we can make it more "official" and drop the elektraChangeTracking prefix of its accompanying methods?

Yes, what about simply calling it elektraDiff*.

src/libs/plugin/plugin.c
kodebach: It's created in `kdbNew`, i.e. almost immediately after the `struct _KDB` memory is allocated. 

https://github.com/ElektraInitiative/libelektra/blob/0082b7012d7a974cee5ee82098fd0dcb38274828/src/libs/elektra/kdb.c#L295-L296

IIRC I added this key, for a purpose exactly like this one. But I didn't create a function in the plugin API, because I wasn't sure whether letting plugins access the whole `KDB` instance is a good idea.

That's also why `elektraPluginFromMountpoint` takes a `Plugin *` and retrieves the necessary data internally. If we keep the `elektraPluginGetKdb` function, we should definitely ensure that plugins cannot call any of the `kdb*` functions on the `KDB *` the receive. For example, via a `callInProgress` flag in `struct _KDB`.

src/plugins/dbus/dbus.c
kodebach: For once, I'll be the one to say it: Before we think about performance, we need benchmarks. Maybe not caching the diff is actually not a noticeable performance impact.

src/libs/plugin/plugin.c
atmaxinger: Yeah, I agree it's better to create a function that takes a `Plugin *` and returns the `ChangeTrackingContext *`. It's also cleaner for the tests.

src/include/kdbchangetracking.h
```diff
@@ -0,0 +1,47 @@
...
+KeySet * elektraChangeTrackingGetAddedKeys (const KeySetDiff * ksd);
+KeySet * elektraChangeTrackingGetRemovedKeys (const KeySetDiff * ksd);
+KeySet * elektraChangeTrackingGetModifiedKeys (const KeySetDiff * ksd); // Returns old keys (pre-modification)
+
+bool elektraChangeTrackingValueChanged (const KeySetDiff * ksd, Key * key);
+bool elektraChangeTrackingMetaChanged (const KeySetDiff * ksd, Key * key);
+
+KeySet * elektraChangeTrackingGetAddedMetaKeys (const KeySetDiff * ksd, Key * key);
+KeySet * elektraChangeTrackingGetRemovedMetaKeys (const KeySetDiff * ksd, Key * key);
+KeySet * elektraChangeTrackingGetModifiedMetaKeys (const KeySetDiff * ksd, Key * key); // Returns old meta keys (pre-modification)
```
kodebach, line 38: Since none of these actually involve change tracking (no `ChangeTrackingContext`), I'd move them to a separate header (or even library) for diffs.

I'd say we use the prefix `elektraDiff*`, call the struct `ElektraDiff` and if it's a separate library call it `libelektra-diff`.

src/libs/elektra/kdb.c
```diff
@@ -61,6 +62,19 @@ static void clearAllSync (KeySet * ks)
 	}
 }
 
+static void ksAppendDeepDup (KeySet * ks, const KeySet * toAppend)
```
kodebach, line 65:
```suggestion
static void ksAppendDup (KeySet * ks, const KeySet * toAppend)
```

I think adding `Dup` is enough to signify it calls `keyDup` compared to `ksAppend` which does nothing.

src/include/kdbchangetracking.h
```diff
@@ -0,0 +1,47 @@
...
+{
+extern "C" {
+#endif
+
+typedef struct _ChangeTrackingContext ChangeTrackingContext;
+typedef struct _KeySetDiff KeySetDiff;
+
+const ChangeTrackingContext * elektraChangeTrackingGetContextFromKdb (KDB * kdb);
+
+KeySetDiff * elektraChangeTrackingCalculateFromContext (KeySet * newKeys, const ChangeTrackingContext * context, Key * parentKey);
```
kodebach, line 20:
```suggestion
KeySetDiff * elektraChangeTrackingCalculateDiff (KeySet * newKeys, const ChangeTrackingContext * context, Key * parentKey);
```

src/include/kdbchangetracking.h
```diff
@@ -0,0 +1,47 @@
...
+KeySetDiff * elektraChangeTrackingCalculateFromKeySets (KeySet * newKeys, KeySet * oldKeys, Key * parentKey);
```
kodebach, line 21:
```suggestion
KeySetDiff * elektraDiffCalculate (KeySet * newKeys, KeySet * oldKeys, Key * parentKey);
```

src/libs/elektra/changetracking.c
```diff
@@ -0,0 +1,601 @@
...
+ */
+ChangeTrackingContext * elektraChangeTrackingCreateContextForTesting (KeySet * oldKeys)
+{
+	ChangeTrackingContext * context = elektraCalloc (sizeof (ChangeTrackingContext));
+
+	ksIncRef (oldKeys);
+	context->oldKeys = oldKeys;
+
+	return context;
+}
```
kodebach, line 61: Do we need this here? Couldn't this be in the tests somewhere? If it's needed in multiple places, you could create a header `src/include/kdbchangetracking_tests.h` with:

```c
static inline ChangeTrackingContext * elektraChangeTrackingCreateContextForTesting (KeySet * oldKeys)
{
	ChangeTrackingContext * context = elektraCalloc (sizeof (ChangeTrackingContext));

	ksIncRef (oldKeys);
	context->oldKeys = oldKeys;

	return context;
}
```

src/libs/elektra/changetracking.c
```diff
@@ -0,0 +1,601 @@
...
+
+/**
+ * @internal
+ *
+ * @brief For testing purposes only: Delete a ChangeTrackingContext object
+ *
+ * @param context the object to delete
+ */
+void elektraChangeTrackingContextDel (ChangeTrackingContext * context)
```
kodebach, line 70: Same as above, if it's just for testing it should be separated somehow.

src/libs/elektra/changetracking.c
```diff
@@ -0,0 +1,601 @@
...
+	for (elektraCursor itOld = 0; itOld < ksGetSize (old); itOld++)
+	{
+		Key * needle = ksAtCursor (old, itOld);
+
+		if (parentKey != NULL && !keyIsBelowOrSame (parentKey, needle))
+		{
+			continue;
+		}
+
+		Key * found = ksLookup (new, needle, 0);
```
kodebach, line 160: Of course I don't have benchmarks to proof it, but this `ksLookup` looks very much like a bottleneck. Both `new` and `old` are sorted the same way. You could just iterate them both at the same time, and compare the keys with `keyCmp (oldKey, newKey)`:
- If both have the same name (`keyCmp == 0`), do a comparison (`else` branch below). Increment both cursors.
- If `keyCmp(oldKey, newKey) < 0`, `oldKey` was removed from `new`. Increment just cursor for `old` and check again.
- If `keyCmp(oldKey, newKey) > 0`, `newKey` was added to `new`. Increment just cursor for `new` and check again.

src/libs/elektra/changetracking.c
```diff
@@ -0,0 +1,601 @@
...
+				if (ksGetSize (addedKeys) > 0 || ksGetSize (removedKeys) > 0 || ksGetSize (modifiedKeys) > 0)
+				{
+					// there was a change in the meta keys --> modified
+					ksAppendKey (modifiedKeys, needle);
+				}
+			}
+		}
+	}
+
+	for (elektraCursor itNew = 0; itNew < ksGetSize (new); itNew++)
```
kodebach, line 198: If you do the simultaneous iteration and `keyCmp` from above, this loop is also unnecessary.

kodebach (approved):
Looks great!

Regarding names: Using `elektraDiff*` will improve things already. Maybe we can also find a better name for the other `elektraChangetracking*`  functions, possibly just `elektraChanges*`. But by using `elektraDiff*` I think we only leave a handful of `elektraChangetracking*`  functions, so the long names might not be as bad.

Also in general: In plugins and for other non-public functions (as in functions which aren't declared in a public header), you don't have to use these long prefixes. You should still use `elektra*` to avoid collisions, but we could e.g. use `elektraIN*` for the `internalnotification` functions.

I also like how you actually used the refcounting properly. @markus2330 we should have a lot more `ksIncRef` and `keyIncRef` like done here, all over the place.

src/include/kdbchangetracking.h
atmaxinger: Yes I was thinking about the same thing. However, not sure how much effort it is to put it in its own library.

src/libs/elektra/changetracking.c
atmaxinger: But then the problem would be that the tests themselves would need to have access to the internal structure of the `ChangeTrackingContext` struct.

src/include/kdbchangetracking.h
kodebach: Shouldn't be too much effort, just a bit of CMake. But you also just leave it in here for now. I'll need to move everything around at some point anyway to make things fit with the new [header concept](https://github.com/ElektraInitiative/libelektra/blob/master/doc/decisions/4_decided/header_file_structure.md)

src/libs/elektra/changetracking.c
kodebach: That's not really an issue in the tests. You can just include the private headers.

src/libs/elektra/changetracking.c
atmaxinger: Good catch! It took me a while to implement, but you're right. Based on some benchmarking (500 keys - 100 unmodified, 100 modified, 100 added, 100 removed) the current implementation averages at about ~350 microseconds, and your proposed implementation at about 27 microseconds.

src/libs/elektra/kdb.c
atmaxinger: yup

src/include/kdbchangetracking.h
atmaxinger: I split it out into separate files. I'll let you do the library then 😅

atmaxinger:
Can we do something about DBus not being found on macOS?

kodebach:
AFAICT the issue is the otherway around, the `dbus` plugin code can't find the `elektraDiff*` functions. I think there is a missing dependency from the plugin to the  `kdb`.

atmaxinger:
Yes, thats another error. I'm talking about

```
Error: Formula `dbus` has not implemented #plist, #service or installed a locatable service file
```

https://cirrus-ci.com/task/5449741161463808?logs=install#L974

atmaxinger:
@Eiskasten what did you do when the build on macOS could not find the [kdb libs](https://github.com/ElektraInitiative/libelektra/pull/4843/checks?check_run_id=11659163112)?

tests/ctest/CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ target_link_elektra (test_cmerge elektra-merge)
 
 target_link_elektra (test_sha-256 elektra-ease)
 
+target_link_libraries (test_changetracking elektra-kdb)
```
Eiskasten, line 40: In my case the problem was that the linker was not able to find the xfconf library so this case may be a bit different. However, I would try to replace `target_link_libraries` with `target_link_elektra`. The test `test_backends` e.g. includes `kdb.h` already and it seems there was no issue on the past.

tests/ctest/CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ target_link_elektra (test_cmerge elektra-merge)
...
+
+target_link_libraries (test_diff elektra-kdb)
+target_link_libraries (test_diffprivate elektra-kdb)
```
Eiskasten, line 43:
```suggestion
target_link_elektra (test_changetracking elektra-kdb)

target_link_elektra (test_diff elektra-kdb)
target_link_elektra (test_diffprivate elektra-kdb)
```

Eiskasten: In my case the problem was that the linker was not able to find the xfconf library so this case may be a bit different. However, I would try to replace use these macros. The test `test_backends` e.g. includes `kdb.h` already and it seems there was no issue on the past.

src/include/kdbchangetracking.h
```diff
@@ -0,0 +1,32 @@
...
+#ifdef __cplusplus
+namespace ckdb
+{
+extern "C" {
+#endif
+
+typedef struct _ChangeTrackingContext ChangeTrackingContext;
+
+const ChangeTrackingContext * elektraChangeTrackingGetContextFromKdb (KDB * kdb) ELEKTRA_WEAK;
+const ChangeTrackingContext * elektraChangeTrackingGetContextFromPlugin (Plugin * plugin) ELEKTRA_WEAK;
```
atmaxinger, line 18: As it turns out, we need to declare these two functions as `weak` on macOS, else the tests will not pass.

src/plugins/internalnotification/internalnotification.c
```diff
@@ -250,7 +316,7 @@ static int keySetContainsSameOrBelow (Key * check, KeySet * ks)
  *                  e.g. elektraInternalnotificationGet or elektraInternalnotificationCommit)
  *
  */
-void elektraInternalnotificationUpdateRegisteredKeys (Plugin * plugin, KeySet * keySet)
+/*void elektraInternalnotificationUpdateRegisteredKeys (Plugin * plugin, KeySet * keySet)
```
flo91, line 319: Please remove code instead of commenting out, esp. before merging into master.

src/plugins/internalnotification/internalnotification.h
```diff
@@ -22,7 +23,8 @@ int elektraInternalnotificationOpen (Plugin * handle, Key * errorKey);
 Plugin * ELEKTRA_PLUGIN_EXPORT;
 
 // Not exported by plugin; used for testing
-void elektraInternalnotificationUpdateRegisteredKeys (Plugin * plugin, KeySet * keySet);
+// void elektraInternalnotificationUpdateRegisteredKeys (Plugin * plugin, KeySet * keySet);
```
flo91, line 26: Please remove commented out function when finished.

src/libs/elektra/changetracking.c
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return &kdb->changeTrackingContext;
+}
+
+/**
+ * Returns the changetracking context for the KDB instance associated with the specified plugin
+ *
+ * @param plugin the plugin
+ * @return ChangeTrackingContext or @p NULL if @p plugin is NULL or does not have a @p KDB instance associated with it
```
flo91, line 24:
```suggestion
 * @return ChangeTrackingContext or @p NULL if @p plugin is @p NULL or does not have a @p KDB instance associated with it
```

src/libs/elektra/changetracking.c
```diff
@@ -0,0 +1,98 @@
+#include <kdbchangetracking.h>
+#include <kdbprivate.h>
+
+/**
+ * Returns the changetracking context of the given KDB instance
+ *
+ * @param kdb the KDB instance
+ * @return ChangeTrackingContext or @p NULL if @p kdb is NULL
```
flo91, line 8:
```suggestion
 * @return ChangeTrackingContext or @p NULL if @p kdb is @p NULL
```

src/libs/elektra/diff.c
```diff
@@ -0,0 +1,593 @@
...
+ * @note The reference counter can never exceed `UINT16_MAX - 1`. `UINT16_MAX` is
+ * reserved as an error code.
+ *
+ * @post @p ksd's reference counter is > 0
+ * @post @p ksd's reference counter is <= UINT16_MAX - 1
+ *
+ * @param ksd the ElektraDiff object whose reference counter should be increased
+ *
+ * @return the updated value of the reference counter
+ * @retval UINT16_MAX on NULL pointer
```
flo91, line 528:
```suggestion
 * @retval UINT16_MAX on @p NULL pointer
```

src/libs/elektra/diff.c
```diff
@@ -0,0 +1,593 @@
...
+/**
+ * @brief Decrement the reference counter of a ElektraDiff object
+ *
+ * @post @p ksd's reference counter is >= 0
+ * @post @p ksd's reference counter is < SSIZE_MAX
+ *
+ * @param ksd the ElektraDiff object whose reference counter should get decreased
+ *
+ * @return the updated value of the reference counter
+ * @retval UINT16_MAX on NULL pointer
```
flo91, line 557:
```suggestion
 * @retval UINT16_MAX on @p NULL pointer
```

src/libs/elektra/diff.c
```diff
@@ -0,0 +1,593 @@
...
+	return ksd->refs;
+}
+
+/**
+ * @brief Returns the reference counter of a ElektraDiff object
+
+ * @param ksd the ElektraDiff object whose reference counter should get returned
+ *
+ * @return the value of the reference counter
+ * @retval UINT16_MAX on NULL pointer
```
flo91, line 583:
```suggestion
 * @retval UINT16_MAX on @p NULL pointer
```

flo91 (changes requested):
Great progress with a good decision and tutorial for getting started as well as lots of unit tests! 🥇 
The coding-style and -structure helps with reading and understanding the code.

I've noticed that the formatting of the C-Codeblocks in the `.md` files only works correctly if the code starts with a block (e.g. `{`, function-definition, if-statement, ...).
Esp. the function-calls are not highlighted otherwise.


As @kodebach already mentioned, the parts for testing look a bit inappropriate in the public parts of the library. Maybe you can move this to dedicated testing-code.

For more details, please see my comments and suggestions I made directly on specific parts of your awesome contribution. :sunglasses:

src/libs/elektra/symbols.map
```diff
@@ -144,6 +144,33 @@ libelektra_1.0 {
 	ksIncRef;
 	ksDecRef;
 	ksGetRef;
+
+	# kdbchangetracking.h
+	elektraChangeTrackingGetContextFromKdb;
+	elektraChangeTrackingGetContextFromPlugin;
+	elektraChangeTrackingContextDel;
```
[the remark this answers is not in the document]
atmaxinger: As long as we don't have the function anywhere else, we need to export it.

src/libs/elektra/diff.c
```diff
@@ -0,0 +1,593 @@
...
+	ssize_t newSize = ksGetSize (new);
+
+	elektraCursor iOld = 0;
+	elektraCursor iNew = 0;
+
+	for (ssize_t i = 0; i < oldSize + newSize; i++)
+	{
+		Key * oldKey = ksAtCursor (old, iOld);
+		Key * newKey = ksAtCursor (new, iNew);
+
```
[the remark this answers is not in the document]
atmaxinger: good catch

src/plugins/internalnotification/internalnotification.c
atmaxinger: done

src/plugins/internalnotification/internalnotification.h
atmaxinger: done

src/libs/elektra/diff.c
```diff
@@ -0,0 +1,593 @@
...
+					// there was a change in the meta keys --> modified
+					ksAppendKey (modifiedKeys, oldKey);
+				}
+				else if (oldMetaSize > 0 || newMetaSize > 0)
+				{
+					ksClear (metaAdded);
+					ksClear (metaRemoved);
+					ksClear (metaModified);
+
+					findDifferences (newMeta, oldMeta, metaAdded, metaRemoved, metaModified, NULL);
```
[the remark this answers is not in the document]
atmaxinger: yep me too ;)

src/libs/elektra/diff.c
```diff
@@ -0,0 +1,593 @@
...
+	if (oldKey == NULL)
+	{
+		return NULL;
+	}
+
+	KeySet * addedKeys = ksNew (0, KS_END);
+	KeySet * removedKeys = ksNew (0, KS_END);
+	KeySet * modifiedKeys = ksNew (0, KS_END);
+
+	findDifferences (keyMeta (key), keyMeta (oldKey), addedKeys, removedKeys, modifiedKeys, NULL);
```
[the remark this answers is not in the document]
atmaxinger: Great idea! I've implemented this now.

atmaxinger:
Good news everyone! After having replaced all change tracking implementations (that I found) in the plugins with our new version, I ran some benchmarks with the `dbus` plugin.

The benchmark consisted of creating X (between 50 and 50k) keys into an empty KDB. Then read that again and modify exactly half of the keys. I also ran it with the `dbus` plugin enabled and disabled. As it turns out, The performance difference between those implementations is IMHO negligible. 

grafik

grafik

kodebach:
1. Since your X-axis is in log-scale and the runtime seems to scale linearly, it might make the graphs easier to read, if the Y-axis is also log scale.
2. Would be interesting to plot just the overhead of tracking. So unless there is a way to measure it directly, simply plot the difference between "with tracking" and "without tracking" for "old" and "new". I think that would show best whether the difference is significant.
3. Why is inserting slower with the new version?
4. Where these a single run each, or is each of the bars an average of a few runs? If it's a single run, it could explain the inconsistent scaling for inserting. Sometimes new is slower, sometimes it's faster.

atmaxinger:
1. Yeah, you're right. I've updated the graphs.
2. Good idea! I've added two graphs.
3. I don't know. I think it might just be variance.
4. Yes. For each amount of keys, I executed a program that does both inserting and modifying 10 times.

kodebach:
The graphs are definitely a lot better now. It's still a bit weird, that with inserting 5000 keys the overhead for new is so much bigger. How did you create the datasets? Is it just numbered keys, i.e. more keys also means larger keynames? Because than it's probably related to that.

atmaxinger:
> Is it just numbered keys, i.e. more keys also means larger keynames

Yes, this is indeed the case. The keys are `user:/test/key0` to `user:/test/key50000`, and the value is `value0` to `value50000`.

Here is the program:

```c
#include <benchmarks.h>

int main (int argc, char **argv)
{
	Key * parentKey = keyNew ("user:/test", KEY_END);
	KeySet * contract = ksNew (0, KS_END);
	KDB * kdb = kdbOpen (contract, parentKey);

	KeySet * ks = ksNew (0, KS_END);
	kdbGet (kdb, ks, parentKey);

	size_t keys = 50000;
	if (argc == 2)
	{
		keys = atoi (argv[1]);
	}


	char nameBuffer [1024] = "";
	char valueBuffer [1024] = "";

	for (size_t i = 0; i < keys; i++)
	{
		snprintf (nameBuffer, 1023, "user:/test/key%zu", i);
		snprintf (valueBuffer, 1023, "value%zu", i);
		ksAppendKey (ks, keyNew (nameBuffer, KEY_VALUE, valueBuffer, KEY_END));
	}

	timeInit ();
	kdbSet (kdb, ks, parentKey);
	int insertingTime = timeGetDiffMicroseconds ();
	//printf ("Inserting %zu keys took %d us (%d ms)\n", keys, insertingTime, insertingTime / 1000);

	ksClear (ks);
	kdbGet (kdb, ks, parentKey);

	size_t modified = 0;
	for (size_t i = keys/2; i < keys; i++)
	{
		snprintf (nameBuffer, 1023, "user:/test/key%zu", i);
		snprintf (valueBuffer, 1023, "value-modified%zu", i);
		ksAppendKey (ks, keyNew (nameBuffer, KEY_VALUE, valueBuffer, KEY_END));
		modified++;
	}


	timeInit ();
	kdbSet (kdb, ks, parentKey);
	int modifyTime = timeGetDiffMicroseconds ();
	//printf ("Modifying %zu keys took %d us (%d ms)\n", modified, modifyTime, modifyTime / 1000);

	printf ("%d;%d\n", insertingTime, modifyTime);

	return 0;
}
```

kodebach:
You could try if using `%05zu` instead of `%zu` makes a difference (would also be interesting for future benchmarks.

atmaxinger:
So I dug a bit deeper, and it seems those benchmarks are mostly I/O (especially `dbus`) bound. 

Sending DBus messages accounts for 42% of the time, while diff calculations is 1%. Which makes sense, as calculating the diff for 400 000 keys only takes about 20 milliseconds on this machine.

grafik

atmaxinger:
I guess it suffices to say that performance of this implementation is good enough.

kodebach:
> So I dug a bit deeper, and it seems those benchmarks are mostly I/O (especially `dbus`) bound.

That was to be expected, but the I/O overhead should be the same for both implementations, so the comparison is still valid.

> Which makes sense, as calculating the diff for 400 000 keys only takes about 20 milliseconds on this machine.

That is impressive, although the questions is whether that is an average or a best-case number ;) I mean in terms of dataset, not regarding the machine in question.

kodebach:
- `multifile`: I'd say remove it. It seems like this was written for one specific use case, but the use case isn't document. I say that because it is oddly limited (no arbitrary set of files, just a glob), yet fully featured (as in has complete get & set).
- `rename`: As you said the sync-flag stuff has to stay there as long as there is another use of `keyNeedSync`. But then it can be removed.
-  `xmltool`: I'd say remove it. @markus2330 Which tests depend on the plugin **and** actually test something other than the plugin itself? I only found a handful of examples in the man pages with could easily use another format.
- `set_keyset`: Remove it
- `backendsDivideInternal`: Yes, switch to `elektraDiff`. We already saw the impact is quite minimal. If we always calculate the diff in `kdbSet` this also allows more granular backend implementations. For the current formats, it's not helpful, you need to rewrite the whole file anyway. But you could imagine formats that may use the diff to know exactly which parts of the file to change. Also database backends could probably benefit from using a diff. So basically, using `elektraDiff` gives more features. IMO that outways the cost.

markus2330:
> Which tests depend on the plugin and actually test something other than the plugin itself?

I once tried to remove the xmltool plugin and there were several substantial tests relying on them. But this was also quite some time ago, so maybe some of these parts are gone now anyway (at least one of them, the tcl plugin, actually already is removed).

> I only found a handful of examples in the man pages with could easily use another format.

Yes, if someone puts in the work and does it carefully it definitely won't be a problem to remove the xmltool. It doesn't have a purpose anymore.

> Yes, switch to elektraDiff.

:+1:

kodebach:
> I once tried to remove the xmltool plugin [...]

So in other words: You don't really now.

@atmaxinger Please try to remove `xmltool` and report if there are any issues.

atmaxinger:
@kodebach what shall we do with cascading keys regarding diffing? While replacing `backendData->keysNeedSync` with `elektraDiff` I noticed that some tests use cascading keys in the keysets they pass to `kdbSet`. Now if we diff those, they will always come up as removed, because the backends themselves do not contain cascading keys. 

We've got 2 options:
1. Generally ignore cascading keys in the diffing algorithm
2. Use a meta key on the `parentkey` parameter of the diffing algorithm to specify whether to ignore cascading keys or not
3. Providing functions to filter out certain namespaces in a diff

I think we had this discussion already for some functions during CoW, and there was talk about disallowing cascading keys in keysets in the future. So from that, I think it would probably be best to go with option 1, and just ignore cascading keys when calculating the diff.

atmaxinger:

atmaxinger:
@markus2330 @flo91 this is ready from my POV. Today I added some benchmark results to the Highlights section of the release notes and put the test-specific functions into kdb-private.

doc/news/_preparation_next_release.md
```diff
@@ -44,11 +44,46 @@ docker run -it elektra/elektra
...
+We ran some memory benchmarks and found a slightly increased memory usage on a stock instance of Elektra.
+If you're using many plugins that do changetracking, the overhead will decrease.
+
+| Number of Keys | Old Implementation (bytes) | New Implementation (bytes) | Memory Increase (%) |
+| -------------: | -------------------------: | -------------------------: | ------------------: |
+|             50 |               225792 bytes |               229580 bytes |              1,68 % |
+|            500 |               383180 bytes |               411238 bytes |              7,32 % |
+|           5000 |              1992294 bytes |              2306867 bytes |             15,79 % |
+|          50000 |             18245222 bytes |             21181235 bytes |             16,09 % |
+|         500000 |            178782208 bytes |            207827763 bytes |             16,25 % |
```
markus2330, line 72: Interesting non-linearity. In later research you might look into it more closely.

doc/tutorials/changetracking.md
```diff
@@ -0,0 +1,112 @@
+# Changetracking
+
+Elektra provides developers of applications and plugins a way to determine changes made to a `KeySet` relative to the last known state of the key database.
+This can be useful if you are writing a plugin that is only supposed to working with changed data.
```
markus2330, line 4:
```suggestion
This can be useful if you are writing a plugin that works with changes of the configuration.
```

doc/tutorials/changetracking.md
```diff
@@ -0,0 +1,112 @@
...
+
+```c
+KeySet * originalKeySet;
+KeySet * modifiedKeySet;
+Key * parentKey;
+
+// Calculate the difference of the keys below and including parentKey
+ElektraDiff * diff = elektraDiffCalculate (modifiedKeySet, originalKeySet, parentKey);
+
+// Extract useful information, see section 'Working with the diff'
```
markus2330, line 25: Forward references should be avoided in scientific texts but okay, to keep the funny style here, simply write something like "Don't be impatient, we will soon look into ..." :wink:

doc/tutorials/changetracking.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Getting the changes as an application developer
+
+If you are elektrifying your app, you can use `elektraChangeTrackingGetContextFromKdb` to get the current `ChangeTrackingContext` for your KDB instance.
+Then, use `elektraChangeTrackingCalculateDiff` to calculate the changes to the KDB.
+
+```c
+KDB * kdb;
+
+ChangeTrackingContext * ctx = elektraChangeTrackingGetContextFromPlugin (kdb);
```
markus2330, line 56: I assume you meant:

```suggestion
ChangeTrackingContext * ctx = elektraChangeTrackingGetContextFromKdb (kdb);
```

Please create an example in examples/ and see if it really compiles.

doc/news/_preparation_next_release.md
```diff
@@ -44,11 +44,46 @@ docker run -it elektra/elektra
...
 
-### <<HIGHLIGHT>>
+### New Changetracking API
+
+We've created a new KeySet diffing and changetracking API for both internal and external use.
+Several plugins already had their own implementations to detect changed made to the key database.
+Unfortunately, every implementation did something different and results varied.
+With this new release every plugin now uses the same shared implementation!
+
+If you want to use the new API yourself, take a look at [the tutorial](../tutorials/changetracking.md)!
```
markus2330, line 58:
```suggestion
To try the amazing new API yourself, take a look at [the tutorial](../tutorials/changetracking.md)!
```

doc/tutorials/changetracking.md
```diff
@@ -0,0 +1,112 @@
...
+
+ksDel (addedKeys);
+ksDel (removedKeys);
+ksDel (modifiedKeys);
+```
+
+Use these methods to get information about single keys:
+
+```c
+Key * myKey;
```
markus2330, line 91: It is unclear where myKey comes from or what it might contain.

src/include/kdbdiff.h
```diff
@@ -0,0 +1,43 @@
...
+uint16_t elektraDiffDecRef (ElektraDiff * ksd);
+uint16_t elektraDiffGetRef (ElektraDiff * ksd);
+
+bool elektraDiffIsEmpty (const ElektraDiff * ksd);
+
+const Key * elektraDiffGetParentKey (const ElektraDiff * ksd);
+
+KeySet * elektraDiffGetAddedKeys (const ElektraDiff * ksd);
+KeySet * elektraDiffGetRemovedKeys (const ElektraDiff * ksd);
+KeySet * elektraDiffGetModifiedKeys (const ElektraDiff * ksd); // Returns old keys (pre-modification)
```
markus2330, line 29: People should read API documentation instead:

```suggestion
KeySet * elektraDiffGetModifiedKeys (const ElektraDiff * ksd);
```

src/include/kdbos.h
```diff
@@ -65,6 +65,15 @@
 #define ELEKTRA_NOINLINE
 #endif
 
+#ifdef __GNUC__
+#define ELEKTRA_WEAK __attribute__ ((weak))
+#endif
+
+#ifdef __clang__
```
markus2330, line 72:
```suggestion
#if defined(__GNUC__) || defined(__clang__)
```

markus2330 (changes requested):
Amazing job! :1st_place_medal: :partying_face:

doc/news/_preparation_next_release.md
```diff
@@ -44,11 +44,46 @@ docker run -it elektra/elektra
...
+
+Elektra's internal code now also uses this new API to detect which backends to execute.
+This can lead to better performance in I/O bound cases, where previously certain keys would have been detected as changed when they weren't.
+
+We ran some memory benchmarks and found a slightly increased memory usage on a stock instance of Elektra.
+If you're using many plugins that do changetracking, the overhead will decrease.
+
+| Number of Keys | Old Implementation (bytes) | New Implementation (bytes) | Memory Increase (%) |
```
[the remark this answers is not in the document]
atmaxinger: Yes!

doc/news/_preparation_next_release.md
atmaxinger: Yeah I've looked very closely into it. It's because at a certain number of keys they become the dominant factor of memory consumption. So at 50 keys the overall memory consumption of Elektra is not really affected by the keys. But at > 5000 the other memory consumption doesn't matter much.

doc/tutorials/changetracking.md
atmaxinger: Good that it's not a scientific text then 😃

doc/tutorials/changetracking.md
atmaxinger: I added a comment to clarify now. It is the key that we want to check. So for example, if the key has been modified, or what the old metadata of the key looked like.

markus2330 (approved):
Great job! The capitalization of titles (in the tutorial) is untypical, usually longer names get capitalized.

@flo91 please also have a look and merge if it is fine.

flo91 (changes requested):
Outstanding work and progress! :star2:
I've just found some small errors regarding missing cleanup or documentation.
After you've fixed them, I think we can merge this. :satisfied:

atmaxinger:
@flo91 I've fixed all the errors you've pointed out. I think you can merge it then.

flo91 (approved):

PR #4853, opened 2023-02-28
Proposals for new API

kodebach:
This PR has 3 variants of the potential new API. The basic ideas are the same, but the differ in how they use public structs.


doc/contrib/new_api/byref/builders/public.h
```diff
@@ -0,0 +1,69 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#define ELEKTRA_KEYSET(...)                                                                                                                \
+	(elektraSetBuild ((ElektraEntry *[]){ __VA_ARGS__ }, sizeof ((ElektraEntry *[]){ __VA_ARGS__ }) / sizeof (ElektraEntry *)))
+
+#define ELEKTRA_KEYNAME(ns_, name_)                                                                                                        \
+	(ElektraName)                                                                                                                      \
+	{                                                                                                                                  \
+		.ns = (ns_), .name = (name_), .size = sizeof (name_)                                                                       \
```
atmaxinger, line 16: That `sizeof` will break as soon as someone passes a dynamically allocated string for `name_`.

kodebach: I know, one solution would be
```suggestion
		.ns = (ns_), .name = (name_), .size = sizeof ("" name_)                                                                       \
```

But the error messages wouldn't be particularly nice. It would probably be something like `expected ')'` and then list the macro expansion.

I just didn't want to clutter this proposal too much with obscure macro tricks ;)

atmaxinger:
So I haven't looked in too much detail yet. However, from a usability perspective, I really dislike the idea of the unescaped names in public API. I find it much harder to read the key names. 

So I would prefer the opaque struct, and provide suitable creator and getter setter methods to work with escaped names in "user" code. At the same time, internally we would only work with the unescaped name.

doc/contrib/new_api/opaque/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+For names we simply expose the opaque `ElektraNameCow` struct used internally as `ElektraName`.
+To that end, we introduce public APIs for creating and deleting a `ElektraName`.
+Because, `ElektraNameCow` is refcounted this can be done safely.
+
+Because everything is opaque, we have the most flexibility with this approach.
+Note, however, that at this moment there is no known case were a breaking change could be avoided with these opaque structs, but not with the ["by-value" variant](../byvalue/README.md)
+
+A clear disadvantage here is that a name cannot live on the stack.
+That means any API that might use temporary names (e.g., `elektraSetLookup`, or `elektraSetFindHierarchy`) is more difficult to use without memory leaks.
```
atmaxinger, line 19: Yes, but that's also the case for the `byref` variant or am I missing something?

kodebach: No, in `byref` the struct isn't opaque, its definition is public. You can do e.g.

```c
ElektraName name = { .ns = ELEKTRA_NS_USER, .name = "foo", .size = 4 };
elektraSetLookup (set, &name);
```

kodebach:
> I really dislike the idea of the unescaped names in public API. I find it much harder to read the key names.

That is true. However, the processing of escaped names isn't particularly cheap. The concept of escaped names is also entirely unnecessary inside `libelektra-core`. That's why the idea has been (for a while now) to remove escaped names from `libelektra-core` entirely, but provide a separate library to deal with escaped names. That separate library could be used by "user" code, when simpler code is preferred over performance.

For the `opaque` variant we could easily add:

```c
ElektraName * elektraNameBuild (ElektraNamespace ns, const char * part, ...) __attribute__ ((sentinel));
```

Which makes a call like this possible:

```c
elektraSetLookup (set, elektraNameBuild (ELEKTRA_NS_USER, "foo", "bar", "123", NULL));
```

For the other variants you could create a (limited) macro. See e.g. https://godbolt.org/z/c6ans6Wr3

Also it has to be noted that:
1. In many other more modern languages (e.g. C++, Rust) it's possible to build the unescaped name from (constant) name parts at compile time. In those cases you don't need the complicated macro tricks or a function that does `malloc`.
2. In such languages it might even be possible to unescape names at compile time.
3. Also most languages other than C, have proper string types which allow embedding `\0` characters. So in bindings the `ElektraName` type in `byref` and `byvalue` would just map to the native string type.

doc/contrib/new_api/byref/core/public.h
```diff
@@ -0,0 +1,127 @@
...
+ElektraReturnCode elektraSetName (ElektraEntry * key, const ElektraName * name);
+// increments nameLock
+void elektraEntryLockName (ElektraEntry * key);
+// decrements nameLock
+void elektraEntryUnlockName (ElektraEntry * key);
+bool elektraEntryIsNameLocked (const ElektraEntry * key);
+
+const ElektraValue * elektraEntryGetValue (const ElektraEntry * key);
+// Note: creates copy of value
+ElektraReturnCode elektraSetValue (ElektraEntry * key, const ElektraValue * value);
```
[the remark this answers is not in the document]
kodebach: > Is there a reason why the set-functions don't have Entry in their name?

Just a copy-paste error

doc/contrib/new_api/README.md
```diff
@@ -0,0 +1,118 @@
...
+For completeness, here are some more example of unsafe and illegal operations.
+These operations compile and "work" without reporting errors at runtime, but will cause various problems.
+An "illegal" operation, immediately break something and will cause wrong results, segfaults, etc. down the line.
+An "unsafe" operation, does not immediately break anything, but it is still an incorrect use of the API and often creates new "unsafe" or even "illegal" operations for the future.
+
+```c
+ElektraEntry * k1 = elektraEntryNew (&(ElektraName){
+    .ns = ELEKTRA_NS_SYSTEM,
+    .name = "foo\0bar\0baz",
+    .size = 14
```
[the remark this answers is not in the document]
kodebach: Yup you're right, I probably highlighted the text in the editor and just blindly used the number of selected characters.

doc/contrib/new_api/opaque/names/examples.c
```diff
@@ -0,0 +1,65 @@
...
+
+	ElektraNameBuffer * buffer2 = elektraNameBufferNew ();
+	elektraNameBufferAppend (buffer2, "foo\0bar", 9);
+	elektraNameBufferAppendPart (buffer2, "part1");
+	elektraNameBufferAppendPart (buffer2, "part2");
+	elektraNameBufferAppendPart (buffer2, "part3");
+	elektraNameBufferRemoveLastPart (buffer2);
+	elektraNameBufferReplaceLastPart (buffer2, "part11");
+
+	ElektraEntry * key2 = elektraEntryNew (elektraNameBufferFinish (buffer2));
```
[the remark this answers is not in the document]
kodebach: Another copy-paste error, I forgot to add that line

lawli3t:
From a usability POV (which is probably where I can give useful input) I would strongly prefer either **byRef** or **byValue**, since those would imo be the easiest to use. The functions would clearly state that they expect a name / value, and the purpose of the arguments would be clear. Additionally, this introduces safeguards against misusing functions and reduces the need for validation logic in every single function leading to less verbose code in every function afaict.

For the intricacies of picking **byRef** or **byValue**, I think others are more qualified to judge the technical implications of those solutions.

markus2330:
Please don't masquerade decisions in different way. It is completely unclear which problem you want to solve with having the names on the stack.

> I hope you can provide an example where opaque works better than byvalue

An answer would get lost here. Please provide a proper decision, there we can add the arguments.

Such an PR is only useful for an overview how the API would look like **based on already decided decisions** with variants on points **which are technically equal** (like different names for the same thing).

PR #4940, opened 2023-05-22
Move headers Part 2

kodebach:
> **Warning**: This PR is only a preview for what I've planned after #4920 and should be treated accordingly. I will not watch the CI results or fix any errors. Please do not review for now.

## What happens in this PR?

Continuing from #4920, this PR includes further progress on the decisions mentioned in #4920. 

- 710728f73a36f6124dedd31dbdca1195a092da04 splits the `old_helper.h` header into more modular units [Link](https://github.com/ElektraInitiative/libelektra/pull/XXXX/files/710728f73a36f6124dedd31dbdca1195a092da04)
- f6cc86fb32d45be8e39058c4701c4cc88c2e6cf9 is the only automated commit in this PR, it updates the files that used `old_helper.h` [Link](https://github.com/ElektraInitiative/libelektra/pull/4920/files/f6cc86fb32d45be8e39058c4701c4cc88c2e6cf9)
- 122790d1508938ffc4232d7f31bb5c51a8a76dd1 and 1565b56200a5dba72337d635abed3fee8e3ae8a6 are the formatting fixes and manual updates necessary after the automated commit [Link](https://github.com/ElektraInitiative/libelektra/pull/4920/files/122790d1508938ffc4232d7f31bb5c51a8a76dd1..1565b56200a5dba72337d635abed3fee8e3ae8a6)
- 8939ff8d3c81ea302e692fa92adaa5a42bc8075b until e1780acb17c0a35cc3e0e5b9df36e5485bec453a were done manually. They split `kdbprivate.h` into more modular units. I did this manually, because I wanted to verify where `kdbprivate.h` is used and why. Partly to ensure there isn't some place that shouldn't use it and partly to inform how `kdbprivate.h` should be split up. [Link](https://github.com/ElektraInitiative/libelektra/pull/4920/files/8939ff8d3c81ea302e692fa92adaa5a42bc8075b..e1780acb17c0a35cc3e0e5b9df36e5485bec453a)
- Allow following commits are additional fixes required to make things actually work

### Hints for reviewers

As with #4920 this PR has a large number of changed files. Most of them once again are only updated `#include`s. When #4920 is merged and this PR is ready for review, I will again provide a list of files with changes other than `#include`s.

## Basics

- [ ] Short descriptions of your changes are in the release notes
      (added as entry in `doc/news/_preparation_next_release.md` which contains `_(my name)_`)
      **Please always add them to the release notes.**
- [ ] Details of what you changed are in commit messages
[...]

PR #4920, opened 2023-05-03
Move headers Part 1

kodebach:
> **Note** The release notes currently don't  mention the moved/changed headers. I'll write a highlight section in one of the follow up PRs. 

This PR **starts** implementing various decisions related to headers and repo structure. It is hard to define which decisions are implemented in this PR, because the changes will be split across multiple PRs for ease of reviewing.

The entire series of PRs, will implement:

- `doc/decisions/4_decided/header_include.md`
- `doc/decisions/4_decided/header_file_structure.md`
- `doc/decisions/4_decided/library_directory_structure.md`
- `doc/decisions/4_decided/library_split.md`

## What happens in this PR?

This PR is very big so here is an overview what happens by commit (range):

1. 0db575f95f7cdf6ba1d8f537c588f86f23cb7ae9 adds the script `scripts/dev/check-includes.sh`. This script checks the new rules for includes and is also executed in a new Github Actions workflow. [Link](https://github.com/ElektraInitiative/libelektra/pull/4920/files/0db575f95f7cdf6ba1d8f537c588f86f23cb7ae9)
3. 496d1ad455ffa3d5fbbd07916471bb2fb8c72644 until d226f6be27672c44bd4d754dba2c6c7fac9dffcb move the existing headers to their new locations. Commits fall into one of these groups: (1) moving/splitting headers (2) automatic refactoring (3) reformatting (4) manual fixes of things missed by automatic script [Link](https://github.com/ElektraInitiative/libelektra/pull/4920/files/496d1ad455ffa3d5fbbd07916471bb2fb8c72644..d226f6be27672c44bd4d754dba2c6c7fac9dffcb)
4. efa1537653b00485b56592997cbe5a05fa836483 until 0a65e50bce37508943e97fd1d1841fdbcae13aa0 fix various problems resulting from the previous commits. **Note**: fd1280afb35f0bd2757be5081c4c9fe170e7f9dd is a simple regex replace (using the regex from `check-includes.sh`), which adds the now required `./` to `#include "..."` lines. [Link](https://github.com/ElektraInitiative/libelektra/pull/4920/files/efa1537653b00485b56592997cbe5a05fa836483..0a65e50bce37508943e97fd1d1841fdbcae13aa0)
[...]

PR #4941, opened 2023-05-22
Move headers Part 3

kodebach:
## What happens in this PR?

With #4940 the pure header updates for should be concluded. However, the decisions listed in #4920 also include some additional changes like new libraries.

This PR includes many CMake changes and finally splits `src/libs/elektra` properly into `src/libs/core` and `src/libs/kdb`.

Among the CMake changes, is a change in how `BUILD_FULL` and `BUILD_STATIC` work. Instead of collecting all the source and header files and compiling them again, we just collect all the CMake targets and use the `$<TARGET_OBJECTS:...>` syntax to add everything as dependencies.

The PR also introduces two new libraries `src/libs/base` and `src/libs/pluginload`. These are not actual libraries that can be installed. Instead, as described in the decisions, they are `OBJECT` libraries in CMake, meaning the `.c` files are compiled into `.o` files but nothing more. We do this, so we can extract base functionality (like `elektraMalloc`) and plugin loading from `elektra-core` and `elektra-kdb`, without introducing additional dependencies.

The PR also introduces a new `src/libs/type`, which is just a bit of `src/libs/ease` extracted into a separate library.

Finally, this PR has a change that was hinted at in https://github.com/ElektraInitiative/libelektra/pull/4940#issuecomment-1558002419. It turns quite a few functions which are just small wrappers around stdlib functions (`elektraMalloc`, `elektraFree`, etc.) into `static inline` functions in **internal** headers. The way this is done is a bit quirky, but it works and allowed the compiler to do more in-depth analysis (see 679085bc80f95a0b5bfd581278bc81591ee9e3a2).

### Hints for reviewers

The PR is a bit bigger, because there are many (similar) changes to CMake files, but otherwise it should be like any other PR in terms of reviews. All commits where done manually.

Most of the CMake changes are just removing `add_headers` and other stuff that isn't actually needed. If you don't understand something, just ask.

## Basics

  - make all unit tests pass
  - fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


kodebach:
As a preview: After this PR, I will tackle `doc/decisions/4_decided/elektra_prefix.md` with one PR per library (i.e. per `src/libs/*` folder in this PR). After that PR-series, the "gigantic but simplistic" PRs should be done and further PRs will be focused on updating a one or two functions at a time. (e.g. removing the automatic `meta:/` from `keyGetMeta` and `keySetMeta`).

kodebach:
@atmaxinger @hannes99 This PR might take a bit more time to review. The simple `#include` changes are all done and this PR is more focused on the library split and CMake updates. However, the CMake updates are all pretty similar, so it should still be quicker to review than 300 changed files might make it appear.

doc/news/_preparation_next_release.md
```diff
@@ -48,7 +48,52 @@ docker run -it elektra/elektra
...
+// almost all libraries provide such a header
+#include <elektra/[LIB].h>
+
+// include only foo symbols of library libelektra-[LIB].so
+// not all libraries provide modularized headers like this
+#include <elektra/[LIB]/foo.h>
+```
+
+(\*) For now, some bindings (e.g., the C++ binding) still install additional headers located in other places.
+These will also be moved to `src/include/elektra` in the future.
```
kodebach, line 83: We forgot to discuss bindings in the decision. IMO their headers should be moved to `src/include/elektra` too, so everything is in one place and we can actually install headers by simply copying a single directory. That would eliminate the risk of accidentally having stuff on the include path during build, but not after install.

However, since it wasn't discussed, I didn't move the headers for now.

kodebach:
Okay, something went wrong in one of the rebases. I'll have to investigate the CI problems further. Probably a small CMake issue somewhere...

src/include/internal/utility/alloc_impl.h
```diff
@@ -0,0 +1,194 @@
...
+	{
+		// TODO: is this safe? couldn't the old buffer already be destroyed?
+		*buffer = svr; /* restore old buffer*/
+		return -1;
+	}
+	else
+	{
+		*buffer = ptr;
+		return 0;
+	}
```
hannes99, line 67:
```suggestion
	ptr = realloc (*buffer, size);
        if (ptr == NULL)
        {
            return -1;
        }
        *buffer = ptr;
        return 0;
```
AFAIK the behavior of realloc is that it only frees the old memory if it successfully allocates the new memory. [1] So, there's no need to revert `*buffer` back to `svr`. Instead, you can just return -1 directly.

[1] https://linux.die.net/man/3/realloc

hannes99 (approved):
other than this small thing LGTM

src/include/internal/utility/alloc_impl.h
kodebach: I actually just added the `TODO`, because I found the code odd. The rest is copied from the old `internal.c`. But you're right, the `svr` stuff is completely unnecessary. Not just because of how `realloc` works, but simply because `realloc` has no way of changing the value of `*buffer`. It could only change the data in `**buffer`.

However, your suggestion would make `svr` an unused variable, so I won't use it directly.

kodebach, line 67:
```suggestion
	ELEKTRA_ASSERT (size, "Size to allocate is zero (implementation defined behavior)");

	void * ptr = realloc (*buffer, size);
	ELEKTRA_ASSERT (ptr, "Memory (re)allocation failed with size %zu", size);
	if (ptr == NULL)
	{
		return -1;
	}

	*buffer = ptr;
	return 0;
```

Fix for https://github.com/ElektraInitiative/libelektra/pull/4941#discussion_r1215376240

kodebach:

kodebach:

kodebach:

kodebach:
@atmaxinger @flo91 Could one of you take a look at this PR?

src/libs/kdb/symbols.map
```diff
@@ -0,0 +1,53 @@
...
+	elektraGOptsContract;
+	elektraGOptsContractFromStrings;
+
+	# FIXME: should be separate lib
+	# kdbchangetracking.h
+	elektraChangeTrackingGetContextFromKdb;
+	elektraChangeTrackingGetContextFromPlugin;
+	elektraChangeTrackingCalculateDiff;
+
+	# FIXME: separate lib? part of merge?
```
atmaxinger, line 26: I don't see why diff should be a part of merge TBH. A separate lib maybe, but on the other hand it's an integral part of KDB so IDK.

src/libs/utility/symbols.map
```diff
@@ -3,4 +3,24 @@ libelektra_0.8 {
...
+	elektraLog;
+	elektraCalloc;
+	elektraFormat;
+	elektraFree;
+	elektraMalloc;
+	elektraMemCaseCmp;
+	elektraMemDup;
+	elektraRealloc;
+	elektraStrCaseCmp;
+	elektraStrCmp;
```
atmaxinger, line 19: Maybe a bit off-topic, but why exactly do we need our own implementation of `strcmp`?

kodebach:

src/libs/kdb/symbols.map
kodebach: The change tracking part is integral to KDB, but the pure diff part could be used standalone (just with core), AFAIK. I thought of merge, since diff & merge are somewhat related concepts.

src/libs/utility/symbols.map
kodebach: The `symbols.map` files actually need to be cleanup when all this restructuring is done. `elektraStrCmp` isn't actually public anymore. It's now just an internal `static inline` function with extra asserts around standard `strcmp`.

https://github.com/ElektraInitiative/libelektra/blob/3af13a20f5fb48d6a2e6180add245c82da357f7c/src/include/internal/utility/compare.h#L40-L45

This PR actually turns most of these "just a standard libc function with extra asserts" into internal `static inline` functions. For the ones that need to be public (`elektraMalloc`, etc.) we re-export them via `libelektra-utility` as normal functions too, but internally use the `static inline` versions. More complicated things like `elektraFormat` are still normal functions everywhere, but are linked statically (but unexported) into `libelektra-core` to avoid a dependency.

atmaxinger (approved):
Another great PR. 🚀

I left a few comments, but overall LGTM.
However, I'm not sure if it's a good idea to have the Latex package changes in this PR. I'd much rather see them fixed in master, and then rebase this branch to master again.

benchmarks/cmp.c
```diff
@@ -86,7 +86,7 @@ int main (void)
 	long long nrIterations = 100000000;
 	const char str1[] = "some string to be compared with/some\\/more with a long common part, and only a bit different";
 	char * str2 = elektraMalloc (sizeof (str1));
-	strcat (str2, str1);
+	strcpy (str2, str1);
```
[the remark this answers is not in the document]
kodebach: Because somehow it always worked ^^ But now that `elektraMalloc` is `static inline` gcc complains about this line. That's actually the only reason I noticed this

doc/decisions/2_solutions_clear/simplify_api.md
```diff
@@ -2,7 +2,7 @@
 
 ## Problem
 
-According to [symbols.map](/src/libs/elektra/symbols.map) the public core API has 124 symbols¹, which is arguable too much for a key-value API which has as highest goal simplicity.
+According to [symbols.map](/src/libs/core/symbols.map) the public core API has 124 symbols¹, which is arguable too much for a key-value API which has as highest goal simplicity.
```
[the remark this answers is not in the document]
kodebach: Going by the detailed list in the notes below it may be down to ~110, but I'm not actually sure.

kodebach:
> However, I'm not sure if it's a good idea to have the Latex package changes in this PR. I'd much rather see them fixed in master, and then rebase this branch to master again.

I cherry picked the commits (without change) from #4967, so if all goes well, they should disappear when this is eventually merged into master. 3ec05e11be35f28c62e565de559384e7e9ee4489 and f3b783b8907ef2baa28741c056a920b7364022e7 might need to be applied to `master` too, I opened #4969.

kodebach:

PR #4943, opened 2023-05-23
Kdb rewrite base

hannes99:
This is the "framework" for the new `C` cli.
I'll mention PRs for commands later here.

This can be merged as is, more commands can be added later.

1. ~add spec support for external commands~: #4951 (reviewed and now in `kdb_rewrite_base`)
2. ~add spec for `C++` commands~: #4949 (reviewed and now in `kdb_rewrite_base`)
3. ~add set command~: #4954 (reviewed and now in `kdb_rewrite_base`)
4. add meta commands: #4955 
5. this one

## Basics

  - make all unit tests pass
  - fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


src/tools/kdb/colors.c
```diff
@@ -0,0 +1,26 @@
...
+
+#include <colors.h>
+#include <kdberrors.h>
+#include <stdio.h>
+
+const char * formatString (char ** buffer, int colorMode, const char * escapeCode, const char * text)
+{
+	if (colorMode == CLI_COLOR_NEVER) return text;
+
+	size_t len = elektraStrLen (escapeCode) + elektraStrLen (text) + elektraStrLen (RESET) + 1;
```
kodebach, line 17: You don't need the `+ 1` at the end, `elektraStrLen` already includes the zero terminator (i.e. `elektraStrLen (X) == strlen (X) + 1`).

src/tools/kdb/get.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+int execGet (KeySet * options, Key * errorKey);
+
+
+// helper functions for the get command
+const char * getCascadingName (const char * str);
+Key * printTrace (KeySet * ks, Key * key, Key * found, elektraLookupFlags options);
+Key * warnOnMeta (KeySet * ks, Key * key, Key * found, elektraLookupFlags options);
+void printOptions (elektraLookupFlags options);
+void setCallback (Key * key, Key * (*f) (KeySet * ks, Key * key, Key * found, elektraLookupFlags flags));
```
kodebach, line 40: If they're helper functions why declare them here? Are the used outside `get.c` too? If not just declare them in `get.c` as `static` functions.

src/tools/kdb/main.c
```diff
@@ -0,0 +1,181 @@
...
+			keyDel (parentKey);
+			parentKey = keyNew (CLI_BASE_KEY, KEY_END);
+			result = subcommands[i].exec (options, parentKey);
+			printError (parentKey);
+			printWarnings (parentKey);
+			goto cleanup;
+		}
+	}
+cpp:
+	result = cpp_main (argc, argv);
```
kodebach, line 175: It would be a bit nicer, if you explicitly defined the subcommands that are implemented in C++ and only delegated to C++, if one of those subcommands was used. Ideally directly to the subcommand, but since the `argv` parsing is different you may still have to call `cpp_main`.

Then if the command is also not in the list of C++ commands, you could do the `tryExternalCommand` stuff from C++ (but translated to C).

If that also fails to find a command, report an error.

src/tools/kdb/get.c
```diff
@@ -0,0 +1,225 @@
...
+	}
+	keySetMeta (key, "callback/print_trace/last_key_name", name);
+	return found;
+}
+
+void setCallback (Key * key, Key * (*f) (KeySet * ks, Key * key, Key * found, elektraLookupFlags flags))
+{
+	union
+	{
+		Key * (*f) (KeySet * ks, Key * key, Key * found, elektraLookupFlags flags);
```
kodebach, line 218:
```suggestion
		lookup_fn f;
```

Please use a `typedef` for this ungodly type:

```c
typedef Key * (*lookup_fn) (KeySet * ks, Key * key, Key * found, elektraLookupFlags flags);
```

src/tools/kdb/get.c
```diff
@@ -0,0 +1,225 @@
...
+			ELEKTRA_ADD_RESOURCE_WARNINGF (
+				key, "%s is context dependent, shown result might be wrong, -v shows you the trace to the key\n",
+				keyName (found));
+		}
+	}
+	return found;
+}
+
+
+Key * printTrace (ELEKTRA_UNUSED KeySet * ks, Key * key, Key * found, elektraLookupFlags options)
```
kodebach, line 173: Why add `KeySet * ks` as an argument, if it's not needed?

src/tools/kdb/get.c
```diff
@@ -0,0 +1,225 @@
...
+}
+
+const char * getCascadingName (const char * str)
+{
+	if (str == NULL) return "/";
+	const char * r = strchr (str, '/');
+	return r == NULL ? "/" : r;
+}
+
+Key * warnOnMeta (ELEKTRA_UNUSED KeySet * ks, ELEKTRA_UNUSED Key * key, Key * found, elektraLookupFlags options)
```
kodebach, line 157: Why add `KeySet * ks` as an argument, if it's not needed?

src/tools/kdb/cpp-main.h
```diff
@@ -0,0 +1,9 @@
+#ifdef __cplusplus
```
kodebach, line 1: Missing include guard

src/tools/kdb/command.h
```diff
@@ -0,0 +1,204 @@
...
+		{                                                                                                                          \
+			perror ("timeout");                                                                                                \
+			return -1;                                                                                                         \
+		}                                                                                                                          \
+		sleep (1);                                                                                                                 \
+	}
+
+// only print if we are at least as 'verbose' as minLogLevel
+#define CLI_PRINT(minLogLevel, fmt, ...) cliPrint (logLevel, minLogLevel, fmt, __VA_ARGS__)
+#define CLI_ERROR_PRINT(minLogLevel, fmt, ...) cliErrorPrint (logLevel, minLogLevel, fmt, __VA_ARGS__)
```
kodebach, line 147: Not a fan of these macros, the only thing they avoid is having to pass `logLevel` explicitly. However, exactly that will cause weird errors, if the function is used where `logLevel` is undefined or defined differently.

Based on the implementation of `cliPrint` and `cliErrorPrint` it would actually make more sense to define the macros as:

```c
#define CLI_PRINT(minLogLevel, fmt, ...) if (logLevel < minLogLevel) { printf (fmt, __VA_ARGS__); }
#define CLI_ERROR_PRINT(minLogLevel, fmt, ...) if (logLevel < minLogLevel) { fprintf (stderr, fmt, __VA_ARGS__); }
```

Then you at least avoid the `va_list` stuff.

But I'd still prefer using the functions directly for a bit of type-safety

src/tools/kdb/command.c
```diff
@@ -0,0 +1,112 @@
...
+	const char * ret = NULL;
+	if (name == NULL)
+	{
+		return ret;
+	}
+
+	KeySet * bookmarks = ksNew (0, KS_END);
+	Key * bookmarkBase = keyNew (CLI_BASE_KEY "/bookmarks", KEY_END);
+	KDB * handle = kdbOpen (NULL, bookmarkBase);
+	kdbGet (handle, bookmarks, bookmarkBase);
```
kodebach, line 27: I don't know how often this function is actually called. But this hidden `kdbGet` seems problematic. Especially since it always uses the same parent key. It would better if `kdb` loaded the bookmarks once at the start and then simply used the existing `KeySet`.

Also since `bookmarks` is only used in one of the branches below, please move the whole KDB stuff to that branch.

Or just do

```c
	if (name[0] != '+')
	{
		return elektraStrDup (name);
	}
```

before you start the bookmarks stuff.

src/tools/kdb/colors.h
```diff
@@ -0,0 +1,72 @@
...
+#define BG_RED(text) FORMAT_TEXT (BG_COLOR_RED, text)
+#define BG_GREEN(text) FORMAT_TEXT (BG_COLOR_GREEN, text)
+#define BG_YELLOW(text) FORMAT_TEXT (BG_COLOR_YELLOW, text)
+#define BG_BLUE(text) FORMAT_TEXT (BG_COLOR_BLUE, text)
+#define BG_MAGENTA(text) FORMAT_TEXT (BG_COLOR_MAGENTA, text)
+#define BG_CYAN(text) FORMAT_TEXT (BG_COLOR_CYAN, text)
+#define BG_WHITE(text) FORMAT_TEXT (BG_COLOR_WHITE, text)
+
+#define BOLD(text) FORMAT_TEXT (FORMAT_BOLD, text)
+#define UNDERSCORE(text) FORMAT_TEXT (FORMAT_UNDERSCORE, text)
```
kodebach, line 61: The hidden assumptions of macros like this should at least be documented. But I really don't see why

```c
BLACK ("some text")
```

is so much better than

```c
formatString (&fmtBuffer, colorMode, COLOR_BLACK, text);
```

especially since it makes this seem reasonable

```c
printf ("%s: %s\n", RED("ERROR"), errorMessage);
```

when this would do the same without the allocation

```c
if (colorMode != CLI_COLOR_NEVER) {
    printf("%s%s%s", COLOR_RED, "ERROR", RESET);
} else {
    printf("%s", "ERROR");
}
printf (": %s\n", errorMessage);
```

which could be improved with a different macro

```c
#define COLOR_PRINT(color, text) if (colorMode != CLI_COLOR_NEVER) { printf("%s%s%s", color, text, RESET); } else { printf ("%s", text); }

COLOR_PRINT (COLOR_RED, "ERROR");
printf (": %s\n", errorMessage);
```

kodebach (changes requested):
I'm not a fan of using macros when you don't need to. I left some comments about that, but IMO we could also merge this with the macros as they are now, as long as they are better documented (especially hidden assumptions).

The reason looks all very good. I just have some minor suggestions.

I also had an idea how to improve the integration of the old C++ code, but again we can also leave it the way it is now, as long as it's documented properly. Specifically, about the external shell script commands, the docs should state whether that is supported only via C++ or even, if you replace `cpp_main` with a dummy so you can compile without C++.

src/tools/kdb/command.c
hannes99: Moved it into branch. This is called in a few commands, and I think in one or two twice.

src/tools/kdb/command.c
```diff
@@ -0,0 +1,112 @@
...
+ * @copyright BSD License (see LICENSE.md or https://www.libelektra.org)
+ */
+
+#include <command.h>
+
+#include <kdberrors.h>
+#include <stdio.h>
+#include <string.h>
+
+const char * expandKeyName (const char * name, bool * resolved)
```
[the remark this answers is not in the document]
hannes99: Make a lot of sense, done.

src/tools/kdb/command.h
```diff
@@ -0,0 +1,204 @@
...
+			     KEY_META, "opt", "n", KEY_META, "opt/long", "no-newline", KEY_META, "opt/arg", "none", KEY_END));             \
+	ksAppendKey (spec, keyNew (COMMAND_SPEC_KEY (COMMAND_NAME) "/nullterm", KEY_META, "description", "Use binary 0 termination.",      \
+				   KEY_META, "opt", "0", KEY_META, "opt/long", "null", KEY_META, "opt/arg", "none", KEY_END));
+
+#define GET_OPT_KEY(options, key) ksLookupByName (options, key, 0)
+#define GET_OPT(options, key) keyString (GET_OPT_KEY (options, key))
+
+#define OR(value, def)                                                                                                                     \
+	void * tmp = value;                                                                                                                \
+	tmp == NULL ? (def) : tmp
```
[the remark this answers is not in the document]
hannes99: leftover, no reason. removed it

src/tools/kdb/get.c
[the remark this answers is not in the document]
hannes99: done

src/tools/kdb/main.c
hannes99: I'll add that in a different PR

src/tools/kdb/get.h
hannes99: done

src/tools/kdb/colors.c
hannes99: done

src/tools/kdb/command.h
hannes99: All of these macros are supposed to be used in the context of a command, so after GET_BASIC_OPTIONS. The idea was that when implementing a command you just have to think about the command and what it should do, so you don't have to think about the context of a CLI.

src/tools/kdb/command.h
```diff
@@ -0,0 +1,204 @@
...
+				       KEY_META, "opt", "C", KEY_META, "opt/arg/help", "WHEN", KEY_META, "opt/long", "color", KEY_META,    \
+				       "opt/arg", "required", KEY_END));                                                                   \
+	ksAppendKey (baseSpec,                                                                                                             \
+		     keyNew (baseKeyName "/nonewline", KEY_META, "description", "Suppress the newline at the end of the output.",          \
+			     KEY_META, "opt", "n", KEY_META, "opt/long", "no-newline", KEY_META, "opt/arg", "none", KEY_END));             \
+	ksAppendKey (spec, keyNew (COMMAND_SPEC_KEY (COMMAND_NAME) "/nullterm", KEY_META, "description", "Use binary 0 termination.",      \
+				   KEY_META, "opt", "0", KEY_META, "opt/long", "null", KEY_META, "opt/arg", "none", KEY_END));
+
+#define GET_OPT_KEY(options, key) ksLookupByName (options, key, 0)
+#define GET_OPT(options, key) keyString (GET_OPT_KEY (options, key))
```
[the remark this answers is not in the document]
hannes99: I guess, this felt easier to read.

src/tools/kdb/command.h
```diff
@@ -0,0 +1,204 @@
...
+
+#define OR(value, def)                                                                                                                     \
+	void * tmp = value;                                                                                                                \
+	tmp == NULL ? (def) : tmp
+#define HAS_ERR(errorKey) keyGetMeta (errorKey, "error/reason") != NULL
+#define HAS_ERR_CODE(errorKey) keyGetMeta (errorKey, "error/number") != NULL
+#define GET_ERR(errorKey) keyString (keyGetMeta (errorKey, "error/reason"))
+#define GET_ERR_CODE(errorKey) keyString (keyGetMeta (errorKey, "error/number"))
+#define GET_ERR_DESC(errorKey) keyString (keyGetMeta (errorKey, "error/description"))
```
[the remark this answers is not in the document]
hannes99: I guess, this felt easier to read.

src/tools/kdb/get.c
hannes99: The signature has to look like this for setCallback

src/tools/kdb/get.c
hannes99: The signature has to look like this for setCallback

src/tools/kdb/get.c
hannes99: done

hannes99:
Thanks a lot for the feedback!

doc/man/man1/kdb-get.1
```diff
@@ -1,6 +1,6 @@
 .\" generated with Ronn-NG/v0.10.1
 .\" http://github.com/apjanke/ronn-ng/tree/0.10.1.pre3
-.TH "KDB\-GET" "1" "August 2021" ""
+.TH "KDB\-GET" "1" "September 2022" ""
```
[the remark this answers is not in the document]
hannes99: I think it works now properly

src/tools/kdb/command.h
kodebach: Hm okay, I get the idea now... Maybe you could do something like this:


In `command.h`:
```c
typedef struct {
   int logLevel;
   char * fmtBuffer;
   // ...
} CliContext;

void loadCliContext (CliContext * ctx, KeySet * options);

static inline getOptionKey (KeySet * options, const char * name)
{
    return ksLookupByName (options, name, 0);
}

void cliPrint (const CliContext * ctx, int minLogLevel, const char * fmt, ...);
// ...
```

In `comand.h` file:
```c
// might need errorKey argument, and error return code
void loadCliContext (struct CliContext * ctx, KeySet * options) {
   // not quite right, but you get the idea
   bool debug;
   elektraKeyToBoolean (GET_OPTION_KEY (options, "debug"), &debug);
   bool verbose;
   elektraKeyToBoolean (GET_OPTION_KEY (options, "verbose"), &verbose);
   ctx->logLevel = verbose + debug;
   ctx->fmtBuffer = elektraMalloc(1024);
}
```

And used like this:

```c
int execGet (KeySet * options, Key * errorKey)
{
  CliContext ctx;
  loadCliContext (&ctx, options);

  // ...

  cliPrint (&ctx, CLI_LOG_VERBOSE, "The resulting keyname is %s\n", keyName (found));
}
```

Then you avoid all the issues of complex macros and still get the same "I don't care how CLI options work" style of code in the actual command implementations.

But for this PR, just leave it as is. It's just an idea, you could look into for the next PR (if you find the time).

src/tools/kdb/command.h
```diff
@@ -0,0 +1,177 @@
...
+	ksAppendKey (baseSpec, keyNew (baseKeyName "/color", KEY_META, "description", "Print never/auto(default)/always colored output.",  \
+				       KEY_META, "opt", "C", KEY_META, "opt/arg/help", "WHEN", KEY_META, "opt/long", "color", KEY_META,    \
+				       "opt/arg", "required", KEY_END));                                                                   \
+	ksAppendKey (baseSpec,                                                                                                             \
+		     keyNew (baseKeyName "/nonewline", KEY_META, "description", "Suppress the newline at the end of the output.",          \
+			     KEY_META, "opt", "n", KEY_META, "opt/long", "no-newline", KEY_META, "opt/arg", "none", KEY_END));             \
+	ksAppendKey (spec, keyNew (COMMAND_SPEC_KEY (COMMAND_NAME) "/nullterm", KEY_META, "description", "Use binary 0 termination.",      \
+				   KEY_META, "opt", "0", KEY_META, "opt/long", "null", KEY_META, "opt/arg", "none", KEY_END));
+
+#define GET_OPT_KEY(options, key) ksLookupByName (options, key, 0)
```
kodebach, line 37: Just noticed, I think this should be:

```suggestion
#define GET_OPT_KEY(options, key) ksLookupByName (options, CLI_BASE_KEY key, 0)
```

In which case use a macro makes sense, because otherwise you can't build the name like that...

It's used as e.g. `GET_OPTION_KEY (options, "debug")`, so unless I'm completely on the wrong path that would lead to `ksLookupByName (options, "debug", 0)` which always returns `NULL`, since `"debug"` is not a valid keyname.

Is there no test that verifies that CLI options (like `--debug`) actually work?

src/tools/kdb/command.h
```diff
@@ -0,0 +1,177 @@
...
+	{                                                                                                                                  \
+	}; /* disable unused variable warning  */                                                                                          \
+	if (!isatty (STDOUT_FILENO))                                                                                                       \
+	{                                                                                                                                  \
+		colorMode = CLI_COLOR_NEVER;                                                                                               \
+	}                                                                                                                                  \
+                                                                                                                                           \
+	int logLevel = verbose + debug;                                                                                                    \
```
kodebach, line 109: Is that really what we want? I.e. do want `kdb --debug` and `kdb --verbose` to behave the same?

The man page makes it sound like the options should be completely orthogonal, that would mean we need to separate `bool`s not a single `int LogLevel`.

`enum LOG_LEVEL` implies that `debug` should include more messages than `verbose`. That would be:

```c
// 1 + 1 if debug (verbose doesn't matter)
// 1 + 0 if verbose but not debug
// 0 + 0 if neither
int logLevel = (verbose || debug) + debug;
```

IMO all these variants are acceptable, it's just inconsistent and/or not properly documented right now.

kodebach (changes requested):
Noticed a two minor inconsistencies/mistakes, please check. Otherwise, LGTM.

src/tools/kdb/command.h
hannes99: By default `verbose` is set to whatever `debug` is, if `-v` should be set, `verbose` will be set to 1.

`-d`:
1. `debug = 1`
2. `verbose = debug`
3. _no `-v` so `verbose` stays_
-> `verbose` + `debug` = 2

`-v`:
1. `debug = 0`
2. `verbose = debug`
3. `-v` so `verbose = 1`
-> `verbose` + `debug` = 1

neither:
1. `debug = 0`
2. `verbose = debug`
3. _no `-v` so `verbose` stays_
-> `verbose` + `debug` = 0

This should work, no?

src/tools/kdb/command.h
hannes99: This macro expects the whole key name since it also needs the command name, `GET_OPT_KEY (options, COMMAND_BASE_KEY (COMMAND_NAME) "/" name)` is how it is used.

kodebach: In that case `GET_BASIC_OPTIONS` is wrong, because I copied the `"debug"` example from there.

Also, please add doxygen comments to the macros that are supposed to be used by command implementations. If there was a comment stating what should be passed as `key`, it would have been quite obvious where the mistake is.

src/tools/kdb/command.h
kodebach: Yes, you're right (technically). I missed that `verbose` is initialized with the value of `debug`. It effectively means that `verbose + debug == (verbose || debug) + debug`.

But only because of two things:
1. `opt/arg = none` means that the `proc:/` key for `--verbose` will always set to `1` or not touched by `elektraGetOpts` at all.
2. Currently only the `proc:/` keys created by `elektraGetOpts` are in `KeySet * options`.

If there was e.g. `user:/sw/elektra/kdb/#0/current/get/verbose` which, would be used as a fallback if `proc:/sw/elektra/kdb/#0/current/get/verbose` doesn't exist. If `user:/sw/elektra/kdb/#0/current/get/verbose` was explicitly set to `0`, then your code would produce `verbose = false`. If `debug = true`, it would mean 
```
verbose + debug = 0 + 1 = 1 = CLI_LOG_VERBOSE
```

So for now it's fine, but `(verbose || debug) + debug` works for the general case too.

hannes99: ohh, I see. You're right

src/tools/kdb/command.h
```diff
@@ -0,0 +1,177 @@
...
+#define GET_ERR(errorKey) keyString (keyGetMeta (errorKey, "error/reason"))
+#define GET_ERR_CODE(errorKey) keyString (keyGetMeta (errorKey, "error/number"))
+#define GET_ERR_DESC(errorKey) keyString (keyGetMeta (errorKey, "error/description"))
+#define COMMAND_BASE_KEY(name) CLI_BASE_KEY "/" name
+#define COMMAND_SPEC_KEY(name) "spec:" COMMAND_BASE_KEY (name)
+
+// only use in the context of a command/sub-command (options variable and GET_OPTION_KEY macro have to be in scope)
+#define GET_BASIC_OPTIONS                                                                                                                  \
+	bool debug = false;                                                                                                                \
+	Key * tmp = GET_OPTION_KEY (options, "debug");                                                                                     \
```
kodebach, line 51: Based on https://github.com/ElektraInitiative/libelektra/pull/4943/files#r1208288747, this line is wrong. But I'm not actually sure how fix it. You could do 

```c
Key * tmp = GET_OPTION_KEY (options, COMMAND_BASE_KEY (COMMAND_NAME) "debug");
```

and it should work, but it would make the macro even harder to use, because now the `COMMAND_NAME` macro needs to be defined. You could also add a `name` parameter to the macro and use

```c
Key * tmp = GET_OPTION_KEY (options, COMMAND_BASE_KEY (name) "debug");
```

which is a bit easier to use, but I actually think the proper solution is using:

```c
Key * tmp = GET_OPTION_KEY (options, CLI_BASE_KEY  "/debug");
```

**and** instead of using the `ADD_BASIC_OPTIONS` macro to add the `--debug`, `--verbose` etc. options to all commands, you add them to the base `KeySet * options` in `main()`. The difference would be:

```c
# works with old kdb, and with current PR, but not with change proposed above
kdb get --verbose

# works with old kdb, and with change proposed above, but not with current PR
kdb --verbose get
```

That's because when using sub-commands with `elektraGetOpts` the location of options like `--verbose` relative to command arguments matters (same as with e.g. `git`). Personally, I think the `kdb --verbose get` option actually makes more sense, since `--verbose` is an option that is not specific to `get`.

src/tools/kdb/main.c
```diff
@@ -0,0 +1,181 @@
...
+		}
+		if (elektraStrNCmp (errorMessage, "Unknown short option:", 21) == 0 ||
+		    elektraStrNCmp (errorMessage, "Unknown long option:", 20) == 0)
+		{
+			result = 1;
+		}
+		else
+		{
+			result = 5;
+		}
```
kodebach, line 148: I created #4953 to discuss a way of avoiding these ugly and delicate string comparisons.

src/tools/kdb/command.h
hannes99: I see why this is a little confusing, I'll document it better,
```c
#define GET_OPTION_KEY(options, name) GET_OPT_KEY (options, COMMAND_BASE_KEY (COMMAND_NAME) "/" name)
```
is defined in every command file, so `GET_OPTION_KEY` in `GET_BASIC_OPTIONS` is different in every command file. `command.h` doesn't define GET_OPTION_KEY.

hannes99:

src/tools/kdb/colors.h
```diff
@@ -0,0 +1,94 @@
...
+#define FORMAT_BOLD "\x1b[1m"
+#define FORMAT_UNDERSCORE "\x1b[4m"
+
+/**
+ * @def FORMAT_TEXT
+ * @brief Formats the text according to the given escape code. Used by the color/formatting macros below.
+ * @param escapeCode The escape code for the desired formatting.
+ * @param text The text to be formatted.
+ */
+#define FORMAT_TEXT(escapeCode, text) formatString (&fmtBuffer, colorMode, escapeCode, text)
```
kodebach, line 46:
```suggestion
/**
 * @def FORMAT_TEXT
 * @brief Formats the text according to the given escape code. Used by the color/formatting macros below.
 *
 * `char * fmtBuffer` and `int colorMode` must be defined in the calling context.
 * Normally, this is achieved by first using #GET_BASIC_OPTIONS
 *
 * @param escapeCode The escape code for the desired formatting.
 * @param text The text to be formatted.
 */
#define FORMAT_TEXT(escapeCode, text) formatString (&fmtBuffer, colorMode, escapeCode, text)
```

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+				       "opt/arg", "required", KEY_END));                                                                   \
+	ksAppendKey (baseSpec,                                                                                                             \
+		     keyNew (baseKeyName "/nonewline", KEY_META, "description", "Suppress the newline at the end of the output.",          \
+			     KEY_META, "opt", "n", KEY_META, "opt/long", "no-newline", KEY_META, "opt/arg", "none", KEY_END));             \
+	ksAppendKey (spec, keyNew (COMMAND_SPEC_KEY (COMMAND_NAME) "/nullterm", KEY_META, "description", "Use binary 0 termination.",      \
+				   KEY_META, "opt", "0", KEY_META, "opt/long", "null", KEY_META, "opt/arg", "none", KEY_END));
+
+/**
+ * @def GET_OPT_KEY
+ * @brief This macro retrieves the key with a given name from an options key set. This is the more general version that GET_OPTION_KEY in
```
kodebach, line 45:
```suggestion
 * @brief This macro retrieves the key with a given name from an options key set. This is the more general version that #GET_OPTION_KEY in
```

The `#` is needed so that doxygen generates a link to the other macro

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+ * command files should use.
+ * @param options The options KeySet.
+ * @param key The name of the key to retrieve.
+ */
+#define GET_OPT_KEY(options, key) ksLookupByName (options, key, 0)
+
+/**
+ * @def GET_OPT
+ * @brief This macro retrieves the string value of the key with a given name from an options key set. This is the more general version that
+ * GET_OPTION in command files should use.
```
kodebach, line 55:
```suggestion
 * #GET_OPTION in command files should use.
```

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * @def GET_ERR_DESC
+ * @brief This macro retrieves the string value of the error/description meta-key in the given key.
+ * @param errorKey The key to get the error/description from.
+ */
+#define GET_ERR_DESC(errorKey) keyString (keyGetMeta (errorKey, "error/description"))
+
+/**
+ * @def COMMAND_BASE_KEY
+ * @brief This macro generates the base key name for a given command. Used in command files for defining GET_OPTION_KEY and GET_OPTION.
```
kodebach, line 98:
```suggestion
 * @brief This macro generates the base key name for a given command. Used in command files for defining #GET_OPTION_KEY and #GET_OPTION.
```

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+	}                                                                                                                                  \
+                                                                                                                                           \
+	int logLevel = (verbose || debug) + debug;                                                                                         \
+	if (logLevel)                                                                                                                      \
+	{                                                                                                                                  \
+	}; /* disable unused variable warning  */
+
+/**
+ * @def RETURN
+ * @brief This macro frees the fmtBuffer and returns the given value. Use only after GET_BASIC_OPTIONS
```
kodebach, line 182:
```suggestion
 * @brief This macro frees the fmtBuffer and returns the given value. Use only after #GET_BASIC_OPTIONS
```

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+ */
+#define RETURN(c)                                                                                                                          \
+	elektraFree (fmtBuffer);                                                                                                           \
+	return c;
+
+/**
+ * @def CLI_PRINT
+ * @brief This macro prints a formatted message if the log level is at least as high as a given level.
+ * @param minLogLevel The minimum log level to print the message.
+ * @param fmt The format string for the message. Is setup by GET_BASIC_OPTIONS.
```
kodebach, line 193:
```suggestion
 * @param fmt The format string for the message. Is setup by #GET_BASIC_OPTIONS.
```

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+ */
+#define CLI_PRINT(minLogLevel, fmt, ...) cliPrint (logLevel, minLogLevel, fmt, __VA_ARGS__)
+
+/**
+ * @def CLI_ERROR_PRINT
+ * @brief This macro prints a formatted error message if the log level is at least as high as a given level.
+ * @param minLogLevel The minimum log level to print the message.
+ * @param fmt The format string for the message. Is setup by GET_BASIC_OPTIONS.
```
kodebach, line 201:
```suggestion
 * @param fmt The format string for the message. Is setup by #GET_BASIC_OPTIONS.
```

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+
+/**                                                                                                                                        \
+ * Expands a keyname if it contains a bookmark. If @name does not contain a bookmark ref a copy of @name is returned.                      \
+ *                                                                                                                                         \
+ * @param name the keyname that might contain a bookmark, and where the expanded name should be saved                                      \
+ * @param resolved will be set to true iff a bookmark was resolved successfully                                                            \
+ *                                                                                                                                         \
+ * @return NULL if the bookmark could not be resolved, NULL was passed as @ks or @name                                                     \
+ * @return key with expanded name, has to be freed after usage                                                                   \
+ */
```
kodebach, line 213:
```suggestion
/**
 * Expands a keyname if it contains a bookmark. If @name does not contain a bookmark ref a copy of @name is returned.
 * 
 * @param name the keyname that might contain a bookmark, and where the expanded name should be saved
 * @param resolved will be set to true iff a bookmark was resolved successfully
 * 
 * @return NULL if the bookmark could not be resolved, NULL was passed as @ks or @name
 * @return key with expanded name, has to be freed after usage 
 */
```

src/tools/kdb/command.h
```diff
@@ -0,0 +1,260 @@
...
+ * @param name The name of the command.
+ */
+#define COMMAND_SPEC_KEY(name) "spec:" COMMAND_BASE_KEY (name)
+
+/**
+ * @def GET_BASIC_OPTIONS
+ * @brief This macro retrieves the values of the basic options from an options key set into local variables. This has to be used in the
+ * context of a command, so GET_OPTION_KEY has to be defined.
+ */
+#define GET_BASIC_OPTIONS                                                                                                                  \
```
kodebach, line 115: Needs a little explaination what `GET_OPTION_KEY` actually should do.

```suggestion
/**
 * @def GET_BASIC_OPTIONS
 * @brief This macro retrieves the values of the basic options from an options key set into local variables.
 * This has to be used in the context of a command, so #GET_OPTION_KEY has to be defined.
 *
 * #GET_OPTION_KEY should be `#define GET_OPTION_KEY(options, name)` and evaluate to a `Key *` given a `KeySet * options` and a `char * name`.
 * 
 */
#define GET_BASIC_OPTIONS                                                                                                                  \
```

src/tools/kdb/command.h
kodebach: I marked this as resolved, since the `struct` is a nice idea, but given the fact that the other PRs use the same structure and are already done, it's something for another day.

kodebach (approved):
LGTM now, just some small improvements to the new docs.

lukashartl (approved):
lgtm

hannes99:

doc/help/kdb-complete.md
```diff
@@ -5,8 +5,7 @@
 `kdb complete [path]`
 
 Where `path` is the path for which the user would like to receive completion suggestion.
-If `path` is not specified, it will show every possible completion. It's synonymous
-to calling `kdb complete ""`.
+Calling `kdb complete ""` will show every possible completion.
```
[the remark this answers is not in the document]
hannes99: added a note

doc/help/kdb-complete.md
```diff
@@ -5,8 +5,9 @@
...
+
+_NOTE: `kdb complete` used to work, in the new `C` implementation "" has to be added_
```
markus2330, line 10: This should be release notes, not here.

doc/tutorials/external-commands.md
```diff
@@ -0,0 +1,51 @@
+# How-to: Add external commands
+
```
markus2330, line 2: Please start with some introduction what the tutorial is going to tell us.

markus2330 (changes requested):
Please describe your framework architecture: how does it work, how to implement new commands, what are the general principles/common functionalities.

markus2330:
Also important is to explain all differences in the release notes, e.g.:

```
kdb get                                                                                                                                                                                                                          
Sorry, I could not process the arguments: Need one argument

Usage: kdb get <name>

Get the value of an individual key.
When the key starts with / a cascading lookup will be done.

Example:

   kdb get system:/elektra/version/constants/KDB_VERSION

-a --all                 Consider all of the keys.
-d --debug               Give debug information or ask debug questions (in interactive mode).
-H --help                Show the man page.
-n --no-newline          Suppress the newline at the end of the output.
-p --profile <name>      Use a different profile for kdb configuration.
-v --verbose             Explain what is happening.
-V --version             Print version info.
-C --color <when>       Print never/auto(default)/always colored output.
```

vs. (in the new version)

```
kdb get                                                                                                                                                                                                                      
ERROR: Expected at least 1 non-option arguments, but only got 0
```


shouldn't come as surprise for the user.

doc/news/_preparation_next_release.md
```diff
@@ -196,11 +196,13 @@ This section keeps you up-to-date with the multi-language support provided by El
 
 ## Tools
 
-### <<Tool>>
+### KDB
 
+- add basic C framework, helper functions and new `get` implementation _(@hannes99)_
```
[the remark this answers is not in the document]
hannes99: done

doc/tutorials/external-commands.md
hannes99: done

doc/help/kdb-complete.md
```diff
@@ -5,8 +5,9 @@
 `kdb complete [path]`
 
 Where `path` is the path for which the user would like to receive completion suggestion.
-If `path` is not specified, it will show every possible completion. It's synonymous
-to calling `kdb complete ""`.
+Calling `kdb complete ""` will show every possible completion.
```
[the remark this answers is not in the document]
hannes99: done

doc/help/kdb-complete.md
hannes99: done

hannes99:

doc/news/_preparation_next_release.md
```diff
@@ -281,11 +285,25 @@ This section keeps you up-to-date with the multi-language support provided by El
...
+  - Logging levels and error/warning outputs are now handled my the new CLI framework, and not by commands individually.
```
markus2330, line 291:
```suggestion
  - Logging levels and error/warning outputs are now handled by the new CLI framework, and not by commands individually.
```

doc/news/_preparation_next_release.md
```diff
@@ -58,14 +58,16 @@ docker run -it elektra/elektra
 ### Session Recording
 
 Elektra now comes with a powerful new feature that allows users to record and export changes made to the KDB: session recording.
-This feature enables you to easily track changes made to the configuration database over time, which helps troubleshoot issues, diagnose errors, and improve system performance.
+This feature enables you to easily track changes made to the configuration database over time, which helps troubleshoot issues, diagnose
```
markus2330, line 61: Please avoid such unrelated changes. It also doesn't confirm to our documentation guidelines.

doc/news/_preparation_next_release.md
```diff
@@ -281,11 +285,25 @@ This section keeps you up-to-date with the multi-language support provided by El
...
+  - Add specification for commands written in `C++`, this enables `opts` to generate the usage message and makes sure all commands(
+    regardless of their implementation) have the same help message format and generally behave the same.
+  - _NOTE_: The `path` argument for `kdb complete` is now required, so instead of `kdb complete` `kdb complete ""` has to be used. The
+    reason for this is described in issue #4952.
 - <<TODO>>
-- <<TODO>>
+- add support for external commands, with and without spec _(@hannes99)_
+  - It is still possible to execute external binaries that are placed in a specific director, same as it worked in the old implementation.
+  - It is now possible to provide a specification to external commands to `kdb`, so `kdb` check/parses arguments and then calls the
+    external program with them.
```
markus2330, line 306: Please show how this can be done.

```suggestion
  - It is now possible to provide a specification to external commands to `kdb`, so `kdb` check/parses arguments and then calls the external program with them.
```

doc/tutorials/external-commands.md
```diff
@@ -0,0 +1,57 @@
...
+This tutorial will describe how to provide `kdb` with the specification of external programs.
+So `kdb` can parse and check the provided options and arguments according to the provided specification.
+This allows you to have, for example, a shell script but its args are checked by `kdb` before running it.
+It is possible to either mount(`kdb mount`) the specification, or set the keys manually using `kdb set` and `kdb meta-set`.
+Both options will be described in the following.
+
+For a reference of how the specification can look like [Command Line Options](command-line-options.md).
+
```
markus2330, line 10: Now you need to write what *specifically* you will do in this tutorial. It looks like this would be for something like `rm`? Please give a full example for how to write a specification for a useful script (e.g. doing `rm`).

I would prefer if it is an example for something related to Elektra.

doc/help/kdb-cache.md
```diff
@@ -19,11 +19,9 @@ subcommand can only remove a user's cache files (i.e. not system wide).
 ## OPTIONS
 
 - `-H`, `--help`:
-  Show the man page.
+  Show usage of command.
 - `-V`, `--version`:
   Print version info.
-- `-p`, `--profile <profile>`:
```
markus2330, line 25: Are profiles not supported anymore? This isn't mentioned in the readme. Please write an issue with all features that are missing.

markus2330 (changes requested):
Thx for the updates. Unfortunately, something went wrong with the rebase/formatting. There are now many irrelevant changes and I didn't find a commit that addresses my review, which made it difficult to see what changed.

src/tools/kdb/main.c
```diff
@@ -0,0 +1,322 @@
...
+		printWarnings (parentKey);
+		keyDel (parentKey);
+		ksDel (options);
+		ksDel (externalBinaries);
+		return 0;
+	}
+	if (result == -1)
+	{ // error
+		const char * errorMessage = GET_ERR (parentKey);
+		if (elektraStrNCmp (errorMessage, "Unknown sub-command:", 20) == 0)
```
atmaxinger, line 253: A bit hacky, but I guess there isn't really another option. Let's just hope, that these strings here will never get localized ...

hannes99: there is already an issue for that: #4953

doc/news/_preparation_next_release.md
hannes99: Not sure how this happened, but reverted it

doc/news/_preparation_next_release.md
hannes99: Added link to the tutorial

doc/tutorials/external-commands.md
hannes99: Added a small script that the tutorial is structured around. I couldn't really think of something that can't already be done by the normal commands.

doc/help/kdb-cache.md
hannes99: I'll mention it in the readme and add an issue(#4965). I think this was discussed a while ago, but maybe I misremember, regardless, it should be mentioned.

doc/news/_preparation_next_release.md
hannes99: done

hannes99:
The compare should work now, I mostly used `commit --fixup=`, should I have also pushed the `!fixup` commits? I felt like that would be weird in the git history.

hannes99:

hannes99:

hannes99:

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
...
+}
+```
+
+The `ADD_BASIC_OPTIONS` at the end adds options that are common across all commands to its spec, this includes things like `verbose: -v` or `debug: -d`.
+It if your command should support these, it makes sense to use the macro.
+
+#### Execution
+
+The `int exec... (KeySet * options, Key * errorKey)` receives two parameters, `options` and `errorKey`.
+Where `options` contains the passed options and arguments and `errorKey` is where any errors or warnings have to be written to so the framework can properly print them.
```
kodebach, line 88:
```suggestion
Where `options` contains the passed options (as parsed by `elektraGetOpts`) and arguments and `errorKey` is where any errors or warnings have to be written to so the framework can properly print them.
```

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
...
+
+#define GET_OPTION_KEY(options, name) GET_OPT_KEY (options, COMMAND_BASE_KEY (COMMAND_NAME) "/" name)
+#define GET_OPTION(options, name) GET_OPT (options, COMMAND_BASE_KEY (COMMAND_NAME) "/" name)
+```
+
+It should also `#include <colors.h>` and `#include <command.h>`, those contain all the needed helper macros and function.
+
+#### Specification
+
+`void add...Spec (KeySet * spec)` will contain the specification of the command new command.
```
kodebach, line 64:
```suggestion
`void add...Spec (KeySet * spec)` will contain the specification (for `elektraGetOpts`) of the new command.
```

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
...
+```c
+ELEKTRA_SET_INTERNAL_ERROR (errorKey, "could not copy key meta to errorKey");
+```
+
+These error setting macros are available in `kdberrors.h`.
+By setting the like this, they are later printed properly formatted by the framework.
+
+## 4. "registering" the new command
+
+You just have to `#include` your header file in `main.c` and add an element to the `command subcommands[]` array.
```
kodebach, line 159:
```suggestion
You just have to `#include` your header file in [`main.c`](/src/tools/kdb/main.c) and add an element to the `command subcommands[]` array.
```

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
+# Implement new commands in C
+
+Here we'll go over what has to be done in order to implement a new command in C and integrate it in to the `kdb` CLI framework.
+
+## 1. Files
+
+First we need a header as well as a c file for the new command.
```
kodebach, line 7:
```suggestion
First we need a header as well as a c file for the new command in `src/tools/kdb`.
```

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
+# Implement new commands in C
+
+Here we'll go over what has to be done in order to implement a new command in C and integrate it in to the `kdb` CLI framework.
```
kodebach, line 3:
```suggestion
Here we'll go over the basic requirements to implement a new command in C and integrate it in to the `kdb` CLI framework.
You can also look at the existing commands in `src/tools/kdb` to get an idea how various details (e.g. calling `kdbGet`/`kdbSet`) should be done.
```

kodebach (approved):
LGTM just a few small suggestions for the new docs.

hannes99:
Thanks for the feedback!

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
...
+				   "opt", "a", KEY_META, "opt/long", "all", KEY_META, "opt/arg", "none", KEY_END));
+	ksAppendKey (spec, keyNew (COMMAND_SPEC_KEY (COMMAND_NAME) "/name", KEY_META, "description", "The name of the key", KEY_META,
+				   "args", "indexed", KEY_META, "args/index", "0", KEY_END));
+
+	ADD_BASIC_OPTIONS (spec, COMMAND_SPEC_KEY (COMMAND_NAME))
+}
+```
+
+The `ADD_BASIC_OPTIONS` at the end adds options that are common across all commands to its spec, this includes things like `verbose: -v` or `debug: -d`.
+It if your command should support these, it makes sense to use the macro.
```
markus2330, line 83:
```suggestion
It if your command should support these, use the macro.
```

Are there exceptions where they should not be supported? If not remove the sentence:

```suggestion
```

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
...
+Available logging levels are `CLI_LOG_NONE`, `CLI_LOG_VERBOSE` and `CLI_LOG_DEBUG`.
+By using `CLI_PRINT` or `CLI_ERROR_PRINT` logging levels are handled automatically.
+
+## 3. Errors/Warning
+
+If errors occur during the execution of your command, they should be set in `errorKey` and not printed directly.
+This could look something like this:
+
+```c
+ELEKTRA_SET_INTERNAL_ERROR (errorKey, "could not copy key meta to errorKey");
```
markus2330, line 151: Doing like this we don't know which errors are from Elektra and which are from CLI? Probably we should provide ELEKTRA_SET_CLI_ERROR?

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
...
+```c
+command subcommands[] = {
+	...
+	{ "your-new-command-name", add...Spec, exec... },
+};
+```
+
+## _5. Validate_
+
+After following these stape the new command will show up when typing `kdb --help` and `kdb <your-cmd-name> --help` will show a usage message based on the specification you have provided in the `add...Spec` function.
```
markus2330, line 170: stape?

doc/tutorials/external-commands.md
```diff
@@ -0,0 +1,121 @@
...
+This tutorial will describe how to provide `kdb` with the specification of external programs.
+So `kdb` can parse and check the provided options and arguments according to the provided specification.
+This allows you to have, for example, a shell script but its args are checked by `kdb` before running it.
+It is possible to either mount(`kdb mount`) the specification, or set the keys manually using `kdb set` and `kdb meta-set`.
+Both options will be described in the following.
+Here we'll define the specification for a simple script that removes files by moving them to a trash folder instead of directly deleting them.
+
+Since the arguments are already checked by `kdb`, the script knows these two are true
+
+1. The script can assume that if `$#` is `1`, `$1` has to be the file.
```
markus2330, line 12: which file?

doc/tutorials/external-commands.md
```diff
@@ -0,0 +1,121 @@
...
+2. And if `$#` is not `1` it hast to be `2`, with `$1` being the `-f` flag and `$2` being the filename.
```
markus2330, line 13: hast?

doc/tutorials/external-commands.md
```diff
@@ -0,0 +1,121 @@
...
+
+if [ $# -eq 1 ]; then
+    # 1. was only called with the filename
+    mkdir -p $HOME/.trash
+
+    mv $1 $HOME/.trash
+    echo moved $1 to trash
+else
+    # 2. was called with -f flag and the filename
+    rm -f $2
```
markus2330, line 28: spaces in $2?

```suggestion
    rm -f "$2"
```

markus2330 (changes requested):
Tutorials are much better now, still quite a few formatting problems etc. (I only marked a few **exemplary**, please check for others yourself.)

doc/dev/kdb-add-new-command.md
```diff
@@ -0,0 +1,170 @@
...
+
+### `C` file
+
+The `<name>.c` to implement the two functions already described in the header file, in addition to those two functions `COMMAND_NAME`, `GET_OPTION_KEY` and `GET_OPTION` have to be defined as follows:
+
+```c
+#define COMMAND_NAME "..." // your command name
+
+#define GET_OPTION_KEY(options, name) GET_OPT_KEY (options, COMMAND_BASE_KEY (COMMAND_NAME) "/" name)
+#define GET_OPTION(options, name) GET_OPT (options, COMMAND_BASE_KEY (COMMAND_NAME) "/" name)
```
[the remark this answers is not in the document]
hannes99: makes sense, moved them to `command.h` and updated the docs since now `#define COMMAND_NAME` has to be before `#include <command.h>`.

hannes99:
Thanks a lot for the feedback @markus2330 !

hannes99:

hannes99:

PR #4951, opened 2023-05-27
Kdb rewrite external

hannes99:
## Basics

  - make all unit tests pass
  - fix all memleaks

## Checklist

      (not in the PR description)

## Review

## Labels


src/tools/kdb/external.c
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+	map->size = 0;
+}
+
+int tryLoadExternal (char * commandName, StringMap * map)
+{
+	char * execPathPtr = getenv ("KDB_EXEC_PATH");
+	bool found = false;
+	char path[1024] = { 0 };
```
kodebach, line 79: If you're not using `malloc` to support arbitrary path lengths, please use

```suggestion
	char path[PATH_MAX] = { 0 };
```

It's still possible that paths are longer than `PATH_MAX`, but it's at least better than a magic number

src/tools/kdb/external.c
```diff
@@ -0,0 +1,205 @@
...
+
+	if (!found)
+	{
+		snprintf (path, sizeof (path), "%s/%s", BUILTIN_EXEC_FOLDER, commandName);
+		found = stat (path, &buf) != -1;
+	}
+
+	if (found)
+	{
+		insert (map, elektraStrDup (commandName), strdup (path));
```
kodebach, line 103: As an example, with a `KeySet * map`, this would be:

```c
Key * k = keyNew ("/", KEY_VALUE, path, KEY_END);
keyAddBaseName (commandName);
ksAppendKey (map, k);
```

src/tools/kdb/external.c
```diff
@@ -0,0 +1,205 @@
...
+		cur = ksAtCursor (part, it);
+		const Key * externalMeta = keyGetMeta (cur, "external");
+		const Key * externalBinary = keyGetMeta (cur, "bin");
+		bool isExternal = false;
+		if (externalBinary != NULL && externalMeta != NULL && elektraKeyToBoolean (externalMeta, &isExternal) && isExternal)
+		{ // add external spec and save path to binary
+			KeySet * externalCommandSpec = ksCut (part, cur);
+			insert (map, elektraStrDup (keyBaseName (cur)), elektraStrDup (keyString (externalBinary)));
+			ksAppend (spec, externalCommandSpec);
+			ksDel (externalCommandSpec);
```
kodebach, line 136:
```suggestion
			KeySet * externalCommandSpec = ksCut (part, cur);
			insert (map, elektraStrDup (keyBaseName (cur)), elektraStrDup (keyString (externalBinary)));
			ksAppend (spec, externalCommandSpec);
			ksDel (externalCommandSpec);
			--it;
```

Without the `--it` you'd be skipping a key because of `ksCut` (removes `cur` from `part`), I think.

src/tools/kdb/external.c
```diff
@@ -0,0 +1,205 @@
...
+
+	pid_t pid;
+	int status = 0;
+	posix_spawnattr_t attr;
+	posix_spawn_file_actions_t fact;
+
+	posix_spawnattr_init (&attr);
+	posix_spawn_file_actions_init (&fact);
+
+	if (posix_spawn (&pid, bin, &fact, &attr, &(argv[1]), __environ) != 0)
```
kodebach, line 159: I like that we `fork` now and even avoid `fork` itself by using `posix_spawn`. Not sure, it's actually needed (a simple `execve` probably would be enough). However, is that really supported on Windows/MinGW? You're already switching to the Win32 API below, so you may as well do the entire process creating with the Win32 API.

doc/tutorials/external-commands.md
kodebach, line 1: 1. AFAIK you don't actually enforce an absolute path anywhere. You just pass the path to `posix_spawn`, which will search `PATH` if the path contains no `/`. That should probably be explained here. 
2. You should explain that external commands can be added without a spec too.
3. For commands with spec you should explain, that the spec is used to validate `argv`, but the results of `elektraGetOpts` aren't used. The raw `argv` is always passed to external commands (with or without spec).

kodebach (approved):
Very nice. Assuming you fix the CI errors, this can be merged IMO. The docs could be better and the new map could just be a `KeySet`, but nothing major. I really like that we can now have specifications for external commands, but it is still entirely optional.

doc/tutorials/external-commands.md
kodebach: Also I'm not sure `elektraGetOpts` enforces a maximum size of `argv`. So if it's possible that `argv` has more elements than defined in the spec, then that should also be clarified here. Otherwise external commands with spec could be surprised by extra arguments

src/tools/kdb/external.c
hannes99: I wanted to keep as much code as possible shared. It was about waiting for it, so error printing/cleanup is still done by the CLI.

doc/tutorials/external-commands.md
hannes99: 1. Actually yes, I'll add that it should be an absolute path. (And mention that it could be a relative one)
2. I think it makes sense that the minimal thing necessary is setting the meta keys `external` and `bin` on `spec:/sw/elektra/kdb/#0/current/CMD_NAME`, so it shows up in --help.
3. So you mean `opts` stops checking `argv` after 100 or so? I'd assume if it detects more than it supports it fails somehow, no?

kodebach: > 2\. I think it makes sense that the minimal thing necessary is setting the meta keys `external` and `bin` on `spec:/sw/elektra/kdb/#0/current/CMD_NAME`, so it shows up in --help.

Yes, definitely better than `KDB_EXEC_PATH` I think. I originally thought that is still supported, but it seems not to be the case.

To show how simple it really is you should probably add an example like:

```c
kdb meta-set spec:/sw/elektra/kdb/#0/current/echo bin /usr/bin/echo
kdb meta-set spec:/sw/elektra/kdb/#0/current/echo external 1
```

You could even simplify the schema to only require:

```c
kdb meta-set spec:/sw/elektra/kdb/#0/current/echo external/bin /usr/bin/echo
```

> So you mean opts stops checking argv after 100 or so? I'd assume if it detects more than it supports it fails somehow, no?

A specification like this

```ni
[file]
 meta:/description = the file that shall be deleted
 meta:/args = indexed
 meta:/args/index = 0
[]
 meta:/command = somecommand
 meta:/description = Simple script
 meta:/external = 1
 meta:/bin = /path/to/test.sh
```

might look like the external command will only be called with a single argument. That's not the case AFAICT. `elektraGetOpts` will not throw an error, if more than one argument is passed (they'll just be ignored). So because you just pass the entire `argv` to the external binary, it may receive more than one argument. In 99% of cases that won't matter, but just in case, the docs should be clear how external binaries are called.

hannes99: No, it is supported, and it would work if a binary is put in that directory. I just don't wanna mention it explicitly, it's more of a backwards compat thing.

kodebach:
Removing `KDB_EXEC_PATH` will also break this part of the `run_env` script:

https://github.com/ElektraInitiative/libelektra/blob/83af03b69cb4d663db6d932ad9f3d6dcf623fa58/scripts/dev/run_env#L26-L30

Also the `kdb list-tools` command will have to be written in C.

doc/tutorials/external-commands.md
hannes99: > might look like the external command will only be called with a single argument. That's not the case AFAICT.

Just tried it, and it is actually fine with it and ignores it, I thought it would complain, but it doesn't. I'll mention that in the docs. Thanks!

kodebach: > No, it is supported, and it would work if a binary is put in that directory. I just don't wanna mention it explicitly, it's more of a backwards compat thing.

Ah yes that also makes sense. I'd still show an example with `kdb set` too, because `kdb mount` always seems like a lot of effort.

hannes99: done

hannes99:

src/tools/kdb/external.c
```diff
@@ -0,0 +1,195 @@
...
+	si.cb = sizeof (si);
+	ZeroMemory (&pi, sizeof (pi));
+
+	// Construct command line string
+	char cmdline[MAX_PATH] = "";
+	for (int i = 1; argv[i]; ++i)
+	{
+		strcat (cmdline, argv[i]);
+		strcat (cmdline, " ");
+	}
```
kodebach, line 147: I'm not sure this would actually work, if `argv[i]` contains a space.

However, I think you can just do:

```suggestion
	char cmdline[MAX_PATH] = "";
	for (int i = 1; argv[i]; ++i)
	{
		strcat (cmdline, "\"");
		strcat (cmdline, argv[i]);
		strcat (cmdline, "\" ");
	}
```

kodebach (changes requested):
Small issue with spaces on Windows, I think

